PR #4601, opened 2020-05-06
Fix intial agglomerate loading

### LINKED ISSUE #4599: Backend reports existing buckets as missing when activating an agglomerate file mapping
Most of the time when I activate an agglomerate file mapping, the segmentation buckets in the middle of the viewports never appear.

agglobug

I checked the requests and it looks like the buckets in the very first requests are always marked as missing by the backend, although they actually exist. This can be seen when re-activating the same mapping as suddenly the buckets are no longer reported as missing and appear in the viewport.

First Request Payload:
```json
[
  {"position":[64,64,64],"zoomStep":0,"cubeSize":32,"fourBit":false,"applyAgglomerate":"agglomerate_view_35"},
  {"position":[96,64,64],"zoomStep":0,"cubeSize":32,"fourBit":false,"applyAgglomerate":"agglomerate_view_35"},
  {"position":[96,96,64],"zoomStep":0,"cubeSize":32,"fourBit":false,"applyAgglomerate":"agglomerate_view_35"}
]
```
First Response: missing-buckets: [0, 1, 2]

Second Request Payload (2/3 buckets are the same as in the first request):
```json
[
  {"position":[64,64,64],"zoomStep":0,"cubeSize":32,"fourBit":false,"applyAgglomerate":"agglomerate_view_35"},
  {"position":[64,96,64],"zoomStep":0,"cubeSize":32,"fourBit":false,"applyAgglomerate":"agglomerate_view_35"},
  {"position":[96,64,64],"zoomStep":0,"cubeSize":32,"fourBit":false,"applyAgglomerate":"agglomerate_view_35"}
]
```
Second Response: missing-buckets: []

youri-k:
- https://fixagglomeratecache.webknossos.xyz

### Steps to test:
- load a new agglomerate mapping
- no black buckets should appear

### Issues:
- fixes #4599 

------
- [x] Ready for review


daniel-wer:
Thanks for the quick fix @youri-k :pray:

youri-k:
@fm3 Tbh I'm not entirely sure, why the different position of the `finishAccess()` caused the bug. However, my theory is the following:
The first bucket with a new mapping is requested and the agglomerate file is not in cache. The agglomerate file is loading and the Future is not completed. When the Future is completed, there are other buckets that request the same future, which is now completed, so the backend responds to those immediately. When all other requests are finished, the original first requests get processed. However, when they try to read from the hdf5 file, it is already closed because the other reads were already performed and the `finishAccess()` call was too early. 

The only thing, I don't know is, why the file was closed after the last read because the cache should be large enough to retain it and not call `removeElement()`.

fm3:
> The first bucket with a new mapping is requested and the agglomerate file is not in cache. The agglomerate file is loading and the Future is not completed. When the Future is completed, there are other buckets that request the same future, which is now completed, so the backend responds to those immediately. When all other requests are finished, the original first requests get processed. However, when they try to read from the hdf5 file, it is already closed because the other reads were finished and the `finishAccess()` call was too early.

Okay, that sounds about right. Good catch!


> The only thing, I don't know is, why the file was closed after the last read because the cache should be large enough to retain it and not call `removeElement()`.

Maybe we can add some logging to see if items get removed from the cache earlier than we expect? This could have performance implications. That’s independent of this PR, though.

PR #4602, opened 2020-05-07
Save Compressed Volume Tracing Buckets 

### LINKED ISSUE #4592: Use lz4 compression in volume tracing
…to save load on tracing store server / fossildb
(check first that this is not already done)
volume buckets are saved with wkw headers, right? so we can probably just use wkw compression, and not create backwards incompatibility

fm3:
- Compress volume tracing buckets with lz4 before writing them into fossildb.
- When loading buckets, only buckets that are shorter than their expected uncompressed bucket length are assumed to be compressed, for backwards compatibility
- When writing buckets, only buckets that actually get shorter by compressing them are written compressed
- Compression factor for my test tracings was about 100x for sparse annotation (single cell) and 6x–12x for dense annotation (based on L4 oversegmentation)

### Questions
 - [x] Is the perf impact noticable? (speed up due to less db load, slowdown due to compression/decompression routines) → compression and decompression incl detection each took about 200–300 µs per bucket in my tests, which I’d consider acceptable. speedup due to less I/O difficult to measure in small setting, large variance in put times, but I assume we can expect some.
 - [x] Do 64 bit tracings need special handling? → 64 bit tracings are not allowed in wk, added limit to avoid overestimating expected bucket size
 - [x] How to mark which buckets in the db are compressed. There are several competing design ideas (This PR implements B (size), and there is a draft for D (key suffix) here https://github.com/scalableminds/webknossos/pull/4610)
  - A: Prefix in the data. Prefix the bytes with a fixed string, check for that string when loading.
     - ⁺ No changes to db loading abstraction levels
     - ⁻ Prefix could have collisions with existing data
     - ⁻ Introducing an optional header feels somewhat hacky
     - ⁻ Every bucket grows by a few bytes
  - B: Size. Calculate expected bucket size from data type and bucket dimensions. If the returned byte array is shorter, decompress it. (when saving, only save compressed version if it is actually shorter than that number)
     - ⁺ No additional I/O, no additional bytes
     - ⁺ Implemented in a few easily readable lines
     - ⁻ Stored information is very implicit, feels somewhat hacky
  - C: Additional boolean column family in fossildb, always fetch both, if the boolean is set, decompress
     - ⁺ Explicit marking of compressed buckets
     - ⁻ Need to fetch second column family key for every bucket request
  - D: Make use of alphabetical key sorting in rocksdb and use suffix for compressed bucket keys. When fetching, check for duplicates and discard the older version.
     - ⁺ Explicit marking of compressed buckets
     - ⁻ Need to use range queries for every bucket request
     - ⁻ Requests for tracings where old (uncompressed) buckets were overwritten by new (compressed) buckets will return both the newest compressed and the newest uncompressed version, superfluous I/O (Also for volume range queries e.g. for download). It may be argued that those tracings will become rare over time.
  - E: Create fossildb migration, compressing all existing buckets, so no distinction needs to be made
    - ⁺ Clean state of db, no distinction code
    - ⁻ Requires downtime, for duration of the migration requests can’t distinguish buckets.
    - ⁻ Writing fossildb migrations is somewhat tedious

### Follow-Ups:
 - [ ] Change Front-end/Back-end communication, should the frontend compress before sending? how about get requests? (future issue)
 - [ ] Change format of zip downloads, should we write lz4 buckets to zip? (future issue)

- https://___.webknossos.xyz

### Steps to test:
- open old volume tracing, should still load fine
- edit it, reload, should still load fine (new buckets using compression, the old ones not)
- upload/download/copy-to-my-account of volume tracings should still work

### Issues:
- fixes #4592 

------
- [x] Needs tracing update after deployment
- [x] Ready for review

### DIFF
--- a/webknossos-tracingstore/app/com/scalableminds/webknossos/tracingstore/tracings/volume/VolumeTracingBucketHelper.scala
+++ b/webknossos-tracingstore/app/com/scalableminds/webknossos/tracingstore/tracings/volume/VolumeTracingBucketHelper.scala
@@ -3,17 +3,12 @@ package com.scalableminds.webknossos.tracingstore.tracings.volume
 import com.scalableminds.util.geometry.Point3D
 import com.scalableminds.util.tools.{Fox, FoxImplicits}
 import com.scalableminds.webknossos.datastore.models.BucketPosition
-import com.scalableminds.webknossos.datastore.models.datasource.DataLayer
-import java.nio.ByteBuffer
+import com.scalableminds.webknossos.datastore.models.datasource.{DataLayer, ElementClass}
 
-import com.scalableminds.webknossos.tracingstore.tracings.{
-  FossilDBClient,
-  KeyValueStoreImplicits,
-  VersionedKeyValuePair
-}
+import com.scalableminds.webknossos.tracingstore.tracings.{FossilDBClient, KeyValueStoreImplicits, VersionedKeyValuePair}
 import com.scalableminds.webknossos.wrap.WKWMortonHelper
 import com.typesafe.scalalogging.LazyLogging
-import net.jpountz.lz4.{LZ4Compressor, LZ4Decompressor, LZ4Factory, LZ4FastDecompressor}
+import net.jpountz.lz4.{LZ4Compressor, LZ4Factory, LZ4FastDecompressor}
 
 import scala.concurrent.ExecutionContext.Implicits.global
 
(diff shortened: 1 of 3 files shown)


fm3:
@youri-k do you have opinions on the backwards compatibility? (or any other ideas, really)

webknossos-tracingstore/app/com/scalableminds/webknossos/tracingstore/tracings/volume/VolumeTracingBucketHelper.scala
```diff
@@ -3,17 +3,54 @@ package com.scalableminds.webknossos.tracingstore.tracings.volume
...
+trait VolumeBucketCompression extends LazyLogging {
+
+  private val lz4factory = LZ4Factory.fastestInstance
+  val compressor: LZ4Compressor = lz4factory.fastCompressor
+  val decompressor: LZ4FastDecompressor = lz4factory.fastDecompressor
+
+  def compressVolumeBucket(data: Array[Byte]): Array[Byte] = {
+    val compressedData = compressor.compress(data)
+
+    logger.info(s"${data.length} -> ${compressedData.length}")
```
fm3, line 28: note to self: remove logging before merge

PR #4604, opened 2020-05-07
Add finer control over interval range in timetracking view

### LINKED ISSUE #4581: allow specifying start and end time in time tracking view
The highest “zoom level” is one day. if you want to inspect the time tracking for short tasks, it would be nice to look at a shorter time frame in more detail

MichaelBuessemeyer:
This PR adds a finer control to the time interval in the timetracking view.

- https://finertimetrackingviewinterval.webknossos.xyz

### Steps to test:
- Trace a task.
- Open the timetracking view and select the last two hours.
- The requested and displayed interval should only show those two hours.

### Issues:
- fixes #4581

------
- [x] Ready for review


frontend/javascripts/admin/time/time_line_view.js
```diff
@@ -300,6 +300,7 @@ class TimeLineView extends React.PureComponent<Props, State> {
               </FormItem>
               <FormItem {...formItemLayout} label="Date">
                 <RangePicker
+                  showTime
```
daniel-wer, line 303: I would propose to disable the minutes and seconds if that makes the ui less complicated. According to the antd docs you can supply an object to `showTime` which will be passed on to the TimePicker component (https://ant.design/components/time-picker/#API). Does that work for our antd version and if so does it make the UI simpler?

MichaelBuessemeyer: Sadly the less complicated version seems not to be available in our antd version. We would need to upgrade antd for this. On the newest node version the `RangePicker` could look like [this](https://codesandbox.io/s/oz6iv).

I choose to keep the option to select minutes and removed only the seconds as the selection of minutes gives the possibility for detailed insight into the tracked time of users. 
Additionally only having hours looks pretty awful in our antd version

Issue #123, opened 2014-01-03
Verify link on user list is not working


tmbo:
In the user list, the verify link is not working. Normally there should be a popup to select the team of the user, but currently nothing happens after a click.

<a href="https://timer.scm.io/repos/537237ce1a00001a003c4a14/issues/123/create?referer=github" target="_blank">Log Time</a>

PR #4611, opened 2020-05-12
[Discarded] NodeJS NML Parser

fm3:
Replaces the backend scala-based NML parser by calling nodejs from scala, to run the JS one.

Edit Daniel:
_This PR contains code that makes it possible to reuse frontend javascript code in the backend by declaring an additional webpack target which extracts part of the frontend code in a separate bundle. This bundle can then be called via nodejs from the backend._
_We ultimately decided against merging this PR, but the proof of concept may be useful in the future :)_

- https://___.webknossos.xyz

### Steps to test:
- abc

### Issues:
- fixes #123

------
- [ ] Ready for review


philippotto:
@fm3 Can this be closed? I'm wondering since the title says `Discarded`? :)

fm3:
fine by me, but I’m not sure how to deal with this. @daniel-wer what do you think, does it make sense to keep this around?

daniel-wer:
I would also suggest to close this. I'll extend the PR description to make it easier to find this if we ever plan to reuse code from the frontend in the backend again.

Issue #4572, opened 2020-04-27
Clean up console logs


fm3:
We should check whether all console logging is useful. Since users are instructed to check those /send the content, it would for example be helpful if the missing sharing token would not appear as error all the time (since this is expected behavior)

PR #4612, opened 2020-05-12
Don't render ActionBarView during loading

### LINKED ISSUE #4597: Hide most action bar entries during initialization
For example, the position input box is orange during initialization. Instead, it should be hidden.

philippotto:
This kills two birds with one stone: The action bar view doesn't show incorrect data during loading anymore (e.g., invalid position). Additionally, the sharing modals are not instantiated before loading which avoids trying to load sharing tokens for the dummy dataset which the store holds initially (which always fails).

- https://avoidtokenerrors.webknossos.xyz

### Steps to test:
- load a dataset or tracing
- ensure that the action bar is there after loading
- check the network tab that no bogus sharing token requests were performed

### Issues:
- fixes #4597
- contributes to #4572

------
- [X] Ready for review

### DIFF
--- a/frontend/javascripts/oxalis/view/right-menu/dataset_info_tab_view.js
+++ b/frontend/javascripts/oxalis/view/right-menu/dataset_info_tab_view.js
@@ -287,8 +287,8 @@ class DatasetInfoTabView extends React.PureComponent<Props> {
 
     return (
       <div className="flex-overflow">
-        <p>{annotationTypeLabel}</p>
-        <p>{descriptionEditField}</p>
+        <div>{annotationTypeLabel}</div>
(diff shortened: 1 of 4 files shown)


frontend/javascripts/oxalis/view/right-menu/dataset_info_tab_view.js
```diff
@@ -287,8 +287,8 @@ class DatasetInfoTabView extends React.PureComponent<Props> {
 
     return (
       <div className="flex-overflow">
-        <p>{annotationTypeLabel}</p>
```
philippotto, line 290: This produced a warning because `p` tags shouldn't contain `div`s.

PR #4619, opened 2020-05-14
Avoid double info request for compound annotations

### LINKED ISSUE #4616: View button for task instance does not work properly
> Wenn ich nach einem Task suche und dann über "view" mir alle fertigen Instanzen angucken will, dann lande ich seit dem Update immer auf einer Seite mit dem Hinweis "Either the dataset does not exist or you do not have the necessary access rights". Wenn ich die Seite refreshe, dann kann ich mir die Tracings normal angucken.

fm3:
Compound Annotations are created lazily, and then cached.
The frontend checks the permissions for logged-out users for an annotation in parallel with loading it. That’s why two info requests are sent.
If the second info request fires before the first one has completed, the compound annotation is created twice, and only one of the two will be in the (key-value) cache. That is the race condition.
They will have different skeleton tracing IDs.
If the frontend then assumes the wrong skeleton ID, the reverse lookup for permission check fails.

This PR changes the frontend to only check for logged-out users *if the user is logged out*, avoiding this double info request.

- https://fixannotationcache.webknossos.xyz

### Steps to test:
- view compound tasks (of at least two instances), should not fail, network tab should show only one info request
- viewing public annotations (on public datasets) while logged out should still work

### Issues:
- fixes #4616

------
- [x] Ready for review

### DIFF
--- a/frontend/javascripts/messages.js
+++ b/frontend/javascripts/messages.js
@@ -76,6 +76,8 @@ In order to restore the current window, a reload is necessary.`,
   "data.disabled_render_missing_data_black": `You just disabled the option to render missing
 data black. This means that in case of missing data, data of lower quality is rendered
 instead. Only enable this option if you understand its effect. All layers will now be reloaded.`,
+  "tracing.unhandled_initialization_error":
+    "Unfortunately, an error occurred during initialization. Please reload the page and try again. This might fix the error.",
   "tracing.out_of_dataset_bounds":
     "The current position is outside of the dataset's bounding box. No data will be shown here.",
   "tracing.out_of_task_bounds": "The current position is outside of the task's bounding box.",
--- a/frontend/javascripts/messages.js
+++ b/frontend/javascripts/messages.js
@@ -77,7 +77,7 @@ In order to restore the current window, a reload is necessary.`,
 data black. This means that in case of missing data, data of lower quality is rendered
 instead. Only enable this option if you understand its effect. All layers will now be reloaded.`,
   "tracing.unhandled_initialization_error":
-    "Unfortunately, an error occurred during initialization. Please reload the page and try again. This might fix the error.",
+    "Initialization error. Please refresh the page to retry. If the error persists, please contact an administrator.",
   "tracing.out_of_dataset_bounds":
     "The current position is outside of the dataset's bounding box. No data will be shown here.",
   "tracing.out_of_task_bounds": "The current position is outside of the task's bounding box.",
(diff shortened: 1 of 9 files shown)


MichaelBuessemeyer:
@fm3 As we discussed I tried to make the error handling while loading the tracing better by displaying unhandled error messages better to the user:
Screenshot_20200514_212750

I'll push the changes for this to this branch

philippotto:
> Compound Annotations are created lazily, and then cached.
> If the second info request fires before the first one has completed, the compound annotation is created twice, and only one of the two will be in the (key-value) cache. That is the race condition.
> They will have different skeleton tracing IDs.
> If the frontend then assumes the wrong skeleton ID, the reverse lookup for permission check fails.

I don't really understand where these two info requests come from. I guess, one of them is `getAnnotationInformation` in the `<SecuredRoute />` part which is also changed in this PR? Which info request is executed in parallel to that?

I also don't understand why returning `false` in `serverAuthenticationCallback` helps in case of compound annotations. :thinking: Wouldn't this show an unauthorized error if the race condition goes bad?

Sorry, I'm a bit lost here :see_no_evil: I understand the bug from the description in your PR, but I don't understand how this matches to the code diff.

fm3:
Ok I guess the real question is: why is this auth callback code executed even if the user is logged in. My guess is that it should handle logged-out users.
It executes this info request, and in parallel the model initialization already does its own.
The diff in this PR prevents this auth check for the compound annotations (for they are always non-public and the get shouldn’t be necessary to determine that) but I suppose the real issue is elsewhere. I’ll have another look on monday, trying to figure out what this callback mechanism is supposed to do and what it actually does.

fm3:
Apparently, the serverAuthenticationCallback is *always* executed for the routes that have it. The reply is then **OR**ed with the normal isAuthenticated. So this PR should work correctly. @daniel-wer do you think we should change the design in SecuredRoute completely so that the serverAuthenticationCallback is only called if isAuthenticated is false? Or do you think this solution is ok?

frontend/javascripts/messages.js
```diff
@@ -76,6 +76,8 @@ In order to restore the current window, a reload is necessary.`,
   "data.disabled_render_missing_data_black": `You just disabled the option to render missing
 data black. This means that in case of missing data, data of lower quality is rendered
 instead. Only enable this option if you understand its effect. All layers will now be reloaded.`,
+  "tracing.unhandled_initialization_error":
+    "Unfortunately, an error occurred during initialization. Please reload the page and try again. This might fix the error.",
```
fm3, line 80:
```suggestion
    "Initialization error. Please refresh the page to retry. If the error persists, please contact an administrator.",
```

daniel-wer:
@fm3 Thanks for looking and digging into this! I would say the more general and cleaner solution would be to only call the serverAuthenticationCallback if isAuthenticated is false. This should lead to fewer unnecessary requests, avoids the race condition in question and doesn't rely on assumptions of routes that are always non-public. It should be enough to change line 32 in `secured_route.js` to `if (!this.props.isAuthenticated && this.props.serverAuthenticationCallback != null) {` (untested) and revert the changes in the router.

Issue #4594, opened 2020-05-05
Very long-running save requests cause problems


philippotto:
It looks like there have been very long running save requests (longer than 15 minutes?) which cause problems for users.

- [x] front-end should communicate slow save requests (see https://github.com/scalableminds/webknossos/pull/4593)
- [x] why does the back-end take so long for save requests (could https://github.com/scalableminds/webknossos/issues/4592 help?)

PR #4623, opened 2020-05-18
Compress Volume Buckets Already in Frontend

fm3:
- https://lzfrontend.webknossos.xyz

### Steps to test:
- Do some volume tracing, save, reload, buckets should load correctly
- download + upload

### Issues:
- contributes to #4594 

------
- ~[ ] Updated [(unreleased) migration guide](../blob/master/MIGRATIONS.unreleased.md#unreleased) if applicable~
- ~[ ] Updated [documentation](../blob/master/docs) if applicable~
- ~[ ] Adapted [wk-connect](https://github.com/scalableminds/webknossos-connect) if datastore API changes~
- [x] Ready for review


fm3:
I’m at a bit of a loss here why the frontend and backend lz4 implementations seem to be incompatible. the frontend-encoded data raises exceptions when decoding in the backend. I can also see in the base64/hex output that the content is different. Any ideas appreciated @youri-k

frontend/javascripts/oxalis/model/bucket_data_handling/wkstore_adapter.js
```diff
@@ -187,6 +188,12 @@ export async function sendToStore(batch: Array<DataBucket>): Promise<void> {
...
     const byteArray = new Uint8Array(data.buffer, data.byteOffset, data.byteLength);
+    console.log("sendToStore");
+    const compressed = lz4.compress(byteArray);
+    // eslint-disable-next-line no-await-in-loop
+    const compressedBase64 = await byteArrayToBase64(compressed);
+    console.log("frontend uncompressed length: ", byteArray.length, " bytes");
+    console.log("frontend compressed (", compressed.length, " bytes): ", compressedBase64);
     // eslint-disable-next-line no-await-in-loop
     const base64 = await byteArrayToBase64(byteArray);
     items.push(updateBucket(bucketInfo, base64));
```
youri-k, line 199: The only thing I noticed is, that the compressed bucket doesn't get used here. Or did you just compare the encoded base64 strings?

fm3:
Got it now :) LZ4 comes in two flavors (framed block stream and single block). The backend expected the single block variant, whereas lz4js generated a framed block stream (with a single block inside). I changed the frontend so that it now also uses the frame-less variant. This works for me now :)

frontend/javascripts/oxalis/model/bucket_data_handling/wkstore_adapter.js
fm3: exactly, that was just my intermediate testing code to compare the outputs :)

fm3:
I’m unsure where to place the `compressLz4Block` function as I’m not familiar with the frontend code structure. I’m also not sure how to adapt the failing `sendToStore: Request Handling should send the correct request parameters` test. Could you have a look at that? 🙈

daniel-wer:
I did some (unscientific) tests regarding request sizes that I wanted to document here. I opened a volume tracing with a fallback layer and set a large brush size. Then, I repeatedly:
- clicked to make a single circular brush paint
- saved

In between, I hit save without brushing, where I only moved a little bit (these are the smaller requests in each dataseries).


gzip and lz4 | lz4 | gzip | uncompressed
---|---|---|---
 |  |  | 

So one can see that requests may become slightly larger with the lz4 compression step, but the amount of data the backend needs to handle after un-gzipping the request data is greatly reduced. One can also see that we should keep gzip-compressing the save requests (factor 3 smaller request size).

fm3:
Interesting! So the bucket requests (at least for single-brush-click buckets) are about 40% larger with this PR. That’s more than I expected. I think we should try to figure out how much of an impact this has for the users. But still, as the best guess for the volume saving performance bottleneck is the server for now, I’d guess we should merge this anyway and if we find that we were wrong about this assumption, we can still undo it, as no incompatible data is created. Do you agree @daniel-wer ?

PR #4630, opened 2020-05-27
Add possibility to edit min and max of intensity range

### LINKED ISSUE #4526: Expose min/max for float layers
With float layers the histogram is necessary for any viewing of the dataset. Sometimes the sampling does not yield sensible results. For these datasets, it would be nice to specify min and max with an input field. Please note that the range of floats can be both very small (e.g. 4.8e-15) and very large (e.g. 3.2e+12).

MichaelBuessemeyer:
This PR adds the possibility for the user to edit the minimum and maximum possible value for the intensity range. 
The use case is described in #4526.

- https://addminmaxediting.webknossos.xyz

### Steps to test:
- Open a dataset (view mode is enough)
- Toggle the edit function for min and max. The button is on the left of the reload-layer button.
- enter some desired values
- the histogram and the `intensityRange` slider's range should adjust to the entered values.
- then adjust the `intensityRange` and you should be able to see something again :)

- please try breaking the feature by entering some random and uncommon values

### Issues:
- fixes #4526

------
- [x] Ready for review


MichaelBuessemeyer:
Thanks a lot for finding those bugs :bug:. Hopefully, I fixed them all. Could you please check that again?

>I'm not sure about the scientific notation feature. Try entering a large max value like 400123 by hand - you end up with 4.0012e+43 which is not what you wanted to type. It's confusing that the value changes while you type. If we still want the scientific notation feature to avoid overly long numbers, I would suggest to add a mechanism to confirm the entered number by clicking on a button. Then, the number could change to scientific notation. However, I don't think this is very important.


Thus we should not support displying in scientific notation? Is that what you meant?

daniel-wer:
> Thus we should not support displying in scientific notation? Is that what you meant?

Yes, I don't think it is very important, and I don't see an easy way to implement it with good UX, so I wouldn't add it for now.

frontend/javascripts/oxalis/view/settings/histogram_view.js
```diff
@@ -138,8 +138,10 @@ class Histogram extends React.PureComponent<HistogramProps> {
     }
     ctx.stroke();
     ctx.closePath();
-    activeRegion.lineTo(((intensityRangeMax - minRange) / fullLength) * canvasWidth, 0);
-    activeRegion.lineTo(((intensityRangeMin - minRange) / fullLength) * canvasWidth, 0);
+    const activeRegionRightLimit = Math.min(histogramMax, intensityRangeMax);
+    const activeRegionLeftLimit = Math.max(histogramMin, intensityRangeMin);
+    activeRegion.lineTo(((activeRegionRightLimit - minRange) / fullLength) * canvasWidth, 0);
+    activeRegion.lineTo(((activeRegionLeftLimit - minRange) / fullLength) * canvasWidth, 0);
```
daniel-wer, line 144: @MichaelBuessemeyer My assumption would be that `histogramMax` is always <= `intensityRangeMax` and therefore, it should be sufficient to simply use `histogramMax` in the `activeRegion.lineTo` call. Is that assumption correct?

daniel-wer: Maybe I was confusing the `histogramMax` and `intensityRangeMax`. Hmm, could you distinguish what each of those is again for me? Maybe we can find more descriptive names that make the difference more clear.

MichaelBuessemeyer: I think you really confused both of them:
- `histogramMax` is the maximum color value that is included in the histogram data send by the backend.
- `intensityRangeMax` is the maximum color value of the displayed color range and is equal to the higher value of the slide. What's new now is that this value can be greater than `histogramMax` as you can see it in your example:
max_range_float

To be honest I think the names are pretty good as the `intensityRange` is the range given by the slider and thus `intensityRangeMax` as the upper limit of that range is not far-fetched (in my opinion) -> Feel free to argue about that.

The same goes for `histogramMax` as this is the highest color value included in the histogram data.

daniel-wer: Ah understood, that makes sense, sorry for the confusion :)

MichaelBuessemeyer:
> The behavior when entering an entirely new number is kind of weird. For example, the max value is set to 100, now you enter 255 instead. When entering the 2, the max slider of the histogram is automatically set to 2, and when entering the 55 afterwards, the max slider remains at 2. However, the max slider was set to 100 before and it's a little unexpected that this changed (although I understand why it happened). A possible solution for this could be to debounce the onChange of the input field so that the histogram only changes after a second of no change or so. What do you think?


I tried to tackle this by only updating when the events `onPressEnter` and `onBlur` are executed but apparently `onPressEnter` is not available for NumberInput in our antd version. :confounded: Thus I choose to just debounce the update. I tested this and it feels pretty smooth :+1: 

>  [Low Pri] Maybe it's an easy fix, if not nevermind: The two vertical side lines of the active area are not vertical, but slightly angled to the middle. To see this enter 100 and 102 for min and max and look at the active area.

I cannot find the cause for this :confused: 

@daniel-wer Could you please check the small updates I just pushed?

frontend/javascripts/oxalis/view/settings/histogram_view.js
```diff
@@ -32,6 +32,11 @@ type HistogramProps = {
   ) => void,
 };
 
+type HistorgramState = {
```
daniel-wer, line 35:
```suggestion
type HistogramState = {
```

frontend/javascripts/oxalis/view/settings/histogram_view.js
```diff
@@ -45,9 +50,25 @@ export function isHistogramSupported(elementClass: ElementClass): boolean {
...
+  const { min, max, data } = props;
+  if (min != null && max != null) {
+    return { min, max };
+  }
+  const dataMin = Math.min(...data.map(({ min: minOfHistPart }) => minOfHistPart));
+  const dataMax = Math.max(...data.map(({ max: maxOfHistPart }) => maxOfHistPart));
+  return { min: dataMin, max: dataMax };
+}
+
+class Histogram extends React.PureComponent<HistogramProps, HistorgramState> {
```
daniel-wer, line 63:
```suggestion
class Histogram extends React.PureComponent<HistogramProps, HistogramState> {
```

frontend/javascripts/oxalis/view/settings/histogram_view.js
```diff
@@ -183,7 +220,8 @@ class Histogram extends React.PureComponent<HistogramProps> {
       layerName,
       onChangeLayer,
     } = this.props;
-    const { min: minRange, max: maxRange } = this.getMinAndMax();
+    const { currentMin, currentMax } = this.state;
+    const { min: minRange, max: maxRange } = getMinAndMax(this.props);
```
daniel-wer, line 223: Looks like `this.getMinAndMax` can be removed now :)

frontend/javascripts/oxalis/view/settings/histogram_view.js
```diff
@@ -132,9 +197,41 @@ class Histogram extends React.PureComponent<HistogramProps> {
     }
   };
 
+  tipFormatter = (value: number) =>
+    value > 10000 ? value.toExponential() : roundTo(value, 2).toString();
+
+  // eslint-disable-next-line react/sort-comp
+  updateMinimumDebounced = _.debounce(
+    (value, layerName) => this.props.onChangeLayer(layerName, "min", value),
+    1000,
```
daniel-wer, line 206:
```suggestion
    500,
```

I would suggest to half the debounce time to make it feel a little more snappy

frontend/javascripts/oxalis/view/settings/histogram_view.js
```diff
@@ -132,9 +197,41 @@ class Histogram extends React.PureComponent<HistogramProps> {
...
+  );
+
+  updateMaximumDebounced = _.debounce(
+    (value, layerName) => this.props.onChangeLayer(layerName, "max", value),
+    1000,
```
daniel-wer, line 211:
```suggestion
    500,
```

MichaelBuessemeyer:
@daniel-wer Thanks for the suggestions. I just applied them.

PR #4634, opened 2020-05-28
Apply-Agglomerate Performance Increase #1

### LINKED ISSUE #4577: agglomerate mapping too slow for large amounts of segments
if you have downsampled segmentations you tend to request buckets with a lot of different segments. this makes looking up the IDs in the agglomerate file extremely slow, even overloading the server. what to do about this?

youri-k:
- https://___.webknossos.xyz

### Steps to test:
- abc

### Issues:
- fixes #4577 

------
- [ ] Ready for review


youri-k:
This is a first "Aufschlag" :tennis: to increase the performance of the apply-agglomerate step of data loading.
I chose to remove the `Fox` and `Try` safety nets because they increase the execution time slightly and play now displays an error message if an agglomerate file doesn't exist rather than just loading black data.
Additionally, we're populating the cache faster because we load more items from the hdf5 file with each read.

amotta:
In case that's helpful: Our file systems have subblock sizes of 65,536 bytes and 262,144 bytes. The block sizes are much larger: 2,097,152 bytes and 8,388,608 bytes, respectively. I believe that best I/O performance is achieved when roughly matching these block sizes.

PR #4641, opened 2020-06-02
Update react-debounce-render to fix warning and clean up some logging

### LINKED ISSUE #4572: Clean up console logs

philippotto:
- https://cleanuplogging.webknossos.xyz

### Steps to test:
- open a tracing and ensure that the logging is not too verbose. I'm open to suggestions to improve it further.

### Issues:
- fixes #4572

------
- [x] Ready for review


philippotto:
Also see https://github.com/scalableminds/webknossos/pull/4612 which removed some superfluous logging.

PR #4644, opened 2020-06-03
When requesting data from fallback layer, skip version field

fm3:
webknkossos-connect is strict about request formats and fails on unexpected fields. When requesting fallback layer data for volume tracings, the version field was sent to the datastore, but datastores don’t know what to do with this field. This PR removes it in this case.
This fixes https://github.com/scalableminds/webknossos-connect/issues/90 (except for the still-wrong largest segment ID, but that is tracked in separate issue https://github.com/scalableminds/webknossos-connect/issues/92)

The scala tracingstores and datastores ignore superfluous fields rather than fail. Do you think we should adapt this behavior for wkconnect also? @philippotto @jstriebel 

### Steps to test:
- start a volume tracing with fallback layer on neuroglancer dataset (e.g. harris2015 which is preconfigured in webknossos-connect). should work as expected (except for segment ID overflow)

### Issues:
- fixes https://github.com/scalableminds/webknossos-connect/issues/90

------
- [x] Ready for review

### DIFF
--- a/CHANGELOG.unreleased.md
+++ b/CHANGELOG.unreleased.md
@@ -20,7 +20,7 @@ For upgrade instructions, please check the [migration guide](MIGRATIONS.released
 
 ### Fixed
 
--
+- Fixed a bug where requesting volume tracing fallback layer data from webknossos-connect failed. [#4644](https://github.com/scalableminds/webknossos/pull/4644)
 
 ### Removed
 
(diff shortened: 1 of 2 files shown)


philippotto:
> The scala tracingstores and datastores ignore superfluous fields rather than fail. Do you think we should adapt this behavior for wkconnect also? @philippotto @jstriebel

Yes, I think this makes sense. Either way, it should be consistent. So, either both implementations should ignore superfluous fields or be strict about it. I'd do the change which is less effort and less likely to break things (so, presumably ignoring superfluous fields?).

philippotto:
> fixes scalableminds/webknossos-connect#90

Cool! Didn't know the closing works across repositories :)

PR #4647, opened 2020-06-05
Add more information to task csv download

### LINKED ISSUE #4642: Task CSV download should contain all info for bulk upload
The task list has a button for downloading a CSV of the tasks that are currently searched. It would be great if this CSV contains all the columns that are required for the bulk task import in order quickly duplicate or correctly recreate (through excel) these tasks.

Screenshot 2020-05-12 at 16 58 44

MichaelBuessemeyer:
- https://verbosetaskcsvdownload.webknossos.xyz

### Steps to test:
- open the task view and download some tasks
- the first two entries are the task id and the creation time -> remove them from each line
- open the task bulk creation and paste in the left-over content of the csv
- try creating those tasks
- each step should work 

### Issues:
- fixes #4642
------
- [x] Ready for review


frontend/javascripts/admin/task/task_create_form_view.js
```diff
@@ -71,11 +71,41 @@ type State = {
...
+    editPosition,
+    editRotation,
+    status,
+    boundingBoxVec6,
+    projectName,
+    script,
+  } = task;
+  const neededExperienceAsString = `${neededExperience.domain},${neededExperience.value}`;
+  const [editPositionAsString, editRotationAsString] = [editPosition, editRotation].map(
+    position => `${position[0]},${position[1]},${position[2]}`,
```
daniel-wer, line 95:
```suggestion
    vector3 => vector3.join(","),
```

frontend/javascripts/admin/task/task_create_form_view.js
```diff
@@ -71,11 +71,41 @@ type State = {
...
+  );
+  const totalNumberOfInstances = status.open + status.active + status.finished;
+  const boundingBoxAsString = boundingBoxVec6
+    ? boundingBoxVec6.reduce((partialString, val) => `${partialString}${val},`)
```
daniel-wer, line 99:
```suggestion
    ? boundingBoxVec6.join(",")
```

Or does Flow complain about this? This way you would also get rid of the trailing comma.

frontend/javascripts/admin/task/task_create_form_view.js
```diff
@@ -71,11 +71,41 @@ type State = {
...
+    : "0,0,0,0,0,0,";
```
daniel-wer, line 100: If the above works, remove the trailing comma here as well ...

```suggestion
    : "0,0,0,0,0,0";
```

frontend/javascripts/admin/task/task_create_form_view.js
```diff
@@ -71,11 +71,41 @@ type State = {
...
+  const scriptId = script ? `,${script.id}` : "";
+
+  const taskAsString =
+    `${id},${created},${dataSet},${type.id},${neededExperienceAsString},${editPositionAsString},` +
+    `${editRotationAsString},${totalNumberOfInstances},${boundingBoxAsString}${projectName}${scriptId}`;
```
daniel-wer, line 105:
```suggestion
    `${editRotationAsString},${totalNumberOfInstances},${boundingBoxAsString},${projectName}${scriptId}`;
```

... and add it here.

frontend/javascripts/admin/task/task_create_form_view.js
```diff
@@ -71,11 +71,41 @@ type State = {
...
+  } = task;
+  const neededExperienceAsString = `${neededExperience.domain},${neededExperience.value}`;
+  const [editPositionAsString, editRotationAsString] = [editPosition, editRotation].map(
+    position => `${position[0]},${position[1]},${position[2]}`,
+  );
+  const totalNumberOfInstances = status.open + status.active + status.finished;
+  const boundingBoxAsString = boundingBoxVec6
+    ? boundingBoxVec6.reduce((partialString, val) => `${partialString}${val},`)
+    : "0,0,0,0,0,0,";
+  const scriptId = script ? `,${script.id}` : "";
```
daniel-wer, line 101: In my tests it was fine to always add the comma, even when there is no scriptId. So I would suggest to always add the comma in line 105 instead of conditionally, here. This will also make it easier to include a header in the CSV, see my general comment :)

MichaelBuessemeyer:
@daniel-wer Thanks a lot for your valuable feedback and for already suggesting changes. This really helped me a lot :smile: .

> Also, what do you think about dropping the created column. I don't think it adds valuable information and it's one more column that needs to be removed if the values should be used for the bulk creation.

That sounds like a good idea to :+1: 

> Additionally, I think it would be great to include a header in the CSV file, otherwise it's pretty hard to find out what those values are exactly, at a later point in time.

I also had that in mind but forgot to mention it as a suggestion :see_no_evil:. Thanks for coming up with this suggestion.

MichaelBuessemeyer:
@daniel-wer Could you please check the newest version again? I added a header and applied all your feedback :smile:

PR #4654, opened 2020-06-09
Make CI more robust by skipping chromium when installing puppeteer

### LINKED ISSUE #4583: CI occasionally fails with "Failed to download Chromium r656675"
This happened a couple of times now. Is there something we can do?

philippotto:
The CI failed quite a lot in the past, since chromium couldn't be downloaded. At least, for the branch builds, this doesn't seem to be necessary, anyway, which is why this can be skipped. The nightlies seem fairly stable which is why I left them as is (they need chromium; so the fix wouldn't work).

- https://___.webknossos.xyz

### Steps to test:
- CI 

### Issues:
- fixes #4583

------
- [X] Ready for review

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -50,7 +50,7 @@ jobs:
           command: tools/postgres/assert_unique_evolution_numbers.sh
       - run:
           name: Install frontend dependencies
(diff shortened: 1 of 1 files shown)


philippotto:
> LGTM, I assume the downloaded chromium is only used for the screenshot tests?

Exactly!

Issue #4068, opened 2019-05-06
Select user when clicking on the row


MichaelBuessemeyer:
Currently, it is not possible to select a user in the user view (only for admins) by clicking anywhere on the row. Adding this would increase the UX imo.

Let's add this feature to wk :slightly_smiling_face:.

MichaelBuessemeyer:
Here is a spike made by philipp https://codesandbox.io/s/000vqw38rl.

philippotto:
Btw, I didn't create the code ;) It's linked from the antd docs.

fm3:
closed via #4663 :)

PR #4663, opened 2020-06-15
Separate Dataset Manager from Team Manager role

### LINKED ISSUE #4652: Prevent Team Managers from access to all datasets
Currently, wK `team managers` have full access to all datasets. This convenient for many but "breaks" the permission system. Organizations that rely on strict team separation can not use team managers. To make matters worse, it breaks the mental model of a team manager being limited to just his team.

How can we rethink / fix this?

Ideas from earlier discussion:
- Restrict `team manager` access to only datasets with their team. 
- Create dedicated team with visibility to all datasets
- Apply the correct `accepted teams` to each dataset on import automatically. Perhaps have a default team per orga. 
- Closely mimic GitHub's behavior since they already solved this issue

fm3:
This PR makes changes the permission system of wK. Team Managers are now actually scoped to teams in regard to dataset visibility.

- https://cleanupdatasetaccess.webknossos.xyz

### Steps to test:
- check scenarios from permission table (see changes in docs)

### Issues:
- fixes #4652

------
- [x] Ready for review


fm3:
@MichaelBuessemeyer a first version of the backend changes is done. the user update form now accepts `isDatasetManager` updates, the user json will contain that. the `isEditable`-filter when requesting datasets respects the new rules.
Note that the upload request for datasets expects an (optional) new field `initialTeams` (strings containing team ids). The allowed teams field should already be visible in the dataset upload form. The frontend should enforce that uploaders who are neither admin nor dataset manager select at least one team they can manage (or they will immediately lose access to their uploaded dataset)

MichaelBuessemeyer:
TODOs:
- [ ] As a team manager: edit a dataset you have access to and revoke your team(s) from the access list -> results in an error but the update is still executed

@fm3 
It looks like the backend does not allow updating the teams a user is assigned to for team managers:
image
In this example, I am logged in as the first user in the list (which is a team manager of the teams `blabla` and `lala`). I try updating the team roles of the other user but get an error from the backend :/

MichaelBuessemeyer:
I can confirm, that team managers are allowed to edit their datasets :tada:

MichaelBuessemeyer:
Question: 
Is a dataset manager allowed to edit the team access of a dataset? Or must this user be a team manager or admin

frontend/javascripts/admin/dataset/dataset_upload_view.js
```diff
@@ -114,6 +143,13 @@ class DatasetUploadView extends React.PureComponent<PropsWithForm, State> {
                   <Upload.Dragger
                     name="files"
                     beforeUpload={file => {
+                      if (!form.getFieldValue("name") || form.getFieldValue("name") === "") {
+                        const filename = file.name
+                          .split(".")
+                          .slice(0, -1)
+                          .join(".");
+                        form.setFieldsValue({ name: filename });
+                      }
```
MichaelBuessemeyer, line 152: @grittaweisheit Wanted the file name to be the default dataset name if no name was entered before. I just added this here as I think this is a very good idea

frontend/javascripts/admin/user/permissions_and_teams_modal_view.js
```diff
@@ -0,0 +1,271 @@
+// @flow
+import { Modal, Button, Radio, Col, Row, Checkbox } from "antd";
```
MichaelBuessemeyer, line 2: Git did not understand that I renamed the file and manipulated it :/

MichaelBuessemeyer:
@fm3 Thanks for adjusting the e2e tests :rocket: :tada: 

@daniel-wer The frontend code should be fine now. Could you please review it and prioritize the review? :D

fm3:
> Is a dataset manager allowed to edit the team access of a dataset? Or must this user be a team manager or admin

Yes, dataset managers should be allowed to do that. I’ll check that the handles does this correctly, and I’ll also check why you get an error editing other user’s team memberships when logged in as appropriate team manager.

docs/users.md
```diff
@@ -23,23 +23,55 @@ Teams are useful for managing dataset access permissions or simply to organize u
...
+There are four different roles for webKnossos users divided into global, organization-wide roles and team-based permissions:
 
-  - __Team Manager:__ Manages a specific team. Team Managers are able to administrate [Tasks, Task Types and Projects](./tasks.md) belonging to that specific team. They are also allowed to activate newly registered users. Team managers are able to access all tracings that belong to users of their team.
+**On a per team basis:**
 
-  - __Admin:__ Manages the whole organization with all teams. Admins are able to do the same things as team managers but across all teams. They can also give admin permissions to other users by using the `Grant Admin Rights` button at the top of the user list. Admins are able to access all tracings that belong to their organization.
+  - __Member:__ A regular wK user account. Members are only able to access datasets of their respective teams and can create annotations for those. Further, the can work on tasks created by their `Team Manager` or an `Admin`. They are only able to access their own annotations or [annotations shared with them](./sharing.md).
```
daniel-wer, line 32:
```suggestion
  - __Member:__ A regular wK user account. Members are only able to access datasets of their respective teams and can create annotations for those. Further, they can work on tasks created by their `Team Manager` or an `Admin`. They are only able to access their own annotations or [annotations shared with them](./sharing.md).
```

docs/users.md
```diff
@@ -23,23 +23,55 @@ Teams are useful for managing dataset access permissions or simply to organize u
...
 
-Only *Admins* and *Team managers* are able to see/access the `Admin` menu options in the navigation bar.
+  - __Team Manager:__ Manage a specific team. Team Managers can administrate and create [Tasks, Task Types, and Projects](./tasks.md) belonging to their respective teams. They also can activate newly registered users and assign/remove new users to theirs teams. Team managers can access all annotations belonging to users of their respective teams. Similarly to regular `Members`, they are only able to access datasets of their respective teams and can create annotations for those.
```
daniel-wer, line 34:
```suggestion
  - __Team Manager:__ Manage a specific team. Team Managers can administrate and create [Tasks, Task Types, and Projects](./tasks.md) belonging to their respective teams. They also can activate newly registered users and assign/remove users to/from their teams. Team managers can access all annotations belonging to users of their respective teams. Similarly to regular `Members`, they are only able to access datasets of their respective teams and can create annotations for those. 
```

docs/users.md
```diff
@@ -23,23 +23,55 @@ Teams are useful for managing dataset access permissions or simply to organize u
...
 
-Only *Admins* and *Team managers* are able to see/access the `Admin` menu options in the navigation bar.
+  - __Team Manager:__ Manage a specific team. Team Managers can administrate and create [Tasks, Task Types, and Projects](./tasks.md) belonging to their respective teams. They also can activate newly registered users and assign/remove new users to theirs teams. Team managers can access all annotations belonging to users of their respective teams. Similarly to regular `Members`, they are only able to access datasets of their respective teams and can create annotations for those. 
+
+**On an organizational level:** 
+
+  - __Admin:__ Manage a whole organization including full access to all datasets, teams, and projects. Admins can access all administrative settings - similiar to `Team Managers` but for all teams - and have full control over all dataset - similar to `Dataset Managers`. They can promote other users to admin or to `Dataset Manager` by using the `Edit Teams and Permissions` modal at the top of the user list. Admins can access all annotations, datasets, projects, tasks, etc belonging to their respective organization.
```
daniel-wer, line 38:
```suggestion
  - __Admin:__ Manage a whole organization including full access to all datasets, teams, and projects. Admins can access all administrative settings - similiar to `Team Managers` but for all teams - and have full control over all datasets - similar to `Dataset Managers`. They can promote other users to `Admin` or to `Dataset Manager` by using the `Edit Teams and Permissions` modal at the top of the user list. Admins can access all annotations, datasets, projects, tasks, etc belonging to their respective organization.
```

docs/users.md
```diff
@@ -23,23 +23,55 @@ Teams are useful for managing dataset access permissions or simply to organize u
...
+
+  - __Dataset Manager:__ Manage all datasets of an organization. Dataset Managers have full read/wrote access to all datasets within their respective organizations regardless of if a dataset is only been made available to a certain team. Use this role for power users who regularly upload datasets or who need access to all datasets regardless of who created them.
```
daniel-wer, line 40:
```suggestion
  - __Dataset Manager:__ Manage all datasets of an organization. Dataset Managers have full read/write access to all datasets within their respective organizations regardless of whether a dataset has been made available only to a specific team. Use this role for power users who regularly upload datasets or who need access to all datasets regardless of who created them.
```

frontend/javascripts/admin/user/permissions_and_teams_modal_view.js
daniel-wer: Thanks for notifying me, I used http://www.mergely.com/editor to compare the two files which worked very well :)

frontend/javascripts/dashboard/advanced_dataset/dataset_action_view.js
```diff
@@ -143,9 +145,9 @@ class DatasetActionView extends React.PureComponent<Props, State> {
             <div className="text-danger">{dataset.dataSource.status}</div>
           </div>
         ) : null}
-        {dataset.isActive ? (
+        {managesDataset && dataset.isActive ? (
```
daniel-wer, line 148: I don't think this change is equivalent. Only the Edit Dataset and Reload functionality should be restricted to Dataset Mangers/Admins, but all users with access should be able to view and trace the datasets, right? To achieve the same behavior as before, the `managesDataset` should be added to the `dataset.isEditable` check.

frontend/javascripts/dashboard/dataset_view.js
```diff
@@ -151,6 +151,7 @@ function DatasetView(props: Props) {
         datasets={filteredDatasets}
         searchQuery={searchQuery}
         isUserAdmin={Utils.isUserAdmin(props.user)}
```
daniel-wer, line 153: I was wondering why this utils method is needed and noticed that all team managers are considered admins when using this method. At least in this context, this appears to be wrong with the new permission system. I would recommend to check all usages of this method and whether they still make sense!

For reference the utils method:
```
export function isUserAdmin(user: APIUser): boolean {
  return user.isAdmin || isUserTeamManager(user);
}
```

frontend/javascripts/oxalis/view/action-bar/share_modal_view.js
```diff
@@ -119,7 +119,7 @@ export default function ShareModalView(props: Props) {
         "The dataset of this annotation is not public. The Sharing Link will make the dataset accessible to everyone you share it with.";
     } else if (visibility === "Private") {
       message =
-        "The annotation is currently private, so Team Sharing is disabled and only admins and team managers can use the Sharing Link.";
+        "The annotation is currently private, so Team Sharing is disabled and only admins and dataset managers can use the Sharing Link.";
```
daniel-wer, line 122: Reminder to test that this actually works for dataset managers and doesn't work for team managers.

frontend/javascripts/admin/user/user_list_view.js
```diff
@@ -303,42 +285,6 @@ class UserListView extends React.PureComponent<PropsWithRouter, State> {
           >
             Change Experience
           </Button>
-          {this.props.activeUser.isAdmin ? (
-            <React.Fragment>
-              <Button
-                onClick={() =>
-                  Modal.confirm({
-                    title: messages["users.grant_admin_rights_title"],
```
daniel-wer, line 311: There was an extra confirmation (modal) needed to grant or revoke admin rights, before, but I couldn't find this functionality in the new version. Did you discuss and agree to remove this?

frontend/javascripts/dashboard/dataset_view.js
hotzenklotz: I suggest to add a new function that is called `isUserAdminOrTeamManager` in addition to Ì `isUserAdmin`. Than we can do fine-grained control in all components.

hotzenklotz: I went through all calls of `Utils.isUserAdmin` and replaced them with `Utils.isUserAdminOrTeamManager` where appropriate. see commit 2567a1b

hotzenklotz:
@MichaelBuessemeyer @fm3 Any thoughts on this: Right now only Admins can see all users/teams with access to a dataset in the `DatasetAccessListView`? Should Dataset Managers see those too? Probably, right?

That raises the issue again, the dataset manager may be able to see too many user names, emails etc. (Or maybe we have to remove/refactor the `DatasetAccessListView` to only show teams, which is fine.)

https://github.com/scalableminds/webknossos/blob/94c944bc4122763d2e6df3370128b2180cf8d174/frontend/javascripts/dashboard/advanced_dataset/dataset_table.js#L157

frontend/javascripts/dashboard/dataset_view.js
```diff
@@ -184,30 +183,32 @@ function DatasetView(props: Props) {
...
-          </Badge>
-        </Button>
-      </Dropdown>
-    </InputGroup>
-  ) : (
-    searchBox
-  );
+    );
+    
+    const isUserAdminOrDatasetManager = Utils.isUserAdmin(props.user) || Utils.isUserDatasetManager(props.user);;
```
daniel-wer, line 193: @hotzenklotz It looks like you don't have prettier enabled, some of the indentations changed and there is a double semicolon here. Prettier should take care of this automatically, alternatively run `yarn pretty` before committing :)

hotzenklotz: > @MichaelBuessemeyer @fm3 Any thoughts on this: Right now only Admins can see all users/teams with access to a dataset in the `DatasetAccessListView`? Should Dataset Managers see those too? Probably, right?
> 
> That raises the issue again, the dataset manager may be able to see too many user names, emails etc. (Or maybe we have to remove/refactor the `DatasetAccessListView` to only show teams, which is fine.)
> 
> https://github.com/scalableminds/webknossos/blob/94c944bc4122763d2e6df3370128b2180cf8d174/frontend/javascripts/dashboard/advanced_dataset/dataset_table.js#L157

According to @fm3 this view should be:
- viewable by admins (see all users with access)
- viewable by team managers (see only users within their teams)
- viewable by dataset managers (see only users within their teams)

fm3: (meaning that being dataset manager does not help here, only team managers or admins can see content there, so it should be hidden to anyone who is not admin or team manager)

MichaelBuessemeyer: So
https://github.com/scalableminds/webknossos/blob/94c944bc4122763d2e6df3370128b2180cf8d174/frontend/javascripts/dashboard/advanced_dataset/dataset_table.js#L157

should be changed to 
```
isUserAdminOrTeamManager ? dataset => <DatasetAccessListView dataset={datset} /> : null
```
?
If that's the case, I would only show the DatasetAccessListView to TeamManagers if they manage a team of the specific dataset, right?

frontend/javascripts/dashboard/advanced_dataset/dataset_action_view.js
MichaelBuessemeyer: Thanks for paying attention, I accidentally move the change a few lines up before committing :see_no_evil:

frontend/javascripts/oxalis/view/action-bar/share_modal_view.js
MichaelBuessemeyer: @daniel-wer This change was a mistake. Private shared annotations should only be visible to admins or team managers but not to dataset managers. 

@hotzenklotz You wanted to test this PR, correct? Could you please test this too?

frontend/javascripts/admin/user/user_list_view.js
MichaelBuessemeyer: No we did not discuss anything I just removed it without thinking :see_no_evil:. Do you think we should keep a confirmation, if the permissions (not team membership) of a user are changed?

MichaelBuessemeyer:
Open TODOs
- [ ] Currently, it is not possible to give admin rights to a user. The update is silently ignored. :/

MichaelBuessemeyer:
@daniel-wer Could you please check my newest changes :smile: ?

MichaelBuessemeyer:
Additionally:
I came across the problem that if I want to select a user in the user list I have to use the checkbox. 
I already suggested in #4068 to make the whole row clickable.
As this was a quick change I just added it. If you (@daniel-wer @hotzenklotz) can find any negative thing about this change, just let me know and I'll revert the change.

On "bad" thing introduced by this change is that e.g. when you change the email of a user, the use is immediately selected

frontend/javascripts/messages.js
```diff
@@ -266,13 +266,15 @@ instead. Only enable this option if you understand its effect. All layers will n
   "users.needs_admin_rights": "Admin rights are required to change the permissions of users.",
   "users.multiple_selected_users":
     "You selected more than one user. To change the permissions of users you need to select them individually.",
-  "users.grant_admin_rights_title": "Do you really want to grant admin rights?",
-  "users.grant_admin_rights": _.template(
-    "You are about to grant admin privileges to <%- numUsers %> user(s) giving them access to all teams, datasets and annotations. Do you want to proceed?",
+  "users.change_permissions_title": "Do you really want to change the permissions of this user?",
+  "users.revoke_all_permissions": _.template(
+    "Ths user <%- userName %> will no longer have any extra access permissions on datasets.",
```
daniel-wer, line 271: Why the "on datasets"? This message will also be shown when changing from Admin status to Member and Admins have many more permissions, so I would probably change it to "will no longer have any extra permissions."

fm3:
My bad, I fixed the ordering of isAdmin and isDatasetManager attributes and this time got around to actually test it, as the frontend is now also functional :)

MichaelBuessemeyer:
> My bad, I fixed the ordering of isAdmin and isDatasetManager attributes and this time got around to actually test it, as the frontend is now also functional :)

❤️ 🚀

frontend/javascripts/admin/user/permissions_and_teams_modal_view.js
```diff
@@ -0,0 +1,350 @@
...
+  }
+
+  getPermissionSelection(onlyEditingSingleUser: boolean, isUserAdmin: boolean) {
+    const roleStyle = { fontWeight: "bold" };
+    const explanationStyle = { color: "rgba(0, 0, 0, 0.55", paddingBottom: 12 };
+    return (
+      <React.Fragment>
+        <h4>
+          Organization Permissions{" "}
+          <a href="https://docs.webknossos.org/guides/users">
```
daniel-wer, line 235: @MichaelBuessemeyer Could you make this open in a new tab, that's what we usually do for documentation links :)

daniel-wer:
Frontend code LGTM, except for the two small issues I commented on:
- [ ] Wording of `users.revoke_all_permissions` message
- [ ] Organization Permissions link should open in new tab

hotzenklotz:
Flo and I just sat done and did the final testing. We changed some small issues with Team Managers not having enough rights. But other than that it looks good.

MichaelBuessemeyer:
Thanks @fm3 @hotzenklotz @daniel-wer 😃

hotzenklotz:
> @MichaelBuessemeyer Thanks for quickly taking care of my feedback!
> 
> > I came across the problem that if I want to select a user in the user list I have to use the checkbox.
> > I already suggested in #4068 to make the whole row clickable.
> 
> I'm not entirely convinced by this feature as (as you said) users get selected when changing the email or when copying text from the row, but we could try it out and see what users say. Any opinion on this @hotzenklotz?
> 

I fixed the event propagation when selecting an `experience` or changing the `email`. That makes a bit better, We still have the same issue when copying text, though.

PR #4669, opened 2020-06-16
Make merger mode work for all skeleton modifications

### LINKED ISSUE #4514: Toast indicate updating of the active mapping in merger mode when there has been an activated server side mapping before
When directly starting the merger mode no information toasts at the center of the screen are displayed. But when first selecting a server-side mapping and then switching to the merger mode for every update of the mapping toasts are being displayed.

merger-mode-tool-tips_downscaled

### LINKED ISSUE #4655: Undo in merger mode does not work as expected
When merging two nodes, the corresponding segments are merged (i.e., they share the same color / segment id). When hitting undo, the nodes are split again but the segments are still merged.

daniel-wer:
Merger mode only worked for basic node/tree creation/deletion cases (see #4655 for an explanation), but didn't work for undo/redo or when merging/splitting trees.

By switching from overwriting specific redux actions to acting on the result of our tree differ, all of these cases are now handled. The resulting code is even easier as the merger mode only needs to react to two diffing actions: `createNode` and `deleteNode`. Complex redux actions like mergeTree, result in low-level diffing actions:
- `deleteNode` actions for each node of the tree that was deleted as part of the merge
- `createNode` actions for all of the same nodes, but in the tree that "survived" the merge
- a `deleteTree` action that merger mode doesn't care about, similar an `updateTree` action which only modifies tree properties
Similar for undo/redo/tree merging/splitting. This is actually an advantage of the "primitive" low-level diffing we do.

I fixed a couple of other bugs while I was at it (not all of them have issues):
- Progress notifications are no longer displayed if another mapping was activated before (#4514)
- The ID Mapping toggle in the segmentation tab was buggy and remained disabled even though a mapping was activated programmatically or by selecting a value from the mapping dropdown
- The merger mode mapping was not deactivated when deactivating merger mode and remained applied
- There was a race condition in merger mode tasks where sometimes a texture wasn't correctly initialized, resulting in all segments being red
- Not a bug, but I replaced the "You set a node too close to gray ..." modal with a Toast, as was wished for in this Discuss post: https://discuss.webknossos.org/t/remove-popup-from-merger-mode-script/1052.

- https://___.webknossos.xyz

### Steps to test:
- Open a tracing with a segmentation and activate merger mode. Execute all kinds of tree-modifying actions - the merger mode mapping should remain consistent. (Undo, Redo, Node Deletion, Tree Split, Tree Merge, ...)
- Activate another mapping, then activate merger mode and set nodes. No progress notifications should appear.
- Deactivate Merger Mode. The mapping should no longer be applied.
- Activate a mapping through the API: `webknossos.apiReady(3).then(api => api.data.setMapping("segmentation", {1: 2, 3: 4}))`. The ID Mapping toggle in the segmentation tab should become activated.

### Issues:
- fixes #4655
- fixes #4514 

------
- [x] Ready for review


hotzenklotz:
@daniel-wer I just gave it a quick test and seems to work very well :-) Nice!

MichaelBuessemeyer:
I'll try to review this when I got time. Currently, #4663 has a higher priority

frontend/javascripts/oxalis/merger_mode.js
```diff
@@ -80,33 +85,32 @@ function getAllNodesWithTreeId(): Array<NodeWithTreeId> {
...
-    Modal.info({ title: "The created node could not be detected." });
-    return;
-  }
   // If there is no segment id, the node was set too close to a border between segments.
   if (!segmentId) {
-    Modal.info({ title: "You've set a point too close to grey. The node will be removed now." });
-    api.tracing.deleteNode(activeNodeId, activeTreeId);
+    api.utils.showToast(
+      "warning",
+      "You've set a node too close to grey. The node will be removed now.",
```
hotzenklotz, line 103: I am not a big fan of the warning text. What is the `grey`? I don't quite understand the message. 

(Also this message should probably belong into `messages.js`)

hotzenklotz: I like your suggestion. Makes it a lot clearer and does not require intimate knowledge about internals and colors. 

```suggestion
      "You have placed a node outside of a segment. The node will be removed now.",
```

daniel-wer:
I have noticed one more bug that I would like to fix. If the user sets a node outside of a segment, the node is removed automatically while displaying a toast with an explanation. However, this leads to an infinite loop when using undo. If you press undo afterwards, the just automatically removed node is created again, but immediately automatically removed again by the merger mode script. However, by removing the node a new entry is added to the undo stack, so we're in the beginning again...
The solution should be to avoid that the node is ever set at all by registering an overwrite for the CREATE_NODE action and dropping the action if the node is set outside a segment.

PR #4671, opened 2020-06-17
Display Loading... instead of No Data when creating a task

### LINKED ISSUE #3558: Drop-down menus should say »Loading…« instead of »Not found« when appropriate

grittaweisheit:
This PR changes the behavior of several views. Selection fields that use loaded data now dislpay a loading spinner instead of "No Data" while fetching data.   
Affected select fields are loading projects, users, taskTypes, teams, scripts or datasets

- https://loadingintaskcreationselects.webknossos.xyz

### Steps to test:
- login as a user
- have a big set of data for projects, datasets, scripts and/or task types
- (alternatively you could put something like a sleep method in the fetchData function of view you are testing to simulate the wait for the data )
- visit the view whose fields you want to test
- you should see a loading spinner when trying to select a Project, TaskType, Dataset, Team, User or Script until all data has been fetched (try to reload the page if it loads too fast)
- If there is no data for one of the mentioned fields, it should display "No Data" after the spinner disappears

### Issues:
- fixes #3558

------
- [x] talked about possible refactoring
- [x] Ready for review


grittaweisheit:
It could be nicer to do this in a more general way, if we want this behavior in other places as well. What do you think?

frontend/javascripts/admin/task/task_create_form_view.js
```diff
@@ -191,14 +193,15 @@ class TaskCreateFormView extends React.PureComponent<Props, State> {
...
+    this.setState({ isFetchingData: true });
     const [datasets, projects, scripts, taskTypes] = await Promise.all([
       getActiveDatasets(),
       getProjects(),
       getScripts(),
       getTaskTypes(),
     ]);
-
     this.setState({ datasets, projects, scripts, taskTypes });
+    this.setState({ isFetchingData: false });
```
daniel-wer, line 204: You should be able to combine these two setState calls into one, simply add the `isFetchingData: false` in the first call

frontend/javascripts/admin/task/task_create_form_view.js
```diff
@@ -390,11 +393,17 @@ class TaskCreateFormView extends React.PureComponent<Props, State> {
...
+                    Loading...
                   </Option>
-                ))}
+                ) : (
+                  this.state.datasets.map((dataset: APIDataset) => (
+                    <Option key={dataset.name} value={dataset.name}>
+                      {dataset.name}
+                    </Option>
+                  ))
+                )}
```
daniel-wer, line 406: This is a good solution, but (as is often the case) the developers of antd (the UI library we use) have thought of this use case and included a `loading` prop in the `Select` component. Setting it to `this.state.isFetchingData` will show a loading spinner during loading :) (See the docs for an example and description: https://ant.design/components/select/#Select-props)

grittaweisheit: Nice! Thanks a lot for pointing it out to me :)

grittaweisheit: The loading property does not work, so I used the notFoundContent property instead.

grittaweisheit:
I'm not quite sure about those files with a Select component:
- time_line_view.js. Should it be changed, too?
- dataset_components.js loads datastores. Are there enough datastores to make someone see the "No Data" icon while it's loading?
- user_selection_component.js and user_scripts_modal_view.js is already doing what we want, but in a different way. Should I change it or leave it be?

frontend/javascripts/admin/task/task_create_form_view.js
daniel-wer: Ah, I see, the `loading` property seems to only be available in antd 3.11.0 which we do not have yet. Your solution works very well too :+1: 
I'll create an issue to update antd sometime in the future and include a note about the `loading` property :)

PR #4695, opened 2020-07-06
Show Warning When Creating Inaccessible Tasks

### LINKED ISSUE #3597: rethink dataset permissions: can we avoid this “are you sure” situation?
by default no teams are allowed access. this leads to an “are you sure” dialogue because this default does not match the usual workflow… can we create a seamless yet secure experience?

### LINKED ISSUE #4502: During Task creation, warn user about permission issues
e.g. if the dataset cannot be viewed by the users for which the project is created.
I’m not sure how to integrate this into the flow. I think it would be nice if this would come up not as an error message but asynchronously while the user fills out the form. But I don’t know how that would work during bulk task creation. Let’s discuss this!

Related: #3597

fm3:
- https://___.webknossos.xyz

### Steps to test:
- create tasks for dataset with no allowed teams (bulk + individual) → warning should be shown
- create tasks for dataset with allowed team matching the project team (bulk + individual) → no warning should be shown
- edit dataset: no annoying “are you sure?” popup anymore

### Issues:
- fixes #4502 
- fixes #3597

------
- [x] Ready for review

### DIFF
--- a/app/controllers/TaskController.scala
+++ b/app/controllers/TaskController.scala
@@ -15,10 +15,10 @@ import javax.inject.Inject
 import models.annotation.nml.NmlResults.NmlParseResult
 import models.annotation.nml.NmlService
 import models.annotation._
-import models.binary.{DataSetDAO, DataSetService}
-import models.project.ProjectDAO
+import models.binary.{DataSet, DataSetDAO, DataSetService}
+import models.project.{Project, ProjectDAO}
 import models.task._
-import models.team.TeamDAO
+import models.team.{Team, TeamDAO}
 import models.user._
 import net.liftweb.common.{Box, Full}
 import oxalis.security.WkEnv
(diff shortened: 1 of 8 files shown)


app/controllers/TaskController.scala
```diff
@@ -420,17 +420,31 @@ class TaskController @Inject()(annotationDAO: AnnotationDAO,
...
     } yield Ok(Json.toJson(result))
   }
 
+  private def warnIfTeamHasNoAccess(requestedTasks: List[TaskParameters], dataSet: DataSet)(
+      implicit ctx: DBAccessContext): Fox[List[String]] = {
+    val projectNames = requestedTasks.map(_.projectName).distinct
+    for {
+      projects: List[Project] <- Fox.serialCombined(projectNames)(projectDAO.findOneByName(_))
+      dataSetTeams <- teamDAO.findAllForDataSet(dataSet._id)
+      noAccessTeamIds = projects.map(_._team).filter(projectTeam => !dataSetTeams.map(_._id).contains(projectTeam))
```
youri-k, line 441:
```suggestion
      noAccessTeamIds = projects.map(_._team).diff(dataSetTeams.map(_._id)))
```

I feel like a diff would be more concise here :slightly_smiling_face:

PR #4700, opened 2020-07-06
fix initial color of user bounding boxes

### LINKED ISSUE #4699: initial color in viewport when adding a user bbox does not match color in menu
…becomes correct when you change the color manually.

MichaelBuessemeyer:
This PR fixes that the initial displayed color of a user bounding box in the settings is not correct.

- https://fixinitialuserbbcolor.webknossos.xyz

### Steps to test:
- Open a dataset in view mode and add a user bounding box. The color of the box in the settings should match the color displayed in the viewports.

### Issues:
- fixes #4699

------
- [x] Ready for review


frontend/javascripts/libs/utils.js
```diff
@@ -151,7 +151,8 @@ export function colorObjectToRGBArray({ r, g, b }: ColorObject): Vector3 {
 }
 
 export function getRandomColor(): Vector3 {
-  const randomColor = [0, 1, 2].map(() => Math.random());
+  // Here we generate three values between 0 and 1 that multipled with 255 will be integers.
```
daniel-wer, line 154:
```suggestion
  // Generate three values between 0 and 1 that multiplied with 255 will be integers.
```

MichaelBuessemeyer:
> Do you know why the color difference was so visible?

The `UserBoundingBoxInput` uses the `Utils.rgbToHex` function. I guess this causes the color to be completely wrong. I'll take a quick look and look whether I can also fix the bug in this method.

MichaelBuessemeyer:
@daniel-wer Could you please give the newest changes a quick check again? :smile:

daniel-wer:
Works very well :rocket:

PR #4709, opened 2020-07-09
Merge Volume Tracings

### LINKED ISSUE #3870: allow merging of volume tracings
should be straightforward if all tracings have distinct bounding boxes (which seems to be the standard case)

fm3:
### TODO
 - [x] implement mergeFromIds route
 - [x] implement mergedFromContents route, send volume zips in initialData (in a tar? or multipart-request?)
 - [x] implement deterministic ordering (last modified annotation wins)
 - [x] save buckets to temporary store when appropriate, handle this when loading buckets
 - [x] send hard errors when loading buckets for tracing that is removed from temporary store, encouraging user to reload. Q: where to get this information? [store in redis.] How long should the cache timeout be? [1h] Can we re-merge in the backend if tracing is still active? [no, let user reload]
 - [x] relabel to avoid merging cells?
 - [x] copy to my account for temporary volume data?
 - [x] test for volume tracings with 8/16/32-bit segmentation fallback layer

- https://___.webknossos.xyz

### Steps to test:
- upload zipfile with multiple volume annotations (create by e.g. downloading a volume project with multiple finished tasks)
 → should be merged
 → note that cell IDs are changed to be distinct
- click “view“ on a volume project with multiple finished tasks
 → should be merged, download should still give individual task data.zips
- click “copy to my account” on this compound volume project
 → should be merged, now download should also contain merged volume data
- test this also with fallback layer (also to check that ids >255 are handled correctly) (this is probably only possible via zip upload)

### Issues:
- fixes #3870 

------
- [x] Ready for review


webknossos-tracingstore/app/com/scalableminds/webknossos/tracingstore/tracings/volume/VolumeTracingBucketHelper.scala
```diff
@@ -82,9 +83,22 @@ trait VolumeTracingBucketHelper
...
+                 data: Array[Byte],
+                 version: Long,
+                 toCache: Boolean = false): Fox[Unit] = {
     val key = buildBucketKey(dataLayer.name, bucket)
-    volumeDataStore.put(key, version, compressVolumeBucket(data, expectedUncompressedBucketSizeFor(dataLayer)))
+    val compressedBucket = compressVolumeBucket(data, expectedUncompressedBucketSizeFor(dataLayer))
+    if (toCache) {
+      // Note that this cache is for temporary volumes only (e.g. compound projects)
+      // and cannot be used for download or versioning
+      Fox.successful(volumeDataCache.insert(key, compressedBucket, Some(20 minutes)))
```
youri-k, line 98: I think this shouldn't be hardcoded to 20 minutes. I don't know if we need to expose this as a config settings, but there might be different preferences based on the RAM, etc.

webknossos-datastore/app/com/scalableminds/webknossos/datastore/storage/TemporaryStore.scala
```diff
@@ -27,6 +27,13 @@ class TemporaryStore[K, V] @Inject()(system: ActorSystem) {
       map.values.toList
     }
 
+  def findAllConditionalWithKey(predicate: K => Boolean): collection.Set[(K, V)] =
+    map.synchronized {
+      map.keySet.filter(predicate).map { key =>
```
youri-k, line 32: I don't know if there is a reason you want to return a `Set`, but if there is not 
```suggestion
      map.filterKeys(predicate)
```
should have the same result and is more concise (and hopefully more performant)

webknossos-tracingstore/app/com/scalableminds/webknossos/tracingstore/tracings/volume/VolumeTracingService.scala
```diff
@@ -149,6 +155,105 @@ class VolumeTracingService @Inject()(
...
+    ZipIO.withUnziped(initialData) {
+      case (_, is) =>
+        val labelSet: mutable.Set[Byte] = scala.collection.mutable.Set()
+        ZipIO.withUnziped(is) {
+          case (_, is) =>
+            WKWFile.read(is) {
+              case (header, buckets) =>
+                if (header.numBlocksPerCube == 1) {
+                  val data: Array[Byte] = buckets.next()
+                  val nonZeroData: Array[Byte] = filterBytes(data, byte => byte != 0x00)
```
youri-k, line 173: In my opinion, the extraction isn't really necessary here, but it's not something that'll block this pr :slightly_smiling_face:

fm3: I agree, it does not improve readability, in fact it may slightly degrade it. However, inlining this function breaks the code because of implicit conversions. This is why I wrote the `FilterBytes` trait, which does *not* extend `ProtoGeometryImplicits` and `KeyValueStoreImplicits`

youri-k: Ah, I see. Thanks for the explanation

fm3:
@youri-k I added support for different data types, could you have another look?

youri-k (approved):
LGTM :shipit:

Issue #4687, opened 2020-06-29
Improve visual separation of layer-specific sections in settings


philippotto:
See https://discuss.webknossos.org/t/opacity-or-opacity/1602/3?u=potto.

I'd vote for improving the vertical padding between the sections. One needs to be careful that this also looks good with the various states of collapsed/expanded sections which can be adjacent to each other.

daniel-wer:
Maybe adding a light gray horizontal divider between the different layers would help in addition to the increased vertical padding.

philippotto:
Fixed by #4722.

@grittaweisheit If you use `fixes #4687` in the PR description (instead of `solves #4687`), this issue would have been closed automatically when the PR is merged. Just as a tip.. :)

PR #4722, opened 2020-07-16
Improve visual separation of layer-specific sections in settings

grittaweisheit:
This PR improves the visual separation of the layer-specific sections in the settings by adding light grey lines between them.

- https://layersectionseparation.webknossos.xyz

### Steps to test:
- open a tracing for a dataset with several layers
- open the settings
- the layer sections should be divided by light grey lines

### Issues:
- solves #4687

------
- [x] Ready for review


grittaweisheit:
Do you want the Invert-button to be vertically aligned to the histogram's slider, or something else? I left it for now, but I should be able to align it easily to something else :)

philippotto:
Awesome! Looks way better to me :)

The only thing missing is some margin-adaption for the segmentation layer settings. You probably only opened a dataset which does not have a segmentation layer. However, the segmentation layer setting has the same problems (namely, the opacity slider needs to move closer to the layer name. also the margin-bottom to the next layer should be be double-checked). If you change/check these two margins, this PR should be good to go :)

Here's an example dataset with a segmentation layer: https://layersectionseparation.webknossos.xyz/datasets/sample_organization/TestUpload/view#905,595,512,0,1.000

image



> Do you want the Invert-button to be vertically aligned to the histogram's slider, or something else?

I meant, that the invert-button should be vertically aligned with its siblings of the same row. So, it needs to be moved a few pixels down :) However, I just thought that this probably should not happen in this PR. I noticed that the invert-button uses absolute positioning which should be refactored, anyway. Also the entire layer settings could need some further iterations, but this is independent of this PR, since it doesn't really target the "visual separation of the layers" and has been like this before. I created another [issue](https://github.com/scalableminds/webknossos/issues/4732) for this. If you feel like doing this, you could assign this issue to yourself. However, it doesn't have high priority, so it should only be an option when you are running out of issues :)

grittaweisheit:
The segmentation layer settings should be better now :)

PR #4728, opened 2020-07-23
Apply d/f switching option to copying slices

### LINKED ISSUE #4708: Volume segmentation copy shortcuts are flipped under certain conditions
Under certain conditions the functionalities of keyboard shortcuts `V` and `SHIFT + V` are flipped, i.e. `V` copies from the next slice and `SHIFT + V` copies from the previous slice instead. One way of reproducing this:

(d/f Switching option disabled, and volume segmentation exists in slice 1)

1. Navigate to slice 2 and press `V`. The volume annotation from slice 1 is copied as expected.
2. Erase the volume annotation in slice 2 using the brush mode.
3. Navigate to slice 1 and change the segmentation in some way by painting or erasing.
4. Navigate to slice 2. Pressing `V` now does not do anything, and pressing `SHIFT + V` copies the volume segmentation from slice 1 instead.

The procedure above is obviously an artificial situation, but I have encountered the same bug under conditions that are probably related. Similarly, under certain conditions I need to use `V` instead of `SHIFT + V` to copy from the next slice.

MichaelBuessemeyer:
This PR changes the copy slice volume functionality so that the considered "previous" slice will always be z-1 when the d/f option is disabled.
Additionally, this PR adds a small fix that keeps the `spaceDirectionOrtho` value of the direction, where it did not change. Previously, if a slice was annotated twice, the second update always caused the dimension that did not change (e.g. z coordinate) to be 1. Now the value the previous value will be kept. -> (This is not so easy to test without using console output)

- https://fixdirectionwhencopyingvolume.webknossos.xyz

### Steps to test:
- Open a volume tracing without an existing volume layer.
- Just use the x-y viewport for this and turn off d/f switching:
1. annotate something
2. move to the next slice and use `v` to copy the annotation
3. Then erase the copied annotation and move back to the previously annotated slice
4. Annotate this slice a bit further and the move to the next slice again
5. copy that slice using `v`. The slice should be copied.
6. Move back to the first slice, erase everything, and use `shift + v` to copy the "second" slice. This should work.

### Issues:
- fixes  #4708

------
- [x] Ready for review


CHANGELOG.unreleased.md
```diff
@@ -14,7 +14,7 @@ For upgrade instructions, please check the [migration guide](MIGRATIONS.released
 -
 
 ### Changed
-- 
+- When d/f switching is turned off and a slice is copied with the shortcut `v`, the previous slice used as source will always be slice - 1 and `shift + v` will always take slice + 1 as the slice to copy from. [#4728](https://github.com/scalableminds/webknossos/pull/4728)
```
daniel-wer, line 17:
```suggestion
- When d/f switching is turned off and a slice is copied with the shortcut `v`, the previous slice used as the source will always be slice - 1 and `shift + v` will always take slice + 1 as the slice to copy from. [#4728](https://github.com/scalableminds/webknossos/pull/4728)
```

frontend/javascripts/oxalis/model/reducers/flycam_reducer.js
```diff
@@ -132,7 +132,13 @@ export function zoomReducer(state: OxalisState, zoomStep: number): OxalisState {
 }
 
 export function setDirectionReducer(state: OxalisState, direction: Vector3) {
-  const spaceDirectionOrtho = direction.map(el => (el < 0 ? -1 : 1));
+  const previousSpaceDirectionOrtho = state.flycam.spaceDirectionOrtho;
```
daniel-wer, line 135: Nice find and fix :+1:

PR #4740, opened 2020-07-29
Show failing bucket requests

### LINKED ISSUE #4725: show error toasts for failed bucket requests
currently bucket request errors are silently ignored. I think this is not necessary.

bucket requests for non-existing buckets should still be answered with 200, containing the missing-buckets header field.
if the server is down, there is already the health check handling, which does the trick.
if something other than 200 is returned, there should be an error toast (ideally, not one per bucket, but one per distinct error?). e.g. due to expired temporary volume, as introduced in #4709, or expired access tokens

MichaelBuessemeyer:
This PR fixes that errors that are thrown in a worker are not correctly transported to the main thread. This includes an update of the comlink library. This fix is used to show good error messages when bucket requests fail.

- https://___.webknossos.xyz

### Steps to test:
- Suggestion: Test this locally:
in `webknossos/datastore/controllers/BinaryDataController.scala` add `_ <- Fox.failure("nope")` as a single line after line 62.
- Then view a dataset.
- An error toast with code 400 should appear for each layer. (The toast is not sticky, as this error could only be a temporary thing).

### Issues:
- fixes  #4725

------
- [x] Ready for review


MichaelBuessemeyer:
@philippotto I assigned you to this review as I already talked with daniel about it. By assigning you I'l like to share the knowledge about the workaround with you so more people know whats going on here :)

frontend/javascripts/oxalis/model/bucket_data_handling/wkstore_adapter.js
```diff
@@ -145,23 +146,40 @@ export async function requestFromStore(
...
-    let resultBuffer = responseBuffer;
-    if (fourBit) {
-      resultBuffer = await decodeFourBit(resultBuffer);
-    }
-
-    return sliceBufferIntoPieces(layerInfo, batch, missingBuckets, new Uint8Array(resultBuffer));
-  });
+  try {
+    return doWithToken(async token => {
+      const stuff = await Request.sendJSONReceiveArraybufferWithHeaders(
```
philippotto, line 151: Maybe `response` or `bufferWithHeaders` or something along these lines instead of `stuff`?

philippotto: Or simply undo this and stick to the destructuring? I guess, this was only for debugging purposes?

frontend/javascripts/oxalis/model/bucket_data_handling/wkstore_adapter.js
```diff
@@ -145,23 +146,40 @@ export async function requestFromStore(
...
+        `${dataUrl}/data?token=${token}`,
+        {
+          data: bucketInfo,
+          timeout: REQUEST_TIMEOUT,
+          showErrorToast: false,
+          doNotInvestigate: true,
```
philippotto, line 157: Does `doNotInvestigate: true` make sense? If buckets cannot be served, the chance is high that this is due to a datastore which is down (which would be investigated if doNotInvestigate was false).

frontend/javascripts/oxalis/workers/headers_transfer_handler.js
```diff
@@ -1,19 +1,94 @@
...
+    stack: string,
+  },
+  response?: {
+    status: number,
+    statusText: string,
+    headers: SerializedHeaders,
+  },
+};
+
+// This handler is designed to overwrite the default "throw" handler of the comlink library, as this handler does not convert object that cannot be send via postMessage.
```
philippotto, line 49:
```suggestion
// This handler is designed to overwrite the default "throw" handler of the comlink library, as this handler does not convert objects that cannot be send via postMessage.
```

frontend/javascripts/oxalis/model/bucket_data_handling/wkstore_adapter.js
```diff
@@ -145,23 +146,40 @@ export async function requestFromStore(
...
+    });
+  } catch (errorResponse) {
+    const errorMessage = `Requesting buckets from layer "${
+      layerInfo.name
+    }" failed with status code ${errorResponse.status} - "${errorResponse.statusText}".`;
+    const urlAsString = `URL - ${dataUrl}`;
+    console.error(`${errorMessage} ${urlAsString}`);
+    console.error(errorResponse);
+    Toast.renderDetailedErrorMessage(errorMessage, urlAsString, {
+      sticky: true,
```
philippotto, line 178: `sticky: true` means that the toast cannot be closed, right? Maybe that's a bit too aggressive, since there was no error message at all before? Just wondering if we go from one extreme to the other with this change..

philippotto: Also this seems to contradict the PR description which states:

>  (The toast is not sticky, as this error could only be a temporary thing)

Or am I missing something? If it's not sticky, then this code probably needs some throttling (I'd suggest one minute; the throttling key could be the same as the toast-key). Otherwise, the user could close the toast and the toast would immediately re-appear as soon as the next bucket request fails.

philippotto:
I just gave this branch a test by opening a dataset and then uninstalling the dev-branch again. Failing bucket requests were visible in the console with 404s. However, there were no toasts. Is this expected?

frontend/javascripts/oxalis/model/bucket_data_handling/wkstore_adapter.js
MichaelBuessemeyer: Indeed :see_no_evil:

MichaelBuessemeyer:
Thanks for your feedback :+1: 

>I just gave this branch a test by opening a dataset and then uninstalling the dev-branch again. Failing bucket requests were visible in the console with 404s. However, there were no toasts. Is this expected?

Thanks for finding this. I'll have a look at it :)

frontend/javascripts/oxalis/model/bucket_data_handling/wkstore_adapter.js
MichaelBuessemeyer: > Or am I missing something?

Nope, thanks for finding this. Somehow I messed up here as I thought I removed it and overlooked it during the last check ...

frontend/javascripts/oxalis/model/bucket_data_handling/wkstore_adapter.js
MichaelBuessemeyer: Thanks for nagging, somehow I thought that the error will be handled earlier when investigating the error. But a look at the code told me that the error is passed on to the upper level (`Promise.reject`)

frontend/javascripts/oxalis/model/bucket_data_handling/wkstore_adapter.js
```diff
@@ -145,23 +146,41 @@ export async function requestFromStore(
     createRequestBucketInfo(zoomedAddress, resolutions, fourBit, applyAgglomerates, version),
   );
 
-  return doWithToken(async token => {
-    const { buffer: responseBuffer, headers } = await Request.sendJSONReceiveArraybufferWithHeaders(
-      `${dataUrl}/data?token=${token}`,
-      {
+  try {
+    return await doWithToken(async token => {
```
MichaelBuessemeyer, line 150: Finally found why the errors were not caught: The `await` statement was missing.
@philippotto Could you please have another look at the code again? I applied your feedback and all errors should now be caught properly.

philippotto: Hmm, not sure I understand this fix. `doWithToken` returns a promise and there shouldn't be a difference between simply returning this promise or awaiting that promise -- thus creating a new promise -- and returning that new promise.
See the highlighted section here for another example: https://javascript.info/async-await#:~:text=We%20could%20explicitly%20return%20a%20promise,%20which%20would%20be%20the%20same:

Or am I missing something?

MichaelBuessemeyer: Oh, there wasn't a bug in the beginning but I added one temporarily:
 If the await statement is missing, the promise of `doWithToken` is directly returned and the catch statement will never catch a failing request as the promise is already returned. But if I wait for `doWithToken` to succeed or fail, the catch block correctly catches the error :+1:

philippotto: Ahh, thanks for the clarification :+1: I didn't consider the try-catch-block in this context. Makes total sense :+1:

Issue #3055, opened 2018-08-14
Update immutability-helper to make use of TS typing


philippotto:
### /edit
The text below is not up-to-date. Instead, we should update the [immutability-helper](https://github.com/kolodny/immutability-helper) package to make use of the TS typing in its newest version.

### Outdated:
Unfortunately, the clever syntax of `update`'s API, most of our reducers are barely typed. Flow simply accepts all possible objects for `update` even though only specific structures yield the intended benefit.

One way to solve this would be to always use object spreads. Since this becomes quite verbose, we should try to find good helper functions which itself are typed. For example, instead of

```js
const oldState = state;
const oldTracing = state.tracing;
{
  ...oldState,
  tracing: {
    ...oldTracing,
    someAttr: someNewValue
  }
}
```

we could have:
```js
updateTracing(state, {
	someAttr: someNewValue
})
```

with
```js
function updateTracing(oldState: StoreType, tracingShape: $Shape<TracingType>): StoreType {
	return {
		...oldState,
		tracing: {
			...oldState,
			...tracingShape
		}
	}
}
```

philippotto:
To do:

- [X] implement typed update utility functions
- [X] adapt settings reducer
- [X] adapt annotations reducer
- [ ] adapt view mode reducer
- [ ] adapt task reducer
- [ ] adapt ui reducer
- [ ] adapt user reducer
- [ ] adapt flycam reducer
- [ ] adapt save reducer
- [ ] adapt volume tracing reducer
- [ ] adapt skeletontracing reducer

I suggest to not do everything at once, but instead do it successively (starting with the easier ones).

hotzenklotz:
@philippotto Is this still an issue with typescript?

philippotto:
yes, it's still relevant, but easier and a bit different: we should update the [immutability-helper](https://github.com/kolodny/immutability-helper) package to make use of its TS typing. I adapted the issue accordingly.

PR #4743, opened 2020-07-30
Add possibility to move nodes

daniel-wer:
Add the possibility to move nodes by adding an action/reducer and by making the WebGL buffer updater (`skeleton.js`) update node and edge positions.
The active node can be moved using "Ctrl + Arrow Keys" or by using "Ctrl + LeftMouseDrag".

- https://movenodes.webknossos.xyz

### Steps to test:
- Open a tracing and move nodes using the mentioned keyboard and mouse controls.
  - Does it work in each viewport?
  - Is the node position persisted after a reload?
  - Does it work for branchpoints?

- Testing the merger mode:
  - open a skeleton tracing with a segmentation layer
  - enable the merger mode and create two nodes that merge to segments
  - Now move one of the nodes around. The mapping should update as expected.

### Issues:
- This will allow users to easily correct ground truth skeleton data. Previously one needed to remove a node, add a new node and reconnect the split trees.

------
- [x] Ready for review


daniel-wer:
- [ ] Use Ctrl + MouseDrag instead of Shift + MouseDrag, because Shift + Click selects a node and the potential to mistakenly move nodes feels rather high
- [ ] Undo buffer should not be completely filled with updateNode actions, take latest instead

frontend/javascripts/oxalis/geometries/skeleton.js
```diff
@@ -485,6 +489,36 @@ class Skeleton {
...
+        return [positionAttribute];
+      });
+    };
+
+    for (const edge of tree.edges.getOutgoingEdgesForNode(nodeId)) {
+      edgePositionUpdater(edge, 0);
+    }
+    for (const edge of tree.edges.getIngoingEdgesForNode(nodeId)) {
+      // The changed node is the target node of the edge which is saved
+      // after the source node in the buffer
```
philippotto, line 517: Instead of passing the `indexOffset`, I'd pass a `isIngoingEdge`. Then, `edgePositionUpdater` can do `const indexOffset = isIngoingEdge ? 3 : 0`. Seems like the better abstraction to me.

frontend/javascripts/oxalis/geometries/skeleton.js
```diff
@@ -485,6 +489,36 @@ class Skeleton {
     });
   }
 
+  /**
+   * Updates a node's position and that of its edges in a WebGL buffer.
+   */
+  updateNodePosition(treeId: number, nodeId: number, position: Vector3, tree: Tree) {
```
philippotto, line 495: Isn't the `treeId` already contained in the `tree` object?

philippotto: Alternatively, I think this method could do the `const tree = skeletonTracing.trees[treeId];` part on its own, since the `tree` is not used on the call site (except for passing it).

frontend/javascripts/oxalis/geometries/skeleton.js
```diff
@@ -485,6 +489,36 @@ class Skeleton {
...
+    const bufferNodeId = this.combineIds(nodeId, treeId);
```
philippotto, line 496: This might be premature-optimization speaking in me, but I'd feel a bit more comfortable, if this method checked whether the `position` really changed with the `updateAction`. A simple `if (._isEqual(tree.position, position)) return;` should already solve this, in my opinion. Since this code path will be taken quite rarely, this is not really important (at least now), but I'd like to avoid establishing the pattern of doing redundant stuff :)

frontend/javascripts/oxalis/model/reducers/skeletontracing_reducer.js
```diff
@@ -252,6 +252,28 @@ function SkeletonTracingReducer(state: OxalisState, action: Action): OxalisState
...
+        case "SET_NODE_POSITION": {
+          const { position, nodeId, treeId } = action;
+          return getNodeAndTree(skeletonTracing, nodeId, treeId)
+            .map(([tree, node]) => {
+              const diffableMap = skeletonTracing.trees[tree.treeId].nodes;
+              const newDiffableMap = diffableMap.set(
+                node.id,
+                update(node, { position: { $set: position } }),
+              );
+              return update(state, {
```
philippotto, line 264: Could you use `updateKey` and `updateKey4` (?) from `deep_update` instead of the update function? See this issue for context: https://github.com/scalableminds/webknossos/issues/3055

I know that the issue is quite old (and hidden), but I'd find it cool when new code uses the type-checked version for updating things :)

frontend/javascripts/oxalis/model/sagas/save_saga.js
```diff
@@ -68,16 +71,21 @@ export function* collectUndoStates(): Saga<void> {
     const curTracing = yield* select(state => enforceSkeletonTracing(state.tracing));
     if (userAction) {
       if (curTracing !== prevTracing) {
+        if (!shallSkipPreviousState(userAction, previousAction)) {
```
philippotto, line 74: I'd negate the semantics of this helper method and name it `shouldAddToUndoStack`, because it seems clearer and easier to parse to me (since there is no double negation).

frontend/javascripts/oxalis/geometries/skeleton.js
daniel-wer: I also thought about this (as I wrote this code), however, I don't think it's easily possible.

> A simple if (._isEqual(tree.position, position)) return; should already solve this, in my opinion.

The tree does not have a position, but each node has one. However, even if we would check the node's position, it would already be the new one, wouldn't it? We would need a reference to the old tracing to check the previous position of the node.

philippotto: Oops, very good point. I glanced over this too quickly. One could look up the old position in the nodeBuffer, but this doesn't appeal to me anymore ^^ Then, half of the function's code is already executed. So, let's leave it as is :)

MichaelBuessemeyer:
> Also, please add the new shortcuts to the shortcut overview page :)


good catch :+1:

frontend/javascripts/oxalis/model/reducers/skeletontracing_reducer.js
MichaelBuessemeyer: Unfortunate, using `updateKey4` here isn't possible because the flow definition expects that all key must exist and cannot be null or undefined, but the `skeleton` property is optional and thus can be undefined/null.

FYI: no `updateKeyX` function is used in the whole `skeletontracing_reducer`, thus making this change quite some work.

Is it fine with you to ignore this suggested change under these circumstances @philippotto?

MichaelBuessemeyer:
@philippotto I applied all of your feedback except one thing (see my comment). Could you please have a look at this PR again?

frontend/javascripts/oxalis/model/reducers/skeletontracing_reducer.js
philippotto: Oh, ok, then nevermind :)

frontend/javascripts/oxalis/controller/combinations/volumetracing_plane_controller.js
```diff
@@ -70,17 +71,21 @@ const simulateTracing = async (): Promise<void> => {
...
       const tool = getVolumeTool(volumeTracing);
       const contourTracingMode = getContourTracingMode(volumeTracing);
 
       if (tool === VolumeToolEnum.MOVE) {
         const state = Store.getState();
-        const { activeViewport } = state.viewModeData.plane;
-        const v = [-delta.x, -delta.y, 0];
-        Store.dispatch(movePlaneFlycamOrthoAction(v, activeViewport, true));
+        if (state.tracing.skeleton != null && event.ctrlKey) {
+          skeletonController.moveNode(delta.x, delta.y);
```
philippotto, line 83: Can we avoid that the volume tracing controller imports the skeleton controller? The `plane_controller` already switches between volume vs. skeleton controller depending on the current tool (see `createToolDependentHandler` usage in `plane_controller`). Therefore, I would have thought that the volume tracing planer controller doesn't have to know about moving nodes. Consequently, one couldn't move nodes when being in trace or brush mode, but this is perfectly fine.
Or is there another reason why this wouldn't work?

frontend/javascripts/oxalis/controller/viewmodes/plane_controller.js
```diff
@@ -328,6 +333,12 @@ class PlaneController extends React.PureComponent<Props> {
     };
   }
 
+  getLoopedKeyboardControls() {
+    return this.props.tracing.skeleton != null
```
philippotto, line 337:
```suggestion
    // Note that this code needs to be adapted in case the volumeController also starts to expose
    // looped keyboard controls. For the hybrid case, these two controls would need t be combined then.
    return this.props.tracing.skeleton != null
```

normanrz:
Btw moving nodes does not seem to work with merger mode yet.

frontend/javascripts/oxalis/controller/combinations/volumetracing_plane_controller.js
MichaelBuessemeyer: I just resolved that. The code is cleaner now. Thanks for mentioning this :smile:

MichaelBuessemeyer:
@philippotto I just adapted the merger mode to work correctly when nodes are moved. Could you please check this PR again? I added a section on how to test the merger mode to the description. The description is rather easy :smile:

CHANGELOG.unreleased.md
```diff
@@ -11,6 +11,7 @@ For upgrade instructions, please check the [migration guide](MIGRATIONS.released
 [Commits](https://github.com/scalableminds/webknossos/compare/20.07.0...HEAD)
 
 ### Added
+- Added the possibility to move nodes. [#4743](https://github.com/scalableminds/webknossos/pull/4743)
```
```suggestion
- Added the possibility to move nodes in skeleton tracings. This can be done either by pressing CTRL + arrow key or by dragging while holding CTRL. [#4743](https://github.com/scalableminds/webknossos/pull/4743)
```

frontend/javascripts/oxalis/merger_mode.js
```diff
@@ -97,56 +96,81 @@ async function createNodeOverwrite(store, call, action, mergerModeState: MergerM
   if (!segmentId) {
     api.utils.showToast("warning", messages["tracing.merger_mode_node_outside_segment"]);
   } else {
-    call(action);
+    await call(action);
+    // We center the create cell manually, as we implicit override the action that centers the new cell automatically.
```
philippotto, line 100:
```suggestion
    // We center the create cell manually, as we implicitly override the action that centers the new cell automatically.
```

philippotto: Interesting. I would have thought, that `call(action);` still dispatches the original action which is why no behavior is lost (i.e., the centering should still happen). But I'm probably wrong?

frontend/javascripts/oxalis/merger_mode.js
```diff
@@ -97,56 +96,81 @@ async function createNodeOverwrite(store, call, action, mergerModeState: MergerM
...
-   nodes are deleted from it. */
-function deleteNode(mergerModeState: MergerModeState, nodeId: number) {
-  const noNodesLeftForTheSegment = onNodeDeleted(mergerModeState, nodeId);
-  if (noNodesLeftForTheSegment) {
+async function onUpdateNode(mergerModeState: MergerModeState, node: NodeWithTreeId) {
+  const { position, id, treeId } = node;
+  const { segmentationLayerName, nodeSegmentMap } = mergerModeState;
+  const segmentId = await api.data.getDataValue(segmentationLayerName, position);
+  if (nodeSegmentMap[id] !== segmentId) {
+    // If the segment of the node changed, it is like the node got got deleted and a copy got created somewhere else.
```
philippotto, line 162:
```suggestion
    // If the segment of the node changed, it is like the node got deleted and a copy got created somewhere else.
```

frontend/javascripts/oxalis/merger_mode.js
```diff
@@ -97,56 +96,81 @@ async function createNodeOverwrite(store, call, action, mergerModeState: MergerM
...
+    // Thus we use the onNodeDelete and onNodeCreate method to update the mapping.
+    if (nodeSegmentMap[id] != null) {
+      await onDeleteNode(mergerModeState, id, false);
+      // debugger;
```
philippotto, line 166: can be removed

frontend/javascripts/oxalis/merger_mode.js
MichaelBuessemeyer: I just tried out why I added the explicit call to center the active node.

The reason is that somehow the centering messes up and centers the last created node. By adding the centerActiveNode` call this issue is fixed.

As this was easy to fix, I did not debug why actually the last node is centered and the one that just got added.

Do you think it is fine to keep it this way, @philippotto? Or shall I look why the last cell is centered instead of the one that just got added?

philippotto: I think, it's fine as is. But maybe extend the comment with the information you just stated here :)

PR #4745, opened 2020-07-30
List of projects with a specific TaskType

### LINKED ISSUE #4420: Allow to list projects which use a specific task type
[Link to Discuss](https://discuss.webknossos.org/t/task-types-view/1568)

In the task types view, there should be a "view projects" button for each task type which lists the projects which use said task type.

Maybe, this can replace the "view" function (we need to clarify whether there aren't users that need this feature).

grittaweisheit:
This PR adds the possibility to access a list of all projects that contain tasks with a specific taskType.

- https://___.webknossos.xyz

### Steps to test:
- go to the Task Types list view
- click on "Projects" of a task type of your choosing
- there should be a list of exactly those projects that contain a task of this type

### Issues:
- fixes #4420

------
- [x] Ready for review


grittaweisheit:
Should I write documentation anywhere for this?

frontend/javascripts/admin/project/project_list_view.js
```diff
@@ -85,11 +89,18 @@ class ProjectListView extends React.PureComponent<PropsWithRouter, State> {
   }
 
   async fetchData(): Promise<void> {
-    const projects = await getProjectsWithOpenAssignments();
-
+    let projects;
+    let taskTypeName = null;
+    if (this.props.taskTypeId) {
+      projects = await getProjectsForTaskType(this.props.taskTypeId);
+      taskTypeName = await getTaskType(this.props.taskTypeId);
```
philippotto, line 96: You can combine these two lines via `Promise.all` so that only one `await` is used. As a result, both requests are processed in parallel instead of sequentially :)

frontend/javascripts/admin/project/project_list_view.js
```diff
@@ -48,6 +51,7 @@ type State = {
   searchQuery: string,
   isTransferTasksVisible: boolean,
   selectedProject: ?APIProjectWithAssignments,
+  taskTypeName: string,
```
philippotto, line 54:
```suggestion
  taskTypeName: ?string,
```

since the name can be null

philippotto:
> Should I write documentation anywhere for this?

I think, that's not necessary, as the feature is quite self-explanatory.

app/models/project/Project.scala
```diff
@@ -100,6 +100,20 @@ class ProjectDAO @Inject()(sqlClient: SQLClient)(implicit ec: ExecutionContext)
...
+  def findAllWithTaskType(taskTypeId: String)(implicit ctx: DBAccessContext): Fox[List[Project]] =
+    for{
+      accessQuery <- readAccessQuery
+      r <- run(
+        sql"""select distinct p._id, p._team, p._owner, p.name, p.priority, p.paused, p.expectedTime, p.isBlacklistedFromReport, p.created, p.isDeleted
+          from webknossos.projects_ p
+          join webknossos.tasks_ t on t._project = p._id 
+          join webknossos.taskTypes_ tt on t._taskType = tt._id
+          where tt._id = '#${taskTypeId}'"""
+          .as[ProjectsRow]
```
fm3, line 112:
```suggestion
        sql"""select #${columnsWithPrefix("a.")}
          from webknossos.projects_ p
          join webknossos.tasks_ t on t._project = p._id 
          join webknossos.taskTypes_ tt on t._taskType = tt._id
          where tt._id = '${taskTypeId}'"""
          .as[ProjectsRow]
```

app/models/project/Project.scala
```diff
@@ -100,6 +100,20 @@ class ProjectDAO @Inject()(sqlClient: SQLClient)(implicit ec: ExecutionContext)
       parsed <- Fox.combined(r.toList.map(parse))
     } yield parsed
 
+  def findAllWithTaskType(taskTypeId: String)(implicit ctx: DBAccessContext): Fox[List[Project]] =
+    for{
+      accessQuery <- readAccessQuery
```
fm3, line 105:
```suggestion
```
(access query is not used in the complex query, so we don’t need to generate it)

frontend/javascripts/admin/project/project_list_view.js
grittaweisheit: That's cool! Thanks :)

app/models/project/Project.scala
grittaweisheit: This is great, thanks! :)
I need the distinct, though. Otherwise one project shows up several times if it contains more than one task of the task type.

fm3: oh, yes, you are right! I did not mean to remove the "distinct". good catch :)

philippotto:
I just tried to test this branch, but got an SQL error when trying to create a new task. But I don't think this is related to this PR (?). @fm3 Do you know what's going on?

See screenshot:
image

fm3:
@philippotto thanks for pointing this out, that is in fact an independent bug introduced in #4711. I added a commit here that fixes it.

Also, backend LGTM.

@grittaweisheit I noticed that when navigating away via menu from this task-type specific project page to the main project page, the content does not seem to be updated, including the title. That’s probably a react problem?

grittaweisheit:
@fm3 Yes, it was a react problem; thanks for pointing that out! The component didn't update properly when the props changed. It works now :)

PR #4750, opened 2020-08-03
Keep zoom during rotation using the buttons of the 3D-viewport

### LINKED ISSUE #4739: Switching 3D view via button zooms away
When using the buttons in the 3D view the scene center and scale is way off. It would be great if the current scale and center (of the remaining axes) could be kept, or the whole dataset is visible in the viewport using all available space.

recording (1)

[Link to the tracing](https://webknossos.brain.mpg.de/annotations/Explorational/5f213b3c0100000e00a86c11#150609,159787,443,0,0.527,206)

might be related to #4021

MichaelBuessemeyer:
This PR reworks the functionality of the rotation buttons of the 3D-Viewport. The zoom is now kept during rotation and the calculation is now shorted, commented and easier to understand.

- https://keepzoomduringtdbuttonrotation.webknossos.xyz

### Steps to test:
- View a dataset. The initial view of the 3d-viewport should show all three planes (independent of the z coordinate of the flycam). The whole width of the dataset should be displayed in the viewport.
- Now rotate the dataset in the viewport a bit and change the zoom and use the rotation buttons as you like. The rotation of the camera should always be as expected.
- One exception: The rotation of the buttons might look a bit off due to not using always the shortest path because we do not use quaternions for interpolation.

### Issues:
- fixes #4739

------
- [x] Ready for review


frontend/javascripts/oxalis/controller/camera_controller.js
```diff
@@ -171,87 +173,64 @@ type TweenState = {
...
-      upX: 0,
-      upY: 0,
-      upZ: -1,
-      l: squareCenterX - squareWidth / 2,
-      r: squareCenterX + squareWidth / 2,
-      t: squareCenterY + height / 2,
-      b: squareCenterY - height / 2,
-    };
+  // Factor to reduce the clipping offset of the z coordinate to get a better angle on the dataset.
+  const zReductionFactor = 0.7;
```
daniel-wer, line 195: Is this needed because of the different z-scale in relation to x and y? If so, shouldn't it be computed from the dataset scale automatically as dataset scales differ widely?

frontend/javascripts/oxalis/controller/camera_controller.js
```diff
@@ -171,87 +173,64 @@ type TweenState = {
...
+  // Use width and height to keep the same zoom.
+  let width = tdCamera.right - tdCamera.left;
+  let height = tdCamera.top - tdCamera.bottom;
+
+  let position: Vector3;
+  let up: Vector3;
+  // Way to calculate the position and rotation of the camera:
+  // First, the camera is either positioned at the current center of the flycam or in the dataset center.
+  // Second, the camera is the moved backwards by a clipping factor into the wanted direction.
```
daniel-wer, line 204:
```suggestion
  // Second, the camera is moved backwards by a clipping offset into the wanted direction.
```

frontend/javascripts/oxalis/controller/camera_controller.js
```diff
@@ -171,87 +173,64 @@ type TweenState = {
...
+  // Together with matching lookUp (up) vectors and keeping the width and height, the position and rotation updates correctly.
+  if (id === OrthoViews.TDView && (height <= 0 || width <= 0)) {
+    // This should only be the case when initializing the 3D-viewport.
+    const aspectRatio = getInputCatcherAspectRatio(state, OrthoViews.TDView);
+    const datasetCenter = voxelToNm(dataset.dataSource.scale, getDatasetCenter(dataset));
+    // The camera has not width and height which might be due to a bug or the camera has not been initialized.
```
daniel-wer, line 210:
```suggestion
    // The camera has no width and height which might be due to a bug or the camera has not been initialized.
```

frontend/javascripts/oxalis/controller/camera_controller.js
```diff
@@ -171,87 +173,64 @@ type TweenState = {
...
+    // Thus we zoom out to show the whole dataset.
+    const paddingFactor = 1.1;
+    width =
+      Math.sqrt(
+        datasetExtent.width * datasetExtent.width + datasetExtent.height * datasetExtent.height,
+      ) * paddingFactor;
+    height = width / aspectRatio;
+    up = [0, 0, -1];
+    // For very high datasets that have a very low or hight z starting coordinate, the planes might not be visible.
```
daniel-wer, line 219:
```suggestion
    // For very tall datasets that have a very low or high z starting coordinate, the planes might not be visible.
```

daniel-wer: I don't understand this part yet. To be in line with the current behavior, shouldn't the TD view be zoomed such that the whole dataset bounding box is visible, initially? Why would the planes not be visible in that case?

frontend/javascripts/oxalis/controller/camera_controller.js
```diff
@@ -262,54 +241,60 @@ export function rotate3DViewTo(id: OrthoView, animate: boolean = true): void {
...
-      b: t - height,
-    };
+    up = upVector[id];
+    position = [
+      positionOffset[id][0] + flycamPos[0],
+      positionOffset[id][1] + flycamPos[1],
+      positionOffset[id][2] + flycamPos[2],
+    ];
   }
+  // Ensure a valid width and height if camera ended up with no width and height due to a bug.
```
daniel-wer, line 251: Is this actually a bug or does this only happen exactly once when the camera is not initialized, yet?

MichaelBuessemeyer:
>This distracts me more than I would've thought. The "crazy" rotations that happen sometimes make it hard to keep the correct spatial picture in my head. I tested the buttons using the old code, and there the rotations seem to always use the shortest path - What's the difference to the new code, why does it not work that way any longer?


To be honest, I had trouble understanding how the previous code worked. That's why I cannot say what actually the difference is between before and this version. But maybe I should really investigate some more time into fine-tuning the rotation to not do something "crazy"

frontend/javascripts/oxalis/controller/camera_controller.js
MichaelBuessemeyer: I thought of this as a way to simply change the view angle at the dataset. Here are two examples:
With `zReductionFactor = 1`:
image
 With `zReductionFactor = 0.7`:
image
As you can see: The angle changed.

But if you compare these I think the first version is fine too and thus the factor can be omitted.

MichaelBuessemeyer:
@daniel-wer I tried out using quaternions to smooth the rotation, but somehow the camera does not update after applying the desired quaternion. I pushed the code I used to test rotating a not used camera to branch `try-out-quaternions`.

-> Thus I would not invest more time into this unless a customer or so is really irritated by the weird rotation. Previously, the rotation was also a bit off in some cases. 
=> I suggest to keep the code as it is. Is that ok with you?

frontend/javascripts/oxalis/controller/camera_controller.js
MichaelBuessemeyer: This happened during refactoring the rotation, but in the current code this should not happen anymore.

I simply removed it :+1:

MichaelBuessemeyer:
@daniel-wer I had another chat with @philippotto about the weird rotation and we found out that the "3D" preset had "wrong" values and corrected them to work fine with the other presets. Thus the rotation between all presets does not look weird anymore.

Nevertheless, the problem that the rotation in some cases is looks weird still exists. For that, we need to interpolate the position on a sphere. But this should be done in another pr :slightly_smiling_face:

MichaelBuessemeyer:
@philippotto As daniel is currently unavailable, do you want to take over reviewing this pr?

philippotto:
> @philippotto As daniel is currently unavailable, do you want to take over reviewing this pr?

Sure thing! I just gave it another test and this PR vastly improves the rotation :+1: Very pleasant to use.

However, I noticed one issue concerning the aspect ratio / scale.

If the 3D viewport is not square, the rendered plane is squished somehow:
image

Compare this to the current master:
image


In both cases, I clicked the XY button. Probably `left-right` and `top-bottom` need to be adapted to the aspect ratio of the 3D view? Something in this direction, I'd guess. If this is fixed, this PR should be good to merge :)

MichaelBuessemeyer:
@philippotto Thanks for finding this bug. I accidentally forgot to remove a reassignment of width and height. The bug should be fixed now

frontend/javascripts/oxalis/controller/camera_controller.js
```diff
@@ -171,87 +173,64 @@ type TweenState = {
...
+  if (id === OrthoViews.TDView && (height <= 0 || width <= 0)) {
+    // This should only be the case when initializing the 3D-viewport.
+    const aspectRatio = getInputCatcherAspectRatio(state, OrthoViews.TDView);
+    const datasetCenter = voxelToNm(dataset.dataSource.scale, getDatasetCenter(dataset));
+    // The camera has no width and height which might be due to a bug or the camera has not been initialized.
+    // Thus we zoom out to show the whole dataset.
+    const paddingFactor = 1.1;
+    width =
+      Math.sqrt(
+        datasetExtent.width * datasetExtent.width + datasetExtent.height * datasetExtent.height,
```
philippotto, line 213:
```suggestion
        datasetExtent.width **2 + datasetExtent.height ** 2,
```

if I'm not mistaken

MichaelBuessemeyer: Thanks, I forgot that js has a "shortcut" for exponational expressions :+1:

daniel-wer:
Just wanted to say, the rotation is much better now! Really cool work :rocket:

PR #4755, opened 2020-08-06
Multi-Resolution Volume Tracings

### LINKED ISSUE #286: Handle multiple volume resolutions for hybrid annotations
<a href="https://timer.scm.io/repos/537237ce1a00001a003c4a14/issues/286/create?referer=github" target="_blank">Log Time</a>

### LINKED ISSUE #4686: Available resolutions list per layer
In the dataset view, there is a list of available resolutions in a tooltip. It would be useful to have that on a per layer basis in case it differs from layer to layer. (e.g. if segmentation is only available in 4-4-1)

fm3:
### Introducing Volume Annotations in Multiple Resolutions.

- Volume tracing objects now contain a list of resolutions in which buckets should be created and saved.
- The front-end simultaneously updates buckets in all of those resolutions whenever the user changes the segmentation.
- Every brush stroke / fill action etc. is up/downsampled to all those resolutions.
- The back-end volume adds explicit downsampling functionality so that legacy volume annotations can be converted in this way.
- The list of resolutions is determined by the fallback layer (use its resolutions if it is present) or the union of all layers (use these resolutions if there is no fallback layer).
- When merging multiple volume annotations, the resolution sets have to be identical.
- Task Types can restrict resolutions. If a volume/hybrid task has resolution restrictions, only those resolutions are present in the task volumes.

Related [design doc in notion](https://www.notion.so/scalableminds/Design-Doc-Multi-Res-Volume-Tracings-42b530ea74e4430f94c845c7e01e9468) (some points may be outdated, see comments here)

**Note that a secondary PR https://github.com/scalableminds/webknossos/pull/4891 has been merged into this one, overwriting some of the discussions AND steps-to-test here.**

- https://multiresvolumes.webknossos.xyz

### TODO:
- [X] ensure all crucial checkboxes in the PR comments are done
- [X] ensure the PR diff itself does not contain "todo" comments anymore or that they link to follow up issues (see https://github.com/scalableminds/webknossos/issues/4838)

### Steps to test:
( **Note that a secondary PR https://github.com/scalableminds/webknossos/pull/4891 has been merged into this one, overwriting some of the discussions AND steps-to-test here.** )
#### Back-end focussing tests
- [x] open legacy volume annotation, brushing + filling + tracing should still work and save mag1-buckets
- [x] open legacy volume annotation, zooming out should display warning about invisible segmentation
- [x] create new volume annotation with fallback layer, tracing should have mags of fallback layer
- [x] create new volume annotation without fallback layer, tracing should have all mags (union) of all layers
- [x] upload legacy volume annotation via dashboard, resulting tracing should have the specified mag list, and have its data downsampled accordingly
- [x] create new volume annotation with fallback layer, zooming out to resolution where fallbacklayer/volume data is not present should display warning about invisible segmentation
- [x] create new volume annotation with fallback layer only present in mag2 (but color data in all mags), volume annotation should have mag2 only (and create mag2 data only)
- [x]  create new volume annotation, brush some, download (data.zip should have bucket data in each mag)
- [x] reupload this multi-resolution volume annotation, should have same resolution list and volume data
- [x] open legacy  or new volume annotation, click »unlink fallback segmentation«, should reload with new empty volume tracing that has all mags (union) of all present dataset layers
- [x] create new *hybrid* annotation, there should not be differences from volume-only concerning the volume data
- [x] check that both upload/download and brushing work with isotropically and anisotropically downsampled datasets

#### Front-end focussing tests
- [x] test with volume tracing **without** fallback layer
	- [x] test brush
		- [x] brush in mag 1, zoom out and check that this is also visible in higher mags
		- [x] brush in mag 2-2-1, zoom in and check that this is also visible in mag1. note that multiple slices should be annotated.
	- [x] "polygon"-trace
		- [x] "polygon"-trace in mag 1, zoom out and check that this is also visible in higher mags
		- [x] "polygon"-trace in mag 2-2-1, zoom in and check that this is also visible in mag1. note that multiple slices should be annotated.
	- [x] use "copy segmentation from last slice"-tool (V shortcut)
		- [x] copy segmentation in mag 1, zoom out and check that this is also visible in higher mags
		- [x] copy segmentation in mag 2-2-1, zoom in and check that this is also visible in mag1. note that multiple slices should be annotated.
  		  - [x] when zooming between mags, copy-segmentation sometimes does not work? double check why.
- [x] test with volume tracing **with** fallback layer (which has the same mags as the color layer)
	- [x] test brush
		- [x] brush in mag 1, zoom out and check that this is also visible in higher mags
		- [x] brush in mag 2-2-1, zoom in and check that this is also visible in mag1. note that multiple slices should be annotated.
	- [x] "polygon"-trace
		- [x] "polygon"-trace in mag 1, zoom out and check that this is also visible in higher mags
		- [x] "polygon"-trace in mag 2-2-1, zoom in and check that this is also visible in mag1. note that multiple slices should be annotated.
	- [x] use "copy segmentation from last slice"-tool (V shortcut)
		- [x] copy segmentation in mag 1, zoom out and check that this is also visible in higher mags
		- [x] copy segmentation in mag 2-2-1, zoom in and check that this is also visible in mag1. note that multiple slices should be annotated.
- [x] test with volume tracing **with** fallback layer (which does not have mag1)
	- [x] ensure that "render missing data" black option is disabled
	- [x] test brush
		- [x] brush in mag 1, zoom out and check that this is also visible in higher mags
		- [x] brush in mag 2-2-1, zoom in and check that this is also visible in mag1. note that multiple slices should be annotated.
	- [x] "polygon"-trace
		- [x] "polygon"-trace in mag 1, zoom out and check that this is also visible in higher mags
		- [x] "polygon"-trace in mag 2-2-1, zoom in and check that this is also visible in mag1. note that multiple slices should be annotated.
	- [x] use "copy segmentation from last slice"-tool (V shortcut)
		- [x] copy segmentation in mag 1, zoom out and check that this is also visible in higher mags
		- [x] copy segmentation in mag 2-2-1, zoom in and check that this is also visible in mag1. note that multiple slices should be annotated.
- [X] test with volume tracing **with** fallback layer (which has "holes" in the mag list; e.g., mag 1 and mag 4 exists, but mag 2 does not)
	- [x] ensure that an warning is shown in the toolbar when trying to annotate something --> volume annotation should be disabled in the missing mags (if it can't be rendered)

### Issues:
- fixes #286 
- fixes #4686

------
- ~[ ] Updated [documentation](../blob/master/docs) if applicable~
- ~[ ] Adapted [wk-connect](https://github.com/scalableminds/webknossos-connect) if datastore API changes~
- [x] Ready for review


MichaelBuessemeyer:
TODO:
 - [ ] Test whether anisotropic resolutions of the volumetracing layer ([4,4,1] instead of [4,4,4]) are sent to the frontend correctly.
 And they should be used by the frontend correctly.

fm3:
we should double-check in the backend if the blank “volume tracing contains all mags” check can cause harm. it may be a good idea to actually hold the list of mags with the volume tracing. Also: what happens if the dataset resolutions change after a tracing has been created?

daniel-wer:
@MichaelBuessemeyer Because I just saw the one Todo "When the trace tool is disabled: just do it in grey, not and "aggressive" orange". I just fixed the disabled style for buttons in the header in another merged PR, so simply setting the `disabled` prop should be enough to "gray out" the Trace button :)

fm3:
todos:
- [ ] unlink fallback layer button should create *new* volume tracing in that annotation
- [ ] backend should handle downsampling when merging annotations that have different resolution lists
- [ ] backend should store the present resolutions for each tracing
- [ ] front-end has to handle “legacy” tracings, showing the “zoom in to view segmentation” toast again.

fm3:
Hi @MichaelBuessemeyer as discussed earlier I added the backend route `/annotations/:typ/:id/unlinkFallback`. The idea is that the unlinkink of the fallback segmentation is no longer an update action for existing volume tracings, but rather creates a new volume tracing in the annotation. Could you adapt the front-end button to use that route rather than creating an update action?
The route returns an annotation json again (containing a new volumeTracingId), so I suppose a page refresh/reinitialization is required. Specifically the list of resolutions in the volume tracing may be different after this operation (also version, largestSegmentId etc.). The user should also be warned that what they already brushed in this annotation is lost and this operation cannot be undone. (Our justification for this is that the typical use-case for removing the segmentation fallback is done *before* first brushing anyway).

fm3:
@youri-k  Could you have a first look at the backend code? Here is a high-level summary of backend changes
- upload/merge routes now call downsampling
- tracingstore has to figure out which mags to create, new RPC to webKnossos, querying datasource
- volume tracing proto now contains organization name (for the above rpc), so the NML parser writes this directly to the tracing object, no longer to the parseSuccess container (nml unit test suite adapted accrdingly)
- unlinkFallback route in wK calls unlinkFallback in tracingStore, mutates annotation and returns new annotation object to caller
- extracted BucketKeys trait from VolumeTracingBucketHelper

frontend/javascripts/admin/api_flow_types.js
```diff
@@ -615,6 +615,7 @@ export type ServerVolumeTracing = {|
   elementClass: ElementClass,
   fallbackLayer?: string,
   largestSegmentId: number,
+  resolutions?: Array<Point3>,
```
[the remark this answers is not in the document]
fm3: yes

frontend/javascripts/oxalis/model_initialization.js
```diff
@@ -452,14 +453,43 @@ function setupLayerForVolumeTracing(
   const fallbackLayer = layers[fallbackLayerIndex];
   const boundaries = getBoundaries(dataset);
 
+  console.log(tracing.resolutions);
+
+  const tracingResolutions = tracing.resolutions
```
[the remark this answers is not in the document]
fm3: yes, should be so

webknossos-tracingstore/proto/SkeletonTracing.proto
```diff
@@ -63,6 +63,7 @@ message SkeletonTracing {
     optional BoundingBox userBoundingBox = 10;
     repeated TreeGroup treeGroups = 11;
     repeated NamedBoundingBox userBoundingBoxes = 12;
+    optional string organizationName = 13;
```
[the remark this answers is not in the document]
fm3: yes, because the tracingstore now has to be able to ask wK for the dataset layer propeties, so the organization name is now stored with the tracings. I wanted to do this both in skeleton and volume tracings, for uniformity

webknossos-tracingstore/app/com/scalableminds/webknossos/tracingstore/tracings/volume/VolumeTracingDownsampling.scala
```diff
@@ -0,0 +1,209 @@
...
+    val originalMag = Point3D(1, 1, 1)
+    for {
+      requiredMags <- getRequiredMags(tracing)
+      elementClass = elementClassFromProto(tracing.elementClass)
+      bucketDataMap = new mutable.HashMap[BucketPosition, Array[Byte]]() {
+        override def default(key: BucketPosition): Array[Byte] = Array[Byte](0)
+      }
+      _ = fillMapWithInitialBucketsInplace(bucketDataMap, tracingId, dataLayer)
+      originalBucketPositions: List[BucketPosition] = bucketDataMap.keys.toList
+      updatedBuckets = new mutable.HashSet[BucketPosition]()
```
youri-k, line 76: Change to `List`

webknossos-tracingstore/app/com/scalableminds/webknossos/tracingstore/tracings/volume/VolumeTracingDownsampling.scala
```diff
@@ -0,0 +1,209 @@
...
+        Array.fill[Byte](bucketVolume * bytesPerElement)(0)
+      } else sourceBucketData
+    }
+
+  private def downsampleData[T: ClassTag](data: Array[Array[T]],
+                                          downScaleFactor: Point3D,
+                                          bucketVolume: Int): Array[T] = {
+    val result = new Array[T](bucketVolume)
+    for {
+      z <- 0 until 32
```
youri-k, line 166: Extract magic number

philippotto:
@daniel-wer Feel free to already have a look at the front-end code. As discussed, it's probably a good idea to talk in person about the high level details of this PR. Just ping me tomorrow :)

philippotto:
I created a follow up issue with remaining todos from which I think that they can be done separately. @fm3 @daniel-wer feel free to check the issue out and object if you see things differently: https://github.com/scalableminds/webknossos/issues/4838

From my POV, the PR could be tested now, too, but I'd suggest to do a "loose" testing round and then after the review + iteration do the thorough testing plan, so that this is the last step before merging.

-------------
@fm3 Since the "segmentation-layer-unlinking" changed again in this PR, I'm wondering whether we need to keep the old update-action for this forever in our code base? Future unlinkings won't produce an update-action, but we probably need to handle tracings where the old update-action was added to? Or do we migrate this somehow? I left the old update-action code in the front-end for now, but wasn't sure about that.

frontend/javascripts/oxalis/model_initialization.js
```diff
@@ -452,21 +421,53 @@ function setupLayerForVolumeTracing(
   const fallbackLayer = layers[fallbackLayerIndex];
   const boundaries = getBoundaries(dataset);
 
+  console.log(tracing.resolutions);
```
daniel-wer, line 424: If this should be kept, I would add a label to it in the output, otherwise I would remove it.

frontend/javascripts/oxalis/model_initialization.js
```diff
@@ -452,21 +421,53 @@ function setupLayerForVolumeTracing(
...
+
+  // Legacy tracings don't have the `tracing.resolutions` property
+  // since they were created before WK started to maintain multiple resolution
+  // in volume annotations. Therefore, this code falls back to mag (1, 1, 1) for
+  // that case.
+  const tracingResolutions = tracing.resolutions
+    ? tracing.resolutions.map(({ x, y, z }) => [x, y, z])
+    : [[1, 1, 1]];
+
+  console.log(tracingResolutions);
```
daniel-wer, line 434: See the above comment

frontend/javascripts/oxalis/view/action-bar/volume_actions_view.js
```diff
@@ -36,30 +46,42 @@ class VolumeActionsView extends PureComponent<Props> {
...
+    const { activeTool, activeResolution, isInMergerMode } = this.props;
+    const hasResolutionWithHigherDimension = activeResolution.some(val => val > 1);
+    const multiSliceAnnotationInfoIcon = hasResolutionWithHigherDimension ? (
+      <Tooltip title="You are annotating in a low resolution. You might be annotating multiple slices at once.">
+        <i className="fas fa-layer-group" />
+      </Tooltip>
+    ) : null;
+    const isTraceToolDisabled = isZoomStepTooHighForTraceTool();
+    const traceToolDisabledTooltip = isTraceToolDisabled
+      ? "Your zoom is low to use the trace tool. Please zoom in further to use it."
```
daniel-wer, line 58:
```suggestion
      ? "Your zoom is too low to use the trace tool. Please zoom in further to use it."
```

frontend/javascripts/test/model/binary/cube.spec.js
```diff
@@ -98,7 +100,7 @@ test("GetBucket should only create one bucket on getOrCreateBucket()", t => {
 
 test("Voxel Labeling should request buckets when temporal buckets are created", t => {
   const { cube, pullQueue } = t.context;
-  cube.labelVoxel([1, 1, 1], 42);
+  cube.labelVoxelInResolution([1, 1, 1], 42, 0);
```
daniel-wer, line 103: The zoomStep is the 3rd parameter and the cellId is the 4th. Same for the following calls.

frontend/javascripts/oxalis/model/volumetracing/volumelayer.js
```diff
@@ -193,6 +239,13 @@ class VolumeLayer {
     }
     const maxCoord2d = this.get2DCoordinate(this.maxCoord);
 
+    // The maximum area is scaled by 3 as the min and maxCoord will always be three slices apart,
+    // because in lines 171 + 172 a value of 2 is subtracted / added when the values get updated.
```
daniel-wer, line 243: I would rather reference the location by function name, the line number reference is very brittle.

daniel-wer (commented):
Very cool stuff, I'm excited that we finally support this feature!

I have reviewed all but the three frontend files with the most changes. I'll finish the review with those files tomorrow. Wanted to submit my review-in-progress, though :)

Missing files:
frontend/javascripts/test/model/volumetracing/volume_annotation_sampling.spec.js
frontend/javascripts/oxalis/model/volumetracing/volume_annotation_sampling.js
frontend/javascripts/oxalis/model/sagas/volumetracing_saga.js

philippotto:
Just an aside: I just added a new dataset (`ROI2017_wkw_sparse_segmentation`) to the dev instances. Its segmentation is missing in mag 1 and mag 16. When using the fallback segmentation, this covers some crucial edge cases.

frontend/javascripts/oxalis/constants.js
```diff
@@ -141,6 +143,8 @@ const Constants = {
   BUCKET_WIDTH: 32,
   BUCKET_SIZE: 32 ** 3,
   VIEWPORT_WIDTH,
+  // About the area the brush reaches at maximum radius (pi * 300 ^ 2).
+  AUTO_FILL_AREA_LIMIT: 200000,
```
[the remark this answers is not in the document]
philippotto: Good point. Also, I'd think that we can go higher than a brush radius, since the brush has to be drawn on every mouse movement, whereas the trace tool works in "atomic steps" (where it's okay to wait a bit). So, maybe we can go as far as "10 brush strokes".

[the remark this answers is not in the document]
daniel-wer: Agreed :+1:

fm3:
@philippotto the update action will have to remain in the backend code, as the materialized tracings are created lazily from update actions and we cannot guarantee that there are no such update actions in the live db already.
However, I don’t see why the frontend needs to keep the code for this update action. So feel free to remove that :)

philippotto:
> However, I don’t see why the frontend needs to keep the code for this update action. So feel free to remove that :)

Aren't these actions exposed to the front-end at some points? Maybe it's the only case, but the Version Restore View fetches the update actions from the back-end to display what changed. And maybe there are other cases, too?

fm3:
Oh yes, you are of course right

CHANGELOG.unreleased.md
```diff
@@ -16,9 +16,10 @@ For upgrade instructions, please check the [migration guide](MIGRATIONS.released
 - Added the possibility to remove the fallback segmentation layer from a hybrid/volume tracing. Accessible by a minus button next to the layer's settings. [#4741](https://github.com/scalableminds/webknossos/pull/4766)
 - Added the possibility to undo and redo volume annotation strokes. [#4771](https://github.com/scalableminds/webknossos/pull/4771)
 - Added the possibility to navigate to the preceding/subsequent node by pressing "ctrl + ," or "ctrl + ." in a skeleton tracing. [#4147](https://github.com/scalableminds/webknossos/pull/4784)
+- Added multi-resolution volume annotations. [#4755](https://github.com/scalableminds/webknossos/pull/4755)
```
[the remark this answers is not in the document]
philippotto: I added a note about the old legacy tracings which won't have the multi-res support automatically. Other than that, nothing else comes to my mind.

frontend/javascripts/oxalis/model/helpers/position_converter.js
```diff
@@ -74,6 +75,7 @@ export function getResolutionsFactors(resolutionA: Vector3, resolutionB: Vector3
   ];
 }
 
+// TODO (1): zoomedAddressToAnotherZoomStep usages should be converted to zoomedAddressToAnotherZoomStepWithInfo
```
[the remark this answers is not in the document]
philippotto: I adapted the comment to point to the follow-up issue I created. Let me know if you disagree with this :)

frontend/javascripts/oxalis/model/accessors/dataset_accessor.js
```diff
@@ -27,6 +27,123 @@ import { aggregateBoundingBox } from "libs/utils";
...
+
+  getHighestResolutionIndex(): number {
+    return Math.log2(this.getHighestResolutionPowerOf2());
+  }
+
+  getHighestResolutionPowerOf2(): number {
+    return _.max(Array.from(this.resolutionMap.keys()));
+  }
+
+  // todo: consider to only go upwards instead of both directions.
```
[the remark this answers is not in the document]
philippotto: For some cases, I'm not really sure whether it's the best thing to always go upwards. For example, it could be nice that the cell selection even works when the segmentation cannot be rendered (since this is only a shader limitation and the cell picking is done in JS). However, for the prefetching (as you say) it doesn't make sense to go in both directions.
I added another method (`getIndexOrClosestHigherIndex`) which only goes in the poorer mag direction and used it where appropriate. Let me know in case you disagree!

frontend/javascripts/oxalis/model/accessors/volumetracing_accessor.js
```diff
@@ -44,10 +45,14 @@ export function getContourTracingMode(volumeTracing: VolumeTracing): ContourMode
   return contourTracingMode;
 }
 
-export function isVolumeTracingDisallowed(state: OxalisState) {
-  const isVolumeTracing = state.tracing.volume != null;
-  const isWrongZoomStep = getRequestLogZoomStep(state) > 1;
-  return isVolumeTracing && isWrongZoomStep;
+export function isVolumeTraceToolDisallowed(state: OxalisState) {
+  if (state.tracing.volume == null) {
+    return false;
```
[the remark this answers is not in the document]
philippotto: I checked the callers and I think that the function is only called when there is a volume tracing. So, it doesn't really matter probably. However, I agree with your general take on the semantics and changed it to return true.

frontend/javascripts/oxalis/model/accessors/dataset_accessor.js
```diff
@@ -27,6 +27,123 @@ import { aggregateBoundingBox } from "libs/utils";
...
+    return resolution;
+  }
+
+  getResolutionByIndexWithFallback(index: number): Vector3 {
+    const resolutionMaybe = this.getResolutionByIndex(index);
+    if (resolutionMaybe) {
+      return resolutionMaybe;
+    } else {
+      const powerOf2 = this.indexToPowerOf2(index);
+      return [powerOf2, powerOf2, powerOf2];
```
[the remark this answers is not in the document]
philippotto: > This fallback doesn't have anything to do with the fallback rendering, right?

Correct, the reason for this code is that the previous code always assumed that all layers have all magnifications (trusting that missing mags would be caught later on). Out of caution, I implemented this fallback here, since I didn't want to check all the callers for their compatibility with the new assumption.

However, I just noticed that this fallback is not really correct, since it always assumes an isotropic magnification. The old code solved this by actually falling back to the magnifications of other layers, however, this information is not really available here. I could add it here, but I'll try to work around this somehow else probably. Thanks for raising attention to this here :+1:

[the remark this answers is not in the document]
philippotto: Hm, I tried to get rid of the with-fallback code but couldn't do it completely. See my latest commit here: https://github.com/scalableminds/webknossos/pull/4755/commits/74aa5fd13a4b3c09505060da66e3876faf73d494

One major use case is having a global position and then transforming that position to a given mag. If mag 1 is missing, there definitely needs to be a fallback somewhere. For mag 1, this is easy, since it is not ambiguous (all components of [1,1,1] have to be 1, otherwise it wouldn't be mag 1). But for other mags, one would actually need the mag information of other layers.
I'm afraid the entire resolution code still needs some refactoring. The `ResolutionInfo` class helps already a bit, but it's still not perfect (as the above issue shows). Also, `zoomedAddressToAnotherZoomStep` is used in quite a lot of places. I added this topic to https://github.com/scalableminds/webknossos/issues/4838.

frontend/javascripts/oxalis/model/bucket_data_handling/bucket.js
```diff
@@ -192,17 +200,16 @@ export class DataBucket {
...
         isVoxelOutside = true;
-        neighbourBucketAddress[dimension] -= Math.ceil(
-          -voxel[dimensionIndex] / Constants.BUCKET_WIDTH,
-        );
+        const offset = Math.ceil(-voxel[dimensionIndex] / Constants.BUCKET_WIDTH);
+        neighbourBucketAddress[dimension] -= offset;
         // Add a full bucket width to the coordinate below 0 to avoid error's
         // caused by the modulo operation used in getVoxelOffset.
-        adjustedVoxel[dimensionIndex] += Constants.BUCKET_WIDTH;
+        adjustedVoxel[dimensionIndex] += Constants.BUCKET_WIDTH * offset;
```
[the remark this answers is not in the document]
philippotto: Good idea :+1: I integrated our `mod` function from utils and also simplified some bits here. Also, the method mutated the input variable, which was not intended, I think. I fixed that too.

frontend/javascripts/oxalis/model/sagas/volumetracing_saga.js
```diff
@@ -277,13 +364,140 @@ export function* floodFill(): Saga<void> {
...
+      cellId,
+      get3DAddress,
+      numberOfSlices,
+      dimensionIndices[2],
+      shouldOverwrite,
+    );
+  }
+  currentLabeledVoxelMap = labeledVoxelMapToApply;
+
+  // Next we downsamplesample the annotation and apply it.
```
daniel-wer, line 472:
```suggestion
  // Next we downsample the annotation and apply it.
```

Unless this was a play of words, but in that case it would've fit better in the upsampling section :grin:

frontend/javascripts/oxalis/model/sagas/volumetracing_saga.js
```diff
@@ -277,13 +364,140 @@ export function* floodFill(): Saga<void> {
...
+  // sourceZoomStep will be lower than targetZoomStep
+  for (const [source, target] of pairwise(downsampleSequence)) {
```
daniel-wer, line 474: It's a little hard to compare but to me it looks like lines 436-469 are almost exactly the same as lines 472-502, except that the `downsampleSequence` is used and the number of slices is set to 1 when calling `applyVoxelMap`. If that's the case I would deduplicate this code.

frontend/javascripts/oxalis/model/sagas/volumetracing_saga.js
```diff
@@ -298,18 +512,30 @@ export function* finishLayer(
...
 
-export function* disallowVolumeTracingWarning(): Saga<*> {
+export function* ensureNoTraceToolInLowResolutions(): Saga<*> {
+  yield* take("INITIALIZE_VOLUMETRACING");
   while (true) {
-    yield* take(["SET_TOOL", "CYCLE_TOOL"]);
-    if (yield* select(state => isVolumeTracingDisallowed(state))) {
-      Toast.warning("Volume tracing is not possible at this zoom level. Please zoom in further.");
+    yield* take(["ZOOM_IN", "ZOOM_OUT", "ZOOM_BY_DELTA", "SET_ZOOM_STEP"]);
+    const isResolutionToLowForTraceTool = yield* select(state =>
```
daniel-wer, line 531:
```suggestion
    const isResolutionTooLowForTraceTool = yield* select(state =>
```

frontend/javascripts/oxalis/model/sagas/volumetracing_saga.js
```diff
@@ -298,18 +512,30 @@ export function* finishLayer(
...
+      isVolumeTraceToolDisallowed(state),
+    );
+    const isTraceToolActive = yield* select(
+      state => enforceVolumeTracing(state.tracing).activeTool === VolumeToolEnum.TRACE,
+    );
+    if (isResolutionToLowForTraceTool && isTraceToolActive) {
```
daniel-wer, line 537:
```suggestion
    if (isResolutionTooLowForTraceTool && isTraceToolActive) {
```

frontend/javascripts/oxalis/model/volumetracing/volume_annotation_sampling.js
```diff
@@ -0,0 +1,324 @@
...
+    let bucket = dataCube.getOrCreateBucket(labeledBucketZoomedAddress);
+    if (bucket.type === "null") {
+      continue;
+    }
+    bucket.markAndAddBucketForUndo();
+    let data = bucket.getOrCreateData();
+    const thirdDimensionValueInBucket = get3DAddress([0, 0])[2];
+    for (let sliceCount = 0; sliceCount < numberOfSlicesToApply; sliceCount++) {
+      if (thirdDimensionValueInBucket + sliceCount === constants.BUCKET_WIDTH) {
+        // The current slice is in the next bucket in the third direction.
```
daniel-wer, line 297: While reading this code, I'm wondering why it happens that the slice is in the next bucket in the third direction, but this doesn't happen in the x and y direction. Is this taken care of when creating the voxel map? I don't think I completely grasped the voxel map concept in conjunction with the `numberOfSlicesToApply` yet.

daniel-wer (commented):
I had a look at the three remaining large files and added some suggestions and comments. :+1: Next I'll have a look at the new changes from the last two days.

CHANGELOG.unreleased.md
```diff
@@ -11,12 +11,22 @@ For upgrade instructions, please check the [migration guide](MIGRATIONS.released
 [Commits](https://github.com/scalableminds/webknossos/compare/20.10.0...HEAD)
 
 ### Added
+<<<<<<< HEAD
```
daniel-wer, line 14: Unresolved merge error

frontend/javascripts/oxalis/model/accessors/accessor_helpers.js
```diff
@@ -0,0 +1,42 @@
...
+  So, instead of carefully decomposing all selectors to ensure maximum
+  instance-reusages, the `reuseInstanceOnEquality` can be used.
+
+  As a rule of thumb, this wrapper should be used for selectors which need
+  the entire store state.
+
+  Note that this function isn't of any use if the return type of the passed
+  function is a primitive value.
+ */
+export function reuseInstanceOnEquality<R, F: (...args: Array<any>) => R>(
```
daniel-wer, line 24: That's pretty cool! :1st_place_medal:

daniel-wer: You've only used this in one spot for now. How did you notice that this specific spot was a bottleneck? Performance tab measurement? There's probably other good candidates where this should be used :)

frontend/javascripts/oxalis/model/accessors/dataset_accessor.js
```diff
@@ -83,19 +87,32 @@ export class ResolutionInfo {
...
+    if (resolutionMaybe) {
+      return resolutionMaybe;
+    }
+
+    if (index === 0) {
+      // If the index is 0, only mag 1-1-1 can be meant.
+      return [1, 1, 1];
+    }
+
+    throw new Error("");
```
daniel-wer, line 115: Reminder that this issue exists. Could it be an option to return the isotropic resolution here as you did before for now?

frontend/javascripts/oxalis/model/accessors/dataset_accessor.js
```diff
@@ -515,6 +544,40 @@ export function getEnabledLayers(
...
+  const missingLayers = getEnabledLayers(dataset, state.datasetConfiguration)
+    .map((layer: DataLayerType) => ({
+      layer,
+      resolutionInfo: getResolutionInfo(layer.resolutions),
+    }))
+    .filter(({ resolutionInfo }) => {
+      const isMissing = !resolutionInfo.hasIndex(zoomStep);
+      if (!isMissing || renderMissingDataBlack) {
+        return isMissing;
+      }
```
daniel-wer, line 563: I'm having a little trouble wrapping my head around this, is it correct that only `isMissing` is returned without `renderMissingDataBlack` influencing the return value?

frontend/javascripts/oxalis/shaders/main_data_fragment.glsl.js
```diff
@@ -148,47 +149,58 @@ void main() {
...
+      vec3 flooredMousePosUVW = transDim(floor(globalMousePosition));
+
+      // When hovering an isosurface in the 3D viewport, the hoveredIsosurfaceId contains
+      // the hovered cell id. Otherwise, we use the mouse position to look up the active cell id.
+      // Passing the mouse position from the 3D viewport is not an option here, since that position
+      // isn't on the orthogonal planes necessarily.
+      cellIdUnderMouse = length(hoveredIsosurfaceId) > 0.1 ? hoveredIsosurfaceId : getSegmentationId(flooredMousePosUVW);
+    } else {
+
+    }
```
daniel-wer, line 168: The `else` should not be needed

frontend/javascripts/oxalis/view/action_bar_view.js
```diff
@@ -181,6 +181,7 @@ class ActionBarView extends React.PureComponent<Props, State> {
           )}
           {showVersionRestore ? VersionRestoreWarning : null}
           <DatasetPositionView />
+          <Icon type="warning" />
```
daniel-wer, line 184: What is this warning icon about?

frontend/javascripts/oxalis/view/viewport_status_indicator.js
```diff
@@ -0,0 +1,48 @@
+// @flow
```
daniel-wer, line 1: Very cool! :heart:

daniel-wer (commented):
Phew, done with the first review round. Thanks a lot to you @MichaelBuessemeyer for doing all the ground work and to @philippotto for taking over and finishing up the frontend for this feature. Really impressive work! Let's go on to testing :100:

frontend/javascripts/oxalis/view/action_bar_view.js
philippotto: Whoops, that creeped in when I experimented with different warning designs for missing magnifications ^^

frontend/javascripts/oxalis/model/accessors/dataset_accessor.js
philippotto: I checked the callers of this method and this should not be called for not-existing magnifications (except for mag 1 which is kind of a special case, since our global positioning system is always in mag 1). So, the error should not be thrown in theory. Since returning the isotropic resolution is potentially wrong, I'd prefer to simply throw the error and fix the call site once we see this error in the wild. Do you agree?

frontend/javascripts/oxalis/model/accessors/dataset_accessor.js
philippotto: I agree that this is kind of hard to read. I think I'll restructure the code a bit to make it more intuitive. Probably this:

```js
const isMissing = !resolutionInfo.hasIndex(zoomStep);
if (!isMissing) {
  // The layer exists. Thus, it is not missing.
  return false;
}

if (renderMissingDataBlack) {
  // We already know that the layer is missing. Since `renderMissingDataBlack`
  // is enabled, the fallback magnifications don't matter. The layer cannot be
  // rendered.
  return true;
}

...
```

philippotto: I'll also change `missing` to `unrenderable` since this fits better in my opinion.

frontend/javascripts/oxalis/model/sagas/volumetracing_saga.js
```diff
@@ -245,6 +311,16 @@ function* copySegmentationLayer(action: CopySegmentationLayerAction): Saga<void>
...
+    labeledVoxelMapOfCopiedVoxel,
+    labeledZoomStep,
+    dimensionIndices,
+    resolutionInfo,
+    cube,
+    activeCellId,
+    z,
+    false,
+  );
   yield* put(finishAnnotationStrokeAction());
```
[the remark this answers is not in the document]
philippotto: This is done in `api.data.labelVoxelsInAllResolutions` which is called by the brush/trace saga, so it should be covered.

frontend/javascripts/oxalis/model/accessors/accessor_helpers.js
philippotto: This is a topic, I pondered on quite a bit. Since the new `ViewportStatusIndicator` uses a complex selector, I wanted to be very careful about its performance. I tried to measure the react rendering performance with the chrome performance tools and with the react profiler, but I couldn't get good insights, unfortunately. However, via logging I figured out that the `render` method was called with high frequency. That's why I wrote this helper function which fixes this. However, I'm not super sold on this solution. As I mentioned, in the above docstring, "carefully decomposing all selectors to ensure maximum instance-reusages" might be a better approach here (but also takes more effort).

Additionally, it might make sense to re-think how we listen to the store. Right now every store change will trigger countless subscriptions and each of them check whether their selector changed (due to memoization it might not be a big deal, but I have a bad feeling about this, since the subscribers might also do other things which trigger an unnecessary re-render). Instead, one could iterate over all used selectors and only trigger the updates where necessary. For example, if 10 subscribers listen on the same selector, one would only need to compare that selector once instead of 10 times. Not super sure how much this brings, but in combination with the selector-decomposition it seems at least intuitive to me that this should perform better.

webknossos-tracingstore/app/com/scalableminds/webknossos/tracingstore/tracings/volume/VolumeTracingDownsampling.scala
philippotto: @youri-k Was this a note for yourself? Or for @fm3 ?

frontend/javascripts/oxalis/model/accessors/dataset_accessor.js
daniel-wer: Yes agreed, but I would add an error message, then :)

webknossos-tracingstore/app/com/scalableminds/webknossos/tracingstore/tracings/volume/VolumeTracingDownsampling.scala
youri-k: It was for @fm3, but we decided in person to leave it as it is.

frontend/javascripts/oxalis/model/sagas/volumetracing_saga.js
```diff
@@ -277,13 +364,140 @@ export function* floodFill(): Saga<void> {
...
+      sourceResolution,
+      sourceZoomStep,
+      targetResolution,
+      targetZoomStep,
+      dimensionIndices,
+      thirdDimensionOfSlice,
+    );
+    // Adjust thirdDimensionValue so get3DAddress returns the third dimension value
+    // in the target resolution to apply the voxelMap correctly.
+    thirdDimensionValue =
```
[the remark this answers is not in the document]
philippotto: Very good point! Also took me a while to understand what exactly is passed around here. I went for the curried solution and included the main calculation in the outer part of the helper.

frontend/javascripts/oxalis/model/sagas/volumetracing_saga.js
philippotto: I was a bit hesitant at first to do this, since I was worried that the deduplication could make it more difficult to read, but the result turned out quite nice in my opinion. Thank you for this suggestion! :rocket:

frontend/javascripts/oxalis/model/accessors/dataset_accessor.js
philippotto: Done ^^

webknossos-tracingstore/app/com/scalableminds/webknossos/tracingstore/tracings/volume/VolumeTracingDownsampling.scala
philippotto: @youri-k is this also resolved/a won't fix?

youri-k: I think it would be a nice to have, but it isn't necessary. If you want, I can make the change now.

philippotto: That's your call ;) Just wanted to check whether this can be resolved.

frontend/javascripts/test/model/volumetracing/volume_annotation_sampling.spec.js
```diff
@@ -0,0 +1,482 @@
...
+  [[5, 5], [5, 6], [6, 5], [6, 6]].forEach(([firstDim, secondDim]) =>
+    labelVoxelInVoxelMap(firstDim, secondDim, sourceVoxelMap),
+  );
+  const goalVoxelMap = getEmptyVoxelMap();
+  [[10, 10], [10, 11], [10, 12], [10, 13], [11, 10], [11, 11], [11, 12], [11, 13]].forEach(
+    ([firstDim, secondDim]) => labelVoxelInVoxelMap(firstDim, secondDim, goalVoxelMap),
+  );
+  [[12, 10], [12, 11], [12, 12], [12, 13], [13, 10], [13, 11], [13, 12], [13, 13]].forEach(
+    ([firstDim, secondDim]) => labelVoxelInVoxelMap(firstDim, secondDim, goalVoxelMap),
+  );
```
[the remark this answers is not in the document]
philippotto: Yes, it's not super straightforward. I added a docstring to the `LabeledVoxelsMap` which should explain the 2D nature a bit (which is why the out-of-bucket check is only relevant for the third dimension). Looks like this:

```
// A LabeledVoxelsMap maps from a bucket address
// to a 2D slice of labeled voxels. These labeled voxels
// are stored in a Uint8Array in a binary way (which cell
// id the voxels should be changed to is not encoded).
export type LabeledVoxelsMap = Map<Vector4, Uint8Array>;
```

I hope this helps a bit.

philippotto:
Thank you so much, @daniel-wer, for reviewing this monstrous PR! Your feedback was really helpful. I think I incorporated most of it now. Regarding the performance of the brushing, I think, I'll do this on a separate branch to not blow this PR any further :see_no_evil: I hope I'll get to it today!

hotzenklotz:
- Brushing in mag 16 failed: 
  https://performantbrush.webknossos.xyz/annotations/Explorational/5f7d835001000017009441a5#512,512,514,0,10.109
  image

- This warning comes up a lot: 
  image

- (Maybe not introduced by this PR) Moving the mouse from window to the next jumps the brush cursor: 
  Kapture 2020-10-07 at 11 29 28

- Fallback Layers fro uint64 not allowed: 
  image
  https://performantbrush.webknossos.xyz/datasets/sample_organization/TestUpload/createExplorative/hybrid/true

- This happened with one dataset. I think it was a dataset with 3 float layers. Can't find it anymore: image

- This annotation is only was traced in mag1 and mag16. (all other mags were "interpolated" correctly) After reloading, the "interpolated" mags are gone. This is an anisotropic dataset (Dataset: 2017-05-31_mSEM_aniso-test): 
  https://performantbrush.webknossos.xyz/annotations/Explorational/5f7d880c01000073009441db#800,487,46,3,1.100
  Kapture 2020-10-07 at 11 27 40



-

hotzenklotz:
Failed Test:
- create new volume annotation, brush some, download (data.zip should have bucket data in each mag)
- reupload this multi-resolution volume annotation, should have same resolution list and volume data
- open legacy or new volume annotation, click »unlink fallback segmentation«, should reload with new empty volume tracing that has all mags (union) of all present dataset layers

frontend/javascripts/oxalis/model/sagas/volumetracing_saga.js
```diff
@@ -150,45 +162,151 @@ export function* editVolumeLayerAsync(): Generator<any, any, any> {
...
+    const bucketZoomedAddress = zoomedPositionToZoomedAddress(min, labeledZoomStep);
+
+    if (currentLabeledVoxelMap.get(bucketZoomedAddress)) {
+      throw new Error("When iterating over the buckets, we shouldn't visit the same bucket twice");
+    }
+
+    const labelMapOfBucket = new Uint8Array(Constants.BUCKET_WIDTH ** 2);
+    currentLabeledVoxelMap.set(bucketZoomedAddress, labelMapOfBucket);
+
+    // globalA (first dim) and globalB (secondB) are global coordinates
```
daniel-wer, line 244:
```suggestion
    // globalA (first dim) and globalB (second dim) are global coordinates
```

fm3:
I pushed a commit (hopefully) fixing the download/reupload bugs. Did not look into anisotropic issues yet, some of that may be backend-related too.

Some other things I noticed during testing:

- [ ] brushing should be disabled when a resolution is shown that is not in the tracing (I found out I had brushed only after zooming)
- [ ] the “layer cannot be rendered” message is too verbose. also, the currently active magnification value should be mentioned. in annotation mode, maybe mention volume annotation layer (not only “segmentation”)?
- [ ] unify “magnification” vs “resolution” in front-end (I’m in favor of “resolution”)
- [ ] #4686 would be really helpful for this feature
- [ ] volume layer in layer menu should maybe include the information whether there is a fallback layer or not
- [ ] (possibly not introduced in this PR): when downloading, “The fallback volume data is excluded” should only appear in the info toast if there is a fallback layer

philippotto:
I think, I merged the master correctly into this PR, but something is still off with the continuous brush feature. I assume, there are still some spots which need to be adapted to the multi-res feature, but weirdly I mainly experience performance problems (but the output looks fine). Since it's currently a major perf bottleneck, I disabled the feature for now :(

- [ ] either re-enable continuous brushing or create a follow-up issue

philippotto:
> I can only brush in mag1 in this annotation. All other mags are forbidden. (Dataset: Test):
https://performantbrush.webknossos.xyz/annotations/Explorational/5f7d8b0601000073009441e3#4732,4996,2176,0,1.653

This is expected, since this tracing uses a fallback layer which only exists in mag 1. Therefore, the new hybrid annotation can also only exist in mag 1 (since there is a warning indicator in the viewport for this, I think this is fine).


>  This annotation is only was traced in mag1 and mag16. (all other mags were "interpolated" correctly) After reloading, the "interpolated" mags are gone. This is an anisotropic dataset (Dataset: 2017-05-31_mSEM_aniso-test):

I fixed another bug which might have solved this, too. At least I cannot reproduce this. We should doublecheck this again when the new dev instance is deployed.

> Fallback Layers fro uint64 not allowed:

@fm3 Do you have an idea how we want to deal with this? It's probably not blocking this PR (since it has been like this before), but obviously it's a problem if we want to create new annotations always with a fallback. Maybe the back-end could create the annotation without fallback and send a warning message to the user? If the user was not explicit about having the fallback (i.e., merely selected "create annotation"), the warning could be dropped altogether maybe.

fm3:
Great work, also with the recent changes! :)

> @fm3 Do you have an idea how we want to deal with this? It's probably not blocking this PR (since it has been like this before), but obviously it's a problem if we want to create new annotations always with a fallback. Maybe the back-end could create the annotation without fallback and send a warning message to the user? If the user was not explicit about having the fallback (i.e., merely selected "create annotation"), the warning could be dropped altogether maybe.

Good question. That’s issue #4852 . It is not easy to send warnings to the front-end while creating a new annotation, as there is no protocol for that. I’d say we can probably leave it as is for the moment (I changed the wording from “not allowed” to “not supported”). I don’t know how common uint64 segmentations are. If many users encounter this problem, solutions may include (1) detecting this ahead of time and avoid showing the normal generic “create annotation” button but only show the fallback-free version. (2) silently (or with a new communication protocol) convert to no-fallback, (3) actually implement uint64 support in the front-end (I think daniel mentioned this may be available now)

philippotto:
> This happened with one dataset. I think it was a dataset with 3 float layers. Can't find it anymore:

Hm, I cannot reproduce this :thinking:

> This annotation is only was traced in mag1 and mag16. (all other mags were "interpolated" correctly) After reloading, the "interpolated" mags are gone. This is an anisotropic dataset (Dataset: 2017-05-31_mSEM_aniso-test):
>
> I fixed another bug which might have solved this, too. At least I cannot reproduce this. We should doublecheck this again when the new dev instance is deployed.

Unfortunately, I can reproduce these issues now...

> I don’t know how common uint64 segmentations are. If many users encounter this problem, solutions may include (1) detecting this ahead of time and avoid showing the normal generic “create annotation” button but only show the fallback-free version. (2) silently (or with a new communication protocol) convert to no-fallback, (3) actually implement uint64 support in the front-end (I think daniel mentioned this may be available now)

I think, our team uses them from time to time. I think option (1) should be relatively easy to do, since the necessary information should be available in the dashboard if I'm not mistaken.

philippotto:
I resolved all conflicts which originated from the reworked toolbar. However, there might still be some problems, as I didn't re-test the branch. Will do so tomorrow probably.

fm3:
feedback has been that this feature should not be enabled by default for tasks. instead, the task type should set the range of resolutions the task annotations should have. This information needs to be propagated when
- [ ] creating tasks from csv
- [  ] creating individual tasks
- [ ] creating tasks from base annotations (via id)
- [ ] creating tasks from base annoatation (via zip upload)

This also means that the currently active …
- [ ] assertion that checks that an annotation with a resolutions list has the *right* resolutions list, as determined by the dataset needs to be changed, this also has to take into account the task type settings

When a task with an incomplete resolutions list is downloaded and re-uploaded as explorational, the current code will create downsampled resolutions, we should decide if this behavior should change
- [  ] decide how to handle re-uploaded tasks with incomplete resolution sets

philippotto:
I just got the following back-end exception:

```
play.api.http.HttpErrorHandlerExceptions$$anon$1: Execution exception[[ArithmeticException: / by zero]]
        at play.api.http.HttpErrorHandlerExceptions$.throwableToUsefulException(HttpErrorHandler.scala:351)
        at play.api.http.DefaultHttpErrorHandler.onServerError(HttpErrorHandler.scala:267)
        at play.core.server.AkkaHttpServer$$anonfun$1.applyOrElse(AkkaHttpServer.scala:448)
        at play.core.server.AkkaHttpServer$$anonfun$1.applyOrElse(AkkaHttpServer.scala:446)
        at scala.concurrent.Future.$anonfun$recoverWith$1(Future.scala:413)
        at scala.concurrent.impl.Promise.$anonfun$transformWith$1(Promise.scala:37)
        at scala.concurrent.impl.CallbackRunnable.run(Promise.scala:60)
        at akka.dispatch.BatchingExecutor$AbstractBatch.processBatch(BatchingExecutor.scala:55)
        at akka.dispatch.BatchingExecutor$BlockableBatch.$anonfun$run$1(BatchingExecutor.scala:92)
        at scala.runtime.java8.JFunction0$mcV$sp.apply(JFunction0$mcV$sp.java:12)
Caused by: java.lang.ArithmeticException: / by zero
        at com.scalableminds.webknossos.datastore.models.BucketPosition.<init>(Positions.scala:52)
        at com.scalableminds.webknossos.tracingstore.tracings.volume.VolumeTracingService.$anonfun$handleUpdateGroup$2(VolumeTracingService.scala:94)
        at scala.concurrent.Future.$anonfun$flatMap$1(Future.scala:303)
        at scala.concurrent.impl.Promise.$anonfun$transformWith$1(Promise.scala:37)
        at scala.concurrent.impl.CallbackRunnable.run(Promise.scala:60)
        at java.util.concurrent.ForkJoinTask$RunnableExecuteAction.exec(ForkJoinTask.java:1402)
        at java.util.concurrent.ForkJoinTask.doExec(ForkJoinTask.java:289)
        at java.util.concurrent.ForkJoinPool$WorkQueue.runTask(ForkJoinPool.java:1056)
        at java.util.concurrent.ForkJoinPool.runWorker(ForkJoinPool.java:1692)
        at java.util.concurrent.ForkJoinWorkerThread.run(ForkJoinWorkerThread.java:175)
```

I opened a hybrid tracing with fallback segmentation and brushed a bit. Not sure yet, what exactly this is causing. Do you have a spontaneous idea, @fm3, or should I investigate?

philippotto:
Here's a screenshot of the save queue content:

image

fm3:
I think there may be a mismatch between the zoomstep and the tracing’s resolution list. 
Can you reproduce it?
Could you replace VolumeTracingService.scala’s line 131 with       } else throw new Exception(s"Received bucket with zoomStep ($zoomStep), could not look up that resolution tracing.resolutions (${tracing.resolutions})")
(I suppose we should do that anyway)

Also, just to make sure: is the zoomStep’s semantic that it is the index in the (sorted) tracing resolutions list? or should we check zoomStep**2 against maxDim?

philippotto:
> Could you replace VolumeTracingService.scala’s line 131 with } else throw new Exception(s"Received bucket with zoomStep ($zoomStep), could not look up that resolution tracing.resolutions (${tracing.resolutions})")
> (I suppose we should do that anyway)

Yes, this seems like a good solution! The error was raised:

`Exception: Received bucket with zoomStep (3), could not look up that resolution tracing.resolutions (Vector(Point3D(2,2,2), Point3D(8,8,8)))`



> Also, just to make sure: is the zoomStep’s semantic that it is the index in the (sorted) tracing resolutions list? or should we check zoomStep**2 against maxDim?

Yes, this seems to be the problem. Before this PR, the `zoomStep` was always the index AND the exponent for the power-of-two-mag sequence. So, it didn't really matter. With the changes of this PR, the zoomStep should be interpreted as the exponent, so that the power-of-two is used to find the correct resolution (using maxDim). So, in this case zoomStep == 3 should map to [8, 8, 8]. Can you change this? I'll commit the fix you mentioned in the meantime :)

fm3:
I changed that. I *think* this was the only volume tracing related spot where this was done as index lookup. Data loading already did it with the power-of-two matching. Good catch!

philippotto:
Awesome! Works perfectly now :) I think, I also worked out the last kinks which were caused by the new toolbar.

philippotto (approved):
Let's roll :ferris_wheel:

Issue #4751, opened 2020-08-05
Guard against corrupted settings


philippotto:
There are multiple ways to do this:
- enforce that the layer settings match our schema (e.g., if a layer opacity is none, change it to the default layer opacity)
- catch react errors in the settings component and reset all settings to default

The case which we observed twice now is that the layer alpha value was set to null for some reason, causing wk to crash as soon as the settings sidebar was opened.

philippotto:
I think this is covered now by https://github.com/scalableminds/webknossos/pull/4845/files

PR #4758, opened 2020-08-07
Make number slider more robust against weird input values

philippotto:
This is a workaround for layer opacity values which become undefined in some circumstances. The actual cause for this is unknown to me, but making the number slider more robust should be an okay way to tackle this.

- https://___.webknossos.xyz

### Steps to test:
- I patched the code locally so that undefined is passed to the number slider and it worked as expected. One can also test it by using the front-end api to set an undefined opacity, but I think it's not crucial to reproduce this. A review should be sufficient.

### Issues:
- workaround for #4751

------
- [X] Ready for review

### DIFF
--- a/frontend/javascripts/oxalis/view/settings/setting_input_views.js
+++ b/frontend/javascripts/oxalis/view/settings/setting_input_views.js
@@ -24,17 +24,22 @@ export class NumberSliderSetting extends React.PureComponent<NumberSliderSetting
   };
 
   _onChange = (_value: number) => {
-    if (this.props.min <= _value && _value <= this.props.max) {
+    if (this.isValueValid(_value)) {
       this.props.onChange(_value);
     }
   };
+
+  isValueValid = (_value: number) => {
+    return _.isNumber(_value) && _value >= this.props.min && _value <= this.props.max;
+  }
 
   render() {
     const { value: originalValue, label, max, min, step, onChange, disabled } = this.props;
 
     // Validate the provided value. If it's not valid, fallback to the midpoint between min and max.
-    const isValueValid = _.isNumber(originalValue) && originalValue >= min && originalValue <= max;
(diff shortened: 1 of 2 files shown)


frontend/javascripts/oxalis/view/settings/setting_input_views.js
```diff
@@ -29,7 +30,11 @@ export class NumberSliderSetting extends React.PureComponent<NumberSliderSetting
   };
 
   render() {
-    const { value, label, max, min, step, onChange, disabled } = this.props;
+    const { value: originalValue, label, max, min, step, onChange, disabled } = this.props;
+
+    // Validate the provided value. If it's not valid, fallback to the midpoint between min and max.
+    const isValueValid = _.isNumber(originalValue) && originalValue >= min && originalValue <= max;
+    const value = isValueValid ? originalValue : Math.floor((min + max) / 2);
```
MichaelBuessemeyer, line 37:
```suggestion
```

philippotto:
I also add a validation check to the onblur code. Thanks for finding this bug, @MichaelBuessemeyer :pray:

frontend/javascripts/oxalis/view/settings/setting_input_views.js
```diff
@@ -24,17 +24,21 @@ export class NumberSliderSetting extends React.PureComponent<NumberSliderSetting
...
   _onChange = (_value: number) => {
-    if (this.props.min <= _value && _value <= this.props.max) {
+    if (this.isValueValid(_value)) {
       this.props.onChange(_value);
     }
   };
 
+  isValueValid = (_value: number) =>
+    _.isNumber(_value) && _value >= this.props.min && _value <= this.props.max;
+
```
MichaelBuessemeyer, line 34: Nice that you extracted the check into an own method to reduce code duplication :+1:

PR #4766, opened 2020-08-14
Allow to remove fallback layer

### LINKED ISSUE #4741: Allow to remove fallback layer from a volume tracing
Consequently, we can remove the "do you want a fallback layer or not?" decision when creating a new tracing. By default, a fallback layer is always used (if it exists on disk). If the user does not want this, they can chose to remove the fallback layer from the tracing.

For volume (or hybrid) tracings, the "Volume Layer" in the settings could have a button to remove this.

grittaweisheit:
This PR makes it possible to remove the fallback layer in a volume tracing. To do this a button in the layer settings header was added.

- https://___.webknossos.xyz

### Steps to test:
- start a volume or hybrid tracing with existing segmentation layer
- go to the dataset settings
- next to the segmentation layer's settings header should be a minus button
- on click this should open a confirm pop up
- when selecting ok the layers segmentation should be removed and the layer should be reloaded
- the minus button should disappear
- when doing a volume/hybrid tracing with a task the button should not be there

### Issues:
- fixes #4741

------
- [x] Ready for review


frontend/javascripts/oxalis/model/sagas/update_actions.js
```diff
@@ -136,6 +136,10 @@ export type RevertToVersionUpdateAction = {|
     sourceVersion: number,
   |},
 |};
+export type RemoveFallbackLayerAction = {|
+  name: "removeFallbackLayer",
+  value: {},
```
philippotto, line 141: Can the value property be removed? Or is there some code which requires this type to look that way?

grittaweisheit: The value property is necessary for the `AsServerAction<...>` conversion.

frontend/javascripts/oxalis/view/settings/dataset_settings_view.js
```diff
@@ -136,6 +138,34 @@ class DatasetSettings extends React.PureComponent<DatasetSettingsProps> {
...
+        }}
+      />
+    </Tooltip>
+  );
+
+  removeFallbackLayer = (layerName: string) => {
+    Modal.confirm({
+      title: messages["tracing.confirm_remove_fallback_layer.title"],
+      content: messages["tracing.confirm_remove_fallback_layer.content"],
+      onOk: async () => {
```
philippotto, line 162: The modal looks quite packed now with the longer text. I think, you can pass a width parameter here so that the modal gets a bit wider (e.g., `width: 600` could work).
Also, splitting the content into two paragraphs would probably be a good idea, too. Since the `messages` file doesn't contain react elements, I'd suggest to just split the `content` property into two parts. So, the usage would look like this:

```
content: <div>
  <p>messages["tracing.confirm_remove_fallback_layer.content_explanation"]</p>
  <p>messages["tracing.confirm_remove_fallback_layer.content_notes"]</p>
</div>
```

philippotto (approved):
Awesome! I left some final comment about tweaking the modal's appearance a bit. After this is done, this should be ready to be merged :)

PR #4773, opened 2020-08-19
Update flow to 0.110.0

daniel-wer:
Update Flow to 0.110 to fix a bug @MichaelBuessemeyer encountered while working on the volume undo.
Updating further will require a little more work due to https://medium.com/flow-type/spreads-common-errors-fixes-9701012e9d58 which is cool in itself, but flow now no longer permits unions in computed properties (see the last section in the link) (/cc @philippotto). This is then an error:

```
type Union = 'member1' | 'member2';
declare var computed_key: Union;
const x = {[computed_key]: 3};
```

This leads to around 50 errors in our code :/
Nevertheless, fine for now and I'll play around with it a little bit to see whether we can upgrade further :)

- https://___.webknossos.xyz

### Steps to test:
- CI is enough

------
- [x] Ready for review


frontend/javascripts/oxalis/model/sagas/effect-generators.js.flow
```diff
@@ -51,12 +51,13 @@ declare type SelectFn6<T1, T2, T3, T4, T5, T6, R> = (state: OxalisState, t1: T1,
 
 declare type SelectFn =
   & (<R, Fn: SelectFn0<R>>(selector: Fn, ...rest: Array<void>) => Generator<*, R, *>)
-  & (<T1, R, Fn: SelectFn1<T1, R>>(selector: Fn, t1: T1, ...rest: Array<void>) => Generator<*, R, *>)
-  & (<T1, T2, R, Fn: SelectFn2<T1, T2, R>>(selector: Fn, t1: T1, t2: T2, ...rest: Array<void>) => Generator<*, R, *>)
-  & (<T1, T2, T3, R, Fn: SelectFn3<T1, T2, T3, R>>(selector: Fn, t1: T1, t2: T2, t3: T3, ...rest: Array<void>) => Generator<*, R, *>)
-  & (<T1, T2, T3, T4, R, Fn: SelectFn4<T1, T2, T3, T4, R>>(selector: Fn, t1: T1, t2: T2, t3: T3, t4: T4, ...rest: Array<void>) => Generator<*, R, *>)
-  & (<T1, T2, T3, T4, T5, R, Fn: SelectFn5<T1, T2, T3, T4, T5, R>>(selector: Fn, t1: T1, t2: T2, t3: T3, t4: T4, t5: T5, ...rest: Array<void>) => Generator<*, R, *>)
-  & (<T1, T2, T3, T4, T5, T6, R, Fn: SelectFn6<T1, T2, T3, T4, T5, T6, R>>(selector: Fn, t1: T1, t2: T2, t3: T3, t4: T4, t5: T5, t6: T6, ...rest: Array<void>) => Generator<*, R, *>)
+  // We are only using SelectFn0 and the others lead to flow errors
```
daniel-wer, line 54: This was causing errors like:

```
Error ┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈ frontend/javascripts/oxalis/model/sagas/isosurface_saga.js:121:32

Could not decide which case to select, since case 1 [1] may work but if it doesn't case 2 [2] looks promising too. To
fix add a type annotation to state [3].

     frontend/javascripts/oxalis/model/sagas/isosurface_saga.js
     118│ // This function either returns the activeCellId of the current volume tracing
     119│ // or the view-only active cell id
     120│ function* getCurrentCellId(): Saga<number> {
 [3] 121│   const volumeTracing = yield* select(state => state.tracing.volume);
     122│   if (volumeTracing != null) {
     123│     return volumeTracing.activeCellId;
     124│   }

     frontend/javascripts/oxalis/model/sagas/effect-generators.js.flow
 [1]  53│   & (<R, Fn: SelectFn0<R>>(selector: Fn, ...rest: Array<void>) => Generator<*, R, *>)
      54│   // We are only using SelectFn0 and the others lead to flow errors
 [2]  55│   & (<T1, R, Fn: SelectFn1<T1, R>>(selector: Fn, t1: T1, ...rest: Array<void>) => Generator<*, R, *>)
```

which I do not understand. I don't see the ambiguity that flow complains about. Case [2] has more arguments and we only give 1 - the select function. That's why I commented out the useless type annotations which lead to this error.

MichaelBuessemeyer: I also do not understand the error and think this is a valid fix :+1:

MichaelBuessemeyer (approved):
@daniel-wer Thanks for updating our flow version :tada:

As the CI passes, this PR looks like it is ready to be merged.

PR #4774, opened 2020-08-20
Fix nightly test

philippotto:
This PR updates the screenshots for the nightly test which apparently changed due to #4768. Strictly speaking, the quaternions shouldn't have changed the default orientation, but I think it's not worth to investigate the subtle differences further.


daniel-wer (changes requested):
Thanks for updating, I noticed two things though:

- The new screenshot for `frontend/javascripts/test/screenshots/fluro-rgb_knossos.png` doesn't look right, it only shows an error message.
- The circle ci artifacts diff shows changes in the font. In the past I had this problem when I regenerated the screenshots using docker (`yarn refresh-screenshots`). Did you also use this command in the past? If I remember correctly `yarn test-screenshot` didn't have this problem (that's what the CI uses). Or you could grab the screenshots from the circle ci artifacts.

philippotto:
> The new screenshot for frontend/javascripts/test/screenshots/fluro-rgb_knossos.png doesn't look right, it only shows an error message.

Just fixed it :)

> The circle ci artifacts diff shows changes in the font. In the past I had this problem when I regenerated the screenshots using docker (yarn refresh-screenshots). Did you also use this command in the past? If I remember correctly yarn test-screenshot didn't have this problem (that's what the CI uses). Or you could grab the screenshots from the circle ci artifacts.

Yeah, I noticed this, too. However, I already picked the artifacts from the circle ci artifacts :thinking:  Maybe the font changes are so subtle that they are tolerated by our tests?

daniel-wer:
Nice!

> Yeah, I noticed this, too. However, I already picked the artifacts from the circle ci artifacts 🤔 Maybe the font changes are so subtle that they are tolerated by our tests?

That could very well be :)

`frontend/javascripts/test/screenshots/2017-05-31_mSEM_aniso-test.new.png.1` seems to have been committed by accident or you forgot to rename it.

daniel-wer (approved):
:+1:

PR #4778, opened 2020-08-20
UI improvement for 2D data

### LINKED ISSUE #4738: UI improvements for 2D data
The wk UI is optimized for 3D data and works very well for that. When using 2D data, the UI is not optimal. With a couple of small changes we could fix that:

- [ ] Show only one viewport (XY)
- [ ] Disable moving in Z

In order to activate these modes, the depth of the bounding box could be detected.

See also #4736

grittaweisheit:
This PR enables only the xy viewport and disables z-movement via keyboard input when navigating 2d datasets

- https://___.webknossos.xyz

### Steps to test:
- open a dataset with a depth of 1 voxel
- there should be only one viewport 
- when using the d- or f-key nothing should happen

### Issues:
- fixes #4738

------
- [x] Ready for review


grittaweisheit:
The decision weather to use a 2d layout is currently made solely with the dataset's depth. I'm not familiar enough with the common use cases to know if this is reasonable, but it seemed to be a save choice. Is it sensible to switch to 2D layouts when tracing a task with a bounding box that's 2-dimensional?

Furthermore I currently only disabled the z-movement with keyboard input. Should it be absolutely impossible to move on the z-axis?

philippotto (approved):
Awesome! Works very well :+1:

Please add the change to the changelog before merging :)

PR #4784, opened 2020-08-28
Shortcuts for navigating to next/prev node

### LINKED ISSUE #4147: Add shortcuts for navigating to next/prev node
Right now, it is very hard to "navigate" along a given tree. There is no easy way to to jump/navigate to the next node within a tree. Only comment nodes can be accessed directly. Selected node in the 3D view is a workaround, but not great.

I propose to have a keyboard shortcut (and perhaps a button?) to navigate to the next or previous node in a tree/skeleton. A primary use case, is the inspection of auto-generated  skeletons.

(We already have buttons/shortcuts to navigate between the trees. This features should offer the same functionality for nodes.)

grittaweisheit:
This PR makes it possible to navigate between nodes in a skeleton tracing via "," and "ctr + ." .

- https://navigatenodes.webknossos.xyz

### Steps to test:
- open a skeleton tracing
- have some nodes
- navigate with "," to the preceding, or with "ctr + ." to the subsequent node
- when on a branch point the node with the highest (subsequent) / lowest (preceding) id is selected 
- if navigating back over a branch that was already navigated over, the path that was already taken is chosen again 

### Issues:
- fixes #4147

------
- [x] Ready for review


frontend/javascripts/oxalis/controller/combinations/skeletontracing_plane_controller.js
```diff
@@ -144,6 +146,74 @@ export function moveNode(dx: number, dy: number) {
   );
 }
 
+function otherNodeOfEdge(edge: Edge, nodeId: number): number {
+  return edge.source === nodeId ? edge.target : edge.source;
+}
+function getSubsequentNodeFromTree(tree: Tree, node: Node): number {
+  const nodes = tree.edges.getEdgesForNode(node.id).map(edge => otherNodeOfEdge(edge, node.id));
+  const next = Math.max(node.id, ...nodes);
```
daniel-wer, line 154: Why do you put `node.id` in there as well? I would say only the node ids of the neighbors need to be checked. Otherwise users will get stuck in places where the current node has a higher id than all of its neighbors. (Same for the `getPrecedingNodeFromTree` function)

daniel-wer (changes requested):
Very nice, I like this feature a lot and you did a good job implementing it, it's already a pleasure to use :tada: 
There are a couple of edge-cases which I annotated and I suggested a more edge-case-resistant approach to detect invalid navigation lists :)

frontend/javascripts/oxalis/model/actions/skeletontracing_actions.js
```diff
@@ -189,6 +195,7 @@ export const SkeletonTracingSaveRelevantActions = [
   "TOGGLE_TREE_GROUP",
   "TOGGLE_ALL_TREES",
   "TOGGLE_INACTIVE_TREES",
+  "UPDATE_NAVIGATION_LIST",
```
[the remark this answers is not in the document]
grittaweisheit: Cool, thanks!

frontend/javascripts/oxalis/controller/combinations/skeletontracing_plane_controller.js
grittaweisheit: Otherwise there can be loops. For example if we have three nodes (1, 2, 3) and 2 is active. Then we move forward, so 3 is active and the list is [2, 3]. Now we move to the next again. Because there is no next in the list we search for it in the tree. If we don't consider 3 we move to 2 and have the list [2, 3, 2]. Moving forward again extends this even further. And when moving back you would jump between nodes 2 and 3 several times before arriving at node 1.
So both ways lead to problems.

daniel-wer: What about checking what the last node was (depending on the direction the user is going with the shortcuts) by looking it up in the list (if one exists) and then making sure to exclude that node when looking for the new node to travel to? Does that make sense?

grittaweisheit: I think that should work :)

frontend/javascripts/oxalis/controller/combinations/skeletontracing_plane_controller.js
```diff
@@ -229,6 +303,7 @@ function onClick(
       );
     } else {
       Store.dispatch(setActiveNodeAction(nodeId));
+      Store.dispatch(updateNavigationListAction([nodeId], 0));
```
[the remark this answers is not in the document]
grittaweisheit: I adjusted it a little bit and just check it in advance by comparing the active node with the list's node at the active index.

frontend/javascripts/oxalis/controller/combinations/skeletontracing_plane_controller.js
```diff
@@ -149,25 +149,40 @@ export function moveNode(dx: number, dy: number) {
...
+    .filter(id => id !== excludedId);
+  const prev = nodes.length ? Math.min(...nodes) : node.id;
   return prev;
 }
 
 function toSubsequentNode(): void {
   const tracing = enforceSkeletonTracing(Store.getState().tracing);
   const { navigationList, activeNodeId, activeTreeId } = tracing;
+  let isValidList = true;
+  if (activeNodeId == null || activeNodeId === undefined) return;
```
daniel-wer, line 173:
```suggestion
  if (activeNodeId == null) return;
```
The `== null` check checks for `null` and `undefined`.

frontend/javascripts/oxalis/controller/combinations/skeletontracing_plane_controller.js
```diff
@@ -149,25 +149,40 @@ export function moveNode(dx: number, dy: number) {
...
+
+  if (
+    activeNodeId !== navigationList.list[navigationList.activeIndex] ||
+    !navigationList.list.length
+  ) {
+    isValidList = false;
+  }
```
daniel-wer, line 180: You could get rid of the `let isValidList` by directly assigning the result of the comparison here, like:

```javascript
const isValidList = navigationList.list.length && activeNodeId === navigationList.list[navigationList.activeIndex];
```

daniel-wer: Same in the other function.

frontend/javascripts/oxalis/controller/combinations/skeletontracing_plane_controller.js
```diff
@@ -144,6 +146,113 @@ export function moveNode(dx: number, dy: number) {
...
+    Store.dispatch(setActiveNodeAction(navigationList.list[navigationList.activeIndex + 1]));
+    Store.dispatch(updateNavigationListAction(navigationList.list, navigationList.activeIndex + 1));
+  } else {
+    // search for subsequent node in tree
+    const { tree, node } = getNodeAndTree(tracing, activeNodeId, activeTreeId)
+      .map(([maybeTree, maybeNode]) => ({ tree: maybeTree, node: maybeNode }))
+      .getOrElse({
+        tree: null,
+        node: null,
+      });
```
daniel-wer, line 197: Maybe simplify this to

```javascript
const [ tree, node ] = getNodeAndTree(tracing, activeNodeId, activeTreeId)
  .getOrElse([null, null]);
```
Same in the other method.

daniel-wer (commented):
Works like a charm in my tests, very cool!

I made some suggestions to simplify the code. I'm not too happy about the code duplication, but as the two cases differ in quite a few lines I don't see an easy way to unify the two functions. So I'd say we should aim to simplify the methods as much as possible but keep them separate :)

Please also add the two new shortcuts to the list of keyboard shortcuts (`keyboard_shortcuts.md`)

frontend/javascripts/oxalis/controller/combinations/skeletontracing_plane_controller.js
grittaweisheit: That's pretty neat! Thanks :)

frontend/javascripts/oxalis/controller/combinations/skeletontracing_plane_controller.js
grittaweisheit: Flow is not happy with that :( . Maybe I could extract it in an extra method so it at least doesn't appear twice?

daniel-wer: :/ Yes you could put it into a `getNodeAndTreeOrNull` method in the skeleton tracing accessor file or you could annotate my proposed solution with a `// $FlowFixMe` to silence flow (as we know that the code is correct).

daniel-wer (approved):
Well done, lgtm :+1:

PR #4802, opened 2020-08-31
Histogram float range shows decimal places

### LINKED ISSUE #4753: Histogram float range should show decimal places
if the dataset brightness range is from 0.0 to 0.1, both the tooltip and the text boxes should support inputting and displaying the appropriate values

grittaweisheit:
- https://___.webknossos.xyz

### Steps to test:
- view or trace a dataset with at least one color layer
- manually adjust the borders of a histogram's range to a value with decimal places
- this should be possible
- when adjusting the range with the slider there should be decimal values displayed

### Issues:
- fixes #4753

------
- [x] Ready for review


grittaweisheit:
This changes the current appearance of the displayed values a little bit. Now the inserted values are rounded only after the input field loses focus and not right at the moment of input. Furthermore the two decimal places are always displayed.

philippotto:
> Furthermore the two decimal places are always displayed.

Is it possible to keep the formatter function in addition to the precision parameter? Then, one could clip the decimal digits, if the number is an integer? Would definitely look nicer. If it's not possible, it's also fine, but I think we should give this a try :)

grittaweisheit:
I'm pretty sure it's not possible, because the formatter is applied immediately. So it's not possible to differentiate between unfinished input and input with trailing zeroes.
The only way I see to make this happen is to refrain from using onChange and only apply the input values on enter or blur. Then we could be sure the input is finished and can be formatted to our liking.

philippotto (approved):
> I'm pretty sure it's not possible, because the formatter is applied immediately. So it's not possible to differentiate between unfinished input and input with trailing zeroes.

Ok, then let's go with this :)

PR #4803, opened 2020-08-31
make decimal position input possible

### LINKED ISSUE #4777: Position input should accept floats
Sometimes, one wants to paste a decimal position (e.g., `0.5, 3.3, 4.4`) from somewhere into webKnossos which is not supported. However, our internal position is stored as floats, so it would make sense to support this input format. Proposed approach:

- correctly parse float positions
- also jump to that precise float position
- on blur, render the position as integers (as it's done currently)

grittaweisheit:
- https://___.webknossos.xyz

### Steps to test:
- open a dataset
- paste a visibly different position with decimal places to the position input 
- the view should jump to that position
- on refocusing on the input field the decimal places should vanish

### Issues:
- fixes #4777

------
- [x] Ready for review


grittaweisheit:
I think currently the position is only adjusted to integers when refocusing and not on blur. That's why I did it the same way for now. If we want it, I could probably change it to update the display on blur.

philippotto:
Works great!

> I think currently the position is only adjusted to integers when refocusing and not on blur. That's why I did it the same way for now. If we want it, I could probably change it to update the display on blur.

Yes, that would be better, I think. Could you give it a try and change it so that it's adjusted to integer on blur? :)

philippotto:
> I think currently the position is only adjusted to integers when refocusing and not on blur

I think,you can change this in general for all adjustments of the position input. So, for example, when inputting an invalid position value, the value is only adjusted when refocusing, but this should also happen on blur.

grittaweisheit:
Now the value is also adjusted on blur.

philippotto (commented):
Awesome :+1: Please add a changelog entry before merging :scroll:

philippotto (approved):

PR #4805, opened 2020-09-01
do not try to decompress reverted buckets

### LINKED ISSUE #4801: broken volume data stream
This happened for a single volume annotation only (so far): viewing it works fine, but duplicate fails with a LZ4 decompression exception and download fails silently (containing 0 byte long data.zip). Is there broken data in the database?
We may need to deploy a new version with more debugging output.

fm3:
The revertToVersion update action of volume tracings overwrites buckets that are empty in the old version with a single zero byte to indicate that there is no data there, without actually deleting the old versions.

This special bucket value was assumed to be a compressed bucket by the loading algorithm, which subsequently failed to decompress it. However, this special value is not compressed and should thus just be passed as is. It is then correctly interpreted as an absent bucket (and the fallback layer data is loaded).

- https://___.webknossos.xyz

### Steps to test:
- create volume tracing, brush some, save.
- revert to previous version (e.g. version 0), reload
- the fallback layer should be rendered again at those areas
- tracing should be downloadable (some zero-sized wkw files are created there, which is not ideal, but this is the old behavior too)

### Issues:
- fixes #4801 

------
- [x] Ready for review

### DIFF
--- a/CHANGELOG.unreleased.md
+++ b/CHANGELOG.unreleased.md
@@ -18,7 +18,7 @@ For upgrade instructions, please check the [migration guide](MIGRATIONS.released
 -
 
 ### Fixed
-- 
+- Fixed a bug where volume tracings some that had been reverted to a previous version became un-downloadable. [#4805](https://github.com/scalableminds/webknossos/pull/4805)
 
 ### Removed
 -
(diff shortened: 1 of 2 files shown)


philippotto (approved):
Kudos for debugging this :+1: Didn't test, but the change looks straightforward and I assume you tested this, too.

daniel-wer:
Wow, nice find @fm3! :1st_place_medal:

PR #4822, opened 2020-09-10
Continuous brush strokes

### LINKED ISSUE #4785: Ensure brush strokes are continuous
Fast mouse movements will result in spotty brush strokes, since we simply brush circles for every captured mouse position. This can be very annoying (especially, when trying to split a cell).
A simple solution to this problem should be to connect multiple mouse positions with a rectangle and fill that rectangle in addition to the circles.

See this picture:
image

If the two adjacent mouse positions are very close (smaller than the radius maybe?), one could skip the rectangle filling.

The implementation should be fast (probably some kind of scan-line approach).

grittaweisheit:
This PR ensures that brush strokes are continuous by drawing a rectangle between two circles when necessary.

- https://continuousbrushstrokes.webknossos.xyz

### Steps to test:
- open a volume tracing
- select the brush tool
- brush fast
- there should be no gaps in your stroke

### Issues:
- fixes #4785 

------
- [x] Ready for review


grittaweisheit:
Sometimes there is a gap between the very first and second circle of a stroke. I'm not sure why that happens and how to fix it.

daniel-wer:
> Sometimes there is a gap between the very first and second circle of a stroke. I'm not sure why that happens and how to fix it.

@valentin-pinkau observed that the gap was always there when the mouse was already moving before the brushing started which gave me a clue to where the bug was originating from. The first circle is painted as a result of the `startEditingAction` and all following circles (and rectangles) are painted as a result of subsequent `addToLayerAction`s. As a result, only the very first rectangle was not painted. I fixed that by initializing `lastPosition` not with `null` but with `startEditingAction.position` :)

frontend/javascripts/oxalis/model/sagas/volumetracing_saga.js
```diff
@@ -110,13 +110,17 @@ export function* editVolumeLayerAsync(): Generator<any, any, any> {
...
+    let lastPosition = startEditingAction.position;
     while (true) {
       const { addToLayerAction, finishEditingAction } = yield* race({
         addToLayerAction: _take("ADD_TO_LAYER"),
         finishEditingAction: _take("FINISH_EDITING"),
       });
 
-      if (finishEditingAction) break;
+      if (finishEditingAction) {
+        lastPosition = null;
```
daniel-wer, line 121: Does `lastPosition` actually need to be set to `null` here? I would expect not, because we break out of the loop anyways.

frontend/javascripts/oxalis/model/sagas/volumetracing_saga.js
```diff
@@ -133,14 +137,24 @@ export function* editVolumeLayerAsync(): Generator<any, any, any> {
       }
       if (activeTool === VolumeToolEnum.BRUSH) {
         const currentViewportBounding = yield* call(getBoundingsFromPosition, activeViewport);
+        const rectangleIterator = lastPosition
```
daniel-wer, line 140: If `lastPosition` is never `null` you can then also remove this check.

frontend/javascripts/oxalis/model/volumetracing/volumelayer.js
```diff
@@ -238,6 +239,130 @@ class VolumeLayer {
...
+      const norm = this.normOfVector2(perpendicularVector);
+      perpendicularVector = this.scalarMultiplication(perpendicularVector, 1 / norm);
+    }
+    return perpendicularVector;
+  }
+
+  normOfVector2(vector: Vector2): number {
+    let norm = 0;
+    let i;
+    for (i of Vector2Indicies) {
```
daniel-wer, line 259: To be in line with the remainder of our code and to avoid `let` variables, I would replace this with 
```suggestion
    for (const i of Vector2Indicies) {
```
which is easier to reason about. Would be great if you could change this for alle the vector2 methods you created.

frontend/javascripts/oxalis/model/volumetracing/volumelayer.js
```diff
@@ -238,6 +239,130 @@ class VolumeLayer {
...
+  vector2Distance(pos1: Vector2, pos2: Vector2): number {
+    let distance = 0;
+    let i;
+    for (i of Vector2Indicies) {
+      distance += Math.pow(pos2[i] - pos1[i], 2);
+    }
+    return Math.sqrt(distance);
+  }
+
+  vectorWithScale(vector: Vector2, scale: Vector2): Vector2 {
```
daniel-wer, line 292: As all these new methods are for Vector2, I would suggest to use a more similar naming pattern for all of them and let all of the method names include the term `vector2`.

daniel-wer (commented):
This is great! Nice implementation and a biiiiiiig improvement to volume annotations, feels awesome to volume trace now :+1: 

Some cosmetic suggestions and then this should be ready to go :)

frontend/javascripts/oxalis/model/sagas/volumetracing_saga.js
```diff
@@ -133,14 +137,24 @@ export function* editVolumeLayerAsync(): Generator<any, any, any> {
...
+          ? currentLayer.getRectangleVoxelIterator(
+              lastPosition,
+              addToLayerAction.position,
+              currentViewportBounding,
+            )
+          : null;
+        if (rectangleIterator) {
+          yield* call(labelWithIterator, rectangleIterator, contourTracingMode);
```
philippotto, line 148: This means that the rectangle is always drawn between two points, right? In the original issue description, I put the idea that the rectangle drawing might be only necessary if the distance between the two brush positions is greater than a certain threshold. This could save the entire rectangle drawing if the user moves the mouse slowly (--> better performance).
Maybe you could try this out? E.g., threshold = radius or threshold = radius/10. There will definitely be a trade-off between how the brush stroke will look and how much time is saved. Maybe it doesn't save much performance, but might be worth to measure.

grittaweisheit: The decision whether to paint the rectangle or not is made before drawing the rectangle, but in the drawing method. I put it there so that I could reuse some of the necessary calculations when drawing the rectangle.

daniel-wer (commented):
Thanks for applying the feedback!
I just noticed one more issue during my final testing. When drawing a closed circle with the brush, the inner area of the circle is filled automatically. However, when drawing a closed circle with the brush **quickly**, the inner area is not filled. I don't quite remember how the filling works, but it probably only knows about the brush circles and not the rectangles now. Could you have a look whether you can spot how to fix it? If the code is too obscure let me know and I'll also have a look :)

Also, this cool improvement deserves a changelog entry :tada:

frontend/javascripts/oxalis/model/volumetracing/volumelayer.js
```diff
@@ -304,6 +304,27 @@ class VolumeLayer {
     return map;
   }
 
+  rectangleToCircles(
```
daniel-wer, line 307: Maybe `getRectangleBetweenCircles` or so would be a more descriptive name?

frontend/javascripts/oxalis/model/volumetracing/volumelayer.js
```diff
@@ -324,32 +345,36 @@ class VolumeLayer {
...
     const setMap = (x, y) => {
       map[x][y] = true;
     };
-    Drawing.fillRectangle(ax, ay, bx, by, cx, cy, dx, dy, setMap);
+
+    // translate the coordinates so the containing box originates in (0|0)
+    const [diffX, diffY] = [
+      Math.floor(Math.min(xa, xb, xc, xd)),
+      Math.floor(Math.min(ya, yb, yc, yd)),
+    ];
```
daniel-wer, line 367: Isn't this the same as `minCoord2d`?

```suggestion
    const [diffX, diffY] = minCoord2d;
```

frontend/javascripts/oxalis/model/volumetracing/volumelayer.js
```diff
@@ -408,11 +433,17 @@ class VolumeLayer {
     let p2;
     for (let i = 0; i < contourList.length; i++) {
       p1 = this.get2DCoordinate(contourList[i]);
-
+      p2 = this.get2DCoordinate(contourList[(i + 1) % contourList.length]);
```
daniel-wer, line 436: With this change, brush circles will be closed automatically now (in the same way the trace tool closes the contour in the end), because for the last value of `i`, `p1` will be the last entry of the `contourList` and `p2` will be the first entry of the `contourList` (see the modulo). However in the brushing case, we don't want to do that, we only want to fill the structure if it was closed by the user, not auto-close it.
I would suggest to check whether `i !== contourList.length - 1` in line 445 and only call the `fillRectangle` function if that's the case and add a comment explaining why.

daniel-wer (commented):
There's a small issue remaining where brush stroked structures are auto-closed now, but apart from that looks very good to me :)

frontend/javascripts/oxalis/model/volumetracing/volumelayer.js
grittaweisheit: That's a way better name, thanks!

frontend/javascripts/oxalis/model/volumetracing/volumelayer.js
grittaweisheit: Oh, yes, it is!

daniel-wer (approved):
Awesome, great feature, LGTM :+1:

PR #4826, opened 2020-09-14
Call Long-Running Jobs from webKnossos

### LINKED ISSUE #4337: Integrated cubing
For new users, one of the key challenge is to get data into webKnossos. To make things easier, we should integrate cubing for tiff files. This would be geared towards smaller datasets (≤1 GB). Could be an external service based on wkcuber.

### LINKED ISSUE #4643: Allow executing long running jobs
Tasks:
- [x] Deploy celery with workers https://github.com/scalableminds/kube/pull/281
- [x] Build cubing task (see https://github.com/scalableminds/webknossos-worker)
- [x] Upload tiff stack to temp dir in wk
- [x] Submit tasks from wk
- [x] (Simple) Progress UI in wk
- [ ] Authentication for workers

Use cases:
- importing/converting datasets #5380
- exporting datasets #4622
- merging volume tracings
- apply merger mode #4603

fm3:
Run long-running jobs converting tiff stacks to wkw, via https://github.com/scalableminds/webknossos-worker

### Steps to test:

- set `jobsEnabled = true` in `features` block in `conf/application.conf`
- flower and celery need to run and have the `tiff_cubing` task registered, compare https://github.com/scalableminds/webknossos-worker
- During dataset upload, check “needs conversion” for zipped tif stack dataset.
- After upload you should be redirected to new page Administration→Jobs, where cubing job should be listed. Page should auto-refresh.
- After successful cubing job, a view link to the now usable dataset should be shown.

- This uses new backend api routes `/jobs/run/cubing`, `/jobs`, `/jobs/:id/status` 

### Issues:
- fixes #4643 
- fixes #4337 

------
- [x] Ready for review


fm3:
an open question is how to pass the dataset scale to the cuber. I see two options:
(1) required field in the upload form if needs-cubing checkbox is checked
(2) assume (1,1,1), let the user edit later from wk edit view (may give away the chance for anisotropic downsampling though)
@normanrz what would you prefer?

fm3:
other todos I noticed
- [ ] add explaining info text to tooltip for needs-conversion checkbox in upload view
- [ ] add feature flag to enable/disable this in instances (or enable only for isDemoInstance?)
- [ ] auto-refresh for jobs list (every 3 seconds?)
- [ ] add action button to view dataset once conversion is done

normanrz:
> (1) required field in the upload form if needs-cubing checkbox is checked
(2) assume (1,1,1), let the user edit later from wk edit view (may give away the chance for anisotropic downsampling though)
@normanrz what would you prefer?

I'd prefer (1), because the scale has an effect on the downsampling (anisotropic)

fm3:
@grittaweisheit thanks for the updates, the auto-refresh is really cool :) 

We should figure out a way to pass the scale the user entered in the form to the start job request. Options I can think of
- (1) add url param to cubing request like `?scale=11.24,11.24,24`
- (2) make cubing request POST and require json input data like `{"scale":[11.24,11.24,11.24]}`
- (3) leave it in upload form request, backend has to save scale to file / db to use it in cubing request

If possible I’d avoid option (3). Which of the first two would you prefer @grittaweisheit or do you have a different idea?

grittaweisheit:
>     * (1) add url param to cubing request like `?scale=11.24,11.24,24`
> 
>     * (2) make cubing request POST and require json input data like `{"scale":[11.24,11.24,11.24]}`
> 
>     * (3) leave it in upload form request, backend has to save scale to file / db to use it in cubing request

They seem equally suitable to me. I think (1) is a little easier to implement, but that's probably just my preference.

fm3:
cool, let’s go ahead with that then :)

fm3:
@youri-k Could you already have a look at the backend changs? I think the front-end should also be almost ready for review, I think the only thing left is that the
- [ ] frontend should show new features only if feature `jobsEnabled` flag.

app/controllers/JobsController.scala
```diff
@@ -0,0 +1,178 @@
...
+        celeryInfoJson <- rpc(s"${wkConf.Jobs.Flower.uri}/api/tasks")
+          .withBasicAuth(wkConf.Jobs.Flower.username, wkConf.Jobs.Flower.password)
+          .getWithJsonResponse[JsObject]
+        celeryInfoMap <- celeryInfoJson
+          .validate[Map[String, JsObject]] ?~> "Could not validate celery response as json map"
+        _ <- Fox.serialCombined(celeryInfoMap.keys.toList)(jobId =>
+          jobDAO.updateCeleryInfoByCeleryId(jobId, celeryInfoMap(jobId)))
+      } yield ()
+      for {
+        updateBox <- updateResult.futureBox
```
youri-k, line 101: I'd prefer using a map here.
```
updateResult.futureBox.map {
    case ... =>
```

app/controllers/JobsController.scala
```diff
@@ -0,0 +1,178 @@
...
+  private val celeryInfosMinIntervalMillis = 3 * 1000 // do not fetch new status more often than once every 3s
+
+  def updateCeleryInfos(): Future[Unit] =
+    if (celeryInfosLastUpdated > System.currentTimeMillis() - celeryInfosMinIntervalMillis) {
+      Future.successful(())
+    } else {
+      val updateResult = for {
+        _ <- Fox.successful(celeryInfosLastUpdated = System.currentTimeMillis())
+        celeryInfoJson <- rpc(s"${wkConf.Jobs.Flower.uri}/api/tasks")
+          .withBasicAuth(wkConf.Jobs.Flower.username, wkConf.Jobs.Flower.password)
```
youri-k, line 93: What do you think about extracting this `withBasicAuth` call?

youri-k (commented):
Backend changes look good to me. We just need to make sure that there are no incompatibilities with the resumable uploads.

grittaweisheit:
The frontend should be ready now, too. @philippotto could you have a look that?

frontend/javascripts/admin/job/job_list_view.js
```diff
@@ -0,0 +1,162 @@
...
+                title="Created at"
+                key="createdAt"
+                render={job => new Date(job.createdAt).toUTCString()}
+                sorter={Utils.compareBy(typeHint, job => job.createdAt)}
+              />
+              <Column
+                title="State"
+                dataIndex="state"
+                key="state"
+                sorter={Utils.localeCompareBy(typeHint, job => job.state)}
```
philippotto, line 127: Could you format the `state` from "SUCCESS" to "Success" (etc.)? The capslock formatting looks a bit rough :)

frontend/javascripts/admin/job/job_list_view.js
```diff
@@ -0,0 +1,162 @@
...
+  componentWillUnmount() {
+    clearTimeout(this.intervalID);
+  }
+
+  async fetchData(): Promise<void> {
+    const jobs = await getJobs();
+    this.setState(
+      {
+        isLoading: false,
+        jobs: jobs || [],
```
philippotto, line 65: In which case is `|| []` necessary? Shouldn't `getJobs` always yield an array?

philippotto (changes requested):
Awesome stuff! The front-end code looks very good :+1: Only added a couple of small suggestions :)

conf/application.conf
```diff
@@ -209,6 +210,14 @@ silhouette {
   }
 }
 
+jobs {
+  flower {
+    uri = "http://localhost:8888"
+    username = "user1"
```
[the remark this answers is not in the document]
grittaweisheit: I honestly don't know. Maybe @fm3 knows more about that?

frontend/javascripts/navbar.js
```diff
@@ -108,6 +108,11 @@ function AdministrationSubMenu({ collapse, ...menuProps }) {
       <Menu.Item key="/taskTypes">
         <Link to="/taskTypes">Task Types</Link>
       </Menu.Item>
+      {features().jobsEnabled && (
+        <Menu.Item key="/jobs">
+          <Link to="/jobs">Jobs</Link>
```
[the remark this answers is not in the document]
grittaweisheit: Fair point, I didn't think about that!

conf/application.conf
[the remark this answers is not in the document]
fm3: yes it needs to be set, but I’d argue that a default value is not harmful. if the user enables the jobs api but not sets the correct password here, starting jobs will fail with authentication errors. This is also how we do it for e.g. postgres

frontend/javascripts/messages.js
```diff
@@ -179,6 +179,7 @@ instead. Only enable this option if you understand its effect. All layers will n
     "The dataset <%- datasetName %> was successfully deleted on disk. Redirecting to dashboard...",
   ),
   "task.no_tasks_to_download": "There are no tasks available to download.",
+  "job.start": "Do you really want to start this job?",
```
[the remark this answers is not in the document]
grittaweisheit: Oh, no, I only forgot to remove it after removing the start action for jobs.

frontend/javascripts/admin/job/job_list_view.js
grittaweisheit: Is all lower case okay, too?

frontend/javascripts/admin/job/job_list_view.js
grittaweisheit: Oh, yes. It's not necessary at all!

frontend/javascripts/admin/job/job_list_view.js
philippotto: I think, capitalized would be nicer. You can use `_.capitalize` for that :)

grittaweisheit: That's neat, thanks!

philippotto:
>  frontend should show new features only if feature jobsEnabled flag.

I think, this is done, right?

frontend/javascripts/admin/admin_rest_api.js
```diff
@@ -742,6 +743,28 @@ export async function getDatasets(
...
 }
 
+export async function getJobs(): Promise<Array<APIJob>> {
+  const jobs = await Request.receiveJSON("/api/jobs");
+  assertResponseLimit(jobs);
+  return jobs.map(job => ({
+    id: job.id,
+    type: job.command,
+    datasetName: job.commandArgs.kwargs.dataset_name,
+    state: _.capitalize(job.celeryInfo.state.toLowerCase()),
```
philippotto, line 753: I think, it would be better to not call the `capitalize` function here. It is only needed for formatting purposes during rendering, which is why I wouldn't change the internal representation here (usually, it's nice to have the server's and the client's representation as similar as possible).

Instead, you can use the `render={(__, job: APIJob) => ...` prop in the State-`<Column />` in the table to do the capitalization there.

grittaweisheit: Should the toLowerCase consequently happen there, too?

philippotto: Ah, yes, sorry, I forgot to mention this. Both should be moved so that this line is only `state: job.celeryInfo.state`.

philippotto (approved):
Looks good from my point of view :)

PR #4828, opened 2020-09-15
specifies initial selected dataset in publication tile

normanrz:
Previously the initially selected dataset in the publication tile was somewhat random. However, we have a sorting key that we use in the dataset list. This PR selects the first dataset initially.

------
- [x] Ready for review


daniel-wer (approved):
LGTM :+1:

Issue #4806, opened 2020-09-01
Some timespans not correctly merged.


fm3:
Some timespans seem to not get correctly merged. This is rare, and clearly visible in the time tracking view. This was supposed to be fixed by #4445 but reoccured in a new version.

fm3:
looks like #4830 actually solved this :)

PR #4830, opened 2020-09-17
Drop overlaps when saving timespans

fm3:
There have been new reports of overlapping timespans. I have two vaguely unlikely theories how this could have happened:
(1) the singleton creation of the timespan service fails, there is more than one timespan service, so the merging fails
(2) there are negative durations in reported timespans due to weird request ordering or some unknown bug in the front-end

This PR deals with (2). Before, negative durations were considered interrupted, so new timespans were created in the db, with overlaps. Now, the overlap is detected and the timestamps that overlap into an existing timespan are dropped. Also, a warning is logged in this case (also including the service hash/id, hopefully giving us insights into both theories).

With the insights gained we should also be able to debug where the negative timespans, if any, come from.

- https://___.webknossos.xyz

### Steps to test:
- modify the frontend to randomize timestamps of update actions and/or info/finish requests
- backend logging should warn about negative durations
- the time tracking/db should not show overlaps

### Issues:
- may help fixing #4806 

------
- [x] Ready for review


youri-k (approved):
LGTM :shipit:

PR #4831, opened 2020-09-18
Don't default to 2D layout for empty dummy dataset

philippotto:
The new 2D layout was also used for the empty dummy dataset with which the store is initialized. This PR ensures that the 2D layout is only used if the depth of the dataset is exactly 1 (and not 0, for example). I've yet to test whether this really fixes the nightly test, but I also noticed that the layout can switch between the two modes quickly when a dataset is opened.

- https://___.webknossos.xyz

### Steps to test:
- execute screenshot tests

### Issues:
- fixes nightly CI

------
- [x] Ready for review

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -249,7 +249,7 @@ jobs:
             curl
             -X POST
             -H "X-Auth-Token: $RELEASE_API_TOKEN"
-            https://kube.scm.io/hooks/remove/webknossos/dev/master?user=CI+%28nightly%29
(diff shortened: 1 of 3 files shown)


philippotto:
This change seems to fix the CI :tada:  See https://app.circleci.com/pipelines/github/scalableminds/webknossos/1167/workflows/1e5537dd-0f83-4190-9347-a07843e66a0a/jobs/12358

fm3 (approved):
Great find :+1:

PR #4834, opened 2020-09-22
filter nan values when generating histogram

### LINKED ISSUE #4817: Histogram sampling fails for float layer
See this discuss thread and the linked tracing: https://discuss.webknossos.org/t/dynamic-interpolation-of-sparse-data/1618

The histogram for the `probs_sparse` layer doesn't show any data and the min/max range cannot be edited, because the request to get the histogram data fails. The logged backend exception says: `java.lang.NumberFormatException: null` with the stack trace pointing to the json formatter. Looks like one of the numbers is becoming null before.

fm3:
float datasets can contain NaN values, the histogram should ignore them. Before this PR, the histogram request failed.
(also slipped in some minor fix in time tracking logging)

### Issues:
- fixes #4817 

------
- [x] Ready for review


philippotto (approved):
Awesome find :)

PR #4841, opened 2020-09-30
Update flow to v0.133.0

daniel-wer:
- The react-dom type annotations were moved from flow core to flow-typed
- No more arbitrary suppression comments (see https://flow.org/en/docs/errors/) and no more generic suppressions, instead the error code needs to be supplied. As I had to go through all suppression comments to add the error code, I also tried to use the semantically correct suppression comment everywhere ($FlowIssue for flow issues, $FlowExpectError where we know that we're causing type errors, etc.)
- Many issues were caused because flow no longer allows union types in computed properties (see last section of https://medium.com/flow-type/spreads-common-errors-fixes-9701012e9d58) :/
- Flow disallowed using any-typed values type annotations. [example](https://flow.org/try/#0CYUwxgNghgTiAEA3W8AeAueUB2BPA3AFCFgD22AzgC7y6arwC88AzPkA) (All of our THREE.<...> annotations)

- [x] There are 6 remaining flow errors that make no sense to me. Would be great if you could have a look :pray: If they don't make sense to you either and you don't find out what's causing them, we'll add the suppression comments...

- https://___.webknossos.xyz

### Steps to test:
- I only made very minimal changes to the running code, everything else were type annotations, so CI should be sufficient

------
- [x] Ready for review


daniel-wer:
Flow coverage calculated using `yarn flow batch-coverage frontend/javascripts/`

Before:
Files                : 379
Expressions          :
  Covered            : 115067
  Total              : 129708
  Covered Percentage : 88.71%

After:
Files                : 379
Expressions          :
  Covered            : 115231
  Total              : 129948
  Covered Percentage : 88.67%

daniel-wer:
Unfortunately `yarn flow-coverage` fails with:

```
yarn flow-coverage
yarn run v1.21.1
$ node_modules/flow-coverage-report/bin/flow-coverage-report.js -i "./frontend/javascripts/**/*.js" -x "./frontend/javascripts/test/**/*.js" -t html -t text
Error while generating Flow Coverage Report: SyntaxError: Unexpected token I in JSON at position 105090 SyntaxError: Unexpected token I in JSON at position 105090
    at JSON.parse (<anonymous>)
    at /home/daniel/projects/oxalis/node_modules/flow-annotation-check/lib/flow.js:80:17
    at processTicksAndRejections (internal/process/task_queues.js:97:5)
error Command failed with exit code 255.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

philippotto (approved):
Thank you for putting the effort into this :+1: :bee: I hope future flow updates will get us some worthy benefits :see_no_evil:

PR #4845, opened 2020-10-01
Fix Dataset Configurations

### LINKED ISSUE #4650: Volume layer opacity does not persist
After changing the opacity of volume layer in tracing mode (say to 50), if I view a dataset with default segmentation opacity that is different (say 20), then this new opacity (20) is also applied to the tracing mode. Would it be possible to make this setting persistent in tracing mode?

youri-k:
- https://___.webknossos.xyz

### Steps to test:
- open Dataset in View Mode and change segmentation settings (e.g. opacity)
- open Volume tracing without fallback on the same dataset and change volume layer settings (e.g. opacity)
- reopen Dataset in View Mode => segmentation settings should persist
- reopen tracing => volume layer settings should persist

### Issues:
- fixes #4650 

------
- [x] Ready for review


CHANGELOG.unreleased.md
```diff
@@ -18,6 +18,7 @@ For upgrade instructions, please check the [migration guide](MIGRATIONS.released
 
 ### Fixed
 - Fixed failing histogram requests for float layers with NaN values. [#4834](https://github.com/scalableminds/webknossos/pull/4834)
+- Fixed the disappearing of dataset settings after switching between view mode and trace mode. [#4845](https://github.com/scalableminds/webknossos/pull/4845)
```
fm3, line 21:
```suggestion
- Fixed the disappearing of dataset settings after switching between view mode and annotation mode. [#4845](https://github.com/scalableminds/webknossos/pull/4845)
```

conf/evolutions/056-add-layer-specific-view-configs.sql
```diff
@@ -0,0 +1,36 @@
+-- https://github.com/scalableminds/webknossos/pull/4XXX
```
fm3, line 1:
```suggestion
-- https://github.com/scalableminds/webknossos/pull/4845
```

app/models/user/UserService.scala
```diff
@@ -146,15 +147,24 @@ class UserService @Inject()(conf: WkConf,
...
+                                                                                               dataSet._id,
+                                                                                               name,
+                                                                                               config)
+      }
       _ <- userDataSetConfigurationDAO.updateDatasetConfigurationForUserAndDataset(user._id,
                                                                                    dataSet._id,
-                                                                                   configuration.configuration)
-      _ = userCache.invalidateUser(user._id)
+                                                                                   dataSetConfiguration.configuration)
+      _ = userCache.invalidateUser(user._id) // TODO Why do we need to invalidate the user here?
```
fm3, line 167: I believe this is a relic from when the dataset configuration was stored (and requested) with the user. Feel free to remove it :)

fm3 (approved):
Backend looks good to me, I think this is a clean solution :) I just added a few minor comments.

frontend/javascripts/oxalis/model_initialization.js
```diff
@@ -109,12 +109,29 @@ export async function initialize(
...
+      isSegmentationLayer: layer.category !== "color",
+    }));
+  }
+
+  if (tracing != null && tracing.volume != null) {
+    if (tracing.volume.fallbackLayer == null) {
+      displayedLayers = displayedLayers.filter(layer => !layer.isSegmentationLayer);
+    }
+    displayedLayers.push({ name: tracing.volume.id, isSegmentationLayer: true });
+  }
```
philippotto, line 131: I'm afraid I don't really understand why the front-end needs to send this layer list to the back-end for the dataset settings. Is it only so that the volume layer specific settings can be stored per tracing? Wouldn't it be sufficient to only send the volume tracing id in that case (since the back-end should already know all the layers)?

If this needs to stay, I'd suggest to try to deduplicate this "replace segmentation layer with fallback layer"-logic somehow, since it also appears in `setupLayerForVolumeTracing`.

philippotto (changes requested):
Great stuff! Also kudos for cleaning up the obsolete quality-setting and fighting through the front-end tests :)

I'm just a bit unclear about why it is necessary to send the dataset layers to the back-end when requesting the settings. Since the volume-vs-fallback-layer-specific logic is now duplicated, I'd prefer to refactor this a bit. See my other comment for context/details :)

frontend/javascripts/oxalis/model_initialization.js
youri-k: The basic concept is that the frontend can tell the backend, which layer settings it needs, to prevent the old overwriting behavior. However, as the backend only updates the layers sent back from the frontend, it is not necessary to send the complete list, as you mentioned. I included the filtering to not access irrelevant settings in the backend + db, but I think the overhead is not large, so I will adapt it according to your suggestion.

philippotto (approved):
Excellent :)

youri-k:
@fm3 Could you have a look at the backend again after the newest frontend changes. I don't know how to feel about the abstraction level in the `ConfigurationController`.

youri-k:
ToDo:
- [ ] migration for `defaultConfiguration` in DataSet

fm3 (approved):
Backend LGTM, great work :tada: Thanks for going through all the renamings :) I think it is a lot more clear and consistent now

frontend/javascripts/dashboard/dataset/default_config_component.js
```diff
@@ -1,18 +1,45 @@
...
+      dataIndex: "value",
+    },
+    {
+      title: "Comment",
+      dataIndex: "comment",
+    },
+  ];
+
+  const comments = {
+    alpha: "20 for segementation layer",
```
philippotto, line 39:
```suggestion
    alpha: "20 for segmentation layer",
```

frontend/javascripts/types/schemas/dataset_view_configuration_defaults.js
```diff
@@ -0,0 +1,64 @@
...
+import { validateObjectWithType } from "types/validation";
+
+const eliminateErrors = (instance: Object, errors: Array<*>, defaults: Object) => {
+  errors.forEach(error => {
+    if (error.name === "required") {
+      instance[error.argument] = defaults[error.argument];
+    } else if (error.name === "additionalProperties") {
+      delete instance[error.argument];
+    } else {
+      const invalidFieldName = error.property.split(".")[1];
```
philippotto, line 18: This line assumes that the incorrect field name is in the first level of the object. If this was used for a nested structure like this:
```
{
  someProperty: {
    someInvalidProperty: 1
  }
}
```

the enforcement would remove `someProperty`, too, if I read this right. We probably don't have any nesting right now, but to avoid future problems, I suggest to do one of the following:
- assert that `error.property.split(".").length === 2` so that this will throw an error if we should have deeper nesting somewhere
- handle the depth correctly. not sure whether you can test this easily? Also note: It would be important that the library reports `someProperty.invalid1.invalid2.invalid3` with `error.property === "someProperty.invalid1"`. If it reports `invalid1.invalid2.invalid3`, then removing the last entry would not be enough.

frontend/javascripts/types/schemas/dataset_view_configuration_defaults.js
```diff
@@ -0,0 +1,64 @@
...
+import { type APIDataset, type APIDataLayer } from "types/api_flow_types";
+import { getDefaultIntensityRangeOfLayer } from "oxalis/model/accessors/dataset_accessor";
+import { validateObjectWithType } from "types/validation";
+
+const eliminateErrors = (instance: Object, errors: Array<*>, defaults: Object) => {
+  errors.forEach(error => {
+    if (error.name === "required") {
+      instance[error.argument] = defaults[error.argument];
+    } else if (error.name === "additionalProperties") {
+      delete instance[error.argument];
```
philippotto, line 16: Does `error.argument` always refer to a top-level property or can it be nested?

philippotto (changes requested):
Awesome! Great job making this beautiful :+1: See my two comments about validation + nesting, though :)

frontend/javascripts/types/schemas/dataset_view_configuration_defaults.js
youri-k: I think, this is fine because if there is a nested property, its type will be `object`, which is disallowed for every property except `layers`, which are corrected elsewhere. Therefore, the incorrect property will always be in the top-level. If we have nested properties in the future, I suggest handling them explicitly like the layers or adapt the code then.

frontend/javascripts/types/schemas/dataset_view_configuration_defaults.js
youri-k: It always refers to a top-level property.

philippotto (approved):
Awesome! Looking forward to seeing this merged :)

PR #4847, opened 2020-10-05
Find data for volumes

### LINKED ISSUE #4840: “find data” for volume tracing layer
could be implemented by first sending a find-data request to the tracingstore (answered with the first valid bucketposition of the tracing from fossildb), and if that fails, send a standard find-data request to the datastore using the fallback layer

youri-k:
- https://___.webknossos.xyz

### Steps to test:
- open a volume tracing => find data button works and finds volume data
- open an empty tracing with fallback layer => find data button looks for volumes first and checks fallback layer afterwards

### Issues:
- fixes #4840 

------
- [x] Ready for review


fm3 (changes requested):
Works well! I added a comment concerning code duplication, otherwise this should be good to go :)

frontend/javascripts/oxalis/view/settings/dataset_settings_view.js
```diff
@@ -407,14 +400,31 @@ class DatasetSettings extends React.PureComponent<DatasetSettingsProps> {
...
-    );
-    if (!position || !resolution) {
+    let foundPosition;
+    let foundResolution;
+
+    if (volume && !isColorLayer) {
+      const requestUrl = `${tracingStore.url}/tracings/volume/${volume.tracingId}`;
+      const { position, resolution } = await findDataPositionForLayer(requestUrl);
+      if ((!position || !resolution) && volume.fallbackLayer) {
+        await this.handleFindData(volume.fallbackLayer, true);
```
philippotto, line 413: This feels a bit hacky, since you are passing `isColorLayer=true` here, even though this call is meant for the fallback layer of the segmentation. Can you restructure this code to avoid this?

fm3 (approved):
:tada:

frontend/javascripts/admin/admin_rest_api.js
```diff
@@ -981,16 +981,10 @@ export async function getOrganizationForDataset(datasetName: string): Promise<st
...
-  layerName: string,
+  requestUrl: string,
 ): Promise<{ position: ?Vector3, resolution: ?Vector3 }> {
   const { position, resolution } = await doWithToken(token =>
-    Request.receiveJSON(
-      `${datastoreUrl}/data/datasets/${datasetId.owningOrganization}/${
-        datasetId.name
-      }/layers/${layerName}/findData?token=${token}`,
-    ),
+    Request.receiveJSON(`${requestUrl}/findData?token=${token}`),
```
philippotto, line 987: I think, this function becomes hard to use when only requiring a `requestUrl` since it's not clear how that has to be constructed. I assume you changed this since the requirements for data layers vs. volume tracings differ too much, which is why I suggest to have two API functions here: `findDataPositionForLayer` and `findDataPositionForVolumeTracing`.

frontend/javascripts/oxalis/view/settings/dataset_settings_view.js
philippotto: There's probably a better way, but renaming `isColorLayer` to `isDataLayer` might work here (isDataLayer being false means that the volume tracing should be looked up).

philippotto (changes requested):
Awesome feature :) I added two suggestions for the front-end part :t-rex:

philippotto (approved):
Awesome :)

PR #4851, opened 2020-10-06
Skip reverted buckets

### LINKED ISSUE #4808: skip reverted buckets in volume tracing download
revertToVersion overwrites some buckets with a single zero-buckets to indicate the absence of data there.
those result in empty wkw files in the volume download. should be skipped entirely instead.

Also, let’s check if the volume merger behaves correctly with those.

Also, let’s try to unify all places where those zero-bytes are checked (some are equality checks, some test length == 1), should all be a function call instead. It *may* even be possible and sensible to adapt the iterator to skip these buckets.

youri-k:
- https://___.webknossos.xyz

### Steps to test:
- Download volume tracing with reverted buckets
- => empty buckets shouldn't be contained in the file

### Issues:
- fixes #4808 

------
- [x] Ready for review


fm3 (approved):
Good stuff! Also great refactoring :) I think this simplification will prevent a lot of future headaches.

PR #4860, opened 2020-10-08
resumable dataset upload

### LINKED ISSUE #4812: Make dataset upload robust
Right now, the dataset upload will fail as soon as the internet connection is a bit spotty (or the server restarted). Especially, for large datasets (> 1 GB), this can be a deal breaker.

To improve this we should support resumable file uploads. For the front-end, this should be quite easy to do with [resumable-js](https://github.com/23/resumable.js/). The back-end needs to be adapted to support this too. In the resumable.js repository, there is a reference implementation in java. Hopefully, this can be easily adapted to our scala codebase: https://github.com/23/resumable.js/tree/master/samples/java/src/main/java/resumable/js/upload

We could also think about storing the chunks on the server's disk, so that the upload even survives server restarts. However, this is probably overkill for now, since this feature will be most useful on webknossos.org and that instance is typically only updated once a month.

youri-k:
- https://___.webknossos.xyz

### Steps to test:
- upload dataset => should work

### Issues:
- fixes #4812 

------
- [x] Ready for review


youri-k:
ToDo:
- [ ] check how unique resumable `uniqueIdentifier`is

youri-k:
Regarding this ToDo:
> check how unique resumable `uniqueIdentifier` is

resumable uses this code to generate it:
```
var relativePath = file.webkitRelativePath||file.relativePath||file.fileName||file.name;
var size = file.size;
return(size + '-' + relativePath.replace(/[^0-9a-zA-Z_-]/img, ''));
```

In my opinion, this isn't unique enough, but I am not sure if it would be enough to add a time code or if we should add the user token in the backend, but that could break it, if the token changes while uploading.

Any ideas @philippotto, @fm3 ?

CHANGELOG.unreleased.md
```diff
@@ -21,6 +21,7 @@ For upgrade instructions, please check the [migration guide](MIGRATIONS.released
 - webknossos.org only: Accounts associated with new organizations can now be created even when a datastore is unreachable. The necessary folders are created lazily when needed. [#4846](https://github.com/scalableminds/webknossos/pull/4846)
 - When downloading a volume tracing, buckets containing a single 0 byte, that were created to restore older versions, are now skipped. [#4851](https://github.com/scalableminds/webknossos/pull/4851)
 - Task information CSV now contains additional column `creationInfo`, containing the original NML filename for tasks based on existing NMLs. [#4866](https://github.com/scalableminds/webknossos/pull/4866)
+- The dataset upload is now more robust due to chunking. [#4860](https://github.com/scalableminds/webknossos/pull/4860)
```
philippotto, line 24:
```suggestion
- The dataset upload is now more robust and can recover from a failed upload of a data chunk. [#4860](https://github.com/scalableminds/webknossos/pull/4860)
```

frontend/javascripts/admin/dataset/dataset_upload_view.js
```diff
@@ -48,9 +54,9 @@ class DatasetUploadView extends React.PureComponent<PropsWithForm, State> {
 
   normFile = e => {
     if (Array.isArray(e)) {
```
philippotto, line 56:
```suggestion
    // Ensure that only one dataset can be uploaded simultaneously
    if (Array.isArray(e)) {
```

frontend/javascripts/admin/dataset/dataset_upload_view.js
```diff
@@ -63,30 +69,65 @@ class DatasetUploadView extends React.PureComponent<PropsWithForm, State> {
...
+                isRetrying: false,
+                uploadProgress: 0,
+              });
+            },
+          );
+        });
+
+        resumableUpload.on("fileAdded", () => {
+          resumableUpload.upload();
+        });
```
philippotto, line 115: Instead of attaching this event handler, couldn't you move `resumableUpload.upload();` below the `addFiles` call in line 130? I'd assume that this is called only once (since only one file is added?). So, doing it explicitly without any listeners, seems more straightforward to me.

Maybe even `resumableUpload.addFiles(formValues.zipFile).upload();` would work as a shorthand.

frontend/javascripts/admin/dataset/dataset_upload_view.js
```diff
@@ -63,30 +69,65 @@ class DatasetUploadView extends React.PureComponent<PropsWithForm, State> {
...
+              await Utils.sleep(3000); // wait for 3 seconds so the server can catch up / do its thing
+              this.props.onUploaded(activeUser.organization, formValues.name);
+            },
+            () => {
+              this.setState({
+                isUploading: false,
+                isFinished: false,
+                isRetrying: false,
+                uploadProgress: 0,
+              });
```
philippotto, line 108: Shouldn't an error be shown to the user in this case? One could even think about retrying `finishDatasetUpload`, but it's probably hard to test.

philippotto (commented):
Awesome stuff! Just left a few small comments and you probably want to change the unique identifier thing, right?

frontend/javascripts/admin/dataset/dataset_upload_view.js
youri-k: Sadly, this didn't work because `addFiles` is asynchrous and only finished when `fileAdded` is fired

frontend/javascripts/admin/dataset/dataset_upload_view.js
youri-k: Yes, you're right

frontend/javascripts/admin/dataset/dataset_upload_view.js
philippotto: Ah, I see, then nevermind :)

app/WebKnossosModule.scala
```diff
@@ -22,6 +23,7 @@ class WebKnossosModule extends AbstractModule {
     bind(classOf[UserCache]).asEagerSingleton()
     bind(classOf[AnnotationStore]).asEagerSingleton()
     bind(classOf[DataSetService]).asEagerSingleton()
+    bind(classOf[DataSourceService]).asEagerSingleton()
```
fm3, line 26: Do you think it would be possible to extract the part of DataSourceService’s functionality that needs it to be singleton to another, smaller, service? I feel like it already has a lot of tasks that may perform better if the injector can create multiple instances dynamically. (things to double-check there: the classes the singletons are injected into don’t automatically become singletons too, right?). This may be a good strategy for other services as well, in the future.

webknossos-datastore/app/com/scalableminds/webknossos/datastore/services/DataSourceService.scala
```diff
@@ -23,10 +23,16 @@ import play.api.inject.ApplicationLifecycle
 import play.api.libs.json.Json
 import org.joda.time.format.ISODateTimeFormat
 
+import scala.collection.mutable
 import scala.concurrent.ExecutionContext.Implicits.global
 import scala.concurrent.duration._
 import scala.io.Source
 
+case class ResumableUploadInformation(chunkSize: Int, totalChunkNumber: Long)
```
fm3, line 31: consider renaming totalChunkNumber to totalChunkCount to distinguish it from the currentChunkNumber

webknossos-datastore/app/com/scalableminds/webknossos/datastore/services/DataSourceService.scala
```diff
@@ -109,7 +117,49 @@ class DataSourceService @Inject()(
     if (emptyDirs.nonEmpty) logger.warn(s"Empty organization dataset dirs: ${emptyDirs.mkString(", ")}")
   }
 
-  def handleUpload(id: DataSourceId, dataSetZip: File): Fox[Unit] = {
+  def handleUpload(uploadId: String,
```
fm3, line 120: consider renaming this to handleUploadChunk

philippotto (approved):
Looks good from my side :man_dancing:

webknossos-datastore/app/com/scalableminds/webknossos/datastore/services/DataSourceService.scala
```diff
@@ -109,7 +117,49 @@ class DataSourceService @Inject()(
...
+                   datasourceId: DataSourceId,
+                   resumableUploadInformation: ResumableUploadInformation,
+                   currentChunkNumber: Int,
+                   chunkFile: File): Fox[Unit] = {
+    val needsWrite = pendingUploadChunks.synchronized {
```
fm3, line 125: I find it a bit hard to grasp what `needsWrite` expresses, especially since it is determined via the return value of `set.add`? What is the semantic of this return value? Could needsWrite be renamed to make its use clearer?

fm3 (changes requested):
I like the approach and separating uploading of chunks and a final finish request seems fair. To be honest I had some troubles understanding some of the semantics of `pendingUploadChunks` and `needsWrite` in the `DataSourceService`. Maybe we can figure something out for that (or maybe I just need a verbal explanation)

app/WebKnossosModule.scala
youri-k: I just tested it and it works :tada: 
> the classes the singletons are injected into don’t automatically become singletons too

No, they can still have multiple instances

fm3 (approved):
Thanks, with the renamings everything gets clear :+1: Works for me and Backend LGTM

philippotto:
:man_dancing: :tada:

PR #4861, opened 2020-10-09
Load histogram data sequentially

### LINKED ISSUE #4829: Loading histogram blocks data loading
I did not check the code, but it looks to me like loading the histogram blocks the first bucket data loading requests in the front-end. However, histogram loading may be slow (especially in remote wk-connect datasets, when https://github.com/scalableminds/webknossos-connect/pull/96 is merged). Could it be changed so that data-loading starts faster and the histogram is loaded later? I realize that for float layers the histogram min/max values are required to calibrate the data range, but maybe it can be initialized with something heuristically and then be reset when the histogram request completes?

grittaweisheit:
This PR adjusts the parallel sending of histogram data requests to a sequential procedure. That results in less blockage of bucket data requests.

- https://___.webknossos.xyz

### Steps to test:
- open the dev tools of your browser
- open a tracing in webknossos
- the histogram requests should be sent one after the other
- the data requests should be sent parallel 

### Issues:
- fixes #4829

------
- [x] Ready for review


frontend/javascripts/oxalis/model/sagas/load_histogram_data_saga.js
```diff
@@ -15,17 +15,17 @@ async function fetchAllHistogramsForLayers(
   dataset: APIDataset,
 ): Promise<HistogramDataForAllLayers> {
   const histograms: HistogramDataForAllLayers = {};
-  const histogramPromises = dataLayers.map(async dataLayer => {
+  for (const dataLayer of dataLayers) {
     try {
+      // eslint-disable-next-line no-await-in-loop
```
philippotto, line 20: Great find! Please add a comment above the `eslint-disable` line which explains why this is done (basically, what you already included in the PR description), so that it's clear, why this is done sequentially.

philippotto (approved):

PR #4862, opened 2020-10-09
add admins and dataset managers to dataset access list

### LINKED ISSUE #4855: Dataset user access list is incomplete
`/api/datasets/…/…/accessList` should include users that have dataset access because they are admin or dataset manager.
(Special care should be taken to avoid listing users the requesting user is not allowed to see)

fm3:
### Steps to test:
- click the plus sign left of dataset name in dataset list
- should list admins and team managers of the dataset’s organization (if the requesting user is allowed to see them)

### Issues:
- fixes #4855

------
- [x] Ready for review


youri-k (approved):
LGTM :shipit:

PR #4867, opened 2020-10-12
word break in upload zone nml filename

### LINKED ISSUE #4250: Text out of bounds
When importing a NML file with a long file name to a dataset the file name is outside the box and therefore not very readable.

fm3:
CSS again behaving rather weird… `word-wrap: break-word` does not do the trick in this case (it was set already before this pr), but `word-break: break-word` does so in major browsers, even though it is not in the standard. Compare [this SO thread](https://stackoverflow.com/questions/36150458/flex-item-overflows-container-due-to-long-word-even-after-using-word-wrap).

@leowe I hope I interpreted your issue description correctly in assuming that you are referring to the upload zone modal?

### Steps to test:
- drag nml with long filename (no spaces or dashs) into dataset / annotation
- filename in import modal should be readable

### Issues:
- fixes #4250

------
- [x] Ready for review


philippotto (approved):
:1st_place_medal: for dealing with the css!

PR #4869, opened 2020-10-12
Do not show toasts for bucket errors but notify airbrake instead

### LINKED ISSUE #4819: Don't show bucket-failed-to-fetch warning immediately
For some users, the warnings appear quite often (every 20 min?). We should double-check that the warning only appears if some requests fail repeatedly.

daniel-wer:
For historical context, initially the toasts were enabled as part of https://github.com/scalableminds/webknossos/issues/4725.

Check airbrake and search for `requesting data` to see the error that is logged to airbrake.

- https://donotshowbucketerrors.webknossos.xyz

### Steps to test:
- Open a tracing on the deployed dev branch, open dev tools and choose "Offline" in the network tab. Move, so that buckets are requested. There should be no toast regarding the failed bucket requests. Check airbrake for the errors.

### Issues:
- fixes #4819 

------
- [ ] Ready for review

### DIFF
--- a/frontend/javascripts/libs/request.js
+++ b/frontend/javascripts/libs/request.js
@@ -266,13 +266,6 @@ class Request {
       setTimeout(() => resolve("timeout"), timeout);
     });
 
-  handleStatus = (response: Response): Promise<Response> => {
-    if (response.status >= 200 && response.status < 400) {
-      return Promise.resolve(response);
-    }
-    return Promise.reject(response);
-  };
-
   handleError = (
     requestedUrl: string,
     showErrorToast: boolean,
--- a/frontend/javascripts/oxalis/model/bucket_data_handling/wkstore_adapter.js
+++ b/frontend/javascripts/oxalis/model/bucket_data_handling/wkstore_adapter.js
@@ -10,6 +10,7 @@ import {
   isSegmentationLayer,
   getByteCountFromLayer,
 } from "oxalis/model/accessors/dataset_accessor";
+import ErrorHandling from "libs/error_handling";
 import { parseAsMaybe } from "libs/utils";
 import { pushSaveQueueTransaction } from "oxalis/model/actions/save_actions";
 import { updateBucket } from "oxalis/model/sagas/update_actions";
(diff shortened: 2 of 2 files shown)


frontend/javascripts/oxalis/model/bucket_data_handling/wkstore_adapter.js
```diff
@@ -171,22 +171,19 @@ export async function requestFromStore(
...
-    let detailedError =
+    const detailedError =
       errorResponse.status != null
-        ? `Status code ${errorResponse.status} - "${errorResponse.statusText}".`
+        ? `Status code ${errorResponse.status} - "${errorResponse.statusText}" - URL: ${
+            errorResponse.url
+          }.`
         : errorResponse.message;
 
-    detailedError += `(URL: ${dataUrl})`;
```
daniel-wer, line 179: If the error is not a response with a status code, the requests lib already appends the URL which is why I moved this from here to the conditional above.

frontend/javascripts/libs/request.js
```diff
@@ -266,13 +266,6 @@ class Request {
       setTimeout(() => resolve("timeout"), timeout);
     });
 
-  handleStatus = (response: Response): Promise<Response> => {
```
daniel-wer, line 269: This method was unused, `handle_http_status.js` is used everywhere, instead.

frontend/javascripts/oxalis/model/bucket_data_handling/wkstore_adapter.js
```diff
@@ -171,22 +171,19 @@ export async function requestFromStore(
...
-    Toast.warning(
-      errorMessage,
-      {
-        key: errorMessage,
-        sticky: false,
-      },
+    ErrorHandling.notify(new Error(errorMessage), {
       detailedError,
-    );
+      isOnline: window.navigator.onLine,
```
philippotto, line 185: Cool beans! Didn't know about that.

philippotto (approved):
Awesome! :+1: for the additional clean up!

frontend/javascripts/oxalis/model/bucket_data_handling/wkstore_adapter.js
```diff
@@ -171,22 +171,19 @@ export async function requestFromStore(
     const errorMessage = `Requesting data from layer "${
       layerInfo.name
```
philippotto, line 172: One thing I noticed: Having the layer name in the error message inhibits the aggregation in airbrake. Maybe we can omit the layer name when sending the message to airbrake?

daniel-wer: For reference, airbrake is able to figure out that the layer name is dynamic and abstracts from that, so the aggregation still works :+1:

PR #4870, opened 2020-10-14
Some smaller performance improvements

philippotto:
While digging through the dev tools, I noticed some smaller potentials for performance improvements and tackled the easy ones. The impact is hard to measure, but I think it's fair to say that all of the changes save CPU cycles without adding to much complexity. So, I think, these changes are fair.

- https://___.webknossos.xyz

### Steps to test:
- I worked with a very big and a normal sized tracing to ensure that the changes work
- normal tracing should still work

### Issues:
- #performance

------
- [X] Ready for review


frontend/javascripts/oxalis/view/right-menu/mapping_info_view.js
```diff
@@ -445,21 +445,11 @@ class MappingInfoView extends React.Component<Props, State> {
...
-  },
-  setHideUnmappedIds(hideUnmappedIds: boolean): void {
-    dispatch(setHideUnmappedIdsAction(hideUnmappedIds));
-  },
-});
+const mapDispatchToProps = {
+  setMappingEnabled: setMappingEnabledAction,
+  setAvailableMappingsForLayer: setLayerMappingsAction,
+  setHideUnmappedIds: setHideUnmappedIdsAction,
+};
```
philippotto, line 452: This change doesn't have an actual impact, it's simply a shorthand.

frontend/javascripts/oxalis/view/right-menu/mapping_info_view.js
```diff
@@ -489,7 +479,4 @@ export default connect<Props, OwnProps, _, _, _, _>(
   mapStateToProps,
   mapDispatchToProps,
   null,
-  {
-    pure: false,
-  },
```
philippotto, line 494: I didn't notice any problems with the mapping UI during testing. The component is-rerendered when the mouse position changes, so it's probably hard to get an outdated mapping view.

frontend/javascripts/oxalis/view/right-menu/trees_tab_view.js
```diff
@@ -661,98 +662,109 @@ class TreesTabView extends React.PureComponent<Props, State> {
...
-            onCancel={this.hideDeleteGroupsModal}
-            onJustDeleteGroup={() => {
-              this.deleteGroupAndHideModal(groupToDelete, false);
-            }}
-            onDeleteGroupAndTrees={() => {
-              this.deleteGroupAndHideModal(groupToDelete, true);
-            }}
-          />
-        ) : null}
+        <DomVisibilityObserver targetId={treeTabId}>
```
philippotto, line 665: The trees tab view eats a considerable amount of performance for large tracings. However, I didn't find easy gains, so I went with this solution which should give some relief if one is willing to hide the trees tab for a moment.

frontend/javascripts/oxalis/view/right-menu/trees_tab_view.js
```diff
@@ -661,98 +662,109 @@ class TreesTabView extends React.PureComponent<Props, State> {
...
+                  <DeleteGroupModalView
+                    onCancel={this.hideDeleteGroupsModal}
+                    onJustDeleteGroup={() => {
+                      this.deleteGroupAndHideModal(groupToDelete, false);
+                    }}
+                    onDeleteGroupAndTrees={() => {
+                      this.deleteGroupAndHideModal(groupToDelete, true);
+                    }}
+                  />
+                ) : null}
```
philippotto, line 763: Only indentation should have changed.

frontend/javascripts/oxalis/geometries/cube.js
```diff
@@ -68,9 +69,10 @@ class Cube {
       this.setCorners(this.min, this.max);
     }
 
-    Store.subscribe(() => {
-      this.updatePosition(getPosition(Store.getState().flycam));
-    });
```
daniel-wer, line 73: Uff, nice find :+1:

frontend/javascripts/oxalis/view/right-menu/mapping_info_view.js
daniel-wer: Hmm, I couldn't find out why this was added. It was added as part of the Custom Viewports PR, but I didn't see any specific reason. I agree that it should be fine :)

frontend/javascripts/oxalis/geometries/plane.js
```diff
@@ -148,14 +148,13 @@ class Plane {
...
   };
 
-  setPosition = (posVec: typeof THREE.Vector3): void => {
-    this.TDViewBorders.position.copy(posVec);
-    this.crosshair[0].position.copy(posVec);
-    this.crosshair[1].position.copy(posVec);
-    this.plane.position.copy(posVec);
-
-    const globalPosition = getPosition(Store.getState().flycam);
-    this.plane.material.setGlobalPosition(globalPosition);
```
daniel-wer, line 158: This is kind of weird, why wasn't `posVec` given as the argument here, before?!

daniel-wer (approved):
Thanks for spending the time to investigate this in detail and for the many performance improvements, great work! :+1:

frontend/javascripts/oxalis/geometries/plane.js
philippotto: Ah, dang. I checked the caller in `SceneController.update()` where it should not matter. However, there's also `SceneController.updateSceneForCam()` where the third dimension is offset so that the skeleton is before the plane. So, there can be a difference.

Do you know what the difference between `update` and `updateSceneForCam` is? I'm a bit confused about their purposes :thinking: Happy to have a call about this in case it's easier :see_no_evil:

daniel-wer: To me it looks like the `update` method is used to update the scale and position of planes if the user moved/zoomed. The `updateSceneForCam` method is used during rendering of the planes, where if plane A is rendered, planes B and C are hidden to avoid artifacts (plane A is also offset to allow to see the skeleton tracing behind the plane).

philippotto: Hmm, at least for the position, it's not clear to me why the two update mechanisms are necessary. In the end, updating the position in the plane material should only be relevant for rendering (if it's not rendered, it shouldn't matter). So, only `updateSceneForCam`'s `setPosition` call seems crucial to me. Or am I missing something?

philippotto:
I just pushed some changes which make some improvements to the event system. Mainly, I replaced the backbone events in the bucket class with `NanoEvents`. In addition to be a bit more elegant, the code should also perform better (since we avoid the `_.extend` call on each bucket instance).
While doing this, I also found a bug in the event-unsubscription in the TextureBucketManager.

I also created this issue (https://github.com/scalableminds/webknossos/issues/4873) since I noticed that the entire event system could use some love.

daniel-wer:
Looks good, nice find with the bug in the event unsubscription!

PR #4872, opened 2020-10-14
Backend NML Parser: Performance Improvements

fm3:
I had wrongly assumed the bulk of the upload nml complexity to be in validateTrees.

Analysis with a 350 MB real-world NML yielded:

```
45 s  load xml from input stream
961 s  parse trees
3 s  splitMultiComponentTrees
6 s  validateTrees
0.7 s  send to tracingstore & save
```
This PR tackles two properties of `parseTrees`:
(1) the accessing of XMLNode attributes via shortcut method `\` is not very efficient, new helper method `getSingleAttribute` (speedup factor ×2–4)
(2) replace inefficient handling of comments and branchpoints by hashmap lookup (speeup factor ×10–15 for this particular NML, which notably has 80k node comments)

so the parse trees line after this change reads with speedup factor ×36
```
27 s  parse trees
```
with the others unchanged.

Other changes:
- increase request timeouts from 1000s (16.6min) to 10000s (166min)
- use the same request timeouts in dev mode (has to be defined in built.sbt)
- when killing the proxy.js, send sigkill rather than sigterm to child processes, avoiding port-in-use problems

### TODO
- [x] use `getSingleAttribute` in remaining places (won’t bring much more speedup, just for consistency)
- [x] cleanup (remove time-logging, re-enable startup scripts etc)

### Steps to test:
- upload very large NML via dashboard
- should be faster, should yield the same SkeletonTracing

------
- [x] Ready for review


youri-k (approved):
LGTM :shipit:

PR #4876, opened 2020-10-15
Time-track movements in the 3D viewport

### LINKED ISSUE #4825: Time-track movements in the 3D viewport

daniel-wer:
- Added a new `updateTdCamera` update action which is triggered if the user modified the 3D view, so that these actions can be time tracked. The action is empty otherwise.
- While I was at it I spent some love on the version view, added a lot of missing descriptions for certain update actions (tree/group visibility, user bounding boxes, fallback layer removal, node updates, edge deletion/addition) and extended the tree deletion message as by now multiple trees can be deleted at once.
- I added save queue compaction for `updateNode` update actions as otherwise lots of new versions are created if the radius of a node is changed or a node is moved. `updateTdCamera` actions are also compacted and only the last one within one save interval is kept.
- I've also noticed that the Undo of a tree deletion is misinterpreted as an NML upload by the version view (as multiple nodes are created at once which is not possible otherwise). Looking at the update actions one cannot be distinguished from the other, so I'm not sure how to fix this.

TODO:
- [x] Rotation in the 3D view is not picked up as a change yet, because the action that is used for that is also used when hovering over the 3D view and when initializing the tracing (those should not be counted as actions). I'll probably dispatch another action for the 3D view rotation so that it can be recognized and told apart from the other actions. _(Solved by introducing a duplicate action that doesn't trigger the save queue diffing which is used accordingly)_

- https://___.webknossos.xyz

### Steps to test:
- Move/Zoom/Rotate the 3D view - the save button should indicate that there are changes which need to be saved.
- Create a task and exclusively do stuff in the 3D view - time should be tracked.
- Check the version view to see whether the action is correctly labeled and also that subsequent movements in the 3D view within one save interval are compacted to only one version.
- Do some of the other mentioned actions with new descriptions/compactions and check the version view afterwards.

### Issues:
- fixes #4825 

------
- [x] Ready for review

### DIFF
--- a/frontend/javascripts/oxalis/view/version_entry.js
+++ b/frontend/javascripts/oxalis/view/version_entry.js
@@ -11,18 +11,46 @@ import type {
   UpdateTreeUpdateAction,
   DeleteTreeUpdateAction,
   RevertToVersionUpdateAction,
+  UpdateNodeUpdateAction,
+  UpdateTreeVisibilityUpdateAction,
+  UpdateTreeGroupVisibilityUpdateAction,
+  CreateEdgeUpdateAction,
+  DeleteEdgeUpdateAction,
 } from "oxalis/model/sagas/update_actions";
 import FormattedDate from "components/formatted_date";
+import { MISSING_GROUP_ID } from "oxalis/view/right-menu/tree_hierarchy_view_helpers";
 
 type Description = { description: string, type: string };
 
+// The order in which the update actions are added to this object,
+// determines the order in which update actions are checked
+// to describe an update action batch. See also the comment
+// of the `getDescriptionForBatch` function.
 const descriptionFns = {
-  deleteTree: (action: DeleteTreeUpdateAction): Description => ({
-    description: `Deleted the tree with id ${action.value.id}.`,
+  importVolumeTracing: (): Description => ({
+    description: "Imported a volume tracing.",
+    type: "plus",
+  }),
+  createTracing: (): Description => ({
+    description: "Created the annotation.",
+    type: "rocket",
+  }),
+  updateUserBoundingBoxes: (): Description => ({
+    description: "Updated a user bounding box.",
+    type: "codepen",
+  }),
+  removeFallbackLayer: (): Description => ({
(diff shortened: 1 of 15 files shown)


frontend/javascripts/oxalis/view/version_entry.js
```diff
@@ -96,6 +157,7 @@ function getDescriptionForBatch(actions: Array<ServerUpdateAction>): Description
 
   // If more than one createNode update actions are part of one batch, that is not a tree merge or split
   // an NML was uploaded.
+  // TODO: This could've also been an undo/redo action.
```
philippotto, line 160: Hmm, tricky. A simple solution would be to generalize the description message (e.g., "X nodes were added.") or to just leave this as is. Figuring out what the correct reason is would obviously be better, but I can't think of a good solution right now :/

philippotto (approved):
Great stuff & clean up :+1: Worked well for me!

One thing I noticed during testing: Merely opening the version-restore-view seems to clear the undo stack. After opening and closing the version-restore-view, I cannot undo an action. But that's been like this probably before this PR? I cannot really see why this is happening. Maybe you know anything about this?

daniel-wer:
> One thing I noticed during testing: Merely opening the version-restore-view seems to clear the undo stack. After opening and closing the version-restore-view, I cannot undo an action. But that's been like this probably before this PR? I cannot really see why this is happening. Maybe you know anything about this?

I'm afraid this is by design (so far). When a version is previewed (or the version view is closed) the tracing is restarted (`api.tracing.restart`) which also restarts all sagas to get a clean state.

frontend/javascripts/oxalis/view/version_entry.js
daniel-wer: Yes, I think we cannot do much about it easily apart from generalizing the message as you suggested. I'll create an issue for that.

daniel-wer: For reference https://github.com/scalableminds/webknossos/issues/4883

PR #4881, opened 2020-10-19
Don't crash in arbitrary *view* mode when invoking skeleton-shortcuts

philippotto:
For example, opening a dataset in view mode, switching to arbitrary mode and then pressing shift+space raised an exception.

- https://fixarbitrarywithoutskeleton.webknossos.xyz

### Steps to test:
- open a dataset in view mode, switch to arbitrary mode and try to break it somehow

### Issues:
- fixes rare "Can't extract the value of a Nothing" error (see https://scm.airbrake.io/projects/95438/groups/2773243668026156845?tab=notice-detail)

------
- [X] Ready for review


daniel-wer (approved):
Nice find and fix! :+1:

PR #4889, opened 2020-10-23
Resumable dataset upload (second try)

### LINKED ISSUE #4888: re-activate resumable dataset upload
#4860 was reverted in #4887 because it broke in production. My theory is that the setup that has the datastore separate from webKnossos fails because the DataSourceService is explicitly instantiated in `WebKnossosModule`. Instead, it should be instantiated in the datastore module. (Also shouldn’t it be the `UploadService` instead?)

fm3:
#4860 was reverted in #4887 because it broke in production. this is the second try.

The first one broke because webKnossos instances without a local datastore tried to instantiate `DataSourceService` but it needs resources that are not available outside of the datastore module.

fixes #4888


philippotto (approved):
Can't say much about the actual fix, but I guess we'll see in prod whether it works :+1:

PR #4891, opened 2020-10-23
Enforce TaskType Resolution Restrictions for Multi-Res Volume Tasks

fm3:
### Changes for the  larger PR #4755 

- creating tasks with volume tracings results in tracings with only the resolutions allowed by tasktype resolution restrictions
- upload does no longer downsample (assumption that single-res-tracings are always “legacy” no longer holds)
- merge does no longer downsample (merge for compound uses only data of intersection of the resolution of all the tracings (excluding empty tracings))
- new explicit downsampling route `PATCH         /annotations/:typ/:id/downsample` (explorative only) takes lowest present mag (e.g.  mag1) and creates pyramid on top of that one, returns annotation object with new volume tracing id. may be slow.

### TODO
- [x] save correct resolution list to tasks that have no `initialData`
- [x] what to do when uploading multiple zips containing volumes with differing resolutions? (`initializeWithDataMultiple`) → assert they are equal.
- [x] determine if resolution restrictions should also influence zooming (pending feedback)
 → they should not
- [x] front-end for explicit downsampling route (button, loading modal, reinitialization)
- [x] fix task creation with base annotations
- [x] make restrict-resolutions immutable for task types (backend)
- [x] sort resolution list for volume tracings
- [x] remove debugging output in volumetracingservice.scala
- [x] create migration that sets resolution restriction to mag1-only for old volume task types
  - [x] what about existing ones that *do restrict* but not precisely mag1-only? → does not happen
- [x] make restrict-resolutions immutable for task types (frontend)
- [x] implement new front-end behavior of resolution restrictions (zoom freely, but allow editing only in right mag, show 
explaining toast)
- [x] hide downsample button in case resolution pyramid is already complete
- [x] merge master
- refactor resolution restrictions (not optional + bool) → let’s do that in follow-up PR, issue #4901 

- https://multiresvolumesrestricted.webknossos.xyz

### Steps to test:
- [x] create task type (volume / hybrid), check »restrict magnifications«, select a restriction
  - [x] create plain task with this task type, request that task. its mag list should be intersection of dataset mags + task type restrictions
  - [x] create task where this intersection becomes empty (e.g. in task type restrict min mag2 and use dataset that has only mag1) → should fail with error message, task should *not* be in db afterwards
  - [x] create task with *no* restrictions, should have full dataset / fallback layer mag list as before
  - [x] create hybrid task based on NML upload (*no* base volume data included), with restrictions, should have intersection of dataset mags + task type restrictions
  - [x] create volume task based on NML upload (with base volume data included), with restrictions, should have intersection of dataset mags + task type restrictions + mags present in uploaded base data
  - [x] create volume task based on NML upload *with fallback layer* (with base volume data included), with restrictions, 
  - [x] should have intersection of dataset mags + task type restrictions + mags present in uploaded base data
    - [x] fallback layer data should load for those mags
    - [x] volume data from zip should load for those mags
  - [x] create volume task based on previous volume task (by id), should have intersection of base task mags + task type restrictions
- [x] download a task with restricted mag list, should contain exactly those mags in data.zip (anisotropic resolutions marked as 2-2-1, isotropic marced as 2, each mag should also have header.wkw)
- [x] copy to my account should not change mag list
- [x] upload single volume annotation as explorational, should have exactly the mags present in its data.zip
- [ ] upload multiple volume annotation at once as explorational, should be merged and have exactly the mags present in its data.zip IF all data.zips have the same resolution set
     → Could not test this due to #4903 deferred to later PR
- [ ] upload multiple volume annotation at once as explorational, should fail with readable error message IF data.zips do not have the seme resolution set. should not be present in user explorationals
     → Could not test this due to #4903 deferred to later PR
- [x] downloading & re-uploading of single explorational with fallback layer should look identical
  - [x] should have all data 
  - [x] data should not be relabeled
- [x] on explorational with limited mag set (either legacy, or uploaded, or task copied to my account), hit »create resolution pyramid« button, page should refresh, volume should now have full mag set of dataset STARTING at the lowest mag that was previously existing (e.g. mag2 + pyramid)
- [x] this button should not be present
    - [x] in tasks
    - [x] if the annotation already has its full possible mag range (limited by fallback layer if present, otherwise dataset)
    - [x] in skeleton only annotations
- [x] check out master (before multi-res-volmes), create volume task (no restrictions), check out this branch, request the task. should have mag1 only
- [x] drag n drop import of volume annotation into existing annotation should import data IF resolution sets match

------
- [x] Ready for review


philippotto:
Awesome :+1: 

> new explicit downsampling route PATCH /annotations/:typ/:id/downsample (explorative only) takes lowest present mag (e.g. mag1) and creates pyramid on top of that one, returns annotation object with new volume tracing id. may be slow.

Just for my understanding: will it be slower than the previous downsample-on-upload approach or "equally slow"?

fm3:
about equal, I’d think. not sure if fetching the data from fossildb vs unpacking the data from the zip has much influence, but i’d guess the bulk time comes from the actual downsampling computation. I just wrote that note to indicate that this should probably be communicated to the user with a spinner. for my sample volume annotation it took 10s, but it may also be several minutes for larger ones. in the future we may of course also consider performance measurements and think about speedup :)

philippotto:
Ok, cool, thanks for the info! If the request takes longer than a minute, we might run into request timeouts again due to the browser, though :see_no_evil: Could you add a `sleep` to the downsample request for me? Then, I could easily test this :)

philippotto:
> front-end for explicit downsampling route (button, loading modal, reinitialization)

I've implemented this now. Still a bit hesitant about the details, but might work for now. The entire page will reload which should be sufficient.

>  this button should not be present [...]  if the annotation already has its full possible mag range (limited by fallback layer if present, otherwise dataset)

This is currently not implemented. I'm wondering if there are use cases where one wants a "fresh" downsampled version of the tracing. Our ad-hoc downsampling approach is a bit of a heuristic, which is why I could imagine that downsampling from scratch could be useful from time to time. What do you think, @fm3?

image

image

An argument against showing the downsample-option all the time is probably that it could confuse the user, as to why this button is there in the first place. I also thought about moving it into the layer-sidebar (but that's already quite crammed) or in the dropdown next to the save button (semantically, this seems to be the worst option).

Would be happy to hear your opinion on that :)

fm3:
Good question! 

I agree that in some cases this strictly “mode”-based downsampling may be desired, overwriting the heuristic one. However, I believe it will be a lot more common to just have an annotation with an incomplete resolution pyramid. The explanation in your screenshot is good but it targets mostly the overwriting case (“re-create”). I think if we really want to support both use cases, we need two differing explanations, rather than one that tries to explain both. 

So if the resolution pyramid is already present, show a text similar to your draft, but if not, show a simpler version, that does not mention re-creation. Also both explanations may benefit from actual values for the resolutions present and those about to be created (this information should be present in the front-end, right?).

However, my assumption is that it would be fine to actually support only the one use case, and hide the button if the resolution pyramid is already complete. This will reduce user confusion. If the difference between the mode downsampling and the heuristic downsampling is so significant that we need to explain it to the users and offer both, maybe the heuristic is too bad to use ever (but I think it’s ok).

Let me know what you think.

On the button location, I think I’d have looked for it in the layer settings, but info tab is fine too. (Maybe even both? Not sure what the UI guidelines on redundant buttons are)

philippotto:
Completely agree with you!

I changed the modal to look like this:

image

Can be opened from here:
image

Looks a tad wonky in my opinion, but I don't know how to improve this. Might be okay for now.

app/models/annotation/AnnotationSettings.scala
```diff
@@ -1,18 +1,31 @@
...
+    allowedMagnifications: Option[AllowedMagnifications] = None
+) {
+  def resolutionRestrictions: ResolutionRestrictions =
+    allowedMagnifications match {
+      case None => ResolutionRestrictions.empty
+      case Some(allowedMags) =>
+        if (allowedMags.shouldRestrict) ResolutionRestrictions(Some(allowedMags.min), Some(allowedMags.max))
+        else ResolutionRestrictions.empty
+    }
+}
```
fm3, line 28: note for reviewer: follow-up refactoring issue #4901

webknossos-tracingstore/app/com/scalableminds/webknossos/tracingstore/tracings/volume/VolumeTracingService.scala
```diff
@@ -205,8 +213,8 @@ class VolumeTracingService @Inject()(
             WKWFile.read(is) {
               case (header, buckets) =>
                 if (header.numBlocksPerCube == 1) {
-                  parseWKWFilePath(fileName.toString).map { bucketPosition: BucketPosition =>
-                    if (resolutionsMatch || bucketPosition.resolution == Point3D(1, 1, 1)) {
+                  parseWKWFilePath(fileName.toString).map { _ =>
+                    if (buckets.hasNext) {
```
fm3, line 217: note for reviewer: buckets.hasNext should always be true here, but bug #4905 makes it not so sometimes.

webknossos-tracingstore/app/com/scalableminds/webknossos/tracingstore/tracings/volume/VolumeTracingService.scala
```diff
@@ -253,7 +263,7 @@ class VolumeTracingService @Inject()(
     def addLabelSet(labelSet: mutable.Set[UnsignedInteger]): Unit = labelSets += labelSet
 
     private def prepareLabelMaps(): Unit =
-      if (labelSets.isEmpty || labelMaps.nonEmpty) {
+      if (labelSets.length <= 1 || labelMaps.nonEmpty) {
```
fm3, line 266: if there is exactly one labelSet no remapping is required (upload of single annotation)

MIGRATIONS.unreleased.md
```diff
@@ -6,7 +6,13 @@ This project adheres to [Calendar Versioning](http://calver.org/) `0Y.0M.MICRO`.
 User-facing changes are documented in the [changelog](CHANGELOG.released.md).
 
 ## Unreleased
-- To convert individual legacy volume annotations to multi-resolution volume annotations, download and re-upload them (upload may take a while).
+- As volume annotations in arbitrary magnifications are now supported and the behavior of magnification restrictions of tasks has changed (allow full zoom, but disable tools unless in correct magnification), you may want to restrict all volume and hybrid task types to 1-1 to achieve the old behavior (mag1-only):
```
philippotto, line 9:
```suggestion
- As volume annotations in arbitrary magnifications are now supported and the behavior of magnification restrictions of tasks has changed (allow full zoom, but disable tools unless in correct magnification), you may want to restrict all volume and hybrid task types to mag 1 to achieve the old behavior (mag1-only):
```

frontend/javascripts/oxalis/view/settings/dataset_settings_view.js
```diff
@@ -76,9 +78,76 @@ type DatasetSettingsProps = {|
...
+      visible={visible}
+      footer={null}
+      width={800}
+      maskClosable={false}
+    >
+      <p>
+        This annotation does not have volume annotation data in all resolutions. Consequently,
+        annotation data cannot be rendered at all zoom values. By clicking &quot;Downsample&quot;,
+        webKnossos will use the best resolution of the volume data to create all dependent
+        resolutions.
```
daniel-wer, line 106: This may overwrite existing resolution, for example, if mags 1 and 2 exist, mag 2 is overwritten, right? If so, this should be noted here or maybe better, in the next paragraph: "The following resolutions will be overwritten ...".

frontend/javascripts/oxalis/view/settings/dataset_settings_view.js
```diff
@@ -76,9 +78,76 @@ type DatasetSettingsProps = {|
...
+          </li>
+          <li>An old annotation was uploaded which did not include all resolutions.</li>
+          <li>The annotation was created in a task that was restricted to certain resolutions.</li>
+          <li>The dataset was mutated to have more resolutions.</li>
+        </ul>
+      </p>
+
+      <p style={{ fontWeight: "bold" }}>
+        Note that this action might take a few minutes. Afterwards, the annotation is reloaded.
+        Also, the version history of the volume data will be reset.
```
daniel-wer, line 128: Out of curiosity, why does the history need to be reset?

frontend/javascripts/oxalis/view/settings/dataset_settings_view.js
```diff
@@ -76,9 +78,76 @@ type DatasetSettingsProps = {|
...
+  isDownsampleVolumeModalVisible: boolean,
+|};
+
+class DatasetSettings extends React.PureComponent<DatasetSettingsProps, State> {
+  constructor() {
+    super();
+    this.state = {
+      isDownsampleVolumeModalVisible: false,
+    };
+  }
```
daniel-wer, line 149:
```suggestion
  state = {
    isDownsampleVolumeModalVisible: false,
  };
```

youri-k (commented):
Backend LGTM apart from my one suggestion

frontend/javascripts/oxalis/view/settings/dataset_settings_view.js
daniel-wer: If my previous comment holds true, is there a technical reason why the best resolution is used for downsampling, possibly overwriting existing resolutions, instead of downsampling from the "worst" existing resolution?

daniel-wer (commented):
Very cool :+1: I like that you are clearly communicating to the user why things are the way they are!

I have a couple of clarifying questions and suggestions, but overall LGTM :)

frontend/javascripts/oxalis/api/api_latest.js
```diff
@@ -744,6 +745,24 @@ class TracingApi {
...
+     has finished.
+     This function can only be used for non-tasks.
+   */
+  async downsampleSegmentation() {
+    const state = Store.getState();
+    const { annotationId, annotationType } = state.tracing;
+    if (state.task != null) {
+      throw new Error("Cannot downsample segmentation for a task.");
+    }
+
```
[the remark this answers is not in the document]
philippotto: Good point! I added an extra check and also created a separate issue for the refactoring part: https://github.com/scalableminds/webknossos/issues/4913

frontend/javascripts/oxalis/view/settings/dataset_settings_view.js
philippotto: I think, the entire `tracing` object within the annotation is swapped out with a new one. Since history is only persisted for a tracing, such swap-outs cannot be recorded. The reason for swapping out the entire volume tracing is probably, since the new tracing needs to have a different resolution set?

Correct me if I'm wrong, @fm3!

Obviously, it would be cool if we could keep the history. @fm3 Would this be feasible for a later iteration?

frontend/javascripts/oxalis/view/settings/dataset_settings_view.js
philippotto: Technically, you are correct, but conceptionally the mags should always be in sync. Thus, discarding downsampled mags and re-creating them should not do any harm, which is why I wouldn't want to unsettle/confuse the users with that information. Do you agree?

frontend/javascripts/oxalis/model/sagas/task_saga.js
```diff
@@ -124,3 +131,58 @@ export default function* watchTasksAsync(): Saga<void> {
...
+  yield* warnMaybe();
+
+  while (true) {
+    yield* take([
+      "ZOOM_IN",
+      "ZOOM_OUT",
+      "ZOOM_BY_DELTA",
+      "SET_ZOOM_STEP",
+      "SET_STORED_LAYOUTS",
+      "SET_TOOL",
```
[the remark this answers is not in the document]
philippotto: Good catch regarding the `SET_TOOL` action :+1: This is a leftover from a version where the saga switched the tool (instead of the volume tools view). I'll remove it.

> Also, if I remember correctly, the zoom step could also be influenced by the SET_INPUT_CATCHER_RECT and SET_INPUT_CATCHER_RECTS actions.

I just verified that `SET_STORED_LAYOUTS` is called when resizing the viewports and when toggling the fullscreen mode. So, the other actions should not be necessary to listen to. Let me know if other cases come to your mind where it could be important, too.

Probably, the store definition for the input_catcher_rects and the layouts could use some clean up, as their semantics overlap here. However, I'm hesitant to touch that code :see_no_evil:

frontend/javascripts/oxalis/view/settings/dataset_settings_view.js
fm3: Yes, that’s correct, the downsample call is a variant of duplicate. This was the easiest way to implement this and ensure no version conflicts etc. However, it does not have to be done like this, a future iteration could probably implement the downsampling as writing new version in an existing volume tracing.

The backend stdout logging also includes both the old and the new tracing id along with the annotation id if this happens, so if a user *really* needs to undo this, we can do it for them in the db.

fm3:
- [ ] I noticed a bug when importing a volume via dnd into an existing one: the id remapping seems broken, works on master. I’ll have a look.

frontend/javascripts/oxalis/model/sagas/task_saga.js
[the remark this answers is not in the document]
daniel-wer: > I just verified that SET_STORED_LAYOUTS is called when resizing the viewports and when toggling the fullscreen mode. So, the other actions should not be necessary to listen to. Let me know if other cases come to your mind where it could be important, too.

Ah okay, cool, thanks for checking :+1:

frontend/javascripts/oxalis/view/settings/dataset_settings_view.js
daniel-wer: > Technically, you are correct, but conceptionally the mags should always be in sync.

Yes if that's the case, I agree :)

daniel-wer (approved):
LGTM :+1:

PR #4892, opened 2020-10-23
Organization restricted datastore

### LINKED ISSUE #4849: Organization-specific datastore
It would be cool, if users of webknossos.org could attach their own datastores/wkconnect (in their HPC). For that, I think it is necessary to restrict the datastore to a single organization.

youri-k:
- https://___.webknossos.xyz

### Steps to test:
- set up local standalone datastore
- change db column `onlyAllowedOrganization` to random id => upload to that ds should fail
- change db column `onlyAllowedOrganization` to organization id => upload to that ds should work when logged in as user of that org

### Issues:
- fixes #4849 

------
- [x] Ready for review


fm3:
Changes thus far LGTM, but I think another change is required: datasets reported by that datastore should be ignored (probably with command line warning) if they are not for its organization

conf/messages
```diff
@@ -116,6 +116,8 @@ dataSet.initialTeams.teamsNotEmpty=Dataset already has allowed teams
 dataset.initialTeams.invalidTeams=Can only assign teams of user
 dataset.initialTeams.timeout=Timeout while setting initial teams. Was the request sent manually?
 dataset.delete.disabled=Dataset deletion is disabled for this webKnossos instance
+dataSet.upload.Datastore.restricted=Your organization is not allowed to upload datasets to this datastore. Please choose another datastore.
+dataSet.upload.validation.failed=Failed to validate Dataset information for upload.
```
fm3, line 120: Will the user see an error chain with the real cause here? If not, maybe mention possible causes/actions so that the user is not stuck when encountering this unspecific message.

app/models/binary/DataSetService.scala
```diff
@@ -127,6 +127,10 @@ class DataSetService @Inject()(organizationDAO: OrganizationDAO,
           .findOneByName(orgaTuple._1)
           .futureBox
           .flatMap {
+            case Full(organization) if dataStore.onlyAllowedOrganization.exists(_ != organization._id) =>
+              logger.info(
+                s"Ignoring ${orgaTuple._2.length} reported datasets for forbidden organization ${orgaTuple._1} on DataStore ${dataStore.name}")
```
fm3, line 132:
```suggestion
                s"Ignoring ${orgaTuple._2.length} reported datasets for forbidden organization ${orgaTuple._1} from organization-specific datastore ${dataStore.name}")
```

fm3 (changes requested):
Backend LGTM.

Sorry, one more concern: should the datastore be visible to users of other organizations? Maybe adapt the accessQuery? (Or if that causes trouble, at least hide it from the upload form dropdown menu? The refresh button should probably also not talk to datastores specific to other organizations)

conf/messages
youri-k: Yes, he will see an error chain.

fm3 (approved):

PR #4899, opened 2020-10-27
Fix unwanted group expansions

daniel-wer:
This fixes a bug which was introduced in #4897.

The `finishSearchCallback` is not only called if the parent groups of the found node are expanded, but also when the tree data changes (where the parent groups of the found node are not expanded). I implemented a fix for that by checking for each element on the path whether it was actually expanded, but that was rather complicated. Therefore, I solved this issue in a more general way by updating the `expandedGroupIds` list whenever the `onChange` handler of the sortable-tree is called because the tree data changed. This way the two structures are always in sync. I think this solution is much less error-prone and the performance impact should be negligible.

- https://___.webknossos.xyz

### Steps to test:
- The bug could be triggered as follows. Create a tracing with this structure:
  - groupA
    - tree1
  - groupB
    - tree2
- Activate tree1 and then collapse groupA. Now change the visibility of groupB or tree2.
- On master this would expand groupA again, which should no longer happen on this branch.

- Also check whether group expansion, toggling of all sub groups works as before.

------
- [x] Ready for review

### DIFF
--- a/frontend/javascripts/oxalis/view/right-menu/tree_hierarchy_view.js
+++ b/frontend/javascripts/oxalis/view/right-menu/tree_hierarchy_view.js
@@ -9,7 +9,6 @@ import SortableTree from "react-sortable-tree";
 import _ from "lodash";
 import type { Dispatch } from "redux";
 import { type Action } from "oxalis/model/actions/actions";
-import update from "immutability-helper";
 import {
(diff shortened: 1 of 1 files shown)


philippotto (approved):
Cool! Works for me and looks a lot cleaner :) Let's hope the performance also turns out to be alright.

PR #4900, opened 2020-10-28
Refactor backend bucket loading

youri-k:
- https://___.webknossos.xyz

### Steps to test:
- view dataset => should still work

### Issues:

------
- [x] Ready for review


fm3 (approved):
LGTM :) maybe even include more information in the BinaryDataController logging (e.g. if agglomerate mapping is active etc. this should occur relatively rarely anyway, right? then we could even print the entire request)

PR #4902, opened 2020-10-29
Hotfix: preserve defaultViewConfiguration

youri-k:
fixes mistake introduced in #4845
- [x] Ready for review


fm3 (approved):
good catch :)

PR #4914, opened 2020-11-05
Bump wkw version to 1.1.9

### LINKED ISSUE #4905: some volume buckets cannot be decompressed after reupload
I can reproduce this on ROI2017_wkw (with uint32 fallback segmentation).
Those buckets were skipped silently as `initializeWithDataMultiple`’s `buckets.next` has undefined behavior and `hasNext` is never checked.
Traced this back to the lz4 decompression in wkw. `net.jpountz.lz4.LZ4Exception: Error decoding offset 50898 of input buffer`
are there wrong assumptions about header length / offsets in scala wkw? does the actual lz4 compression fail? we should investigate.
update: looks like lz4 itself is not the issue but the data passed to it is already corrupt. may be from unpacking the nested volume zip…

fm3:
- new webknossos-wrap version 1.1.9 includes fix in https://github.com/scalableminds/webknossos-wrap/pull/57 so it should fix the volume annotation upload bug where some data is missing.

### Steps to test:
- download + reupload relatively large volume annotation
- all buckets should be present

### Issues:
- fixes #4905 

------
- [x] Needs tracing update after deployment
- [x] Ready for review


youri-k (approved):
LGTM :shipit:

Issue #4835, opened 2020-09-23
Improve isosurface loading for rendering a single cell


philippotto:
Right now, it's a bit tedious to get a nice rendering of a complete cell. As I wrote [here](https://discuss.webknossos.org/t/experimental-isosurface-rendering/1267/8?u=potto), there are two possible culprits:
- our artificial limit for how much isosurface chunks can be loaded at once
- thin processes might abort the full computation

We should have a look at how to improve this. For example, maybe the limit can be removed altogether as long as the computation is throttled reasonably.

- [X] Also, listing the isosurfaces/meshes explicitly could greatly improve the UX. See https://github.com/scalableminds/webknossos/pull/4567#issuecomment-673333038

philippotto:
Also, there seems to be a bug when navigating to another area while an isosurfaces is computed. See: https://discuss.webknossos.org/t/experimental-isosurface-rendering/1267/9?u=potto

philippotto:
I think, this issue got irrelevant with the support for precomputed meshes.

philippotto:
Users are reporting that ad-hoc meshes are not loaded completely. This is likely happening due to the limit mentioned above.

> For example, maybe the limit can be removed altogether as long as the computation is throttled reasonably.

I think, this would be the easiest way to improve the situation right now.

PR #4917, opened 2020-11-06
Isosurface List

grittaweisheit:
This PR adds a list of isosurfaces in the meshes tab while viewing a dataset.

- https://___.webknossos.xyz

### Steps to test:
- view a dataset
- enable isosurfaces
- open meshes tab
- create isosufaces by pressing shift and clicking on a cell
- the list shoul diplay evey isosurface
- delete isosurfaces by pressing ctrl and clicking on the isosurface
- deleted isosurfaces should not be in the list
- download an isosurface by clicking on its download icon in the list

### Issues:
- #4835 
------
- [x] Ready for review

### DIFF
--- a/frontend/javascripts/oxalis/store.js
+++ b/frontend/javascripts/oxalis/store.js
@@ -463,6 +463,7 @@ export type OxalisState = {|
   +viewModeData: ViewModeData,
   +activeUser: ?APIUser,
   +uiInformation: UiInformation,
+  +isosurfaces: Array<number>,
 |};
 
 const sagaMiddleware = createSagaMiddleware();
--- a/frontend/javascripts/oxalis/view/right-menu/meshes_view.js
+++ b/frontend/javascripts/oxalis/view/right-menu/meshes_view.js
@@ -1,6 +1,6 @@
 // @flow
 
-import { Button, Checkbox, Icon, Input, Modal, Spin, Tooltip, Upload } from "antd";
+import { Button, Checkbox, Icon, Input, List, Modal, Spin, Tooltip, Upload } from "antd";
 import type { Dispatch } from "redux";
 import { connect } from "react-redux";
 import React from "react";
@@ -23,6 +23,7 @@ import { readFileAsArrayBuffer } from "libs/read_file";
 import { setImportingMeshStateAction } from "oxalis/model/actions/ui_actions";
 import ButtonComponent from "oxalis/view/components/button_component";
 import { trackAction } from "oxalis/model/helpers/analytics";
+import { jsConvertCellIdToHSLA } from "oxalis/shaders/segmentation.glsl";
 
 const ButtonGroup = Button.Group;
 
@@ -76,6 +77,8 @@ const mapStateToProps = (state: OxalisState) => ({
   meshes: state.tracing != null ? state.tracing.meshes : [],
   isImporting: state.uiInformation.isImportingMesh,
   isHybrid: state.tracing.volume != null && state.tracing.skeleton != null,
+  isosurfaces: state.isosurfaces,
+  mappingColors: state.temporaryConfiguration.activeMapping.mappingColors,
 });
 
 const mapDispatchToProps = (dispatch: Dispatch<*>) => ({
@@ -131,6 +134,23 @@ class MeshesView extends React.Component<Props, { currentlyEditedMesh: ?MeshMeta
   };
 
   render() {
+    const convertHSLAToCSSString = ([h, s, l, a]) =>
+      `hsla(${360 * h}, ${100 * s}%, ${100 * l}%, ${a})`;
+
+    const convertCellIdToCSS = id =>
+      convertHSLAToCSSString(jsConvertCellIdToHSLA(id, this.props.mappingColors));
+
+    const renderIsosurfaceListItem = (cellId: number) => (
+      <List.Item>
+        {" "}
+        <span
+          className="colored-circle"
+          style={{ backgroundColor: convertCellIdToCSS(cellId) }}
+        />{" "}
+        Segment {cellId}
+      </List.Item>
+    );
+
     return (
       <div className="padded-tab-content">
         <ButtonGroup style={{ marginBottom: 12 }}>
@@ -158,6 +178,7 @@ class MeshesView extends React.Component<Props, { currentlyEditedMesh: ?MeshMeta
             </Tooltip>
           )}
         </ButtonGroup>
+        <List dataSource={this.props.isosurfaces} renderItem={renderIsosurfaceListItem} />
         {this.state.currentlyEditedMesh != null ? (
           <EditMeshModal
             initialMesh={this.state.currentlyEditedMesh}
--- a/frontend/javascripts/oxalis/view/right-menu/meshes_view.js
+++ b/frontend/javascripts/oxalis/view/right-menu/meshes_view.js
@@ -1,6 +1,6 @@
 // @flow
 
-import { Button, Checkbox, Icon, Input, Modal, Spin, Table, Tooltip, Upload } from "antd";
+import { Button, Checkbox, Icon, Input, List, Modal, Spin, Tooltip, Upload } from "antd";
 import type { Dispatch } from "redux";
 import { connect } from "react-redux";
 import React from "react";
@@ -28,7 +28,6 @@ import { trackAction } from "oxalis/model/helpers/analytics";
 import { jsConvertCellIdToHSLA } from "oxalis/shaders/segmentation.glsl";
 
 const ButtonGroup = Button.Group;
-const { Column } = Table;
 
 export const stlIsosurfaceConstants = {
   isosurfaceMarker: [105, 115, 111], // ASCII codes for ISO
@@ -138,19 +137,38 @@ class MeshesView extends React.Component<Props, { currentlyEditedMesh: ?MeshMeta
 
   render() {
     const refreshButton = (segmentId: number) => (
-      <Icon key="refresh-button" type="redo" onClick={console.log(segmentId)} />
+      <Tooltip title="refresh isosurface">
+        <Icon key="refresh-button" type="redo" onClick={console.log(segmentId)} />
+      </Tooltip>
     );
     const downloadButton = (segmentId: number) => (
-      <Icon
-        key="download-button"
-        type="vertical-align-bottom"
-        onClick={() => Store.dispatch(triggerIsosurfaceDownloadAction(segmentId))}
-      />
+      <Tooltip title="download isosurface">
+        <Icon
+          key="download-button"
+          type="vertical-align-bottom"
+          onClick={() => Store.dispatch(triggerIsosurfaceDownloadAction(segmentId))}
+        />
+      </Tooltip>
     );
     const convertHSLAToCSSString = ([h, s, l, a]) =>
       `hsla(${360 * h}, ${100 * s}%, ${100 * l}%, ${a})`;
     const convertCellIdToCSS = id =>
       convertHSLAToCSSString(jsConvertCellIdToHSLA(id, this.props.mappingColors));
+
+    const renderListItem = (cellId: number) => (
+      <List.Item
+        actions={[refreshButton(cellId), downloadButton(cellId)]}
+        style={{ marginBottom: -15 }}
+      >
+        <span
+          className="circle"
+          style={{
+            backgroundColor: convertCellIdToCSS(cellId),
+          }}
+        />{" "}
+        Segment {cellId}
+      </List.Item>
+    );
 
     return (
       <div className="padded-tab-content">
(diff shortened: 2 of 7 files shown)


philippotto:
Awesome! This is really, really cool and will make the isosurface feature so much more usable. Also, great job for showing the segment colors in the list as circles :+1: 

I think, we can iterate on this PR a bit to fully leverage the new UI for the isosurfaces. The first things which come to my mind are the following:
- [ ] clicking on "Segment X" could move the current position in webKnossos to that isosurface (where exactly is not too important for now. Maybe store the seed position from where the calculation was started?)
- [ ] a reload-button would be great (functionality would probably be very similar to the existing reload-button in the 3D viewport). however, that reload-button should only turn into a spinner if the segment which belongs to the button is loading. in another iteration, we could even distinguish between "completely reload the isosurface" and "continue computation".
- [ ] a delete-button would be cool (the current way of deleting via ctrl-click is very hidden and could also be removed if it makes things easier)

You probably need to move more state into the global `Store` to enable this. So, instead of having `isosurfaces` as a simple list of ids, you probably need a dict-like structure like this:

```
isosurfaces: {
  3: {
    segmentId: 3, // this is redundant to the dict-key, but this makes passing around a dict-value simpler
    seedPosition: [1, 2, 3],
    isLoading: false
  }
}
```

Note that the `trees` property in the store also uses a dict-like structure, but that is actually a `DiffableMap`. I think, for your usecase that would be a bit overkill. Instead, I would use a plain JS Object. In the reducers, you can make use of the deep-update-helpers to update an isosurface element (something like `updateKey3(state, "isosurfaces", segmentId, "isLoading", true);`

Let me know in case you have any questions!

grittaweisheit:
Thanks for the input! The things you suggested are now implemented. I'm open for further suggestions :).

philippotto:
Very cool! I just tested it and the new features feel very good :+1:

I noticed a few small things which we could improve:
- When the list is empty, nothing is shown. I think, we should use the space to explain the feature a bit. Similar to the empty tree tab: 
image
- It's not clear that the list items are clickable. At least, change the `cursor` via css to `pointer`. Also, the active segment should be highlighted somehow (e.g., some background color as it's done in the tree tab)
- The layout is not quite right if the segment ids differ in their lengths. The buttons should be right-aligned properly. Alternatively, only show the buttons on hover (then, you don't need to align them). Also, I think, that you can remove the vertical bars between the buttons, as the buttons itself are pretty well distinguishable as different click targets.
image
- Currently, the feature is enabled/disabled via a setting in the `Dataset` tab (called "Render Isosurfaces (Beta)"). I think, we can remove it from that spot and instead show the setting in the "Meshes" tab. I'm thinking of a [switch button](https://ant.design/components/switch/); maybe with a label (e.g., see the "Text & icon" label in the antd doc). At that point, we might as well remove the "(Beta)" part :)

Only fyi:
- If one uses the download feature, deletes the isosurface and then re-imports it via the "import" button, the isosurface is not shown in the list. However, this is probably a bit more complicated to fix. I built the feature back then, and as far as I remember, the import merely attaches the STL file to the scene and sets the correct color. The proper isosurface loading logic is completely decoupled from this. I'll create a separate issue for this :)

frontend/javascripts/oxalis/store.js
```diff
@@ -467,6 +467,7 @@ export type OxalisState = {|
   +viewModeData: ViewModeData,
   +activeUser: ?APIUser,
   +uiInformation: UiInformation,
+  +isosurfaces: Object,
```
philippotto, line 470: From a pure JavaScript-perspective, it is true that this is simply an `Object`. However, we can (and should) provide more information to flow so that the static validation is more powerful. See https://flow.org/en/docs/types/objects/#toc-objects-as-maps. So, this should be something like:

```js
type IsosurfaceInfo = {
  ..
}
...

type ... = {
  ...
  +isosurfaces: {[segmentId: string]: IsosurfaceInfo}
  ...
}
```

frontend/javascripts/oxalis/view/right-menu/meshes_view.js
```diff
@@ -131,6 +139,67 @@ class MeshesView extends React.Component<Props, { currentlyEditedMesh: ?MeshMeta
   };
 
   render() {
+    const moveToIsosurface = (seedPosition: Vector3) => {
+      Store.dispatch(setPositionAction(seedPosition));
+    };
+    const downloadButton = (segmentId: number) => (
```
philippotto, line 145: The naming of `downloadButton`, `refreshButton` and `deleteButton` is a bit unusual for these functions. Either rename them to `getDownloadButton`, `getRefreshButton` etc. or capitalize them to `DownloadButton` etc. In the latter case, one would use the functions as react components (i.e., `<DownloadButton segmentId=... />` on the caller side and `({segmentId}) => ...` on the definition side). The first variant is probably easier to adapt to :)

frontend/javascripts/oxalis/view/right-menu/meshes_view.js
```diff
@@ -131,6 +139,67 @@ class MeshesView extends React.Component<Props, { currentlyEditedMesh: ?MeshMeta
...
+      <Tooltip title="download isosurface">
```
philippotto, line 146: We usually use title case for UI labels (https://titlecase.com/). Same for the other button's tooltips :)

frontend/javascripts/oxalis/view/right-menu/meshes_view.js
```diff
@@ -158,6 +227,15 @@ class MeshesView extends React.Component<Props, { currentlyEditedMesh: ?MeshMeta
             </Tooltip>
           )}
         </ButtonGroup>
+
+        <List
+          dataSource={Object.values(this.props.isosurfaces)}
+          size="small"
+          split={false}
+          renderItem={renderListItem}
+          locale={{ emptyText: " " }}
```
philippotto, line 236: Here you can put some small explanation of the list (i.e., what is displayed and how can the user add something to this list).

frontend/javascripts/oxalis/view/right-menu/meshes_view.js
```diff
@@ -131,6 +139,67 @@ class MeshesView extends React.Component<Props, { currentlyEditedMesh: ?MeshMeta
...
+      convertHSLAToCSSString(jsConvertCellIdToHSLA(id, this.props.mappingColors));
+
+    const renderListItem = (isosurface: Object) => {
+      const { segmentId, seedPosition, isLoading } = isosurface;
+      return (
+        <List.Item
+          actions={[
+            downloadButton(segmentId),
+            refreshButton(segmentId, isLoading),
+            deleteButton(segmentId),
```
philippotto, line 186: Ah, I see, the vertical bars and the "wrong" alignment is coming from antd's list. Might be hard to fix then :thinking: For the alignment, I don't have an idea, but for the vertical bar, one could try to simply pass an array with length one, like this:

```
<List.Item
          actions={[
            <div>
              {downloadButton(segmentId)}
              {refreshButton(segmentId, isLoading)}
              {deleteButton(segmentId)}
            </div>
...
```

frontend/javascripts/oxalis/view/right-menu/meshes_view.js
grittaweisheit: Good to know, thanks!

frontend/javascripts/oxalis/store.js
grittaweisheit: Yes, that's definitely better, thank you.

frontend/javascripts/oxalis/view/right-menu/meshes_view.js
grittaweisheit: Works nicely, thanks for the suggestion!

grittaweisheit:
Thanks a lot for your input! I especially like the ideas for the list items.
Let me know if there's something else that can be improved :-).

philippotto:
Great! This is progressing nicely :)

I've got some new ideas, but we are very close, I think.

You probably haven't seen it, but the "Meshes" tab can also show STL files which one has dragged into the view. The feature is quite hidden which is also the reason why it currently clashes with the isosurface part. Right now it would look like this:

image

This is why I had the idea to restructure the view a bit:
 I created a small mock up for my new ideas:

image

The view would have two sections (ad-hoc generated isosurfaces and the uploaded STL list). Next to each header could be an "i"-icon (the same one as can be found next to the layer names in the dataset settings) which explains what the sections are. The plus button would be equal to the import button (for both headers, you could use the same import button to make things easier).

Additionally, I added two buttons which allow to have an isosurface loaded for the currently active cell or for the currently centered position. I think, this will be easier to use then the shift-clicking stuff. An added benefit would be that we can get rid of the user setting altogether :) Having the entire feature hooked to an active cell also caused some problems in the past (since there is no such as thing as an active cell when one only views a dataset). The isosurface saga maintains an own version of the "active cell" for that reason. With the dedicated buttons, we could get rid of this (also the upcoming context menu will become helpful here).
The "Load Isosurface for active Segment"-button could be hidden when one is in dataset-view mode.

---------------
Two other things:
- it would be cool if the the highlighting of the "active" segment would look more similar to the highlighting in the trees tab. right now, the two lists don't feel consistent:
image
- the highlighting of the "active" segment currently only has one meaning: the user clicked on the segment. if I select another cell (via shift-click) or move somewhere else, the segment is still highlighted. therefore, I suggest to adapt the highlighting dynamically. there are two options: (a) always highlight the segment which is at the current position (b) always highlight the segment which is active according to the store state. I think, the latter option would be easier, but wouldn't directly work for the dataset-view-mode (see my section above explaining that we could get rid of that ^^). Maybe a hybrid makes sense? Not sure about this. also it's not super important right now.

---------

It would be very cool if you could have a look at my suggestions and feel free to deviate from them and integrate your own ideas :) Also feel free to hit me up if you want to discuss details or have questions.

grittaweisheit:
I really liked your mock up ideas! The only deviation I made was putting the buttons for loading isosurfaces directly beneath the isosurface heading. I thinks that's a little bit more intuitive. 

I now highlight a list item, when the segment is at the current position.

There are only a few things I would like to mention, that I don't find ideal but don't know how to fix and want to check if it would be worth the effort to figure it out. I would appreciate your input to them!
- When loading the isosurface to the active cell there is no position I can anchor the isosurface to, which makes it impossible to navigate to the cell when clicking on the corresponding list entry. I'm not sure where I could get a suitable position to fix that. 
- The same goes for imported isosurfaces.
- When uploading an isosurface I can include it in the list, but it can't be removed like normally loaded isosurfaces and instead only vanished from the list, while the isosurface itself still exists in the 3D Viewport.

Additionally I have two questions:
- Should the `renderIsosurface` setting be removed entirely? 
- The `shift + click` shortcut behaves differently in different hybrid tracings. There it only loads the isosurface for the active cell and not the one you click on. Should it be removed, fixed or kept?

I would appreciate your input :)

philippotto:
Great stuff!!

> I really liked your mock up ideas! The only deviation I made was putting the buttons for loading isosurfaces directly beneath the isosurface heading. I thinks that's a little bit more intuitive.

Yes, that's good :+1: I'd add some whitespace between the two buttons. However, maybe the load-button for the active cell should be removed, anyway, since it causes problems due to the missing seed position.

> * When loading the isosurface to the active cell there is no position I can anchor the isosurface to, which makes it impossible to navigate to the cell when clicking on the corresponding list entry. I'm not sure where I could get a suitable position to fix that.

Very good point. Then, I'd remove the button for now :smile: 

> * The same goes for imported isosurfaces.

Hm, maybe we should store the seed position in the STL file. Right now, we already embed the segment id in the output file (doesn't belong to the official file format). However, it's fine for now if this does not work yet in my opinion.


> * When uploading an isosurface I can include it in the list, but it can't be removed like normally loaded isosurfaces and instead only vanished from the list, while the isosurface itself still exists in the 3D Viewport.

Do you know why this does not work? From what I see the `SceneController` adds a threejs group to `this.isosurfacesGroupsPerSegmentationId` from which later deletion should be possible. We can also call about this if that's easier :) 

> * Should the `renderIsosurface` setting be removed entirely?

Very good question. I thought about this for a bit and came to the following conclusion. until now, the setting had two uses: 1) initiate the isosurface rendering 2) update the isosurface. for 1) we now have the button (and later the context menu). and for 2) I think that the isosurface should always be updated for a flycam action (see `yield _takeEvery(FlycamActions, ensureSuitableIsosurface);` in the isosurface saga) *if* there is an activated isosurface entry in the list. So, if the user has not used the isosurfaces, nothing is loaded (equal to an disabled isosurface setting). As soon as the user has loaded an isosurface, that isosurface is updated until it's deleted or another isosurface is activated.
Does this make sense? Then, the setting could be removed.

> * The `shift + click` shortcut behaves differently in different hybrid tracings. There it only loads the isosurface for the active cell and not the one you click on. Should it be removed, fixed or kept?

I think, we should remove the (isosurface related) shortcut (shift+click to activate a cell should still work in a volume tracing). An explicit button works better for now and later on we can re-add it to the context menu which will land soon.


----------

Other than that, I have some minor suggestions:

image

- the blue background for the active list item should ideally only go as far as the segment text (similar to the active tree). Then, the icons which appear on hover are also on a white background automatically which increases the contrast.
- the meshes-headline should have some `margin-top`
- the meshes-section should also have an empty-placeholder (e.g., "There are no meshes. You can import an STL file to change this.")

grittaweisheit:
Thanks again for your input! 

I removed the "load active..." button, the shortcut for isosurface-loading, the renderIsosurface setting and adjusted the setting of the active isosurface a little bit so that the last used isosurface is the active one.
The deletion of imported isosurfaces works again. I'm not sure why it didn't work but works now again.
Furthermore I made the lists of isosurfaces prettier according to your suggestions.

philippotto (approved):
Awesome stuff! Thanks for going through all the iterations. This will be a huge improvement and boost for the isosurfaces :tada: :tada: :tada: 

I took the brazen liberty to push some finishing touches which caught my eye when I tested the last version. I think, this was easier than writing the stuff down again and having you go through it :) I hope this was alright!

It now looks like this:
image

I'm still not 100% satisfied with the prominent "load-button", but I think it's better to have it next to the other buttons instead of in a separate line :thinking: 

From my point of view, this PR should be ready for merging :) Go ahead if you agree!

philippotto:
We forgot to add a changelog entry :see_no_evil: I added that in https://github.com/scalableminds/webknossos/pull/5102, though.

PR #4920, opened 2020-11-09
Fix broken overwrite-mode behavior

philippotto:
- https://___.webknossos.xyz

### Steps to test:
- I used both brush and trace tool with left/right click (for draw vs delete) to test both overwrite-behaviors (via ctrl and mode switch)

### Issues:
- this bug was introduced with #4755 

------
- [X] Ready for review


daniel-wer (approved):
:+1:

PR #4922, opened 2020-11-09
Refactor TaskType Resolution Restrictions

### LINKED ISSUE #4901: Refactor TaskType AllowedMagnifications
Currently redundantly coded, and saved as jsValue in database, should be structured. compare ResolutionRestrictions case class introduced in #4891

fm3:
The old format for tasktype settings
```
  +allowedMagnifications?: {
    shouldRestrict: boolean,
    min?: number,
    max?: number,
  },
```
had threefold redundancy (field is optional, shouldRestrict boolean, numbers are optional), creating a lot of possibilities for inconsistent states. I propose we change that to 
```
  +resolutionRestrictions: {
    min?: number,
    max?: number,
  },
```
The old format fitted the form in tasktype creat view better, though, there needs to be an adapter for the formValues.

### TODO:
- [x] update schema + scala code
- [x] sql evolution
- [x] use new format in task type create view
- [x] use new format in flycam accessor

### Steps to test:
- create task type with restrictions, create task
- fetch task, resolutions should be restricted

### Issues:
- fixes #4901 

------
- [x] Ready for review


MIGRATIONS.unreleased.md
```diff
@@ -16,4 +16,5 @@ and (settings_allowedmagnifications is null or settings_allowedmagnifications::j
 
 ### Postgres Evolutions:
 - [057-add-layer-specific-view-configs.sql](conf/evolutions/056-add-layer-specific-view-configs.sql)
```
philippotto, line 18: I know that this has been like this before, but it seems to be wrong? Shouldn't it be `57` in the path?

frontend/javascripts/oxalis/model/accessors/flycam_accessor.js
```diff
@@ -294,9 +294,9 @@ export function getValidTaskZoomRange(
   respectRestriction: boolean = false,
 ): [number, number] {
   const defaultRange = [userSettings.zoom.minimum, Infinity];
-  const { allowedMagnifications } = state.tracing.restrictions;
+  const { resolutionRestrictions } = state.tracing.restrictions;
 
-  if (!respectRestriction || !allowedMagnifications || !allowedMagnifications.shouldRestrict) {
+  if (!respectRestriction || !resolutionRestrictions) {
```
philippotto, line 299: Isn't the new format

```
  +resolutionRestrictions: {
    min?: number,
    max?: number,
  },
```

? Then, this would need to check whether min and max are both null?

frontend/javascripts/admin/tasktype/task_type_create_view.js
```diff
@@ -110,8 +110,13 @@ class TaskTypeCreateView extends React.PureComponent<Props, State> {
...
     formValues.recommendedConfiguration = jsonStringify(formValues.recommendedConfiguration);
-    formValues.settings.allowedMagnificationsmin = formValues.settings.allowedMagnifications.min;
-    formValues.settings.allowedMagnificationsmax = formValues.settings.allowedMagnifications.max;
+    formValues.settings.resolutionRestrictionsForm = {
+      shouldRestrict:
+        formValues.settings.resolutionRestrictions.min != null ||
+        formValues.settings.resolutionRestrictions.max != null,
+      min: formValues.settings.resolutionRestrictions.min || 1,
+      max: formValues.settings.resolutionRestrictions.max || 512,
+    };
```
philippotto, line 119: I like that you separated the form vs not-form aspects explicitly here. However, I'm wondering whether we could do it without having both values simultaneously in the `formValues`?

Something like this:
```
function getFormValuesForTaskType() {
  const taskType = _.merge({}, defaultValues, taskType);
  formValues.recommendedConfiguration = jsonStringify(formValues.recommendedConfiguration);

  // explain here what's going on
  const resolutionRestrictions = taskType.settings.resolutionRestrictions;
  delete taskType.settings.resolutionRestrictions;
  taskType.settings.resolutionRestrictionsForm = ...;
}

const formValues = getFormValuesForTaskType(taskType)
```

philippotto (commented):
Great clean up! Just see my two comments :)

frontend/javascripts/oxalis/model/accessors/flycam_accessor.js
fm3: The way I understand this the code below does that for min and max individually, and essentially results in the default range if both are null. Am I misreading that? In fact we could probably even remove the `!resolutionRestrictions` check here too, as it should now always be present. Do you agree?

philippotto: > In fact we could probably even remove the !resolutionRestrictions check here too, as it should now always be present.

Yes, that was the point which confused me initially :)

> The way I understand this the code below does that for min and max individually, and essentially results in the default range if both are null. 

Yes, that should be correct!

fm3:
Thanks for your feedback! I removed the superfluous check and added the delete statement and comment for the formValues :) Could you have another look?

philippotto (commented):
Awesome! Looks very good :+1: I'll leave the actual "approval" to @youri-k :)

youri-k (approved):
LGTM :shipit:

Issue #3598, opened 2019-01-04
Use object pooling for array buffers to avoid unnecessary GC


philippotto:
Instead of using an object pool for buckets (which would make lifecycle handling of buckets more cumbersome, probably), we could have a pool of array buffers, which is used when setting data for buckets.

The pool could provide a low-level interface such as:

```
requestArrayBuffer(type: "Uint8Array" | ..., length: number): ArrayBuffer
markArrayBufferAsFree(arrayBuffer: ArrayBuffer): void
```

Internally, the pool would either allocate an array buffer when needed or return an unused array buffer. `markArrayBufferAsFree` would be called when we "collect" a bucket (i.e., when we remove the last active reference).
That way, we would avoid unnecessary allocations of 32 kb per bucket. Additionally, we should have a good way of limiting/determining the current memory consumption defined by array buffers, since everything would go through this abstraction.

philippotto:
One random thought:

- When we request the data via `fetch`, I'm not sure whether we can use pre-allocated memory for the response. So, maybe the pooling is only appropriate for the case where we instantiate arraybuffers by ourselves (and this is probably only true for volume annotations)

PR #4924, opened 2020-11-09
Re-enable continuous brush and fix perf + edge-case bugs

### LINKED ISSUE #4859: Re-enable continuous brushing
When I merged the code changes from the continuous brushing feature into the multiresolution pr (https://github.com/scalableminds/webknossos/pull/4755), I might have broken something. At least, the performance was considerably poorer (which I didn't notice on `master`). To reduce the complexity of the already quite big multi-res PR, I disabled that feature.

Let's re-enable it while doublechecking/debugging the performance

### LINKED ISSUE #4877: error when brushing into negative coordinates
if in a volume annotation I brush into negative coordinates, an error alert comes up:
```
Internal error.
Please reload the page to avoid losing data.

TypeError: k[e] is undefined S@main.js:1:380703
drawHorizontalLine2d
``'

### PR DESCRIPTION
**Pull request by philippotto**
I had disabled the continuous brushing feature after I merged it into the multi-res volume branch. One reason was performance and the other was an issue with the scale in higher mags (which was likely produced during the merge).
With this PR the continuous brush feature gets re-enabled. The scale-bug should be fixed now and I also fixed #4877. Additionally, I could greatly improve the performance by ensuring that the allocated buffer only covers the area from "min to max" (instead of from "0 to max"). This caused major problems for me since my dataset's origin wasn't at 0,0,0.

- https://fixcontinuousbrush.webknossos.xyz

### Steps to test:
- use the brush in a very quick manner
- brush in different mags

### Issues:
- fixes #4859 
- fixes #4877

------
- [X] Ready for review

### DIFF
--- a/frontend/javascripts/oxalis/model/volumetracing/volumelayer.js
+++ b/frontend/javascripts/oxalis/model/volumetracing/volumelayer.js
@@ -166,7 +166,7 @@ class VolumeLayer {
     return this.getContourList().length === 0;
   }
 
-  getVoxelBuffer2D(mode: VolumeTool): VoxelBuffer2D {
+  getFillingVoxelBuffer2D(mode: VolumeTool): VoxelBuffer2D {
     if (this.isEmpty() || this.minCoord == null) {
       return VoxelBuffer2D.empty();
     }
@@ -176,6 +176,24 @@ class VolumeLayer {
       return VoxelBuffer2D.empty();
     }
     const maxCoord2d = this.get2DCoordinate(this.maxCoord);
+
+    if (mode === VolumeToolEnum.BRUSH) {
+      // If the brush is used, only perform the "filling" operation
+      // when start- and end coordinate are close enough to each other
+      const contourList = this.getContourList();
+      if (contourList.length < 2) {
+        return VoxelBuffer2D.empty();
+      }
+
+      const startEndDist = V3.length(V3.sub(contourList[0], contourList[contourList.length - 1]));
+
+      const state = Store.getState();
+      const { brushSize } = state.userConfiguration;
+      const radius = Math.round(brushSize / 2);
+      if (startEndDist > 2 * radius) {
+        return VoxelBuffer2D.empty();
+      }
+    }
 
     // The maximum area is scaled by 3 as the min and maxCoord will always be three slices apart,
     // because in `updateArea` a value of 2 is subtracted / added when the values get updated.
@@ -209,9 +227,9 @@ class VolumeLayer {
     // area if you consider narrow shapes.
     // Also, it will be very clear where to start the filling
     // algorithm.
-    this.drawOutlineVoxels((x, y) => setMap(x, y, 0), mode);
+    this.drawOutlineVoxels((x, y) => setMap(x, y, 0));
     this.fillOutsideArea(map, width, height);
-    this.drawOutlineVoxels(setMap, mode);
+    this.drawOutlineVoxels(setMap);
 
     const buffer2D = new VoxelBuffer2D(
       map,
@@ -397,42 +415,15 @@ class VolumeLayer {
     return buffer2D;
   }
 
-  drawOutlineVoxels(setMap: (number, number) => void, mode: VolumeTool): void {
+  drawOutlineVoxels(setMap: (number, number) => void): void {
     const contourList = this.getContourList();
-    const state = Store.getState();
-    const dimIndices = Dimensions.getIndices(this.plane);
-    const [scaleX, scaleY] = this.get2DCoordinate(
-      getBaseVoxelFactors(state.dataset.dataSource.scale),
-    );
-    const { brushSize } = state.userConfiguration;
-    const radius = Math.round(brushSize / 2);
     let p1;
     let p2;
     for (let i = 0; i < contourList.length; i++) {
       p1 = this.get2DCoordinate(contourList[i]);
       p2 = this.get2DCoordinate(contourList[(i + 1) % contourList.length]);
-      if (mode === VolumeToolEnum.TRACE) {
-        Drawing.drawLine2d(p1[0], p1[1], p2[0], p2[1], setMap);
-      } else if (mode === VolumeToolEnum.BRUSH) {
-        // we don't want to connect the last and the first circle with a rectangle
-        if (i !== contourList.length - 1) {
-          // const [xa, ya, xb, yb, xc, yc, xd, yd] = this.getRectangleBetweenCircles(p1, p2, radius, [
-          //   scaleX / this.activeResolution[dimIndices[0]],
-          //   scaleY / this.activeResolution[dimIndices[1]],
-          // ]);
-          // if (this.vector2DistanceWithScale(p1, p2, [scaleX, scaleY]) > 1.5 * radius) {
-          //   Drawing.fillRectangle(xa, ya, xb, yb, xc, yc, xd, yd, setMap);
-          // }
-        }
-        Drawing.fillCircle(
-          p1[0],
-          p1[1],
-          radius,
-          scaleX / this.activeResolution[dimIndices[0]],
-          scaleY / this.activeResolution[dimIndices[1]],
-          setMap,
-        );
-      }
+
+      Drawing.drawLine2d(p1[0], p1[1], p2[0], p2[1], setMap);
     }
   }
 
--- a/frontend/javascripts/oxalis/model/volumetracing/volumelayer.js
+++ b/frontend/javascripts/oxalis/model/volumetracing/volumelayer.js
@@ -20,7 +20,7 @@ import Constants, {
   type VolumeTool,
   Vector2Indicies,
 } from "oxalis/constants";
-import { V3 } from "libs/mjs";
+import { V2, V3 } from "libs/mjs";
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: frontend/javascripts/oxalis/model/volumetracing/volumelayer.js [hunk 1]
```diff
@@ -289,17 +269,14 @@ class VolumeLayer {
     scale: Vector2,
   ): [number, number, number, number, number, number, number, number] {
     const normedPerpendicularVector = this.vector2PerpendicularVector(centre1, centre2);
-    const shiftVector = this.vector2WithScale(
-      this.vector2ScalarMultiplication(normedPerpendicularVector, radius),
-      scale,
-    );
-    const negShiftVector = this.vector2ScalarMultiplication(shiftVector, -1);
+    const shiftVector = V2.scale2(normedPerpendicularVector, V2.scale(scale, radius));
```
philippotto, line 272: This was the scale-related bug (the scale had to be applied to the radius).

frontend/javascripts/oxalis/model/volumetracing/volumelayer.js
```diff
@@ -434,13 +416,13 @@ class VolumeLayer {
...
-          if (this.vector2DistanceWithScale(p1, p2, [scaleX, scaleY]) > 1.5 * radius) {
-            Drawing.fillRectangle(xa, ya, xb, yb, xc, yc, xd, yd, setMap);
-          }
+          // const [xa, ya, xb, yb, xc, yc, xd, yd] = this.getRectangleBetweenCircles(p1, p2, radius, [
+          //   scaleX / this.activeResolution[dimIndices[0]],
+          //   scaleY / this.activeResolution[dimIndices[1]],
+          // ]);
+          // if (this.vector2DistanceWithScale(p1, p2, [scaleX, scaleY]) > 1.5 * radius) {
+          //   Drawing.fillRectangle(xa, ya, xb, yb, xc, yc, xd, yd, setMap);
+          // }
```
philippotto, line 425: @daniel-wer At first I thought this code was wrong, since the rectangles should already have been drawn, but then I read that it's necessary for the implementation of the auto-fill-feature to work (since that code doesn't "know" about the rectangles which were drawn earlier?).
I'm a bit hesitant to keep this code here, as all rectangles are effectively drawn twice then. I wonder whether we should simply remove the auto-fill feature (especially, now that we have the paint-bucket-tool). What do you think? :see_no_evil:

philippotto (commented):
By the way, I also experimented with a `BufferPool` to avoid frequent memory allocations. It looked somewhat like this:

```js
class BufferPool {
  unusedBuffers: Map<number, ArrayBuffer>;

  constructor() {
    this.unusedBuffers = new Map();
  }

  getUint8Array(size: number, fillValue: number = 0): { array: Uint8Array, release: () => void } {
    let buffer;
    // Divide size by 1024 and multiply from the outside with 1024, to make
    // one kiloByte the "smallest unit". Possible buffer sizes will be:
    // 1 kB, 2 kB, 4 kB, 8 kB, 16 kB etc.
    const roundedSize = 1024 * 2 ** Math.ceil(Math.log2(Math.ceil(size / 1024)));
    if (this.unusedBuffers.has(roundedSize)) {
      buffer = this.unusedBuffers.get(roundedSize);
      this.unusedBuffers.delete(roundedSize);
    } else {
      // console.log("Allocating buffer with", roundedSize);
      buffer = new ArrayBuffer(roundedSize);
    }

    const release = this._release.bind(this, buffer, roundedSize);

    return {
      array: new Uint8Array(buffer, 0, size).fill(fillValue),
      release,
    };
  }

  _release(buffer, roundedSize) {
    this.unusedBuffers.set(roundedSize, buffer);
  }
}
```

Unfortunately, I couldn't really measure any performance improvement or see less GC churn :man_shrugging: 

image

frontend/javascripts/oxalis/model/volumetracing/volumelayer.js
daniel-wer: > At first I thought this code was wrong, since the rectangles should already have been drawn, but then I read that it's necessary for the implementation of the auto-fill-feature to work (since that code doesn't "know" about the rectangles which were drawn earlier?).

To explain the status quo: The key here is that only voxels that were drawn with _this_ brush stroke should affect whether an area is auto-filled or not. Therefore, the `contourList` which saves the coordinates of all brush events is used to recreate a map of the brushed voxels (and one cannot simply look up the voxels in the buckets). This map needs to mirror the brushed voxels exactly to allow for correct auto-fill which is why the circles and rectangles need to be redrawn from the `contourList`.

> I'm a bit hesitant to keep this code here, as all rectangles are effectively drawn twice then.

To avoid drawing twice, we could also remember the painted voxels for the current brush stroke (instead of only the coordinates of the brush events) and reuse these here for the auto-fill at the end of the brush stroke. I think you can estimate the best whether this would be easy/feasible with the current code, but I assume it's not that easy as we don't know the size of the map in advance :thinking:

> I wonder whether we should simply remove the auto-fill feature (especially, now that we have the paint-bucket-tool).

I probably wouldn't remove it for now, regardless of whether we change anything in this PR or did you notice performance problems related to the auto-fill?

frontend/javascripts/oxalis/model/volumetracing/volumelayer.js
daniel-wer: Nice find!

frontend/javascripts/oxalis/model/volumetracing/volumelayer.js
```diff
@@ -335,7 +335,7 @@ class VolumeLayer {
     );
     const minCoord2d = [Math.floor(Math.min(xa, xb, xc, xd)), Math.floor(Math.min(ya, yb, yc, yd))];
     const maxCoord2d = [Math.ceil(Math.max(xa, xb, xc, xd)), Math.ceil(Math.max(ya, yb, yc, yd))];
-    const [width, height] = maxCoord2d;
+    const [width, height] = V3.sub(maxCoord2d, minCoord2d);
```
daniel-wer, line 338: Also nice find! :open_mouth:

daniel-wer (approved):
Nice debugging and fixes! LGTM :+1:

daniel-wer:
> By the way, I also experimented with a BufferPool to avoid frequent memory allocations.
Unfortunately, I couldn't really measure any performance improvement or see less GC churn man_shrugging

Hmm, interesting, could be that the browser already handles this efficiently for these kinds of (smaller?) allocations. Thinking about it, I would be somewhat surprised if a "general" buffer pooling strategy would outperform the memory management of the browser. I would expect it to be more performant if we use information that the browser does not have, though. For example when looking at our data buckets which for one layer all have the same size and where we know the maximum number of buffers that should be allocated, the pooling (we already do) probably has better performance than always allocating new buffers.
However, I am no expert in this and it could very well be that we can get rid of some of the gc pressure by re-using buffers cleverly. I would love to do a controlled deep-dive into the whole gc issue in the near future :)

frontend/javascripts/oxalis/model/volumetracing/volumelayer.js
philippotto: Thank you for the summary! As we discussed in person, the current implementation of this PR restricts the auto-fill-feature by only kicking in if the start and end of the drawn "path" are so that the first and last brush circle are touching. The downside is that closing a circle and releasing the mouse somewhere else (e.g., when correcting some "bumps" in the outline or just when being a bit imprecise), won't trigger the auto-fill. However, the UI for that could also be improved by highlighting the start point and the "will perform auto-fill"-state.
In general, I think that the auto-fill-feature is not super crucial and would like to choose the more performant route. In case of negative feedback, we can re-think this :)

> did you notice performance problems related to the auto-fill?

Yes, when brushing a rather large area, there was a noticeable lag for me when releasing the brush :/ This felt especially weird to me, since everything was already brushed (the auto-fill-feature didn't fill anything) and no additional computation was necessary.

philippotto:
> > By the way, I also experimented with a BufferPool to avoid frequent memory allocations.
> > Unfortunately, I couldn't really measure any performance improvement or see less GC churn man_shrugging
> 
> Hmm, interesting, could be that the browser already handles this efficiently for these kinds of (smaller?) allocations. Thinking about it, I would be somewhat surprised if a "general" buffer pooling strategy would outperform the memory management of the browser.

Yeah, that's probably the case. However, I thought that a very dumb pool might already help, since the typical usage pattern is very simple: acquire buffer, release buffer, acquire buffer, release buffer... Acquire and release should be super fast, since the same buffer is simply re-used in 99% of the cases. Therefore, I thought we could "buy" less GC pressure by reducing the memory efficiency (since the pool is very basic and potentially leaves lots of memory unused).

> [...], the pooling (we already do) probably has better performance than always allocating new buffers.

Which pooling are we already doing for bucket buffers?

> I would love to do a controlled deep-dive into the whole gc issue in the near future :)

Definitely, I'm all for that, too :)

daniel-wer:
> Which pooling are we already doing for bucket buffers?

I mistakenly thought we were doing some kind of buffer reusing in the `data_cube.js:addBucketToGarbageCollection` function. However, this function only seems to ensure that a certain maximum number of buckets is never exceeded. I probably mixed this up with your proposal in https://github.com/scalableminds/webknossos/issues/3598 :)

PR #4926, opened 2020-11-09
Fix histogram scale

### LINKED ISSUE #4925: Fix histogram scale
Due to miscalculations, the histogram scale seems to be linear instead of logarithmic.

How to fix:
First, apply the logarithm and then scale the resulting values according to the maximum value and the histogram height.

MichaelBuessemeyer:
This PR fixes the histogram scale. Previously we downscaled the data before applying the logarithm. This leads to a linear scale.
With this fix, the logarithm is applied before scaling the data to the histogram height.

During testing, I found 2 errors in the console produced by the new downscaling modal added by the multi-resolution PR. I just fixed those on the fly.

- https://fixhistogramscale.webknossos.xyz

### Steps to test:
- Open a dataset and look at the histogram. Come this with the master. The small values should now be much higher in the histogram. The details in the data should also be increased for small values.
- As I also fiddled with the modal mentioned above please check whether it looks alright and does not throw errors anymore.
- For that: Open a only tracing that did not get downscaled on your local machine and then open the modal using the orange icon above the segmentation layer settings in the tracing view.

### Issues:
- fixes #4925

------
- [x] Ready for review


MichaelBuessemeyer:
After a discussion, we decided to use a linear scale :chart_with_upwards_trend:

MichaelBuessemeyer:
@daniel-wer I got an exciting update. The scale is now linear and there are now even values on the y-axis :exploding_head: 
image

Could you please check out my changes?

MichaelBuessemeyer:
A quick internal discussion stated that we do not need the y-axis as the values are sampled and as the scale now is linear (which is what a user should expect)

frontend/javascripts/oxalis/view/settings/histogram_view.js
```diff
@@ -133,15 +133,12 @@ class Histogram extends React.PureComponent<HistogramProps, HistogramState> {
...
     ctx.strokeStyle = `rgba(${color.join(",")})`;
-    // Here we normalize all values to the interval of 0 - 9 and then add 1
-    // to gain an interval reaching from 1 - 10, since values between 0 and 1 would be negative, otherwise.
-    const downscalingFactor = 9 / maxValue;
-    const downscaledData = elementCounts.map(
-      value => Math.log10(downscalingFactor * value + 1) * canvasHeight,
-    );
+    // Here we apply the logarithm to all elements and divide by the highest logarithmic value to have an interval from [0,1].
+    // Then we scale this up to the whole height of the histogram canvas.
+    // Note: We add the 1 as otherwise a value of zero would lead to negative infinity and not 0.
```
daniel-wer, line 140:
```suggestion
    // Scale data to the height of the histogram canvas
```

daniel-wer (commented):
Thanks @MichaelBuessemeyer for fixing this and reacting dynamically to the feedback others and I gave you :pray: 

One weird thing I noticed on this branch, but not on master. At first the histogram line is thin, but when moving a histogram handle for the first time, the line becomes thicker all of a sudden and there's also a line at the bottom. See this gif:
histo_width

MichaelBuessemeyer:
> One weird thing I noticed on this branch, but not on master. At first the histogram line is thin, but when moving a histogram handle for the first time, the line becomes thicker all of a sudden and there's also a line at the bottom.

Oh, thanks for finding this bug. This was caused, because the canvas always used the same path. Refreshing was done in the drawYAxis method. I just added using a new path directly to the beginning of the drawHistogram method to fix this :+1:

daniel-wer (approved):
Nice find with the fix for the double line drawing bug :100:

PR #4936, opened 2020-11-11
Clean up and improve how mismatching resolutions are detected

philippotto:
The nightly test fails in one test case due to "mismatching resolutions". However that failure was a false positive, which is caused by the fact that we keep the original resolutions per layer instead of storing the "densified version". The code had problems with a dataset where the resolutions were like this:
- mag 1, mag 2, mag 4, mag 8
- mag 32

The old code densified the resolutions (i.e., padded all the missing resolutions for the second layer) and assumed that the longest resolution list covers all resolutions. The new code constructs a union of all resolutions and checks that there are no mismatches while doing so.

- https://___.webknossos.xyz

### Steps to test:
- see new automated tests
- [x] execute nightly

### Issues:
- follow-up to #4755 

------
- [X] Ready for review

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -260,12 +260,13 @@ jobs:
       #       -X POST
       #       -H "X-Auth-Token: $RELEASE_API_TOKEN"
       #       https://kube.scm.io/hooks/install/webknossos/dev/master?user=CI+%28nightly%29
-      # - run:
-      #     name: Install dependencies and sleep at least 3min
-      #     command: |
-      #       yarn install --frozen-lockfile &
-      #       sleep 180 &
-      #       wait
+      - run:
+          name: Install dependencies and sleep at least 3min
+          command: |
+            yarn install --frozen-lockfile
+      # &
+      # sleep 180 &
+      # wait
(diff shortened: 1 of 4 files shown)


philippotto:
The nightly succeeded :tada: See https://app.circleci.com/pipelines/github/scalableminds/webknossos/1683/workflows/d378dc18-d171-478f-8181-51bb928acff1/jobs/12880

daniel-wer (approved):
Nice fix and it makes sense to me. Impressive how well the longest-list-covers-all-resolutions heuristic hold up :see_no_evil: Thanks for adding the tests and for confirming the fix by running the screenshot tests :tada:

philippotto:
> Impressive how well the longest-list-covers-all-resolutions heuristic hold up see_no_evil

I think (but maybe I'm missing something), the previous logic was quite "precise" (but not elegant :laughing:), since the resolutions were always "densified" (starting at mag 1 until the maximum) in the initialization. Therefore, essentially the range from "mag 1" to "highest mag" was used, which is probably equivalent to the new way (building the union and then densifying it).

Issue #4937, opened 2020-11-11
volume annotation’s largest segment id is never bumped from 0 to 1


fm3:
when opening a new volume annotation (no fallback layer) and brushing some, there is no update action setting the largest segment id to 1, even though there is now data with segment id 1.
this leads to wrong id remapping when importing a volume data zip.
when creating more cells later the update actions setting the largest segment id to 2, 3, etc are sent correctly

daniel-wer:
I found the issue, but am wondering what the best fix would be. It's a combination of three things.

First of all, the frontend always needs an activeCell which is why one is created automatically when creating a new volume tracing. Secondly, this is hidden from the diffing to avoid always having an unsaved state when creating a tracing. Thirdly, the largest segment id is immediately updated when creating a new cell (using "C" or the "Create Cell" button). Strictly speaking, this should only happen once the user annotated at least one voxel using the new cell id.

I feel like the proper fix would be for the frontend to only create a new cell in a new tracing, once the user annotates the first voxel. This would mirror the behavior in skeleton tracings where the first tree is only created once the first node is set. This would be a bit of work as there are quite a few spots which assume there always is an activeCellId.

Alternatively (or in addition) the largest segment id could be updated only once a voxel using that id was annotated. This would lead to cases where the activeCellId is larger than the largestSegmentId, though, not sure whether that is problematic/unintuitive.

Any opinions? (/cc @philippotto)

philippotto:
> [...] the largest segment id could be updated only once a voxel using that id was annotated. This would lead to cases where the activeCellId is larger than the largestSegmentId, though, not sure whether that is problematic/unintuitive.

This sounds like a fair approach to me. In general, the user is probably not aware of the largest segment id, anyway, since we only use that to be able to create a fresh id. The only downside I see is that hitting "create cell" multiple times will only bump the id once (since the user didn't label anything and thus didn't bump the id). That's a bit counter-intuitive probably. Two possible solutions come to my mind for this, though:

- Disable the button if activeCellId > largestSegmentId. A tooltip could explain that the id wasn't used yet. However, this also feels quite weird to me. First, because one can still set the ID arbitrarily in the settings. Second, hitting C to cycle through the available "colors" (as I sometimes do :D) wouldn't be possible. Three, we already have a lot of tooltips :laughing: 
- Create cell generates `Math.max(activeCellId, largestSegmentId) + 1`.

[...]

PR #4915, opened 2020-11-05
Fix uploading multiple volume tracings

youri-k:
### URL of deployed dev instance (used for testing):
- https://___.webknossos.xyz

### Steps to test:
- upload zip with multiple similar named data zips => should not fail

### Issues:
- fixes #4903 

------
- [x] Updated [(unreleased) changelog](../blob/master/CHANGELOG.unreleased.md#unreleased)
- ~~[ ] Updated [(unreleased) migration guide](../blob/master/MIGRATIONS.unreleased.md#unreleased) if applicable~~
- ~~[ ] Updated [documentation](../blob/master/docs) if applicable~~
- ~~[ ] Adapted [wk-connect](https://github.com/scalableminds/webknossos-connect) if datastore API changes~~
- ~~[ ] Needs datastore update after deployment~~
- [x] Ready for review

PR #4938, opened 2020-11-11
Refactor Volume Annotation Merging

### LINKED ISSUE #4916: Refactor MergedVolume usages
I feel like the usages of MergedVolume can be written a lot DRYer with some convenience functions for reading entire volumes from zip or from the db. Also, MergedVolume should probably be moved to its own scala file.

fm3:
Adds some convenience methods to `MergedVolume` to make its usages DRYer and easier to read. Also don’t silently ignore zip unpacking Failures anymore.

- https://refactorvolumemerging.webknossos.xyz

### Steps to test:
- open volume annotation, brush some, drag in other volume zip, should be merged correctly (be wary of #4937)
- upload multiple volume zips, should be merged correctly (blocked by #4915 )
- create, brush and finish volume tasks, click view on volume project, should merge data correctly in CompoundProject

### Issues:
- fixes #4916 

------
- [x] Ready for review


webknossos-tracingstore/app/com/scalableminds/webknossos/tracingstore/tracings/volume/MergedVolume.scala
```diff
@@ -0,0 +1,135 @@
...
+import scala.concurrent.ExecutionContext
+
+class MergedVolume(elementClass: ElementClass, initialLargestSegmentId: Long = 0)
+    extends DataConverter
+    with ByteUtils
+    with VolumeDataZipHelper
+    with ProtoGeometryImplicits {
+  private var mergedVolume = mutable.HashMap.empty[BucketPosition, Array[UnsignedInteger]]
+  private var labelSets = mutable.ListBuffer[mutable.Set[UnsignedInteger]]()
+  private var labelMaps = mutable.ListBuffer[mutable.HashMap[UnsignedInteger, UnsignedInteger]]()
```
youri-k, line 23: If we're using mutable types here, we can declare them as `val`s to prevent unwanted new assignments

webknossos-tracingstore/app/com/scalableminds/webknossos/tracingstore/tracings/volume/MergedVolume.scala
```diff
@@ -0,0 +1,135 @@
...
+  private def prepareLabelMaps(): Unit =
+    if (labelSets.isEmpty || (labelSets.length == 1 && initialLargestSegmentId == 0) || labelMaps.nonEmpty) {
+      ()
+    } else {
+      var i: UnsignedInteger = UnsignedInteger.zeroFromElementClass(elementClass)
+      if (initialLargestSegmentId > 0) {
+        labelMaps += mutable.HashMap.empty[UnsignedInteger, UnsignedInteger]
+        i = UnsignedInteger.fromLongWithElementClass(initialLargestSegmentId, elementClass)
+      }
+      labelSets.toList.foreach { labelSet =>
```
youri-k, line 65: The `toList` call is unnecessary in my opinion here because we don't rely on a return type here

webknossos-tracingstore/app/com/scalableminds/webknossos/tracingstore/tracings/volume/MergedVolume.scala
```diff
@@ -0,0 +1,135 @@
...
+        }
+    }
+
+  def addFromDataZip(sourceVolumeIndex: Int, zipFile: File)(implicit ec: ExecutionContext): Box[Unit] =
+    withBucketsFromZip(zipFile) { (bucketPosition, bytes) =>
+      add(sourceVolumeIndex, bucketPosition, bytes)
+    }
+
+  def add(sourceVolumeIndex: Int, bucketPosition: BucketPosition, data: Array[Byte]): Unit =
+    if (data.length > 1) { // skip reverted buckets
```
youri-k, line 92: One question: do we need to keep this for backwards compatibility with old downloaded tracings because the iterators no longer return reverted buckets. Maybe we can put this check into the `addFromDataZip` call then.

webknossos-tracingstore/app/com/scalableminds/webknossos/tracingstore/tracings/volume/VolumeDataZipHelper.scala
```diff
@@ -0,0 +1,73 @@
...
+                    block(bucketPosition, data)
+                  }
+                }
+              }
+            }
+        }
+    }
+    for {
+      _ <- unzipResult
+    } yield ()
```
youri-k, line 39: I find it hard to understand what this does. Is it the same as `unzipResult.map(_ => ())`? Why do we want to change the return type? Is it meant to deter someone from using the return value?

webknossos-tracingstore/app/com/scalableminds/webknossos/tracingstore/tracings/volume/VolumeTracingService.scala
```diff
@@ -175,161 +176,30 @@ class VolumeTracingService @Inject()(
   }
 
   def initializeWithDataMultiple(tracingId: String, tracing: VolumeTracing, initialData: File): Fox[Set[Point3D]] = {
-    if (tracing.version != 0L) {
-      return Failure("Tracing has already been edited.")
-    }
+    if (tracing.version != 0L) return Failure("Tracing has already been edited.")
```
youri-k, line 179: That looks so much better. Thank you :tada:

webknossos-tracingstore/app/com/scalableminds/webknossos/tracingstore/tracings/volume/VolumeTracingService.scala
```diff
@@ -343,36 +213,20 @@ class VolumeTracingService @Inject()(
...
+        Fox.successful(())
+      } else {
+        savedResolutions.add(bucketPosition.resolution)
+        saveBucket(dataLayer, bucketPosition, bytes, tracing.version)
+      }
     }
 
     for {
       _ <- unzipResult
     } yield savedResolutions.toSet
```
youri-k, line 227: I would prefer a simple  `.map()` when only having one statement in a `for ... yield`, but that's personal preference I guess.

Maybe we should discuss in person, also for the future

youri-k (changes requested):
Wow, that is a huge improvement :tada: . Feels like the Atacama because it's sooo DRY :desert:

My comments are only about some smaller improvements :slightly_smiling_face:

webknossos-tracingstore/app/com/scalableminds/webknossos/tracingstore/tracings/volume/MergedVolume.scala
fm3: true enough :) and the data zip unpacking method drops all-zero buckets already also :) I’ll remove it here.

webknossos-tracingstore/app/com/scalableminds/webknossos/tracingstore/tracings/volume/VolumeDataZipHelper.scala
fm3: `map` is fine with me :) It does not make sense to use the return value, so I wanted to hide it in the signature, yes.

webknossos-tracingstore/app/com/scalableminds/webknossos/tracingstore/tracings/volume/VolumeTracingService.scala
fm3: I agree, map is better. I had originally written `unzipResult.map(savedResolutions.toSet)` (forgetting the `_ =>`) and had changed it to the for comprehension when that didn’t compile. I changed it to the correct map format now :)

fm3:
Thanks for your feedback! I implemented the suggestions, feel free to have another look.

youri-k (approved):
LGTM :slightly_smiling_face:

PR #4940, opened 2020-11-11
Abort bucket pulls if a layer is reloaded

### LINKED ISSUE #4895: Error: Unexpected state: UNREQUESTED
Airbrake shows several occurrences of `Error: Unexpected state: UNREQUESTED`. I think, this can happen if a bucket is fetched and in the mean-time canceled, but the response is still processed.

I think, the error itself doesn't have a high impact, apart from being annoying and confusing. A simple try-catch around it would probably work, but I'd vote for refactoring the corresponding logic to make this error not happen in the first place.

https://scm.airbrake.io/projects/95438/groups/2737723584692415595?tab=notice-detail

daniel-wer:
When a layer is reloaded, all buckets are "collected" which means they are deleted and reset. However, bucket requests that were still in progress were not aborted, before, which lead to `Error: Unexpected state: UNREQUESTED` errors (and potentially caused the rendering of old data if the actual data on the hard disk really changed).

Initially, I focused on aborting the actual fetch requests. However, this is fairly complex as AbortSignals cannot be transferred to web workers and in the end it didn't prevent the "UNREQUESTED" errors. There's a lot more that happens, besides the `fetch` call, in the time from when the pullqueue triggers the bucket fetching procedure until the actual array buffer is saved in the bucket. Therefore, I opted for a more general approach, allowing to abort the bucket fetching procedure at any time. I used the same `AbortController` and `AbortSignal` concept for that as it is fairly elegant. In theory, the signal could be forwarded to the actual fetch call so that it can be aborted in-flight as well, however, I'm not sure whether this limited time save is worth the substantial additional complexity.
In my tests, the layer reloading was very quick with the current approach.

- https://___.webknossos.xyz

### Steps to test:
- Move around and (while buckets are still loading) reload a layer using the button in the layer settings. Alternatively, reload a layer twice in quick succession.
- Open the version view and quickly switch between different volume versions.
- No errors should be shown in the console.

### Issues:
- fixes #4895 

------
- [x] Ready for review


frontend/javascripts/test/model/binary/pullqueue.spec.js
```diff
@@ -107,7 +107,9 @@ test.serial("Successful pulling: should receive the correct data", t => {
 
 function prepare() {
   WkstoreAdapterMock.requestWithFallback = sinon.stub();
-  WkstoreAdapterMock.requestWithFallback.onFirstCall().returns(Promise.reject());
+  WkstoreAdapterMock.requestWithFallback
+    .onFirstCall()
+    .returns(Promise.reject(new Error("Expected promise rejection in tests. Can be ignored.")));
```
daniel-wer, line 112: I spent ages hunting down this promise rejection which let the test fail (for a good reason!, I found a bug that way), but because no rejection reason was given, there was no stack trace or anything, so I didn't find it. I even updated ava to the newest version in the process because I thought I would get more information :see_no_evil: In the end, I found it through the debugger and we use the newest ava version now, so there's that :)

frontend/javascripts/oxalis/model/bucket_data_handling/pullqueue.js
```diff
@@ -84,8 +94,19 @@ class PullQueue {
 
     const { renderMissingDataBlack } = Store.getState().datasetConfiguration;
 
+    const { signal } = this.abortController;
     try {
-      const bucketBuffers = await requestWithFallback(layerInfo, batch);
+      const bucketBuffers = await new Promise(async (resolve, reject) => {
+        const abort = () => reject(PULL_ABORTION_ERROR);
+        signal.addEventListener("abort", abort);
+        try {
```
daniel-wer, line 102: This try/catch inside a promise inside a try/catch construction is sub-optimal to say the least, but it's the only way I got it to work. The inner try/catch is needed to avoid unhandled promise rejections. The promise is needed to be able to trigger an error from the event listener that is catched by the outer try/catch. Please let me know if you see a better solution :see_no_evil:

philippotto: Since there is an inner and outer promise, it makes sense to me that there has to be an inner and outer try-catch. So, I'm assuming this cannot easily be avoided. However, maybe the abortable-logic can be extracted into a helper? A usage like this might be more concise (and reusable):

```js
try {
  const bucketBuffers = await asAbortable(
    requestWithFallback(layerInfo, batch),
    this.abortController.signal,
    PULL_ABORTION_ERROR
  );
  ...
} catch (...) {
  ...
}
```

with `asAbortable` looking somewhat like this:

```js
function asAbortable(promise, signal, error) {
  return new Promise(async (resolve, reject) => {
        const abort = () => reject(error);
        signal.addEventListener("abort", abort);
        try {
          const value = await promise;
          resolve(value);
        } catch (error) {
          reject(error)
        }
        signal.removeEventListener("abort", abort);
}
```

philippotto (approved):
Awesome job! Really cool that you found the actual culprit for this error :+1: I only left a small refactoring idea as comment.

One last question: Should we set the bucket state to `UNREQUESTED` for the buckets for which we abort the requests? In the specific use case, it doesn't matter, as the buckets are collected anyway. However, if we use the abort method later in some other use case, this might cause trouble :thinking: 
No strong opinion, though. Probably alright as it is right now :)

philippotto:
> However, if we use the abort method later in some other use case, this might cause trouble thinking

Another use case just came to my mind: disabling a visible layer could immediately abort all bucket requests for that layer. However, since the abort is not "real" (as in: the requests still go through), there is probably not much use in that.

daniel-wer:
> Another use case just came to my mind: disabling a visible layer could immediately abort all bucket requests for that layer. However, since the abort is not "real" (as in: the requests still go through), there is probably not much use in that.

I just checked this: After disabling a layer, no new requests for that layer are issued, so yes there won't be any real use in using the abortion functionality for that.

daniel-wer:
@philippotto Thanks a lot for your feedback, the `asAbortable` functionality was a great candidate to extract as a function :+1: 

> One last question: Should we set the bucket state to UNREQUESTED for the buckets for which we abort the requests? In the specific use case, it doesn't matter, as the buckets are collected anyway. However, if we use the abort method later in some other use case, this might cause trouble thinking

Absolutely, I simply had to remove the early return in the pullqueue error handler which takes care of exactly that :+1:

philippotto (approved):
Excellent :)

PR #4944, opened 2020-11-13
Improve Dataset Settings to not adapt suggestion automatically

### LINKED ISSUE #4672: separate saving dataset sharing from updating properties
a user could not share a dataset with a team because wk guessed invalid json properties. this should not block the other tabs of the edit view

MichaelBuessemeyer:
This PR changes how wk handles the suggestions of settings in the datasource-properties.json that could be inferred.
In short, it does not automatically apply those suggestions but offers the user to adapt them. It makes it easier for the user if properties in the datasource-properties.json are missing or invalid.

Before you are reading all this. I myself had problems with keeping an overview of all cases and how they should/could be tested and how they should behave. That's why I wrote all this.

There are several cases that might happen. From now on datasource-properties.json will be names datasource.json:
- The datasource.json on the server is correct and the server has no additional things to infer.
   - In that case, nothing needs to be done :).
- The datasource.json on the server is correct and the server has additional things to infer.
   - In that case, the user should be offered to accept these suggestions. Previously, they were adapted directly which might have caused errors that prevented a person from saving other changes done.
- The datasource.json on the server is a correct JSON but has some missing or invalid properties that violate the JSON-schema.
   - In that case, the form will be filled with the invalid JSON and an error should show the user, what in invalid.
   - In case the user does not change the datasource settings, not update will be sent to the backend. This prevents the update request to fail because the datasource settings are still invalid.
- The datasource.json on the server is in an invalid JSON format and therefore cannot be parsed. 
  - If thats the case, then the datasource settings will be filled with the suggested changes by the backend.
  - TODO: Discuss which option to pick:
     - option 1 (Currently the case): Saving the settings is prevented even when the user does not change the datasource settings. Therefore the user is forced to enter valid datasource settings to fix the datasource.json in on the server. 
    - option 2: Saving is not prevented when the datasource settings weren't changed. This makes it possible to e.g. set "Sharing & Permissions"  without having to fix the datasource.
   -> I think we should go for option 1, as ([afaik](https://www.macmillandictionary.com/dictionary/british/afaik)) a dataset with an invalid datasource.json cannot be used anyway.

I hope this covers all cases.

- https://___.webknossos.xyz

### Steps to test:
- Choose one dataset, lets call it A.
- Testing should be done locally as there you can easily change the datasource-properties.json (datasource.json).
- First: Create a backup of the datasource-properties.json of dataset A to recover it after testing. It is easier than using wk's backup file.

- Open the edit view of dataset A. If it shows a banner that suggestions could be inferred, do the next testing case first.
- If it shows nothing and changing "Sharing & Permissions" works everything is fine :heavy_check_mark: 

- To prepare this test case, open the datasource.json of dataset A and remove a resolution from one layer.
- Then go to the dashboard and reload that dataset.
- Then open the edit view. The data tab should now have a blue banner on top that suggests to merge additional findings by the backend into the current settings.
   - First, do not change anything in the datasource settings and change something in one of the other tabs. Saving should work and by reloading/reopening the edit view, the suggested settings should not have been applied and the blue banner should still be there.
  - Next, try to accept the suggested settings and then save the settings. Then reload/reopen the edit view. The blue banner should be gone and the suggested settings should be applied.

- Open the datasource.json of dataset A, set the largestSegmentId to -1 and remove a resolution again.
- Then reload by using the reload button in the dashboard.
- Next, open the edit view of dataset A. The datasource settings should show the settings correct settings except that the largestSegmentId now is set to -1 and marked as invalid. Additionally, a blue banner should be displayed that shows that the datasource settings have invalid properties. It should have the option of accepting the suggested settings by webknossos. But these suggestions are not helpful in this case, as wk cannot guess the largestSegmentId.
  - First, do the same thing as in the previous test: Try editing something in another tab and the save and do not edit the datasource settings. The saving should work out and be persisted when reloaded.
  - Second, fix the invalid setting by setting the largestSegment to e.g. 100, accept the suggested settings by the backend and then save those settings. This should work fine and the updated settings should be persisted.
  - Next, set the largestSegmentId again to -1 in the datasource.json and reload the settings from the dashboard (Leave the resolutions as they are). 
    The dataset settings should only show a blue banner telling the user to repair the broken settings but the banner does not suggest any improvements from the backend that could be applied.
   Then fix the largestSegmentId again and save the changes. This should work without any error and relaoding the edit page should not show any blue banner anymore.

- To prepare that case simply remove one comma from the datasource.json of dataset A where a comma is supposed to be. This will break the parsing.
- Reload the dataset from the dashboard to ensure that the datasource-properties.json is up-to-date.
- Then open the edit view. The banner display should be yellow/orange and a warning. The loaded datasource setting should now be the settings that were suggested by the backend. But the backend cannot guess settings like bounding boxes or voxel size. Those settings should be set to "0,0,0,0,0,0" / "0,0,0".
- Try to change something in another tab and then save. Saving should not work and the data tab should open with the incorrect settings.
- Then fix those settings and save them. This should now work and be persisted.

That's it. These were all cases I could think off :)

### Issues:
- fixes #4672

------
- [x] Ready for review


frontend/javascripts/dashboard/dataset/dataset_import_view.js
```diff
@@ -182,44 +215,106 @@ class DatasetImportView extends React.PureComponent<Props, State> {
...
+      );
+      type = "warning";
+    } else if (!isDatasourceJSONValid(savedDataSourceOnServer)) {
+      // TODO: why it this never the case? Shall this case be removed?
+      message = (
+        <div>
+          The current datasource-properties.json on the server seems to be invalid. Thus the
+          datasource-properties.json settings below were suggested by webKnossos. The current
+          content of the datasource-properties.json can be viewed{" "}
+          <a
```
MichaelBuessemeyer, line 239: I got a question regarding this case. @fm3 I think you can answer my question here.

This case happens/should happen when the datasource-properties.json is in a valid JSON Format but the JSON does not match the datasource-properties.json schema. One reason might be that a properties is missing or so.
To reproduce the intended error to catch with this case, simply remove the "largestSegmentId" from the JSON.

While working at this PR, I noticed that the backend also validated the format and if that fails, simply sends an error and not the invalid JSON :/
The message the backend sends to the frontend when the largestSegmentId property is missing is:
```
status: "Error: Invalid json format in binaryData/sample_organization/e2006_knossos/datasource-properties.json: Failure(Invalid json: List((/dataLayers(1)/dataFormat/category/largestSegmentId,List(JsonValidationError(List(error.path.missing),WrappedArray())))),Empty,Empty)"
```

@fm3 Is it possible that the backend still sends the invalid JSON? This would be helpful as then the only invalid option is largestSegmentId and the rest of the set options could be used to fill the settings view. This way the user does only need to set the largestSegmentId and not the other fields like the bounding boxes or so.

fm3: I would have expected that the values »suggested by webknossos« as sent by the backend do contain the values of those fields that were valid in the original? I think some merging goes on there in the backend already.

fm3: if that is not the case or that does not support the optimal workflow, sure, we can change the backend behavior, do you think that would be beneficial here?

MichaelBuessemeyer: I just check this and the backend does not merge the bounding boxes, scale, voxel size and largest segment id. 

The use case would be that repairing a broken datasource-properties.json would be easier as only the missing / broken attribute has to be adjusted and not very of the attribute listed above.

I think what is against this case is whether this really happens or if a broken datasource-properties.json is very rare. 
I think: if it is rare then it should not be necessary to add this to the backend but if this happens from time to time I think that would make fixing the JSON file easier.

youri-k: @MichaelBuessemeyer and I just had a discussion. We agreed on the following approach:
If the backend can not validate the `datasource_properties.json`, it creates an `UnusableDatasource` as before, but checks if the `datasource_properties.json` is valid json. If it is valid json, it is saved alongside the current status.
When exploring an `UnusableDatasource`, the `previousDatasource` response field is set to the valid, but unusable json, or to the normal `UnusableDatasource` if the json is invalid.

MichaelBuessemeyer: Thanks, @youri-k :smile: 
@fm3 just for your notice:
Youri's changes made case three of the listed ones above possible. Without those changes, case three would be exactly handled as case 4.

frontend/javascripts/dashboard/dataset/dataset_import_view.js
```diff
@@ -86,8 +89,10 @@ class DatasetImportView extends React.PureComponent<Props, State> {
     messages: [],
     activeDataSourceEditMode: "simple",
     activeTabKey: "data",
+    savedDataSourceOnServer: null,
+    inferredDataSource: null,
     differenceBetweenDatasources: null,
-    previousDataSource: null,
+    didApplyDifferenceBetweenDatasources: false,
```
daniel-wer, line 95: Could you unify the spelling of dataSource in this file as best as possible, so that the S is uppercase?

frontend/javascripts/dashboard/dataset/dataset_import_view.js
```diff
@@ -97,28 +102,49 @@ class DatasetImportView extends React.PureComponent<Props, State> {
...
+        const didParsingTheSavedDataSourceJSONSucceed =
+          savedDataSourceOnServer != null && savedDataSourceOnServer.status == null;
+        if (didParsingTheSavedDataSourceJSONSucceed) {
+          dataSource = savedDataSourceOnServer;
+        } else {
+          // If the current datasource json is invalid, the inferred version should be used automatically.
+          dataSource = inferredDataSource;
+          this.setState({ didApplyDifferenceBetweenDatasources: true });
+        }
+        this.setState({ savedDataSourceOnServer, inferredDataSource });
```
daniel-wer, line 128: To get rid of those let variables, could you save the `dataSourceMessages` to the state here instead of in line 180? Then you should be able to get rid of the let variable declarations in lines 107, 112, 113 and use `const = ({` in line 114.
Please ignore if this doesn't make sense.

frontend/javascripts/dashboard/dataset/dataset_import_view.js
```diff
@@ -97,28 +102,49 @@ class DatasetImportView extends React.PureComponent<Props, State> {
...
+            didApplyDifferenceBetweenDatasources: areObjectsEqual,
+          });
         }
       }
       if (dataSource == null) {
         throw new Error("No datasource received from server.");
       }
+      if (dataset.dataSource.status != null && dataset.dataSource.status.includes("Error")) {
+        // If the datasource-properties.json could not be parsed due to schema errors,
+        // we replace it with version that is at least parsable.
```
daniel-wer, line 143:
```suggestion
        // we replace it with the version that is at least parsable.
```

frontend/javascripts/dashboard/dataset/dataset_import_view.js
```diff
@@ -162,10 +188,21 @@ class DatasetImportView extends React.PureComponent<Props, State> {
...
   getDatasourceDiffAlert() {
+    // Only show if the option did not apply
+    const {
+      savedDataSourceOnServer,
+      differenceBetweenDatasources,
+      didApplyDifferenceBetweenDatasources,
+      inferredDataSource,
+    } = this.state;
+    const didUserAppliedSuggestions =
+      didApplyDifferenceBetweenDatasources === true &&
```
daniel-wer, line 199:
```suggestion
      didApplyDifferenceBetweenDatasources &&
```

frontend/javascripts/dashboard/dataset/dataset_import_view.js
```diff
@@ -162,10 +188,21 @@ class DatasetImportView extends React.PureComponent<Props, State> {
...
+      isDatasourceJSONValid(savedDataSourceOnServer);
```
daniel-wer, line 200: In which case would `didApplyDifferenceBetweenDatasources` be true but `isDatasourceJSONValid(savedDataSourceOnServer)` be false? In other words, I don't understand why the second condition is needed, yet.

frontend/javascripts/dashboard/dataset/dataset_import_view.js
```diff
@@ -182,44 +219,104 @@ class DatasetImportView extends React.PureComponent<Props, State> {
...
+    } else if (!isDatasourceJSONValid(savedDataSourceOnServer)) {
+      // If the datasource-properties.json on the server has invalid or missing properties but is a valid JSON.
+      message = (
+        <div>
+          The current datasource-properties.json on the server seems to have invalid or missing
+          properties. <br />
+          <a
+            href="#"
+            onClick={() =>
+              showJSONModal("Current and invalid datasource-properties.json", inferredDataSource)
```
daniel-wer, line 246: Doesn't this only show the inferred data source? The description says "Current and invalid datasource-properties.json", though.

frontend/javascripts/dashboard/dataset/dataset_import_view.js
```diff
@@ -162,10 +188,21 @@ class DatasetImportView extends React.PureComponent<Props, State> {
...
+      didApplyDifferenceBetweenDatasources === true &&
+      isDatasourceJSONValid(savedDataSourceOnServer);
     if (
-      this.state.previousDataSource == null ||
-      this.state.previousDataSource.status === notImportedYetStatus ||
-      _.size(this.state.differenceBetweenDatasources) === 0
+      savedDataSourceOnServer == null ||
+      savedDataSourceOnServer.status === notImportedYetStatus ||
+      _.size(differenceBetweenDatasources) === 0 ||
+      didUserAppliedSuggestions
```
daniel-wer, line 205: Shouldn't the diff always be zero when `didUserAppliedSuggestions == true`? Do we need the check for `didUserAppliedSuggestions`, then?

daniel-wer (commented):
@MichaelBuessemeyer Thanks a lot for documenting the different cases in such detail. This will make it much easier to test this PR and also helped me to understand the code! For now, I've only reviewed the code and added minor comments. I'll test in the next days and will report back.

frontend/javascripts/dashboard/dataset/dataset_import_view.js
MichaelBuessemeyer: Oh thx you just found a bug in the logic. :bug: 

As I went over this again and not having worked with this PR for more than a week I came up the the idea of better handling the "state of the datasource"-setting. I just implemented this and this rework also affects the logic.
This means that you have to take a look at the whole logic again :see_no_evil: 
But hopefully this change makes the logic more expressive / simple and therefore easier to understand.

Please tell me what you think about this and if you have suggestions, that could improve the understandability even more :smile:

frontend/javascripts/dashboard/dataset/dataset_import_view.js
```diff
@@ -65,8 +70,8 @@ type State = {
   activeTabKey: TabKey,
   savedDataSourceOnServer: ?APIDataSource,
   inferredDataSource: ?APIDataSource,
-  differenceBetweenDatasources: ?Object,
-  didApplyDifferenceBetweenDatasources: boolean,
+  differenceBetweenDataSources: ?Object,
```
daniel-wer, line 73:
```suggestion
  differenceBetweenDataSources: Object,
```

At least from what I can see.

frontend/javascripts/dashboard/dataset/dataset_import_view.js
```diff
@@ -219,10 +233,20 @@ class DatasetImportView extends React.PureComponent<Props, State> {
...
     let type = "info";
-    const didParsingTheSavedDataSourceJSONSucceed = savedDataSourceOnServer.status == null;
-    if (!didParsingTheSavedDataSourceJSONSucceed) {
+    const applySuggestedSettings = () => {
+      this.props.form.setFieldsValue({
+        dataSourceJson: jsonStringify(inferredDataSource),
+      });
+      this.setState(currentState => {
+        const updatedStatus = _.cloneDeep(currentState.dataSourceSettingsStatus);
+        updatedStatus.appliedSuggestions = "YES";
```
daniel-wer, line 244:
```suggestion
        const updatedStatus = {
          ...currentState.dataSourceSettingsStatus, appliedSuggestions: "YES"
        };
```

No need to get out the `_.cloneDeep` hammer I think :)

frontend/javascripts/dashboard/dataset/dataset_import_view.js
```diff
@@ -56,6 +56,11 @@ type Props = {| ...OwnProps, ...StateProps |};
 
 type TabKey = "data" | "general" | "defaultConfig";
 
+type DataSourceSettingsStatus = {
```
daniel-wer, line 59: Could you add a small comment indicating what "NONE" means? It becomes apparent that it means "The server has no suggestions." when reading the code, but it's not really clear here in my opinion.

frontend/javascripts/dashboard/dataset/dataset_import_view.js
```diff
@@ -56,6 +56,11 @@ type Props = {| ...OwnProps, ...StateProps |};
...
+  appliedSuggestions: "YES" | "NO" | "NONE",
+  isJSONFormatValid: "YES" | "NO" | "BROKEN_FORMAT",
```
daniel-wer, line 61: The name of this variable (`isJSONFormatValid`) in combination with the enum values is not really clear in my opinion. `jsonFormat` is ambiguous, it could either mean the "inner structure" of the json (keys, values, types) which the first two values of the enum refer to I assume or it could mean the json specification itself which the last enum value refers to.
Maybe it's more understandable if you rename the last enum value to "BROKEN_JSON" to avoid the duplicate reference of "format" - what do you think?

daniel-wer (commented):
Great @MichaelBuessemeyer, I agree that using the enums the code became easier to understand :+1: 

I added some small suggestions. You could also think about using dict enums (see constants.js) or Enumjs (see api_flow_types.js) which would make it easier to rename enum values later and would help to avoid confusion between the two different enum types.

I still haven't tested this, but will hopefully manage tomorrow!

frontend/javascripts/dashboard/dataset/dataset_import_view.js
MichaelBuessemeyer: > Maybe it's more understandable if you rename the last enum value to "BROKEN_JSON" to avoid the duplicate reference of "format" - what do you think?

That sounds like a good suggestion to me :D

MichaelBuessemeyer:
> I added some small suggestions. You could also think about using dict enums (see constants.js) or Enumjs (see api_flow_types.js) which would make it easier to rename enum values later and would help to avoid confusion between the two different enum types.


I already had this in mind. I tried to give each possible state an expressive name but this lead to more confusing names than the current option uses (in my opinion).
Nevertheless here are the states I came up with. Can you think of a way / names that express each state better?
```
const DataSourceSettingsStatusEnum = {
 jsonIsParseableAndFormatIsCorrectAndAppliedSuggestions: "JSON_IS_PARSEABLE_AND_FORMAT_IS_CORRECT_AND_APPLIED_SUGGESTIONS"
  jsonIsParseableAndFormatISCorrectNoSuggestionsAvailable: "JSON_IS_PARSEABLE_ANDFORMAT_IS_CORRECT_AND_APPLIED_SUGGESTIONS"
...
}
```
I think this shows why I did not implement it this way.


But I think a mixture of both is the best: I will give each "substate" of `DataSourceSettingsStatus` an enum. :+1:

frontend/javascripts/dashboard/dataset/dataset_import_view.js
MichaelBuessemeyer: I renamed it to `NoAvailableSuggestions` :+1:

fm3 (commented):
Backend LGTM

daniel-wer:
> But I think a mixture of both is the best: I will give each "substate" of DataSourceSettingsStatus an enum.

Nice, that is what I had in mind :+1:

frontend/javascripts/dashboard/dataset/dataset_import_view.js
```diff
@@ -97,28 +122,59 @@ class DatasetImportView extends React.PureComponent<Props, State> {
...
+        });
       }
       if (dataSource == null) {
         throw new Error("No datasource received from server.");
       }
+      if (dataset.dataSource.status != null && dataset.dataSource.status.includes("Error")) {
+        // If the datasource-properties.json could not be parsed due to schema errors,
+        // we replace it with the version that is at least parsable.
+        const datasetClone = (_.cloneDeep(dataset): any);
+        ("YES");
```
daniel-wer, line 174:
```suggestion
```

frontend/javascripts/dashboard/dataset/dataset_import_view.js
```diff
@@ -55,15 +56,34 @@ type Props = {| ...OwnProps, ...StateProps |};
...
+const AppliedSuggestionsEnum = {
+  Yes: "Yes",
+  No: "No",
+  NoAvailableSuggestions: "NoAvailableSuggestions",
+};
+
+const IsJSONFormatValidEnum = {
+  Yes: "Yes",
+  No: "No",
+  BrokenFormat: "BrokenFormat",
```
daniel-wer, line 68: I would still opt to rename this to BrokenJson

daniel-wer (commented):
I just did the testing which was an extremely pleasant procedure with your detailed instructions and descriptions of the different cases - Thanks a lot!

Everything worked as described and I like that the current situation is always accurately described in the banner, so that the user understands what to do next.

I did not really understand in which case I was redirected to the dashboard after saving and when not - seemed pretty random.

See my small suggestions and then this should be good to go :tada:

youri-k:
@MichaelBuessemeyer I just merged the master and noticed 2 things:
- [ ] Since #4959, DataSets with no layers are considered invalid, but the bar at the top shows that this config is valid
- [ ] When importing a DataSet, which has a previousDataSource, the scale of the previousDataSource gets overwritten with (0,0,0) and the `defaultViewConfiguration` from the previousDataSource is deleted

frontend/javascripts/dashboard/dataset/dataset_import_view.js
MichaelBuessemeyer: Sorry I mix this up when I added the enums :see_no_evil:

MichaelBuessemeyer:
> @MichaelBuessemeyer I just merged the master and noticed 2 things:
- [ ] Since #4959, DataSets with no layers are considered invalid, but the bar at the top shows that this config is valid
- [ ] When importing a DataSet, which has a previousDataSource, the scale of the previousDataSource gets overwritten with (0,0,0) and the `defaultViewConfiguration` from the previousDataSource is deleted

Thanks for telling me @youri-k. 
- The first point was pretty easy to fix by adding a minLength to the datalayer array in the schema validation.
- About your second point: I do not really get what you mean. This PR is only about the datasource-properties.json and the default view configurations are not saved in there (at least as far as I know).
How can I fix this issue conceptually in your opinion?

youri-k:
Now that we send the most correct version of the datasource-properties.json in the `previousDataSource` field, we lose the `status` field of the DataSet inside the `ImportView`. Do you think this is a problem?

MichaelBuessemeyer:
@daniel-wer I fixed a few things youri mentioned.
Now the error message the backend generates when trying to parse an invalid datasource-properties.json file is displayed to the user.

MichaelBuessemeyer:
@youri-k I just added the `defaultViewConfiguration`to the datasource-properties.json schema in the frontend. You also mentioned something like `defaultLayerViewConfiguration` for each layer. Where exactly is this located in the json? 
Is it at the same level as the "category" property of a layer and how is that property called?

daniel-wer:
> When importing a DataSet, which has a previousDataSource, the scale of the previousDataSource gets overwritten with (0,0,0)

@MichaelBuessemeyer could you reproduce this and if so did you fix it as well?

youri-k:
It is similarly named, so `defaultViewConfiguration` and the format is described [here](https://github.com/scalableminds/webknossos/blob/master/frontend/javascripts/types/schemas/dataset_view_configuration.schema.js#L91-L96)

frontend/javascripts/dashboard/dataset/dataset_import_view.js
```diff
@@ -97,28 +140,64 @@ class DatasetImportView extends React.PureComponent<Props, State> {
...
+        } else {
+          // If the current datasource json is invalid, the inferred version should be used automatically.
+          dataSource = inferredDataSource;
+          dataSourceSettingsStatus.isJSONFormatValid = IsJSONFormatValidEnum.BrokenJson;
+          dataSourceSettingsStatus.appliedSuggestions = AppliedSuggestionsEnum.Yes;
         }
+        const inferredDataSourceWithCorrectedScale = ensureValidScaleOnInferredDataSource(
+          savedDataSourceOnServer,
+          inferredDataSource,
+        );
```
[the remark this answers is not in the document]
youri-k: I can confirm this for the bounding box and the largest segment id, but not for the voxel size (, if this is the same as `elementClass`). This should be set in the WKW header, so it should always be correct. Therefore, you can exclude this from the ensuring method I think.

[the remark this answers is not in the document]
MichaelBuessemeyer: An internal conversation showed the following results:
The backend cannot guess the bounding boxes for knossos data layers but this is fine as knossos cubes are not first class citizen.
The guessing of bounding boxes works fine for wkw datalayers.
Thus the "ensuring"-method only needs to be adjusted to ensure the usage of the saved largestSegmentId in the datasource-properties.json.

You can check whether the `largestSegmentId` is used correctly by removing the `elementClass` from the segmentation layer of the dataset you are using to experiment with.
Then the suggested settings should include the saved `largestSegmentId` instead of `0`.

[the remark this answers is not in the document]
MichaelBuessemeyer: The voxel size is no in the  ensuring method :+1:

daniel-wer (commented):
Nice, code looks mostly good, I added an optional suggestion regarding type safety.

During testing, I encountered an issue while following your instructions:

> - Create a copy of a dataset.
> - Open the datasource-properties.json file of that copy and remove the layers so that the dataLayers properties looks like this: "dataLayers": [],
> - Now reload all datasets in the dashboard. The copy should appear with the "Zero layer error" and you should be suggested to import this dataset.
> - In the import view there should be an error message at the top listing the "Zero layer error". If you are using the "use wk suggested settings" option in the blue banner, the scale should be equal to the scale in the datasource.json and not [0,0,0].

I did this, but for a dataset with a segmentation layer. When clicking import after accepting the suggested settings, nothing happened, but also no validation error message was displayed. I activated the "Advanced" view and clicked import again, which showed a validation message that one layer was neither a color nor a segmentation layer. I was able to import the dataset, after updating the `largestSegmentId` from 0 to 1.
Could you check why the validation error message was not displayed in the simple mode. Ideally, it would've also hinted at the `largestSegmentId` instead of that the layer was neither color nor segmentation.

Additionally, you said
> You can check whether the largestSegmentId is used correctly by removing the elementClass from the segmentation layer of the dataset you are using to experiment with.
Then the suggested settings should include the saved largestSegmentId instead of 0.

I tested this as well and the suggested setting were correct. However, after confirming the suggestion, the scale and largestSegmentId fields were empty and the dataset couldn't be imported. Only after switching to "Advanced" mode, those fields were filled and the dataset could be imported.

frontend/javascripts/dashboard/dataset/dataset_import_view.js
```diff
@@ -94,6 +94,40 @@ export type FormData = {
...
+      layer => layer.category === "segmentation",
+    );
+    const savedSegmentationLayerSettings = savedDataSourceOnServer.dataLayers.find(
+      layer => layer.category === "segmentation",
+    );
+    if (
+      segmentationLayerSettings != null &&
+      savedSegmentationLayerSettings != null &&
+      segmentationLayerSettings.largestSegmentId === 0 &&
+      savedSegmentationLayerSettings.category === "segmentation"
```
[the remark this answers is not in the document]
MichaelBuessemeyer: This additional check is needed because flow is not smart enough to know that savedSegmentationLayerSettings must be settings for a segmentation layer and not for a color layer.

I added a comment for this :+1:

MichaelBuessemeyer:
> I did this, but for a dataset with a segmentation layer. When clicking import after accepting the suggested settings, nothing happened, but also no validation error message was displayed. I activated the "Advanced" view and clicked import again, which showed a validation message that one layer was neither a color nor a segmentation layer. I was able to import the dataset, after updating the largestSegmentId from 0 to 1.


Uff this was quite hard to find out:
When you accept the suggestions, new form fields are added to the form.
Somehow when a new form element is added, antd's `getFieldDecorator` does not set the initial value of the new form element, although the value is already provided indirectly. For this to work, you have to set the initial value yourself. Afterward, you must also call `validateFields` to enforce validation/error messages on the new field values.

> Could you check why the validation error message was not displayed in the simple mode. Ideally, it would've also hinted at the largestSegmentId instead of that the layer was neither color nor segmentation.

This works now for the simple form. I think fixing the JSON Schema validation errors is not possible / out of scope as they are provided by the library.

> I tested this as well and the suggested setting were correct. However, after confirming the suggestion, the scale and largestSegmentId fields were empty and the dataset couldn't be imported. Only after switching to "Advanced" mode, those fields were filled and the dataset could be imported.

This is the same error as before. I resolved it by simply setting the initial value manually.

frontend/javascripts/dashboard/dataset/dataset_import_view.js
```diff
@@ -302,13 +302,17 @@ class DatasetImportView extends React.PureComponent<Props, State> {
...
+        currentState => {
+          const updatedStatus = {
+            ...currentState.dataSourceSettingsStatus,
+            appliedSuggestions: AppliedSuggestionsEnum.Yes,
+          };
+          return { dataSourceSettingsStatus: updatedStatus };
+        },
+        // Enforce validation as antd does not do this automatically.
+        () => this.props.form.validateFields(),
+      );
```
MichaelBuessemeyer: I tried to enforce the field validation after line 304 (after updating the values with `setFieldsValue`) but this did not work correctly and I do not know why. I do not think that `setFieldsValue` is asynchronous :/.

Therefore I just added it after the state update (line 314) and this worked fine a few times I tried. Can you think of a better option or do you think that's alright @daniel-wer?

MichaelBuessemeyer:
@daniel-wer Could you please give this another and hopefully last testing session? :test_tube:

frontend/javascripts/dashboard/dataset/dataset_import_view.js
daniel-wer: Hmm that's strange. By "this did not work correctly" - do you mean that the validation failed or was there some other error?

frontend/javascripts/oxalis/model_initialization.js
```diff
@@ -322,17 +324,15 @@ function initializeDataset(
...
+  mutableDataset.dataSource.dataLayers = updatedDataLayers;
 
   serverTracingAsVolumeTracingMaybe(tracing).map(volumeTracing => {
     const newDataLayers = setupLayerForVolumeTracing(dataset, volumeTracing);
-    // $FlowExpectedError[incompatible-use] We mutate the dataset here to avoid that an outdated version is used somewhere else
-    dataset.dataSource.dataLayers = newDataLayers;
+    mutableDataset.dataSource.dataLayers = newDataLayers;
   });
 
   ensureMatchingLayerResolutions(dataset);
```
daniel-wer, line 334: The `mutableDataset` should probably also be used in this line.

daniel-wer (commented):
Great fixes, my previously failing test cases now work as expected :+1: 

I tested some of the other cases as well and the only thing I noticed was that in case 2:

> The datasource.json on the server is correct and the server has additional things to infer.
In that case, the user should be offered to accept these suggestions. Previously, they were adapted directly which might have caused errors that prevented a person from saving other changes done.

the diff that is shown when clicking on `The JSON-encoded difference between the current and the suggested version can be inspected here.` doesn't seem to be correct. For testing, I removed a single layer and the diff was very large, showing various differences in each layer. I wasn't able to detect a pattern. Could you check whether there's an easy fix for that?

frontend/javascripts/dashboard/dataset/dataset_import_view.js
MichaelBuessemeyer: The validation did succeed without an error, but there should have been an error as the default value of the bounding boxes that was used is `0,0,0,0,0,0`.

MichaelBuessemeyer:
> the diff that is shown when clicking on The JSON-encoded difference between the current and the suggested version can be inspected here. doesn't seem to be correct. For testing, I removed a single layer and the diff was very large, showing various differences in each layer. I wasn't able to detect a pattern. Could you check whether there's an easy fix for that?

I tried to find a rather quick workaround but the problem here is that the data layers are compared by index resulting in the following mismatch: In the currently saved settings, the 2nd layer is at index 0 while in the inferred settings this layer is at index 1. As the compare method uses the index for comparison, both data layers in the settings do not match.
I could not find a quick and suitable solution for this :/

frontend/javascripts/dashboard/dataset/dataset_import_view.js
```diff
@@ -174,54 +288,125 @@ class DatasetImportView extends React.PureComponent<Props, State> {
       Modal.info({
         title,
         width: 800,
-        content: (
-          <Input.TextArea rows={20} style={jsonEditStyle}>
-            {jsonStringify(object)}
-          </Input.TextArea>
-        ),
+        content: <Input.TextArea rows={20} style={jsonEditStyle} value={jsonStringify(object)} />,
```
MichaelBuessemeyer, line 291: I had to make this change after merging the newest master into this branch. Otherwise, the text area would be empty. 

I do not know why this change is needed. Maybe antd was updated? But at least it works :)

MichaelBuessemeyer: I just noticed that antd indeed was upgraded

daniel-wer (approved):
Thank you for the clarifications and for sticking with this PR, LGTM :+1:

PR #4945, opened 2020-11-15
Clean up temporary upload files

### LINKED ISSUE #4884: Follow-Up resumable upload: clean up orphan file chunks
If resumable uploads, as introduced in #4860 fail (server down etc), orphan file chunks may be left on disk. Those should be cleaned up periodically. Check if their name is present in the live wK’s upload chunk registry

youri-k:
- https://___.webknossos.xyz

### Steps to test:
- start upload, but do not complete it
- restart wk
- => temporary upload file should be gone

### Issues:
- fixes #4884 

------
- [x] Ready for review


youri-k:
This checks for orphan file chunks only on server restart because that is the only way elements are removed from the map without being deleted from the disk as well. I don't know, if that's enough, given that some wK instances run for an extended period of time. We could introduce some long timeout, e.g. 10 days, maybe even a month, that ensures this cleanup job is not dependent on server restarts.

fm3 (commented):
LGTM. I think that rare execution on startup is ok.
Do you know if it impacts wk startup time?
If so, maybe do this asynchronously, like 60s after startup?
I’d also suggest adding logging to see which files are deleted (in case anything is ever wrongly deleted then we will have a means to debug that)

youri-k:
> Do you know if it impacts wk startup time?
I tried to measure it with around 10 broken uploads and it took around 50ms, so I wouldn't expect any negative impact. Do you expect more broken uploads?

I added some logging now :slightly_smiling_face:

fm3 (approved):
Ok cool that shouldn’t be a problem then :)  Thanks for checking!

PR #4946, opened 2020-11-15
Prevent postgres errors when reporting datalayer

### LINKED ISSUE #4483: SQL Error: duplicate key value violates unique constraint "dataset_layers_pkey
This happens every now and again when the datastores report their datasets to wK. Sometimes the user sees this after hitting “refresh”. If we can’t get to the bottom of this, we should at least add a retry / make sure the user does not see this in the frontend

youri-k:
- https://___.webknossos.xyz

### Steps to test:
- abc

### Issues:
- tackles #4483 

------
- [x] Ready for review


youri-k:
I found the the SQL Error when uploading a zero-layer dataset.
Sadly, I cannot reproduce the `p_key` bug locally, even after reintroducing an absent dataset.
I noticed that the serializable level was not set like in other composed queries, so I suggest trying this and see if it still happens in production.

app/models/binary/DataSet.scala
```diff
@@ -321,20 +320,19 @@ class DataSetResolutionsDAO @Inject()(sqlClient: SQLClient)(implicit ec: Executi
...
             }
           }
-        }.flatten
-      }
+        }
       case _ => List()
     }
     for {
-      _ <- run(DBIO.sequence(List(clearQuery) ++ insertQueries).transactionally)
+      _ <- run(DBIO.sequence(List(clearQuery) ++ insertQueries).transactionally.withTransactionIsolation(Serializable))
```
fm3, line 335: I believe anything that is run as `Serializable` aborts on conflict, so we should probably add a retry around it, as in the other places where this happens.

app/models/binary/DataSet.scala
```diff
@@ -445,10 +443,11 @@ class DataSetDataLayerDAO @Inject()(sqlClient: SQLClient, dataSetResolutionsDAO:
         dataLayers.map(d => sanitize(d.name)))}"
     val clearQuery = sqlu"delete from webknossos.dataset_layers where _dataSet = ${_dataSet}"
 
+    // TODO if we redefine dataSet.usable, this can be changed back
```
fm3, line 446: please move this todo line to the relevant issue instead

fm3 (changes requested):
Cool, thanks for looking into that :) Also thanks for fixing the formatting!
Since you said you could not reproduce the duplicate key error, I’d keep the original issue open until we’ve found a new clue there or until we are sure that it does not occur again. Instead, we can remove the line about zero-layer datasets from the original issue.

fm3 (approved):
LGTM

PR #2131, opened 2017-11-10
Support cyclic trees

daniel-wer:
This PR adds support for cyclic trees (they're not really trees any longer ^^). The warning in the abstract tree viewer regarding cyclic trees was broken, opening the tab resulted in an infinite loop that nearly crashed my system. I've added a small cycle check and render an error message if a cyclic tree was detected.
The only other place I could find that assumed non-cyclic trees was the deleteNode/deleteEdge functionality. I've fixed this as well.

We should also think about whether we want to allow to create cyclic trees from within webknossos. Currently the merge functionality will not allow this, and maybe this is fine - we could wait for the lab to request that feature.

Does anyone else know of other places in the code that assume there are no cyclic trees? (@normanrz @hotzenklotz @philippotto)

@jfrohnhofen As far as I know, the backend makes no assumptions about the absence of cycles in trees, right?

### Mailable description of changes:
 - Cyclic trees/graphs are now officially supported by webKnossos. Splitting/Merging actions on those trees will no longer break tracings.

### Steps to test:
- Upload this nml [cyclic.nml.zip](https://github.com/scalableminds/webknossos/files/1461491/cyclic.nml.zip) or create other cyclic nmls
- Split trees, undo, redo, try to break webknossos - webknossos should not break :D

### Issues:
- fixes #2130

------
- [x] Ready for review

Issue #1857, opened 2017-06-01
Better update action diffing


daniel-wer:
Adapt our diffing algorithm to detect tree merges and splits to:

  - avoid possibly humongous JSON requests to the server (that fail if too big)
  - avoid broken tracings due to those big requests
  - increase overall tree merge/split performance, which takes a couple of minutes currently for a 30000 node tree

<a href="https://timer.scm.io/repos/537237ce1a00001a003c4a14/issues/1857/create?referer=github" target="_blank">Log Time</a>

PR #4950, opened 2020-11-23
Add context menu for nodes

### LINKED ISSUE #4906: Shortest path (length) between two nodes / More UI
It should be possible to get the shortest distance between two nodes (See https://discuss.webknossos.org/t/show-pathlength-for-a-skeleton/1620/8.)

I think, the crucial part will be to do the UI well. For such scenarios, a context menu might be a good option? Shift+Rightclick on a node could open a menu with the following options:

- "create edge to this node" (if possible)
- "delete edge to this node" (if possible)
- "delete this node"
- "measure shortest distance to this node"
- "measure total distance of this tree"

etc.

What do you think? @normanrz @fm3 @daniel-wer ?

MichaelBuessemeyer:
This PR adds a context menu for nodes that offers various actions.

Currently, the node context menu is only supported on nodes.
I could also change it, so that it is possible to open it up anywhere, no matter whether there is a node or not.
The following operations are currently implemented and are available if the menu is opened on an existing node:
- One of the following:
    - Create Edge To This Node (When the active node and the clicked node are in the same tree)
    - Create Edge & Merge With This Tree (When the active tree and the tree the clicked node is in are not the same)
- Delete Edge To This Node (Only enabled, when the active node is directly connected to this node)
- Measure Length To This Node (Only enabled, when the active node and the clicked node are in the same tree)
- Measure Length Of This Tree

Additional useful feature suggestions are welcome :smile_cat: 

- https://___.webknossos.xyz

### Steps to test:
- Open a skeleton tracing and create some trees.
- The context menu should open with shift + right-click and should automatically close when an option is clicked or the user clicks somewhere else.
    - Currently, firefox still opens up the usual context menu which is kinda bothersome. I'll try to work out a fix for this.
- Each option in the context menu should do what is says and should only be enabled if the action makes sense and is possible.
- Play around with the options and try out edge cases like deleting a node in a "branch" should split the current tree.

### Issues:
- fixes #4906

------
- [x] Ready for review


frontend/javascripts/oxalis/controller/combinations/skeletontracing_plane_controller.js
```diff
@@ -340,6 +345,42 @@ function onClick(
...
+
+  const { volume } = state.tracing;
+  if (!volume || volume.activeTool === VolumeToolEnum.MOVE) {
+    // We avoid creating nodes when in brushing mode.
+    const nodeId = maybeGetNodeIdFromPosition(planeView, position, plane, isTouch);
+    if (nodeId != null && nodeId > 0) {
+      showNodeContextMenuAt(event.pageX, event.pageY, nodeId);
+    } else {
+      setWaypoint(calculateGlobalPos(position), ctrlPressed);
+    }
```
philippotto, line 380: If I'm reading the code correctly, a context menu is opened on right click IFF the cursor is over an existing node. Is this correct? I think, it would be safer to only open the context menu when an additional modifier is pressed (e.g., shift). Otherwise, creating a new node very close to another node would not work, anymore. Also, we would lose the possibility to open context menus on spots where no node was placed (e.g., the context menu could also show "create node", "create branchpoint" there).

MichaelBuessemeyer: > If I'm reading the code correctly, a context menu is opened on right click IFF the cursor is over an existing node. Is this correct?

Indeed this is correct :+1: 

> I think, it would be safer to only open the context menu when an additional modifier is pressed (e.g., shift). Otherwise, creating a new node very close to another node would not work, anymore. Also, we would lose the possibility to open context menus on spots where no node was placed (e.g., the context menu could also show "create node", "create branchpoint" there).

Thanks for the valuable feedback. I'll apply it now :smile_cat:

MichaelBuessemeyer:
Suggested Features:
- [ ] directly show the euclidian distance between the clicked and the active node in the menu.
- [ ] Add hint about the move node shortcut in the menu
- [ ] "Hide this tree"
- [ ] Add a version of the menu that has no selected node and has options like: 
  - [ ] create node here 
  - [ ] create new tree here
  - and maybe even more

frontend/javascripts/oxalis/view/node_context_menu.js
```diff
@@ -0,0 +1,314 @@
...
+      <tr>
+        <td>CTRL + ,</td>
+        <td>Navigate to preceding Active Node</td>
+      </tr>
+      <tr>
+        <td>CTRL + Left Click / CTRL + Arrow Keys</td>
+        <td>Move the Active Node</td>
+      </tr>
+    </tbody>
+  </table>
```
MichaelBuessemeyer, line 70: What else could we list here?

MichaelBuessemeyer:
I would stop here adding no more features as otherwise this PR only becomes bigger and therefore harder to review.

@philippotto Could you please review this PR?

And do you have an idea on how to handle that firefox opens a context menu on `right click + shift`? I could not fix this / catch this event in firefox.

frontend/javascripts/oxalis/controller/combinations/skeletontracing_plane_controller.js
```diff
@@ -340,11 +345,51 @@ function onClick(
...
     return;
   }
+
+  const { volume } = state.tracing;
+  if (!volume || volume.activeTool === VolumeToolEnum.MOVE) {
+    // We avoid creating nodes when in brushing mode.
+    const nodeId = event.shiftKey
+      ? maybeGetNodeIdFromPosition(planeView, position, plane, isTouch)
+      : null;
+    const rotation = getRotationOrtho(activeViewport);
```
philippotto, line 382: What do you think about passing `activeViewport` to `showNodeContextMenuAt` so that this function can calculate the rotation on its own. This feels more general to me (in case the active viewport is later needed for something else), as we don't pass derived data but instead the "source data".

frontend/javascripts/oxalis/model/actions/skeletontracing_actions.js
```diff
@@ -233,6 +239,12 @@ export const createNodeAction = (
   timestamp,
 });
 
+export const createEdgeAction = (sourceNodeId: number, targetNodeId: number): CreateEdgeAction => ({
+  type: "CREATE_EDGE",
+  sourceNodeId,
+  targetNodeId,
+});
```
philippotto, line 246: @daniel-wer added support for cyclic trees (aka graphs ^^) to webKnossos in https://github.com/scalableminds/webknossos/pull/2131, but since then we didn't really encourage creating those. I'm inclined to keep it that way, since there haven't been any requests for this (or were there any?).
So, maybe just reuse the `MergeTreesAction` (and only show it for different trees)?
Do you agree?

frontend/javascripts/oxalis/model/reducers/skeletontracing_reducer.js
```diff
@@ -155,6 +156,26 @@ function SkeletonTracingReducer(state: OxalisState, action: Action): OxalisState
...
+              update(state, {
+                tracing: {
+                  skeleton: {
+                    trees: { $set: trees },
+                  },
+                },
+              }),
+            )
+            .getOrElse(state);
+        }
```
philippotto, line 177: see previous comment

frontend/javascripts/oxalis/model/reducers/skeletontracing_reducer_helpers.js
```diff
@@ -633,12 +633,43 @@ export function deleteTree(
...
+  };
+
+  const newTrees = update(trees, {
+    [targetTree.treeId]: {
+      edges: {
+        $set: targetTree.edges.addEdges(sourceTree.edges.asArray().concat(newEdge)),
+      },
+      comments: { $set: targetTree.comments.concat(sourceTree.comments) },
+      branchPoints: { $set: targetTree.branchPoints.concat(sourceTree.branchPoints) },
+    },
```
philippotto, line 662: see previous comments

frontend/stylesheets/_general.less
```diff
@@ -35,3 +35,19 @@ p {
   filter: FlipH;
   -ms-filter: "FlipH";
 }
+
+.table-with-borders-between-columns {
+  td {
+    padding: 2px 10px;
+    border-left: 1px solid rgba(0, 0, 0, 0.65);
```
philippotto, line 42: I'd remove the border :)

frontend/javascripts/oxalis/view/node_context_menu.js
```diff
@@ -0,0 +1,314 @@
...
+  createTree: () => void,
+|};
+
+type StateProps = {| skeletonTracing: ?SkeletonTracing, datasetScale: Vector3 |};
+
+type Props = {| ...OwnProps, ...StateProps, ...DispatchProps |};
+type NodeContextMenuOptionsProps = {| ...Props, nodeContextMenuNodeId: number |};
+type NoneNodeContextMenuProps = {| ...Props |};
+
+const NAVBAR_HEIGHT = 48;
```
philippotto, line 48: Can we get rid of this constant somehow? Maybe compute it ad-hoc via the DOM element? Seems very brittle in case the navbar height changes at some point.

frontend/javascripts/oxalis/view/node_context_menu.js
```diff
@@ -0,0 +1,314 @@
...
+  const length = api.tracing.measureTreeLength(treeId);
+  notification.open({
+    message: messages["tracing.tree_length_notification"](treeName, formatNumberToLength(length)),
+    icon: <i className="fas fa-ruler" />,
+  });
+}
+
+function NodeContextMenuOptions({
+  skeletonTracing,
+  nodeContextMenuNodeId,
```
philippotto, line 91: How about `clickedNodeId`?

frontend/javascripts/oxalis/view/node_context_menu.js
```diff
@@ -0,0 +1,314 @@
...
+            ? measureAndShowLengthBetweenNodes(activeNodeId, nodeContextMenuNodeId)
+            : null
+        }
+      >
+        Path Length to this Node
+      </Menu.Item>
+      <Menu.Item
+        className="node-context-menu-item"
+        key="measure-whole-tree-length"
+        disabled={activeNodeId == null || !areInSameTree || isTheSameNode}
```
philippotto, line 178: Why does this need disabling? Shouldn't one be always able to calculate the total path length of the clicked tree?

frontend/javascripts/oxalis/view/node_context_menu.js
```diff
@@ -0,0 +1,314 @@
...
+        key="hide-tree"
+        onClick={() => hideTree(nodeContextTree.treeId)}
+      >
+        Hide this Tree
+      </Menu.Item>
+    </Menu>
+  );
+}
+
+function NoneNodeContextMenu({
```
philippotto, line 194: Maybe rename to `NoNodeContextMenuOptions` to keep it analoguous to `NodeContextMenuOptions` (also note `No` instead of `None`).

frontend/javascripts/oxalis/view/node_context_menu.js
```diff
@@ -0,0 +1,314 @@
...
+  mergeTrees,
+  deleteNode,
+  setActiveNode,
+  hideTree,
+}: NodeContextMenuOptionsProps) {
+  if (skeletonTracing == null) {
+    return null;
+  }
+  const { activeTreeId, trees, activeNodeId } = skeletonTracing;
+  const nodeContextTree = findTreeByNodeId(trees, nodeContextMenuNodeId).get();
```
philippotto, line 104: Similarly, I suggest `clickedTree` instead of `nodeContextTree`.

philippotto (changes requested):
Awesome stuff! This will be a great usability improvement :)

I left some code comments and also noted some things during testing:



- In a hybrid tracing, pressing the shift modifier will open a tooltip in the toolbar. however, that tooltip doesn't mention "rightclick to open contextmenu" (it only says that a node can be selected).
- The "info"-stuff in the toolbar is cool! However, I'd move it to the bottom, so that one doesn't need to move the mouse across the info stuff to trigger an action.
- The info part could also show the node id of the clicked node and the corresponding tree id. Also, the clicked (node's) position could be shown in another line (again, clicking it would copy that position).
- Maybe rename "Navigate to subsequent Active Node" to "Activate next node" (same for previous)
- It's cool that the direct distance to the active node is immediately visible in the context menu. However, copying that distance is hard since the contextmenu immediately vanishes when selecting something. In general, this seems fine to me, but maybe we can simply copy the distance to the clipboard when clicking the entry?
- Why is the "Activate Next/Previous Node" action under "possible interactions" ? since it's a simple trigger (as opposed to the move action which needs additional interaction), we could simply list it in the top-level menu? however, it might be a bit weird, since it doesn't have to do anything with the clicked position :thinking: but that's a problem in general. Overall I'm a bit hesitant about the "possible interactions" UI. It's not very typical (afaik) to have such a help section in a context menu. What do you think about removing this for the first iteration?

I'll have a look regarding the context menu bug in firefox!

frontend/javascripts/oxalis/view/node_context_menu.js
philippotto: I think, you can completely get rid of this if you move the contextmenu within `tracing_layout_view.js` into the `<div id={canvasAndLayoutContainerID} style={{ position: "relative" }}>` component. That way, the absolute position of the context menu is relative to that child (which is positioned below the navbar).

philippotto:
> Currently, firefox still opens up the usual context menu which is kinda bothersome. I'll try to work out a fix for this.

This seems to be by design :( Then, I suggest to also allow `Meta + Rightclick` to open the wk context menu. It's a bit weird (which is why I wouldn't make it the default), but would be a workaround for firefox. When the user uses firefox and uses `shift + rightclick`, you could show a toast that they should use `Win/Cmd + Right click` instead, since they are using firefox.

Btw, I wouldn't use ctrl + rightclick, since mac users without a right mouse button would use ctrl + leftclick to get a rightclick. And alt + rightlick wouldn't work on default linux setups since that opens an OS context menu.

frontend/javascripts/oxalis/controller/combinations/skeletontracing_plane_controller.js
MichaelBuessemeyer: I do not have a strong opinion about this. My goal here was that the context menu component simply gets all the needed values. 

The way you suggested seems fine to me too :+1:

frontend/javascripts/oxalis/model/actions/skeletontracing_actions.js
MichaelBuessemeyer: Hmm, I see what you mean. By adding this option, it is very easy to create a cyclic tree and thus this option encourages this.

I removed the possibility to add an edge between two nodes of the same tree :+1:

frontend/javascripts/oxalis/view/node_context_menu.js
MichaelBuessemeyer: The problem here is that the position from the event is relative to the whole page but the context menu is positioned within the main part of the tracing view. This container is relative and has the y offset that is equal to the navbar height. That's why the context menu is positioned too low if the navbar height is not subtracted.

MichaelBuessemeyer: But placing it at the same hierarchy level as the nml upload container works fine as the container has the position set to relative :+1:

frontend/stylesheets/_general.less
MichaelBuessemeyer: Hmmmmm..... :thinking:  ok :)

MichaelBuessemeyer:
TODO List:
> - [ ] In a hybrid tracing, pressing the shift modifier will open a tooltip in the toolbar. however, that tooltip doesn't mention "rightclick to open contextmenu" (it only says that a node can be selected).
> - [x] The "info"-stuff in the toolbar is cool! However, I'd move it to the bottom, so that one doesn't need to move the mouse across the info stuff to trigger an action.
> - [ ] The info part could also show the node id of the clicked node and the corresponding tree id. Also, the clicked (node's) position could be shown in another line (again, clicking it would copy that position).
> - [ ] Maybe rename "Navigate to subsequent Active Node" to "Activate next node" (same for previous)
> - [ ] It's cool that the direct distance to the active node is immediately visible in the context menu. However, copying that distance is hard since the contextmenu immediately vanishes when selecting something. In general, this seems fine to me, but maybe we can simply copy the distance to the clipboard when clicking the entry?
> - [ ] Why is the "Activate Next/Previous Node" action under "possible interactions" ? since it's a simple trigger (as opposed to the move action which needs additional interaction), we could simply list it in the top-level menu? however, it might be a bit weird, since it doesn't have to do anything with the clicked position :thinking: but that's a problem in general. Overall I'm a bit hesitant about the "possible interactions" UI. It's not very typical (afaik) to have such a help section in a context menu. What do you think about removing this for the first iteration?
- [ ] Implement the firefox workaround

MichaelBuessemeyer:
> Why is the "Activate Next/Previous Node" action under "possible interactions" ? since it's a simple trigger (as opposed to the move action which needs additional interaction), we could simply list it in the top-level menu? however, it might be a bit weird, since it doesn't have to do anything with the clicked position :thinking: but that's a problem in general. Overall I'm a bit hesitant about the "possible interactions" UI. It's not very typical (afaik) to have such a help section in a context menu. What do you think about removing this for the first iteration?

I think removing it is fine.

MichaelBuessemeyer:
On hold because of planned keyboard shortcuts / UI toolbar rework in #4997

MichaelBuessemeyer:
@philippotto I applied all the feedback, added the context menu to the docs, and tested each feature a bit. Could you please check this PR again?

frontend/javascripts/libs/mjs.js
```diff
@@ -86,7 +87,7 @@ M4x4.transformPoints = function transformPoints(
   if (r == null) {
     r = new Float32Array(points.length);
   }
-  for (let i = 0; i < points.length; i += 3) {
+  for (let i = 0; i < points.length; i = i + 3) {
```
philippotto, line 90: Was there any specific reason for this?

philippotto (approved):
Awesome stuff! Works very well for me and the code looks good, too :)

frontend/javascripts/libs/mjs.js
MichaelBuessemeyer: Yeah, I came accross the comment in line 56 "// DO NOT CHANGE to let compound assignment (+=) as V8 cannot optimize this" because my IDE optimized the assignment in line 57 to `+=`.

Therefore I thought that if the comment is correct, the compound assignment in line 89 should is wrong. That why I changed this. I personally do not know why this comment was added, but I can look it up:
Here we go: This is where this comment/change was introduced 
https://github.com/scalableminds/webknossos/issues/1857#issuecomment-305641037

MichaelBuessemeyer: However, I did not do any performance testing on this

MichaelBuessemeyer: so can I merge this?

philippotto: Ah, interesting. I assume that's not really up to date anymore, since the comment is 4 years old and already then stated:

> So it looks like the new V8 optimizing compiler Turbofan will be turned on by default in the next chrome version (59), which doesn't suffer from those deoptimization issues.

Maybe just undo your change :)

PR #4957, opened 2020-11-25
Disable 'use existing segmentation layer' when that layer doesn't exist

philippotto:
- https://___.webknossos.xyz

### Steps to test:
- open dashboard and use "create volume tracing" and "create hybrid tracing". verify that "use existing segmentation layer" is disabled if it doesn't exist

Superseded by https://github.com/scalableminds/webknossos/pull/4939, but this PR is meant as a a hotfix.

### DIFF
--- a/frontend/javascripts/dashboard/advanced_dataset/dataset_action_view.js
+++ b/frontend/javascripts/dashboard/advanced_dataset/dataset_action_view.js
@@ -11,8 +11,11 @@ import messages from "messages";
 const createTracingOverlayMenu = (dataset: APIMaybeUnimportedDataset, type: TracingType) => {
   const typeCapitalized = type.charAt(0).toUpperCase() + type.slice(1);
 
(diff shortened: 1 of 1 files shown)


daniel-wer (approved):
LGTM, unless @normanrz wanted the whole "Use ... Segmentation Layer" sub menu to be hidden if no segmentation layer exists (in that case a click on "Start Hybrid Annotation" could directly start a hybrid annotation with a new segmentation layer as there is no other option).

normanrz:
Actually, I was hoping that the menu wouldn't be there at all and the annotation would be created right away.

philippotto:
Yeah, #4939 will take care of this properly, but for this hotfix I decided to go with this minimal change, since the proper solution would be a bit more involved (and would cause several merge conflicts). Is this alright?

normanrz:
Alright

PR #4958, opened 2020-11-26
Generate agglomerate skeletons from voxelytics agglomerate file

### LINKED ISSUE #4886: Add / import skeleton from agglomerate file
[Link to Discuss](https://discuss.webknossos.org/t/add-import-skeleton-from-agglomerate-file/1623)

fm3:
Generate skeletons from an activated voxelytics agglomerate file mapping by `shift` middle-clicking onto a cell of a segmentation layer.

- https://___.webknossos.xyz

### TODO
 - [x] readable error toasts (from backend)
 - [x] antd messages about having sent a request ”loading skeleton for agglomerate 123”
 - [x] frontend eats the error message? (seeing empty toast)
 - [x] error message if clicked with shortcut but action is not available (e.g. no agglomerate mapping was selected)
 - [x] measure time in backend
 - [x] if “render missing data black” is deactivated, clicking on rendered mag2 if zoomed in does not send requests.
 - [x] decide if skeletons should be saved in annotation (or should this entire thing be a view-mode/sandbox feature?)
 - [x] do not load isosurface on ctrl+shift click, even if it is active in the settings
 - [x] should be able to request an agglomerate again if selected mapping changed
 - [x] skeletons can be loaded multiple times, currently - kind of related to "decide if skeletons should be saved in annotation ..." - _We decided that this is not a problem, the user will notice that there are duplicated trees_
 - [x] if requests fail, progress message should disappear, should be able to try again
 - [x] re-enable full tests before merging
 - [x] send request to fallback layer instead of volume annotation layer for volume/hybrid annotations
 - [x] show error if there is a volume annotation layer but no fallback layer -> _solved because there is no agglomerate file mapping in that case_
 - [x] find new shortcut, `ctrl + shift + leftclick` is already taken for the fill tool :see_no_evil: 
 - [x] document keyboard shortcut _Shift + Middle Click_

### Steps to test:
- Locally, use a dataset with agglomerate file mappings (for example, by linking the voxelytics test output). If you don't have one, I can send a test dataset.
- Activate a mapping and `shift + middle click` onto a cell -> Progress notifications should be shown and a skeleton should be loaded.
- Test all the error cases:
  - without activating a mapping
  - with an activated json mapping instead of hdf5
  - clicking on the background (segment 0)
- Check the TODOs above for further test cases

### Issues:
- fixes #4886 

------
- [x] Ready for review


webknossos-datastore/conf/com.scalableminds.webknossos.datastore.routes
```diff
@@ -26,6 +26,8 @@ GET         /datasets/:organizationName/:dataSetName/layers/:dataLayerName/mappi
 
 # Agglomerate files
 GET         /datasets/:organizationName/:dataSetName/layers/:dataLayerName/agglomerates                          @com.scalableminds.webknossos.datastore.controllers.DataSourceController.listAgglomerates(organizationName: String, dataSetName: String, dataLayerName: String)
+GET         /datasets/:organizationName/:dataSetName/layers/:dataLayerName/agglomerates/:mappingName/skeleton/:agglomerateId                 @com.scalableminds.webknossos.datastore.controllers.DataSourceController.generateAgglomerateSkeleton(organizationName: String, dataSetName: String, dataLayerName: String, mappingName: String, agglomerateId: Long)
+GET         /datasets/:organizationName/:dataSetName/layers/:dataLayerName/agglomerates/:mappingName/skeletonUnsafe/:agglomerateId                 @com.scalableminds.webknossos.datastore.controllers.DataSourceController.generateAgglomerateSkeletonUnsafe(organizationName: String, dataSetName: String, dataLayerName: String, mappingName: String, agglomerateId: Long)
```
youri-k, line 30: Don't forget to remove this debug route :slightly_smiling_face:

youri-k (commented):
LGTM :slightly_smiling_face:

webknossos-datastore/app/com/scalableminds/webknossos/datastore/services/AgglomerateService.scala
```diff
@@ -130,77 +135,73 @@ class AgglomerateService @Inject()(config: DataStoreConfig) extends DataConverte
...
+      )
+      val duration = System.nanoTime() - startTime
+      if (duration > 100 * 1e6) {
+        logger.debug(s"Generating skeleton from agglomerate file took ${math
+          .round(duration / 1e6)} ms (${skeletonEdges.length} edges, ${nodes.length} nodes).")
+      }
 
-    val skeletonEdges = edges.map { e =>
-      Edge(source = e(0).toInt, target = e(1).toInt)
+      Fox.successful(skeleton)
```
youri-k, line 203: I'd prefer a simple box here, but that is personal preference, so leave it as it is.

daniel-wer:
> frontend eats the error message? (seeing empty toast)

I can reproduce this and see the `messages` json in the network tab, however, when consuming the erroneous reponse with our generic error handler, the response text is shown as "". I have not found out why, yet. I assume this is the same status code and response the backend usually sends for other errors, right?

fm3:
yes, we checked that other errors, for example »there is no task available« comes with the same error format and also http status 400

daniel-wer:
Thanks for checking! I just found the culprit. The error message was swallowed because array buffers are fetched via web workers per default and the error response doesn't seem to be transferred correctly.

daniel-wer:
@fm3 I took care of the remaining frontend To Do's, thanks for maintaining that list! I will do a full test tomorrow. Could you merge the master and solve the merge errors once you find the time? :)

fm3:
Cool, thanks! I resolved the merge conflicts :) 
One thing I noticed during testing: If the shortcut is clicked on a hybrid annotation, the request is sent with the volume tracing id as layer name, which fails. Would it be possible to either disable this completely if a volume tracing is present, or, maybe better, send it to the fallback segmentation layer only (if one is present)? That would make this more compatible with the hybrid-as-default path

daniel-wer:
@fm3 Thanks a lot! I've added and fixed a couple more TODOs in the list in the PR description.

I also noticed that the `Ctrl + Click` shortcut is already taken for volume tracings :see_no_evil: This could be solved by adding a general context menu which could include this functionality (but I think #4950 only adds a menu for nodes). Maybe we need to resort to a slightly more exotic shortcut for now.

daniel-wer:
I've tested this with a "real-world" agglomerate file which worked nicely and was really fast :+1: 

I hit the `agglomerateSkeletonEdgeLimit` quite a few times, but to play it safe we can still test the performance impact first, before increasing it. Can the `agglomerateSkeletonEdgeLimit` be configured per deployment or do we need to release a new version for that?

frontend/javascripts/oxalis/controller/combinations/segmentation_plane_controller.js
```diff
@@ -4,12 +4,106 @@ import Model from "oxalis/model";
...
+  const { renderMissingDataBlack } = state.datasetConfiguration;
+  const zoomStep = getRequestLogZoomStep(state);
+  const position = calculateGlobalPos(pos);
+
+  // While render missing data black is not active and there is no segmentation for the current zoom step,
+  // the segmentation of a higher zoom step is shown. Here we determine the the next zoom step of the
+  // displayed segmentation data to get the correct segment ids of the cell that was clicked on.
+  const usableZoomStep = renderMissingDataBlack
+    ? zoomStep
+    : segmentation.cube.getNextUsableZoomStepForPosition(position, zoomStep);
```
philippotto, line 57: What do you think about extracting this code into an `getRenderedZoomStepAtPosition` api method?

philippotto (commented):
Cool stuff :tada: :tada: :tada: Will test tomorrow :)

philippotto:
> Locally, use a dataset with agglomerate file mappings (for example, by linking the voxelytics test output). If you don't have one, I can send a test dataset.

@fm3 or @daniel-wer (don't know who wrote this), can you send the test agglomerate file to me? :)

philippotto (approved):
Awesome! Testing worked flawlessly for me :rocket: Will be a nice christmas present for our users :smiley:

frontend/javascripts/admin/admin_rest_api.js
```diff
@@ -1396,3 +1396,20 @@ export function computeIsosurface(
...
+  layerName: string,
+  mappingId: string,
+  agglomerateId: number,
+): Promise<ArrayBuffer> {
+  return doWithToken(token =>
+    Request.receiveArraybuffer(
+      `${dataStoreUrl}/data/datasets/${datasetId.owningOrganization}/${
+        datasetId.name
+      }/layers/${layerName}/agglomerates/${mappingId}/skeleton/${agglomerateId}?token=${token}`,
+      { useWebworkerForArrayBuffer: false },
```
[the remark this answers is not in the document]
daniel-wer: This is intended, I added a comment:

```
// The webworker code cannot do proper error handling and always expects an array buffer from the server.
// In this case, the server sends an error json instead of an array buffer sometimes. Therefore, don't use the webworker code.
```

frontend/javascripts/oxalis/controller/combinations/segmentation_plane_controller.js
```diff
@@ -4,12 +4,106 @@ import Model from "oxalis/model";
...
+    }
+
+    Store.dispatch(
+      addTreesAndGroupsAction(
+        createMutableTreeMapFromTreeArray(parsedTracing.trees),
+        parsedTracing.treeGroups,
+      ),
+    );
+  } catch (e) {
+    console.error(e);
```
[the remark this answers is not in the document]
daniel-wer: Yes I think there will be, but the "expected" error cases (REST response fails or contains backend error) will show a message. I'm not sure whether we should add error handling for all other possible cases. When calling `parseProtoTracing`, for example, we also expect it to go right in the other spots where we call it. However, I've adapted the catch handler to only hide the notification and re-throw the error, so that the error is logged correctly (but still, no Toast is shown to the user, I think). If we want to show an error Toast to users every time an error is thrown, we should add a general error handler that does that. Currently, this wouldn't be trivial as for some errors a toast was already shown (request lib).
What do you think?

daniel-wer:
@philippotto Thank you for your feedback, I should've addressed all of it :)

Regarding the shortcut to trigger this, what do you think about middle-mouse-click? I know that there are two-button mouses, but most of the time there should be a method to emulate the middle mouse click. Also, it's only targeted at a few users for now and I only see this as a temporary solution until we maybe introduce a context menu where we could place the functionality.

frontend/javascripts/oxalis/controller/combinations/segmentation_plane_controller.js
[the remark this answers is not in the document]
philippotto: Fully agree! I think the fact that the error wouldn't be logged to airbrake was the actual thing which bothered me subconsciously (but instead of identifying that I could only pinpoint the user-notification-thing). So, your recent change covers the issue nicely :+1:

philippotto:
Excellent!

> Regarding the shortcut to trigger this, what do you think about middle-mouse-click? I know that there are two-button mouses, but most of the time there should be a method to emulate the middle mouse click. Also, it's only targeted at a few users for now and I only see this as a temporary solution until we maybe introduce a context menu where we could place the functionality.

Hmm, middle-mouse-drag is already taken for moving the current position. Therefore, I wouldn't use middle-click..
Maybe a normal button in the segmentation-tab would work for now? If a HDF5 mapping is enabled, a button could appear for generating the skeleton at the centered position? And as soon as we have the global context menu, we can put the action there.

daniel-wer:
As agreed upon in our internal discussion, I changed the shortcut to Shift + Middle Click and also added a button to the Segmentation Tab which allows to load the skeleton for the centered cell. While working on this, I noticed a bug concerning tree ids that were too large (see https://github.com/scalableminds/webknossos/issues/5009) and added a workaround as this would otherwise lead to errors when using the skeleton import functionality of this PR in real world scenarios.
Would be great if you could have a final look :)

frontend/javascripts/oxalis/model/reducers/skeletontracing_reducer_helpers.js
```diff
@@ -524,14 +524,18 @@ export function addTreesAndGroups(
...
+      for (const treeId of treeIds) {
+        const tree = trees[treeId];
         if (tree.name === "") {
           tree.name = generateTreeName(state, tree.timestamp, tree.treeId);
         }
       }
 
-      const needsReassignedIds = Object.keys(skeletonTracing.trees).length > 0;
+      // TreeIds > 1024^2 break webKnossos, see https://github.com/scalableminds/webknossos/issues/5009
+      const hasTreeIdsLargerThanMaximum = treeIds.filter(treeId => treeId > 1048576);
```
philippotto, line 536:
```suggestion
      const hasTreeIdsLargerThanMaximum = treeIds.some(treeId => treeId > 1048576);
```

Otherwise, this would return an array, which is always truthy (so, the check in the next line would always evaluate to true). Additionally, this should be better perf-wise (since this can do an early-out).

philippotto (commented):
Awesome stuff & good catch with the tree ids! However, I think, I found a small bug in the workaround for the tree ids. See my one suggestion. Afterwards, this should be good to go :+1: Probably in the new year then, right?

frontend/javascripts/oxalis/model/reducers/skeletontracing_reducer_helpers.js
daniel-wer: :man_facepalming: I forgot to add `.length` in the end. Thanks for catching that and your suggestion is even better with the early-out :+1:

daniel-wer:
> Afterwards, this should be good to go +1 Probably in the new year then, right?

Sounds good to me, start the new year with a nice new feature :)

PR #4959, opened 2020-11-26
Treat zero layer dataset as unusable

### LINKED ISSUE #4941: consider zero-layer datasets as inactive
right now a dataset with a valid datasource-properties.json but zero layers is considered usable/active by the backend.
it might make sense to consider it unusable, which has an influence of the UI options in the front-end

youri-k:
- https://___.webknossos.xyz

### Steps to test:
- upload a zero layer dataset
- should be marked as unusable in the frontend, but not throw an error

### Issues:
- fixes #4941 

------
- [x] Ready for review

### DIFF
--- a/CHANGELOG.unreleased.md
+++ b/CHANGELOG.unreleased.md
@@ -11,7 +11,7 @@ For upgrade instructions, please check the [migration guide](MIGRATIONS.released
 [Commits](https://github.com/scalableminds/webknossos/compare/20.12.0...HEAD)
 
 ### Added
--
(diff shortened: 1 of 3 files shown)


fm3 (approved):
LGTM, I also tried it and seems to work as expected :) 

Does this interact with https://github.com/scalableminds/webknossos/pull/4944 ? Will the scale still be available there for zero-layer datasets?

youri-k:
Yes, this does interact with #4944 and I will make the necessary changes there to make the scale still available.

PR #4960, opened 2020-11-27
Reload Dataset when refreshing single Dataset

### LINKED ISSUE #4562: Show `reload` button in datasets table even if dataset is not (properly) imported
When the datasource-properties json is invalid, the dataset is shown as "not imported" and there is an "import" link. When I fix the json on disk, I have to click "reload" for all datasets. Instead, it would be enough if I could click "reload" for the broken dataset. However, that link is not shown.

youri-k:
- https://reloadunimporteddatasets.webknossos.xyz

### Steps to test:
- When reloading a single Dataset in the `DatasetActionView`, refresh the data from the server
- The original issue was concerned about unimported Datasets, but I decided to reload the data for every Dataset 

### Issues:
- fixes #4562 

------
- [x] Ready for review


frontend/javascripts/dashboard/dataset/dataset_cache_provider.js
```diff
@@ -89,8 +91,34 @@ export default function DatasetCacheProvider({ children }: { children: Node }) {
...
+    try {
+      setIsLoading(true);
+      const updateIdx = datasets.findIndex(
+        dataset =>
+          dataset.name === datasetId.name &&
+          dataset.owningOrganization === datasetId.owningOrganization,
+      );
+      if (updateIdx !== -1) {
+        const updatedDataset = await getDataset(datasetId);
+        // TODO is this bad practice?
```
youri-k, line 106: One open question would be, if it is fine to reuse the `datasets` array.

frontend/javascripts/dashboard/advanced_dataset/dataset_action_view.js
```diff
@@ -135,18 +137,32 @@ class DatasetActionView extends React.PureComponent<Props, State> {
...
-            <Link
-              to={`/datasets/${dataset.owningOrganization}/${dataset.name}/import`}
-              className="import-dataset"
-            >
-              <Icon type="plus-circle-o" />
-              Import
-            </Link>
-
+          <div className="dataset-actions">
+            <React.Fragment>
```
daniel-wer, line 156: The `<React.Fragment>` should not be needed as far as I can see (since there is the wrapping div).

frontend/javascripts/dashboard/dataset/dataset_cache_provider.js
daniel-wer: No, this is considered bad practice as you suspected. In short, mutating state might lead to missed rendering updates because the object reference stays the same (see [this post](https://stackoverflow.com/questions/55178507/mutating-state-from-reacts-usestate-hook), for example).

You could simply do
```javascript
const newDatasets = datasets.slice();
newDatasets[updateIdx] = updatedDataset;
setDatasets(newDatasets);
```

frontend/javascripts/dashboard/dataset/dataset_cache_provider.js
```diff
@@ -89,8 +91,34 @@ export default function DatasetCacheProvider({ children }: { children: Node }) {
...
+        datasets[updateIdx] = updatedDataset;
+        // this is unnecessary, but maybe more expressive
```
daniel-wer, line 108: See my comment above for why this isn't unnecessary :)

frontend/javascripts/dashboard/dataset/dataset_cache_provider.js
```diff
@@ -89,8 +91,34 @@ export default function DatasetCacheProvider({ children }: { children: Node }) {
       setIsLoading(false);
     }
   }
+
+  async function updateDataset(datasetId: APIDatasetId) {
+    if (isLoading) return;
```
daniel-wer, line 96: Could you also add this line to the `fetchDatasets` method above. Otherwise, I'm worried about race conditions where this method is invoked first, then `fetchDatasets` is called and the results are applied in undefined order.

daniel-wer (commented):
Nice work @youri-k, please see my comments regarding your question :)

frontend/javascripts/dashboard/dataset/dataset_cache_provider.js
youri-k: I noticed that this would lead to the abortion of fetching the datasets after triggering the check in `checkDatasets`. Please see if my solution is fine :slightly_smiling_face:

youri-k:
To merge the latest master, I had to restructure the code a bit, so please also check whether that's okay.

One more thing I noticed when testing was that the newly reloaded dataset is automatically bumped to the beginning of the dataset list. This is because it is the most recent dataset that the user interacted with. I don't think this is expected behavior, but I don't know how easy it is to prevent the resorting, but don't break other stuff :thinking:

daniel-wer:
> One more thing I noticed when testing was that the newly reloaded dataset is automatically bumped to the beginning of the dataset list. This is because it is the most recent dataset that the user interacted with. I don't think this is expected behavior, but I don't know how easy it is to prevent the resorting, but don't break other stuff :thinking:

Good point! At least the table is scrolled to the top after hitting Reload, but I'm not able to test whether it jumps to page 1 if reloading a dataset on another page. Maybe the respective updated table entry could be highlighted after the update (with a fade out). But still, the jump is distracting. Any idea, @philippotto ?

daniel-wer:
@philippotto Any opinion on this?

philippotto:
Sorry, this slipped through the cracks!

> Maybe the respective updated table entry could be highlighted after the update (with a fade out). But still, the jump is distracting. Any idea, @philippotto ?

In my opinion, we should try to avoid the jump altogether. It's quite weird to update only one dataset and then have everything change. The reason why this happens right now is as follows: The rendered datasets are loaded from a cache. Asynchronously, we update the cache, but don't propagate the new data to the UI. This means the data always "lags one page-load-cycle behind" which has the advantages that the datasets are quickly rendered (due to the cache) AND that there is no sudden change in the table. This behavior is achieved by passing an `applyUpdatePredicate` parameter to `context.fetchDatasets` in `dataset_view.js`. In essence, the cache is updated, but we ask the cache context to not propagate the new state.

Long story short-ish: This PR implements the `updateDataset` method which updates the newest dataset state and then propagates that (thus, re-introducing the jump). My suggestion would be to make the `updateDataset`-method accept an additional optional parameter which would be called `datasets`. If that parameter is provided, the CacheContext would not update the new dataset within its own datasets state but instead in the passed `datasets` variable (and that is then passed to `setDatasets`).
In my opinion, this would solve the problem relatively easy.

Let me know if you have questions or if you disagree :)

youri-k:
Unfortunately, this does not solve the problem. I figured out that the request to fetch the dataset from the backend updates the `lastUsedByUser` timestamp. Therefore, the dataset gets automatically sorted to the top of the list. I am not sure, how to fix this. A simple solution would be to add an additional parameter to the request to stop the backend from updating the timestamp. What would you think about that?

Additionally, I extended your approach to update both the optional datasets parameter and the cache internal datasets to not lose the update once the user reopens the site.

philippotto:
> Unfortunately, this does not solve the problem. I figured out that the request to fetch the dataset from the backend updates the lastUsedByUser timestamp. Therefore, the dataset gets automatically sorted to the top of the list. I am not sure, how to fix this. A simple solution would be to add an additional parameter to the request to stop the backend from updating the timestamp. What would you think about that?

Ah, very good point. How about this: the back-end still updates the lastUsedByUser property and also returns it to the front-end. BUT: The frontend remembers the old lastUsedByUser value and restores that in the dataset object which is used to update the optional datasets array. The cache's "internal" datasets list, receives the original object. This would work, right?

> Additionally, I extended your approach to update both the optional datasets parameter and the cache internal datasets to not lose the update once the user reopens the site.

Nice, very good observation :+1:

youri-k:
> This would work, right?

Yes it does :tada: 

I had a little trouble with overwriting the property because of flow. Please see if my implementation is okay :slightly_smiling_face:

frontend/javascripts/dashboard/dataset/dataset_cache_provider.js
```diff
@@ -122,18 +123,20 @@ export default function DatasetCacheProvider({ children }: { children: Node }) {
...
           if (!datasetsToUpdate) setDatasets(newDatasets);
         }
+
+        if (optionalUpdateIndex !== -1 && datasetsToUpdate) {
+          const newDatasets = datasetsToUpdate.slice();
+          const copyOfUpdatedDataset = (_.cloneDeep(updatedDataset): any);
+          copyOfUpdatedDataset.lastUsedByUser = newDatasets[optionalUpdateIndex].lastUsedByUser;
+          newDatasets[optionalUpdateIndex] = (copyOfUpdatedDataset: APIMaybeUnimportedDataset);
+          setDatasets(newDatasets);
+        }
```
philippotto, line 139: I think, the entire update logic can be simplified by using a `map` operation like this:

```js
if (datasetsToUpdate) {
  const newDatasets = datasetsToUpdate.map(dataset => {
    if (dataset.name === datasetId.name && ...) {
      const { lastUsedByUser } = dataset;
      return {
        ...updatedDataset,
        lastUsedByUser,
      }
    } else {
      return dataset;
    }
  })
}
```

That way, you don't have to remember indices and chop/join arrays :) The same goes for "internal dataset". You will probably need to always fetch the requested dataset (even if it does not appear in the datasets-arrays,) however, that shouldn't be a problem.

philippotto (changes requested):
One last style recommendation :)

frontend/javascripts/dashboard/dataset/dataset_cache_provider.js
youri-k: Thx for the suggestion. The code looks way cleaner now, but for some reason flow does not like it. Would you mind taking another look?

philippotto (approved):
Thank you for incorporating my nitpicks :) I think, this should be good now (probably merge on monday?).

frontend/javascripts/dashboard/dataset/dataset_cache_provider.js
philippotto: code looks great! for the record: I couldn't figure out how to tame flow. suppressing the error seems like an okay workaround for now :+1:

PR #4962, opened 2020-12-01
Persist the selected overwrite-mode for hybrids and volume tracings

### LINKED ISSUE #4894: Remember "overwrite mode" in the new toolbar
With the new toolbar currently the overwrite mode overwrites other cell IDs both for adding annotation (i.e. left click) and deleting annotation (i.e. right click). This is a change in behavior from the previous version, where only adding annotation overwrites other cell IDs and deleting annotation does not overwrite.

I could not think of a clear use case for overwriting other cell IDs on deletion, and given that this changes the previous default behavior and causes inconvenience in certain cases, maybe it is worth considering no overwriting on deletion for either mode (i.e. changing the mode only affects adding annotation)?

philippotto:
- https://rememberoverwritemode.webknossos.xyz

### Steps to test:
- open a volume or hybrid tracing
- select the brush tool
- verify that the different overwrite behaviors work (can be switched in the toolbar)
- reload the page and check that the last overwrite behavior is active after the reload

### Issues:
- fixes #4894

------
- [X] Ready for review

### DIFF
--- a/frontend/javascripts/messages.js
+++ b/frontend/javascripts/messages.js
@@ -40,7 +40,7 @@ export const settings = {
   mergerMode: "Enable Merger Mode",
   gpuMemoryFactor: "Hardware Utilization",
   autoBrush: "Automatic Brush (Beta)",
-  overwriteMode: "Overwrite Mode (controls the overwrite behavior for volume annotation tools)",
+  overwriteMode: "Volume Annotation Overwrite Mode",
 };
 
 export const layerViewConfigurations = {
--- a/frontend/javascripts/messages.js
+++ b/frontend/javascripts/messages.js
@@ -40,6 +40,7 @@ export const settings = {
   mergerMode: "Enable Merger Mode",
   gpuMemoryFactor: "Hardware Utilization",
   autoBrush: "Automatic Brush (Beta)",
+  overwriteMode: "Overwrite Mode (controls the overwrite behavior for volume annotation tools)",
 };
 
 export const layerViewConfigurations = {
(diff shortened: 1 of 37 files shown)


daniel-wer (commented):
Nicely solved :+1: 

If you add a label for this to the `settings` object in the `messages.js`, this label would be shown to the user when this setting is used as part of a recommended configuration. Currently, no label is displayed. Maybe the typing can be improved, so flow complains if that `settings` object doesn't contain an entry for all user settings, but don't spend too much time on it.

philippotto:
> If you add a label for this to the `settings` object in the `messages.js`, this label would be shown to the user when this setting is used as part of a recommended configuration. Currently, no label is displayed. Maybe the typing can be improved, so flow complains if that `settings` object doesn't contain an entry for all user settings, but don't spend too much time on it.

Good catch! I added a message now.

I also tried to make it more "typesafe" as you suggested, but got quickly saw the rabbit hole which is why I canceled that :) My attempt was something like this:

```
export const settings: $ObjMap<{ ...UserConfiguration, ...DatasetConfiguration }, () => str> = {
  ...
}
```

However, merging UserConfiguration and DatasetConfiguration is not enough, since the object also includes other values. Also, circular dependencies for the types arose, so I gave up :see_no_evil:

frontend/javascripts/messages.js
```diff
@@ -40,6 +40,7 @@ export const settings = {
   mergerMode: "Enable Merger Mode",
   gpuMemoryFactor: "Hardware Utilization",
   autoBrush: "Automatic Brush (Beta)",
+  overwriteMode: "Overwrite Mode (controls the overwrite behavior for volume annotation tools)",
```
daniel-wer, line 43:
```suggestion
  overwriteMode: "Volume Annotation Overwrite Mode",
```

Sorry for the nitpicking, but this label is shown in tables (task type creation and if the user opens a task with recommended settings) and should be rather concise. The `loadingStrategyDescription` which is longer is a special case which is probably shown in a tooltip or so.

daniel-wer (approved):
LGTM if my one suggestion is addressed :)

frontend/javascripts/messages.js
philippotto: Very good point! Thank you for being so attentive!

PR #4973, opened 2020-12-03
Remove unused imports

### LINKED ISSUE #4885: Backend Linter: Enforce absence of unused imports
we have a lot of unused imports in the backend. this could slow down compilation (and increase code complexity needlessly)

maybe we could also include automatic import cleaning in `yarn pretty-backend`

youri-k:
### Issues:
- fixes #4885 

------
- [x] Ready for review


fm3 (approved):
LGTM
There seem to be some false warnings that we may need to suppress in the future (e.g. in routes files and html templates), but let’s go with this for now, and clean up the true warnings over the coming weeks, Then we can decide how we want to change it further :) Thanks for looking into that!

Issue #4971, opened 2020-12-03
make tiff conversion repeatable without changing name


fm3:
if tiff import fails, a retry will fail as the unzip target directory is already present. should be re-tryable.
the cubing request needs to find the right unzipped directory though, which needs some mechanism
related: #4963

PR #4977, opened 2020-12-03
Fix single file uploads 

### LINKED ISSUE #4963: Error uploading dataset zip: directory exists
when uploading a dataset zip (for conversion in this case) I got “is a directory” errors. I think ZipIO.unzipToFolder creates the directory, but there’s a `ensureDirectory` before that in `finishUpload` – why did that ever work? Do you have an idea @youri-k ?

youri-k:
- https://___.webknossos.xyz

### Steps to test:
- upload zip with only file and no empty directories => should now work

### Issues:
- fixes #4963 

------
- [x] Ready for review

### DIFF
--- a/util/src/main/scala/com/scalableminds/util/io/ZipIO.scala
+++ b/util/src/main/scala/com/scalableminds/util/io/ZipIO.scala
@@ -165,7 +165,9 @@ object ZipIO extends LazyLogging {
                      excludeFromPrefix: Option[List[String]] = None)(f: (Path, InputStream) => Box[A])(
       implicit ec: ExecutionContext): Box[List[A]] = {
 
-    def stripPathFrom(path: Path, excludeFromPrefix: List[String]): Path = {
+    def isFileNameInPrefix(prefix: Path, fileName: String) = prefix.endsWith(Paths.get(fileName).getFileName)
+
+    def cutOffPathAtLastOccurrenceOf(path: Path, excludeFromPrefix: List[String]): Path = {
       var lastExcludedIndex = -1
       path.iterator().asScala.zipWithIndex.foreach {
         case (subPath, idx) =>
@@ -185,13 +187,12 @@ object ZipIO extends LazyLogging {
 
     val zipEntries = zip.entries.asScala.filter(e => !e.isDirectory && (includeHiddenFiles || !isFileHidden(e))).toList
 
-    //color, mask, segmentation are the values for dataSet layer categories and a folder only has one category
     val commonPrefix = if (truncateCommonPrefix) {
       val commonPrefixNotFixed = PathUtils.commonPrefix(zipEntries.map(e => Paths.get(e.getName)))
-      val strippedPrefix = stripPathFrom(commonPrefixNotFixed, excludeFromPrefix.getOrElse(List.empty))
+      val strippedPrefix = cutOffPathAtLastOccurrenceOf(commonPrefixNotFixed, excludeFromPrefix.getOrElse(List.empty))
       // if only one file is in the zip and no layer name is given as prefix, do not remove file name as common prefix
       zipEntries match {
-        case head :: tl if tl.isEmpty && strippedPrefix.endsWith(Paths.get(head.getName).getFileName) =>
+        case head :: tl if tl.isEmpty && isFileNameInPrefix(strippedPrefix, head.getName) =>
           strippedPrefix.subpath(0, strippedPrefix.getNameCount - 1)
         case _ => strippedPrefix
       }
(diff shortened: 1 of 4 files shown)


webknossos-datastore/app/com/scalableminds/webknossos/datastore/services/UploadService.scala
```diff
@@ -125,6 +124,8 @@ class UploadService @Inject()(dataSourceRepository: DataSourceRepository, dataSo
             dataSourceRepository.updateDataSource(
               dataSourceService.dataSourceFromFolder(dataSourceDir, dataSourceId.team))
         case e =>
+          // TODO check whether this is too aggressive
+          PathUtils.deleteDirectoryWithContents(dataSourceDir)
```
youri-k, line 128: While testing this PR, I noticed that the Dataset is still reported to wk because of the empty directory that remains after the unsuccessful unzipping. Therefore, I created this method to delete the erroneous `dataSourceDir`, but maybe this too aggressive :thinking:

youri-k: I think this could already solve #4971. I will move the code there and another request to wk to remove the dataset from the db if the unzipping fails. Additionally, we should think about "reserving" the name already for the first request, so similar named datasets fail early and not only in the `finishUpload` request.

util/src/main/scala/com/scalableminds/util/io/ZipIO.scala
```diff
@@ -169,30 +165,35 @@ object ZipIO extends LazyLogging {
...
     val commonPrefix = if (truncateCommonPrefix) {
       val commonPrefixNotFixed = PathUtils.commonPrefix(zipEntries.map(e => Paths.get(e.getName)))
-      val strippedPaths = excludeFromPrefix.getOrElse(List()).flatMap(stripPathFrom(commonPrefixNotFixed, _))
-      strippedPaths.headOption match {
-        case Some(path) => path
-        case None       => commonPrefixNotFixed
+      val strippedPrefix = stripPathFrom(commonPrefixNotFixed, excludeFromPrefix.getOrElse(List.empty))
+      // if only one file is in the zip and no layer name is given as prefix, do not remove file name as common prefix
+      zipEntries match {
+        case head :: tl if tl.isEmpty && strippedPrefix.endsWith(Paths.get(head.getName).getFileName) =>
```
fm3, line 194: Maybe this logic can be extracted into a function? I think it’s fairly hard to read

util/src/main/scala/com/scalableminds/util/io/ZipIO.scala
```diff
@@ -169,30 +165,35 @@ object ZipIO extends LazyLogging {
...
+        case -1 => path
+        // subpath(0, 0) is forbidden, therefore we handle this special case ourselves
+        case 0 => Paths.get("")
+        case i => path.subpath(0, i)
+      }
     }
 
     val zipEntries = zip.entries.asScala.filter(e => !e.isDirectory && (includeHiddenFiles || !isFileHidden(e))).toList
 
     //color, mask, segmentation are the values for dataSet layer categories and a folder only has one category
```
fm3, line 188: While we’re here, could you remove this comment? I don’t see what it’s good for in this place

webknossos-datastore/app/com/scalableminds/webknossos/datastore/services/UploadService.scala
fm3: Cool that you looked into that already :) I think atually deleting is indeed too aggressive, as we may not be 100% certain that this is never wrongly triggered. I’d suggest moving to `.trash` instead (but still double-checking that this really only happens in the error case).
The other ideas sound good to me too, but, as you say, let’s do all of that in a separate PR

util/src/main/scala/com/scalableminds/util/io/ZipIO.scala
```diff
@@ -169,30 +165,35 @@ object ZipIO extends LazyLogging {
...
-        else if (path.subpath(i, i + 1).toString.contains(string)) //sample paths: "test/test/color_1", "segmentation"
-          if (i == 0) // path.subpath(0,0) is invalid and means that there is no commonPrefix which isn't started with a layer
-            Some(Paths.get(""))
-          else // this is the common prefix but stripped from the layer
-            Some(path.subpath(0, i))
-        else
-          stripPathHelper(i + 1)
-
-      stripPathHelper(0)
+    def stripPathFrom(path: Path, excludeFromPrefix: List[String]): Path = {
```
fm3, line 168: Maybe the function could have a more descriptive name now

fm3 (commented):
Mostly LGTM, please see my minor comments

fm3 (approved):

PR #4984, opened 2020-12-07
Allow Users to Join Multiple Organizations

### LINKED ISSUE #4761: Global users on webknossos.org
I think the current organization and user account structure is still confusing for new users. I mean that user accounts are tied to organizations, but you can't have the same email address in multiple organizations and you cannot easily share data/annotations across organizations.

I think a global user model such as Github has, would be clearer: People sign up for a user account on webknossos.org, can create new organizations, and invite others to join their organization. This would be a model that most users are familier with. Perhaps we could utilize the team capabilities of wk for that?

@fm3 @philippotto @hotzenklotz  thoughts?

fm3:
Users can now join multiple organizations, admins can now invite users by email address, skipping the manual user activation step.

### TODO Backend
 - [x] move email calls to multiuser
 - [x] only superusers should be able to change emails if multiuser has more than one identity
 - [x] superuser switch route (by email) needs multiuser disambiguation
 - [x] new user switch route by orga
 - [x] organization read access query: allow multiuser indirection
 - [x] set lastLoggedInIdentity upon login and upon switch
 - [x] use lastLoggedInIdentity upon login
 - [x] superusers should see all orgas
 - [x] forbid switching to deactivated users
 - [x] user switch route should lazily create user for superusers
 - [x] invite table in schema, invite case class + service + DAO
 - [x] create + save invite tokens
 - [x] add route to get organization name for invite token
 - [x] assert valid invite tokens in join request
 - [x] invalidate invite tokens after they are used to join
 - [x] also allow *registering* by invite (make sure logic doesn’t break single-orga behavior)
 - [x] send new-user notification upon user join to orga admin
 - [x] send invite emails (remove link from stdout)
 - [x] double-check that change password still works (both reset and change while logged in)
 - [x] repair createOrganizationWithAdmin request (create multi-user, handle loginInfo correctly)
 - [x] add sql constraint for users that (_multiUser, _organization) should be unique
 - [x] make GET routes POST where appropriate
 - [x] db migration
   - [x] adapt schema
   - [x] insert multiusers for each user
   - [x] connect the multiuser references
   - [x] migrate tokens (replace user email with user id)
   - [x] reversion?
 - [x] adapt e2e-test db
 - [x] re-enable unused-warnings in build.sbt
 - [x] switch back `isDemoInstance` default in application.conf to `false`
 - [x] switch back `email.logToStdout` default in application.conf to `false`

### TODO Frontend
 - [x] in user menu (top right of screen), offer to switch to other organizations
   - [x] also make it a bit more obvious that there are multiple organizations
 - [x] change invite UI, no more link, instead, enter recipients, add autoActivate checkbox (default false), send to backend
 - [x] new view `/invite/<tokenValue>`
   - [x] if logged in, show button to join organization with currently active account.
   - [x] if logged out, enforce authentication (either, via registration which should send the `inviteToken` as additional form field) or by logging in (should do an explicit join afterwards)

### New Routes
 - `GET api/auth/joinOrganization/:inviteToken` (allows the user to switchOrganization afterwards)
 - `GET api/auth/switchOrganization/:organizationName` (exchanges user cookie, rediercts to dashboard)
 - `GET api/organizations` (list all you can switch to)
 - `GET api/organizations/byInvite/:inviteToken` (list the organization you have been invited to)
 - `POST api/auth/sendInvites` expects something like `{recipients: ["a@b.com”, "x@y.com"], autoActivate: false}`, for development now logs the invite token to stdout

### Changed Routes
 - `api/auth/register` has new optional form field `inviteToken` (it is required for instances with >1 orga). It uses `organizationName` only if there is no token (to support legacy invite links for the moment) and does not use `organizationDisplayName` anymore. (Those are still used in `api/auth/createOrganizationWithAdmin` though)

### Steps to test:
Hint: set `mail.logToStdout = true` in application.conf to see email contents.

- as demo instance (`features.isDemoInstance = true`) 
  -  create a few organizations
  - invite user 
  - as logged-in user of another org, click invite link, join
  - as logged-in user in that org, click invite link, try to join, should warn that you are already in that org
  - as logged-out user go to invite link, should offer option to register, or log in and then join
  - register via invite link should get you into the matching organization
  - as a user with multiple orgs, use switch menu
  - as superuser, use switch menu, should see all orgs, should be able to switch to all orgs
- as single-org instance
  - invite user
  - register upon invite should land you in that org
  - register without invite should also land you in that org

### Issues:
- fixes #4761 

------
- [x] Ready for review


frontend/javascripts/oxalis/view/right-menu/abstract_tree_tab_view.js
```diff
@@ -16,6 +16,7 @@ import AbstractTreeRenderer, {
 import window from "libs/window";
 
 type OwnProps = {|
+  // eslint-disable-next-line react/no-unused-prop-types
```
philippotto, line 19: eslint got a bit smarter at some spots, but occasionally the smartness also backfired since it didn't catch some usages. the `portalKey` is an extreme example which we use in the golden layout adapter.

tools/test.sh
```diff
@@ -21,7 +20,7 @@ function prepare {
   pbjs -t json "webknossos-tracingstore/proto/SkeletonTracing.proto" > "$testBundlePath/SkeletonTracing.proto.json"
   pbjs -t json "webknossos-tracingstore/proto/VolumeTracing.proto" > "$testBundlePath/VolumeTracing.proto.json"
   # --copy-files will copy files that are present in the source dir but are not transpiled (e.g.: json files)
-  BABEL_ENV=test babel "$jsPath" --out-dir "$testBundlePath" --copy-files $additionalParams
+  BABEL_ENV=test babel "$jsPath" --out-dir "$testBundlePath" --copy-files
```
philippotto, line 23: suddenly, babel complained about the `--timeout` option which was passed to it. the args to this bash script are passed to `prepare` and ava. I'm not aware of any specific args which `babel` needs, which is why this should work alright.

philippotto (commented):
Ok, I think, I'm done with the front-end :tada:

If a user is member of multiple orgs, the avatar changes to:
image

I hope this is a good balance of drawing some attention to the multi-users-aspect, but not too much.


One other note:
I updated the minor version of antd which caused webpack to fail. After trying numerous things, I got it to work by deleting the entire `yarn.lock` file. I guess, that some of antds internal dependencies weren't updated properly and ignoring our existing `yarn.lock` fixed it consequently. Usually, we don't update the entire `yarn.lock` file, but I hope this is fine for this PR.

frontend/javascripts/libs/user_local_storage.js
```diff
@@ -4,7 +4,7 @@ import Store from "oxalis/store";
 
 function prefixKey(key) {
   const { activeUser } = Store.getState();
-  const prefix = !activeUser ? "Anonymous" : activeUser.email;
+  const prefix = !activeUser ? "Anonymous" : `${activeUser.email}-${activeUser.organization}`;
```
daniel-wer, line 7: I did a quick check what functionality this would affect as it basically "removes" all items stored in a user's local storage. Most things are not a big deal (dataset cache, last finished task, last search tags). However, the tracing view layouts are also saved in local storage (`layout_persistence.js`) and it could be annoying to users if everything is reset. It might make sense to add a backward compatibility bridge for the `prefixKey` in `getItem` (where the `activeUser.email` prefix is tried if the new prefix didn't turn up a result)  for a while, which can then later be removed. This should be opt-in as it shouldn't happen for the dataset cache, for example. What do you think?

daniel-wer (commented):
Very nice work! Please see my initial comments. I'll add my observations from testing later today :)

frontend/javascripts/admin/auth/registration_view.js
```diff
@@ -21,7 +26,9 @@ function RegistrationView({ organizationName }: Props) {
     (async () => {
       setIsLoading(true);
       try {
-        if (organizationName != null) {
+        if (inviteToken != null) {
+          setOrganization(await getOrganizationByInvite(inviteToken));
```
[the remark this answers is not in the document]
philippotto: Good catch! I just noticed that this view doesn't even need the `inviteToken`. This was a leftover from a point where I thought, that the registration view would be used if a user follows an invitation and registers. However, the PR now implements this by showing the accept-invite-view which then is able to open the authentication-modal with the invite-token (which then shows the `RegistrationForm` which is not the same as `RegistrationView` :trollface:).
To come back to your question: the accept-invite-view is the only high-level component which receives the token (from the URL) and that component validates the token, too.

frontend/javascripts/libs/user_local_storage.js
philippotto: You are right, it would be nice to keep this backwards compatible. I think, one could even argue that the layouts should be stored across the user's organizations. So, getItem/setItem could receive an `isOrganizationSpecific`-setting which is true by default and set to false for the layout stuff. That way, this wouldn't even be a compatibility bridge. What do you think? :)

daniel-wer: That's even more elegant, great idea :+1:

philippotto: Done!

daniel-wer (commented):
My minor observations from testing, everything worked flawlessly otherwise:
- Visiting an invite link again, after it was used before, shows the correct message on screen but also a generic error toast: ("Couldn't find the requested resource.")
- I wonder if we could make it a little more visible which organization a user has selected. Maybe the background color of the circle could be organization-dependent. This could allow users to easily spot that they have the wrong organization activated if they switch a lot.

Due to the complexity of this change it would probably be best, if someone else would test this as well :)

app/controllers/OrganizationController.scala
```diff
@@ -32,20 +39,37 @@ class OrganizationController @Inject()(organizationDAO: OrganizationDAO,
...
+  }
+
+  def getDefault: Action[AnyContent] = Action.async { implicit request =>
     for {
       allOrgs <- organizationDAO.findAll(GlobalAccessContext) ?~> "organization.list.failed"
       org <- allOrgs.headOption.toFox ?~> "organization.list.failed"
       js <- organizationService.publicWrites(org)(GlobalAccessContext)
     } yield {
-      if (allOrgs.length > 1) // Cannot list organizations if there are multiple ones due to privacy reasons
+      if (allOrgs.length > 1) // Cannot list organizations publicly if there are multiple ones, due to privacy reasons
```
youri-k, line 55: IntelliJ wants me to replace the `.length > 1` call with `allOrgs.headOption.isDefined && allOrgs.tail.nonEmpty`. I don't know the real runtime impact.

app/models/user/MultiUser.scala
```diff
@@ -0,0 +1,94 @@
...
+import javax.inject.Inject
+import slick.jdbc.PostgresProfile.api._
+import com.scalableminds.webknossos.schema.Tables._
+import slick.lifted.Rep
+import utils.{ObjectId, SQLClient, SQLDAO}
+
+import scala.concurrent.ExecutionContext
+
+case class MultiUser(
+    _id: ObjectId,
```
youri-k, line 15: Is there a reason, you did not put the first- and lastname into the multi user?

app/oxalis/security/WebknossosBearerTokenAuthenticatorService.scala
```diff
@@ -76,7 +77,8 @@ class WebknossosBearerTokenAuthenticatorService(settings: BearerTokenAuthenticat
     for {
       tokenAuthenticator <- repository.findOneByValue(tokenValue) ?~> "auth.invalidToken"
       _ <- bool2Fox(tokenAuthenticator.isValid) ?~> "auth.invalidToken"
-      user <- userService.findOneByEmail(tokenAuthenticator.loginInfo.providerKey)
+      idValidated <- ObjectId.parse(tokenAuthenticator.loginInfo.providerKey) ?~> "auth.tokenNoId" // TODO readable message, warning that token may be old format (or support old format with disambiguation?)
```
youri-k, line 80: TODO reminder

youri-k (commented):
Wow, great stuff! :tada: 

I did not test all functions yet.

philippotto:
> Visiting an invite link again, after it was used before, shows the correct message on screen but also a generic error toast: ("Couldn't find the requested resource.")

Good point! Just pushed a fix for this.

> I wonder if we could make it a little more visible which organization a user has selected. Maybe the background color of the circle could be organization-dependent. This could allow users to easily spot that they have the wrong organization activated if they switch a lot.

Cool idea, but I'm a bit hesitant about the actual usefulness of differently colored icons. First, I think that most users will stay in their "main" organization (a big reason for why we built this was that some users got into the wrong organization and switching was very complicated) and thus don't need to identify the current organization using a color. Second, I think the mapping from color to actual organization is quite hard for the user. There might be power users who switch often and then can identify the color well, too, but I feel like there should be a better way to do this. For example, one could show the full (or abbreviated) organization name in the navbar somewhere. However, getting this right (UI-wise) will be a bit tricky which is why I'd like to postpone this. Is this okay?

daniel-wer:
> However, getting this right (UI-wise) will be a bit tricky which is why I'd like to postpone this. Is this okay?

Yes fine for me :+1:

app/oxalis/security/WebknossosBearerTokenAuthenticatorService.scala
fm3: Fortunately we found a way to migrate the legacy tokens (they now point to the user id instead of the email) so no legacy-specific error message is required :) I’ll just use the normal error message here too.

app/models/user/MultiUser.scala
fm3: Yes, though I’m not 100% convinced myself. There may be situations where a user wants to have specific names depending on the organization (e.g. not share their full name with one). It also made for the smaller change right now. We may want to change this later, though. Or if you think this doesn’t make sense, I’ll also change it here.

app/controllers/OrganizationController.scala
fm3: Hmm, I think this sacrifices readability for performance.
In case of very long lists, this definitely makes sense (do not count millions of links just to compare the result with 1), but we expect allOrgs to stay fairly small. So I’d go for the more readable version (also not chosing `lengthCompare`). Also, if there is a sharp increase in org counts, the place to performance-optimize first is above, by not getting all of them from the db.

app/controllers/Authentication.scala
```diff
@@ -232,90 +247,126 @@ class Authentication @Inject()(actorSystem: ActorSystem,
...
       }
     )
   }
 
-  def autoLogin = Action.async { implicit request =>
+  def switchMultiUser(email: String): Action[AnyContent] = sil.SecuredAction.async { implicit request =>
+    implicit val ctx: GlobalAccessContext.type = GlobalAccessContext
+    for {
+      requestingMultiUser <- multiUserDAO.findOne(request.identity._multiUser)
+      _ <- bool2Fox(requestingMultiUser.isSuperUser) ?~> Messages("user.notAuthorised") ~> FORBIDDEN
```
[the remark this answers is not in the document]
fm3: Yes, (1) It is very unlikely to happen, usually an accident by one of us using it while being logged in with the wrong account. So no security breached there, we never used this logging to debug anything. 
And (2) it outputs entered emails to logfiles, which may become a privacy leak.

app/controllers/Authentication.scala
```diff
@@ -232,90 +247,126 @@ class Authentication @Inject()(actorSystem: ActorSystem,
...
     signInForm.bindFromRequest.fold(
       bogusForm => Future.successful(BadRequest(bogusForm.toString)),
       signInData => {
         val email = signInData.email.toLowerCase
-        val credentials = Credentials(email, signInData.password)
-        credentialsProvider
-          .authenticate(credentials)
+        val userFopt: Future[Option[User]] =
+          userService.userFromMultiUserEmail(email)(GlobalAccessContext).futureBox.map(_.toOption)
+        val idF = userFopt.map(userOpt => userOpt.map(_._id.id).getOrElse("")) // do not fail here if there is no user for email. Fail below.
```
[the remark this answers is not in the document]
fm3: I chose to not fail while fetching the user here, to leave all the failing and error handling code to the below case when the userService cannot retrieve the user. This way, error handling is not duplicated. Not a strong preference, but I think this is slightly cleaner (even though it required an explaining comment).

Concerning the for-yield, I did not so much like nesting these, I feld the code became harder to read then in this case.

philippotto:
> @philippotto if you have a minute could you have a look at the conflict in yarn.lock? (comes from #5003 )

Sure, I just pushed the resolved conflict. For the record, one can simply run `yarn install` and the conflict in the lock file will be resolved automatically :)

fm3:
@youri-k Thanks for your feedback, I implemented most of it :) Could you have another look?

youri-k (approved):
I tested the described steps and everything worked flawlessly :tada: 

See my small code comments

PR #4986, opened 2020-12-07
Fix that largest segment id is never bumped from 0 to 1

### LINKED ISSUE #4937: volume annotation’s largest segment id is never bumped from 0 to 1

daniel-wer:
See [this comment](https://github.com/scalableminds/webknossos/issues/4937#issuecomment-738130505) for an in-depth description of the bug and fix. I've opted to use the `FINISH_ANNOTATION_STROKE` action to signify that voxels were annotated as that is the same action that should always be triggered after a volume annotation to make undo/redo work. It's not checked whether voxels were actually annotated (in the copySegmentationLayer case the copy could be empty), but in my opinion a stricter check wouldn't really have a benefit and complicate things.
I've added integration tests to check that the `maxCellId` is updated whenever voxels are annotated. Are there other cases than floodfill, copyLayer, brush/trace? For undo the `activeCellId` and `maxCellId` are not undone, but that "problem" existed before and doesn't break this fix as far as I can see.

- https://___.webknossos.xyz

### Steps to test:
- Check this branch out locally. In the `save_saga.js` search for `compress: true` and change it to `compress: false`.
- Open a volume tracing without a fallback layer. Open the network tab. Make a single brush stroke and save afterwards. You should see an `updateTracing` update action in the request to `.../update` which contains a `largestSegmentId` of 1.

### Issues:
- fixes #4937 

------
- [x] Ready for review

### DIFF
--- a/frontend/javascripts/oxalis/model/reducers/volumetracing_reducer_helpers.js
+++ b/frontend/javascripts/oxalis/model/reducers/volumetracing_reducer_helpers.js
@@ -46,7 +46,7 @@ export function setActiveCellReducer(state: OxalisState, volumeTracing: VolumeTr
   });
 }
 
-export function createCellReducer(state: OxalisState, volumeTracing: VolumeTracing, id: ?number) {
+export function createCellReducer(state: OxalisState, volumeTracing: VolumeTracing, id?: number) {
   if (id === 0) {
     // cellId 0 means there is no annotation, so there must not be a cell with id 0
     return state;
@@ -59,17 +59,27 @@ export function createCellReducer(state: OxalisState, volumeTracing: VolumeTraci
     id = Math.max(activeCellId, maxCellId) + 1;
   }
 
-  // Create the new VolumeCell
-  const cell: VolumeCell = { id };
+  if (volumeTracing.cells[id] == null) {
+    // Create the new VolumeCell
+    const cell: VolumeCell = { id };
 
-  return update(state, {
-    tracing: {
-      volume: {
-        activeCellId: { $set: cell.id },
-        cells: { [cell.id]: { $set: cell } },
+    return update(state, {
+      tracing: {
+        volume: {
+          activeCellId: { $set: id },
+          cells: { [id]: { $set: cell } },
+        },
       },
-    },
-  });
+    });
+  } else {
+    return update(state, {
+      tracing: {
+        volume: {
+          activeCellId: { $set: id },
+        },
+      },
+    });
+  }
 }
 
 export function updateDirectionReducer(
--- a/frontend/javascripts/oxalis/model/reducers/volumetracing_reducer_helpers.js
+++ b/frontend/javascripts/oxalis/model/reducers/volumetracing_reducer_helpers.js
@@ -51,12 +51,12 @@ export function createCellReducer(state: OxalisState, volumeTracing: VolumeTraci
     // cellId 0 means there is no annotation, so there must not be a cell with id 0
     return state;
   }
-  let newMaxCellId = volumeTracing.maxCellId;
+
+  // The maxCellId is only updated if a voxel using that id was annotated. Therefore, it can happen
+  // that the activeCellId is larger than the maxCellId. Choose the larger of the two ids and increase it by one.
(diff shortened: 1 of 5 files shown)


frontend/javascripts/oxalis/model/reducers/volumetracing_reducer_helpers.js
```diff
@@ -51,12 +51,12 @@ export function createCellReducer(state: OxalisState, volumeTracing: VolumeTraci
...
-  let newMaxCellId = volumeTracing.maxCellId;
+
+  // The maxCellId is only updated if a voxel using that id was annotated. Therefore, it can happen
+  // that the activeCellId is larger than the maxCellId. Choose the larger of the two ids and increase it by one.
+  const { activeCellId, maxCellId } = volumeTracing;
   if (id == null) {
-    newMaxCellId++;
-    id = newMaxCellId;
-  } else {
-    newMaxCellId = Math.max(id, newMaxCellId);
```
philippotto, line 59: This line change raised some questions for me which is why I digged a bit through the whole `cells` code. I have some observations which I cannot really get straight in my head:

- apparently, we have an entire `cells` mapping in the store (I wasn't really aware of that :see_no_evil:). however, this seems to be mostly unused (it's only maintained as far as I can see). for each cell, this mapping only stores its id, so there is no real valuable information there. BUT: we could probably easily build upon that (e.g., to add a position to each cell) which would be quite cool at some point.
- right now, a new cell is created as soon as its ID is made the `activeCellId`. in my opinion, this is a bit conflicting the new semantics of this PR which only considers a new cell id once it was also labeled. I think it would be more intuitive if the maximum cell ID would always equal the maximum key id in `cells`. however, it's probably not really crucial to have this constraint (and also might be hard to implement since the creation should ideally happen before something is labeled).
- passing an explicit cell id to `createCell` seems a bit weird to me. as far as I can see, there are only two places where this is used: in `setActiveCellReducer` and in the tests. For `setActiveCellReducer`, the use case makes sense. however, the `cellId` parameter in the redux `createCell`-action should probably never really be used. so, maybe we should just remove this from the action?
- if a not-null id parameter is only used in `setActiveCellReducer`, the following point is probably irrelevant. however, considered in isolation, this specific line change seems a bit dangerous to me. if this function is called with a cell id for which there is already an object in the store, the function would create a new `cell` object and replace it. to avoid this, this reducer could simply check whether the cell already exists and then do nothing (would be a bit redundant to what `setActiveCellReducer` does right now, but theoretically other callers could be added in the future).

Sorry for this wall of text. I just wanted to share what went through my head. We could also have a call about this, in case it's easier. My tldr-suggestions would be probably be: remove the `cellId` parameter from the `createCell` action & don't do anything in `createCellReducer` if the id already exists. Creating the cell object not-lazily is probably alright, as that structure is not stored on the server anyway (at least for now).

philippotto (commented):
Great fix and clean up :+1: I just tested it and it works beautifully (I was able to set a breakpoint at the right spot to inspect the save queue before the compression was performed so I didn't have to check it out locally).

See my one comment which contains some thoughts and questions which we should clarify before merging this PR :)

frontend/javascripts/oxalis/model/reducers/volumetracing_reducer_helpers.js
daniel-wer: > Sorry for this wall of text. I just wanted to share what went through my head. We could also have a call about this, in case it's easier. My tldr-suggestions would be probably be: remove the cellId parameter from the createCell action & don't do anything in createCellReducer if the id already exists. Creating the cell object not-lazily is probably alright, as that structure is not stored on the server anyway (at least for now).

Thank you for taking the time to think this through! I fully agree with your conclusions and should have adapted the code accordingly.

philippotto (approved):
Excellent! Also kudos for adding and adapting the test suite :man_singer:

PR #4993, opened 2020-12-16
Address Unused Warnings

fm3:
First batch of backend’s unused warnings should be gone.

Some remain (mostly in routes files, html templates, and the DBAccessContext), will be addressed in a later PR

------
- [x] Ready for review


youri-k (approved):
LGTM :shipit:
I did not go through all the changes, but the CI should catch if something does not work as expeceted, right?
Is there anything that I should take another look at?

PR #5003, opened 2020-12-18
Upgrade Saxophone XML Parser to  0.7.1

### LINKED ISSUE #4605: Frontend NML parser doesn't accept ">" in attributes
We just encountered the following nml which was produced by NmlWriter.scala. The `>` in the `description` attribute caused the NML to be rejected by the frontend NML parser. Backend parser accepted it.

```
<things>
  <meta name="writer" content="NmlWriter.scala" />
  ...
  <parameters>
    <experiment name="..." organization="..." description="&lt;no description>" />
```

fm3:
- https://upgradesaxophone.webknossos.xyz

### Steps to test:
- drop a normal NML into tracing view, should still work
- drop in an NML with `>` in an attribute, like `description="a > b"` should also work (fix https://github.com/matteodelabre/saxophone/issues/14 )

### Issues:
- fixes #4605

------
- [x] Ready for review


philippotto:
Should work now :) I had to adapt the webpack config so that hierarchical package dependencies are properly handled. Since this change could have subtle effects, I'd suggest to merge this in January.

daniel-wer (approved):
Awesome, works like a charm, Also, thanks again @philippotto for the debugging marathon and the fix for bundling multiple package versions! :pray:

Issue #5000, opened 2020-12-18
Forced garbage collection of buckets


philippotto:
Airbrake still shows some `Error:  Unhandled Rejection: "Error: Unexpected state: UNREQUESTED"` errors (https://scm.airbrake.io/projects/95438/groups/2737723584692415595?tab=notice-detail). See #4895 where we also noticed this. The latter issue was closed, since we identified one major culprit for these messages. However, I think, I found another culprit.

When looking at the logs in Airbrake, one often sees that this message correlates with the UNREQUESTED-error:
`A bucket was forcefully garbage-collected. This indicates that too many buckets are currently in RAM.`

We should find out why this happens and how to solve the root problem (namely that too many buckets are dealt with). And/or one could improve the handling of this situation. Currently, it's probably something like this:

- a bucket is requested (e.g., for rendering)
- lots of other buckets are also allocated (not exactly clear why)
- for these other buckets, not enough slots are available. therefore, the very first bucket is forcefully collected
- now, the request for the first bucket finishes. the bucket is re-created again and `receiveData` is called. however, the bucket is in the unrequested-state now.

My guess for the reason of the "too many buckets are currently in RAM"-symptom is one of the following:
- ~rapid movement to another position (or rapidly zooming), so that the old requests are basically outdated, but were already initiated. canceling these requests would be helpful, but also a bit difficult.~
  - hm, I just checked the pullqueue code and after the bucket buffer is received, `getBucket` (not `getOrCreateBucket`) is called. if the bucket doesn't exist, the buffer is ignored. so, my theory doesn't really add up.
- upsampling of volume data or other mechanisms which process buckets independent of the rendering pipeline
  - however, the error also occurs for skeleton-only data, so there has to be another reason
- ???

@daniel-wer do you have thoughts on this?

daniel-wer:
@philippotto I had a brief look at this as well, but haven't found a possible explanation yet. I also wasn't able to reproduce these issues.

As you already noted there seem to be two issues: (1) Too many buckets are allocated **and marked as not collectable** (meaning they are either accessed in the last render pass, dirty or currently requested). (2) `Unexpected state: UNREQUESTED` probably, because buckets were forcefully collected.

[...]

PR #5010, opened 2020-12-23
Ensure that bucket picker respect RAM-limit for buckets

philippotto:
If the GPU is able to allocate more than 5000 buckets for a layer, the bucket picker code could theoretically cause that more than 5000 buckets were fetched by the pullqueue. However, each `DataCube` has a hard limit of 5000 which meant that each new bucket kicked another bucket out of the `DataCube`. This PR makes the bucket picker respect the hard limit of 5000 and also reports a proper error to airbrake.

I temporarily reduced the hard limit from 5000 to 100 and checked that the change really improves things. Indeed, before the fix, the rendered output was very spotty (lots of randomly distributed black buckets). After the fix, this was way better.
However, I still noticed some "forced bucket collection" messages. From my understanding, the prefetching is at fault here, since it doesn't respect any bucket limits at all. I think, there is no easy fix for this. Instead, we probably should refactor the entire logic around which buckets are pulled / kept in RAM / kept on GPU so that all limits can be respected correctly.

For now, this fix should help a bit (hopefully) and we can observe airbrake to see whether the other aspect is really happening (or if it's only a theoretical problem (or if it's the main problem)).

- https://___.webknossos.xyz

### Steps to test:
- see explanation above

### Issues:
- contributes to #5000

------
- [X] Ready for review


frontend/javascripts/oxalis/model/bucket_data_handling/data_cube.js
```diff
@@ -259,17 +260,22 @@ class DataCube {
 
   addBucketToGarbageCollection(bucket: DataBucket): void {
     if (this.bucketCount >= this.MAXIMUM_BUCKET_COUNT) {
-      for (let i = 0; i < 2 * this.bucketCount; i++) {
```
philippotto, line 262: I think, earlier versions of `shouldCollect()` had side effects which is why this code iterated twice over all buckets. However, now, the method is sideeffect-free which is why a single iteration should be enough.

frontend/javascripts/oxalis/constants.js
```diff
@@ -178,6 +178,9 @@ const Constants = {
 
   MIN_TREE_ID: 1,
   MIN_NODE_ID: 1,
+
+  // Maximum of how many buckets will be held in RAM (per layer)
+  MAXIMUM_BUCKET_COUNT_PER_LAYER: 100,
```
philippotto, line 183: todo: revert back to 5000

.circleci/not-on-master.sh
```diff
@@ -1,8 +1,8 @@
 #!/usr/bin/env bash
 set -Eeuo pipefail
 
-if [ "${CIRCLE_BRANCH}" == "master" ]; then
+# if [ "${CIRCLE_BRANCH}" == "master" ]; then
```
daniel-wer, line 4: Reminder to revert this as well.

daniel-wer (commented):
Thanks for looking into this, LGTM!

> However, I still noticed some "forced bucket collection" messages. From my understanding, the prefetching is at fault here, since it doesn't respect any bucket limits at all. I think, there is no easy fix for this. Instead, we probably should refactor the entire logic around which buckets are pulled / kept in RAM / kept on GPU so that all limits can be respected correctly.

Agreed, the additional error logging in this PR should help us decide on the priority of that :) Another possibility would be to factor in a certain percentage of prefetched buckets in the calculation in `getBucketCapacity`. For example, the method could return only 0.75 times the theoretical capacity to allow space for prefetched buckets. I'm not sure whether such a static percentage exists and is robust, though.

daniel-wer (approved):
:+1:

philippotto:
> Another possibility would be to factor in a certain percentage of prefetched buckets in the calculation in getBucketCapacity. For example, the method could return only 0.75 times the theoretical capacity to allow space for prefetched buckets. I'm not sure whether such a static percentage exists and is robust, though.

Yes, that could be an easy workaround :+1:

PR #5016, opened 2021-01-04
Improve hint in version view

### LINKED ISSUE #4883: Improve potentially misleading "NML upload" entry in version view
The version view (Settings Dropdown - Restore older version) shows an entry like "Uploaded an NML with x nodes." when one update action batch contains more than one `createNode` update actions as that used to be the only action that could lead to this. However, in the mean time this could've also been caused by an Undo (for example, of the deletion of multiple trees) or Redo action. The error message should be generalized and say something like "Added x trees with y nodes." or so.

Todo in the code: https://github.com/scalableminds/webknossos/blob/43bb9277517a1f74f00fa12ba8507f89c05f5202/frontend/javascripts/oxalis/view/version_entry.js#L160

philippotto:
- https://___.webknossos.xyz

### Steps to test:
- I created a tree, deleted a tree and reverted that via undo
- the version view looks like this:
image

### Issues:
- fixes #4883

------
- [X] Ready for review


daniel-wer (approved):
Nice, LGTM :+1:

PR #5017, opened 2021-01-05
Disable brush in mags higher than mag8

### LINKED ISSUE #4967: Brushing in high mags (e.g., > mag 16) is slow
When being in mag 16-16-16, even a very small brush size will cause lots of buckets to be labeled (also thanks to the continuous brushing we do). One bucket in mag 16, corresponds to 4096 buckets in mag 1. All of these have to be available in RAM which is also a memory problem (in addition to speed).

I'm afraid we need to (re-)introduce some limit for brushing in high mags. Maybe similar to the trace tool. We still have the advantage of being able to trace in low mags and then zoom out, though.

Related to this: We might want to expose a way to "remove" low mags. E.g., if I want to annotate nuclei, I'm probably not interested in maintaining mag 1.

philippotto:
- https://disablebrushinhighmags.webknossos.xyz

### Steps to test:
- open a tracing, switch to brush tool
- ensure that brush is usable in mags 1 to 8
- the brush tool should get disabled (and another tool should be activated) automatically as soon as one zooms to an unusable mag

### Issues:
- fixes #4967

------
- [x] Ready for review


frontend/javascripts/oxalis/view/action-bar/volume_actions_view.js
```diff
@@ -180,7 +183,7 @@ const getExplanationForDisabledVolume = (
     return "Volume annotation is disabled while the merger mode is active.";
   }
 
-  if (!isLabelingPossible) {
+  if (!isSegmentationVisibleForMag) {
```
daniel-wer, line 186: One thing I noticed during testing is that this boolean is also false if the segmentation layer is simply disabled (for example, using `3`). Could you adapt the message accordingly?

daniel-wer (approved):
Works like a charm, only two minor non-blocking suggestions :)

frontend/javascripts/oxalis/view/action-bar/volume_actions_view.js
philippotto: Good catch! Since the existing logic around the disabled tool tooltip (i.e., which text to show) was a bit convoluted, I refactored the corresponding code which made it easier to integrate the case you mentioned. Also, the automatic switching of the tool (if the current one gets disabled), got a bit DRYer due to the refactoring.
Feel free to have another look!

PR #5030, opened 2021-01-06
Fix that agglomerate skeleton button was clickable in view or read-only mode

daniel-wer:
See title. I had to add a rather ugly workaround for an antd tooltip bug, but I wanted to show the disabled button and tooltip in view and read-only mode so that users understand why they cannot import skeletons.

- https://___.webknossos.xyz

### Steps to test:
- Open a dataset with an agglomerate file in View mode. Activate and select an agglomerate file mapping. -> The “Import Skeleton for Centered Cell” button should be disabled and a tooltip should be shown.
- The same should happen in a hybrid tracing if the Version View is opened (which switches the tracing to read-only mode).
- In other cases, the button should still be enabled and work as expected.

### Issues:
- see https://discuss.webknossos.org/t/add-import-skeleton-from-agglomerate-file/1623/4

------
- [x] Ready for review

### DIFF
--- a/frontend/javascripts/oxalis/view/right-menu/mapping_info_view.js
+++ b/frontend/javascripts/oxalis/view/right-menu/mapping_info_view.js
@@ -57,6 +57,7 @@ type StateProps = {|
   activeViewport: OrthoView,
   activeCellId: number,
   isMergerModeEnabled: boolean,
+  allowUpdate: boolean,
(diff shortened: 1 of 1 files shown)


philippotto (approved):
Works like a charm :sparkles: Also :+1: for finding the workaround.

Issue #5034, opened 2021-01-07
Document backend log level configuration


fm3:
logback.xml configs can be used to enable or disable different log-levels for different parts of the backend. We should check the documentation and compile a step-by-step guide for how this works in the wk backend.

Issue #5032, opened 2021-01-07
Iterate on agglomerate skeleton edge limit


fm3:
Let’s have a look at the logs and remove or increase the limit if it seems feasible.

fm3:
Turns out the skeleton generation is fast enough, increased the limit to 100k in production now. Even so, the slowest request was 630ms (73k edges, 47k nodes). I’d recommend against removing the limit entirely or increasing it much more, because the generated skeletons will be saved in the fossildb and create additional load there. Let’s see if users will ask for even higher limits in the future.

PR #5035, opened 2021-01-07
change debug logging to info for agglomerate skeleton perf

fm3:
Stand-alone datastores currently omit DEBUG logging, which I had accidentally used for this particular output.
Since #5034 is still open, the easiest change seemed to change the line in question.

Contributes to #5032


youri-k (approved):
LGTM :shipit:

PR #5044, opened 2021-01-12
dont throw initialization error when logged out

### LINKED ISSUE #4731: Opening dataset while logged-out: Initialization Error [Object object]
The intended error message from the back-end shows also: “Dataset xyz does not exist or could not be accessed. You may need to log in.” But additonally (and this one sticky) we get the initialization error toast. I suspect that is a front-end bug

grittaweisheit:
This PR removes the "Initialization Error. ..." toast that popped up when trying to view a dataset without being logged in.

- https://___.webknossos.xyz

### Steps to test:
- copy a link to a dataset view
- don't log in / log out
- go to the copied link
- you should be moved to the login page with an error toast that says you can not access the dataset
- there should be no "Initialization Error. ..." toast.

### Issues:
- fixes #4731 

------
- [x] Ready for review


frontend/javascripts/oxalis/controller.js
```diff
@@ -113,8 +113,9 @@ class Controller extends React.PureComponent<PropsWithRouter, State> {
       .then(() => this.modelFetchDone())
       .catch(error => {
         this.props.setControllerStatus("failedLoading");
-        // Don't throw errors for errors already handled by the model.
-        if (error !== HANDLED_ERROR) {
+        // Don't throw errors for errors already handled by the model
+        // or when there is no logged in user since then the initialization fails anyways.
+        if (error !== HANDLED_ERROR && this.props.user != null) {
```
philippotto, line 118: Hmm, I'm not sure whether this is the best approach. Basically, this will always tell a user to log in, if an error happened while being not logged in. However, there could be an error which is unrelated to authorization.
For example, if I'm not registered and I view a publicly available dataset and that fails for some reason, I'll be asked to log in.  However, in this case we should show the actual error.

I think, a better approach would be to refine this if-statement à la:

```
const isAuthorizationError = ...;
const encourageLogIn = this.props.user == null && isAuthorizationError;
if (error !== HANDLED_ERROR && !encourageLogIn) {

}
```

The only remaining bit would be to find out how to set `isAuthorizationError`. I'd hope something like `error.statusCode` would be useful for this.

philippotto: Let me know in case we should discuss this in person!

philippotto:
Awesome! Works very well :+1: 

A thing which would also be good: if I'm logged in and open a "private" dataset, a toast encourages me to log in. If I do so, the dataset is immediately loaded (which is good), but the toast is still there (however, it's obsolete now).
Could you have a look whether it's easy to remove the toast after the log-in? If the dataset cannot be loaded again a new toast will appear, anyway. So, the old toast can definitely be removed after log-in.
The `Toast` module provides a `close` function which takes a `key`. If the same key is passed when creating the initial toast (can be a hardcoded value, such as `error-while-loading-dataset-toast`), the closing should work.

grittaweisheit:
I couldn't think of a straight forward way to solve this, because the error toast that prompts you to log in is automatically created from a request response message. This makes it hard to identify specific errors. 
I instead attached the status as key to toasts generated that way and close 404 errors when logging in after being prompted to do so to view a dataset.
I couldn't think of a similarly easy solution.

philippotto (approved):
Excellent! The status-code-as-key solution looks very elegant to me :+1: Feel free to merge on Monday, but please also update the actual issue number in the PR template :)

PR #5047, opened 2021-01-13
Allow handling dataset configurations for datasets not normally visible

### LINKED ISSUE #5045: Accessing dataset via token: saving dataSetConfiguration fails
When logged in to orga A, but viewing dataset of orga B with a token, the `PUT dataSetConfiguration` request fails

fm3:
- https://datasetconfglobal.webknossos.xyz

### Steps to test:
- create a second user not allowed to see a particular dataset
- as orga admin, copy sharing link from dataset settings page, open as that second user
- should be allowed to view and to change view settings in sidebar

### Issues:
- fixes #5045 

### On Privacy
- Privacy impact should be minimal, since dataset default view configurations do not contain any problematic information, and the requester needs to already know the dataset name (usually only happens when they received a sharing link anyway)

------
- [x] Ready for review


youri-k (approved):
Nice catch :tada:

PR #5050, opened 2021-01-15
Disable fill tool for higher mags

### LINKED ISSUE #4923: Fill-tool produces spotty segmentation in other mags
reproduction steps:
- zoom to mag 8-8-1
- use the fill tool on an empty plane --> produces rectangle
- zoom in --> some buckets are missing?

MichaelBuessemeyer:
This PR tackles the problem that the fill tool used in lower resolutions produces spotty segmentations.
The reason for this is that a lot of buckets get annotated at the same time and not all of them can be loaded into the memory. Therefore we restrict the usage of the tool to the first two magnifications (the same as for the trace tool).

- https://___.webknossos.xyz

### Steps to test:
- Open a volume or hybrid tracing.
- Using the fill tool should only work in the two best resolutions.
- If the fill tool is active and you zoom out, the fill tool should be disabled as soon as you zoom to a not-allowed resolution.

### Issues:
- fixes #4923

------
- [x] Ready for review


frontend/javascripts/oxalis/view/action-bar/volume_actions_view.js
```diff
@@ -391,7 +392,13 @@ export default function VolumeActionsView() {
           style={narrowButtonStyle}
           value={VolumeToolEnum.FILL_CELL}
         >
-          <i className="fas fa-fill-drip" />
+          <i
+            className="fas fa-fill-drip"
+            style={{
+              opacity: disabledInfosForTools[VolumeToolEnum.FILL_CELL].isDisabled ? 0.5 : 1,
+            }}
+          />
```
MichaelBuessemeyer, line 400: Somehow changing the opacity of this fa icon triggers a small transition animation that the icon of the trace tool does not have, as it is not a fa icon.
At first look, I could find a way to avoid the animation. Is it alright to keep the animation?

CHANGELOG.unreleased.md
```diff
@@ -17,6 +17,7 @@ For upgrade instructions, please check the [migration guide](MIGRATIONS.released
 
 ### Changed
 - The brush tool is disabled in low magnifications (magnification 16 and lower) to avoid performance problems when annotating. [#5017](https://github.com/scalableminds/webknossos/pull/5017)
+- The fill tool is disabled in low magnifications (magnification 2 and lower) to avoid producing spotty annotations. [#5050](https://github.com/scalableminds/webknossos/pull/5050)
```
philippotto, line 20:
```suggestion
- The fill tool is disabled in low magnifications (magnification 2 and lower) to avoid producing spotty annotations (caused by needing too much main memory). [#5050](https://github.com/scalableminds/webknossos/pull/5050)
```

frontend/javascripts/oxalis/view/action-bar/volume_actions_view.js
philippotto: Yes, I think, that's alright :)

philippotto (approved):
Awesome :+1: Works very well.

PR #5049, opened 2021-01-15
Add beforeunload to dataset upload view

grittaweisheit:
This PR double checks with the user if he/she really wants to leave the dataset upload view. 

### URL of deployed dev instance (used for testing):
- https://___.webknossos.xyz

### Steps to test:
- upload a dataset
- try to navigate elsewhere while it's still uploading
- you should be warned about leaving while it's still loading

### Issues:
- fixes #5051

------
- [x] Updated [(unreleased) changelog](../blob/master/CHANGELOG.unreleased.md#unreleased)
- [ ] ~~Updated [(unreleased) migration guide](../blob/master/MIGRATIONS.unreleased.md#unreleased) if applicable~~
- [ ] ~~Updated [documentation](../blob/master/docs) if applicable~~
- [ ] ~~Adapted [wk-connect](https://github.com/scalableminds/webknossos-connect) if datastore API changes~~
- [ ] ~~Needs datastore update after deployment~~
- [x] Ready for review

PR #5051, opened 2021-01-15
Fewer error messages in view mode

### LINKED ISSUE #5046: Failure in saving dataSetConfiguration causes UI crash
I’m getting the »Failed to save annotation. Retrying.« Toast even in view mode and the »Internal error.
Please reload the page to avoid losing data. [object Object] undefined« alert.

Saving dataSetConfiguration should not be that important, maybe this error can be (more) silently ignored, while still logging it to airbrake

MichaelBuessemeyer:
This PR reduces the possible error messages displayed in view only mode.
The two errors that are now ignored in view only mode are:
- The error that is show when the updating the tracing fails.
- The error that is show when saving the users dataset settings fails.

- https://lesserrormessagesinviewmode.webknossos.xyz

### Steps to test:
- I only found out how to reproduce the second error (locally):
  - Open a dataset in view mode and when it is loaded kill your local server.
  - The open the dataset settings an eg. enable/disable the "Render Isosurfaces (Beta)"-setting. This should not trigger an error.
  - If you try the same on the master, the tracing should enter the "error state" and an alert should be displayed. 

### Issues:
- fixes #5046

------
- [x] Ready for review


daniel-wer (commented):
Could you add that errors are still logged to airbrake? Otherwise we may never know about them.
For example, my assumption would be that the update actions save queue is always empty in view mode and it would be very interesting if (and how/why) this assumption might not hold.

Apart from that, LGTM :+1:

MichaelBuessemeyer:
@daniel-wer I just added the airbrake notification for view mode only.  :). Could you please give this a check (whether I used the notify method correctly)?

daniel-wer:
@MichaelBuessemeyer looks good to me. Please also add this to the other spot where an error is suppressed :+1:

daniel-wer (approved):
LGTM, worked as expected during testing :+1:

philippotto:
Accidentally closed by #5049.

Issue #5038, opened 2021-01-11
Local storage quota exceeded


fm3:
For the second time I got the Firefox error message described here https://support.mozilla.org/de/questions/1309735 while refreshing the dataset list. Clearing site data helps, but, do we need to store that much? Or is there a leak?

Can the error be caught and handled appropriately?

fm3:
A chrome user reported the same problem.

philippotto:
i'll build a hotfix now.

philippotto:
I think, we can close this issue. We've got an error boundary UI now which also shows a button to clear the local storage if this quota should be exceeded. Also, the cached dataset list is more compact now, so the chances should be rather low.

PR #5052, opened 2021-01-18
also clear new datasets cache key for global localStorage key

philippotto:
- https://___.webknossos.xyz

### Steps to test:
- it's a bit hard to test (which is why this slipped in earlier), which is why I would suggest to have a hard look at the one line change. inspecting my current localStorage for the mh wk instance, I see the following keys:
`email@address-orgaName-wk.datasets-v2`
`email@address-wk.datasets-v2`
- the second one is deprecated (doesn't use the orga name), but uses the new dataset key (since the version didn't need to be bumped).
- the line change should delete this entry.

### Issues:
- helps for #5038 
------
- [X] Ready for review


daniel-wer (approved):
So the v2 was introduced before the datasets key was made organization-specific (I thought it was the other way around)? In any case, this shouldn't break anything, so LGTM :)

philippotto:
> So the v2 was introduced before the datasets key was made organization-specific (I thought it was the other way around)?

Yeah, I thought so, too, but I checked it now: v2 was introduced with the hybrid-by-default PR. The organization-specific stuff came later :eyes:

PR #5055, opened 2021-01-18
Disable parallel dataset upload requests for now

fm3:
The name-reserving logic of the wk backend leads to errors if upload chunk requests come in parallel. Since that fix looks nontrivial I suggest we disable parallel requests for the moment, to fix the uploads.

### Steps to test:
- upload a dataset with a fresh name
- should not show name conflict errors

- [x] Ready for review

### DIFF
--- a/frontend/javascripts/admin/admin_rest_api.js
+++ b/frontend/javascripts/admin/admin_rest_api.js
@@ -919,7 +919,7 @@ export function createResumableUpload(datasetId: APIDatasetId, datastoreUrl: str
         query: datasetId,
         chunkSize: 10 * 1024 * 1024, // set chunk size to 10MB
         permanentErrors: [400, 403, 404, 409, 415, 500, 501],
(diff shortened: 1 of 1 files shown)


philippotto (approved):
LGTM!

frontend/javascripts/admin/admin_rest_api.js
```diff
@@ -919,7 +919,7 @@ export function createResumableUpload(datasetId: APIDatasetId, datastoreUrl: str
         query: datasetId,
         chunkSize: 10 * 1024 * 1024, // set chunk size to 10MB
         permanentErrors: [400, 403, 404, 409, 415, 500, 501],
-        simultaneousUploads: 3,
+        simultaneousUploads: 1,
```
philippotto, line 922:
```suggestion
        // Only increase this value if https://github.com/scalableminds/webknossos/issues/5056 is fixed
        simultaneousUploads: 1,
```

Issue #5001, opened 2020-12-18
Auto-Detect “needsConversion” on Dataset Upload


fm3:
We could unpack every dataset to a temp dir first and then check the contents to decide if it needs some conversion or is ready to be moved to destination dir as is

philippotto:
@MichaelBuessemeyer could you have a look whether the front-end can get the file names within a ZIP without having to hold the entire zip file in memory? then, one can iterate over all files to check whether there is a wkw file. in that case, "needsConversion" can be set to false.

Here we are also looking at the file names within a zip file:
https://github.com/scalableminds/webknossos/blob/72dfd4f0a72b95e69e86967ee2030aa32aca35a1/frontend/javascripts/oxalis/view/right-menu/trees_tab_view.js#L208-L211

However, I'm not sure whether this loads the entire zip into main memory. You can verify this by loading a 1 GB zip file into wK and read the file names in the zip. if it's fast and the RAM looks ok, this should be a good approach to close this issue :).

MichaelBuessemeyer:
@philippotto According to this [issue](https://github.com/Stuk/jszip/issues/290) `JSZip` does not support this. However, [this project](https://gildas-lormeau.github.io/zip.js/index.html) does and it can be used pretty easily using an npm package.
The first implementation of this works and is in the upload UI rework PR.

fm3:
Fixed in #5081 :rocket:

PR #5064, opened 2021-01-20
Repair Unused DBAccessContexts

### LINKED ISSUE #4992: Reports: enforce organizations are isolated
The project progress and open tasks reports take team IDs. From a quick glance of the code, there is no back-end side enforcement that the user can see that team. This should be enforced (or at least that you cannot get report data for other organizations).
There is no immediate danger as you still at least have to find out team IDs, but this should be addressed.

### LINKED ISSUE #5013: Clean Up Remaining Backend Compiler Warnings
As a follow-up after #4983 and #5012 the remaining unused warnings (as far as I remember all deal with DBAccessContext). For each, we should check if either a real access check should be added, or where global access is fair, in which case the context needn’t be passed.

fm3:
There were a lot of unused warnings for DBAccessContext objects, which are used on the lowest level to ensure rights management. 
- Most of these were correctly unused, in this case I removed them entirely. 
- Some were unused correctly but could not be removed because of subclassing, in this case I added `@silent`
- Some pointed to missing access checks, which I added, either in form of DB access checks or explicit checks on higher levels
- I also added access checks to the reports controllers to ensure reports can only be obtained for the own organization and teams (even if team IDs are leaked somehow)
- One remaining problem is with ExperienceDomains, tracked in follow-up issue #5071
- I also included some small refactorings here (sorry for the large diff, but most small changes should be obviously harmless)

### Steps to test:
- upload dataset with initial teams (as admin/dataset manager) should still be allowed
- project publicWrites was changed (hide owner if user is not allowed to see owner), should check if that creates problems (try to trace tasks as normal user)
- update description / name of annotation
- dataset token sharing should still work
- public annotations should still be visible as logged-out user
- my annotations list should still work
- you should no longer get a response for report requests even if you know the right team id, if you are not in that team and no admin

### Issues:
- fixes #5013
- fixes #4992

------
- [x] Ready for review


app/models/annotation/handler/AnnotationInformationHandler.scala
```diff
@@ -29,25 +31,26 @@ trait AnnotationInformationHandler extends FoxImplicits {
...
   def provideAnnotation(identifier: ObjectId, user: Option[User])(implicit ctx: DBAccessContext): Fox[Annotation]
 
+  @silent // suppress warning about unused implicit ctx, as it is used in subclasses
   def nameForAnnotation(t: Annotation)(implicit ctx: DBAccessContext): Fox[String] =
     Fox.successful(t.id)
 
   def restrictionsFor(identifier: ObjectId)(implicit ctx: DBAccessContext): Fox[AnnotationRestrictions]
 
   def assertAllOnSameDataset(annotations: List[Annotation]): Fox[Boolean] = {
+    @tailrec
```
youri-k, line 41: Do we want to add this annotation to all tailrecursive functions? There are also other places, where we use tailrecursive functions.

app/utils/SQLHelpers.scala
```diff
@@ -58,32 +59,30 @@ class SimpleSQLDAO @Inject()(sqlClient: SQLClient)(implicit ec: ExecutionContext
...
-        case Success(res) => {
+        case Success(res) =>
           Fox.successful(res)
-        }
-        case Failure(e: Throwable) => {
+        case Failure(e: Throwable) =>
           val msg = e.getMessage
           if (retryIfErrorContains.exists(msg.contains(_)) && retryCount > 0) {
-            logger.debug(s"Retrying SQL Query (${retryCount} remaining) due to ${msg}")
+            logger.debug(s"Retrying SQL Query ($retryCount remaining) due to $msg")
```
youri-k, line 67: I remember you said `debug` output of some instances is suppressed. Do we want to change the logging type here?

webknossos-datastore/app/com/scalableminds/webknossos/datastore/controllers/DataSourceController.scala
```diff
@@ -81,51 +83,52 @@ class DataSourceController @Inject()(
...
+                    _ <- if (!uploadService.isKnownUpload(uploadId))
+                      webKnossosServer.validateDataSourceUpload(id) ?~> "dataSet.upload.validation.failed"
+                    else Fox.successful(())
+                    chunkFile <- request.body.file("file") ?~> "zip.file.notFound"
+                    _ <- uploadService.handleUploadChunk(uploadId,
+                                                         id,
+                                                         resumableUploadInformation,
+                                                         chunkNumber,
+                                                         new File(chunkFile.ref.path.toString))
+                  } yield {
```
youri-k, line 118: You typically omitted the brackets for a single line `yield`, so I'd suggest changing this here, too.

webknossos-datastore/app/com/scalableminds/webknossos/datastore/controllers/DataSourceController.scala
```diff
@@ -81,51 +83,52 @@ class DataSourceController @Inject()(
...
-    }
   }
 
-  def finishUpload = Action.async(validateJson[UploadInformation]) { implicit request =>
+  def finishUpload: Action[UploadInformation] = Action.async(validateJson[UploadInformation]) { implicit request =>
     accessTokenService.validateAccess(UserAccessRequest.administrateDataSources) {
       AllowRemoteOrigin {
         for {
-          (dataSourceId, initialTeams) <- uploadService.finishUpload(request.body)
+          (dataSourceId, initialTeams, needsConversion) <- uploadService.finishUpload(request.body)
```
youri-k, line 131: Do we use the `needsConversion` flag here? I think it can be silenced with `_`.

youri-k (approved):
Nice :tada: 
Thank you especially for the explaining comments in the `UserTokenController`.

See my small comments. They're mostly about consistency throughout the code.

app/utils/SQLHelpers.scala
fm3: If I remember correctly, that is a standalone-datastore-only issue, and debug should be fine here. It is certainly better in terms of semantics.

webknossos-datastore/app/com/scalableminds/webknossos/datastore/controllers/DataSourceController.scala
fm3: Good catch! That was a leftover from my attempts at #5001 which apparently slipped into this branch. I removed the additional return value.

app/models/annotation/handler/AnnotationInformationHandler.scala
fm3: Yes, eventually I would like to see all tail recursive functions annotated. However, they are not super easy to find.
Actually this is part of a somewhat larger campaign to address the code style warnings produced by IntelliJ. This started from my trying to get rid of unnecessarily public members of any classes, and developed into a desire to get rid of all of these warnings. This is also related to the removing of superfluous enclosing blocks and unused vales. However, there is too much code to do all of that manually all at once. So this is why I clean a file here and there by and by. Same goes for the ${id.id} thing and single-line yield statements, let’s defer that to a later PR (or multiple ones to keep conflicts small). This does not have a particularly high priority. Do you think this approach is fair?

PR #5065, opened 2021-01-20
Release 21.02

philippotto:
- [X] Ready for review

### DIFF
--- a/CHANGELOG.released.md
+++ b/CHANGELOG.released.md
@@ -7,6 +7,30 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Calendar Versioning](http://calver.org/) `0Y.0M.MICRO`.
 For upgrade instructions, please check the [migration guide](MIGRATIONS.released.md).
 
+
+## [21.02.0](https://github.com/scalableminds/webknossos/releases/tag/21.02.0) - 2021-01-20
+[Commits](https://github.com/scalableminds/webknossos/compare/21.01.0...21.02.0)
+
+### Highlights
+- The dataset import UI was streamlined by making it clearer when automatic suggestions are available for the dataset properties. Previously, these suggestions were applied automatically which could cause some confusion. [#4944](https://github.com/scalableminds/webknossos/pull/4944)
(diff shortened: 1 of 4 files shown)


CHANGELOG.released.md
```diff
@@ -7,6 +7,30 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Calendar Versioning](http://calver.org/) `0Y.0M.MICRO`.
 For upgrade instructions, please check the [migration guide](MIGRATIONS.released.md).
 
+
+## [21.02.0](https://github.com/scalableminds/webknossos/releases/tag/21.02.0) - 2021-01-20
+[Commits](https://github.com/scalableminds/webknossos/compare/21.01.0...21.02.0)
+
+### Highlights
+- The dataset import UI was streamlined by making it clearer when automatic suggestions are available for the dataset properties. Previously, these suggestions were applied automatically which could cause some confusion. [#4944](https://github.com/scalableminds/webknossos/pull/4944)
```
philippotto, line 15: This is basically a copy of "- Suggestions for the datasource settings of a dataset are no longer applied automatically. They can be applied optionally now. [#4944](https://github.com/scalableminds/webknossos/pull/4944)", but more tuned to the highlights section, which is why I kept both variants in the changelog.

CHANGELOG.released.md
```diff
@@ -7,6 +7,30 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
...
+- Added the possibility to generate skeletons from an HDF5 agglomerate file mapping on-the-fly. With an activated agglomerate file mapping, use `Shift + Middle Mouse Click` to import a skeleton of the cell into the annotation. Alternatively, use the button in the segmentation tab to import a skeleton of the centered cell into the annotation. [#4958](https://github.com/scalableminds/webknossos/pull/4958)
```
fm3, line 16: Do you think we should add more public documentation to the agglomerate mapping files if we advertise them here?

fm3 (approved):

CHANGELOG.released.md
philippotto: I think, it boils down to: do we want to encourage the usage of agglomerate files in the long run? If so, we could supply documentation, conversion tools etc etc. However, we should probably clear the initial question first. I'll open a discussion issue for this.

philippotto: See https://github.com/scalableminds/webknossos/issues/5066

PR #5073, opened 2021-01-21
Run snapshot tests without concurrency for deterministic ordering

### LINKED ISSUE #3798: snapshot tests: ensure deterministic ordering for snapshots
when I run the tests locally, I get a diff because `## annotations-tracing-volume` and `## annotations-updateActions` are swapped

fm3:
Locally produced snapshots did not match those created by the CI in that the order of the different tests was different. Passing `-C 1` to `yarn test-e2e` fixes that.

### Steps to test:
- Successful CI run despite snapshot diff

### Issues:
- fixes #3798

- [x] Ready for review


test/e2e/End2EndSpec.scala
```diff
@@ -39,7 +37,7 @@ class End2EndSpec (arguments: Arguments) extends Specification with GuiceFakeApp
 
   }
 
-  private def runWebdriverTests = "yarn test-e2e".run().exitValue()
+  private def runWebdriverTests = "yarn test-e2e -C 1".run().exitValue()
```
philippotto, line 40: I think, it would be better to incorporate this change in the `test-e2e` entry in the package.json. That way, manual executions of this command will also use the concurrency flag. Do you agree?

fm3: Yes I agree. Not sure where to pass it there, though. Should it be in `test.sh`? Or its call in package.json?

philippotto: Good point, it's probably best to have it in `test.sh` where `*.e2e.js` is passed (line 48).

PR #5078, opened 2021-01-22
Add an issue template

MichaelBuessemeyer:
To support write more details about the issue if it's none trivial, I suggest adding an issue template.

Feedback for the template is welcome :smiley_cat: 

This template was generated using [open-source-templates](https://github.com/TalAter/open-source-templates) and then modified to our needs.

- [x] Ready for review


.github/ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,37 @@
+<!--- Provide a general summary of the issue in the Title above, if the title is not expressive enough. -->
```
daniel-wer, line 1: Conceptual question: Who are these comments aimed at? They somewhat read like the user creating an issue should be able to read them, but looking at the rich diff which shows the rendered markdown, I cannot see them.

.github/ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,37 @@
...
+<!--- Provide a list of actions to reproduce the bug. -->
+1.
+2.
+3.
+4.
+
+
+## Your Environment (for bugs)
+<!--- If you bug only occurs in a special environment, please list as many relevant details about the environment as possible. -->
+* Version used:
```
daniel-wer, line 34: What version is meant here? Maybe not needed.

.github/ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,37 @@
...
+<!--- If you're describing a bug, tell us what should happen -->
+<!--- If you're suggesting a change/improvement, tell us how it should work -->
+
+## Current Behavior
+<!--- If describing a bug, tell us what happens instead of the expected behavior -->
+<!--- If suggesting a change/improvement, explain the difference from current behavior -->
+<!--- If your are suggesting a new feature just check the box below -->
+- [ ] This is a new feature so there isn't any current behavior.
+
+## Possible Solution
```
daniel-wer, line 18: This somewhat overlaps with the "Expected Behavior" section, doesn't it? Maybe we can get rid of this section to slim down the issue template a little bit.

daniel-wer (commented):
Thanks for taking the initiative to create this template! I had a quick look and made some comments. I feel like we should try to slim this down a little bit more to avoid too much friction during issue creation :)

.github/ISSUE_TEMPLATE.md
MichaelBuessemeyer: > Who are these comments aimed at?

The are aimed for the creator of the issue.

> They somewhat read like the user creating an issue should be able to read them, but looking at the rich diff which shows the rendered markdown, I cannot see them.

The comments are meant to only be seen when creating/editing the issue. The information in the comments only exists to give hints on what to write at the position of the comment. 
Making the text in the comments visible as default would only overcrowd the issue with unnecessary information. (At least in my opinion)

.github/ISSUE_TEMPLATE.md
MichaelBuessemeyer: I would make this section optional. (maybe add this to the title)
I think separating the expected behavior and "possible solutions" is a good thing. Maybe "possible solution" is not the a good way of expressing what I meant with this section.

Explanation:
"Expected Behavior": This section should describe what behavior wk should have from a high level perspective leaving the implementation aside. Eg: "When I use `Left-Click`in skeleton mode a new node should be created." 

"Possible Solution": This section should be optional and contain concrete hints or ideas on how to implement the expected behavior. Eg: "In the `skeletontracing_plane_controller.js`, add a method that upon `Right-Click`that creates a new node at the current mouse position using the `createNodeAction`-Action.

An actual example of what could be in the "possible solutions" section can be found [here](https://github.com/scalableminds/webknossos/issues/5001#issuecomment-762281543).

Of cause, if the creator does not have hint on the implementation the whole section can be removed.

.github/ISSUE_TEMPLATE.md
MichaelBuessemeyer: Additionally, those comments kinda guide a non-scalable member through the template to try to ensure they understand what information we need / is helpful. I know this does not happen often, but I did happen in the past :)

daniel-wer: > The comments are meant to only be seen when creating/editing the issue. The information in the comments only exists to give hints on what to write at the position of the comment.

Ah I see, the comments are only visible in non-markdown mode, so when creating/editing the issue. That makes sense :)

.github/ISSUE_TEMPLATE.md
daniel-wer: Yes, I understand the distinction, but I'm thinking of ways to slim down the template a bit. In my opinion, the friction with this current template may be quite high. There's a lot of sections that need to be thought about, be filled out or be deleted.
I'd say the "Possible Solutions" section would only be filled rarely and we could do without it. If someone has concrete implementation ideas, they could add them to another section or simply add a similar section themselves. However, let's see what the others think :)

philippotto: I think, it would help to provide different templates. The current template has a large amount of bug-related stuff, but we often also create feature-issues (where different things would be interesting). GitHub supports having multiple templates from which one has to pick when creating an issue. Have a look at the scikit-learn repository. Here are the templates: https://github.com/scikit-learn/scikit-learn/tree/main/.github/ISSUE_TEMPLATE

You can also test the "new issue" button in their repo to see how it behaves.

What do you think?

MichaelBuessemeyer: > What do you think?

Sounds like a very good idea :+1:.

To do this I need the help of an repo owner like @normanrz or @hotzenklotz. [Here](https://docs.github.com/en/github/building-a-strong-community/configuring-issue-templates-for-your-repository) is the github tutorial on what to do to get this nice template selection when creating an issue like [scikit-learn](https://github.com/scikit-learn/scikit-learn/issues/new/choose) it has.

MichaelBuessemeyer: And btw. I just separated the issue template into bugs and features.

philippotto: > To do this I need the help of an repo owner like @normanrz or @hotzenklotz. Here is the github tutorial on what to do to get this nice template selection when creating an issue like scikit-learn it has.

Has this happened?

normanrz: I did the steps. See https://github.com/scalableminds/webknossos/pull/5122/files. I guess we can merge both PRs into one.

hotzenklotz: I just followed the tutorial and enabled a bug and features request templates with the standard values. This created a new PR: https://github.com/scalableminds/webknossos/pull/5121 

@MichaelBuessemeyer Feel free to take over my PR and edit as required.

MichaelBuessemeyer:
closing this pr in favour of #5121

PR #5085, opened 2021-01-25
Drop support for Knossos cubes

### LINKED ISSUE #4359: Drop support for knossos cubes
We could do a lot of nice refactoring if knossos cube support is dropped. Like, we could get rid of the ElementClass concept in favour of voxelType + voxelSize.
First, let’s check in the databases if anyone still uses knossos datasets in wk.

fm3:
As discussed previously, the Knossos dataformat support has not been tested in a while, and it never was 100% compatible anyway. Since all our datasets have by now been converted to wkw, the Knossos code can be removed.

I also added a few more type annotations to the datastore code.

### Steps to test:
- wkw datasets should still be reported + browsable

### Issues:
- fixes #4359

------
- [x] Ready for review


webknossos-datastore/app/com/scalableminds/webknossos/datastore/dataformats/BucketProvider.scala
```diff
@@ -11,30 +12,28 @@ import scala.concurrent.ExecutionContext
...
-        logger.warn(
-          s"loading file in $className took ${if (duration > 3000) "really " else ""}long.\n"
-            + s"  duration: $duration\n"
-            + s"  dataSource: ${readInstruction.dataSource.id.name}\n"
-            + s"  dataLayer: ${readInstruction.dataLayer.name}\n"
-            + s"  cube: ${readInstruction.cube}"
-        )
-      }
-      result
+      implicit ec: ExecutionContext): Fox[Array[Byte]] =
```
youri-k, line 19: Why is it a `Fox[..]`? I think the DataCubeCache can be rewrote to only use `Box[..]`, but maybe that's another issue.

youri-k (approved):
Wow that makes the code easier tor read 🎉 

I just reviewed the code, so I did not test it yet. Do you think it's necessary before merging?
I'm pretty confident that this shouldn't cause any problems after reading the code.

One style question: Do we want to rename all Formats to `jsonFormat` in the future?

webknossos-datastore/app/com/scalableminds/webknossos/datastore/dataformats/BucketProvider.scala
fm3: Yes could be right, but since this is not a change introduced here, I’d defer that. The Fox does not do any harm here for now :)

fm3:
Thanks for the quick review!
Yes, I think `jsonFormat` is a clearer name for these. Do you agree?
I already tested refreshing the dataset list and browsing the data, so I think we can say it’s fine without you re-testing. If the CI also agrees :)

PR #5091, opened 2021-01-26
Add separate `register` route

### LINKED ISSUE #5076: Separate login and signup pages
In order to integrate the new landing page for webknossos.org, we need separate pages for login and signup.
Preferably, the url would be `/login` and `/signup`.

philippotto:
- https://___.webknossos.xyz

### Steps to test:
- (test with isDemo true and false separately)
- open `/register` and create new account

### Issues:
- fixes #5076

------
- [x] Ready for review


philippotto:
Originally, I also planned to refactor the authentication forms a bit, but gave up in the end, since it got out of hand. Instead, I created a follow-up issue which hopefully captures the insights I gained today: https://github.com/scalableminds/webknossos/issues/5096

frontend/javascripts/router.js
```diff
@@ -427,10 +427,7 @@ class ReactRouter extends React.Component<Props> {
               />
               <Route
                 path="/auth/register"
```
normanrz, line 429: Would you mind renaming the route to `/auth/signup`?

philippotto: Should we keep the old `register` path (and redirect to `signup`)? Or can it be a "breaking change"?

normanrz: should be a redirect or keep both.

philippotto: done

frontend/javascripts/admin/auth/registration_view.js
```diff
@@ -4,15 +4,12 @@ import { Link, useHistory } from "react-router-dom";
...
-import RegistrationForm from "./registration_form";
+import SpotlightRegistrationForm from "admin/auth/spotlight_registration_form";
+import RegistrationForm from "admin/auth/registration_form";
 
-type Props = {
-  organizationName?: string,
-};
-
-function RegistrationView({ organizationName }: Props) {
+function RegistrationViewNotDemo() {
```
daniel-wer, line 12: I was confused how invite links could still work, although no organization name could be supplied, here, but I found out that for invites the AuthenticationModal is used. Glad you created https://github.com/scalableminds/webknossos/issues/5096 :+1:

daniel-wer (approved):
I tested the demo, non-demo and invitation registration case. Everything worked as expected, nice work :+1: Would definitely be in favor of https://github.com/scalableminds/webknossos/issues/5096 :)

PR #5095, opened 2021-01-26
Fix nightly

philippotto:
The new screenshots contain the reload-button for isosurfaces which was enabled by default by #4917.


daniel-wer (approved):
:+1:

PR #5097, opened 2021-01-26
Store Uploader ID with Dataset

### LINKED ISSUE #4987: Store uploader for dataset
For datasets that were uploaded through the UI, it would be good to store which user uploaded it. Can be null for detected dataset from disk.

fm3:
### Steps to test:
- Upload a dataset via UI
- `select _uploader from datasets` should show uploader id
- Datasets reported from disks should have the value null in that column.
- Dataset json representation (e.g. in GET /api/datasets`) should not change.

### Issues:
- fixes #4987 

------
- [x] Ready for review


youri-k (approved):
LGTM :shipit:

normanrz:
@philippotto could we list the uploader also in the dataset list? maybe morph the `creation date` column to a `creation` column with time and uploader.

philippotto:
>@philippotto could we list the uploader also in the dataset list?

should be doable. I assume the back-end would need to send the owner names (not the ids) with the dataset list, if this is not already done.

> maybe morph the creation date column to a creation column with time and uploader.

is an additional column also ok? the dataset layers column could be a bit smaller and we also have a nice overflow-x behavior for the table, so space shouldn't be a big issue. showing two different attributes in one column breaks the sorting UI unfortunately.

fm3:
TBH I find horizontal scrollbars rather annoying :see_no_evil: with the Actions column not being pushed out it’s not too terrible but still I would try to keep the table as narrow as possible, since not everyone has 20″ screens. I understand that the sorting is important, though :thinking: Do you think maybe »Active« could be removed? Inactive datasets can be recognized by the changed action list, right?

philippotto:
continued in https://github.com/scalableminds/webknossos/issues/5147

Issue #5022, opened 2021-01-05
NUX: Add UI hints to show features


philippotto:
For new users (or new features), we could show UI hints (e.g., popovers) to explain some important features. This can be useful to guide users during (or after) onboarding through the most important steps. For example, after the account was created, one could nudge the user to import a dataset or to start an annotation.

When we add new features, we could also add such UI hints to present the feature to users.

In general, we probably want to store a boolean for each user and for each "feature" to remember whether the user has already cancelled the NUX.

nux == new user experience

- [X] gather nux-worthy features:
  - short tour/tutorial to explain wk core UI
  - new "meshes" tab with isosurface list (only when there is a segmention)
- [x] have a look at announce kit
- [x] implement

philippotto:
@daniel-wer researched this bit and found the promising-looking library https://shepherdjs.dev/. I'd suggest to give this library a try and test out the following stuff:

- build a dummy tour for the tracing view ((1) this is the toolbar, it can do this and that (2) these are the tracing viewports etc...)
- highlight some new feature (e.g., the overhauled meshes tab)

This doesn't necessarily need to be very polished. I'm primarily interested in how easy the integration is and what kind of UX impression this would have.

Might be something for @MichaelBuessemeyer when he has free capacities :)

MichaelBuessemeyer:
Ok here are my thoughts/findings about the library:
First, this lib looks nice and could be a good solution. Additionally, it looks like it is pretty easy to use. 
However, the lib supports multiple js frontend frameworks and has specialized versions for them. The same goes for React. But I could not get the react version working correctly for quite some time. One reason might be, that the react version only supports hooks and not the "usual" react way of handling a components state.

Therefore I just stopped trying to use the react version and moved to the "normal" version. The normal version worked out of the box and I was able to create a small tutorial within a short time period.
Note: Everything is built for the OthroView mode. Other view modes might crash.

My first thoughts on what we need to change:
- To directly hint at an element, this element needs a unique css class or id. Therefore components we want to hint at would need a small but easy change.
- To show the user the newest features since his last visit, I suggest the following for the tracing view:
  - We need to maintain a single js file containing an array of steps hinting at the newest features in chronological order. This would be like a changelog.
  - Each user has an id or index of the last seen update hint of that list saved in the db.
  - Each time the user opens the tracing view the frontend checks whether the user has any pending new features to be hinted at.
[...]

PR #5099, opened 2021-01-26
Implement new "welcome header"

### LINKED ISSUE #5020: Make dataset import more prominent
Especially necessary if there are no datasets yet.

### LINKED ISSUE #5023: Iterate on "new user screen"
- [x] link to a hardcoded dataset so that one can easily start annotating something (i.e., testing out wk)
- [ ] link to guide (e.g., dataset upload)

philippotto:
The new welcome header with the empty-datasets-CTA currently looks like this:

image

In addition to the new header, I also removed the redundant headlines from the dashboard tab contents (as they merely repeated the tab titles).

Remaining todo:
- [x] add UI to hide the banner
- [x] persist whether the header was hidden by the user. the old banner was set via a GET parameter, which meant that the user would see this only once. should have a JSON-like property for a user to store stuff like this? @fm3 What do you think? for https://github.com/scalableminds/webknossos/issues/5022 we would need this, too. db migrations can update values within a json object, right?
- [x] clean up (e.g., remove code of old whats next banner)

- https://welcomeheader.webknossos.xyz

### Steps to test:
- log in on the dev instance
- check out the welcome banner in the dashboard and test its different actions
  - feel free to test the close-icon, but note that the header won't re-appear on page refresh. so, do this at the very end of testing :laughing: 
- checkout the empty dataset table with the overhauled CTA cards
  - they will only appear if the dataset list is empty. I hardcoded this for now to simplify testing.

### Issues:
- fixes #5023
- fixes #5020

------
- [x] Ready for review


fm3:
Yes we can add a json field to the user, and yes we can edit those in SQL evolutions. I guess it should be stored with the multiuser, so that it is not reset when joining a new organization. What should it be called? `nuxInfo`?

philippotto:
> Yes we can add a json field to the user, and yes we can edit those in SQL evolutions.

Great!

> I guess it should be stored with the multiuser, so that it is not reset when joining a new organization.

Yes, I agree :+1: What is the current `/api/users/${newUser.id}` route referring to? Is it the multi user?

> What should it be called? `nuxInfo`?

Yes, sounds good to me :+1:

philippotto:
I decided to also incorporate the CTA in case of an empty datasets-table into this PR, since this should go well together with the general welcome-header.

The current version looks like this (for a completely empty account on wk.org):
image

The deployed version can be found here: https://welcomeheader.webknossos.xyz/?showWhatsNextBanner=true (notice that the `showWhatsNextBanner` will disappear after page load; so a reload will hide the banner. this will change once we can persist the information in the DB).

@hotzenklotz @normanrz Maybe you already want to give feedback on the wording?

fm3:
@philippotto I just added the `novelUserExperienceInfos` field to the user json, defaulting to `{}`. When sent to the user update request, the field is completely overwritten for that user (or actually for the multiuser, as this is cross organization). Feel free to test it on this branch, does this work for you? :)

fm3:
I’ll have a look 👀

frontend/javascripts/admin/welcome_ui.js
```diff
@@ -0,0 +1,147 @@
...
+              />
+              <WhatsNextAction
+                title="Learn How To Create Annotations"
+                description="Watch a short video to see how data can be annotated with webKnossos."
+                icon={<Icon type="plus-circle" className="action-icon" />}
+                href="https://youtu.be/W-dosptovEU?t=52"
+              />
+              <WhatsNextAction
+                title="Invite Your Colleagues"
+                description="Send invites to your colleagues that they can join your organization."
```
hotzenklotz, line 128: Find das etwas hackelig in der Formulierung. Evtl.
```suggestion
                description="Send invites to your colleagues and ask them to join your organization."
```

fm3:
> However, the response contains novelUserExperienceInfos: {}. Do you know whats going on?

My bad, typical copy-paste error, should be fixed now :)

frontend/javascripts/admin/welcome_ui.js
```diff
@@ -0,0 +1,147 @@
...
+            <p
+              style={{
+                fontSize: 20,
+                lineHeight: 1.5,
+                marginTop: 0,
+                marginBottom: 8,
+              }}
+            >
+              Congratulations on your new webKnossos account! To hit the ground running, we
+              recommend the following steps for you:
```
hotzenklotz, line 104: Ich denke es heißt `recommend .. to you` https://dict.leo.org/german-english/recommend

```suggestion
              recommend the following steps to you:
```

frontend/javascripts/admin/welcome_ui.js
philippotto: Whoops, I intended to write it as "send invites [...] SO that they can ...", but the word "so" was missing. Your suggestion sounds fine, too, though. I'll take yours :+1:

frontend/javascripts/admin/welcome_ui.js
```diff
@@ -0,0 +1,147 @@
...
+              }}
+            >
+              Congratulations on your new webKnossos account! To hit the ground running, we
+              recommend the following steps to you:
+            </p>
+            <div className="whats-next-actions-grid">
+              <WhatsNextAction
+                title="Open a Demo Dataset"
+                description="Have a look at a public dataset to experience webKnossos in action."
+                href={getDemoDatasetUrl()}
```
daniel-wer, line 110: I think it could be confusing for non-demo users that they are redirected to webknossos.org, here. In the dataset list you implemented a switch to show the "Add Sample Dataset" Modal to non-demo users, instead. This could be a good solution here as well.

conf/application.conf
```diff
@@ -87,11 +87,16 @@ features {
...
+  isDemoInstance = true
   autoBrushReadyDatasets = []
   taskReopenAllowedInSeconds = 30
   allowDeleteDatasets = true
   jobsEnabled = false
+  # For new users, the dashboard will show a banner which encourages the user to check out the following dataset.
+  # If isDemoInstance == true, `/createExplorative/hybrid/true` is appended to the URL so that a new tracing is opened.
+  # If isDemoInstance == false, `/view` is appended to the URL so that it's opened in view mode (since the user might not
+  # have an account).
+  publicDemoDatasetUrl = "https://webknossos.org/datasets/MPI_Brain_Research/2012-09-28_ex145_07x2_ROI2017_connectome"
```
daniel-wer, line 99: The internal feedback was to use this dataset instead: https://webknossos.org/datasets/scalable_minds/l4dense_motta_et_al_demo

daniel-wer (commented):
Great work, massive design step up :1st_place_medal: 

Please see my small comments.

daniel-wer:
Also, thanks for making the testing on the deployed version so easy and frictionless! :grinning:

frontend/javascripts/admin/welcome_ui.js
daniel-wer: I just saw that you discussed this in Slack already and others agreed with the current solution. So feel free to ignore this :)

conf/application.conf
philippotto: Good catch!

frontend/javascripts/admin/admin_rest_api.js
```diff
@@ -170,6 +170,26 @@ export function updateUser(newUser: $Shape<APIUser>): Promise<APIUser> {
...
+      ...novelUserExperienceShape,
+    },
+  };
+  const newUserSync = {
+    ...user,
+    ...updateShape,
+  };
+  const newUserAsync = updateUser(updateShape);
+
+  return [newUserSync, newUserAsync];
```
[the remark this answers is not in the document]
philippotto: Yeah, I was also a bit surprised that we didn't need something similar earlier and was afraid I'd be missing something ^^ But I think the rest of our code is either quite atomic (via forms) or very "sophisticated" (via the save queue). So this landed somewhere in-between.

philippotto:
@fm3 I've got to bother with you two remaining things:

- the migration would need to be adapted so that existing users get `{"hasSeenDashboardWelcomeBanner": true}` as an initial value. I assume it's not as easy as changing it to `ALTER TABLE webknossos.multiUsers ADD COLUMN novelUserExperienceInfos JSONB NOT NULL DEFAULT '{"hasSeenDashboardWelcomeBanner": true}'::json;`, is it?
- hiding the banner didn't work for a non-admin during testing. I created a new account, activated that via the admin, and then clicked the close-icon. then, I get `You are not authorized to view or edit this resource`.  are users not allowed to update their own account?

fm3:
> The migration would need to be adapted so that existing users get {"hasSeenDashboardWelcomeBanner": true} as an initial value. 

Done :) 

> Are users not allowed to update their own account?

Oh, yes that is actually correct 🤔 I’ll talk to youri today about how we can separate that.

conf/evolutions/063-novelUserExperienceinfos.sql
```diff
@@ -0,0 +1,32 @@
+-- https://github.com/scalableminds/webknossos/pull/5099
+
+START TRANSACTION;
+
+DROP VIEW webknossos.userInfos;
+DROP VIEW webknossos.multiUsers_;
+
+
+ALTER TABLE webknossos.multiUsers ADD COLUMN novelUserExperienceInfos JSONB NOT NULL DEFAULT '{}'::json;
```
youri-k, line 9: Maybe cast the type directly to `jsonb`

conf/evolutions/063-novelUserExperienceinfos.sql
```diff
@@ -0,0 +1,32 @@
...
+
+ALTER TABLE webknossos.multiUsers ADD CONSTRAINT nuxInfoIsJsonObject CHECK(jsonb_typeof(novelUserExperienceInfos) = 'object');
+
+-- users that already existed *before* this change should not get the welcome banner
+UPDATE webknossos.multiUsers SET novelUserExperienceInfos = '{"hasSeenDashboardWelcomeBanner": true}'::json;
```
youri-k, line 14: Same as above. Maybe cast the type directly to `jsonb`

youri-k (commented):
Backend LGTM apart from the small comment regarding the cast type in SQL

philippotto:
> Also, should non-admin members actually see the same welcome banner? They are not allowed to do most of what it explains?

I changed the code so that the invite-card is hidden when the user is not a team manager and the dataset-import-card is hidden if he is not a dataset manager. Admins can see all cards and "normal" users only the "open dataset" and "show tutorial video".

daniel-wer (commented):
Frontend LGTM. Should be ready to merge once the user editing permission issue is solved :)

fm3:
I solved the permission issue by moving the novelUserExperienceInfos update to a separate route. @philippotto Could you check that this works for you? Also, I changed the api call in `admin_rest_api.js`, please have a look there too, I’m not sure if for example the `updateShape.id` field is still needed?

philippotto:
> I solved the permission issue by moving the novelUserExperienceInfos update to a separate route. @philippotto Could you check that this works for you? Also, I changed the api call in `admin_rest_api.js`, please have a look there too, I’m not sure if for example the `updateShape.id` field is still needed?

Awesome! The id field wasn't used anymore. I simplified the function a bit and it works great :man_dancing: 

Then, this should be good to go, right, @daniel-wer?

daniel-wer (approved):
> Awesome! The id field wasn't used anymore. I simplified the function a bit and it works great man_dancing
> 
> Then, this should be good to go, right, @daniel-wer?

I cannot see these changes yet, but once they're pushed, this should be ready to go :+1:

philippotto:
> > Awesome! The id field wasn't used anymore. I simplified the function a bit and it works great man_dancing
> > Then, this should be good to go, right, @daniel-wer?
> 
> I cannot see these changes yet, but once they're pushed, this should be ready to go

Hm, github seems to have some problems right now. I'll wait a bit.

Issue #5101, opened 2021-01-27
Failing isosurface requests crash front-end


philippotto:
Failing isosurface requests crash front-end

PR #5102, opened 2021-01-27
Make isosurfaces more robust

philippotto:
- don't crash front-end if isosurface request fails
- retry isosurface requests if they fail
- indicate loading state of isosurfaces also on initial request

- https://robustisosurfaces.webknossos.xyz

### Steps to test:
- open a dataset with segmentation
- set network speed to slow
- request an isosurface --> a loading icon should be visible as long as the request is pending
- set network setting to offline
- request an isosurface --> a loading icon should appear and the console should log details about retries
- set network setting to online --> the isosurface generation should succeed on the next retry

### Issues:
- follow up #5101

------
- [X] Ready for review


frontend/javascripts/oxalis/model/sagas/isosurface_saga.js
```diff
@@ -207,9 +212,10 @@ function* loadIsosurfaceWithNeighbors(
     seedPosition != null ? seedPosition : yield* select(state => getFlooredPosition(state.flycam));
   const clippedPosition = clipPositionToCubeBoundary(position, zoomStep, resolutionInfo);
   let positionsToRequest = [clippedPosition];
-  if (seedPosition) {
```
philippotto, line 210: @grittaweisheit do you know why `addIsosurfaceAction` is only dispatched if there is a `seedPosition` ? 1) The used position is set to the current annotation position and used as a seed if `seedPosition == null`, so it should be fine to use that here, too? 2) I'm not really sure in which cases `seedPosition` is null anyway. The signature states `seedPosition: ?Vector`, but all callers seem to pass it if I'm not missing anything?

daniel-wer (approved):
Very nice, LGTM once your question was answered :+1:

frontend/javascripts/oxalis/model/sagas/isosurface_saga.js
grittaweisheit: I think it is because `ensureSuitableIsosurface` is also called at every flycam movement, but we don't really want to add an isosurface then. And since there is no seed position in that case, that is the way I decided whether to dispatch the `addIsosurfaceAction` or not.

philippotto: Ok, makes sense. Good point with the flycam-movement listener! I missed that. However, the "if seed position is null, we know only it was caused by a flycam movement"-logic is a bit too hidden for me. I just changed the code to simply dispatch `addIsosurfaceAction` everytime **unless** it already exists in the store. This is bit more intuitive and safer in my opinion.

PR #5108, opened 2021-01-29
Prevent deleting teams that are still referenced

### LINKED ISSUE #5077: Deleting team breaks the tasks still linked to it
When a team is deleted, tasks associated to it can no longer be loaded. The same may be true for other database entries. We need to develop a strategy for how to change the dependant entries.

fm3:
Fixed a bug where the user could delete teams that were still referenced in annotations, projects or task types, thus creating invalid state.

Also note follow-up issue #5109

- https://teamsdeleteassertion.webknossos.xyz

### Steps to test:
- Try deleting organization team, should be forbidden
- Try deleting team while not admin, should be forbidden
- Try deleting fresh team, should work
- Create team, create task type or project that uses that team, deleting it should now be forbidden.

### Issues:
- fixes #5077 

------
- [x] Ready for review


app/models/project/Project.scala
```diff
@@ -86,17 +86,15 @@ class ProjectDAO @Inject()(sqlClient: SQLClient)(implicit ec: ExecutionContext)
     for {
       accessQuery <- readAccessQuery
       rList <- run(
-        sql"select #${columns} from #${existingCollectionName} where _id = ${id.id} and #${accessQuery}"
-          .as[ProjectsRow])
+        sql"select #$columns from #$existingCollectionName where _id = ${id.id} and #$accessQuery".as[ProjectsRow])
```
youri-k, line 89: If you are already adapting the query, you can get rid of the `id.id` as well I think.

app/models/project/Project.scala
```diff
@@ -166,9 +163,15 @@ class ProjectDAO @Inject()(sqlClient: SQLClient)(implicit ec: ExecutionContext)
                           where _id = ${p._id.id}""")
     } yield ()
 
-  def updatePaused(id: ObjectId, isPaused: Boolean)(implicit ctx: DBAccessContext) =
+  def updatePaused(id: ObjectId, isPaused: Boolean)(implicit ctx: DBAccessContext): Fox[Unit] =
     updateBooleanCol(id, _.paused, isPaused)
 
+  def countForTeam(teamId: ObjectId): Fox[Int] =
```
youri-k, line 169: The `countForTeam` seem pretty WET. Maybe we can extract them to a trait or in a utils function.

conf/messages
```diff
@@ -40,6 +40,12 @@ team.notAllowed=You are not part of this team. Project can’t be created.
 team.admin.notPossibleBy=User {1} can’t be assigned administrative rights in team {0} because he doesn't belong to the teams parent team.
 team.admin.notAllowed=You are not authorized to administrate this team.
 team.notFound=Team could not be found.
+team.nameInUse=Team name is or was already in use. Please chose a different name.
+team.delete.organizationTeam=Team cannot be deleted as it is the oganization base team
```
youri-k, line 44: small typo here
```suggestion
team.delete.organizationTeam=Team cannot be deleted as it is the organization base team
```

youri-k (changes requested):
General change looks very good, but please see my small comments :slightly_smiling_face:

app/models/project/Project.scala
fm3: I agree that this duplication is not pretty.

However, I’m afraid my ideas of how to extract them do not really make it better
(1) move `coutForTeam` up as is to SecuredSQLDAO
 → problem: not every table has a team column
(2) move up `count` to SecuredSQLDAO, but pass the query
 → would not reduce duplication by a lot, all three would still have their query
(3) move up `count` to SecuredSQLDAO and pass only where clause as string
 → Would mean that we have to construct where clause as string, as opposed to `sql""` literal
(4) extract `countForTeam` to separate trait to have the three DAOs extend
 → that new trait would have to either extend abstract class SecuredSQLDAO itself (with all that is required there, I couldn’t get it to work) or at least declare both `run` and `existingCollectionName` with their full signatures, which is quite a bit of duplication in its own right

I think the current version is actually preferrable, despite the duplication, as it is the easiest to read. As an aside, I somewhat dislike extracting it also because it is kind of random that those tables have a `_team` column and the responsibility for knowing the columns should be with the DAOs. One may want to change this query for only one of the three in the future.
Do you think this is fair?

youri-k: Yes, I agree. This seems fair :slightly_smiling_face:

youri-k (approved):
LGTM :shipit:

PR #5113, opened 2021-02-01
Fix statistics overview graph error

### LINKED ISSUE #4779: Another error on Statistics > Overview page
May be related to #4638. On 20.09.0, same way to trigger the error (clicking on points on the graph). Error message is different this time:

`An unknown error occurred. Please consider refreshing this page to avoid an inconsistent state. Error message: TypeError: Cannot read property '0' of undefined`.

grittaweisheit:
This PR fixes the occurrence of an error when clicking on the graph of the statistics overview page.

- https://___.webknossos.xyz

### Steps to test:
- go to statistics > overview
- click on a point of the Overall Weekly Annotation Time (hours per week) graph 
- there should be no error

### Issues:
- fixes #4779

------
- [x] Ready for review


frontend/javascripts/admin/statistic/statistic_view.js
```diff
@@ -90,7 +90,7 @@ class StatisticView extends React.PureComponent<{}, State> {
 
   selectDataPoint = ({ chartWrapper }: GoogleCharts) => {
     const chart = chartWrapper.getChart();
-    const indicies = chart.getSelection().selection[0];
+    const indicies = chart.getSelection()[0];
```
philippotto, line 93: Could you test this in multiple browsers? I'm wondering whether the library exposes an inconsistent behavior somewhere. Since 2019-02-14 we use `"react-google-charts": "^2.0.0"` and on 2020-06-23, @MichaelBuessemeyer changed this exact line (see https://github.com/scalableminds/webknossos/commit/fe93f6615fa4cf14971a1050170c8210be9e24eb#diff-823e1a6f8f8ef290cfeec2ab2219cc92dea4da813f1c1717b833b70346953b75R93). Now your change would basically revert his change :smile: I assume you both had good reasons to change the code, but they are contradicting each other. I'm afraid that we might get another bug report soon. So, it would be good to find out what's going on. Testing multiple browsers might be a good place to start :thinking: 

Maybe it also depends on where one clicks exactly or something like that?

@MichaelBuessemeyer Do you know more about this?

philippotto (commented):
Awesome :+1:  /edit: see my other comment :see_no_evil:

frontend/javascripts/admin/statistic/statistic_view.js
MichaelBuessemeyer: I just talked with Gritta about this change / bug and it looks like her versions is correct. Maybe I messed this up last year :sweat_smile: :swan:? 

We checked the fix in firefox and chromium. It would be nice if oucld also check the fix in the browsers you have installed @philippotto :robot:

philippotto (approved):
Thanks for doublechecking this! I also tested again with firefox and chrome and it seems to work. Let's hope this is settled then :)

PR #5119, opened 2021-02-02
Fix nightly

philippotto:
I don't really know why this changed, but the nightly often crashed when rendering the roi2017 dataset with the axons mapping. I noticed that the mapping data is quite large (> 200 MB) which is why I changed the test to use the much smaller astrocyte mapping (~ 20 MB). This seems to have helped :tada: 

### Steps to test:
- see https://app.circleci.com/pipelines/github/scalableminds/webknossos/2305/workflows/cb251280-5156-4729-9592-71fabb64ed07/jobs/13515


daniel-wer (approved):
Nice, that seems like a very good solution! :)

PR #5123, opened 2021-02-03
Fix Lookup for Initial Datastore

### LINKED ISSUE #5120: Investigate localhost datastores url
On wK startup, the localhost datastore is checked, and re-inserted if missing.
If there is one but its name is not localhost, the re-insertion is attempted and triggers an SQL Error.
This lookup should not be by name, but uri

fm3:
When inserting initial data on startup, wK checks if there is already a local datastore and tracingstore. If so, none are inserted.
This check was by name (`"localhost"`), which is the default name for these. However, in production this name may differ, causing this check to be falsely negative, resulting in an SQL error. A better way to identify the local datastore in the db is by using its `url` field.
I also did some minor refactoring as suggested by the IDE.

### Steps to test:
- start with fresh schema, local datastore and tracingstore should be inserted
- on second restart, they should not be inserted again
- in the db, change their name to something else
- on another restart, they should still not be inserted again

### Issues:
- fixes #5120

------
- [x] Ready for review


youri-k (approved):
LGTM :shipit:

PR #5124, opened 2021-02-03
Ignore Carriage Return When Diffing Schemas

### LINKED ISSUE #3463: ignore line-feed inconsistencies in diff schema script

fm3:
On some platforms the output of dumping the DB schema has trailing `\r` characters. This should not trigger the schemas-do-not-match error.

### Steps to test:
- Convert a file from the dumped SQL schema to dos endings
- Diffing output should still be empty (if otherwise identical)

### Issues:
- fixes #3463 

------
- [x] Ready for review


jstriebel (approved):
LGTM :+1:

PR #5131, opened 2021-02-04
Adapt nightly to dropping knossos-support

philippotto:
I removed the knossos-related test cases and added a wkw dataset with RGB channels. I think this should cover everything which was tested before.


daniel-wer (approved):
Awesome :tada: 

Did you want to trigger the nightly here, but did so in the welcome branch PR, instead?

philippotto:
> Did you want to trigger the nightly here, but did so in the welcome branch PR, instead?

Whoops, that explains it :laughing: It's running now..

Issue #5057, opened 2021-01-19
Drop target is small


normanrz:
When a list (e.g. the annotation list) has little entries and thus is short, the drop target is also small and doesn't extend to the whole page.
<img width="1493" alt="Screenshot 2021-01-19 at 10 10 22" src="https://user-images.githubusercontent.com/335438/105012885-f207c600-5a3e-11eb-8d17-409baa6be73c.png">

Issue #4909, opened 2020-11-02
Failing to start job blocks user


youri-k:
If this line fails for whatever reason, the user is stuck on the upload screen with a spinner preventing further action.
https://github.com/scalableminds/webknossos/blob/5c06706676ec9a013239e57ce7ff1dff9ad8c43e/frontend/javascripts/admin/dataset/dataset_add_view.js#L63

PR #5137, opened 2021-02-04
Upload multiple dataset files

### LINKED ISSUE #5033: consider upload of multiple tif files to import
can we drag multiple tif files into wk to import them?

fm3:
- https://___.webknossos.xyz

### Steps to test:
- if you test on an instance without job support, you can still set `jobsEnabled=true` in `application.conf`. an error toast will appear when wk tries to schedule the conversion.
- open the dataset import view and experiment with various inputs for the upload zone:
  - dragging a zip should analyze whether the zip contains wkw files. if this is not the case, the "Scale" textbox will appear.
  - dragging multiple zips should show a validation error
  - dragging image files should show the "Scale" textbox, too
  - dragging wkw files shouldn't show the scale box
  - mixing wkw and image files should show a validation error
  - dragging tar or rar files should show an error
- try to actually upload some files
  - DnD should support dragging one file, multiple files and folders
  - opening the file picker should allow to select one or multiple files (but no folders)
  - if these (zipped) files are wkw, the provided dataset name should be used to create a new dataset folder in `binaryData`
  - if the files need conversion, `binaryData/sample_organization/.forConversion` should contain a new folder (named after your provided dataset name) which includes the uploaded files
- if more than 10 files are attached to the form, the file list should be compact (only one line per file; possible with a scrolling overflow). otherwise, the second line will show the relative path.
- removing a file via the "remove" action should work

image

compact view:
image

### To Do Backend
- [x] unpack dataset in backend (auto-detect if unzip or move)
- [x] move dataset to destination in backend
- [x] adapt sample dataset import
- [x] adapt cleanUpOrphanFileChunks
- [x] backend: check if all filles have arrived
- [x] error handling (see compiler warnings)
- [x] scala Path vs java Path vs File vs Directory
- [x] ensure no conflict with #5156 (dataset filesize report)
- [x] double-check that files with equal names are not deduplicated if their paths differ
- [x] remove logging

### To Do Frontend
- [x] in resumable upload, pass upload id *and* file id / path
- [x] use complete / progress events instead of fileSuccess / fileProgress
- [x] merge master (especially the changes from #5081
- [x] adapt UI to accept both multiple files and directories (might not work with a single antd element)
- [x] ~defer content checks until upload button is hit ? (files may be dropped one by one)~ I decided to re-validate on every file change. otherwise, one would think that everything is good to go and when hitting "upload" one gets a red warning saying that the scale has to be added. I like it better when one gets instant feedbac.
- [x] show abbreviated file list if too many files
- [X] show relative path in file list? ~unfortunately, this doesn't seem to be possible with antd 3.~ with react-dropzone this works. however, I only added this when fewer than 10 items are present.
- [x] add explanatory text?
- [x] check that progress bar correctly shows progress across all files
- [x] clean up

### Issues:
- fixes #5033 

------
- [x] Ready for review


philippotto:
@fm3 I just tried to upload ~ 50 files and got:

```
2021-02-22 17:37:42,751 [INFO] com.scalableminds.webknossos.datastore.services.UploadService - for file heart-118707-200 (79th copy).png expected 1 chunks, found 1
```

An error toast also showed `dataSet.import.incomplete.missingFiles`. Do you know what might be the cause for this?

fm3:
Let’s try that together tomorrow :) the Info logging should be safe to ignore (expected 1 got 1 sounds about right)) missingFiles sounds like the number of files received/registered by the server is not equal to the number of files the front-end announced 🤔

philippotto:
whew, I think I'm done (literally) with this PR. react-dropzone simplified things a lot, even though I had to re-implement some stuff. Afterwards, I was able to add some polish, such as more data validation.

I added an extensive test plan above and also tested these things myself (apart from the end-2-end part with the actual conversion, but this should be independent of this PR).

philippotto:
I still have to fix the NML upload which was broken by updating the react-dropzone library.

philippotto:
> I still have to fix the NML upload which was broken by updating the react-dropzone library.

This is done now. I followed this changelog: https://github.com/react-dropzone/react-dropzone/releases/tag/v8.0.0

Also I fixed #5057 because I noticed that dragging a file at the lower part of the page didn't work if that part was empty.

youri-k (changes requested):
Backend code LGTM

However, when testing without enabling jobs, removing all uploaded files leads to an error message that the files aren't converted to wkw.

frontend/javascripts/admin/dataset/dataset_upload_view.js
```diff
@@ -73,11 +75,13 @@ class DatasetUploadView extends React.PureComponent<PropsWithFormAndRouter, Stat
     (this.props.activeUser.isAdmin || this.props.activeUser.isDatasetManager);
 
   normFile = e => {
-    // Ensure that only one dataset can be uploaded simultaneously
+    console.log("normFile returns:", Array.isArray(e) ? e : e && e.fileList);
```
daniel-wer, line 78: Can this be removed?

frontend/javascripts/admin/dataset/dataset_upload_view.js
```diff
@@ -446,6 +506,128 @@ class DatasetUploadView extends React.PureComponent<PropsWithFormAndRouter, Stat
...
+  borderColor: "#00e676",
+};
+
+const rejectStyle = {
+  borderColor: "#ff1744",
+};
+
+function FileUploadArea({ fileList, onChange }) {
+  const onDropAccepted = acceptedFiles => {
+    onChange(_.uniqBy(fileList.concat(acceptedFiles), file => file.path));
```
daniel-wer, line 541: In the admin_rest_api you did something like `file.path || file.webkitRelativePath || file.relativePath || file.name` to get a file identifier. Is `file.path` actually always defined?

frontend/javascripts/admin/dataset/dataset_upload_view.js
```diff
@@ -446,6 +506,128 @@ class DatasetUploadView extends React.PureComponent<PropsWithFormAndRouter, Stat
...
+  const showSmallFileList = files.length > 10;
+  const list = (
+    <List
+      itemLayout="horizontal"
+      dataSource={acceptedFiles}
+      size={showSmallFileList ? "small" : "default"}
+      renderItem={item => (
+        <List.Item
+          actions={[
+            <a key="list-loadmore-edit" onClick={() => removeFile(item)}>
```
daniel-wer, line 572: Shouldn't these List.Items all have a different key?

frontend/javascripts/admin/dataset/dataset_upload_view.js
```diff
@@ -446,6 +506,128 @@ class DatasetUploadView extends React.PureComponent<PropsWithFormAndRouter, Stat
...
+    () => ({
+      ...baseStyle,
+      ...(isDragActive ? activeStyle : {}),
+      ...(isDragAccept ? acceptStyle : {}),
+      ...(isDragReject ? rejectStyle : {}),
+    }),
+    [isDragActive, isDragReject, isDragAccept],
+  );
+
+  const files = acceptedFiles.map(file => <li key={file.path}>{file.path}</li>);
```
daniel-wer, line 561: See my comment above regarding `file.path`.

daniel-wer (commented):
Frontend LGTM, nice work! I added some smaller questions and will report back after testing :)

daniel-wer:
Works really well :tada: The validation is very helpful.

Two things:
- I tried uploading a data.zip (or as a folder without zipping) which looks like:
  ```
  data.zip
    1 ->
      z0 ->
        y0 ->
          x0.wkw
      header.wkw
  ```
  which cannot be imported, because the resolution folder is not found. Looks like wK assumes that the parent folder is zipped/dropped for the upload. I'm wondering whether this could be handled better, but it may not be that important and is not a blocker.
- What do you think about always including the grayed out path and black file name as a single line. Otherwise when having more than 10 entries it gets confusing with all those x0.wkw and header.wkw from the different resolutions. It's hard to get an overview of what has been dropped. It could look like this (with the small icon instead) and would get rid of the `showSmallFileList` switch:
path_and_name

frontend/javascripts/messages.js
```diff
@@ -209,8 +209,8 @@ instead. Only enable this option if you understand its effect. All layers will n
   "task.no_tasks_to_download": "There are no tasks available to download.",
   "dataset.upload_success": "The dataset was uploaded successfully.",
   "dataset.upload_failed": "The dataset upload failed.",
-  "dataset.upload_none_zip_error":
-    "It looks like the selected file is not a zip file. WebKnossos only supports uploading zipped datasets. Please ensure that your dataset is compressed to a zip archive.",
+  "dataset.unsupported_file_type":
+    "It looks like the selected file is not supported. WebKnossos only supports uploading zipped WKW datasets or image files.",
```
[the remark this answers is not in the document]
fm3: Please note that right now only tiff stacks are supported (until https://github.com/scalableminds/webknossos-cuber/pull/268 is finished and integrated into https://github.com/scalableminds/webknossos-worker/blob/master/tasks.py

fm3:
> removing all uploaded files leads to an error message that the files aren't converted to wkw.

@youri-k Thanks for having a look! Could you clarify, what you mean by removing all uploaded files? At what point does the error crop up? What would be the expected behavior?

frontend/javascripts/admin/dataset/dataset_upload_view.js
philippotto: Hmm, doesn't seem to be necessary. There is no warning in the console. I assume that the `<List>` component adds the key property after calling `renderItem` :thinking:

frontend/javascripts/admin/dataset/dataset_upload_view.js
philippotto: Good point. I dug a bit into this and it turns out that `react-dropzone` "enriches" the `File` object with the `path` attribute. See https://github.com/react-dropzone/file-selector/blob/1e23f600ade799c378cc0a65189708f68183cec8/src/file.ts#L18. Therefore, `path` should always exist. However, just to be on the safe side, I changed the code to `file.path || file.name` for both usages of `file.path` and added a comment which explains the rationale.

philippotto: Ok, I changed my mind a bit, since there are more occurences of `file.path` and I didn't want to include the fallback everywhere. I removed the `|| file.name` fallback completely now and added one check to the resumable-upload-creator which throws an exception in case `file.path` doesn't exist, anymore. Theoretically, this shouldn't happen, but since this "enrichment-feature" of react-dropzone seems to be very subtle (and not really documented), I wanted to integrate this into our code base somehow.

frontend/javascripts/messages.js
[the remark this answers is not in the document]
philippotto: I left it vague on purpose, since I didn't want to introduce a dependency to the cuber (if the cuber supports a new format, this would need to be updated here). Instead, I'd rather focus on getting the `cuber` to support 99% of the use cases. But feel free to voice a different opinion :)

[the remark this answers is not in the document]
philippotto: Theoretically, one would also need to explain which file structures/hierarchies are supported. This opens up a can of worms in my opinion.

philippotto:
> * I tried uploading a data.zip (or as a folder without zipping) which looks like:
>   ```
>   data.zip
>     1 ->
>       z0 ->
>         y0 ->
>           x0.wkw
>       header.wkw
>   ```
>   
>   
>   which cannot be imported, because the resolution folder is not found. Looks like wK assumes that the parent folder is zipped/dropped for the upload. I'm wondering whether this could be handled better, but it may not be that important and is not a blocker.

This will probably/hopefully be improved automatically as soon as @youri-k's work on the auto-conversion in the cuber lands. So, I wouldn't worry about it for now.

> * What do you think about always including the grayed out path and black file name as a single line. Otherwise when having more than 10 entries it gets confusing with all those x0.wkw and header.wkw from the different resolutions. It's hard to get an overview of what has been dropped. It could look like this (with the small icon instead) and would get rid of the `showSmallFileList` switch:

Good idea :+1: I made sure to always include the path as you suggested. However, I kept the large/small-icon-switch in the code, since one earlier issue with the upload-UI was that dropping only one file was very subtle in its appearance.

Now it looks like this:

image
and:
image

frontend/javascripts/messages.js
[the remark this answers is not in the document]
daniel-wer: Yes that makes sense, this error message might not be the best spot for this. Still, from a user perspective, I think it would be very helpful to see the supported formats somewhere. Ideally, the cuber should be documenting this and wK could link there. If such a page doesn't exist yet, we can add and link to it later.

fm3:
> This will probably/hopefully be improved automatically as soon as @youri-k's work on the auto-conversion in the cuber lands. So, I wouldn't worry about it for now.
In the wkw case, the cuber is not triggered, so this problem will persist. It is, however, unrelated to this PR. I created follow-up issue https://github.com/scalableminds/webknossos/issues/5219 for that

youri-k:
> Could you clarify, what you mean by removing all uploaded files? At what point does the error crop up? What would be the expected behavior?

The error cropped up after adding files and then removing every file from the list. My expectation would be that the validation asks for files to be uploaded, but not giving me an error that the file format isn't understood. I can try to produce a GIF if it's necessary.

fm3:
Ah ok thanks, I can reproduce it now!
This is probably the frontend validation, @philippotto 
(1) Drop myWKWDataset.zip 
(2) Then click remove in the Files list
(3) List is now empty, warning modal pops up »does not seem to be in the WKW format«
This should only be triggered if there is at least one file selected.

philippotto:
Good find! Should be fixed now.

fm3:
@youri-k and probably also @daniel-wer Could you have another look? If I understand correctly, all feedback has been implemented :)

daniel-wer (approved):
Frontend LGTM and @youri-k's issue has also been fixed, nice work :+1:

philippotto (changes requested):
I just noticed that all the shortcuts are broken in the tracing view (e.g., d, f, space etc). Probably because the dropzone container catches these events somehow. I have to investigate a fix for this before we can merge it.

daniel-wer:
> I just noticed that all the shortcuts are broken in the tracing view (e.g., d, f, space etc). Probably because the dropzone container catches these events somehow. I have to investigate a fix for this before we can merge it.

Uff, nice catch!

fm3:
Do you know if this PR might also fix #4909?

philippotto:
> Do you know if this PR might also fix #4909?

I don't think so, but I can have a look at it.

philippotto:
> I just noticed that all the shortcuts are broken in the tracing view (e.g., d, f, space etc). Probably because the dropzone container catches these events somehow. I have to investigate a fix for this before we can merge it.

Should work now. Was a one-line fix :man_dancing: 



> > Do you know if this PR might also fix #4909?
> 
> I don't think so, but I can have a look at it.

I fixed this by properly closing the modal and showing an error toast which encourages the user to re-try or contact us. I think this is good enough. It's probably more important to make the job submissions stable, anyway.

youri-k (approved):
Now everything works :tada:

philippotto (approved):

PR #5142, opened 2021-02-05
Avoid more null-errors in isosurface and mesh handling

### LINKED ISSUE #4935: Importing mesh in view mode doesn't work
How to reproduce:
- Open this dataset https://webknossos.org/datasets/scalable_minds/l4dense_motta_et_al_demo/view#2850,4318,1770,0,1.000
- Open `Meshes` tab
- Import this mesh: 
[5918_simplified.stl.zip](https://github.com/scalableminds/webknossos/files/5523486/5918_simplified.stl.zip)
- Boom 
Screenshot 2020-11-11 at 11 38 36

### LINKED ISSUE #5106: "Load Isosurface for centered cell" throws error when there is no segmentation layer
"Load Isosurface for centered cell" throws error when there is no segmentation layer available in the dataset.

Either hide the isosurface UI or disable the "load isosurface" (and probably the import button). I tend to the latter option, because this has the benefit that the user will still see that there is a feature when segmentation data is available.

Related: https://github.com/scalableminds/webknossos/issues/4935

philippotto:
The meshes UI could provoke errors when viewing a dataset without segmentation layer or when update-permissions are missing (for example, in view mode).

- https://___.webknossos.xyz

### Steps to test:
- I tested the cross-product of the following factors:
  - dataset with and without segmentation layer (I used ROI2017_wkw and data_types)
  - open dataset in view mode, with a hybrid and with a skeleton-only annotation
- In each constellation:
  - make sure the "load centered isosurface" behaves correctly (it should only be disabled if there is no segmentation layer). if the centered cell is 0, the button is available, but should emit a warning toast.
  - the import buttons are only enabled if allowUpdate is true
  - importing an meshes/isosurface STLs should "work as expected". see [here](https://www.notion.so/scalableminds/Test-Data-c0563be9c4a4499dae4e16d9b2497cfb#f3ca867baabf43cca7f562ec16039a5a) for sample STLs.

### Issues:
- fixes #5106
- fixes #4935

------
- [X] Ready for review


daniel-wer (approved):
LGTM :+1: 

In #4935 you wrote:

> Related: this error only appears in a very obscure report in airbrake: https://scm.airbrake.io/projects/95438/groups/2835680766133594723?tab=notice-detail
> 
> Instead of [object Object], there should be a proper error message. I suggest to investigate this, too, before fixing the actual problem, since it's crucial to have robust error reporting.

Were you able to have a look at that as well? :)

philippotto:
> > Related: this error only appears in a very obscure report in airbrake: https://scm.airbrake.io/projects/95438/groups/2835680766133594723?tab=notice-detail
> > Instead of [object Object], there should be a proper error message. I suggest to investigate this, too, before fixing the actual problem, since it's crucial to have robust error reporting.
> 
> Were you able to have a look at that as well? :)

Good point! I investigated this and if the sagas crash, the try-catch-block potentially receives the rejection reason of a promise, which doesn't need to be an error instance. for this case, the default stringification would produce [object object]. I fixed this by applying json.stringify when necessary. It's a bit of a bummer that rejection reasons and errors seem to be mixed in the catch-clause, but I cannot think of a more elegant solution right now.

PR #5143, opened 2021-02-05
Email templates

### LINKED ISSUE #5019: Update email templates

normanrz:
I updated the email templates for new users and new organizations. They are structured after the new welcome header by @philippotto. This PR depends on #5099 to be merged.

### Issues:
- fixes #5019 

------
- [x] Ready for review


app/views/mail/newAdminWKOrg.scala.html
```diff
@@ -1,25 +1,25 @@
...
     </li>
     <li>
-      Upload your own dataset into your own space private workspace. If desired, you can
-      publish your dataset into our public gallery.
+      <a href="https://webknossos.org/datasets/upload">Upload your own dataset</a>. Datasets are private by default and you can securely share them with your collaborators.
     </li>
     <li>
-      Invite colleagues from your lab or external collaborators to work with your data.
-      webKnossos works great for small teams and supports management of large annotator crowds.
+      Watch a short video to <a href="https://webknossos.org/datasets/upload">learn how to create annotations</a> in webKnossos.
```
philippotto, line 18: This URL should probably point to a youtube vide?

app/views/mail/newUserWKOrg.scala.html
```diff
@@ -1,36 +1,28 @@
...
+      There are <a href="https://webknossos.org/publications">many published datasets</a> from the scientific community ready to be explored.
     </li>
     <li>
-      Upload your own dataset into your own space private workspace. If desired, you can
-      publish your dataset into our public gallery.
-    </li>
-    <li>
-      Invite colleagues from your lab or external collaborators to work with your data.
-      webKnossos works great for small teams and supports management of large annotator crowds.
+      Watch a short video to <a href="https://webknossos.org/datasets/upload">learn how to create annotations</a> in webKnossos.
```
philippotto, line 25: Same here.

app/views/mail/newUser.scala.html
```diff
@@ -6,14 +6,12 @@
 
   @if(!enableAutoVerify) {
     <p>
-      Welcome to your new webKnossos account. You are almost ready to start exploring and to collaboratively
-      annotate large 3D datasets.
+      Welcome to your new webKnossos account. You are almost ready to share and annotate your large 3D datasets.
     </p>
     <p>Your account is currently under review. An administrator will activate you soon.</p>
```
philippotto, line 11: Btw: Do we send an email when a user was activated later? Might be unimportant, since the new invite-by-email-feature uses autoEnable, as far as I know, but in general it would make sense..

app/views/mail/newUser.scala.html
```diff
@@ -6,14 +6,12 @@
...
   } else {
     <p>
-      Welcome to your new webKnossos account. Get ready to start exploring and to collaboratively
-      annotate large 3D datasets.
+      Welcome to your new webKnossos account. Get ready to share and annotate your large 3D datasets.
```
philippotto, line 14: This email doesn't seem to contain a link to wk. Might be a good idea to add this? Something similar to ` There are <a href="https://webknossos.org/publications">many published datasets</a> from the scientific community ready to be explored.` maybe..

philippotto (commented):
Great stuff! Left a few comments and in general, this PR probably needs to be merged after #5099 since that PR adds the `publicDemoDatasetUrl` attribute. Alternatively, copy that line change into this PR.

app/views/mail/newUser.scala.html
normanrz: This email is for the self-hosted instances. I don't think it should direct users to wk.org. What do you think?

app/views/mail/newUser.scala.html
normanrz: I don't think there is an activation-notifier. Would be helpful, tho. cc @fm3

app/views/mail/emailBaseTemplate.scala.html
```diff
@@ -44,7 +44,7 @@
     </p>
     <p style="color: #888; font-size: 0.6rem;">
       <a href="https://webknossos.org" style="color: #888;">webknossos.org</a> is a service by
-      <a href="https://scalableminds.com">scalable minds</a> &bull; <a href="mailto:hello@@scalableminds.com" style="color: #888;">hello@@scalableminds.com</a>
+      <a href="https://scalableminds.com">scalable minds</a> &bull; <a href="mailto:hello@@webknossos.org" style="color: #888;">hello@@webknossos.org</a>
```
hotzenklotz, line 47: ❤️

hotzenklotz:
Text sounds good to me.

app/views/mail/newUser.scala.html
philippotto: Ah, good point, then it's alright for me.

app/views/mail/newUser.scala.html
fm3: I think https://github.com/scalableminds/webknossos/blob/master/app/views/mail/validateUser.scala.html is what you are referring to.

app/views/mail/newAdminWKOrg.scala.html
normanrz: I added the youtube-link that is also used in the welcome-header. Should we put the url in the application.conf? cc @fm3

philippotto: Since this URL doesn't need to be instance specific, I'd be fine with not having it in the application.conf. Updating it fast is also not a big priority imo. But either way would be fine with me.

fm3: The welcome header/onboarding uses three different video URIs. Not sure how to (/if we should) move them all to config.

normanrz: ok. then let's keep it hard-coded.

philippotto (approved):

PR #5152, opened 2021-02-10
Fix multiuser email lookup

### LINKED ISSUE #5127: Broken user lookup?
Not sure yet in which context this appears but we got a single occurence notification of this:
```
Notification from webKnossos at https://webknossos.org
org.postgresql.util.PSQLException: Bad value for type int : <aMongoID>:
Causing query: [select id from webknossos.multiusers where email = ? and true]
```

fm3:
When users signed up with an email adress that was already in use, an SQL error was produced. The end user presumably still got the correct error message on the top level, albeit with a strange error tail.

Also in this PR: removed a superfluous multiUser lookup in user update

### Steps to test:
- CI should be green
- Try registering with existing email, should get readable error

### Issues:
- fixes #5127 
------
- [x] Ready for review


youri-k (approved):
LGTM :shipit:

PR #5156, opened 2021-02-11
Send Analytics Events in Backend

### LINKED ISSUE #5134: Implement Metrics tracking
I would like to implement the tracking of events for analytics on the backend. Events consist typically of
- an event type (e.g. `signup`, `open_annotation`, `upload_dataset`, `invite_colleague`),
- a user identifier (e.g. email address or multiuser database id),
- some event properties (e.g. dataset name, annotation id),
- some user properties (e.g. organization) and,
- a timestamp.

These events are sent to an HTTP API of the analytics provider. I think it would be great to have an abstraction for that, so that we could switch to a different provider in the future. All dashboards and metrics are then derived from the stream of events in the analytics tool.

Events to track (initially):

wk internal:
- [x] Signup (+ source of signup: invite vs frontpage)
- [x] Invite colleagues
- [x] Join organization
- [x] Open Dataset
- [x] Create Annotation
- [x] Open Annotation (+ owner of annotation)
- [x] Trigger long-running job
- [x] Download Annotations

triggered from datastore:
- [x] Upload Dataset (+ dataset name, size, type?)
- [x] Change Dataset Settings
- [x] Trigger Isosurface rendering

triggered from tracingstore
- [x] Save Annotation (maybe per update action)

API definition for sending the events: https://developers.amplitude.com/docs/http-api-v2
Example:
```bash
curl --data '{"api_key": "...", "events": [{"user_id": "john_doe@gmail.com", "event_type": "watch_tutorial",
  "user_properties": {"Cohort": "Test A"}}]}' https://api.amplitude.com/2/httpapi
```

Maybe we'll add some frontend tracking in the future as well (e.g. open share modal, select brush tool etc).

fm3:
Send analytics tracking events to Amplitude API, first iteration.

### TODO
- [x] track actions described in #5134
- [x] ensure non-blocking operation
- [x] use json format
- [x] check if dataset upload is correctly tracked in conversion case
- [x] test with live Amplitude system
- [x] create follow-up issue for tracking bad requests (e.g. failed dataset uploads, failed jobs/dataset conversions) → #5168 
- [x] create follow-up issue for monitoring this in action → #5170

### Steps to test:
- set up amplitude url and key in `application.conf` ( `backendAnalytics.key`, `backendAnalytics.uri`)
- do the actions listed in #5134
- tracking events should be sent
- they should not block the user interaction
- to test this locally without spamming the analytics endpoint, the verbose logging can be activated instead of inserting the correct key

### Issues:
- fixes #5134

------
- [x] Ready for review


app/models/analytics/AnalyticsService.scala
```diff
@@ -0,0 +1,145 @@
...
+import org.joda.time.DateTime
+import play.api.libs.json._
+import utils.{ObjectId, WkConf}
+
+import scala.concurrent.ExecutionContext
+
+class AnalyticsService @Inject()(rpc: RPC, wkConf: WkConf, analyticsLookUpService: AnalyticsLookUpService)(
+    implicit ec: ExecutionContext)
+    extends LazyLogging {
+  def note(analyticsEvent: AnalyticsEvent): Unit = {
```
normanrz, line 21: Typcially a method like that is call `track`

conf/webknossos.latest.routes
```diff
@@ -83,8 +83,9 @@ PUT           /datastores/:name/datasource
 PUT           /datastores/:name/datasources                                     controllers.WKDataStoreController.updateAll(name: String)
 PATCH         /datastores/:name/status                                          controllers.WKDataStoreController.statusUpdate(name: String)
 POST          /datastores/:name/verifyUpload                                    controllers.WKDataStoreController.validateDataSetUpload(name: String)
-POST          /datastores/:name/addInitialTeams                                 controllers.WKDataStoreController.addDataSetInitialTeams(name: String, token: String, dataSetName: String)
+POST          /datastores/:name/reportDatasetUpload                             controllers.WKDataStoreController.reportDatasetUpload(name: String, token: String, dataSetName: String, dataSetSizeBytes: Long)
 POST          /datastores/:name/deleteErroneous                                 controllers.WKDataStoreController.deleteErroneous(name: String)
+POST          /datastores/:name/reportIsosurfaceRequest                         controllers.WKDataStoreController.reportIsosurfaceRequest(name: String, token: Option[String])
```
normanrz, line 88: Should this be tracked from the frontend?

normanrz (approved):
lgtm :+1:

conf/webknossos.latest.routes
fm3: Yes I think I’ll move it back to the front-end when implementing #5166

youri-k (approved):
LGTM :shipit:

PR #5162, opened 2021-02-12
add help block

normanrz:
I thought it might be a good idea to add a help message like this, because I think data upload will still be an issue for some users.
Screenshot 2021-02-12 at 21 57 12

Should be merged as part of #5081.


MichaelBuessemeyer:
This is a good idea + thanks for already doing the coding 👍

MichaelBuessemeyer (approved):

PR #5164, opened 2021-02-15
Allow accessing dataset admin view configuration via sharing token

### LINKED ISSUE #3240: Non-public datasets accessed using a token do not honor default settings
This issue is about accessing non-public datasets using the Sharing Link with a token, when logged out.
Under these circumstances the `/api/dataSetConfigurations/<dataset_name>` request does not return the configured default position and zoomstep, which is why the dataset doesn't open at the correct position in the tracing view.

Steps to reproduce:
- Configure a default position for a non-public dataset
- Copy the Sharing Link for that dataset (should include a token, because the dataset is not public)
- Log out and access the dataset using the copied link
- You will not be at the configured default position, but instead at the center of the bounding box

fm3:
### Steps to test:
- give a dataset an admin view configuration (e.g. layer configuration in settings view)
- open private tab, open dataset there
- should see the respective settings

### Issues:
- fixes #3240

------
- [x] Ready for review


philippotto (approved):
LGTM :+1:

PR #5167, opened 2021-02-15
Indicate loading state in tab title during loading

### LINKED ISSUE #5041: Avoid flashing of “Test Dataset” in tab title during page load
During page load when viewing a dataset or annotation, the string “Test Dataset” appears in the tab title until it is replaced by the real dataset name. Instead, the tab title should still just say »webKnossos – orga « in this initializing state, like it does at first for a few ms.

daniel-wer:
... instead of values from default state. This should avoid confusion :)

- https://___.webknossos.xyz

### Steps to test:
- Previously, during loading, the tab title showed default values from the store, namely "Test Dateset" and "Connectomics Department". This should no longer be the case. During loading the tab title should show "Loading" and the correct values, afterwards.
- Trigger an error, for example by deleting some part of the annotation id. The tab title should switch to "Error" after the loading phase, instead of showing the default store values.

### Issues:
- fixes #5041 

------
- [x] Ready for review

### DIFF
--- a/frontend/javascripts/oxalis/view/layouting/tracing_layout_view.js
+++ b/frontend/javascripts/oxalis/view/layouting/tracing_layout_view.js
@@ -212,11 +212,17 @@ class TracingLayoutView extends React.PureComponent<PropsWithRouter, State> {
   };
 
   getTabTitle = () => {
-    const titleArray: Array<string> = [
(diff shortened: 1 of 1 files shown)


grittaweisheit (approved):
That's very neat! :+1:

PR #5174, opened 2021-02-17
Increase Standalone Stores Request Timeout

### LINKED ISSUE #4271: tracingstore requests under load produce error 502
I thought we disabled timeouts, or at least set them very high?
see
https://discuss.webknossos.org/t/cant-load-annotations-here/1533/2

fm3:
Turns out we had those timeouts set to the large value only in `application.conf` but not for the standalone versions of the stores.

From the changelog:
> Changed the default request timeouts for standalone datastores and tracingstores to match those of local ones (10000s instead of 75s)

### Steps to test

I tested it on data1 with an artificial sleep in a request (got error 502 from nginx after 75s before this change, got my reply after the 5min sleep afterwards)

### Issues
- fixes #4271

- [x] ready for review


youri-k (approved):
LGTM :shipit:

PR #5178, opened 2021-02-18
Clean up and unify config, add typing

### LINKED ISSUE #5173: Prune unused config keys from application.conf
There are a couple of unused fields in the config (e.g. braingames.binary.loadTimeout) – we should go through the list and remove the unused ones (or use them again if it makes sense)

fm3:
I cleaned up the config as much as I could without actually changing any keys (which would require changes in the deployments).
 - Remove unused config fields
 - Combine blocks
 - Unify ordering in application.conf and standalone*.conf
 - Add type hints to ConfigReaders

It may be worth to change the config options keys themselves, creating a unified semantics and naming convention, and use Durations instead of integers with unclear time units. I’d move that to a later PR, though, as it requires operators to change how they override parameters for their individual setups.

- https://cleanupconf.webknossos.xyz

### Steps to test:
- CI should run through
- should still be able to trace, login/logout
- Standalone setup should also still run the same

### Issues:
- fixes #5173

------
- [x] Ready for review


webknossos-datastore/app/com/scalableminds/webknossos/datastore/DataStoreConfig.scala
```diff
@@ -7,41 +7,40 @@ import play.api.Configuration
...
+    val uri: String = get[String]("http.uri")
   }
 
   object Braingames {
     object Binary {
       object ChangeHandler {
-        val enabled = get[Boolean]("braingames.binary.changeHandler.enabled")
-        val tickerInterval = get[Int]("braingames.binary.changeHandler.tickerInterval") minutes
+        val enabled: Boolean = get[Boolean]("braingames.binary.changeHandler.enabled")
+        val tickerInterval: FiniteDuration = get[Int]("braingames.binary.changeHandler.tickerInterval") minutes
```
jstriebel, line 20: Why would this not be inferred automatically? Just curious :cat2:

conf/application.conf
```diff
@@ -1,68 +1,49 @@
...
-  }
-  redis {
-    address = "localhost"
-    port = 6379
-  }
-}
+  # Unset some of the security filters enabled in datastore.Filters
+  filters.headers.contentSecurityPolicy = null
+  filters.headers.contentTypeOptions = null
+  filters.headers.frameOptions = null
```
jstriebel, line 25: It's not clear to me yet which parts should be grouped and which not. E.g. why not the following?
```suggestion
  filters.headers {
    # Unset some of the security filters enabled in datastore.Filters
    contentSecurityPolicy = null
    contentTypeOptions = null
    frameOptions = null
  }
```

conf/application.conf
```diff
@@ -1,68 +1,49 @@
...
 
-http {
-  uri = "http://localhost:9000"
-  address = "127.0.0.1"
-  port = 9000
+  # Timeouts. Note that these take effect only in production mode (timeouts are shorter in dev)
+  server.http.idleTimeout = 10000s
+  server.akka.requestTimeout = 10000s
```
jstriebel, line 29: Same as above

conf/application.conf
```diff
@@ -99,14 +86,41 @@ features {
...
+    uri = "http://localhost:9000"
+    pingIntervalMinutes = 10
+  }
+  fossildb {
+    address = "localhost"
+    port = 7155
+  }
+}
 
-# Binary data settings
```
jstriebel, line 108: binary data :trollface:

jstriebel (approved):
Nice, thanks for arranging this so nicely! Mostly LGTM, just minor questions, so feel free to merge as-is :+1:

webknossos-datastore/app/com/scalableminds/webknossos/datastore/DataStoreConfig.scala
fm3: You’re right, it is inferred automatically, which is why the previous version compiled too. However, Scala style recommendation is to explicitly type-annotate all public members, both for readability and for compiler speed-up.

conf/application.conf
fm3: Good point! Let’s group all the things (that have more than one item)

PR #5193, opened 2021-02-21
Fix public annotation sharing

### LINKED ISSUE #5192: Public annotation links broken?
I’m getting 404 even though the annotation visibility is set to public.
Note that the error does not include the typical “annotation not found” but rather the generic “resource not found” message, indicating that the error may occur later in the code when looking up subsequent information about the annotation.

fm3:
Fixes a bug where publicly shared annotations could not be loaded because the analytics call in the info request was wrapped in a `map` the result of which was flattened into the for comprehension. 

This PR changes the for comprehension so it ignores the result of the `map` (which is `None` if `request.identity` is `None`).

I also scanned the code for other occurrences of `request.identity.map` but the others were already correctly ignored.

The Bug was introduced in #5156 three days ago. It has not been included in a release.

- https://fixpublicannotations.webknossos.xyz

### Steps to test:
- Set an annotation to public (don’t forget to hit save), copy link
- log out
- open link, you should see the annotation

### Issues:
- fixes  #5192

------
- [x] Ready for review

### DIFF
--- a/app/controllers/AnnotationController.scala
+++ b/app/controllers/AnnotationController.scala
@@ -68,7 +68,7 @@ class AnnotationController @Inject()(
             timeSpanService.logUserInteraction(timestamp, user, annotation) // log time when a user starts working
           } else Fox.successful(())
         }
-        _ <- request.identity.map { user =>
+        _ = request.identity.map { user =>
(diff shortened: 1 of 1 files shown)


youri-k (approved):
Thanks for spotting that 🎉

PR #5194, opened 2021-02-22
Fix Dataset Tiff Conversion

fm3:
The finishUpload route needs a valid value for needsConversion to unzip to the correct target directory.
needsConversion is no longer part of formValues, so this was undefined

### Steps to test:
- set up webknossos-worker, enable jobs in application.conf
- upload zipped tiff dataset

------
- [x] Ready for review


philippotto (approved):
:pray: 

Thanks for doublechecking, investigating and fixing this :+1: I meant to test this on demo, too, but forget about it last week -.-

PR #5195, opened 2021-02-22
Tiff Export Job

### LINKED ISSUE #4911: Export Bounding Box to Tiff
Now that we have support for long-running jobs, it would be useful to also have the possibility to generate stuff that can be downloaded. A good first case for this would be a dataset-to-tiff-export. wkcuber can be used to do the actual conversion. Probably, the output needs to be zipped and then somehow the UI should initiate a download for this.

fm3:
Lets the user download a zipfile with tiff files, exported from wkw on disk.

- New api routes `api/jobs/run/tiffExport/:organizationName/:dataSetName/:layerName/:bbox` (with bbox being in format x,y,z,w,h,d)
- `api/jobs/downloadExport/:organizationName/:exportFileName` (jobs json may now include exportFileName, I already adapted api flow types)

### Todo: 
- [x] create new task in webknossos-worker (see https://github.com/scalableminds/webknossos-worker/pull/5 )
- [x] create api route to call that job
- [x] expose exported files to download
- [x] adapt jobs list to show this download link
- [x] Create UI that calls the run api
- [x] Create follow-up issue for expiry + cleanup job → #5202
- [x] Create follow-up issue for exporting volume annotations → #5203

### Steps to test:
- Set up https://github.com/scalableminds/webknossos-worker/pull/5 (Will be merged into master there once this is approved)
- set `features.jobsEnabled = true` in `application.conf`
- Create annotation, create user bounding box
- In user bounding box list, you should see download icons, opens modal
- select layer, see that job is running in jobs page
- when complete, download zip, should contain selected data range

### Issues:
- fixes #4911 

------
- [x] Ready for review


fm3:
@philippotto Would you be interested in discussing how the front-end for this could look like? :)

frontend/javascripts/oxalis/view/settings/export_bounding_box_modal.js
```diff
@@ -0,0 +1,97 @@
...
+
+const ExportBoundingBoxModal = ({ destroy, dataset, boundingBox }: Props) => {
+  const [startedExports, setStartedExports] = useState([]);
+
+  const handleClose = () => {
+    destroy();
+  };
+
+  const handleStartExport = layerName => {
+    console.log("start export for", layerName);
```
philippotto, line 23: can be removed

frontend/javascripts/oxalis/view/settings/export_bounding_box_modal.js
```diff
@@ -0,0 +1,97 @@
...
+  );
+
+  const downloadHint =
+    startedExports.length > 0 ? (
+      <p>
+        Go to{" "}
+        <a href="/jobs" target="_blank">
+          Jobs Overview Page
+        </a>{" "}
+        for running exports and to download the results.
```
philippotto, line 64:
```suggestion
        to see running exports and to download the results.
```

sounds clearer to me. otherwise, `running` might be mistaken as an imperative (i.e., go there if you want to run sth).

frontend/javascripts/oxalis/view/settings/export_bounding_box_modal.js
```diff
@@ -0,0 +1,97 @@
...
+      </p>
+    ) : null;
+
+  console.log(dataset.dataSource.dataLayers);
```
philippotto, line 68: remove before merge

frontend/javascripts/oxalis/view/settings/export_bounding_box_modal.js
```diff
@@ -0,0 +1,97 @@
...
+    <Modal
+      title="Export Bounding Box as Tiff Stack"
+      onCancel={handleClose}
+      visible
+      width={500}
+      footer={[
+        <Button key="close" type="primary" onClick={handleClose}>
+          close
+        </Button>,
+      ]}
```
philippotto, line 82: you can omit this. the modal should have a closing-x, anyway.

frontend/javascripts/oxalis/view/settings/export_bounding_box_modal.js
```diff
@@ -0,0 +1,97 @@
...
+    >
+      <p>
+        Data from the selected bounding box at {bboxText} will be exported as tiff stack zip
```
philippotto, line 85:
```suggestion
        Data from the selected bounding box at {bboxText} will be exported as a tiff stack zip
```

frontend/javascripts/admin/job/job_list_view.js
```diff
@@ -138,19 +138,37 @@ class JobListView extends React.PureComponent<Props, State> {
...
+                            job.exportFileName
+                          }`}
+                          title="Download"
+                        >
+                          <Icon type="download" />
+                          Download
+                        </a>
+                      )}
+                    </span>
+                  </div>
```
philippotto, line 171: I would prefer if the case-by-case distinction would be done via `job.type`. For example something like this:

```
if (job.type === "cubing) {
  if (!job.datasetName) {
    return null;
  }
  return ...
} else if (job.type === "convert_to_tiff") {
  return ...
}
```

youri-k (approved):
Code LGTM :shipit:

frontend/javascripts/oxalis/view/settings/setting_input_views.js
```diff
@@ -338,20 +341,26 @@ export class UserBoundingBoxInput extends React.PureComponent<UserBoundingBoxInp
...
-          <Col span={22}>
+          <Col span={20}>
             <Switch
               size="small"
               onChange={this.handleVisibilityChange}
               checked={isVisible}
               style={{ margin: "auto 0px" }}
             />
           </Col>
+          <Col span={2}>{exportButton}</Col>
```
philippotto, line 363: this allocates space for the button even if it isn't rendered. I'd put the column into the variable (and rename it to `exportColumn`). the `span` property of the first `Col` element could be switched between 20 or 22 depending on `exportColumn != null`.

philippotto (commented):
Awesome! Looks quite cool. I'll give it a test now :)

philippotto:
I just tested the feature and after some hiccups with my wk-worker-setup it worked :tada: It would be very cool if the errors would bubble up into the jobs-page. However, a proper error in the worker-stdout would already go a long way (it just said `process.CalledProcessError: Command '['python', '-m', 'wkcuber.export_wkw_as_tiff', '--layer_name', 'color', '--bbox', '500,500,500,100,100,100', '--source_path', '/data/sample_organization/ROI2017_wkw', '--destination_path', '/data/sample_organization/.converting/20
21-02-55_15-22__ROI2017_wkw__color__h9LBAgVIiNeb/tif']' returned non-zero exit status 1`). First, the file mount was wrong and then I had a permission problem. Obviously, that was a setup mistake on my part, but maybe we can improve the error handling a bit here.

During testing I noticed two things:
- there doesn't seem to be a limit on the bounding box size. shouldn't we disable the export button if the bbox is larger than X voxels?
- the name of the downloaded zip archive could look a bit nicer. for me it was `2021-02-55_15-27__ROI2017_wkw__color__AEVEq1j1JdfR.zip`. maybe we could drop the random postfix?
- the layer name doesn't appear in the job description. if one exports multiple layer, one has to click "download" to see which one is which:
image

fm3:
Thanks for having a look!

>  However, a proper error in the worker-stdout would already go a long way

I could not yet figure out how to do this, but I’ll look again. I agree that it would be very valuable.


> there doesn't seem to be a limit on the bounding box size. shouldn't we disable the export button if the bbox is larger than X voxels?

Yes, good point! What do you think would be a sensible default for this limit? (Also cc @normanrz )

> maybe we could drop the random [filename] postfix?

I added it to avoid name collisions. Do you think it is safe to remove it? I guess we can add seconds to the timestamp to reduce the risk.

philippotto:
> > However, a proper error in the worker-stdout would already go a long way
> 
> I could not yet figure out how to do this, but I’ll look again. I agree that it would be very valuable.

Since my stdout showed a thrown `process.CalledProcessError`, it's probably a good approach to catch that (and potentially re-raise? don't know how the protocol with the worker looks in error cases). This post also shows how to extract the actual output. I hope this works? https://stackoverflow.com/a/47185863

> > there doesn't seem to be a limit on the bounding box size. shouldn't we disable the export button if the bbox is larger than X voxels?
> 
> Yes, good point! What do you think would be a sensible default for this limit? (Also cc @normanrz )

How about 1024**3 Vx ~ 1 GB?



> > maybe we could drop the random [filename] postfix?
> 
> I added it to avoid name collisions. Do you think it is safe to remove it? I guess we can add seconds to the timestamp to reduce the risk.

You mean collisions on the server's file system, right? Ideally, the postfix would stay there, but the default name when downloading wouldn't include this. I think, this is possible by setting a HTTP header à la: `header('Content-Disposition: attachment; filename="2021-02-55_15-27__ROI2017_wkw__color.zip"');`

normanrz:
> Yes, good point! What do you think would be a sensible default for this limit? (Also cc @normanrz )

Initially, I was thinking of (1024vx)**3

normanrz:
Btw is it downloading all layers (including segmentation)?

philippotto:
> Btw is it downloading all layers (including segmentation)?

the user can select the layers they want to download. downloading volume is not supported yet (since that part has to go via the tracingstore). I think, @fm3 suggested to do this in a separate PR.

fm3:
Note to self: Todo:
- [ ] restrict bbox size (both volume and edge length)
- [ ] fix flow + eslint
- [ ] make sure that cubing jobs also still work

fm3:
@philippotto Thanks for your feedback! I think I solved all raised issues, could you have another look?

frontend/javascripts/oxalis/view/settings/export_bounding_box_modal.js
```diff
@@ -0,0 +1,123 @@
...
+    return nameIfColor || nameIfVolume;
+  });
+
+  const exportButtonsList = layerNames.map(layerName =>
+    layerName ? (
+      <p>
+        <Button
+          key={layerName}
+          onClick={() => handleStartExport(layerName)}
+          // disabled={startedExports.includes(layerName)} //TODO re-enable before merge, this is just to make development easier
```
philippotto, line 48: todo

philippotto (approved):
Looks great! :) See my few comments before merging.

frontend/javascripts/admin/job/job_list_view.js
```diff
@@ -81,6 +82,64 @@ class JobListView extends React.PureComponent<Props, State> {
...
+      );
+    } else return null;
+  };
+
+  renderState = (__: any, job: APIJob) => {
+    const stateString = _.capitalize(job.state.toLowerCase());
+    if (job.state === "SUCCESS") return stateString;
+    else {
+      return (
+        <Tooltip title="Something whent wrong when executing this job. Feel free to contact us if you need assistance.">
```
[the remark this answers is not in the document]
fm3: I’m unsure about that. The address may change and is only right for our demo instance. I’d say the contact link in the navbar should be sufficient. Do you agree?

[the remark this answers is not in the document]
philippotto: yes, you are probably right :+1:

PR #5196, opened 2021-02-22
Add snapshot test: public annotations viewable while logged out

### LINKED ISSUE #5191: Add test for public annotation sharing
Snapshot tests should include a test of public annotation sharing (can a logged-out viewer still see the annotation)

fm3:
This was broken for three days, fixed in #5193 – but we should have noticed sooner.

### Steps to test:
- CI run should be enough (I also tested that if I set the annotation visibility to Internal in the csv, the test does fail)

### Issues:
- fixes #5191 

------
- [x] Ready for review


daniel-wer (approved):
Thanks a lot @fm3! :+1:

Issue #5199, opened 2021-02-22
Task scripts are not properly handled when hot-swapping tasks


philippotto:
## Context
Finishing a task and starting a new one via "Finish and get next task" doesn't work correctly when scripts are involved. For example, a script which shows a message upon task start, won't show that message when the second task was started.

## Expected Behavior
Ideally, each task script should be terminated when a task is done, and the new task script should be started.

## Current Behavior
Currently, the page reloads if the task script changes. If the script does not change, the script simply stays alive for the new task. If it does the correct thing, is implementation specific.

## Steps to Reproduce the bug
1. Create task with script which outputs something upon start
2. Switch from one task to another with "finish and get next task"
3. see that the output only is created once.
4.

## Your Environment for bug

* Browser name and version: e.g. Chrome 39
* Operating System and version: e.g. Windows 10
* Version of WebKnossos (Release or Commit):
irrelevant

PR #5200, opened 2021-02-22
Reload page when the next task is loaded which has a different script

philippotto:
- https://___.webknossos.xyz

### Steps to test:
- Create a task with a script which outputs something upon start.
- Switch from one task to another with "finish and get next task"
- The page should reload and the output should be visible after switching the task.

### Issues:
- contributes to https://github.com/scalableminds/webknossos/issues/5199

------
- [X] Ready for review


daniel-wer (approved):
:+1:

PR #5201, opened 2021-02-22
Floor the z coordinate on initial movement-shortcut-firing

philippotto:
- https://floorzwhenfing.webknossos.xyz

### Steps to test:
- Open a dataset and enforce a z position which is very close to it's higher neighbor (e.g., `9.9`) via
```
webknossos.apiReady().then(async api => {
	api.tracing.centerPositionAnimated([28928, 28928, 1282.9], false)
})
```
- press (and keep pressing f) and notice how the z value changes in a consistent manner (on current prod, the first z changes are faster in comparison)

### Issues:
- fixes https://discuss.webknossos.org/t/move-value-not-consistent/1645/2

------
- [x] Ready for review


daniel-wer:
Nice find! I have not tested this specific branch yet as the CI didn't go through, but I think the current solution doesn't entirely get rid of the problem. It works for f-ing, but the problem persists (and is made worse) for subsequent d-ing, because the position is always right at the edge to the previous slice. This could be somewhat mitigated by setting the position to .5 instead of flooring. However, the second update would still be twice as fast as the subsequent updates.
To make sure updates are always evenly spaced one would need to account for the current moving direction and set the position to .0 or .999 accordingly.

Also this shouldn't be special-cased to z but instead to the 3rd dimension, because the problem exists for the other two viewports as well, right?

Let's have a call if it's easier :)

philippotto:
Unfortunately, you are correct :laughing: 

> To make sure updates are always evenly spaced one would need to account for the current moving direction and set the position to .0 or .999 accordingly.

Yes, this seems like a good solution :+1: I'll try to integrate this.

> Also this shouldn't be special-cased to z but instead to the 3rd dimension, because the problem exists for the other two viewports as well, right?

As far as I can see, this special behavior (first event changes coordinate, subsequent events increment in small steps) is only performed for "depth" related movement (`moveZ` has a suboptimal name as the `z` refers to the dimension which is orthogonal to the current viewport). Thus, shortcuts as `f`, `d`, `space` etc should be covered. Moving via arrow keys shouldn't be affected as these always use the normal move speed.

frontend/javascripts/oxalis/model/reducers/flycam_reducer.js
```diff
@@ -105,13 +105,24 @@ function resetMatrix(matrix: Matrix4x4, dataSetScale: Vector3) {
...
+    if (vector[2] > 0) {
+      // If the direction is incrementing, we clamp to .0 so that subsequent movements
+      // go from .0 to 0.999 before the next slice is shown.
+      matrix[14] = Math.floor(matrix[14]);
+    } else {
+      // If the direction is decrementing, we clamp to .999 so that subsequent movements
+      // go from .999 to .0 before the next slice is shown.
+      matrix[14] = Math.floor(matrix[14]) + 0.999;
+    }
+  }
```
daniel-wer, line 125: > As far as I can see, this special behavior (first event changes coordinate, subsequent events increment in small steps) is only performed for "depth" related movement (moveZ has a suboptimal name as the z refers to the dimension which is orthogonal to the current viewport). Thus, shortcuts as f, d, space etc should be covered. Moving via arrow keys shouldn't be affected as these always use the normal move speed.

Yes agreed, that's what I meant as well :) However, wouldn't that mean that you would need to look up which component of `vector` is the "depth" one, here (depending on the `activePlaneId`)? Currently you're always accessing the true z component by using `vector[2]` and `matrix[14]`, but it should vary depending on which is the active viewport. See for example the `MOVE_FLYCAM_ORTHO` reducer which also does something like `const dim = Dimensions.getIndices(planeId)[2];` for the dynamic space direction.

My suggestion would be to add the active `planeId` to the function signature and then something like
```suggestion
  if (clampToEdge) {
    const dim = Dimensions.getIndices(planeId)[2];
    if (vector[dim] > 0) {
      // If the direction is incrementing, we clamp to .0 so that subsequent movements
      // go from .0 to 0.999 before the next slice is shown.
      matrix[12 + dim] = Math.floor(matrix[12 + dim]);
    } else {
      // If the direction is decrementing, we clamp to .999 so that subsequent movements
      // go from .999 to .0 before the next slice is shown.
      matrix[12 + dim] = Math.floor(matrix[12 + dim]) + 0.999;
    }
  }
```

philippotto: Ahh, you are obviously completely right. Thanks for not getting sidetracked by my stubborn assumption that my implementation was working for all dimensions :see_no_evil: Also thank you for coding the fix!

philippotto:
I also covered https://discuss.webknossos.org/t/comments-search-dialog-hides-current-node-comment/1644. looks like this now:

image

"top" didn't work, because there is not enough space and antd will resort to another position automatically.

daniel-wer (approved):
Awesome, works like a charm :+1:

PR #5207, opened 2021-02-24
add proxy uris to sitemap.xml

fm3:
Adding the uris from the proxy config to sitemap.

I also cleaned up the code a bit.

- [x] Ready for review


youri-k (approved):
LGTM :shipit:

PR #5214, opened 2021-02-25
Remove K/L from intro shortcut list

normanrz:
We don't really want people to use K and L anymore, because it is more complicated than resizing with the mouse.

- [x] Ready for review


philippotto (approved):
:+1:

PR #5215, opened 2021-02-25
Fix Shift + Scroll for some mouses on macOS

### LINKED ISSUE #4854: Changing brush size via mouse wheel doesn't work properly on Mac OS X / Chrome
> The shortcut to adjust the brush size (Shift + Scrolling) does not work correctly for me. No matter what direction I scroll in the brush always shrinks. Perhaps a Chrome + OSX thing.

also note that this bug also occurs for mice which are not magic mice

daniel-wer:
For mouses that only have a vertical scroll wheel, MacOS seems to emulate horizontal scrolling when using `Shift` and vertical scrolling. See for example, this issue: https://bugs.mojang.com/browse/MC-121772?page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel&showAll=true
I fixed it by picking the absolutely larger value of `deltaX` or `deltaY`. As we're not making a difference between horizontal/vertical scrolling afaik, this shouldn't negatively affect other parts of the code.

~~During testing, I noticed that neither `Shift + Scroll` nor `Alt + Scroll` seems to work in Firefox. Not sure whether that is a known bug.~~ This problem was caused by the `imwheel` program, not related to webKnossos.

- https://fixmacscroll.webknossos.xyz

### Steps to test:
- Own a Mac
  - Open a volume tracing and activate the "Brush".
  - Use `Shift + Scroll` to change the size of the brush. This should work in both directions (increasing/decreasing size).
- Test that other mouse bindings using the scroll wheel still work:
  - `Scroll` to move in z-direction
  - `Alt + Scroll` to zoom

### Issues:
- fixes #4854 

------
- [x] Ready for review


daniel-wer:
I'd like to do some distributed testing for this (https://fixmacscroll.webknossos.xyz):

@hotzenklotz Could you test whether this works for you now on Mac, in Chrome and in Firefox?
@philippotto @fm3 Could you test whether `Shift` or `Alt` + `Scroll` work for you in Firefox on Linux (see testing instructions)? For me, the `shiftKey`, `ctrlKey`, and `altKey` properties on `WheelEvents` seem to always be false which is why wK only scrolls through the dataset instead of changing the brush size or zooming :thinking:

hotzenklotz:
Works great under Mac OSX with Chrome, Firefox and Safari. Brush resizes as intended :-) Thanks

fm3:
Works fine for me in Firefox Windows, Chrome Windows, Firefox Linux and Chromium Linux

philippotto:
Everything works great on chrome and firefox for me (linux) :tada:

frontend/javascripts/libs/input.js
```diff
@@ -526,10 +526,8 @@ export class InputMouse {
     if (isDragging && this.ignoreScrollingWhileDragging) {
       return;
     }
-    let delta = 0;
-    if (event.deltaY != null) {
-      delta = -Number(event.deltaY);
-    }
+    // For some mouses on MacOS, `shift + wheel` will scroll horizontally instead of vertically
+    const delta = -(Math.abs(event.deltaX) > Math.abs(event.deltaY) ? event.deltaX : event.deltaY);
```
philippotto, line 530: Great workaround :+1:

philippotto (approved):

daniel-wer:
Thanks for testing everyone, I found out that the `imwheel` program I'm using to work around a chrome scrolling-related bug caused the issue for me :disappointed: Adapting its configuration fixed the issue for me as well.

PR #5216, opened 2021-02-26
in datasets table, make _publication a foreign key

fm3:
This constraint should have been in place already. _publication can only be the id of a publication or null.

- [x] Ready for review


youri-k (approved):
LGTM :shipit:

PR #5217, opened 2021-02-26
Add is_superuser property to analytics events

### LINKED ISSUE #5211: Add superuser property to analytics events
Backend analytics user properties should include the information if the respective (multi)user is superuser

fm3:
Superusers (i.e. employees) are noise for analytics, so it may be desirable to filter out their events.

### Steps to test:
- in application.conf set `backendAnalytics.verboseLoggingEnabled = true`, open a dataset
- in analytics logging, the is_superuser property should show (and be true iff the requesting user is a superuser)

### Issues:
- fixes #5211

------
- [x] Ready for review

### DIFF
--- a/app/models/analytics/AnalyticsService.scala
+++ b/app/models/analytics/AnalyticsService.scala
@@ -47,10 +47,9 @@ class AnalyticsLookUpService @Inject()(userDAO: UserDAO, multiUserDAO: MultiUser
     extends LazyLogging {
   implicit val ctx: DBAccessContext = GlobalAccessContext
 
-  def isSuperUser(userId: ObjectId): Fox[Boolean] =
+  def isSuperUser(multiUserId: ObjectId): Fox[Boolean] =
     for {
-      user <- userDAO.findOne(userId)
-      multiUser <- multiUserDAO.findOne(user._multiUser)
+      multiUser <- multiUserDAO.findOne(multiUserId)
     } yield multiUser.isSuperUser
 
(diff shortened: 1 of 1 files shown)


app/models/analytics/AnalyticsService.scala
```diff
@@ -43,10 +43,19 @@ class AnalyticsService @Inject()(rpc: RPC, wkConf: WkConf, analyticsLookUpServic
...
 }
 
-class AnalyticsLookUpService @Inject()(userDAO: UserDAO, wkConf: WkConf) extends LazyLogging {
+class AnalyticsLookUpService @Inject()(userDAO: UserDAO, multiUserDAO: MultiUserDAO, wkConf: WkConf)
+    extends LazyLogging {
+  implicit val ctx: DBAccessContext = GlobalAccessContext
+
+  def isSuperUser(userId: ObjectId): Fox[Boolean] =
+    for {
+      user <- userDAO.findOne(userId)
```
youri-k, line 52: Do we really need to take the detour with the `userId`?
I think we can directly pass the `multiUserId` and save one db call.

youri-k (changes requested):
See my comment about having to look up the user again. I feel there is optimization potential.
Otherwise LGTM :slightly_smiling_face:

app/models/analytics/AnalyticsService.scala
fm3: Absolutely right! Good catch :) Could you have a look again?

youri-k (approved):
LGTM :shipit:

PR #5222, opened 2021-02-27
Adapt Agglomerate Mapping Max Mag to correct value

fm3:
https://github.com/scalableminds/webknossos/pull/4706 increased the threshold from 8 to 16, so message should also appear only starting at 16.

Long-term it would be nice to have a single place for this value, but that is not trivial since the datastore config is not exposed to the frontend. We could instead add a flag to the data requests when this limit was exceeded or something. But for now I think this is ok.

- [x] Ready for review

### DIFF
--- a/frontend/javascripts/oxalis/model/sagas/annotation_saga.js
+++ b/frontend/javascripts/oxalis/model/sagas/annotation_saga.js
@@ -15,6 +15,8 @@ import constants from "oxalis/constants";
 import messages from "messages";
 import { getRequestLogZoomStep } from "oxalis/model/accessors/flycam_accessor";
 
+/* Note that this must stay in sync with the back-end constant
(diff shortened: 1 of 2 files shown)


philippotto:
> Long-term it would be nice to have a single place for this value, but that is not trivial since the datastore config is not exposed to the frontend.

Maybe it would already help a bit to add a comment to each of the constant definitions mentioning the other spot on which this depends? The other threshold also looks like it should be extracted to a constant (at least at the module level): https://github.com/scalableminds/webknossos/pull/4706/files#diff-279361122384dfd2a93e9f0ddaa2a847254e0d925ce220d63762f221c07d261cR58

fm3:
Yes, god point :) I added comments, could you have another look?

philippotto (approved):
Awesome! Thanks for fixing this :rocket:

PR #5224, opened 2021-03-01
adds ctrl+scroll for zooming

### LINKED ISSUE #5213: Zoom with track pad
## Detailed Description
Add pinch-to-zoom with a trackpad. Apple computers support this; linux and windows maybe. I think support is based on the pointer events API.

normanrz:
Adds Ctrl+scroll for zooming in and out of ortho planes. This also enables pinch-to-zoom on mac trackpads.

Notes:
- Flight mode doesn't have a scroll-based zoom
- Should we also use ctrl+move for moving (as analogue to alt+move)?

### Steps to test:
- Open a dataset
- Use ctrl+scroll to zoom in
- Use pinch-to-zoom on a Mac trackpad

### Issues:
- fixes #5213 

------
- [x] Ready for review


philippotto (approved):
Works well :+1: Maybe it's a worth a changelog entry?

normanrz:
> Maybe it's a worth a changelog entry?

Will do

What do you think about Ctrl+Move?

philippotto:
> What do you think about Ctrl+Move?

I don't have a strong opinion about this, but I'm rather leaning against it. In the past, I noticed that it's easy to move accidentally while zooming via alt+scroll. If we separate the modifiers (alt for moving and ctrl for zooming), we could mitigate this. Also, I'm not super convinced by the alt+mouse-move shortcut. Personally, I'd find something like space+drag more intuitive. Changing from alt to ctrl doesn't really improve it imo. Changing it later once again might be too much churn.

PR #5229, opened 2021-03-03
Make project time Long

fm3:
Some projects are long, especially if expressed in milliseconds.

### Steps to test:
- Create project, annotate for at least 25 person-days. (Or just set a time (>=2^32) to a task in the db)
- Loading project list should not raise error

- [x] Ready for review


philippotto (approved):
LGTM :+1: 

> Create project, annotate for at least 25 person-days. (Or just set a time (>=2^32) to a task in the db)

:laughing:

PR #5235, opened 2021-03-03
Add API methods to get/set the active group and to set the camera position

daniel-wer:
These fairly basic api functions have been missing which I noticed while writing a project-related user script.
I also fixed a bug in the tree list, where we have to work around a bug in the `react-sortable-tree` library which messes with the highlighting of the `active{Tree,Group}`. The condition when this workaround should be applied was incomplete.

- https://apiextension.webknossos.xyz

### Steps to test:
- Can be tested from the dev console:
  - `webknossos.apiReady(3).then(api => api.tracing.setActiveGroup(2))`
  - `webknossos.apiReady(3).then(api => console.log(api.tracing.getActiveGroupId()))`
  - `webknossos.apiReady(3).then(api => api.tracing.setCameraPosition([100, 100, 200]))`

------
- [x] Ready for review

### DIFF
--- a/frontend/javascripts/oxalis/api/api_latest.js
+++ b/frontend/javascripts/oxalis/api/api_latest.js
@@ -40,6 +40,7 @@ import {
   getNodeAndTreeOrNull,
   getActiveNode,
   getActiveTree,
+  getActiveGroup,
   getTree,
   getFlatTreeGroups,
   getTreeGroupsMap,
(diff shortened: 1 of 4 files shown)


frontend/javascripts/oxalis/api/api_latest.js
```diff
@@ -754,6 +778,16 @@ class TracingApi {
     return getPosition(Store.getState().flycam);
   }
 
+  /**
+   * Sets the current camera position.
```
philippotto, line 782: You could also mention `centerPositionAnimated` here which would be the animated version of `setCameraPosition`.

philippotto (approved):
LGTM! :+1: Looking forward to using the set-camera-position-api!

PR #5237, opened 2021-03-04
Expose analytics endpoint to frontend

### LINKED ISSUE #5166: Expose analytics endpoint for frontend
## Detailed Description
Add an endpoint for tracking analytics events to the backend API. The controller should enrich the event information (e.g. with user, organization etc) and pass it on to the analytics service.

## Context
We also want to track events from the frontend, but not expose the implementation details or API key to the frontend.

fm3:
This PR
- Adds a new route `POST api/analytics/:eventType`, forwarding the json object posted there as `event_properties` to the backend analytics (user is automatically added from the cookie, analytics is only sent if logged in).
- Uses this new endpoint to track isosurface requests
- Removes tracking isosurface requests via backend communication
- Removes the old analytics postgres table and route that was already unused since 2017

### Steps to test:
- In application.conf set `backendAnalytics.verboseLoggingEnabled = true`
- Request isosurface
- Backend logging should show analytics event
- The request should not block the front-end. If it fails, no error should be shown
- Note that browser tracking blockers do block requests to this route (as is fair)

### Issues:
- fixes #5166 

------
- [x] Updated migration guide
- [x] Ready for review


frontend/javascripts/admin/admin_rest_api.js
```diff
@@ -125,6 +125,13 @@ export function doWithToken<T>(fn: (token: string) => Promise<T>, tries: number
   });
 }
 
+export function trackAnalytics(eventType: string, eventProperties: Object): void {
```
philippotto, line 128: `sendAnalyticsEvent` would be a tad clearer in my opinion

frontend/javascripts/admin/admin_rest_api.js
```diff
@@ -125,6 +125,13 @@ export function doWithToken<T>(fn: (token: string) => Promise<T>, tries: number
...
+  Request.sendJSONReceiveJSON(`/api/analytics/${eventType}`, {
```
philippotto, line 129:
```suggestion
  // Note that the Promise from sendJSONReceiveJSON is not awaited or returned here, since
  // failing analytics events should not have an impact on the application logic.
  Request.sendJSONReceiveJSON(`/api/analytics/${eventType}`, {
```

philippotto (commented):
Front-end lgtm! :+1: 

> Note that browser tracking blockers do block requests to this route (as is fair)

Because some might block all requests to `*/analytics/*`? Have you experienced this on your machine? I would have thought that most adblockers wouldn't be that aggressive.

conf/evolutions/067-drop-analytics.sql
```diff
@@ -0,0 +1,10 @@
+-- https://github.com/scalableminds/webknossos/pull/TODO
```
youri-k, line 1:
```suggestion
-- https://github.com/scalableminds/webknossos/pull/5237
```

youri-k (approved):
Backend LGTM :tada: 
Tested it and it works as expected

fm3:
> Because some might block all requests to `*/analytics/*`? Have you experienced this on your machine? I would have thought that most adblockers wouldn't be that aggressive.

Yes, I assume so. I experienced this with uBlock Origin.

PR #5244, opened 2021-03-08
Fix Bounding Box Checks for Tiff Export

fm3:
Follow-up Bugfix for #5195 

 - Correctly parse bounding box for size assertions (assume width,height,depth instead of xmax,ymax,zmax)
 - Use Long multiplication for volume to avoid int overflow
 - Remove some unused util methods in BoundingBox class
 - Add some typing to public methods in BoundingBox class

### Steps to test:
- with jobsEnabled, go to routes 
- `GET http://localhost:9000/api/jobs/run/tiffExport/sample_organization/testDataset/color?bbox=50,50,50,5000,5000,5000` (should say volume too large)
- `GET http://localhost:9000/api/jobs/run/tiffExport/sample_organization/testDataset/color?bbox=50,50,50,10,10,10` (check should go through, attempt contacting jobs broker. this was broken before, as 10,10,10 was assumed to be bottom-right, which was invalid)
- (I did test these locally)
- CI should run through

- [x] Ready for review

### DIFF
--- a/util/src/main/scala/com/scalableminds/util/geometry/BoundingBox.scala
+++ b/util/src/main/scala/com/scalableminds/util/geometry/BoundingBox.scala
@@ -4,7 +4,7 @@ import net.liftweb.common.{Box, Empty, Full}
 
 case class BoundingBox(topLeft: Point3D, width: Int, height: Int, depth: Int) {
 
-  val bottomRight = topLeft.move(width, height, depth)
+  val bottomRight: Point3D = topLeft.move(width, height, depth)
 
   def intersects(other: BoundingBox): Boolean =
     math.max(topLeft.x, other.topLeft.x) < math.min(bottomRight.x, other.bottomRight.x) &&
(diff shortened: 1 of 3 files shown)


util/src/main/scala/com/scalableminds/util/geometry/BoundingBox.scala
```diff
@@ -47,31 +47,27 @@ object BoundingBox {
...
-        b.topLeft.x,
-        b.topLeft.y,
-        b.topLeft.z,
-        b.topLeft.x + b.width,
-        b.topLeft.y + b.height,
-        b.topLeft.z + b.depth
-      ))
-
-  def fromForm(s: String): Box[BoundingBox] =
+  def createFrom(s: String): Box[BoundingBox] =
```
jstriebel, line 55: Is `createFrom` an established term for constructor-like methods? Couldn't it overload the constructor instead? (Probably not, because of the `Box`?). I found `fromForm` or maybe `fromFormString` quite speaking (also, since it would be similar in python-land), but `createFrom` is fine as-well.

jstriebel (approved):
LGTM :+1:, just one minor nitpick, feel free to leave as-is.

util/src/main/scala/com/scalableminds/util/geometry/BoundingBox.scala
fm3: Yes, overwriting the constructor is not possible as you described. fromX would have been fine by me, but there were already a number of createFrom methods here with different parameters and I didn’t want to make the diff all across wK large by renaming them all now.

Issue #5202, opened 2021-02-23
Add expiry + clean up job for exported files


fm3:
Exported files are stored on the server (for now a zipfile each) – those should not stay there indefinitely, we should introduce expiry and a clean up job.

The expiry should also be communicated to the user.

Also, intermediate files in `.converting` should be cleaned up periodically

The question is where to store the expiry timestamps (they need to survive backend restarts). Redis again? Or directly on disk in filenames/additional files?

We could also use modified timestamps from the filesystem.

frcroth:
- .converting and .export are populated by worker.
- New cronjob (like scan inbox) that deletes stuff
- Use modified date (older than 1 month, should be configurable)
- logging

frcroth:
@normanrz suggested using this instead: https://github.com/scalableminds/kube?tab=readme-ov-file#humerus-cleanup

Issue #4893, opened 2020-10-23
bucket requests: replace zoomStep by actual (vec3) mag 


fm3:
this zoomStep lookup has caused a lot of headaches, as many of the original assumptions don’t hold anymore, can we get rid of it entirely?

fm3:
we discussed that we want to send `mag` directly, compare #6154 – if in GET routes it should be mag=2-2-1 (parse string literal on server), in POST routes it can be sent as [2,2,1]. Should always have three components (no scalar short form)

Issue #4762, opened 2020-08-10
Unify naming of "zoomStep" and "resolution" / "resolutionIndex"


MichaelBuessemeyer:
Currently, the frontend uses "zoomStep" and "resolutionIndex" for the same thing and the resolutions in the store are all saved in as "resolutions".

I suggest sticking to one term for the "resolution"/"resolutionIndex". I think "resolutionIndex" is the better term as it works well with the term "resolutions". But on the other hand "zoomStep" is used much more frequent.

fm3:
related: #4893

also note this discussion https://github.com/scalableminds/webknossos/pull/4755#issuecomment-715367049

fm3:
superseded by #4893

PR #5246, opened 2021-03-08
Export Volume Annotations

### LINKED ISSUE #5203: Tiff Export for Volume Annotations
https://github.com/scalableminds/webknossos/pull/5195 creates the possibility to export data as tiff stacks.
It would be helpful for annotation pipelines if this could include volume annotations created by the webKnossos user.

The long-running job would have to get the volume data from the tracingstore (or fossildb directly?), unpack it, and mix it with the fallback segmentation layer data.

Should probably include the possibility to select the version of the volume tracing.

fm3:
image

Note that #5252 should be merged first (some of the bugs fixed there could crop up here too during testing)

### TODO (includes changes in https://github.com/scalableminds/webknossos-worker/pull/6/ )
 - [x] fetch volume annotations
 - [x] merge with fallback layer
 - [x] pass result to cuber
 - [x] are compound tracings cached long genough? → increased to 70 min, created follow-up issue #5255 
 - [x] use glob instead of exists
 - [x] pass version
 - [x] pass token only for this job
 - [x] pass tracing id from frontend
 - [x] don’t include fallback layer if it is unlinked
 - [x] test resulting files with imagemagick (`convert 000002.tiff -evaluate multiply 200 -depth 8 8.tif`)
 - [x] adapt job description in list view
 - [x] disable buttons already in frontend if mag1 is not present
 - [x] ensure saved state
 - [x] clean up intermediate files? (let’s leave them in for the moment, for debugging. clean-up job issue #5202)

### Steps to test:
- set up https://github.com/scalableminds/webknossos-worker/pull/6/
- set `jobsEnabled = true` in application.conf
- open volume annotation (also try hybrid, skeleton, view mode)
- brush some, create bbox
- click export tif button, modal should show annotation layer
- zip download from jobs list should include segmentation tifs with the annotations
- modals should not contain typos

### Issues:
- fixes #5203 

------
- [x] Ready for review


fm3:
Another idea: might be worth to save the “referrer” link with the job (e.g. to go back to the source (compound) annotation the tiffs were exported from. The tracing id is not sufficient for this lookup in compound case.

frontend/javascripts/oxalis/view/settings/export_bounding_box_modal.js
```diff
@@ -2,54 +2,82 @@
...
   };
 
-  const layerNames = dataset.dataSource.dataLayers.map(layer => {
-    const nameIfFromDataset = layer.category === "color" || !hasVolumeTracing ? layer.name : null;
-    const nameIfVolume =
-      hasVolumeTracing && layer.category === "segmentation" && layer.fallbackLayerInfo != null
-        ? layer.fallbackLayerInfo.name
-        : null;
-    return nameIfFromDataset || nameIfVolume;
+  const hasMag1 = (layer: APIDataLayer) => layer.resolutions.map(r => Math.max(...r)).includes(1);
```
philippotto, line 42: you can use the ResolutionInfo abstraction like this:

```suggestion
  const hasMag1 = (layer: APIDataLayer) => getResolutionInfo(layer.resolutions).hasIndex(0);
```

`getResolutionInfo` needs to be imported from `oxalis/model/accessors/dataset_accessor`

frontend/javascripts/oxalis/view/settings/export_bounding_box_modal.js
```diff
@@ -2,54 +2,82 @@
...
-          onClick={() => handleStartExport(layerName)}
-          disabled={startedExports.includes(layerName)}
+          key={exportKey(layerInfos)}
+          onClick={() => handleStartExport(layerInfos)}
+          disabled={startedExports.includes(exportKey(layerInfos)) || !layerInfos.hasMag1}
         >
-          {layerName}
-          {startedExports.includes(layerName) ? " (started)" : null}
+          {layerInfos.displayName}
+          {!layerInfos.hasMag1 ? " (resolution 1 missing)" : ""}
```
philippotto, line 79: can you create a follow-up issue to support exports where mag1 is missing? is it difficult to support this, too?

frontend/javascripts/oxalis/view/settings/export_bounding_box_modal.js
```diff
@@ -2,54 +2,82 @@
...
+      return {
+        displayName: layer.name,
+        layerName: layer.name,
+        tracingId: null,
+        tracingVersion: null,
+        hasMag1: hasMag1(layer),
+      };
+    if (layer.fallbackLayerInfo != null)
+      return {
+        displayName: "Volume annotation with fallback segmentation",
```
philippotto, line 55: nice :+1:

philippotto (commented):
front-end looks great! will test now.

frontend/javascripts/oxalis/view/settings/export_bounding_box_modal.js
fm3: https://github.com/scalableminds/webknossos/issues/5260
Shouldn’t be too difficult, but didn’t strike me as highest-priority at this time

frontend/javascripts/oxalis/view/settings/export_bounding_box_modal.js
fm3: Done :) Though it reminds me that we should really tackle #4762 and #4893 (probably we should do this together as it is so interlinked in front-end and back-end)

youri-k (approved):
I have not tested it, but the backend code looks good :tada:

PR #5252, opened 2021-03-08
Small Fixes for Tiff Export

### LINKED ISSUE #5248: Tiff export: permission denied without error message
Tiff export should work for all datasets that you can see.
Also, if there is a permission denied issue, there should at least be a readable error message.

### LINKED ISSUE #5249: Tiff export: show bbox in list
The bounding box of export jobs should be included on the Jobs page

### LINKED ISSUE #5251: Tiff export: segmentation layer not listed if in view mode
Turns out you can add user bboxes in view mode. In this case, the tiff export modal does not correctly list the segmentation layer.

Same for skeleton-only annotations

fm3:
### Steps to test:
- set `jobsEnabled = true` in application.conf
- bounding box of export jobs should now be visible in jobs list view
- when attempting to export from a dataset of a different orga, a readable error message should be shown
- when exporting from view mode or skeleton-only annotation, segmentation layer should be available in export modal (volume without fallback is the only case where it should be hidden)

### Issues:
- fixes #5248 
- fixes #5249 
- fixes #5251

------
- [x] Ready for review

### DIFF
--- a/frontend/javascripts/admin/job/job_list_view.js
+++ b/frontend/javascripts/admin/job/job_list_view.js
@@ -92,7 +92,7 @@ class JobListView extends React.PureComponent<Props, State> {
           <Link to={`/datasets/${job.organizationName}/${job.datasetName}/view`}>
             {job.datasetName}
           </Link>{" "}
-          (Box {job.boundingBox})
+          (Bounding Box {job.boundingBox})
         </span>
       );
     } else {
--- a/frontend/javascripts/admin/job/job_list_view.js
+++ b/frontend/javascripts/admin/job/job_list_view.js
@@ -91,7 +91,8 @@ class JobListView extends React.PureComponent<Props, State> {
           Tiff export from {job.layerName || "a"} layer of{" "}
           <Link to={`/datasets/${job.organizationName}/${job.datasetName}/view`}>
             {job.datasetName}
-          </Link>
+          </Link>{" "}
+          (Box {job.boundingBox})
         </span>
       );
     } else {
--- a/frontend/javascripts/oxalis/view/settings/export_bounding_box_modal.js
+++ b/frontend/javascripts/oxalis/view/settings/export_bounding_box_modal.js
@@ -10,10 +10,11 @@ import * as Utils from "libs/utils";
 type Props = {
   destroy: () => void,
   dataset: APIDataset,
+  hasVolumeTracing: boolean,
   boundingBox: BoundingBoxType,
 };
 
-const ExportBoundingBoxModal = ({ destroy, dataset, boundingBox }: Props) => {
+const ExportBoundingBoxModal = ({ destroy, dataset, hasVolumeTracing, boundingBox }: Props) => {
   const [startedExports, setStartedExports] = useState([]);
 
   const handleClose = () => {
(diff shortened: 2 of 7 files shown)


frontend/javascripts/oxalis/view/settings/export_bounding_box_modal.js
```diff
@@ -31,12 +32,15 @@ const ExportBoundingBoxModal = ({ destroy, dataset, boundingBox }: Props) => {
...
 
   const layerNames = dataset.dataSource.dataLayers.map(layer => {
-    const nameIfColor = layer.category === "color" ? layer.name : null;
+    const nameIfFromDataset = layer.category === "color" || !hasVolumeTracing ? layer.name : null;
     const nameIfVolume =
-      layer.category === "segmentation" && layer.fallbackLayerInfo && layer.fallbackLayerInfo.name
+      hasVolumeTracing &&
+      layer.category === "segmentation" &&
+      layer.fallbackLayerInfo &&
+      layer.fallbackLayerInfo.name
```
philippotto, line 40:
```suggestion
      layer.fallbackLayerInfo != null
```

should be enough, I think, since `name` should always exist given that the object is not null.

frontend/javascripts/oxalis/view/settings/export_bounding_box_modal.js
```diff
@@ -31,12 +32,15 @@ const ExportBoundingBoxModal = ({ destroy, dataset, boundingBox }: Props) => {
...
         ? layer.fallbackLayerInfo.name
         : null;
-    return nameIfColor || nameIfVolume;
+    return nameIfFromDataset || nameIfVolume;
```
philippotto, line 43: I find this code a bit hard to read. Using early-returns could avoid the duplicate check of the `category` for example:

```js
if (layer.category === "color" || !hasVolumeTracing) {
	return layer.name;
}

if (layer.fallbackLayerInfo != null) {
	return layer.fallbackLayerInfo.name;
}

// For a volume layer without fallback information, we cannot provide an export currently
return null;
```

However, it might be a matter of taste. Code usually reads better if `reader == writer` :laughing:

frontend/javascripts/admin/job/job_list_view.js
```diff
@@ -91,7 +91,8 @@ class JobListView extends React.PureComponent<Props, State> {
           Tiff export from {job.layerName || "a"} layer of{" "}
           <Link to={`/datasets/${job.organizationName}/${job.datasetName}/view`}>
             {job.datasetName}
-          </Link>
+          </Link>{" "}
+          (Box {job.boundingBox})
```
philippotto, line 95:
```suggestion
          (Bounding Box {job.boundingBox})
```

would be more consistent with the rest of the UI

philippotto (approved):
Front-end looks good to me :+1: I only made "optional" suggestions.

frontend/javascripts/oxalis/view/settings/export_bounding_box_modal.js
fm3: I’ll include this feedback in the next version, which wil be a bit more complex :see_no_evil: , and which is already on the way in https://github.com/scalableminds/webknossos/pull/5246

PR #5257, opened 2021-03-09
Allow shift+clicking isosurfaces even when a skeleton is present

### LINKED ISSUE #5220: Jumping to isosurface position via shift+click in 3D viewport is broken
## Context
After creating an isosurface (e.g., in a hybrid annotation), shift+click in the 3D viewport doesn't work.

## Expected Behavior
The clicked position should be jumped to.

## Current Behavior
Nothing happens.

## Steps to Reproduce the bug
1. Open dataset with segmentation
2. Load an isosurface via the "Meshes" tab
3. shift+click the isosurface in the 3D viewport
4.

## Your Environment for bug

* Browser name and version: e.g. Chrome 39
* Operating System and version: e.g. Windows 10
* Version of WebKnossos (Release or Commit):
chrome / linux

philippotto:
- https://fixisosurfacejumping.webknossos.xyz

### Steps to test:
- open a dataset with a segmentation in view mode, generate an isosurface and use shift+click in 3D viewport to go to the clicked position
- repeat the above step for a hybrid tracing

### Issues:
- fixes #5220

------
- [X] Ready for review


frontend/javascripts/oxalis/controller/td_controller.js
```diff
@@ -161,42 +168,40 @@ class TDController extends React.PureComponent<Props> {
         this.setTargetAndFixPosition();
       },
       pinch: delta => this.zoomTDView(delta, true),
+      mouseMove: (delta: Point2, position: Point2) => {
+        Store.dispatch(setMousePositionAction([position.x, position.y]));
+      },
+      leftClick: (pos: Point2, plane: OrthoView, event: MouseEvent, isTouch: boolean) => {
+        if (skeletonControls != null) {
+          skeletonControls.leftClick(pos, plane, event, isTouch);
+        }
```
philippotto, line 177: In the rare case, that the current mouse position points on a isosurfaces and a node (and the correct modifiers are pressed), it can happen that both actions are executed. However, this shouldn't do any harm in my opinion.

frontend/javascripts/oxalis/controller/td_controller.js
```diff
@@ -161,42 +168,40 @@ class TDController extends React.PureComponent<Props> {
         this.setTargetAndFixPosition();
       },
       pinch: delta => this.zoomTDView(delta, true),
+      mouseMove: (delta: Point2, position: Point2) => {
+        Store.dispatch(setMousePositionAction([position.x, position.y]));
```
philippotto, line 172: Previously, this was only done when no skeleton controller existed. I think, the rationale behind this was that we wouldn't need the mouse position then (because the mapping view wouldn't exist). However, updating the position in all cases seems less brittle to me.

daniel-wer: Do you mean this was only done when a volume controller existed? Because the existence of the mapping view is not related to the existence of a skeleton. If your description is correct, I don't understand how this could've worked before (or I'm misunderstanding :see_no_evil:)

frontend/javascripts/oxalis/controller/td_controller.js
daniel-wer: Yes it's probably harmless. If this becomes a frequent problem and annoys users we can think about only executing one of the two, depending on which element is nearer to the camera.

daniel-wer (approved):
Nice fix :+1:

frontend/javascripts/oxalis/controller/td_controller.js
philippotto: I had another look in the code and my example with the mapping view was a bit wrong. This particular mouse move handler is only responsible for updating the mouse position when moving the mouse in the 3d viewport. And that position is only used for one thing as far as I can see: for isosurface hit tests (i.e., for highlighting and for jumping to isosurfaces). That explains that only the isosurface-related functionality was broken in certain scenarios.

However, in both of the above cases the hit-test is triggered by an event (which contains the mouse position). Therefore, I removed the dispatch of the updateMousePosition-action and instead pass the mouse position directly into the hit test. This also removes the (throttled) call of `performIsosurfaceHitTest` which was done in every render call of the plane view.

Could you have a another look at the latest commit? :)

daniel-wer: Ah, I see, thank you for having another look! The changes make sense to me :+1:

Issue #4772, opened 2020-08-18
Allow to edit dataset from viewer


philippotto:
It would be very convenient to be able to edit the currently opened dataset without going to the dashboard. One easy way to achieve this, would be an edit button in the "dataset info" tab. That edit button could open the "edit dataset" view in a modal. when that modal is saved, the entire dataset could simply reload.

PR #5262, opened 2021-03-11
Link to View mode from Annotation mode info tab

### LINKED ISSUE #5259: Add link to dataset to annotation info tab
The info tab already contains the dataset name. It should be a link to the dataset (view mode)

fm3:
- https://linktoviewmode.webknossos.xyz

### Steps to test:
- open annotation of a dataset
- in info tab, dataset name should be link
- open private dataset’s annotation with sharing token while logged out
- view link should *not* include the token, clicking should yield error, external user is supposed to just see it via the annotation

### Issues:
- fixes #5259
related to #4772 (may be combined with settings link there)

------
- [x] Ready for review


daniel-wer (approved):
Nice, that's helpful! :+1: 

> view link should not include the token, clicking should yield error, external user is supposed to just see it via the annotation

Works as described, but is there any particular reasoning behind that? Including the token would make it work and avoid the error, that's why I'm asking. Still, no strong opinion on that.

fm3:
> Works as described, but is there any particular reasoning behind that? Including the token would make it work and avoid the error, that's why I'm asking. Still, no strong opinion on that.

My opinion on this is not really strong either, but my assumption was that the token-sharing should only work for exactly the url that was shared. If/when a user requests this to change, we can probably change it, but for now this makes the most sense to me.

daniel-wer:
But at least for now according to my tests, the token can be copied and used to access the dataset in view mode. I just wanted to make sure that you are aware of that, there's no harm in it imo.

PR #5263, opened 2021-03-11
breakWord for layer name above histogram

### LINKED ISSUE #5094: View mode layer settings: layer-name is cropped
For very long layer-names the name is cropped in the layer-settings:
Screenshot from 2021-01-26 14-06-39
(In this case the important distinction between the two layers is in the end of the name and not visible)
It was not possible to resize the settings. To see the full name, there should be either the option to resize the settings, or the name should wrap to the next line.

fm3:
image

### Steps to test:
- rename the layer of a dataset to a very long name but no hyphens
- refresh datasource-properties.json
- open in webKnossos viewer, open settings pane
- name should be fully readable
- name should not overlap with other UI elemnts (like the histogram data)

### Issues:
- fixes #5094 

------
- [x] Ready for review

### DIFF
--- a/frontend/javascripts/oxalis/view/settings/dataset_settings_view.js
+++ b/frontend/javascripts/oxalis/view/settings/dataset_settings_view.js
@@ -347,7 +347,7 @@ class DatasetSettings extends React.PureComponent<DatasetSettingsProps, State> {
       <Row>
         <Col span={24}>
           {this.getEnableDisableLayerSwitch(isDisabled, onChange)}
-          <span style={{ fontWeight: 700 }}>
+          <span style={{ fontWeight: 700, wordWrap: "break-word" }}>
             {!isColorLayer && isVolumeTracing ? "Volume Annotation" : layerName}
           </span>
 


frontend/javascripts/oxalis/view/settings/dataset_settings_view.js
```diff
@@ -347,7 +347,7 @@ class DatasetSettings extends React.PureComponent<DatasetSettingsProps, State> {
       <Row>
         <Col span={24}>
           {this.getEnableDisableLayerSwitch(isDisabled, onChange)}
-          <span style={{ fontWeight: 700 }}>
+          <span style={{ fontWeight: 700, wordWrap: "break-word" }}>
```
jstriebel, line 350: I think this should be a semicolon:
```suggestion
          <span style={{ fontWeight: 700; wordWrap: "break-word" }}>
```

MichaelBuessemeyer: The comma is correct. `{ fontWeight: 700; wordWrap: "break-word" }` is a JS object where properties are separated by a comma :nerd_face:

jstriebel: Ah, nevermind, I guess this is a react-component and the inner braces are just an object, so comma it is.

jstriebel (approved):
LGTM :+1:

frontend/javascripts/oxalis/view/settings/dataset_settings_view.js
fm3: yep, hence also wordWrap instead of word-wrap, which would have been the pure css syntax :) thanks you two for having a look :heart:

PR #5264, opened 2021-03-11
Send Analytics Event for Failed Dataset Upload

### LINKED ISSUE #5168: Add analytics event for failed requests
The analytics events introduced in #5156 do not encode if the request was completed.
Additional events could be tracked in case of e.g. failed dataset upload or failed dataset conversion.

fm3:
Track failed dataset uploads.
Sorry for the large diff in formatting, i did not change the `.then` block in dataset_upload_view, only added the catch before.

### Steps to test:
- insert `          _ <- Fox.failure("nope")` in `app/com/scalableminds/webknossos/datastore/controllers/DataSourceController.scala` Line 130
- in `application.conf`, set `backendAnalytics.verboseLoggingEnabled = true`
- Upload a dataset
- request should fail
- backend logging should show analytics event for failed request

### Issues:
- fixes #5168 

------
- [x] Ready for review


frontend/javascripts/admin/dataset/dataset_upload_view.js
```diff
@@ -135,46 +140,53 @@ class DatasetUploadView extends React.PureComponent<PropsWithFormAndRouter, Stat
...
-                isRetrying: false,
-                uploadProgress: 0,
+          finishDatasetUpload(formValues.datastore, uploadInfo)
+            .catch(error => {
+              sendFailedRequestAnalyticsEvent("finish_dataset_upload", error, {
+                dataset_name: datasetId.name,
               });
-            },
-          );
+              throw error;
```
philippotto, line 148: In line 182, there is already a catch handler for this promise (the second parameter to `then` is the onRejected handler). I'd move the send-request there and the `throw` can be removed then.

philippotto (approved):
Nice! See my one comment about moving the catch-handler, though :)

PR #5266, opened 2021-03-11
Unify Backend Enumerations

### LINKED ISSUE #5186: Unify Scala Enum Implementations
There are a few different ways of implementing enums across our backend. Can we always extend `Enumeration` and use our EnumUtils for json formatting?

fm3:
- https://unifybackendenums.webknossos.xyz

### Steps to test:
- Enum fields should still be intact: list task types, create, edit allowed modes
- Should be able to annotate, save, download, try hybrid, skeleton, volume
- CI should run through

### Issues:
- fixes #5186 

------
- [x] Ready for review


app/models/annotation/AnnotationSettings.scala
```diff
@@ -1,12 +1,19 @@
...
 import play.api.libs.json._
 
+object TracingMode extends ExtendedEnumeration {
+  type TracingMode = Value
+  val orthogonal, oblique, flight, volume = Value
+}
+
 case class AnnotationSettings(
-    allowedModes: List[String] = SKELETON_MODES,
+    allowedModes: List[TracingMode.Value] = skeletonModes,
```
fm3, line 16: Note for reviewer: in the files where the enums are defined, their respective type aliases cannot be imported (that causes conflicts), hence this is still `TracingMode.Value` here, but would be `TracingMode` elsewhere.

youri-k (approved):
LGTM :shipit:

PR #5267, opened 2021-03-11
Fix frontend search for objects with null values

daniel-wer:
There was a regression in the frontend search for keys whose value was an object which contained `null` or `undefined` values - leading to a full site crash :see_no_evil: 

- https://___.webknossos.xyz

### Steps to test:
- Open the tasks view and select a project with tasks.
- Enter a random search term in the search field at the very top which doesn't find any tasks. -> The site should not crash. (it does on master)

------
- [x] Ready for review


fm3 (approved):
Works as intended and LGTM :) Thanks for adding more tests as well!
Has this bug been around for a while? In this case, should we maybe add a changelog entry for its fix?

daniel-wer:
I stumbled upon the bug by accident and checked airbrake which only shows a single instance from a couple of days ago. So it looks like it is rather fresh, although from the code I would've expected it to be much older. I'll add a changelog entry anyways :)

fm3:
Ok cool :) To be honest I hadn’t even been aware there was a search bar up there 🙈 Maybe few people have used it.

PR #5277, opened 2021-03-15
More Line Breaks for Readability

### LINKED ISSUE #5276: Max Width for Task Type description
In Task Type table, description columns get extremely wide if there is a lot of text in the markdown description. There should be line-breaks forced by a max width

fm3:
- Adds max width to task type description container. This field contains user-supplied markdown. If the user writes a long paragraph, the task type table column got extremely wide, forcing the user to scroll really far to the right to see the other columns. This also influences the task type description in Dashboard/My Tasks, where the text now also does not fill the whole width of the container anymore. IMO this is a good thing (it’s hard to read very wide paragraphs).
- Add break-word to dataset name in annotation info tab, as dataset names tend to get very long

- https://maxwidth.webknossos.xyz

### Steps to test:
- create task type with a very long paragraph in description, should be readable both in task type table and in dashboard/my tasks
- rename a dataset to a long name, should still be readable without scroll bar in annotation info tab

### Issues:
- fixes #5276 

------
- [x] Ready for review


frontend/stylesheets/_dashboard.less
```diff
@@ -248,6 +248,8 @@
 }
 
 .task-type-description {
+  max-width: 700px;
+
```
daniel-wer, line 252:
```suggestion
```

daniel-wer (approved):
Nice improvements :+1:

PR #5278, opened 2021-03-15
Trim experience domains

### LINKED ISSUE #5271: Trim whitespace from experience domains
Let’s avoid confusing duplicates by always trimming whitespace from experience domains, both in task and user menus.
compare https://discuss.webknossos.org/t/whitespace-and-experience-domains/1652

fm3:
The back-end should remove trailing whitespaces from experience domains during task creation. It’s a regression that was created by not using the fromForm method anymore, but instead use auto-generated JSON reads for the task parameters.
When assigning experiences to users, the trimming already happens as intended.
In this PR I add this for task creation again, and also remove the unused methods from the Experience case class

### Steps to test:
- Create task, use surrounding whitespace in experience domain name
- give user same experience with *different* amount of surrounding whitespace
- they should still get the task (the whitespace should be removed)

### Issues:
- fixes #5271 

------
- [x] Ready for review


jstriebel (approved):
LGTM :+1:

PR #5284, opened 2021-03-16
Slack Notification for critical FossilDB errors, failed Jobs

### LINKED ISSUE #5282: Telemetry message for failed fossildb writes
Failed fossildb writes should be logged to slack (unless they are caused by restart and rocks handle not being initialized)

fm3:
Note that the additional PR  Track failed jobs to slack + analytics (#5294)  is now included in this one.

Adds Slack notifications for failed fossildb `put` requests (as long as they are not just the `UNAVAILABLE` error caused by the fossildb restarting).

Also gives some love to the slack notification service in general:
 - extracted SlackClient class used by both the wK and the tracingStore service
 - Unified notification title schema and colors
 - Pass output from SQL Schema check to slack message
 - Add rate limiting (do not send more than 30 messages per minute)

### Steps to test:
- In application.conf, set a valid slack url or just `verboseLoggingEnabled = true`
- provoke some critical fossildb errors (this is not trivial with a standard fossildb, either change its code or mount its data dir on a tiny partition so you write it full. Or possibly change the wK code so that it sends a broken put request. But compiler may catch that)
- backend logging or slack should show pretty messages
- there should *not* be a notification for fossildb-not-running problems
- slow volume requests should still be logged to slack
- schema mismatches should still be logged, and now include the schema diff
- If you provoke more than 30 messages per minute, those above the limit should be dropped. Note that the rate-limit works once for the wK module and once for the tracingstore module.

### Issues:
- fixes #5282 

------
- [x] Ready for review


youri-k (approved):
LGTM :shipit:

PR #5285, opened 2021-03-16
Add layer + mag dir to wkw uploads if missing

### LINKED ISSUE #5219: Dataset upload: add default layer name and mag if not present
When uploading a dataset via the upload view, the required structure (after stripping common prefix, excluding layer names) is layerName/mag/z0/y0/x0.wkw

If there is only one layer and/or only one mag, the user may skip those top directories. In this case, the dataset import should not fail. Instead, the upload should add reasonable defaults (`color` and `1` I guess)

fm3:
When a user uploads wkw-converted data directly, that is not in a layer directory or not even in a mag directory, the back-end now automatically adds those to the path.

A risk to this method is that layers called `1`, `1-1-1` or `z0` etc would be mistaken for mags/bucket paths and this would trigger wrongly. But I’ll assume this will rarely be the case.

- https://uploadwkwdir.webknossos.xyz

### Steps to test:
- upload only parts of the dataset
- should be moved (console logging), and viewable after import

### Issues:
- fixes #5219 

------
- [x] Ready for review


youri-k (approved):
LGTM :shipit: 

While testing, I stumbled across this typo: https://github.com/scalableminds/webknossos/blob/upload-wkw-dir/frontend/javascripts/admin/dataset/dataset_add_view.js#L99
Would you mind including a fix for this?

PR #5292, opened 2021-03-17
Fix oblique bug II

### LINKED ISSUE #5288: Missing buckets in oblique mode
In some cases, the oblique mode shows black spots.

philippotto:
I adapted two things:
- the amount of scan lines is determined dynamically instead of using a constant (yay, one constant less :tada:) <-- this fixed one error class
- Instead of only casting the scan lines before and after the current z (this is done to ensure buckets are immediately available when moving across bucket boundaries), this is also done at the current z <-- this solved a second error class

- https://___.webknossos.xyz

### Steps to test:
- I set up an affected dataset locally and compared the before VS after scenarios to check that the bug was fixed

### Issues:
- fixes #5288 

------
- [X] Ready for review


daniel-wer (approved):
Nice, the fixes make sense to me. If you tested this, LGTM :+1:

PR #5294, opened 2021-03-18
Track failed jobs to slack + analytics

### LINKED ISSUE #5286: Telemetry messages for failed worker jobs
wk-worker is in its early phase. We want to know about all failed jobs in production.

fm3:
Log failed worker jobs to slack + backend analytics

### Steps to test:
- in application.conf set `features.jobsEnabled = true`, `backendAnalytics.verboseLoggingEnabled = true`, `slackNotifications.verboseLoggingEnabled = true`
- set up worker, but modify it so that jobs fail (e.g. raise an exception in a task)
- call that task from wk
- open jobs list, see failed job
- backend logging should show what messages would have been sent to analytics + slack
- while jobs list stays open, celery infos are repeatedly fetched, but messages should only fire for *newly* failed jobs.

### Issues:
- fixes #5286 

------
- [x] Ready for review

### DIFF
--- a/app/controllers/JobsController.scala
+++ b/app/controllers/JobsController.scala
@@ -4,18 +4,20 @@ import java.nio.file.{Files, Paths}
 import java.util.Date
 
 import com.mohiva.play.silhouette.api.Silhouette
+import com.scalableminds.util.accesscontext.GlobalAccessContext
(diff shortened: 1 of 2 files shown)


youri-k (approved):
LGTM :shipit:

PR #5297, opened 2021-03-18
Ignore escaped carriage return in diff schema

### LINKED ISSUE #5293: Diff schema has false positives
Despite the recent changes, sometimes diff_schema shows a diff where there is none. Could be about trailing carriage returns again, even though the diff command already contains `--strip-trailing-cr`. We should investigate. Compare https://scm.slack.com/archives/CMBMU5684/p1616005193000700

fm3:
Turns out psql escapes carriage returns in function definitions.
These should be ignored when checking schema diffs.

This creates a potential for false negatives (if the real difference is the existence of `\r' in the schema), but as discussed this seems rare enough to take the risk.

### Steps to test:
- Insert '\r' somewhere in the schema (while preserving sql validity)
- run diff_schema.js
- should not show differences induced by `\r`occurences, should still show other differences

### Issues:
- fixes #5293 

------
- [x] Ready for review

### DIFF
--- a/tools/postgres/diff_schema.js
+++ b/tools/postgres/diff_schema.js
@@ -101,7 +101,6 @@ function generateRandomName() {
   return "wk_tmp_" + random;
 }
 
-
 function sortAndClean(dumpedDir) {
   glob.sync(dumpedDir + "/**", { nodir: true }).forEach(function(fileName) {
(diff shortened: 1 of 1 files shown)


jstriebel (approved):
LGTM :+1:

PR #5304, opened 2021-03-18
Add requested url to rejection reason for better error reporting

philippotto:
@fm3 and I noticed that airbrake shows frequent failed bucket requests with a `404` status. It's unclear where this status is coming from. Unfortunately, the failing request reports don't show any url which is why I added that url.

- https://___.webknossos.xyz

### Steps to test:
- I tested the following things:
  - break the actual bucket fetching url (failing url will be printed)
  - break the token request url (failing url will be printed)
  - add a "throw error" into the buffer extraction code (error will be printed)

------
- [X] Ready for review

### DIFF
--- a/frontend/javascripts/libs/request.js
+++ b/frontend/javascripts/libs/request.js
@@ -300,13 +300,15 @@ class Request {
               // to a datastore. Then, ping the datastore
               pingMentionedDataStores(text);
 
-              return Promise.reject(json);
+              /* eslint-disable-next-line prefer-promise-reject-errors */
+              return Promise.reject({ ...json, url: requestedUrl });
             } catch (jsonError) {
               if (showErrorToast) Toast.error(text);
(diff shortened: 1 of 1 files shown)


frontend/javascripts/libs/request.js
```diff
@@ -300,13 +300,15 @@ class Request {
               // to a datastore. Then, ping the datastore
               pingMentionedDataStores(text);
 
-              return Promise.reject(json);
+              /* eslint-disable-next-line prefer-promise-reject-errors */
```
philippotto, line 303: I'm not sure why we don't use a proper error here (line 307 also uses this pattern). If eslint encourages us to do this, we should probably comply. I opened #5305 as a follow-up.

daniel-wer: Hmm, I don't know either, thanks for creating the follow-up issue.

daniel-wer (approved):
:+1:

Issue #5298, opened 2021-03-18
Secure Dev Deployments


fm3:
We may want to use private data in dev deployments.
To allow this we need to
 - [x] change scmboy password (kube)
 - [x] change scmboy token (introduce config var, set in kube) → PR https://github.com/scalableminds/webknossos/pull/5309
 - [x] pass those to nightly run (use circleci environment variables. Turns out only token is used. https://github.com/scalableminds/webknossos/pull/5321)
 - [x] change application.authentication.ssoKey → separate issue #5317 with PR #5318 
 - [x] change datastore.key (not needed, was already secret)
 - [x] change tracingstore.key

An alternative would be to surround the dev deployments by basicauth, but that would also have to be passed to nightly and does not allow the flexibility of creating additional users on individual dev deployments to show stuff to third parties

PR #5309, opened 2021-03-19
Expose default user initial token to config

fm3:
Expose default user token to config, to allow setting it to secret but known value for nightly tests

### Steps to test:
- change `application.authentication.defaultUser.token` in application.conf
- should show in UI at `/auth/token`

### Issues:
- contributes to #5298

------
- [x] Ready for review


jstriebel (approved):
:+1:

PR #5316, opened 2021-03-22
Fix scale json schema

### LINKED ISSUE #5129: Cannot set scale of [1,1,1]
When using 1,1,1, in the advanced tab of the datasource edit view, this error message is shown below the editor:

Invalid schema: instance.scale[0] must be strictly greater than true; instance.scale[1] must be strictly greater than true; instance.scale[2] must be strictly greater than true

The corresponding schema looks like this: `items: { type: "number", minimum: 0, exclusiveMinimum: true },` in `datasource.schema.js` which sounds about right. Needs investigation.

MichaelBuessemeyer:
I fixed the schema by using the [docs](https://json-schema.org/understanding-json-schema/reference/numeric.html#range). The docs say, that there is a difference between how `exclusiveMinimum` is used in version 4  and how it is used in later versions like 6 that we use. We simply used the outdated way of defining a minimum.

- https://___.webknossos.xyz

### Steps to test:
- Open the edit view of a dataset, activate the advanced view, and fiddle around with the scale. 
  - Try to use only two numbers for the scale. -> Validation should fail.
  - Try to use four numbers for the scale. -> Validation should fail.
  - Try to use 0 as any number in the scale array. -> Validation should fail.
  - Try to use 0.4 as any number in the scale array. -> Validation should succeed.
  - Try to enter `[1,1,1]` as a valid scale.  -> Validation should succeed.
  - and more :)

### Issues:
- fixes #5129

------
- [x] Ready for review

### DIFF
--- a/frontend/javascripts/types/schemas/datasource.schema.js
+++ b/frontend/javascripts/types/schemas/datasource.schema.js
@@ -139,7 +139,7 @@ export default {
         },
         scale: {
           type: "array",
-          items: { type: "number", minimum: 0, exclusiveMinimum: true },
+          items: { type: "number", exclusiveMinimum: 0 },
           minItems: 3,
           maxItems: 3,
         },
--- a/frontend/javascripts/types/schemas/datasource.types.js
+++ b/frontend/javascripts/types/schemas/datasource.types.js
@@ -62,6 +62,5 @@ export type DatasourceConfiguration = {
     team: string,
   },
   dataLayers: Array<DataLayer>,
-  // Add minimum=0 and exclusiveMinimum=true to items
   scale: Array<number>,
 };


fm3:
Shouldn’t 0.4 be allowed? In the steps to test it says should fail, but 0.4nm/vx looks like a valid value to me. Am I missing something?

MichaelBuessemeyer:
> Shouldn’t 0.4 be allowed? In the steps to test it says should fail, but 0.4nm/vx looks like a valid value to me. Am I missing something?

Nope, you are correct. This was simply a Copy and Paste mistake from my side :grin:

daniel-wer (approved):
Nice find! :+1:

Issue #5313, opened 2021-03-22
Cache-control should have max-age > 1h


philippotto:
## Detailed Description
CSS and JS files currently have `cache-control: public, max-age=3600`. However, we are already using a cache buster for these files, which means we could use a very high max-age and add the `immutable` attribute.

## Context
Properly caching our resources would reduce the initial page load amount by ~ 2 MB (one of these bigger requests takes 2 seconds on my machine).

philippotto:
@fm3 noted:

- on localhost there are no cache-control headers
- Play might differentiate between dev and production
- the handling of these cache attributes should probably happen in `*RequestHandler.scala`
- also note that on dev, we use the JS proxy which can also "interfere" with the HTTP headers

philippotto:
Fixed via deployment config by @fm3.

PR #5318, opened 2021-03-22
Make SSO opt-in in config

### LINKED ISSUE #5317: ssoKey should have empty default
ssoKey should be empty in application.conf by default (make sso opt-in)

fm3:
The code considers SSO is disabled when the config key is an empty string.

I also added the config part related to #5313 (but the null values are equal to how this behaved before. Only added those lines as hints how to set this in production).

### Steps to test:
- try SSO route, logging should say that it is disabled

### Issues:
- fixes #5317
- contributes to #5298
------
- [x] Ready for review


jstriebel (approved):
LGTM :+1:

PR #5321, opened 2021-03-22
Use auth token from environment variable during nightly

### LINKED ISSUE #5298: Secure Dev Deployments

fm3:
Our dev deployment is no longer using the public token. The nightly has to know the secret token, it is passed via environment variable.

### Steps to test:
- run nightly from this branch, should run green. (I did that already)

### Issues:
- fixes #5298 (in combination with recent kube changes)

------
- [x] Ready for review

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -282,10 +282,11 @@ workflows:
   version: 2
   circleci_build:
     jobs:
-      - build_test_deploy:
-          filters:
-            tags:
-              only: /.*/
+      - nightly
+      #- build_test_deploy:
+      #    filters:
+      #      tags:
+      #        only: /.*/
   circleci_nightly:
     jobs:
       - nightly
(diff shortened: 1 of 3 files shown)


philippotto (approved):
:tada:

Issue #5323, opened 2021-03-22
Antialias screenshots


daniel-wer:
## Detailed Description

Screenshots of the tracing views are created by rendering specific scenes off-screen and converting the resulting buffer to an image. However, WebGL 1 doesn't support antialiasing for this use case, leading to ragged edges in the resulting screenshots. See https://discourse.threejs.org/t/why-is-a-custom-fbo-not-antialiased/1329
The linked thread proposes upgrading to WebGL 2 or using a postprocessing pass (FXAA) to fix this issue.

![segem_test_mag_2__201_197_75__TDView](https://user-images.githubusercontent.com/1702075/112038093-bb375480-8b42-11eb-8329-d5592cbf4177.png)

PR #5324, opened 2021-03-22
Screenshot for 3D view

### LINKED ISSUE #5218: Screenshot for 3D view
## Detailed Description
`Q` should also make a screenshot of the 3D viewport

## Context
To make pretty pictures of meshes, it should be possible to make a screenshot of the 3D viewport

daniel-wer:
Also screenshot the 3D view when using the screenshot functionality.
Allow to screenshot using the menu button in view mode.

During testing I noticed that screenshots near the dataset border included a mirrored version of the actual screenshot. This bug was happening, because the screenshot needs to be flipped horizontally due to the inverted webgl y-axis. For pixels outside of the dataset, the alpha value is set to 0. Therefore, when flipping the image of a canvas in-place and drawing it to the same canvas, you could see-through the unflipped version of the image in some areas. I fixed that by drawing the flipped image to another canvas.

See the followup issue https://github.com/scalableminds/webknossos/issues/5323 regarding antialiasing of the screenshots.

- https://___.webknossos.xyz

### Steps to test:
- Open a dataset, move to the edge of the dataset and press "Q" to create screenshots of all four viewports.
- Alternatively, use the "Screenshot" menu entry from the dropdown next to the save button.
- Screenshot in flight mode.

### Issues:
- fixes #5218 

------
- [x] Ready for review


daniel-wer:
@MichaelBuessemeyer Could you have a look at this? :)

MichaelBuessemeyer:
I just tested this feature and it works very well :tada: 

While looking at the screenshots, I noticed that rendering the "out of dataset" parts with nothing looks a bit weird. Maybe we could render these parts just the same way they are rendered in the viewports (black)? But this is only a guess whether this looks better. What do you think @daniel-wer? What about a follow-up issue for this? 
image

MichaelBuessemeyer:
>During testing I noticed that screenshots near the dataset border included a mirrored version of the actual screenshot. This bug was happening, because the screenshot needs to be flipped horizontally due to the inverted webgl y-axis. 

@daniel-wer could you please provide me with a bit more details on how to reproduce this bug? I wasn't able to reproduce it in the master.

daniel-wer:
> While looking at the screenshots, I noticed that rendering the "out of dataset" parts with nothing looks a bit weird. Maybe we could render these parts just the same way they are rendered in the viewports (black)? But this is only a guess whether this looks better. What do you think @daniel-wer? What about a follow-up issue for this?

Interesting, this seems to be browser-dependent - Chrome rendered the "out of dataset" parts in a yellowish white which was less obtrusive. This is because the "out of dataset" parts get an alpha value of 0 in the shader (which is the cause of the mirroring bug I fixed). We could either change this alpha value in the shader code or initialize the buffer that is rendered into with the correct background color, I think. I'll think about which solution I prefer.

> @daniel-wer could you please provide me with a bit more details on how to reproduce this bug? I wasn't able to reproduce it in the master.

Doing a screenshot so that the upper half of the xy viewport is outside of the dataset bounding box should trigger this bug (I used chrome). The result looked like this:

ROI2017_wkw__249_148_512__PLANE_XY

frontend/javascripts/oxalis/view/action-bar/view_dataset_actions_view.js
```diff
@@ -18,11 +26,12 @@ export default function ViewDatasetActionsView(props: Props) {
   );
   const overlayMenu = (
     <Menu>
-      {props.layoutMenu}
       <Menu.Item key="share-button" onClick={() => setShareDatasetModalVisibility(true)}>
         <Icon type="share-alt" />
         Share
       </Menu.Item>
+      {screenshotMenuItem}
```
MichaelBuessemeyer, line 33: Nice addition :+1:

MichaelBuessemeyer (approved):
The code looks good and works in each view. I also could reproduce the bug on the master (using firefox and chrome) and can confirm that it is solved on this branch :rocket:

frontend/javascripts/oxalis/shaders/main_data_fragment.glsl.js
```diff
@@ -137,7 +137,7 @@ ${compileShader(
 void main() {
   vec3 worldCoordUVW = getWorldCoordUVW();
   if (isOutsideOfBoundingBox(worldCoordUVW)) {
-    gl_FragColor = vec4(0.0);
+    gl_FragColor = vec4(0.0, 0.0, 0.0, 1.0);
```
daniel-wer, line 140: @philippotto Setting the alpha value to 1 for pixels outside of the bounding box, fixes both the mirroring issue (described in the PR description - would make my earlier fix obsolete) as well as the other transparency issue (see Michaels screenshot). Do you see any harm in that? I checked that the node picking is not affected.

philippotto: Sounds perfectly fine to me :+1: I don't think that a lot of reasoning went into this line when it was written. The result probably behaved like "black" in the critical use case (viewport is out of bounds) which was considered good enough. Cool that you found this solution :)

MichaelBuessemeyer: I can't find any case where it would result in a :bug: too.

MichaelBuessemeyer:
I retested the changes on the dev deployment and they seem to work in all cases.

It should be ready for :ship:

PR #5325, opened 2021-03-23
Fix SQL errors during failed job tracking

fm3:
Follow-up for #5294 

SQL lookups with empty lists fail, let’s immediately return in this case.

Also, fix some formatting

- [x] Ready for review


jstriebel (approved):
LGTM :+1:

PR #5330, opened 2021-03-23
Analytics: mark annotation updates that are view-only

### LINKED ISSUE #5243: In save annotation analytics event, mark if movement only
Some annotation saves do not contain meaningful changes to the annotation (e.g. camera movements only). Those should be marked in analytics, so that the statistics show how much actual annotating is done.

We should check which other update actions are supposed to not be “meaningful” annotation work (how about changing the annotation name? Rename a tree? Toggle visibility? etc, make a list.

fm3:
Some tracing update actions are caused by the owner moving around or toggling tree visibility only. We want to be able to filter those in analytics.
Hence, the `update_annotation_view_only` analytics event is introduced and `update_annotation` is only sent if update actions that are considered “significant” are present.

I also added some type annotations in `TracingController.scala`, which led to some reformatting diff. The only real change in that file is in `commitUpdates`

### Steps to test:
- in `application.conf`, set `backendAnalytics.verboseLoggingEnabled = true`
- create annotation, move around, save, create trees or brush, save
- In backend logging, you should see the analytics events that would be sent.
- If both “significant” and “view only” update actions are present in a save, both events should be sent
- Note that for Hybrid tracings, two may be sent per save (for skeleton and for volume).

### Issues:
- fixes #5243

------
- [x] Ready for review

### DIFF
--- a/app/controllers/WKTracingStoreController.scala
+++ b/app/controllers/WKTracingStoreController.scala
@@ -2,8 +2,9 @@ package controllers
 
 import com.scalableminds.util.accesscontext.{DBAccessContext, GlobalAccessContext}
 import com.scalableminds.util.tools.{Fox, FoxImplicits}
+import com.scalableminds.webknossos.tracingstore.TracingUpdatesReport
 import javax.inject.Inject
-import models.analytics.{AnalyticsService, UpdateAnnotationEvent}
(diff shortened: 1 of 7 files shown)


jstriebel (approved):
LGTM :+1:

PR #5332, opened 2021-03-23
Switch to image.sc support forum by default

normanrz:
Switches the support links to https://forum.image.sc/tag/webknossos

------
- [x] Ready for review


frontend/javascripts/navbar.js
```diff
@@ -300,6 +300,18 @@ function LoggedInAvatar({ activeUser, handleLogout, ...other }) {
         <Menu.Item disabled key="organization">
           {orgDisplayName}
         </Menu.Item>
+        <Menu.Item key="switch">
```
fm3, line 303: This does not really seem to work for me. When I click it (in Firefox) the menu closes, so I can’t type in it to see what it does. Also, this should probably only appear in the isMultiMember case, right?
From the code, I don’t really understand what it is supposed to do.

normanrz: This was not supposed to be part of the PR. Just a failed experiment. Would be nice to have something like this, tho :D

philippotto (commented):
Looks good to me :+1: However, the autocomplete-switch-stuff is still in the PR.

philippotto (approved):
:+1:

PR #5333, opened 2021-03-24
In diff schema, replace escaped carriage return by two spaces

fm3:
Replacing by emptystring did not work because the functions file is structured as a table, where the cells are padded with spaces.
So let’s try again by replacing them with the padding that would have been there in their place.

- [x] Ready for review


jstriebel (approved):
:+1:

PR #5339, opened 2021-03-25
Fix bug where some values in project list were displayed incorrectly

### LINKED ISSUE #5336: Pausing project sets time view to NaN temporarily
In project list view, the returned project json is plugged into the list after the pause/resume API call. However, the json returned there does not include the status/time.
Either we need to patch this together with the previously present project object, or those API calls need to return the project *with status*

fm3:
- https://projectupdate.webknossos.xyz

### Steps to test:
- Project list, click on pause/resume
- tracing time and open assignments should not change (certainly not vanish or be set to NaN)

### Issues:
- fixes #5336

------
- [x] Ready for review


philippotto (approved):
:+1:

PR #5340, opened 2021-03-25
Don’t scroll up when showing confirm modals

### LINKED ISSUE #5053: Don't scroll up on "deactivate user"
When clicking "deactivate user" in the user list, I am scrolled to the top of the page which is very annoying when trying to deactivate multiple consecutive users. Maybe this can be fixed with a `preventDefault` call. Might also apply to other links in the list.

fm3:
- https://modaldontscrollup.webknossos.xyz

### Steps to test:
- Create a bunch of tasks
- in task list, scroll down, click »delete«, confirmation modal should show, but page should not scroll up to top
- same for task types, users, projects, scripts, teams
- same for own tasks in dashboard (cancel / reset actions etc)

### Issues:
- fixes #5053 

------
- [x] Ready for review

### DIFF
--- a/CHANGELOG.unreleased.md
+++ b/CHANGELOG.unreleased.md
@@ -11,14 +11,14 @@ For upgrade instructions, please check the [migration guide](MIGRATIONS.released
 [Commits](https://github.com/scalableminds/webknossos/compare/21.04.0...HEAD)
 
 ### Added
-- 
+-
 
 ### Changed
 - webKnossos is now part of the [image.sc support community](https://forum.image.sc/tag/webknossos). [#5332](https://github.com/scalableminds/webknossos/pull/5332)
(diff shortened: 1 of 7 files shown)


philippotto (changes requested):
Thank you for tackling this and finding a workaround! However, I think, it would be better to simply not use `<a href...>` and instead use a proper button with a link appearance. antd provides this with: `<Button type="link" onClick={...}>`. I think, this can act as a drop-in replacement and will also solve the scroll issue without having to deal with `preventDefault()` in all click handlers.
Can you give it a try to see whether this works?

philippotto:
The antd button might introduce unnecessary padding, but we could create a wrapper component which removes the padding.

fm3:
Replaced by new implementation (as per Philipp’s suggestion) in #5371

PR #5341, opened 2021-03-25
Don't re-fill all available recommended settings to task type form

### LINKED ISSUE #5270: Task Type form re-fills all available recommended settings
## Detailed Description
When a task type does not have a value for a particular recommended setting, the task type edit form still auto-fills all available settings to some defaults upon load. When the user hits save again, without removing them again first, they are now set.
Instead, the form should show only the settings that are already set (possibly with the option to add default settings)

## Context
Users reported that some values should *not* be set for some task types so that the tracer’s personal settings can be used for their tasks.

philippotto:
If there is already a non-null recommendation-configuration, don't add the defaults.

- https://___.webknossos.xyz

### Steps to test:
- edit the default task type and add a recommended configuration
- use the "remove" buttons to edit the recommended configuration
- save task type and re-open the edit page
- the recommended configuration should look exactly as before (i.e., no additional attributes should be added)
- disabling the recommendation, saving + reloading, re-enabling the recommendation should show all available properties again

### Issues:
- fixes #5270

------
- [X] Ready for review


philippotto:
@MichaelBuessemeyer ping :)

frontend/javascripts/admin/tasktype/task_type_create_view.js
```diff
@@ -91,6 +91,8 @@ class TaskTypeCreateView extends React.PureComponent<Props, State> {
   }
 
   async applyDefaults() {
+    const taskType = this.props.taskTypeId ? await getTaskType(this.props.taskTypeId) : null;
+    const hasRecommendedConfiguration = taskType.recommendedConfiguration != null;
```
MichaelBuessemeyer, line 95:
```suggestion
    const hasRecommendedConfiguration =
      taskType != null && taskType.recommendedConfiguration != null;
```

MichaelBuessemeyer (approved):
LGTM and works as described :+1: 

There is just one small thing flow is complaining about. I left a suggestion that is hopefully "pretty" :) 

And thanks for the detailed testing steps. They made pretty easy to check.

PR #5345, opened 2021-03-29
Load precomputed meshes

### LINKED ISSUE #4810: Efficient mesh storage
We want to have a workflow that can generate all meshes of a given automatic segmentation. We need to think of a clever format to store them. Some questions/observations:
- One STL file per segment on disk is infeasible 
- We may want to store the meshes in a chunked way
- We may want to store multiple Level-of-Details (LOD) for each mesh
- Database or file system?

We should do some research how other 3D applications (most importantly Neuroglancer) handle the storage. Links: 
- https://github.com/google/neuroglancer/tree/master/src/neuroglancer/datasource/precomputed
- https://github.com/seung-lab/cloud-volume

### LINKED ISSUE #5346: Meshfile UI follow-up
## Detailed Description
- [ ] In the meshes tab, get rid of the mesh section. Uploading meshes is not really required anymore
- [ ] Make two buttons for on-demand rendering of isosurfaces (current button) and loading of pre-rendered mesh (new button, only available if dataset/fallback layer has a meshfile, maybe as dropdown-button if multiple meshfiles are available)
- [ ] List pre-rendered meshes just as computed isosurfaces (with loading indicator, delete and download button)
- [ ] I would get rid of the term "isosurface" and only use "mesh" going forward.
- [ ] Add "load mesh" to context menu (if meshfile is available)

## Context
The mesh file integration from the hackathon is kind of hacky and needs work to be properly integrated.

fm3:
Load precomputed meshes of a segmentation from a hdf5 mesh file.

New datastore routes:

```
GET  /datasets/:organizationName/:dataSetName/layers/:dataLayerName/meshes (list available meshfiles)
POST /datasets/:organizationName/:dataSetName/layers/:dataLayerName/meshes/chunks (list chunks for segment)
POST /datasets/:organizationName/:dataSetName/layers/:dataLayerName/meshes/chunks/data (get chunk data)
```

### To Do Backend
 - [x] implement API
 - [x] error handling, malformed hdf leads to exceptions
 - [x] re-enable token protection for the new requests
 - [x] CI
 - [x] Include caching for read chunk requests
 - [x] in case of exceptions, call finishAccess on cached mesh file
 - [x] Handle compressed meshes (read metadata, may be `stl+gzip`, if so, set gzip header)

### To Do Frontend
 - [x] connect to front-end admin_rest_api
 - [x] merge master (the new code here is still based on old antd)
 - [x] rework UI as specified in #5346, quoting from normanrz:
    - [x] In the meshes tab, get rid of the mesh section. Uploading meshes is not really required anymore
    - [x] Make two buttons for on-demand rendering of isosurfaces (current button) and loading of pre-rendered mesh (new button, only available if dataset/fallback layer has a meshfile, maybe as dropdown-button if multiple meshfiles are available)
    - [x] List pre-rendered meshes just as computed isosurfaces (with loading indicator, delete and download button)
    - [x] I would get rid of the term "isosurface" and only use "mesh" going forward.
    - [x] Add "load mesh" to context menu (if meshfile is available)
    - [x] Add visibility checkboxes to precomputed meshed in the list

### Steps to test:
- Put mesh hdf5 file into `meshes` directory inside a segmentation layer (talk to fm3 or normanrz for access to one)
- abc

### Issues:
- fixes #4810 
- fixes #5346

------
- [x] Ready for review


fm3:
@youri-k If you have time, could you already have a look at the back-end code? The front-end part will probably change quite a bit, but the scala side is probably more or less stable.

.circleci/not-on-master.sh
```diff
@@ -1,8 +1,4 @@
 #!/usr/bin/env bash
 set -Eeuo pipefail
 
-if [ "${CIRCLE_BRANCH}" == "master" ]; then
-  echo "Skipping this step on master..."
-else
-  exec "$@"
-fi
+echo "Skipping this step on hackathon..."
```
youri-k, line 4: Re-enable the CI before merging

youri-k (commented):
Did not test, but the backend code looks good :tada: 
Will test, once the frontend is ready

philippotto:
@grittaweisheit What's the status here? Are the checkboxes in the PR description up to date? If not, please check what's already done and add missing items in case there are any. Would be really cool to see this PR ready for review/testing :rocket:

grittaweisheit:
Sorry, for not keeping everything up to date! 
I now checked the items that are done. Only the context menu integration is missing. 

Furthermore @normanrz mentioned that it would be cool to have checkboxes to toggle the meshes' visibility. If we don't want to just reload from the file but from stored buffers or something similar (analogue to how it is currently done with single meshes from files) it seems to be more to do than just a little bit. (I thought I was on the right path but realized it's not as easy as I thought :/  )

normanrz:
> Furthermore @normanrz mentioned that it would be cool to have checkboxes to toggle the meshes' visibility. If we don't want to just reload from the file but from stored buffers or something similar (analogue to how it is currently done with single meshes from files) it seems to be more to do than just a little bit. (I thought I was on the right path but realized it's not as easy as I thought :/ )

Maybe we could just hide them in the THREE scene?

philippotto:
> Furthermore @normanrz mentioned that it would be cool to have checkboxes to toggle the meshes' visibility. If we don't want to just reload from the file but from stored buffers or something similar (analogue to how it is currently done with single meshes from files) it seems to be more to do than just a little bit. (I thought I was on the right path but realized it's not as easy as I thought :/ )

I'd hope that this is fairly similar to the existing visibility changes for the old (uploadable) meshes (i.e., see `setMeshVisibility` in the scene controller) which merely acts on the THREE scene as @normanrz suggests. However, you probably saw that already.. Let's talk about it in person if there is some complexity I'm missing right now. Just hit me up when you've got time.

grittaweisheit:
I just got it to work! 
There's some cleaning up to do and there are no checkboxes (I used a button to see if it works), but the concept stands.

philippotto:
> I just got it to work!
> There's some cleaning up to do and there are no checkboxes (I used a button to see if it works), but the concept stands.

Awesome! The checkboxes feel cool :+1: 

In general, the UI already works quite well :rocket: However, during testing I noticed a few things which could be improved:
- [ ] The `meshes` GET request is not cached. For example, opening the context menu will always fetch for the available meshes. Also, moving the mouse sometimes triggers this request.
- [ ] A mesh is only listed after it was completely loaded. Ideally, it should behave similarly to the ad-hoc calculated isosurfaces (they show up immediately in the list and a spinner is shown if they are being loaded). It's probably helpful to throttle the internet connection via the browser devtools to test this.
- [ ] "Please select a mesh file first" seems to be unnecessary, if there is only one file anyway.
- [ ] If no mesh file is available, the "load from meshfile" button should be disabled
- [ ] I'd probably rename "load from meshfile" to "Load precomputed meshfile"

frontend/javascripts/admin/admin_rest_api.js
```diff
@@ -1553,3 +1567,65 @@ export function getAgglomerateSkeleton(
     ),
   );
 }
+
+export function getMeshfilesForDatasetLayer(
```
philippotto, line 1571: what's the difference to `getMeshfilesForDatasetLayer` in line 1239?

frontend/javascripts/oxalis/controller/scene_controller.js
```diff
@@ -259,6 +259,10 @@ class SceneController {
     this.stlMeshes[id].visible = visibility;
   }
 
+  setIsoVisibility(id: number, visibility: boolean): void {
```
philippotto, line 262: `setIsosurfaceVisibility` would be better imo

frontend/javascripts/oxalis/controller/combinations/segmentation_plane_controller.js
```diff
@@ -100,6 +100,6 @@ export function isosurfaceLeftClick(pos: Point2, plane: OrthoView, event: MouseE
   cellId = segmentation.cube.getMappedDataValue(position, getRequestLogZoomStep(Store.getState()));
 
   if (cellId > 0) {
-    Store.dispatch(changeActiveIsosurfaceCellAction(cellId, position));
+    Store.dispatch(changeActiveIsosurfaceCellAction(cellId, position, true));
```
philippotto, line 103: `true` is or `shouldReload`, right? Wouldn't this be unnecessary for precomputed meshes? Or is that case caught in the isosurface saga (I couldn't find that, though)?

frontend/javascripts/oxalis/view/right-menu/meshes_view.js
```diff
@@ -243,49 +209,146 @@ class MeshesView extends React.Component<
...
+        Load from Meshfile
+      </DropdownButton>
+    );
+
     const getMeshesHeader = () => (
-      <div style={{ marginTop: 10 }}>
+      <React.Fragment>
         Meshes{" "}
-        <Tooltip title="Meshes are rendered alongside the actual data in the 3D viewport. They are imported from STL files.">
+        <Tooltip title="Meshes are rendered alongside the actual data in the 3D viewport. They are imported from STL or hdf5 files. They can also be computed.">
```
philippotto, line 324:
```suggestion
        <Tooltip title="Meshes are rendered alongside the actual data in the 3D viewport. They can be computed ad-hoc or pre-computed.">
```

since we don't support dropping stls anymore, the file formats seem like an unnecessary detail.

frontend/javascripts/oxalis/view/node_context_menu.js
```diff
@@ -181,9 +201,62 @@ function NoNodeContextMenuOptions({
...
+        layerName,
+        fileName,
+        id,
+        chunkPos,
+      );
+      const geometry = parseStlBuffer(stlData);
+      getSceneController().addIsosurfaceFromGeometry(geometry, id);
+    }
+    createMesh(id, globalPosition);
+  };
```
philippotto, line 251: this code should be in the isosurface saga imo. i think, there is also some duplication with code in the mesh-view itself. both places should be unified and be moved into the saga (or another module).

philippotto (changes requested):
I already had a look at the code and left some comments :)

frontend/javascripts/oxalis/controller/combinations/segmentation_plane_controller.js
grittaweisheit: The function doesn't even seem to be used, so maybe I can just remove it?

philippotto: Yes, good point :+1:

philippotto:
@grittaweisheit Do you think we can get the PR merged this week? :)

grittaweisheit:
I think it's possible :). I just finished implementing the feedback and fixing a bug I noticed when moving to a mesh (it gets "refreshed", which means full new computation for precomputed meshes).

philippotto (commented):
awesome! will test now one last time :)

philippotto:
also a changelog entry is still missing.

frontend/javascripts/oxalis/view/right-menu/meshes_view.js
```diff
@@ -236,49 +218,125 @@ class MeshesView extends React.Component<
...
+            overlay={getMeshFilesDropdown()}
+            buttonsRender={([leftButton, rightButton]) => [
+              React.cloneElement(leftButton, {
+                disabled: !hasCurrenMeshFile,
+                style: { borderRightStyle: "dashed" },
+              }),
+              React.cloneElement(rightButton, { style: { borderLeftStyle: "dashed" } }),
+            ]}
+          >
+            Load precomputed meshfile
```
philippotto, line 303:
```suggestion
            Load Precomputed Mesh
```

frontend/javascripts/oxalis/view/node_context_menu.js
```diff
@@ -181,9 +193,47 @@ function NoNodeContextMenuOptions({
...
+
+  const getLoadMeshMenuItem = () => (
+    <Menu.Item
+      className="node-context-menu-item"
+      key="load-mesh-file"
+      onClick={loadMesh}
+      disabled={!currentMeshFile}
+    >
+      Load Mesh from File
```
philippotto, line 231:
```suggestion
      Load Precomputed Mesh
```

philippotto (approved):
Awesome! Works flawlessly for me :+1: I only have two wording suggestions for the UI which I added as further PR comments :)

Let's merge it on Tuesday (I can do that since you will not be at work, probably) :rocket:

PR #5362, opened 2021-04-06
Fix volume tasks from file without specified Bbox

### LINKED ISSUE #5361: Creating Task from volume zip: empty bbox not replaced by full one
In the task create view (not bulk), using a volume tasktype and „from NML“ option with an existing volume zip that contains an NML without a specified bounding box.

## Expected Behavior
Resulting volume tracing should get full bounding box of the dataset (this is the behavior for explorational volume uploads too)

## Current Behavior
The resulting volume tracing has an empty bounding box (0,0,0,0,0,0), essentially making it useless as no volume data is ever loaded.

fm3:
- https://volumetaskbbox.webknossos.xyz

### Steps to test:
- Create explorational volume, brush some, download
- Create volume task type
- Create task, select the volume task type, select “from NML”, select the downloaded zip
- As user, fetch the task, previously brushed data should be present

### Issues:
- fixes #5361

------
- [x] Ready for review


youri-k (approved):
Works perfectly :tada:

PR #5369, opened 2021-04-07
[WIP] Status bar

### LINKED ISSUE #4995: Add status bar to show important information
Similar to blender, wK could have a thin status bar at the bottom to show what the mouse buttons do (the labels would change when modifiers change the behavior). Additionally, important information such as the current magnification could be shown.

image

Other potential things to show:
- whether buckets are currently loading (could show the speed, too)
- ...?

grittaweisheit:
This PR adds a status bar to a tracing view to display important information and shortcuts.

- https://statusbar.webknossos.xyz

### Steps to test:
- open tracing
- there should be a status bar displaying the current mag 
- when moving the mouse in a (non 3d view) plane the position and cell-id should be displayed
- there should be shortcuts for moving, moving along 3rd axis and zooming
- in the 3d viewport an additional shortcut for rotating the 3d view should be displayed 

### Issues:
- fixes #4995

------
- [x] Ready for review


frontend/javascripts/statusbar.js
```diff
@@ -0,0 +1,177 @@
+// @flow
```
philippotto, line 1: I think this file should be moved to `webknossos/frontend/javascripts/oxalis/view`.

frontend/javascripts/statusbar.js
```diff
@@ -0,0 +1,177 @@
...
+
+  getCellInfo(globalMousePosition: ?Vector3) {
+    const segmentationLayerName = Model.getSegmentationLayer().name;
+    const cube = this.getSegmentationCube();
+    const renderedZoomStepForMousePosition = api.data.getRenderedZoomStepAtPosition(
+      segmentationLayerName,
+      globalMousePosition,
+    );
+    const getIdForPos = (pos, usableZoomStep) =>
+      pos && cube.getDataValue(pos, null, usableZoomStep);
```
philippotto, line 119: In the "Volume" tab (aka mapping-info-tab) one can select/enable a mapping which changes the segment ids according to a JSON (or hdf5) "dictionary". I'd argue that we should show the mapped segment id in that case here. This should be possible via something like this:

```
id = cube.getDataValue(...)
return cube.mapId(id)
```

there's no need to check whether the mapping is enabled, since `mapId` will do so automatically.

Additionally, I would change the appearance of the id so that it shows "Cell: 123 (mapped)" *iff* a mapping is enabled. This should probably happen outside of the `getIdForPos` function, since the result wouldn't be an id anymore (but a string). You can check whether a mapping is enabled via `cube.isMappingEnabled()`.

frontend/javascripts/statusbar.js
```diff
@@ -0,0 +1,177 @@
...
+        <span>
+          <img
+            src="/assets/images/icon-downsampling.svg"
+            style={{ width: 15, height: 15 }}
+            alt="Resolution"
+          />
+          {activeResolution.join("-")}{" "}
+        </span>
+        <span>Pos: [{globalMousePosition ? this.getPosString(globalMousePosition) : "-,-,-"}]</span>
+        {hasSegmentation() && this.getCellInfo(globalMousePosition)}
```
philippotto, line 150: I'd just change this to:
```suggestion
        {this.getCellInfo(globalMousePosition)}
```

and then `return null` in that function if `Model.getSegmentationLayer()` returns null.

philippotto (commented):
Cool stuff :100: Awesome to see it merged with the layout pr. The code looks pretty good :+1: I only left some small code-related comments.

Other than that I would on the following before we can merge this:
- [ ] the shortcut hints should adapt to the current tool. I might have said in an earlier talk, that this is not high-pri for now, but I think, this would need to change to get this merged, since the statusbar would contain wrong information otherwise (e.g., when being in brush-mode, the status bar tells the user that they can move with the left mouse button).
- [ ] the icons need better contrast/colors
- [ ] the alignment might need to be double checked

You can also go through the other internal feedback you got and follow through with that, but the above points would be my personal prioritization :)

grittaweisheit:
Thanks for your feedback! I implemented it as well as a lot of the others' feedback.

frontend/javascripts/oxalis/view/statusbar.js
```diff
@@ -0,0 +1,220 @@
...
+
+const statusbarStyle: Object = {
+  marginLeft: borderToggleButtonMargin,
+  marginRight: borderToggleButtonMargin,
+  fontSize,
+  display: "flex",
+  flexWrap: "wrap",
+  overflow: "hidden",
+};
+const defaultIconStyle = { height: fontSize - 1, marginTop: 2 };
```
philippotto, line 48: Why `fontSize - 1`? At least on my screen, this makes the icons quite blurry. For example, the "movement" dashes next to the left-click-drag-icon becomes barely visible. When I bumped the height via the devtools, the icon became clear again and the height also looked alright. Or do you see any downsides to this?

frontend/javascripts/oxalis/view/statusbar.js
```diff
@@ -0,0 +1,220 @@
...
+          <span style={defaultShortcutStyle}>
+            <img
+              className="keyboard-mouse-icon"
+              src="/assets/images/icon-statusbar-mouse-right.svg"
+              alt="Mouse Left"
+              style={defaultIconStyle}
+            />
+            Place Node
+          </span>
+        )}
```
philippotto, line 119: This part also needs to be adapted for the current tool (since one cannot place nodes in the brush tool). I'd recommend a simple object which maps from tool to right-click-action-description à la:
```
const rightClickToLabel = {
  MOVE: isSkeletonAnnotation ? "Place Node" : null,
  BRUSH: "Erase",
  TRACE: "Erase",
  FILL_CELL: null,
  PICK_CELL: null,
}
```

that way, we can easily add "open context menu"  as soon as we support that (instead of null).

philippotto (approved):
Awesome! This is really cool :) I only left two small to dos, but afterwards, this should be good to merge into @MichaelBuessemeyer's layout PR :+1:

philippotto:
:rocket: :100:

PR #5371, opened 2021-04-07
Don’t scroll up when showing confirm modals

### LINKED ISSUE #5053: Don't scroll up on "deactivate user"

fm3:
Second try, replacing  #5340

- https://linkbutton.webknossos.xyz

### Steps to test:
- Create a bunch of tasks
- in task list, scroll down, click »delete«, confirmation modal should show, but page should not scroll up to top
- same for task types, users, projects, scripts, teams
- same for own tasks in dashboard (cancel / reset actions etc)
- Padding for those links should look the same as before

### Issues:
- fixes #5053 

------
- [x] Ready for review

### DIFF
--- a/frontend/javascripts/components/link_button.js
+++ b/frontend/javascripts/components/link_button.js
@@ -1,0 +1,7 @@
+// @flow
+import { Button } from "antd";
+import * as React from "react";
+
+export default function LinkButton(props: Object) {
+  return <Button type="link" {...props} className="link-button" />;
+}
--- a/frontend/javascripts/components/link_button.js
+++ b/frontend/javascripts/components/link_button.js
@@ -2,6 +2,9 @@
 import { Button } from "antd";
 import * as React from "react";
 
+// This component should be used instead of <a href="#" onClick={} />.
+// One advantage of this component is that clicking the link button won't scroll
+// up (which would happen with <a ... /> if not handled otherwise)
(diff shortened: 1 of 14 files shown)


frontend/javascripts/components/link_button.js
```diff
@@ -0,0 +1,7 @@
+// @flow
+import { Button } from "antd";
+import * as React from "react";
+
+export default function LinkButton(props: Object) {
```
philippotto, line 5:
```suggestion
// This component should be used instead of <a href="#" onClick={} />.
// One advantage of this component is that clicking the link button won't scroll
// up (which would happen with <a ... /> if not handled otherwise)
export default function LinkButton(props: Object) {
```

philippotto (approved):
Awesome! I assume you searched for `<a href="#"` and replaced all occurrences?

fm3:
I hadn’t done the full list (only those where scrolling up could be a problem) but I did now. There is one remaining instance, that is in a Dropdown overlay. It did not look right with the LinkButton, but there is also no scrolling up happening there with the `a` so I’d leave it as is. The remainig `href="#"` are on AsyncLinks which also don’t have this problem.

Issue #5372, opened 2021-04-07
Remove "ignore script error" workaround


philippotto:
With antd4 and safari, there is a weird issue with the resize observer which throws a benign but annoying "script error". As a workaround, we ignore that error, but strictly speaking we don't know whether this is a resize-observer-related error. Therefore, we should find another workaround (or hope that a future antd update will fix this).

PR #5373, opened 2021-04-07
Ignore script error

philippotto:
- https://suppressscripterror.webknossos.xyz/

### Steps to test:
- open wk with safari and check that no "script error" toasts appear

### Issues:
- see https://github.com/scalableminds/webknossos/issues/5372

------
- [X] Ready for review

### DIFF
--- a/frontend/javascripts/libs/error_handling.js
+++ b/frontend/javascripts/libs/error_handling.js
@@ -132,13 +132,13 @@ class ErrorHandling {
         return;
       }
       if (error == null) {
-        // older browsers don't deliver the error parameter
+        // Older browsers (and apparently Safari) don't deliver the error parameter
         error = new Error(message);
       }
       console.error(error);
       this.notify(error);
 
-      if (error.toString() === "Script error.") {
+      if (error.toString() === "Error: Script error.") {
         // Safari and the newest antd version don't play well together. Often, "ResizeObserver loop completed with undelivered notifications." is triggered
         // but that message is lost. Instead, a "Script error." is thrown. Since that error is benign and can be frequent, we will
         // ignore it here to not annoy the user. The message wasn't added to BLACKLISTED_ERROR_MESSAGES so that the error can still be seen via airbrake.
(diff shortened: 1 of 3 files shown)


frontend/javascripts/libs/error_handling.js
```diff
@@ -132,12 +132,20 @@ class ErrorHandling {
...
       }
       if (error == null) {
-        // older browsers don't deliver the error parameter
+        // Older browsers (and apparently Safari) don't deliver the error parameter
         error = new Error(message);
       }
       console.error(error);
       this.notify(error);
 
+      if (error.toString() === "Error: Script error.") {
```
philippotto, line 141: @daniel-wer Maybe you have a better idea? I couldn't think of another solution, since Safari doesn't expose anything useful in the error message :thinking:

daniel-wer: Hmm, that's unfortunate :/ Not sure whether you've seen these:
https://github.com/ant-design/ant-design/issues/15075 is the exact issue, but was closed prematurely. However, a new issue was created: https://github.com/ant-design/ant-design/issues/26621 which mentions that this was fixed in some dependent library in the mean time. Doesn't look like it was really fixed, does it? :/

philippotto: Yes, I saw this, but noticed that the subdependencies are already at the newest version if I'm not mistaken (e.g., `rc-align` is at `4.0.9` according to npm and also according to `yarn.lock`). However, I'll link these issues in the follow-up issue in the hope that there will be a future fix in there :)

daniel-wer: I triggered an update of the antd dependencies by removing/adding it via yarn, but it doesn't look like any dependencies relevant to this error were updated :/ Let's go with this temporary workaround then :)

daniel-wer (approved):

hotzenklotz (approved):
Works for me

frontend/javascripts/libs/error_handling.js
philippotto: Cool, thank you! Having up to date sub dependencies probably will probably save us from other troubles sooner or later :)

PR #5374, opened 2021-04-07
Fix context menu overlay being to high

MichaelBuessemeyer:
In the current version of the master, the invisible overlay behind the node context menu is too high. It does not consider the navbar height.

- <https://fixcontextmenuoverlay.webknossos.xyz>

### Steps to test:
- Open a hybrid / skeleton tracing.
- Use `shift + right click` to open the context menu. No scroll bars should appear at the bottom and at the right side of the page.
- Do the same thing on the master. In this case, scroll bars should appear.

### Issues:
- No issue for this PR.
------
- [x] Ready for review


fm3 (approved):
Works for me 🎉  I can still provoke scroll bars when clicking too low on the screen (such that the context menu actually goes below the screen) – would be cool if the context-menu would open upwards instead then (like the browser/os-created one does)

MichaelBuessemeyer:
> Works for me tada I can still provoke scroll bars when clicking too low on the screen (such that the context menu actually goes below the screen) – would be cool if the context-menu would open upwards instead then (like the browser/os-created one does)

Just as discussed: I created a separate issue for this: #5383

PR #5378, opened 2021-04-08
Add annotation backlink to export job

### LINKED ISSUE #5261: Tiff export: store annotation backlink with job
The Tiff exports that include volume annotations should have a backlink to the annotation stored with the job. This way, in the job list view, the user can get back to the annotation.

The tracing id is not sufficient for this lookup in compound tracing case (where the tracing ids expire)

This requires a little refactoring (probably make the job run a post request, there are a little too many get parameters already)

fm3:
… so that the job list can link directly to the annotation the job was started from.

image

### Steps to test:
- set jobsEnabled = true, set up https://github.com/scalableminds/webknossos-worker/pull/14
- create volume/hybrid annotation, export a bbox
- job list should contain backlink
- old jobs should also still be rendered (without the backlink)

### Issues:
- fixes #5261 

------
- [x] Needs wk-worker update after deployment
- [x] Ready for review


philippotto (approved):
Very cool! LGTM! :100:

PR #5379, opened 2021-04-08
Revert "Make Project and TaskType names unique by Organization"

fm3:
Reverts scalableminds/webknossos#5334

Problem being that the new restriction of unique project names was too tight, deleted projects with duplicates existed in production. Thus, the db evolution did not succeed.

Issue #5337, opened 2021-03-24
Double-Check that localstorage persistence typechecks work


fm3:
In persistence.js, we check `PropTypes.checkPropTypes`. When changing the type as part of #5334 we expected this check to fail, but it didn’t. Is there something wrong there?

PR #5381, opened 2021-04-08
Make Project and TaskType names unique by Organization (second try)

### LINKED ISSUE #4005: Project/Task type names names should not have to be unique across all organizations
Don't know how feasible this is, but right now it's confusing, if you cannot create a project because some other (unknown to you) orga has used the same name.

fm3:
Project and TaskType names had to be globally unique. Now, they are uniqe per organization.

Reverts scalableminds/webknossos#5379 (which reverted the original #5334 )

Difference being that the unique constraint is no longer enforced in the database, but only via scala, compare issue description in #5388 I also made sure that the name is immutable (front-end already grayed out the form value, but backend didn’t enforce it) See last commits for the diff.

Quoted from original PR:

### TODO
 - [x] change unique constraints in schema
 - [x] add organization during insert
 - [x] change project api (by orga+name/by id/by disambiguation)?
 - [x] change edit view route
 - [x] use new api in frontend
 - [x] adapt task search form
 - [x] task search form persistence? → follow-up https://github.com/scalableminds/webknossos/issues/5337
 - [x] how to access project name for rendering form?
 - [x] adapt access queries
 - [x] rest api versioning
 - [x] evolutions
 - [x] snapshot tests

- https://___.webknossos.xyz

### Steps to test:
- List projects, create project, edit project, should all still work as expected
- Create tasks, search for tasks by project, should show.
- Try to create second project with the same name, should fail with readable message.
- Create second organization
- Create project with name used in other org, should work
- Edit projects, the ones across organizations should not influence each other
- same with task types.

### Issues:
- fixes #4005

------
- [x] Ready for review


youri-k (approved):
LGTM :shipit:

PR #5385, opened 2021-04-08
Fix corruption of skeleton when using CTRL+Rightclick in empty tree

### LINKED ISSUE #5303: Ctrl + Rightclick causes skeleton annotation corruption
## Context

I tried to use the `Ctrl + Rightclick` functionality to create nodes that won't be centered automatically.

## Expected Behavior

A node should be created at the click location without centering it.

## Current Behavior

The message "A corruption in the current skeleton annotation was detected. Please contact your supervisor and/or the maintainers of webKnossos to get help for restoring a working version. Please include as much details as possible about your past user interactions. This will be very helpful to investigate the source of this bug." appears.

## Steps to Reproduce the bug

1. Create/Activate an empty tree
2. Use `Ctrl + Rightclick` to create a first node
3. Use `Ctrl + Rightclick` to create a second node
-> The corruption message appears after a short time.

## Your Environment for bug

* Browser name and version: Chrome 88
* Operating System and version: Ubuntu 18.04
* Version of WebKnossos (Release or Commit): Version: 21.02.0-13936

philippotto:
The bug has two facets:
1. Using ctrl+rightclick to add a tree does not select and also not center it. This means, using ctrl+rightclick in an empty tree twice, will try to create two nodes within the same tree. However, if no node is selected, no edge can be constructed for the second node which produces the corruption.
2. The `addNode` reducer helper simply added a node when being told so and only created an edge, if a node was selected. This means, if no node is selected (for whatever reason), corruption can ensue when the tree is not-empty. This should not happen in theory, but (1) provoked this scenario.

To fix this bug, I did 2 things:
1. Ensure that ctrl+rightclick also selects the first node of a tree (it's still not centered). I think, this is a fair behavior.
2. I adapted the `addNode` code so that the corruption cannot happen even if the circumstances would provoke it. The reducer would simply do a noop instead (and log an error message to the console).

- https://fixcorrupttrees.webknossos.xyz

### Steps to test:
- Create/Activate an empty tree
- Use Ctrl + Rightclick to create a first node
- Use Ctrl + Rightclick to create a second node

### Issues:
- fixes #5303

------
- [X] Ready for review

### DIFF
--- a/CHANGELOG.unreleased.md
+++ b/CHANGELOG.unreleased.md
@@ -21,6 +21,7 @@ For upgrade instructions, please check the [migration guide](MIGRATIONS.released
 - Fixed a bug where some values in the project list were displayed incorrectly after pausing/unpausing the project. [#5339](https://github.com/scalableminds/webknossos/pull/5339)
 - Fixed that editing a task type would always re-add the default values to the recommended configuration (if enabled). [#5341](https://github.com/scalableminds/webknossos/pull/5341)
 - Fixed a bug where tasks created from existing volume annotations that did not have a specified bounding box were broken. [#5362](https://github.com/scalableminds/webknossos/pull/5361)
+- Fixed a bug which could cause corrupted trees when CTRL+Rightclick was used in an empty tree. [#5385](https://github.com/scalableminds/webknossos/pull/5385)
 
 ### Removed
 -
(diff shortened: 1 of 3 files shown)


daniel-wer (approved):
Very nice solution! :+1:

PR #5392, opened 2021-04-13
Await Server Replies When Updating User

### LINKED ISSUE #4988: “email has been changed” toast shows even if request has failed
If you edit a user’s email address in the Users admin view, a failed request leads to both an error toast and a success toast. The success message should not be shown, and the value should not be updated in the list.

fm3:
- https://___.webknossos.xyz

### Steps to test:
- Create second user
- In admin view, activate, deactivate, change email, should work
- try deactivating yourself, should fail, and not show as succeeded
- insert `          _ <- Fox.failure("nope")` in line 319 of UserController.scala
- subsequent user email/activation updates should fail, and should not seem successful in frontend

### Issues:
- fixes #4988 

------
- [x] Ready for review

### DIFF
--- a/frontend/javascripts/admin/user/user_list_view.js
+++ b/frontend/javascripts/admin/user/user_list_view.js
@@ -111,10 +111,7 @@ class UserListView extends React.PureComponent<PropsWithRouter, State> {
     const newUserPromises = this.state.users.map(user => {
       if (selectedUser.id === user.id) {
         const newUser = Object.assign({}, user, { isActive });
-        return updateUser(newUser).then(
-          serverUser => Promise.resolve(serverUser),
-          () => Promise.reject(user),
-        );
+        return updateUser(newUser);
       }
       return Promise.resolve(user);
     });
@@ -139,10 +136,7 @@ class UserListView extends React.PureComponent<PropsWithRouter, State> {
     const newUserPromises = this.state.users.map(user => {
       if (selectedUser.id === user.id) {
         const newUser = Object.assign({}, user, { email: newEmail });
-        return updateUser(newUser).then(
-          serverUser => Promise.resolve(serverUser),
-          () => Promise.reject(user),
-        );
+        return updateUser(newUser);
       }
       return Promise.resolve(user);
     });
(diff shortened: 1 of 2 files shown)


frontend/javascripts/admin/user/user_list_view.js
```diff
@@ -107,48 +107,57 @@ class UserListView extends React.PureComponent<PropsWithRouter, State> {
...
-        isTeamRoleModalVisible: isActive,
-      };
+  activateUser = async (selectedUser: APIUser, isActive: boolean = true) => {
+    const newUserPromises = this.state.users.map(user => {
+      if (selectedUser.id === user.id) {
+        const newUser = Object.assign({}, user, { isActive });
+        return updateUser(newUser).then(
+          serverUser => Promise.resolve(serverUser),
+          () => Promise.reject(user),
+        );
```
philippotto, line 117:
```suggestion
        return updateUser(newUser);
```

I think this should be equivalent for the success case and semantically "better" for the reject case. The argument to `reject` should be the "reason" for why the promise was rejected. It's recommended to use an error instance for that and I think that error instance is passed as the first parameter to the arrow-function (swallowed since you didn't list parameters for the error-handler). If the latter was changed, the entire `then` invocation becomes a no-op, if I'm not mistaken (hence, the above suggestion).

frontend/javascripts/admin/user/user_list_view.js
```diff
@@ -107,48 +107,57 @@ class UserListView extends React.PureComponent<PropsWithRouter, State> {
...
-
-      return {
-        users: newUsers,
-        selectedUserIds: [selectedUser.id],
-      };
+  changeEmail = async (selectedUser: APIUser, newEmail: string) => {
+    const newUserPromises = this.state.users.map(user => {
+      if (selectedUser.id === user.id) {
+        const newUser = Object.assign({}, user, { email: newEmail });
+        return updateUser(newUser).then(
```
philippotto, line 142: same here.

philippotto (approved):
Awesome! Works great & elegant solution :+1: I just added one small suggestion.

Issue #5368, opened 2021-04-07
Antd Upgrade Followup


daniel-wer:
Now that antd was upgraded to v4, we can use the loading prop of Selects instead of the notFoundContent workaround introduced in #4671

PR #5395, opened 2021-04-13
Fix style of tracing-button-group

philippotto:
The rounded corners were wrong and the overflow behavior was also missing.
Before:
image

After:
image

The overflow behavior is still not optimal, but I'd argue that the change in the antd upgrade was not really intentional.

### Issues:
- contributes to #5368 

------
- [X] Ready for review


daniel-wer (approved):
:+1:

PR #5396, opened 2021-04-14
Support arbitrary modes even when data layer is not uint8

### LINKED ISSUE #5144: Oblique and arbitrary modes are disabled if color layer is not uint8
Oblique and arbitrary modes are disabled if color layer is not uint8.

In `determineAllowedModes` the following code exists:
```
if (colorLayer != null && colorLayer.elementClass !== "uint8") {
```

Not sure whether there is a valid reason for this restriction, but I hope we can simply remove this.

philippotto:
The check which I removed in this PR was faulty anyway. It only checked the datatype of the "first" color layer. This means that we already supported non-uint8 layers in arbitrary modes as long as the "first" color layer was uint8.
I also had a look at a float layer in arbitrary mode and couldn't spot any problems (also I don't know why there should be a problem). I assume this is a very old relic from CPU-based rendering (?) which is obsolete by now.

- https://supportfloatinarbitrary.webknossos.xyz

### Steps to test:
- open a dataset with float layer and switch to an arbitrary mode

### Issues:
- fixes #5144

------
- [X] Ready for review


daniel-wer (approved):
Nice find! For fun I tried to trace it back, this was the earliest I could find: https://github.com/scalableminds/webknossos/commit/93c01a0a126f4a0577a4858c55c2ab38818604c2 - Seriously old :older_adult:

philippotto:
Wow! Nice research :smile:

Issue #4787, opened 2020-08-28
Rework skeleton and volume toolbar


philippotto:
The current tools for skeleton and volume tracings are not very explorable. A lot of behavior is hidden behind shortcuts and or not intuitive modes. For example, to create skeletons you have to be in the "move" tool :see_no_evil: And the existence of the upcoming fill-tool will also be very hard to find out.

There are multiple ways to improve this situation (and we might want to do this in multiple stages).

My vision for a round experience would be:
- have a tool for each dedicated action and use icons instead of labels (e.g.,  the new fill-tool should have an own button)
  - theoretically, actions such as "removing an edge" or "merging two trees" could be a separate tool, too (using the existing shortcuts could still work)
- categorize the tools according to skeleton vs volume. maybe, it could also make sense to have a general skeleton-vs-volume mode button (so, you are either in the skeleton or in the volume mode). I think, this would also be intuitive for users, since one usually focuses on one tracing type (even in hybrids). Additionally, it would save space in the navbar.
- different modes within these tools, should be presented in the navbar as additional buttons (see screenshot below for how Gimp does this)
- in general, all shortcuts which modify the tool behavior, should immediately reflect the state change in the toolbar (also comparable to how gimp does it). For example, if you press a modifier while using the brush tool, the mode button would switch to show that you are overwriting only the current cell.

![image](https://user-images.githubusercontent.com/2486553/91549847-e3056780-e927-11ea-8525-595994ada36e.png)

normanrz:
Making a toolbar with icons is a good idea. This would also match 99% of existing drawing tools.

hotzenklotz:
I agree. 

(While we are at it: Is merge mode also a tool? Or is it a state? If it is a tool, it should also have a button)

philippotto:
> (While we are at it: Is merge mode also a tool? Or is it a state? If it is a tool, it should also have a button)

I thought about that too and I think that it is a special mode (or state), because one can be in merger mode and still want to switch between moving, creating/deleting nodes (and theoretically also brushing).

PR #5397, opened 2021-04-15
[1/n] Dedicated Skeleton Tool

philippotto:
- https://___.webknossos.xyz

### Steps to test:
- blocked until #5403 is merged

### Issues:
- contributes to #4787 

------
- [X] Ready for review


philippotto:
@daniel-wer maybe we can talk about my work in this pr (and https://github.com/scalableminds/webknossos/pull/5403) on Monday :) There's still a lot of spaghetti going on and it's probably hard to review, but I don't see any good way to avoid this, since a lot of code has to be moved to make things better. The two PRs might be good chunks semantically, but let's talk about this in person :call_me_hand:

frontend/javascripts/oxalis/api/api_latest.js
```diff
@@ -834,24 +834,22 @@ class TracingApi {
   /**
    * Returns the active volume tool which is either
    * "MOVE", "TRACE" or "BRUSH".
-   * _Volume tracing only!_
    */
-  getVolumeTool(): ?VolumeTool {
-    const tracing = assertVolume(Store.getState().tracing);
-    return getVolumeTool(tracing);
+  getAnnotationTool(): AnnotationTool {
```
philippotto, line 838: todo: be backwards compatible.

philippotto: done

philippotto:
@daniel-wer As discussed, feel free to review this PR. In the meantime, I'll continue with #5403. Testing can wait until both PRs are merged with each other.

frontend/javascripts/oxalis/controller/viewmodes/plane_controller.js
```diff
@@ -155,6 +154,22 @@ class PlaneController extends React.PureComponent<Props> {
 
   getPlaneMouseControls(planeId: OrthoView): Object {
     const defaultDragHandler = (delta: Point2) => movePlane([-delta.x, -delta.y, 0]);
+    const rightClickHandler = (
+      pos: Point2,
+      plane: OrthoView,
+      event: MouseEvent,
+      isTouch: boolean,
+    ) => {
+      console.log("skeletonController.onContextMenu");
```
daniel-wer, line 163:
```suggestion
```

frontend/javascripts/oxalis/controller/viewmodes/plane_controller.js
```diff
@@ -539,22 +552,22 @@ class PlaneController extends React.PureComponent<Props> {
...
     return (...args) => {
-      if (skeletonHandler && volumeHandler) {
-        // Deal with both modes
-        const tool = Utils.enforce(getVolumeTool)(this.props.tracing.volume);
-        if (tool === VolumeToolEnum.MOVE) {
+      const tool = this.props.tracing.activeTool;
+      if (tool === AnnotationToolEnum.MOVE) {
+        if (viewHandler != null) {
+          viewHandler(...args);
+        } else if (skeletonHandler != null) {
```
daniel-wer, line 559: Does this mean that there is `skeletonHandler` functionality that should be executed in MOVE mode? I would've expected this to no longer be the case after this PR. Or will this only be cleaned up in the other PR?

frontend/javascripts/oxalis/model/actions/volumetracing_actions.js
```diff
@@ -23,7 +23,7 @@ type AddToLayerAction = { type: "ADD_TO_LAYER", position: Vector3 };
 type FloodFillAction = { type: "FLOOD_FILL", position: Vector3, planeId: OrthoView };
 type FinishEditingAction = { type: "FINISH_EDITING" };
 type SetActiveCellAction = { type: "SET_ACTIVE_CELL", cellId: number };
-type SetToolAction = { type: "SET_TOOL", tool: VolumeTool };
+type SetToolAction = { type: "SET_TOOL", tool: AnnotationTool };
```
daniel-wer, line 26: These two actions should be moved to the annotation actions (as you did for the respective reducers)

frontend/javascripts/oxalis/model/reducers/reducer_helpers.js
```diff
@@ -126,5 +133,22 @@ export function convertServerAnnotationToFrontendAnnotation(annotation: APIAnnot
     tracingStore,
     meshes,
     user,
+    // todop: activeTool needs to be stored in annotation (on server)
```
daniel-wer, line 136: Looks like this is still an active todo

frontend/javascripts/test/api/api_skeleton_latest.spec.js
```diff
@@ -199,7 +199,7 @@ test("Utils Api: registerOverwrite should overwrite an existing function", t =>
 
 test("Calling a volume api function in a skeleton tracing should throw an error", t => {
   const api = t.context.api;
-  t.throws(() => api.tracing.getVolumeTool());
+  t.throws(() => api.tracing.getAnnotationTool());
```
daniel-wer, line 202: I would assume this test to fail, since `getAnnotationTool()` no longer only works in volume tracings. A different volume-only api method should be chosen for this test.

frontend/javascripts/test/api/api_volume_latest.spec.js
```diff
@@ -24,24 +24,23 @@ test("setActiveCell should set the active cell id", t => {
   t.is(api.tracing.getActiveCellId(), 27);
 });
 
-test("getVolumeTool should get the current tool", t => {
-  const { api } = t.context;
-  t.is(api.tracing.getVolumeTool(), VolumeToolEnum.MOVE);
-});
```
daniel-wer, line 30: I think this line was removed in error. Now the next test is somehow inside this one if I see it correctly.

frontend/javascripts/test/reducers/volumetracing_reducer.spec.js
```diff
@@ -240,46 +238,36 @@ test("VolumeTracing should not allow to set trace tool if getRequestLogZoomStep(
...
+
+  // Tool should not have changed
+  t.is(newState.tracing.activeTool, AnnotationToolEnum.MOVE);
 });
 
 test("VolumeTracing should cycle trace/view/brush tool", t => {
   const cycleToolAction = VolumeTracingActions.cycleToolAction();
 
   // Cycle tool to Brush
   let newState = VolumeTracingReducer(initialState, cycleToolAction);
```
daniel-wer, line 250: I would expect two cycling actions to be needed now with the new skeleton tool in between move and brush.

daniel-wer (commented):
Very cool work! Nice to see how the generalization of the tool concept simplifies the code in lots of places :)

frontend/javascripts/oxalis/controller/viewmodes/plane_controller.js
philippotto: The second PR improves this a bit, but in the end it's still a bit mangled. I'll try to summarize it:

- This particular function (with the behavior that skeleton-related handlers are used as fallback when the move tool is active) is only used for keyboard-bindings thanks to the second PR. This means, hitting "c" in the move tool will create a new tree. This is probably up for debate, but it's quite compatible with the old behavior.
- With the new PR, there is still some entanglement of the move tool with other tools. For example, the `scroll` handler of the move tool is also responsible for changing the brush size (if `shift` is pressed). If two tools registered a `scroll` handler, this would be a conflict (which cannot be resolved using the active tool alone). Therefore, I went with avoiding these conflicts by implementing the bindings in the most generic tool. Luckily, this is only the case for very few handlers.

frontend/javascripts/oxalis/model/actions/volumetracing_actions.js
philippotto: good point!

frontend/javascripts/test/reducers/volumetracing_reducer.spec.js
philippotto: on this branch, the test fails indeed, but in the other PR I took care that the cycle-tool functionality only cycles through the enabled/available tools. in a volume annotation, the skeleton tool won't be available, which is why the test works again.

philippotto (commented):
Thank you for your feedback! Some of it is already addressed in #5403 and most of the remaining stuff was just added by my in https://github.com/scalableminds/webknossos/pull/5403/commits/ec0a4f480e569644d22bb79f608c96287df7113f. I integrated your feedback in my other PR to avoid conflicts. I hope this isn't too chaotic. The only remaining thing is the persistence of the annotation tool. I'll take care of this now.

philippotto:
> The only remaining thing is the persistence of the annotation tool. I'll take care of this now.

See https://github.com/scalableminds/webknossos/pull/5403#issuecomment-824772374

frontend/javascripts/oxalis/controller/viewmodes/plane_controller.js
daniel-wer: Thank you for explaining! This seems to be a good solution for now :+1:

philippotto:
Contained in #5422.

PR #5402, opened 2021-04-16
Call webknossos-worker Generate Mesh File Job

fm3:
Call webknossos-worker job to create mesh file from an existing segmentation layer.

Note that this PR does not yet include a UI for this, which will presumably be added later after #5346 is done.

### Steps to test:
- set up https://github.com/scalableminds/webknossos-worker/pull/15
- set jobsEnabled = true
- call `/api/jobs/run/computeMeshFile/sample_organization/l4_sample?layerName=segmentation&mag=4-4-1&agglomerateView=agglomerate_view_20` (replace by values pointing to existing data. agglomerateView is optional)
- job should succeed, dataset should have meshfile

------
- [x] Ready for review


app/controllers/JobsController.scala
```diff
@@ -246,6 +246,29 @@ class JobsController @Inject()(jobDAO: JobDAO,
...
+  def runComputeMeshFileJob(organizationName: String,
+                            dataSetName: String,
+                            layerName: String,
+                            mag: String,
+                            agglomerateView: Option[String]): Action[AnyContent] =
+    sil.SecuredAction.async { implicit request =>
+      for {
+        organization <- organizationDAO.findOneByName(organizationName) ?~> Messages("organization.notFound",
+                                                                                     organizationName)
+        _ <- bool2Fox(request.identity._organization == organization._id) ?~> "job.export.notAllowed.organization" ~> FORBIDDEN
```
youri-k, line 258: `job.export.notAllowed.organization` message is currently limited to tiff export. Maybe word the message differently or create a new one.

youri-k (changes requested):
LGTM except the wrong error message :tada:

youri-k (approved):
Perfect :tada:

frontend/javascripts/admin/admin_rest_api.js
```diff
@@ -876,6 +876,23 @@ export async function startExportTiffJob(
...
 
+export function startComputeMeshFileJob(
+  organizationName: string,
+  datasetName: string,
+  layerName: string,
+  mag: Vector3,
+  agglomerateView?: string,
+) {
+  // TODO: figure out what's going on with the segmentation layer (only fallback "segmentation" is found by worker)
+  // TODO: figure out why this doesn't work at all... :( Not even with the default example
```
philippotto, line 887: @grittaweisheit do you need help with this or could you resolve this together with @fm3 last Friday?

grittaweisheit: We figured out the second one :)
But I'm not sure what's the intended behavior regarding the layer. Should the meshfile always be computed from the fallback (`segmentation`) layer, or is there a case where it should be another layer?

fm3: I’d say it should always be the segmentation layer (fallback in case of volume annotation, for now), but note that it is not always named `segmentation`. That’s why the named should be passed.

frontend/javascripts/oxalis/view/right-menu/meshes_view.js
```diff
@@ -171,6 +179,20 @@ class MeshesView extends React.Component<
       Store.dispatch(setPositionAction(seedPosition));
     };
 
+    const getComputeMeshFileButton = () => (
+      <Button
```
philippotto, line 183: I don't know whether this is already on your to do list, but this button should ideally:
- disappear (or be hidden a bit) if a mesh file was already computed (because then the button is usually not interesting to the user)
- have `loading={true}` if a compute-mesh-job is currently running (for now, you can simply call `getJobs()` and filter by `datasetName` every 60 seconds or so)

Regarding the second point: @fm3 do you think it makes sense to adapt `/api/jobs` so that it optionally accepts a datasetName parameter? Like `/api/jobs?datasetName=...`?

frontend/javascripts/oxalis/view/right-menu/meshes_view.js
```diff
@@ -171,6 +179,20 @@ class MeshesView extends React.Component<
...
+        onClick={() => {
+          startComputeMeshFileJob(
```
philippotto, line 185: in addition to calling this function, a simple toast should also appear. maybe something like this:

```
Toast.success("The computation of a mesh file was started successfully. Depending on the datset's size this may take a bit. You don't need to keep this dataset open for the computation to continue.");
```

frontend/javascripts/oxalis/view/right-menu/meshes_view.js
fm3: Hmm, I’m not sure. Not every job is necessarily associated with a (single) dataset, so maybe we’ll have to find a more generic solution to this filtering 🤔

philippotto: Hm, I see your point. I think that a generic approach could be one of the following:
(1) allow arbitrary filtering (i.e., a get parameter for each field is supported) <-- the front-end would responsible for defining the correct filters (for the compute-mesh case, it would make sense to filter by jobType==compute_mesh + dataset=current_dataset)
(2) provide use-case-driven routes (e.g., "fetch mesh jobs for mesh ui" --> `/api/jobs/mesh?datasetName=...`; "apply merger mode" --> `api/jobs/mesh?annotationId=...`)

So, my above suggestion to allow filtering by datasetName would be a subcase of (1)

grittaweisheit:
Do we still want to be able to import stl files for single meshes, now that we are able to use whole mesh files?  Otherwise we could reuse that button/position for computing mesh files. Then it wouldn't be as prominent in the first place :thinking:

philippotto:
> Do we still want to be able to import stl files for single meshes, now that we are able to use whole mesh files? Otherwise we could reuse that button/position for computing mesh files.

My gut reaction would be: keep the support for the old stl import, but make it less prominent (could be in a dropdown).

> Then it wouldn't be as prominent in the first place

the compute-mesh button can (and should probably) be prominent when no mesh file was computed yet. it should only become less prominent once there is already a mesh file. I would suggest to do the following:

- if no mesh file exists: show "compute mesh" (and hide the button "load precomputed mesh" since this would be disabled anyway). within a dropdown, the old "stl import" could be listed.
- if a mesh file exists: show "load precomputed mesh" and move the "compute mesh" item into the dropdown (next to the "stl import").

frontend/javascripts/oxalis/view/right-menu/meshes_view.js
grittaweisheit: For now I filtered the jobs by datasetName, type and state. :)

grittaweisheit:
Should the `jobsEnabled` config in the `application.conf` be set to true or stay false for now?

fm3:
It should stay `false` in the repo. In the production config it is overwritten.

frontend/javascripts/oxalis/view/right-menu/meshes_view.js
```diff
@@ -163,6 +197,29 @@ class MeshesView extends React.Component<Props, State> {
...
+        "The computation of a mesh file was started successfully. Depending on the dataset's size this may take a bit. You don't need to keep this dataset open for the computation to continue.",
+      );
+    };
+
+    const getComputeMeshFileButton = () => (
+      <Button
+        size="small"
+        loading={this.state.isComputingMeshfile}
+        onClick={startComputingMeshfile}
+      >
```
philippotto, line 218: I think, this button will also appear (and be clickable) if the wk instance does not support jobs. I suggest to use `features().jobsEnabled` to make this button potentially `disabled`. In that case, the button should display a tooltip on hover which says something like this:

`The computation of mesh files is not supported by this webKnossos instance.`

philippotto (dismissed):
The code looks good! However, I noticed that the compute-button needs to be disabled if jobs are disabled. Also, the "computation finished"-toast is not yet implemented (or pushed?) yet. Will test if that's included, too.

frontend/javascripts/oxalis/view/node_context_menu.js
```diff
@@ -226,9 +227,14 @@ function NoNodeContextMenuOptions({
...
       key="load-mesh-file"
       onClick={loadMesh}
-      disabled={!currentMeshFile}
+      disabled={!currentMeshFile || !features().jobsEnabled}
     >
-      Load Precomputed Mesh
+      <Tooltip
+        visible={!features().jobsEnabled}
+        title="The computation of mesh files is not supported by this webKnossos instance."
+      >
```
philippotto, line 235: loading a mesh file should work even if jobs are disabled (only computation of meshes is bound to jobs), since one can manually put a mesh file on disk even without jobs.

frontend/javascripts/oxalis/view/right-menu/meshes_view.js
```diff
@@ -163,6 +212,40 @@ class MeshesView extends React.Component<Props, State> {
...
+        this.props.activeResolution,
+      );
+      this.setState({ isComputingMeshfile: true });
+      Toast.success(
+        "The computation of a mesh file was started successfully. Depending on the dataset's size this may take a bit. You don't need to keep this dataset open for the computation to continue.",
+      );
+    };
+
+    const getJobsDisabledTooltip = node => (
+      <Tooltip
```
philippotto, line 229: Won't this always show the tooltip (even without hovering) if jobs are disabled? I think, you want something like this:

```js
      <Tooltip
        title={features().jobsEnabled ? "Compute a mesh file for this dataset so that meshes for segments can be loaded quickly afterwards." : "The computation of mesh files is not supported by this webKnossos instance."}
      >
```

frontend/javascripts/oxalis/view/right-menu/meshes_view.js
```diff
@@ -124,15 +136,52 @@ type DispatchProps = ExtractReturn<typeof mapDispatchToProps>;
...
+          job.state === "STARTED",
+      ).length > 0;
+    if (wasComputingMeshfile && !isComputingMeshfile) {
+      Toast.info(
+        <React.Fragment>
+          A meshfile generation job for this dataset has finished. Click{" "}
+          <a href="/jobs" target="_blank">
+            here
+          </a>{" "}
+          to see all jobs.
```
philippotto, line 171:
```suggestion
          The requested computation of a mesh file for this dataset has finished. You can now use the "Load Precomputed Mesh" in the "Meshes" tab.
```

grittaweisheit: I'm not sure this sends the right message. We currently don't check, whether it's the requested computation. It only indicates, that there was some running meshfile job for this dataset, and now there's not. That's why I included the link to the jobs, so one could see how the job ended.
We would need to save the job id when starting a job and then check with the job id, if the job is still running, was successful or failed. That's possible, but not the case right now.

frontend/javascripts/oxalis/view/right-menu/meshes_view.js
grittaweisheit: Oh, yes. I don't know how this slipped by without me noticing. Thanks for pointing it out!

frontend/javascripts/oxalis/view/right-menu/meshes_view.js
philippotto: Good point! How about:

```suggestion
          The computation of a mesh file for this dataset has finished. You can now use the "Load Precomputed Mesh" in the "Meshes" tab.
```

This doesn't imply that the user actually requested that mesh file and simply announces that the feature is available now. What do you think? In my opinion, there is no great use in delegating the user to the jobs page, since the exciting stuff happens in the mesh tab :)

frontend/javascripts/oxalis/view/right-menu/meshes_view.js
```diff
@@ -163,6 +213,49 @@ class MeshesView extends React.Component<Props, State> {
       Store.dispatch(setPositionAction(seedPosition, null, false));
     };
 
+    const startComputingMeshfile = async () => {
+      const availableResolutions = getResolutions(this.props.dataset);
+      const meshfileResolution =
+        availableResolutions.find(res => _.isEqual(res, defaultMeshfileGenerationResolution)) !=
+        null
+          ? defaultMeshfileGenerationResolution
+          : this.props.activeResolution;
```
philippotto, line 222: There are two small problems with this code:
- if the mag 4-4-2 exists, it is not used, even though it would probably be a good idea to use it
- if the `defaultMeshfileGenerationResolution` mag does not exist, this simply defaults to the active resolution (which might be mag-1 which is resource-intense)

Both issues can be resolved with `getDatasetResolutionInfo(this.props.dataset).getIndexOrClosestHigherIndex(defaultMeshfileGenerationResolutionIndex)` with `const defaultMeshfileGenerationResolutionIndex = 2` (because 2**2 = mag 4).

philippotto:
@daniel-wer Could you review the two commits I added? Predominantly, I changed the polling logic a bit so that the code remembers for which job id it waits. If that job changes its state, a corresponding toast is shown. The polling even works after reloading the page (this is not new), which also has the side-effect that other users might be notified about a finished mesh job. However, this should be okay.

@fm3 I opened a PR in the worker repository which should be merged and deployed before this PR is merged.

fm3:
The mentioned worker fixes are merged and deployed :rocket:

daniel-wer (commented):
@philippotto's last two commits look good to me. @fm3 Did you already do or are you going to do a final test whether this works as expected?

fm3:
I can do a final test tomorrow

fm3:
I pushed some small changes to the UI, could you have another look? I also tested and noticed that the worker needs the new voxelytics release for this to work as expected. So I created the new worker PR https://github.com/scalableminds/webknossos-worker/pull/37 that is now in review

daniel-wer:
Thank you @fm3, frontend looks good to me :+1:

PR #5403, opened 2021-04-16
[2/n] Refactor mouse controls

philippotto:
- https://___.webknossos.xyz

### To do:

- [X] refactor mouse bindings
	- [X] decouple bindings from actions
	- [X] remove some complexity from plane_controller (i.e., remove zoom-code)
- [X] only show toolbar in ortho-view
- [X] make tool-cycling respect disabled tools
	- [X] move tool-specific code into own accessors module
- [X] fix tests
- [X] integrate with / update to new statusbar
- [X] store active tool only temporary in uiInformation
- [X] add legacy-keybinding config
	- [X] right now, it only controls whether the skeleton tool should be enabled by default

### Steps to test:
- let's go through a review-round first

### Issues:
- contributes to #4787

------
- [X] Ready for review


philippotto:
@daniel-wer I think this PR is ready for review. It mostly contains refactorings, fixes some bugs and also adds the legacy binding config. We will probably need a migration for this, so that old users get that.

I discussed with @fm3 that persisting the current tool is not really necessary for now. I built it so that the MOVE tool is always selected by default **unless one has enabled the legacy-bindings and opened a skeleton-only annotation**. I hope this is an adequate solution.

Next, I will open a 3rd PR which makes use of the legacy-binding switch to allow context-menu everywhere etc. I have yet to figure out how to handle the erase modes for brush/trace. Not sure whether these will become completely indpendent tools in the UI :thinking:

frontend/javascripts/oxalis/controller/combinations/tool_controls.js
```diff
@@ -0,0 +1,271 @@
...
+          }
+          default: // ignore other cases
+        }
+      },
+      over: () => {
+        MoveHandlers.handleOverViewport(planeId);
+      },
+      pinch: delta => MoveHandlers.zoom(delta, true),
+      mouseMove: (delta: Point2, position: Point2, id, event) => {
+        // Always set the correct mouse position. Otherwise, using alt + mouse move and
```
daniel-wer, line 76: As this will result in many store updates, maybe we should check that there is no component that is rendered on every store update. Using the React Dev Tools with the "Highlight Updates" setting should show that.
However, it's probably not that important as this used to be done already in volume tracings to update the displayed segment id.

frontend/javascripts/oxalis/controller/combinations/tool_controls.js
```diff
@@ -0,0 +1,271 @@
...
+  static onClick(
+    planeView: PlaneView,
+    position: Point2,
+    shiftPressed: boolean,
+    altPressed: boolean,
+    ctrlPressed: boolean,
+    plane: OrthoView,
+    isTouch: boolean,
+  ): void {
+    if (!shiftPressed && !isTouch && !ctrlPressed) {
```
daniel-wer, line 156: I think it would be easier to understand if the `else` branch further down would be an `else if` and the branch conditions would show the full conditions that's needed. For example, for the handleMergeTrees alt **and shift** (or ctrl would actually work as well?!) need to be pressed which is very hard to see right now.

frontend/javascripts/oxalis/model/accessors/tool_accessor.js
```diff
@@ -0,0 +1,124 @@
...
+  const isZoomStepTooHighForTracing = isZoomStepTooHighFor(state, AnnotationToolEnum.TRACE);
+  const isZoomStepTooHighForFilling = isZoomStepTooHighFor(state, AnnotationToolEnum.FILL_CELL);
+
+  return {
+    [AnnotationToolEnum.MOVE]: {
+      isDisabled: false,
+      explanation: "",
+    },
+    [AnnotationToolEnum.SKELETON]: {
+      isDisabled: !hasSkeleton,
```
daniel-wer, line 104: Does this mean the skeleton tool is shown (but disabled) in volume-only tracings, whereas the volume tools are not shown in skeleton-only tracings?

frontend/javascripts/oxalis/model_initialization.js
```diff
@@ -280,6 +285,23 @@ function initializeTracing(_annotation: APIAnnotation, tracing: HybridServerTrac
...
+  const { useLegacyBindings } = Store.getState().userConfiguration;
+
+  if (!useLegacyBindings) {
+    // The MOVE tool is already the default
+    return;
+  }
+
+  const { tracing } = Store.getState();
+
+  if (tracing.skeleton != null && tracing.volume == null) {
```
daniel-wer, line 298: Wouldn't the current state be more accurately reflected if in legacy mode the skeleton tool would always be the default (except in volume-only)?

frontend/javascripts/oxalis/view/settings/user_settings_view.js
```diff
@@ -445,6 +445,11 @@ class UserSettingsView extends PureComponent<UserSettingsViewProps> {
             value={this.props.userConfiguration.dynamicSpaceDirection}
             onChange={this.onChangeUser.dynamicSpaceDirection}
           />
+          <SwitchSetting
+            label={settingsLabels.useLegacyBindings}
```
daniel-wer, line 449: What do you think about a small description of what this entails? Or do we simply say, the "old" people know what this means (and have it activated automatically) and nobody else should activate it?

daniel-wer: Thinking about this: Existing organizations may want it that this setting is automatically set even for newly registered users, because they have training material referencing the legacy mouse/keyboard handlers.

frontend/javascripts/test/api/api_volume_latest.spec.js
```diff
@@ -25,19 +25,20 @@ test("setActiveCell should set the active cell id", t => {
 });
 
 test("getAnnotationTool should get the current tool", t => {
-  const {api} = t.context;
+  const { api } = t.context;
```
daniel-wer, line 28: Did prettier not run for this file before? :thinking:

daniel-wer (approved):
Awesome :+1: I wasn't able to really review the large amount of moved code, but had a look at everything else. I left some smaller comments and clarifying questions. Didn't test yet as we'll probably do that once properly in the end :)

frontend/javascripts/oxalis/controller/combinations/segmentation_handlers.js
```diff
@@ -85,21 +84,3 @@ export async function loadAgglomerateSkeletonAtPosition(position: Vector3) {
 
   await progressCallback(true, "Skeleton generation done.");
 }
-
-export function isosurfaceLeftClick(pos: Point2, plane: OrthoView, event: MouseEvent) {
```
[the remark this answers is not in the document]
philippotto: `isosurfaceLeftClick` wasn't moved, but was completely removed since it wasn't used anymore. I think, the code become dead due to the new mesh UI we added some time ago.

frontend/javascripts/oxalis/controller/combinations/tool_controls.js
philippotto: Good point! I think the store update itself is not that bad since (as you say) it's already happening for volume tracings (which is the default due to hybrids). However, the "highlight updates" hint is still very relevant. I noticed that the toolbar updates now on each mouse move. I added this to-do element to #5422 and will tackle this after the other functional changes have been finished.

frontend/javascripts/oxalis/model/accessors/tool_accessor.js
philippotto: With #5422 this becomes a bit cleaner. Then, the logic is quite simple: skeleton-tools are shown if there is a skeleton and volume tools are shown if there's a volume annotation. The move-tool is always shown obviously.
The `isDisabled` property is still computed here, since that  information is used when cycling the current tool.

frontend/javascripts/test/api/api_volume_latest.spec.js
philippotto: I did that change manually since my linter in sublime complained :see_no_evil: With our project-config, eslint marks these only as a warning (which is why the CI passes).

frontend/javascripts/oxalis/controller/combinations/tool_controls.js
philippotto: Very good point! Historically grown, I guess. I rewrote the logic per your suggestion and also noticed that one can select a node with a mere tap (i.e., touch). Wasn't that obvious, either, when reading the code.
I pushed the changes to #5422. See 605e044a1. Probably also fixed that ctrl and alt could also merge trees.

frontend/javascripts/oxalis/model_initialization.js
philippotto: Yes, good point. Not sure whether there was a reason against that. We probably missed that variation or I mixed something up. Do you agree, @fm3? I implemented the change in 3dafe17f2fcdc5e8c76fc0889fa712681a012be2.

frontend/javascripts/oxalis/view/settings/user_settings_view.js
philippotto: > What do you think about a small description of what this entails? Or do we simply say, the "old" people know what this means (and have it activated automatically) and nobody else should activate it?

Good idea! I think, this definitely needs an explanation, so the user isn't left puzzled. I'll add one.

> Thinking about this: Existing organizations may want it that this setting is automatically set even for newly registered users, because they have training material referencing the legacy mouse/keyboard handlers.

Hmm, good point. If we are lucky we could count on the recommended user config which should also cover new users. However, a proper organization-wide setting would probably be safer.. I'll add this to the todo list.

frontend/javascripts/oxalis/model_initialization.js
fm3: I don’t remember exactly what we discussed about this previously, but yes, seems fair to me right now.

philippotto:
Contained in #5422.

PR #5350, opened 2021-03-30
Upgrade antd to version 4

MichaelBuessemeyer:
Upgrade UI library antd to version 4, creating a slightly more modern look and behavior of many UI elements and reducing technical debt.

### URL of deployed dev instance (used for testing):
- https://upgradeantd.webknossos.xyz

### Steps to test:
- see comment below, https://github.com/scalableminds/webknossos/pull/5350#issuecomment-814065777

### Issues:
- fixes #4678
- fixes #5165 🎉 

------
- [x] Updated [(unreleased) changelog](../blob/master/CHANGELOG.unreleased.md#unreleased)
- [X] Ready for review

PR #5404, opened 2021-04-16
Only upload to datastores that allow upload

fm3:
Fixed a bug where dataset uploads were sent to the wrong datastore (first in list, regardless of allowsUpload), and thus failed.

~I’m a bit confused why this bug didn’t occur before. The datastore list from the server is ordered by name, so the “correct” one, webknossos.org is sent last, but the old code took the first.~ regression introduced in #5350 (for explanation see comment below)

### Steps to test:
- set up external datastore, insert into db with allowsUpload = false
- upload dataset, should go to local datastore
- toggle allowsUpload for both datastes
- reload upload page, upload should now go to the other datastore.

------
- [x] Ready for review

### DIFF
--- a/CHANGELOG.unreleased.md
+++ b/CHANGELOG.unreleased.md
@@ -32,6 +32,7 @@ For upgrade instructions, please check the [migration guide](MIGRATIONS.released
 - Fixed artifacts in screenshots near the dataset border. [#5324](https://github.com/scalableminds/webknossos/pull/5324)
 - Fixed a bug where the page would scroll up unexpectedly when showing various confirm modals. [#5371](https://github.com/scalableminds/webknossos/pull/5371)
 - Fixed a bug where user changes (email, activation) would show as successful even if they actually failed. [#5392](https://github.com/scalableminds/webknossos/pull/5392)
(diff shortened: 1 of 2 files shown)


MichaelBuessemeyer:
I just had a quick look at the previous versions of the code:
There are / were at least two versions with that bug.
The most recent version came with the antd upgrade and is currently deployed on the master.

Before the antd upgrade, the code was:
```
 static getDerivedStateFromProps(props) {
    if (
      props.datastores.length === 1 &&
      props.form.getFieldValue("datastore") !== props.datastores[0].url
    ) {
      props.form.setFieldsValue({ datastore: props.datastores[0].url });
    }
    return null;
  }
```
As you can see, I switched from `getDerivedStateFromProps` to `componentDidUpdate` during the antd upgrade.
One significant reason why the bug now occurred on the master might be, that previous to the antd upgrade's version the `datastore` field in the form was **only** updated, if there was exactly one datastore. During the antd upgrade, I changed this to always select a datastore once at least one is available. And this is pretty likely the reason why the bug was only present on wk.org with the antd upgrade because wk.org has more than just one datastore.

@fm3 fix should work imo as it does the required filtering for datastores that allow uploads.

philippotto (approved):

PR #5407, opened 2021-04-19
Dark theme

normanrz:
- Implements a dark theme for webKnossos through a separate stylesheet
- Tweaks the necessary styles for the UI components to adapt to both light and dark themes, i.e.
  - move color codes out of JS, if possible
  - replace specific hex/rgb codes with antd-provided variables (see https://antdtheme.com/default#Color and https://ant.design/docs/spec/colors)
- Reorganizes some stylesheets, e.g.
  - remove unused CSS styles
  - move spotlight stuff out of `main.less`
  - add a `_utils.less` stylesheet for exposing antd-variables as CSS variables
- The charts in `time_line_chart_view.js` and `statistic_view.js` just got a white background, because the Google Chart module doesn't allow easy theming. Should be fixed in a follow-up PR.

### TODO
- [x] persist mode (backend)
- [x] publication gallery
- [x] spotlight_view
- [x] welcome header
- [x] onboarding.js
- [x] dataset_upload_view.js
- [x] dataset_import_view.js
- [x] stacked_bar_chart.js
- [x] nml_upload_zone_container.js
- [x] viewport_status_indicator.js
- [x] dataset_position_view.js
- [x] dataset_settings_view.js
- [x] time_line_chart_view.js
- [x] experience_modal.js
- [x] dataset_action_view.js
- [x] toast.js
- [x] save_button.js
- [x] tracing_layout_view.js
- [x] dataset_info_tab.js
- [x] meshes_view.js
- [x] setting_input_view.js
- [x] share_modal.js
- [x] mapping_info_view.js
- [x] notifications
- [x] context menu
- [x] info tab
- [x] isosurface tab
- [x] tree tab
- [x] abstract tree tab (fixed with a white background)
- [x] segmentation tab
- [x] version restore
- [x] tracing-view navbar
- [x] Merge with #5279 
- [x] Merge with #5369 
- [x] Merge with #5420 

- https://darktheme.webknossos.xyz

### Steps to test:
- Open https://darktheme.webknossos.xyz, switch to dark mode and try everything :-)

------
- [x] Ready for review


normanrz:
Although there are still a few TODOs, I think the code is ready for a second pair of eyes.

frontend/javascripts/router.js
```diff
@@ -449,7 +445,6 @@ class ReactRouter extends React.Component<Props> {
                   return <FinishResetPasswordView resetToken={params.token} />;
                 }}
               />
-              <Route path="/spotlight" component={SpotlightView} />
```
philippotto, line 452: so, this route is not supported anymore, correct?

philippotto (commented):
Impressive work :bee: :bee: :bee: I had a look at the code and everything looks solid. Great clean up, too!

Next, I will test the dark mode a bit and maybe tweak a few things.

@youri-k Can you review the few back-end changes?

@normanrz Should we declare the dark mode as "experimental" for the first release? this might make users more forgiving when they find weird spots in the dark theme..

frontend/javascripts/router.js
normanrz: Yeah. I don't think it was linked anywhere anymore.

frontend/javascripts/oxalis/controller/renderer.js
```diff
@@ -15,6 +15,7 @@ function getRenderer() {
           // This prevents flickering when rendering to a buffer instead of the canvas
           preserveDrawingBuffer: true,
           antialias: true,
+          alpha: true,
```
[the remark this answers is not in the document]
normanrz: Not sure if that would cost more resources. We might also get the same result by setting a black background in Three.

normanrz:
> @normanrz Should we declare the dark mode as "experimental" for the first release? this might make users more forgiving when they find weird spots in the dark theme..

Yeah, that is a good idea.

youri-k (commented):
Backend Code LGTM :tada:

philippotto:
I just pushed several tweaks which caught my eye during testing. From my point of view this PR can be merged. @fm3 Could you double-check the light theme to ensure that no embarrassing regression is introduced? The dark theme is marked as beta, so we are relatively safe there.

frontend/javascripts/oxalis/view/components/border_toggle_button.js
```diff
@@ -23,7 +30,7 @@ function BorderToggleButton(props: Props) {
     inFooter === true ? "footer-button" : "flexlayout__tab_toolbar_button"
   }`;
   const imageSrc = `/assets/images/icon-sidebar-${iconKind}-${side}-${
```
normanrz, line 32: @philippotto If possible, i would try to avoid the JS code knowing about the selected theme, because it will not update reactively when the theme is changed. In this case, the image could be implemented as a background image.

philippotto: ah, good idea with the background image :+1: `--is-dark-mode` felt clever, but using a pure bg image is obviously better. i will change it now

fm3:
Thanks for pinging me about the light theme :) I did notice some changes, only one of which I’d consider blocking:
light theme changes
 - hover color for welcome banner links is brighter (a little less legible but less impact than the recent font change)
 - error toast font color is red now (fine by me)
 - brush preview rendering has trippy new color scheme (Screenshot from Firefox. It’s better in chrome, but still definitely different from master. I think this one should be fixed) 
image
 - toggling the theme leads to “collapse”/“combination” of top menus (I just see a … instead of dashboard/administration/statistics…) as if there was too little space. It’s fixed on reload. 
image

philippotto:
Thanks for testing! I agree with your assessment of the blocking part.

> * brush preview rendering has trippy new color scheme (Screenshot from Firefox. It’s better in chrome, but still definitely different from master. I think this one should be fixed)
>   image

Very interesting. I cannot reproduce this. It might be related to my comment here: https://github.com/scalableminds/webknossos/pull/5407#discussion_r621254276
Could you check out this branch locally and set `alpha` to false just to check whether this fixes the rendering issue?

fm3:
> Could you check out this branch locally and set `alpha` to false just to check whether this fixes the rendering issue?

Indeed, removing the `alpha: true,` line restores the master behavior

normanrz:
Thanks for your testing and suggestions!

> * hover color for welcome banner links is brighter (a little less legible but less impact than the recent font change)

I made the hover color a bit darker

> * brush preview rendering has trippy new color scheme (Screenshot from Firefox. It’s better in chrome, but still definitely different from master. I think this one should be fixed)
>   image

I fixed this.

> * toggling the theme leads to “collapse”/“combination” of top menus (I just see a … instead of dashboard/administration/statistics…) as if there was too little space. It’s fixed on reload.
>   image

I couldn't reproduce this :-(

fm3:
Excellent news :)

Concerning the last point, this seems to work fine in Chrome, but I can consistently reproduce the described problem in Firefox Windows+Linux. I don’t know how much of a problem it is, since reloading fixes the issue and most users won’t be switching themes very often.

normanrz:
> Excellent news :)
> 
> Concerning the last point, this seems to work fine in Chrome, but I can consistently reproduce the described problem in Firefox Windows+Linux. I don’t know how much of a problem it is, since reloading fixes the issue and most users won’t be switching themes very often.

I could also reproduce it in Firefox and fixed it now :-)

fm3:
> I could also reproduce it in Firefox and fixed it now :-)
Works for me 🎉

philippotto (approved):
Excellent! Mergy-merge?

PR #5410, opened 2021-04-20
Improve Layouting

philippotto:
See https://discuss.webknossos.org/t/reworked-layout-system/1658/5

- https://___.webknossos.xyz

### Steps to test:
compare:
- image
- image

------
- [X] Ready for review


fm3 (approved):
LGTM. I think it makes sense that the dataset slider is slightly more to the right (even though this was marked as a problem in the original report), as the whole per-layer block is indented.

I see you removed my additional padding for the info tab. If you think this looks better, fair enough, but maybe the general padded-tab-content class could get one or more pixels padding more? I think it still looks kind of full, especially right next to the (naturally very busy) data viewport.

fm3:
Yes, sounds good :+1:

PR #5384, opened 2021-04-08
Panel content reorganization + Control/Tool rework

MichaelBuessemeyer:
### URL of deployed dev instance (used for testing):
- https://panelcontentreorganization.webknossos.xyz

### Steps to test:
- Look at the checklist below and check whether the panels/tabs have the content as described in the list.
- Please also check the (basic) functionality of the panel elements / settings.

- I had quite a lot of merging conflicts relating the dark mode pr / css. This needs to be tested, so nothing breaks because of this PR.

### Issues:
- fixes #5231

------
- [x] Updated [(unreleased) changelog](../blob/master/CHANGELOG.unreleased.md#unreleased)
- ~~[ ] Updated [(unreleased) migration guide](../blob/master/MIGRATIONS.unreleased.md#unreleased) if applicable~~
- ~~[ ] Updated [documentation](../blob/master/docs) if applicable~~
- ~~[ ] Adapted [wk-connect](https://github.com/scalableminds/webknossos-connect) if datastore API changes~~
- ~~[ ] Needs datastore update after deployment~~
- [x] Ready for review

PR #5422, opened 2021-04-23
[Base] New Controls

### LINKED ISSUE #4787: Rework skeleton and volume toolbar

philippotto:
- [x] Skeleton tool
  - [x] legacy mode should behave as before
    - [X] main controls
    - [ ] also support modifiers in non-legacy mode to merge trees etc (necessary? already available via context menu)
  - [x] left click/drag is "smart" and capable of
    - [x] create node
    - [x] select node
    - [x] pan
    - [x] move node
    - [ ] these actions should switch to the skeleton tool?
  - [x] right click opens context menu
- [x] Trace/Brush tool
  - [x] legacy mode should behave as before
  - [x] right click opens context menu
    - [x] add "select cell"
    - [x] add "copy cell id"
    - [ ] create new cell id? (not convinced, since that isn't context-specific)
    - [x] add fill cell (necessary, since we removed ctrl+shift+click as a shortcut for fill-cell)
    - [ ] optional
      - [ ] copy cell from previous slice
  - [x] support erasure
    - [X] add separate erase-tools (for brush and trace)
    - [x] avoid multiple erase-buttons
      - [X] only render the one which relates to the active tool
      - [ ] if the above isn't good enough, use dropdowns or something else ?
     - [x] CTRL+Shift + Leftclick should erase
- [x] double-check modifiers which can act on active tools
  - [x] pressing shift in "erase"-brush should not switch tools
  - [ ] ... ?
- [x] order context menu entries depending on active tool (in skeleton mode, skeleton options should appear first)
- [x] double-check unnecessary renderings
  - [x] toolbar should not rerender on mouse-move
  - [ ] ... ?
- [x] explain legacy-mode via tooltip in settings
- [x] adapt interactive tooltips in toolbar when modifiers are pressed
- [X] adapt shortcut hints in status bar to non-legacy mode
  - [x] adapt shortcut hints in status bar when modifiers are pressed
- [x] pull master in
- [x] migration strategy for legacy mode
  - [x] organization-wide setting?
  - [x] migrate recommended settings
  - [x] migrate existing user settings
- [ ] turn plane/oblique/flight switch into dropdown with icons?

- https://___.webknossos.xyz

### Steps to test:
- abc

### Issues:
- fixes #4787

------
- [x] Ready for review


philippotto:
@daniel-wer I think it would make sense if you could do a high-level review of this PR. I'll open another PR which pulls the master in separately, so that you can review a clean diff (I just had some trouble with squashing the way I wanted and think this is the safer route to take).

Code-wise I'm not really satisfied with how the tool handling is done, but I don't see a way to improve this without doing an even further refactoring. The tools work as follows:

- the active tool is stored in the `Store`
- depending on the active tool, the mouse handlers of the corresponding Tool class (see `tool_controls.js`) are delegated
- if a modifier is pressed, the current Tool class is responsible for invoking the action of another tool (so, the Draw tool will delegate to the pick-cell-tool if shift is pressed)
- for the toolbar and the status bar, a tool accessor is used which is able to map from an active tool and pressed modifiers to the "effective tool" (if I press shift with the draw tool, the toolbar/statusbar show the pick-cell-tool as active)
- the "effective tool" is used to get the control-hints in the status bar

It's a bit messy, but as I said: I don't see a good way around this currently.

frontend/javascripts/oxalis/model/accessors/tool_accessor.js
```diff
@@ -193,30 +192,30 @@ export function adaptActiveToolToShortcuts(
...
-    return activeTool;
-  }
-
-  if (activeTool === AnnotationToolEnum.SKELETON) {
-    // The "skeleton" tool is not changed right now (since actions such as moving a node
-    // don't have a dedicated tool). The only exception is "Alt" which switches to the move tool.
-    if (isAltPressed) {
-      return AnnotationToolEnum.MOVE;
+    // These tools do not have any modifier-related behavior currently (except for ALT
+    // which is already handled below)
```
daniel-wer, line 196: Hm, I don't see how ALT is being handled below for the `ERASE_BRUSH` and `ERASE_TRACE` tools. Isn't the ALT handling still being done in the else branch of this condition (so never hit for these two tools)?

daniel-wer (commented):
Lots of great fixes :+1:

frontend/javascripts/oxalis/model/accessors/tool_accessor.js
philippotto: It's a bit hidden: https://github.com/scalableminds/webknossos/pull/5422/files/ce5deb14bc2a7090152400ea8d0363d698b174e6..1f6249f3e1913c169bf5ce794565f410dc259cf5#diff-6bae9636ea043e9416c5c662d914cbb44ee20ef5eed61c99d2419472e981b072R222

The logic of this method is a bit convoluted. Ideally, this should probably be also delegated to (and unified with) the tool classes. However, it should work at least :see_no_evil:

philippotto:
I just merged this PR manually into https://github.com/scalableminds/webknossos/pull/5384 which is why I'm closing it here.

PR #5426, opened 2021-04-27
Register new demo instance users with mailchimp

### LINKED ISSUE #5424: Integrate wk with Mailchimp
## Detailed Description
Automatically add new users on webknossos.org to our Mailchimp account and stop sending the onboarding email for these users from wk.

## Context
We want to experiment with the welcome email (and also automated consecutive emails). In Mailchimp, it is much easier to iterate on the templates and the email flow than coding the HTML in wk. Mailchimp is also the tool that we want to use for our newsletters.

fm3:
Do not send welcome mail to new demo instance users, instead add them to mailchimp list and tag them either with `registered_as_admin` or `registered_as_user`

### Steps to test:
- set isDemoInstance = true, fill in mail.mailchimp block in config with valid values
- register new account, mailchimp list should contain new entry

### Issues:
- fixes #5424 

- [x] Ready for review


webknossos-datastore/app/com/scalableminds/webknossos/datastore/rpc/RPCRequest.scala
```diff
@@ -165,7 +165,7 @@ class RPCRequest(val id: Int, val url: String, wsClient: WSClient) extends FoxIm
     request
       .execute()
       .map { result =>
-        if (result.status == OK) {
+        if (Status.isSuccessful(result.status)) {
```
fm3, line 168: mailchimp sends 204 instead of 200, which was wrongly interpreted as an error

youri-k (approved):
LGTM :shipit:

Issue #3475, opened 2018-11-20
Refactor input mouse handling


philippotto:
At the moment, we are manually attaching to a dom node via a selector, which can change (e.g., when the golden layout adapter re-instantiates the layout) though.

daniel-wer:
This is partly resolved in #4031.
Components which still break because of this (but lower prio):
- HammerJS touch handling (as hammerjs doesn't seem to support event delegation, but we can probably imitate it)
- ThreeJS Trackball Controls, i.e. the right click rotation

philippotto:
See https://github.com/scalableminds/webknossos/pull/5429/files#r621310352 which again gave some attention to this issue. The new FlexLayout lazily renders the input-catchers (so, if the viewport is hidden by the layout it won't be rendered at all). Since we are waiting for these DOM elements, the current approach is a bit sub-optimal (perf-wise and also in general a bit messy).

PR #5429, opened 2021-04-27
Fix rendering if 3D viewport is hidden initially

### LINKED ISSUE #5430: Rendering is broken if 3D viewport is hidden initially
See #5429

philippotto:
- https://___.webknossos.xyz

### Steps to test:
- open a dataset or annotation
- re-arrange the viewports so that the 3d viewport is not visible
- wait a few seconds so the layout is persisted (this is not done via the save button)
- reload the page (previous layout should be visible)
- ensure that all viewports look good (despite the 3d viewport being hidden)
- use "reset layout" and test that all mouse bindings work as expected

### Issues:
- fixes #5430

------
- [X] Ready for review

### DIFF
--- a/frontend/javascripts/oxalis/controller/td_controller.js
+++ b/frontend/javascripts/oxalis/controller/td_controller.js
@@ -207,6 +207,9 @@ class TDController extends React.PureComponent<Props> {
   }
 
   setTargetAndFixPosition(position?: Vector3): void {
+    if (this.controls == null) {
+      return;
+    }
     position = position || getPosition(this.props.flycam);
     const nmPosition = voxelToNm(this.props.scale, position);
 
--- a/frontend/javascripts/oxalis/controller/td_controller.js
+++ b/frontend/javascripts/oxalis/controller/td_controller.js
@@ -70,7 +70,7 @@ function maybeGetActiveNodeFromProps(props: Props) {
 }
 
 class TDController extends React.PureComponent<Props> {
-  controls: typeof TrackballControls;
+  controls: ?typeof TrackballControls;
   mouseController: InputMouse;
   oldNmPos: Vector3;
   isStarted: boolean;
@@ -148,7 +148,12 @@ class TDController extends React.PureComponent<Props> {
     this.forceUpdate();
   }
 
-  updateControls = () => this.controls.update(true);
+  updateControls = () => {
+    if (!this.controls) {
+      return;
+    }
+    this.controls.update(true);
+  };
 
   getTDViewMouseControls(): Object {
     const skeletonControls =
@@ -207,14 +212,15 @@ class TDController extends React.PureComponent<Props> {
   }
 
   setTargetAndFixPosition(position?: Vector3): void {
-    if (this.controls == null) {
+    const { controls } = this;
+    if (controls == null) {
       return;
     }
     position = position || getPosition(this.props.flycam);
     const nmPosition = voxelToNm(this.props.scale, position);
 
-    this.controls.target.set(...nmPosition);
-    this.controls.update();
+    controls.target.set(...nmPosition);
+    controls.update();
 
     // The following code is a dirty hack. If someone figures out
     // how the trackball control's target can be set without affecting
@@ -263,10 +269,6 @@ class TDController extends React.PureComponent<Props> {
   }
 
   render() {
-    if (!this.controls) {
-      return null;
-    }
-
     return (
       <CameraController
         cameras={this.props.cameras}
--- a/frontend/javascripts/oxalis/controller/viewmodes/plane_controller.js
+++ b/frontend/javascripts/oxalis/controller/viewmodes/plane_controller.js
@@ -134,21 +134,18 @@ class PlaneController extends React.PureComponent<Props> {
     // catchers (only necessary for HammerJS). We should refactor the
     // InputMouse handling so that this is not necessary anymore.
     // See: https://github.com/scalableminds/webknossos/issues/3475
-    const tdId = `inputcatcher_${OrthoViews.TDView}`;
-    Utils.waitForElementWithId(tdId).then(() => {
-      OrthoViewValuesWithoutTDView.forEach(id => {
-        const inputcatcherId = `inputcatcher_${OrthoViews[id]}`;
-        Utils.waitForElementWithId(inputcatcherId).then(el => {
-          if (!document.body.contains(el)) {
-            console.error("el is not attached anymore");
-          }
-          this.input.mouseControllers[id] = new InputMouse(
-            inputcatcherId,
-            this.getPlaneMouseControls(id),
-            id,
-            true,
-          );
-        });
+    OrthoViewValuesWithoutTDView.forEach(id => {
+      const inputcatcherId = `inputcatcher_${OrthoViews[id]}`;
+      Utils.waitForElementWithId(inputcatcherId).then(el => {
+        if (!document.body.contains(el)) {
+          console.error("el is not attached anymore");
+        }
+        this.input.mouseControllers[id] = new InputMouse(
+          inputcatcherId,
+          this.getPlaneMouseControls(id),
+          id,
+          true,
+        );
       });
     });
   }
--- a/frontend/javascripts/oxalis/view/input_catcher.js
+++ b/frontend/javascripts/oxalis/view/input_catcher.js
@@ -8,7 +8,13 @@ import Scalebar from "oxalis/view/scalebar";
 import ViewportStatusIndicator from "oxalis/view/viewport_status_indicator";
 import Store from "oxalis/store";
 import makeRectRelativeToCanvas from "oxalis/view/layouting/layout_canvas_adapter";
-import { defaultViewportRect } from "oxalis/default_state";
+
+const emptyViewportRect = {
+  top: 0,
+  left: 0,
+  width: 0,
+  height: 0,
+};
 
 type Props = {
   viewportID: Viewport,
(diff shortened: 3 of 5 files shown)


frontend/javascripts/oxalis/controller/viewmodes/plane_controller.js
```diff
@@ -134,21 +134,18 @@ class PlaneController extends React.PureComponent<Props> {
...
-            console.error("el is not attached anymore");
-          }
-          this.input.mouseControllers[id] = new InputMouse(
-            inputcatcherId,
-            this.getPlaneMouseControls(id),
-            id,
-            true,
-          );
-        });
+    OrthoViewValuesWithoutTDView.forEach(id => {
```
philippotto, line 137: Waiting for the 3d viewport did not succeed here, if it's hidden initially. In that case, the mouse handlers were not set up. Since, there is a `waitForElementWithId` two lines below for **each** viewport, I'd hope that this change doesn't break anything.

frontend/javascripts/oxalis/view/input_catcher.js
```diff
@@ -48,7 +49,13 @@ function adaptInputCatcher(inputCatcherDOM: HTMLElement, makeQuadratic: boolean)
 const renderedInputCatchers = new Map();
 
 export function recalculateInputCatcherSizes() {
-  const viewportRects = {};
+  const viewportRects: Object = {
```
philippotto, line 52: This is a bit of a dirty workaround. I'm not sure yet which component cannot deal with the TD viewport rect, but this at least fixes the problem. @MichaelBuessemeyer Do you know how this could be turned into a proper fix?

frontend/javascripts/oxalis/controller/viewmodes/plane_controller.js
MichaelBuessemeyer: I think this won't be a problem, as the `waitForElementWithId` works with promises. The onliest thing I see is that if one of the viewports is never rendered, that there will be endless calls of `window.requestAnimationFrame` as long as wk is opened in a browser tab. 
I think this is not optimal. Maybe we can add a timeout to this function and retrigger adding the mouse controllers for all inputcatchers that were newly rendered when the layout is changed.
But this would be a thing for another PR.

Note:
```
export function waitForElementWithId(elementId: string): Promise<*> {
  const tryToResolve = resolve => {
    const el = document.getElementById(elementId);
    if (el) {
      resolve(el);
    } else {
      window.requestAnimationFrame(() => tryToResolve(resolve));
    }
  };
  return new Promise(tryToResolve);
}
```

MichaelBuessemeyer (changes requested):
The current workaround does not work for me when I stack all ortho views together and have the XY-viewport open.

Lets investigate why :mag:

MichaelBuessemeyer:
My :mag results:
The root cause is that the cameras for the planes are not updated until the 3d viewport is rendered. 
This bug only appeared now, because afaik `gl` always rendered each viewport and not only the viewports that were already visible like `FlexLayout` does it.

Then why are the viewport cameras updated once the 3d viewport is rendered? 
This is because only the `TdController` uses a `CameraController`. A `CameraController` automatically updates the cameras of all viewports once they are changed in the store. Therefore upon rendering the 3D viewport, a `TdController` is rendered and thus a `CameraController` that updates the viewports cameras :tada:.

How to solve this (imo):
I think this is a software architectural problem, that all the viewports rely on the `CameraController` that is rendered by the 3d viewport for updating their cameras. Imo each viewport should have its own "`CameraController`" that is responsible for updating its own camera and not the camera of the other viewports.

It looks like implementing this change is quite some work, but think it is necessary and that a quick fix would not be sufficient / too dirty and might result in related bugs later on.
@philippotto Do you have enough time for the change I suggested?

Note: The changes in `input_catcher` and `default_state` are unnecessary.

MichaelBuessemeyer:
You can verify my thesis by commenting out the loop starting at [this line](https://github.com/scalableminds/webknossos/blob/85e73b82fdb129861dd674aa71c1966775c64015/frontend/javascripts/oxalis/controller/camera_controller.js#L117) and 
- then opening a tracing view with only the XY viewport active, 
- then opening the 3d viewport
- and then opening the XY viewport again should no longer fix the issue.

If you uncomment the lines again, then the 3rd step should work and data should be visible in the xy viewport

frontend/javascripts/oxalis/controller/viewmodes/plane_controller.js
philippotto: Good point! I cross-referenced this comment in the corresponding issue: https://github.com/scalableminds/webknossos/issues/3475

I agree that this is out-of-scope for this PR.

frontend/javascripts/oxalis/controller/td_controller.js
```diff
@@ -260,10 +269,6 @@ class TDController extends React.PureComponent<Props> {
   }
 
   render() {
-    if (!this.controls) {
-      return null;
-    }
```
philippotto, line 265: By postponing this null-check to the exact points where it's necessary (i.e., in `setTargetAndFixPosition`), we can eagerly instantiate the CameraController which should fix the issue.

frontend/javascripts/oxalis/view/input_catcher.js
philippotto: Instead of falling back to the "default viewport rect", I now changed it to the "empty viewport rect". It's still not super elegant, but:

- I find it better to have a key for each viewport (even if it contains a zero-area-rect). I think, most of the other code assumes that they can look up each plane's viewport rect.
- it fixes another issue I had. Merely moving the `if (!this.controls)`-check in the `td_controller`, still caused a small rendering bug for me on initial load. See screenshot:
image

philippotto:
> My  results:
> The root cause is that the cameras for the planes are not updated until the 3d viewport is rendered.
> This bug only appeared now, because afaik `gl` always rendered each viewport and not only the viewports that were already visible like `FlexLayout` does it.
> 
> Then why are the viewport cameras updated once the 3d viewport is rendered?
> This is because only the `TdController` uses a `CameraController`. A `CameraController` automatically updates the cameras of all viewports once they are changed in the store. See [here](https://github.com/scalableminds/webknossos/blob/85e73b82fdb129861dd674aa71c1966775c64015/frontend/javascripts/oxalis/controller/camera_controller.js#L117). Therefore upon rendering the 3D viewport, a `TdController` is rendered and thus a `CameraController` that updates the viewports cameras .
> Note: The `TdController` is only rendered when the 3d viewport exists due to [this guard](https://github.com/scalableminds/webknossos/blob/85e73b82fdb129861dd674aa71c1966775c64015/frontend/javascripts/oxalis/controller/td_controller.js#L263-265).

Thanks for the detailed investigation and write-up :+1: :+1: From your last sentence, I could derive a relatively simple solution. I simply removed the guard and limited its scope so that it doesn't affect the CameraController.

> How to solve this (imo):
> I think this is a software architectural problem, that all the viewports rely on the `CameraController` that is rendered by the 3d viewport for updating their cameras. Imo each viewport should have its own "`CameraController`" that is responsible for updating its own camera and not the camera of the other viewports.
> 
> It looks like implementing this change is quite some work, but think it is necessary and that a quick fix would not be sufficient / too dirty and might result in related bugs later on.
> @philippotto Do you have enough time for the change I suggested?

Yes, I agree, that an architectural change would be even better. However, as a quick fix, I hope that my above workaround is adequate.

> Note: The changes in `input_catcher` and `default_state` are unnecessary.

See my other comments. The changes in `input_catcher` seem to be necessary (and I also think that they make the viewport-rect state more consistent).

frontend/javascripts/oxalis/view/input_catcher.js
MichaelBuessemeyer: > I find it better to have a key for each viewport (even if it contains a zero-area-rect). I think, most of the other code assumes that they can look up each plane's viewport rect.

I do not think that this should be an issue as the default state has a viewport rect for each viewport and the `setInputCatcherRects`-Action only updates the rects of the passed viewports. Therefore there should always be a rect for each viewport.

>it fixes another issue I had. Merely moving the if (!this.controls)-check in the td_controller, still caused a small rendering bug for me on initial load. See screenshot:

:+1: 

Additionally, I think having "empty" viewports as default should also solve @fm3 bug as my guess is that this is caused by having a non-empty viewport for a viewport that is not rendered in the dom.
image

philippotto: > I do not think that this should be an issue as the default state has a viewport rect for each viewport and the setInputCatcherRects-Action only updates the rects of the passed viewports. Therefore there should always be a rect for each viewport.

Ah, that explains why it didn't crash. However, as you indicated, not updating not-rendered viewports leads to stale values. So, I think my changes should stay as is.

MichaelBuessemeyer: >So, I think my changes should stay as is.

Indeed think so too. (But forgot to make this explicit :grin:)

philippotto: Then only an explicit GitHub "Accept" is missing, I think :)

MichaelBuessemeyer (approved):
I was finally able to test the fix and it works :rocket: :roll_of_paper:

MichaelBuessemeyer:
> Yes, I agree, that an architectural change would be even better. However, as a quick fix, I hope that my above workaround is adequate.

I just created an issue for this: #5436

PR #5435, opened 2021-04-29
Make sure you can only see your own long-running job list

fm3:
The jobsController didn’t override findAll, and the generic one from the superclass does not use the access query.

### Steps to test:
- set up long running jobs, run some
- set up second account
- each should see only their own long-running jobs in the list

------
- [x] Ready for review


youri-k (approved):
LGTM :shipit:

PR #5449, opened 2021-05-02
Increase initialTeams timeout from 5m to 1h

fm3:
This timeout ensures that no initialTeams request is accepted after the upload procedure is completed.

However, it was too strict. In case of a busy server, the time between datasource being reported and its initialTeams request being served was larger than 5 minutes. I think it is safe enough to increase the timeout. I also added more verbose error output so that in the future we can easily compare the timestamps when debugging.

### Steps to test:
- Upload dataset with initial teams.
- in finishUpload, introduce a few minutes sleep between `uploadService.finishUpload` and `webKnossosServer.reportUpload`
- reportUpload (including initialTeams) should still run through

------
- [x] Ready for review


youri-k (approved):
LGTM :shipit:

PR #5208, opened 2021-02-24
Clean up config

fm3:
Cleaning up application.conf and its usages, as well as standalone-datastore.conf and standalone-tracingstore.conf.

### Introduced conventions
 - prefer `x.enabled` for feature flags over `x.active`
 - prefer FiniteDuration or Duration data type for durations (no `xInSeconds`: Int anymore)
 - prefer `uri` over `url` except where really not applicable
 - `x.cache.details` for cache section of different blocks

### TODO
- [x] gather feedback on naming/grouping
- [x] print warnings if old keys are still present
- [x] re-introduce http.port (play does not automatically listen to port supplied in http.uri)
- [x] run module health checks via http://localhost:${http.port} again
- [x] should we also tackle the features block? no, follow-up issue #5423
- [x] include relevant subset of the table below to migration guide
- [x] double check that the old fields being present does not do harm
- [x] double check that the new fields being present does not do harm to old wK code

### Steps to test:
- start locally, should still run
- CI should pass
- test that airbrake still runs
- standalone tracingstore should still run
- standalone datastore should still run

### Config key changes
old key | new key | notes
--------|--------|-------
`http.address` | removed | used by play, default is 0.0.0.0, you can still overwrite it if necessary
`actor.defaultTimeout` | removed | was already unused
`js.defaultTimeout` | removed | was already unused
`akka.loggers` | removed | was already unused
`application.name` | removed | was already unused
`application.branch` | removed | was already unused
`application.version` | removed | was already unused
`application.title` | `webKnossos.tabTitle` |
`application.insertInitialData` | `webKnossos.sampleOrganization.enabled` |
`application.insertLocalConnectDatastore` | removed | feature removed, insert manually instead
`application.authentication.defaultuser.email` | `webKnossos.sampleOrganization.user.email` |
`application.authentication.defaultUser.password` | `webKnossos.sampleOrganization.user.password` |
`application.authentication.defaultUser.token` | `webKnossos.sampleOrganization.user.token` |
`application.authentication.defaultUser.isSuperUser` | `webKnossos.sampleOrganization.user.isSuperUser` |
`application.authentication.ssoKey` | `webKnossos.user.ssoKey` |
`application.authentication.inviteExpiry` | `webKnossos.user.inviteExpiry` |
`webKnossos.user.time.tracingPauseInSeconds` | `webKnossos.user.time.tracingPause` | **type changed from Int to FiniteDuration, add ` seconds`**
`webKnossos.query.maxResults` | removed | was already unused
[...]

PR #5451, opened 2021-05-03
Use disabled list in wk module detection

fm3:
Our kube setup does not disable play modules by removing them from the “enabled” list but from adding them to the “disabled” list. This must be reflected in the detection used in our health checks.

Hotfix following up #5208 

- [x] Ready for review


philippotto (approved):
LGTM, but I obviously don't have any experience with this.

PR #5457, opened 2021-05-05
Fix two bugs related to transferring task instances

### LINKED ISSUE #5442: Task search shows duplicates if a user has multiple instances of single task
## Context
Administration -> Tasks

## Expected Behavior
Tasks should not be duplicated.

## Current Behavior
Task instances for the same datasets are duplicated every time certain column headers are clicked.

## Steps to Reproduce the bug

1. Open the tasks view and search for a user that has more than one instance of task for a certain dataset (e.g. 2 instances).
2. Click on the ID/Dataset/Creation Date column header to sort. Now the dataset is listed 3 times.
3. Keep clicking the column header and every time there will be a new duplicated entry.

Sorting by Project or Type didn't cause this behavior in my hands.

## Your Environment for bug

* Browser name and version: Chrome 90
* Operating System and version: Windows 10
* Version of WebKnossos (Release or Commit): 21.05.0

fm3:
- Fixed a bug where the task search showed duplicates if a user had multiple instances of a task (as made possible by the transfer functionality) – this never worked before, it was just such a rare scenario that we never noticed this.

- Fixed a bug where showing active users of a project, and transferring their tasks was broken. – this was a regression introduced in #5334 

- https://fixtransfertasks.webknossos.xyz

### Steps to test:
- Create task with multiple instances, request an instance for yourself
- Sign up second user, assign matching experience, get another instance of said task as this user
- as the admin, click “show active users” in project list, transfer tasks to yourself. you have two open instances now
- in task search, search by user (yourself), result should not be duplicated

### Issues:
- fixes #5442

------
- [x] Ready for review


philippotto (approved):
LGTM! :+1: Maybe have a third look at the SQL change yourself, since I'm not too confident that I could spot an error there. The usage of `in` rings a performance-critical bell for me, but that's probably an unfounded gut reaction..

fm3:
Thanks! Yes, I am fairly sure this `in` is harmless. The inner query is the heavy one that performs join and `where` on a large table, and it remains essentially unchanged, except that the limit is no longer applied there. But the limit already came _after_ the joins before too, so this should not change the performance. The outer select is harmless. I believe that the heaviest part remains serializing the result from sql to scala.
If my assumption is wrong we should notice this fairly quickly in production and we can then think about another way to solve this.

philippotto:
Sounds great!

fm3:
Apparently the order in which tasks are listed has changed, though, so I had to change the snapshots. Do you think this is problematic? There is no explicit order by in the query, just as before, but apparently the behavior of postgres is different with the nested query

philippotto:
> Do you think this is problematic?

I think, the only negative impact of the undefined ordering is that we have to refresh the snapshots every now and then. For now, this seems to be alright I'd say :)

PR #5458, opened 2021-05-05
Fix histogram sampling for tiny datasets

### LINKED ISSUE #5444: Histogram request fails for tiny datasets
If the bounding box of a dataset is tiny (e.g. 50 × 50 × 30 voxel³) the histogram request fails.
The position sampling runs into breaking criterion `spaceBetweenWidth < DataLayer.bucketLength` already in the first iteration. How can we create a sampling strategy that also returns valid data for arbitrarily small datasets?

fm3:
The position sampling for histogram and find data roughly works like a menger sponge, with a stopping criterion when the distance between the sampling points becomes too close. For tiny datasets, the first iteration already runs into the stopping criterion.

The fix is to always add the 32vx³ bucket at 0,0,0 to the position list. This way, the data from this corner (which tends to exist for tiny datasets) is included.

- https://tinyhistogram.webknossos.xyz

### Steps to test:
- open a very small dataset (tested with 50 × 50 × 30 voxel³)
- sidebar should show valid histogram

### Issues:
- fixes #5444 

------
- [x] Ready for review


youri-k (approved):
Nice idea :tada:

fm3:
> Is this relative to the datasource's bounding box?

It will be in the next version 🙈 😆  good idea

PR #5461, opened 2021-05-05
Highlight active tree and active node in comment tab

normanrz:
Highlights the active tree and active node in the comment tab with a background color similar to the active tree in the tree tab.

Screenshot 2021-05-05 at 15 53 06
Screenshot 2021-05-05 at 15 53 12

------
- [x] Ready for review


daniel-wer (approved):
LGTM :) As far as I can see, this shouldn't interfere with comment links (#<node_id> to link to another node) as those links have the same blue color as the node id in front of comments which is also still visible :+1:

philippotto:
Looks good! However, the background color makes the suboptimal vertical alignment more prominent. Maybe this is easy to fix. If not, should be fine for now, too.

normanrz:
Next round with adjusted centering.

Screenshot 2021-05-18 at 18 27 52
Screenshot 2021-05-18 at 18 27 59

I also increased the height of each comment entry to 24, which is the same as in the tree tab.
And, I removed the bold selected comment, because that is redundant with the colored background. wdyt?

philippotto (approved):
Great :rocket:

PR #5473, opened 2021-05-06
More logging for long-running jobs

fm3:
- Send slack notification for successful jobs (additionally to the previously existing notification for failed ones)
- Send slack notification + log to console if the run job request itself fails

### Steps to test:
- set `jobsEnabled = true` and `slackNotifications.verboseLoggingEnabled = true`
- upload dataset that needs conversion (but do not run webknossos-worker), slack notification for the failed request should be logged to backend logging
- do run webknossos-worker upload valid and invalid dataset (that needs conversion), both success and failure should be logged as slack notification to the backend logging

- [x] Ready for review

### DIFF
--- a/conf/application.conf
+++ b/conf/application.conf
@@ -153,7 +153,6 @@ mail {
     pass = ""
   }
   defaultSender = "webKnossos <no-reply@webknossos.org>"
-  demoSender = ""
   reply = "webKnossos <no-reply@webknossos.org>"
   mailchimp {
     host = ""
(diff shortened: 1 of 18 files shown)


conf/application.conf
```diff
@@ -153,7 +153,6 @@ mail {
     pass = ""
   }
   defaultSender = "webKnossos <no-reply@webknossos.org>"
-  demoSender = ""
```
fm3, line 156: small follow-up cleanup for #5426 slipped in here

PR #5474, opened 2021-05-06
Apply mapping when exporting to tiff

### LINKED ISSUE #5415: Export volume annotation should take active mapping into account
Exporting a volume annotation while a mapping (agglomerate or mapping json) is active should produce the annotation with the applied mapping. A notable example/edge case for this is the merger-mode which produces a mapping on-the-fly (not stored on disk).

youri-k:
- https://___.webknossos.xyz

### Steps to test:
- export wkw while mapping enabled
- exported data should look like mapped data

### Issues:
- fixes #5415 

------
- [x] Ready for review


fm3:
Backend LGTM :) @philippotto could you have a quick look at the front-end changes?

Also note that the updated worker needs to be deployed before this upgrade is done on an instance with `jobsEnabled = true`

philippotto (changes requested):
Great work! Just see my few nitpicks and I think it would make sense to also cover the merger-mode-case (for now, it's enough to simply state that this is not supported yet).

frontend/javascripts/oxalis/view/settings/export_bounding_box_modal.js
```diff
@@ -68,6 +88,10 @@ const ExportBoundingBoxModal = ({ destroy, dataset, boundingBox, tracing }: Prop
         annotationType: null,
         tracingVersion: null,
         hasMag1: hasMag1(layer),
+        hideUnmappedIds: existsActivePersistentMapping ? hideUnmappedIds : null,
+        mappingName: existsActivePersistentMapping ? mappingName : null,
+        mappingType: existsActivePersistentMapping ? mappingType : null,
+        isColorLayer: layer.category === "color",
```
philippotto, line 94: What happens when this code runs for a color layer while a persistent mapping is active? This will pass the mapping parameters to the back-end which is confusing, I'd assume? Maybe

```suggestion
        // ... somewhere above this block
        const isColorLayer = layer.category === "color";
        // ...
        hideUnmappedIds: !isColorLayer && existsActivePersistentMapping ? hideUnmappedIds : null,
        mappingName: !isColorLayer && existsActivePersistentMapping ? mappingName : null,
        mappingType: !isColorLayer && existsActivePersistentMapping ? mappingType : null,
        isColorLayer
```

philippotto (approved):
Great! But see my last nitpick :)

philippotto:
Great!

PR #5486, opened 2021-05-11
Preserve row selection in users table

### LINKED ISSUE #5471: In user list, search changes selection in unexpected ways
Quoting from [Discuss post](https://discuss.webknossos.org/t/selecting-users-reset/1665): “sometimes user selection is resetted. It happens when I enter something in the search field.” Also see linked video

philippotto:
- https://___.webknossos.xyz

### Steps to test:
- set up wk instance with at least two users
- open users table
- select a user
- use the search so that the unselected user is the only remaining element
- select that remaining user
- the top left should show that 2 users are selected (before the bug, only one was selected)

### Issues:
- fixes #5471

------
- [X] Ready for review


frontend/javascripts/admin/user/user_list_view.js
```diff
@@ -269,6 +269,7 @@ class UserListView extends React.PureComponent<PropsWithRouter, State> {
   render() {
     const hasRowsSelected = this.state.selectedUserIds.length > 0;
     const rowSelection = {
+      preserveSelectedRowKeys: true,
```
philippotto, line 272: see https://ant.design/components/table/#rowSelection

daniel-wer (approved):
Nice find!

As a followup we should check whether that is relevant for any other tables as well (if we actually have more tables with checkboxes :thinking: ).

PR #5487, opened 2021-05-12
Use bbox topleft instead of 0,0,0 in histogram position sampling

fm3:
@philippotto rightly pointed out that my previous fix  #5458 works only for the (somewhat standard, but not general) case of the dataset starting at 0,0,0. This PR changes the used position to the dataset bounding box topleft corner.

- [x] Ready for review


philippotto (approved):
looks legit :smiley:

PR #5488, opened 2021-05-12
Include superuser hint in slack job notifications

fm3:
Include isSuperUser info in slack job notifications so that we know when we were just testing/debugging the feature ourselves.

- [x] Ready for review


youri-k (approved):
LGTM :shipit:

PR #5489, opened 2021-05-12
Revert "Pass token to wkw conversion job"

fm3:
Reverts scalableminds/webknossos#5485

unfortunately I couldn’t get the new worker version that uses this argument to run, so let’s try this again later.


fm3:
Got it to work in the end, no need to revert this :)

Issue #5296, opened 2021-03-18
Use frontend analytics to track interesting events


philippotto:
## Detailed Description
Events to track (to be extended):
- [x] Open sharing modal (annotation)
- [x] Open dataset settings
- [x] Open upload view
- [x] Add files to upload view
- [x] Customize layout (i.e. move panels, resize panels)
- [ ] Store custom layout
- [x] Load custom layout

## Context
We have analytics on the backend, but we cannot track everything from the backend. See `sendAnalyticsEvent`.

MichaelBuessemeyer:
Closed via #5490

MichaelBuessemeyer:
The reason why not tracking event was implemented for "Store custom layouts" can be found in #5490:

> ... adds the requested analytics tracking of #5445 except for "Store custom layout" as it does not make sense in my opinion as this is tightly coupled to "Customize layout (i.e. move panels, resize panels)" when the auto-saving layout is turned on (which is the default).

PR #5490, opened 2021-05-14
Improve error message & more frontend analytics

### LINKED ISSUE #5445: Clicking "View" on a project which has no tracings produces an error
When creating a task and directly clicking "view" for it (i.e., no tracings were created for that task), the following error is shown:

image

There should be a better explanation of what's happening. Something like: No annotation was yet finished for this task. Please ensure that the task has... (assigned hiwis?)

MichaelBuessemeyer:
This PR adds an additional error Toast for #5296 to better explain the error to the user and adds the requested analytics tracking of # 5445 except for "Store custom layout" as it does not make sense in my opinion as this is tightly coupled to "Customize layout (i.e. move panels, resize panels)" when the auto-saving layout is turned on (which is the default).

If you think tracking this event still makes sense just notify me :smile: 

- https://___.webknossos.xyz

### Steps to test:
- Create a new project (`http://<>/projects/create`) 
- Then directly click on view in the project table.
- A tracing should open up with two errors:
    1. Annotation couldn’t be found
    1. It looks like this project does not have a single task completed. Make sure that at least one task of this project is finished to view it. 
I added the second error, to tell the user why this step failed and how to fix this error.

### Issues:
- fixes #5445, #5296

------
- [x] Ready for review


frontend/javascripts/dashboard/advanced_dataset/dataset_action_view.js
```diff
@@ -237,6 +242,9 @@ class DatasetActionView extends React.PureComponent<Props, State> {
                   to={`/datasets/${dataset.owningOrganization}/${dataset.name}/edit`}
                   title="Open Dataset Settings"
                   disabled={isReloading}
+                  onClick={() =>
+                    sendAnalyticsEvent("open_dataset_settings", { datasetName: dataset.name })
```
philippotto, line 246: I think, it would be better to trigger this event within `componentDidMount` of the settings view. That way, we also catch this event, when another settings link is used (e.g., within the dataset-info-tab of the tracing view).

frontend/javascripts/admin/dataset/dataset_upload_view.js
```diff
@@ -349,11 +354,11 @@ class DatasetUploadView extends React.Component<PropsWithFormAndRouter, State> {
     if (files.length === 0) {
       return;
     }
-
     let needsConversion = true;
     for (const file of files) {
       const filenameParts = file.name.split(".");
       const fileExtension = filenameParts[filenameParts.length - 1].toLowerCase();
+      sendAnalyticsEvent("add_files_to_upload", { fileExtension });
```
philippotto, line 361: If I drop 100 files, this event will be triggered 100 times. Can we group this by fileExtension? Then, the file count could also be transmitted. E.g.

```js
for (const fileExtension of uniqueFileExtensions) {
  const fileCount = amount of files with that extension
  sendAnalyticsEvent("add_files_to_upload", { fileExtension, fileCount });
}
```

philippotto (changes requested):
Nice! Already looking very good.

I had one thought about page visits: Maybe we can track this on a lower level via the react router. So, each page visit/change is tracked. What do you think?

frontend/javascripts/admin/dataset/dataset_upload_view.js
MichaelBuessemeyer: Good catch :+1:

I just applied your feedback :)

philippotto (approved):
:man_dancing:

PR #5491, opened 2021-05-15
Add all csv info to displayed table on successful task creation

### LINKED ISSUE #5359: Unify task creation CSV with displayed table
When creating tasks successfully, a list of the created tasks is shown, and a csv download is offered. The two should have the same content.

Also, the project id / name should be added to the csv filename (if unique?)

comapre https://discuss.webknossos.org/t/show-task-ids-for-created-tasks-miss-filename/1621/4

MichaelBuessemeyer:
This PR unifies the information that is displayed in the table that is shown upon successful task creating with the information of the CSV.

Imo this change does not make so much sense. Maybe I misunderstood something but to me, the display information in the modal is far too much to be easily understood/is not clearly arranged. Additionally, the file name might get too huge if all IDs (like suggested) are added to the file name. Thus I limited this to a maximum of 3 task ids.

I think some general input on my points is needed before we start a code review :)

- https://___.webknossos.xyz

### Steps to test:
- Got to the admin->task view
- Open the task creation
- Create a task and upon success see new content in the modal.
- Then download the CSV and check the file name.

### Issues:
- fixes #5359

------
- [x] Ready for review

### DIFF
--- a/frontend/javascripts/admin/task/task_create_form_view.js
+++ b/frontend/javascripts/admin/task/task_create_form_view.js
@@ -79,11 +79,6 @@ type State = {
   isFetchingData: boolean,
 };
 
-export function taskToShortText(task: APITask) {
-  const { id, creationInfo, editPosition } = task;
-  return `${id},${creationInfo || "null"},(${editPosition.join(",")})`;
-}
-
 export function taskToText(task: APITask) {
   const {
     id,
@@ -113,6 +108,12 @@ export function taskToText(task: APITask) {
   return taskAsString;
 }
 
+function tasksToCSVString(tasks: Array<APITask>) {
+  const allTasksAsStrings = tasks.map(task => taskToText(task)).join("\n");
+  const csv = [TASK_CSV_HEADER, allTasksAsStrings].join("\n");
+  return csv;
+}
+
 export function downloadTasksAsCSV(tasks: Array<APITask>) {
   if (tasks.length < 0) {
     return;
@@ -122,9 +123,12 @@ export function downloadTasksAsCSV(tasks: Array<APITask>) {
   const currentDateAsString = formatDateInLocalTimeZone(lastCreationTime);
   const allTeamNames = _.uniq(tasks.map(task => task.team));
   const teamName = allTeamNames.length > 1 ? "multiple_teams" : allTeamNames[0];
-  const allTasksAsStrings = tasks.map(task => taskToText(task)).join("\n");
-  const csv = [TASK_CSV_HEADER, allTasksAsStrings].join("\n");
-  const filename = `${teamName}-${maybeTaskPlural}-${currentDateAsString}.csv`;
+  const allProjectNames = _.uniq(tasks.map(task => task.projectName));
+  const projectName = allProjectNames.length > 1 ? "multiple_projects" : allProjectNames[0];
+  const taskIds = _.uniq(tasks.map(task => task.id));
+  const taskIdsString = taskIds.length < 4 ? taskIds.join("_") : "multiple_ids";
+  const csv = tasksToCSVString(tasks);
+  const filename = `${teamName}-${maybeTaskPlural}-${taskIdsString}-${projectName}-${currentDateAsString}.csv`;
   const blob = new Blob([csv], { type: "text/plain;charset=utf-8" });
   saveAs(blob, filename);
 }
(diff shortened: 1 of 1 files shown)


MichaelBuessemeyer:
@fm3 As you created the issue for this: Do you have an opinion about my thought about whether the changes actually make sense? Or do you have some suggestions to improve these changes?

fm3:
Thanks for looking into this! You are quite right, these changes create somewhat weird behavior. I wrote a message to the original requesting user asking to clarify, let’s see what they reply :) Should have done that immediatly when writing the issue, I’m sorry if this created redundant work!

daniel-wer:
The expected changes have been clarified, see https://discuss.webknossos.org/t/show-task-ids-for-created-tasks-miss-filename/1621/6.
Please request a review once you incorporated those :)

fm3:
tbh I’d slightly deviate from the `IDs_`-prefix suggested there, and go with the more general `tasks_ids_<projectnames>_<date>.csv` or something like that.

I also don’t like hiding the short success list, as I still believe it to be useful when creating few tasks. But maybe we can make the CSV download button more prominent even if the short list is there. And maybe the list can be styled differently to more obviously be something different than the CSV. That way, users who want to use the CSV workflow can always do that, and users who like instant feedback on (few) created tasks can still get it. What do you think?

daniel-wer:
> tbh I’d slightly deviate from the IDs_-prefix suggested there, and go with the more general tasks_ids_<projectnames>_<date>.csv or something like that.

Sounds fair to me.

> I also don’t like hiding the short success list, as I still believe it to be useful when creating few tasks. But maybe we can make the CSV download button more prominent even if the short list is there. And maybe the list can be styled differently to more obviously be something different than the CSV. That way, users who want to use the CSV workflow can always do that, and users who like instant feedback on (few) created tasks can still get it. What do you think?

It could be hidden behind a "Details" toggle, like this to de-clutter the popup by default?
details-toggle

Also, from the post it sounded like the main issue was that the CSV download button was only visible if a lot of tasks were created. So that should be changed to be always visible.

fm3:
> It could be hidden behind a "Details" toggle, like this to de-clutter the popup by default?

Yes, looks nice :) but that might not even be necessary if the real problem is the accessability of the CSV download button.

> Also, from the post it sounded like the main issue was that the CSV download button was only visible if a lot of tasks were created. So that should be changed to be always visible.

I agree. I think it is always visible already, but should be moved above the list, so the user does not have to scroll to get there.

MichaelBuessemeyer:
Thank you two for already discussing the issue :+1:.
I'll try to implement your ideas :)

MichaelBuessemeyer:
I just pushed the following changes:
- I updated the csv file name according to @fm3 suggestion and removed the spaces in the date formatting.
- I made the Download CSV button more prominent.
- Added a warning when creating tasks for multiple projects at once

frontend/javascripts/admin/task/task_create_form_view.js
```diff
@@ -117,14 +117,13 @@ export function downloadTasksAsCSV(tasks: Array<APITask>) {
   if (tasks.length < 0) {
     return;
   }
-  const maybeTaskPlural = tasks.length > 2 ? "tasks" : "task";
+  const maybeTaskPlural = tasks.length > 2 ? "task_ids" : "task_id";
```
daniel-wer, line 120: Shouldn't this be

```suggestion
  const maybeTaskPlural = tasks.length > 1 ? "task_ids" : "task_id";
```

daniel-wer (commented):
Very nice, works well :+1: 

How would I go about testing the "failed" path, so that some of the tasks fail during creation? If I change the dataset name, taskTypeId, or projectName I only end up with error toasts, but not with the popup and "the creation of x tasks failed".

Also, I noticed that when I try to create a task for a project that is not existing, I end up with an unhelpful error toast `Couldn't find the requested resource.` without further details. @fm3 Would it be easy to improve that? Otherwise, I'll create a follow up issue :)

fm3:
@daniel-wer I’ll have a look at the too-generic message, but I’d still track it in a separate issue.

To add a failed task to the report, you can modify line 426 in TaskCreationService.scala to `result <- TaskCreationResult.fromTaskJsFoxes(Fox.failure("Nope") :: taskJsons, warnings)` (prepends a failure to the result list)

frontend/javascripts/admin/task/task_create_form_view.js
MichaelBuessemeyer: correct :facepalm:

daniel-wer (approved):
Great, works very well :+1:

PR #5492, opened 2021-05-17
Avoid duplicate layers when applying suggestions

### LINKED ISSUE #5401: Dataset Settings View: applySuggestedSettings mixes up layers
An additional color layer directory is linked into an existing dataset, without updating the datasource-properties.json on disk. Then the edit view is opened, where the suggestion to adapt additonal inferred information is accepted.

If this happens while the settings view is in “advanced” view, i.e. shows json, everything works as expected.

However, if the simple view is shown, the layers get mixed up, resulting in a duplicate of a previously existing layer, rather than the addition of the new layer. Switching to “advanced” then already shows the problem, but even if no such switch is performed, the broken settings are sent to the datastore upon hitting “Save”.

I already pushed a back-end side assertion on branch duplicate-layers, but while the front-end bug is in place, that would only move the error from postgres to the datastore, so I’m not opening a PR for that yet.

fm3:
### Steps to test:
- on disk, duplicate a layer of a dataset
- open dataset settings from dashboard, click accept selection, save
- new layer should be included, no duplicate assertion should show

### Issues:
- fixes #5401 

------
- [x] Ready for review


MichaelBuessemeyer:
Note https://github.com/scalableminds/webknossos/issues/5401#issuecomment-846000580

I just pushed a commit that solves the problem in two ways: 
1. The inferred datasource JSON is now also applied to the form fields used in the simple view and not only to the form used in the advanced view.
2. I got rid of the `merge` method in the `syncDataSourceFields` method and only copy the value from one view to the other. I played a bit around with this solution and it seems to work :)

@fm3 Could you please try to confirm that the newest commit fixes the bug and also try to test the new synchronization between the simple <-> advanced view so that the bug fix does not create new bugs?

Thanks :pray:

fm3:
Works great for me now! :tada: I also changed some fields in the respective views and even deleted an entire layer in the json, and the syncing seems to work fine as well

frontend/javascripts/dashboard/dataset/dataset_import_view.js
```diff
@@ -632,23 +633,18 @@ class DatasetImportView extends React.PureComponent<Props, State> {
     if (!form) {
       return;
     }
-
-    const parsedConfig = JSON.parse(form.getFieldValue("dataSourceJson"));
     if (syncTargetTabKey === "advanced") {
       // Copy from simple to advanced: update json
 
-      // parsedConfig has to be used as the base, since `dataSource` will only
-      // contain the fields that antd has registered input elements for
```
philippotto, line 641: I doubletested that this comment is indeed outdated. Opening the edit-view for a dataset and changing information in the basic tab, still has the complete JSON available when syncing the tabs :+1:

philippotto (approved):
Awesome! Took me a while to understand what's going on during testing, but everything seems to work perfectly :rocket:

philippotto:
I'm going to merge this now so that I can include this fix in the next release :100:

Issue #5496, opened 2021-05-19
Iterate on Isosurface Performance (with mappings?)


fm3:
We noticed fairly frequent 30s-timeouts for isosurface chunk requests. The chunks should be fairly small, it is not clear why this takes this long. We should investigate which part is slow. Note that an Agglomerate View Mapping (with a fairly high agglomeration rate) was active. It may make sense to measure the impact of the mapping and check if the caches are used correctly there

fm3:
Server logs show a `java.lang.ArrayIndexOutOfBoundsException` for this request. May be a duplicate of #5238  after all

fm3:
Yes, this is just how the actor + front-end handling of the failed requests looks now. Perf is ok (if there is no exception thrown)

PR #5499, opened 2021-05-19
Fix apply agglomerate 

youri-k:
-  fixes the reader range calculation

- https://___.webknossos.xyz

### Steps to test:
- see slack

### Issues:
- contributes to #5496 

------
- [x] Ready for review

### DIFF
--- a/webknossos-datastore/app/com/scalableminds/webknossos/datastore/storage/AgglomerateFileCache.scala
+++ b/webknossos-datastore/app/com/scalableminds/webknossos/datastore/storage/AgglomerateFileCache.scala
@@ -122,7 +122,7 @@ class BoundingBoxCache(val cache: mutable.HashMap[(Long, Long, Long), BoundingBo
 
     var x = initialBoundingBox._1
     var y = initialBoundingBox._2
-    var z = initialBoundingBox._3 + currDimensions._3
+    var z = initialBoundingBox._3
 
     while (x < requestedCuboid.x && x < dataLayerBox.x) {
       val nextBBinX = (x + currDimensions._1, y, z)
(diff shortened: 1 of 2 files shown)


webknossos-datastore/app/com/scalableminds/webknossos/datastore/storage/AgglomerateFileCache.scala
```diff
@@ -111,26 +111,35 @@ class BoundingBoxCache(val cache: mutable.HashMap[(Long, Long, Long), BoundingBo
...
     val dataLayerBox = request.dataLayer.boundingBox.bottomRight
+
+    // use the values of first bb to initialize the range and dimensions
     val initialValues = cache(initialBoundingBox)
     var range = initialValues.idRange
     var currDimensions = initialValues.dimensions
 
     var x = initialBoundingBox._1
     var y = initialBoundingBox._2
-    var z = initialBoundingBox._3 + currDimensions._3
```
youri-k, line 125: @fm3 The error was caused by this premature advancement of z. Since we already read the `range` and `currDimensions` two lines above, this saved one iteration in the inner loop. However, this corrupted the values of `nextBBinX` and `nextBBinY` because they started at the wrong z value. This seems to never have been problem before, but it is conceptually wrong. Now we do one "unnecessary" inner iteration and cache lookup, but I think this is better than a complicated check whether we are in the first iteration. Let me know if you think the comments and this explanation are sufficient or if there still is something unclear.

fm3: Cool, thanks, I think I get this a bit better now. Also thanks for the comments! They are very helpful!
If you agree, I think it would be great to also have a short higher-level comment block explaining what the cumsum.json is and what wk uses it for. Maybe even also for the agglomerate file itself. Feel free to merge after that :)

fm3 (approved):

PR #5502, opened 2021-05-20
Adapt local docker-compose service to new config format

fm3:
### Steps to test:
- create local docker setup via `./start_docker.sh` as described in readme (note that you may have to `systemctl stop postgresql` first if a postgres is running
- initial data should not be inserted and you should see the onboarding workflow at localhost:9000

------
- [x] Ready for review


youri-k (approved):
LGTM :shipit:

PR #5505, opened 2021-05-25
Add tags to mailchimp for some user interactions

### LINKED ISSUE #5450: On demo instance, tag users that already uploaded/annotated/etc in mailchimp
Send tag to mailchimp when a user
 - views publications
 - annotates own dataset
 - annotates public dataset
 - invites other users
 - uploads dataset

so that call-to-action emails can be sent to those who haven’t done so for some time after signup

Note that similar to the analytics events, those calls should not be blocking so that the user interaction is not interrupted if the request fails.

Some of these may have to be triggered from the front-end (like viewing publications) – note that privacy add-ons may block those trigger requests (as is fair).

fm3:
Adds mailchimp tags to users in the audience. These calls are non-blocking so the normal wk operation is not impeded.
The new tags are `has_annotated`, `has_invited_team`, `has_uploaded_own_dataset`, `has_viewed_published_dataset`.

### Steps to test:
- fill in valid credentials in `application.conf` key block `mailchimp`
- create annotation, view public dataset of other organization, upload dataset, send invites
- corresponding mailchimp user should get those tags.
- to test without credentials, add logging to `MailchimpClient.scala` L54 instead

### Issues:
- fixes #5450 

------
- [x] Ready for review

### DIFF
--- a/app/controllers/DataSetController.scala
+++ b/app/controllers/DataSetController.scala
@@ -18,6 +18,7 @@ import utils.ObjectId
 import javax.inject.Inject
 import models.analytics.{AnalyticsService, ChangeDatasetSettingsEvent, OpenDatasetEvent}
 import models.organization.OrganizationDAO
+import oxalis.mail.{MailchimpClient, MailchimpTag}
 import play.api.mvc.{Action, AnyContent}
 
 import scala.concurrent.duration._
@@ -34,6 +35,7 @@ class DataSetController @Inject()(userService: UserService,
                                   teamDAO: TeamDAO,
                                   dataSetDAO: DataSetDAO,
                                   analyticsService: AnalyticsService,
+                                  mailchimpClient: MailchimpClient,
                                   sil: Silhouette[WkEnv])(implicit ec: ExecutionContext)
     extends Controller
     with RemoteOriginHelpers {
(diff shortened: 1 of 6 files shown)


app/controllers/DataSetController.scala
```diff
@@ -220,6 +222,9 @@ class DataSetController @Inject()(userService: UserService,
           js <- dataSetService.publicWrites(dataSet, request.identity, organization, dataStore)
           _ = request.identity.map { user =>
             analyticsService.track(OpenDatasetEvent(user, dataSet))
+            if (dataSet.isPublic && dataSet._organization != user._organization) {
```
fm3, line 225: @normanrz @Albane-dlv is this the condition we want here? My understanding of the tag is that the user has viewed a dataset that is public and that is of a different organization.
Alternatively we could look at wether it is linked to a publication. What do you think?

Albane-dlv: I was thinking of: user has viewed a public dataset from any organisation. If that is possible, it would be great!

fm3: Sure, I changed it now :)

youri-k (approved):
Tested and everything works as described :tada:

PR #5507, opened 2021-05-25
Schedule fetching job status updates

### LINKED ISSUE #5331: Add ticker to fetch celery job status
Currently the job status is only fetched if a user visits the jobs listing. If they never do, the analytics can never report the failed job. We should add a ticker that does this every few minutes.

fm3:
Checks celery job status list every 5 minutes.
Note that user requests can trigger this in a higher frequency, this is just added in case users don’t list their jobs for a long time.

### Steps to test:
- enable jobs, see in logging that update runs every 5 minutes even if the user does not open jobs overview page

### Issues:
- fixes #5331

------
- [x] Ready for review


youri-k (approved):
LGTM :shipit:

PR #5510, opened 2021-05-26
Add jobs/worker status route

### LINKED ISSUE #5509: Add worker status routes
It would be great to have a route that shows
 - the number of running/pending worker jobs
 - the deployed worker version (compare also https://github.com/scalableminds/webknossos-worker/issues/30)

fm3:
### Steps to test:
- set jobsEnabled = true, visit /api/jobs/status
- should fail if flower is not available
- should otherwise list registered workers, and the number of currently pending / running jobs (across organizations)
- note that it also shows finished jobs, but only those flower still knows about (this is cleared on flower restart, postgres has old job info, though)

### Issues:
- fixes #5509

------
- [x] Ready for review

### DIFF
--- a/app/WebKnossosModule.scala
+++ b/app/WebKnossosModule.scala
@@ -1,5 +1,6 @@
 import com.google.inject.AbstractModule
-import controllers.{InitialDataService, JobService}
+import controllers.InitialDataService
+import models.job.JobService
 import models.annotation.AnnotationStore
 import models.binary.DataSetService
 import models.task.TaskService
--- a/app/controllers/JobsController.scala
+++ b/app/controllers/JobsController.scala
@@ -4,237 +4,19 @@ import java.nio.file.{Files, Paths}
 import java.util.Date
 
 import com.mohiva.play.silhouette.api.Silhouette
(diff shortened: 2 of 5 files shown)


fm3:
@jstriebel @normanrz I had hoped I could also query some kind of build info, but I haven’t found any way to query any custom information from the workers. Do you know of a way to query which container is running? Otherwise, I suppose we always have to use the kubernetes dashboard to see if upgrades have been successful. That’s acceptable too, I guess.

jstriebel:
> Do you know of a way to query which container is running

You would need to add an endpoint in the worker(s) which would yield this information. The information itself would need to be part of the build in CI, similar to the wk or voxelytics version information. Besides that it's possible to use the kubernetes API itself to get the information of Pods, but I think that would be a quite tight coupling.

fm3:
Thanks for your input! The workers are not web servers, so adding an endpiont there would require adding that. Maybe this will be an option in the future if we find this important. 
So for the moment I’d say this PR is already helpful as is, and when we are in doubt whether an update was complete we should for now look at the kubernetes dashboard.

youri-k (approved):
LGTM :shipit:

PR #5513, opened 2021-05-26
Increase threshold for slow request slack notifications from 5 to 10s

fm3:
We haven’t been paying much attention to the slow requests recently, and <10s seems acceptable as the front-end can handle it well.

- [x] Ready for review


youri-k (approved):
LGTM :shipit:

PR #5519, opened 2021-05-27
Mesh follow up

### LINKED ISSUE #5506: Precomputed Mesh Follow-Up
- [x] when reading segment id from centered voxel, also take into account fallback buckets if necessary. currently, the action will default to segment 0 (and crash; see below) if the centered bucket is still loading.
- [x] load all chunks in parallel to improve loading speed (this will block bucket requests, but since the mesh loading is triggered explicitly by the user, this should be alright)
- [x] don't fetch mesh for segment 0
  - [x] more generic: don't crash if the list-request fails / returns empty
- [x] deleting a segment while it's still loading produces an invalid state

philippotto:
- https://___.webknossos.xyz

### Steps to test:
- open a dataset with a mesh (let me know if you need that)
- move around, request a precomputed mesh, delete the mesh from the list (ideally, all of this should happen while data is loading; I did this by throttling the Internet speed)

### Issues:
- fixes #5506 with the following points:
  - [x] when reading segment id from centered voxel, also take into account fallback buckets if necessary. this also makes sure to correctly deal with fallback loading while "render missing data black" is enabled (previously, the code always backed out if the setting was enabled even though fallback rendering is at work when the bucket is still being fetched)
  - [x] load chunks in parallel to improve loading speed. now, a sliding window is applied to process the jobs with 6 "workers". this can clog the request queue of the entire page, but this should be alright, since the user requested the meshes explicitly. the reason I chose (and built) a pool for this is that the request can be easily aborted then. in hindsight (I'm just noticing that), aborting the requests via an `AbortController` would have been alright, too :thinking:  However, now, we can tweak the parallelization degree better.
  - [x] don't fetch mesh for segment 0
    - [x] more generic: don't crash if the list-request fails / returns empty
  - [x] deleting a segment while it's still loading produces an invalid state
- addtionally, the coloring of the meshes was tweaked a bit

------
- [X] Ready for review

### DIFF
--- a/frontend/javascripts/libs/task_pool.js
+++ b/frontend/javascripts/libs/task_pool.js
@@ -26,7 +26,7 @@ export default function processTaskWithPool<T>(
 
       // If that promise is done, process a new one (that way,
       // the pool size stays constant until the queue is almost empty.)
-      newPromise.then(startNextTask);
+      newPromise.then(startNextTask, startNextTask);
     };
 
     for (let i = 0; i < poolSize; i++) {
--- a/frontend/javascripts/libs/task_pool.js
+++ b/frontend/javascripts/libs/task_pool.js
@@ -1,0 +1,36 @@
+// @flow
+
+export default function processTaskWithPool<T>(
+  tasks: Array<() => Promise<T>>,
+  poolSize: number,
+): Promise<Array<T>> {
+  return new Promise((resolve, reject) => {
+    const promises = [];
+    let isFinalResolveScheduled = false;
+
+    const startNextTask = () => {
+      if (tasks.length === 0) {
+        if (!isFinalResolveScheduled) {
+          isFinalResolveScheduled = true;
+
+          // All tasks were kicked off, which is why all promises can be
+          // awaited now together.
+          Promise.all(promises).then(resolve, reject);
+        }
+        return;
+      }
+
+      const task = tasks.shift();
+      const newPromise = task();
+      promises.push(newPromise);
+
+      // If that promise is done, process a new one (that way,
+      // the pool size stays constant until the queue is almost empty.)
+      newPromise.then(startNextTask);
+    };
+
+    for (let i = 0; i < poolSize; i++) {
+      startNextTask();
+    }
+  });
+}
--- a/frontend/javascripts/oxalis/model/bucket_data_handling/data_cube.js
+++ b/frontend/javascripts/oxalis/model/bucket_data_handling/data_cube.js
@@ -510,24 +510,28 @@ class DataCube {
     this.trigger("volumeLabeled");
   }
 
-  hasDataAtPositionAndZoomStep(voxel: Vector3, zoomStep: number = 0) {
+  isZoomStepRenderableForVoxel(voxel: Vector3, zoomStep: number = 0): boolean {
+    // When this method returns false, this means that the next resolution (if it exists)
+    // needs to be examined for rendering.
+
     const bucket = this.getBucket(this.positionToZoomedAddress(voxel, zoomStep));
 
     if (!(bucket instanceof DataBucket)) {
       // This is a NullBucket (e.g., because it's out of the bounding box or in a not-existing
-      // magnification). Don't assume fallback rendering.
+      // magnification). This zoomstep is as good as all the other zoomsteps (as these will only
+      // hold null buckets, too).
       return true;
     }
 
     if (bucket.hasData() || bucket.isLoaded()) {
       // The data exists or the bucket was loaded at least (the latter case
-      // occurs when fallback rendering is disabled but the bucket is missing.
-      // Then, the bucket has the "isLoaded" state).
+      // occurs when renderMissingDataBlack is *enabled* but the bucket is missing.
+      // Then, the bucket has the "isLoaded" state and should be used for rendering).
       return true;
     }
 
     if (bucket.isMissing()) {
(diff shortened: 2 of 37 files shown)


frontend/javascripts/oxalis/model/bucket_data_handling/data_cube.js
```diff
@@ -511,7 +511,30 @@ class DataCube {
   }
 
   hasDataAtPositionAndZoomStep(voxel: Vector3, zoomStep: number = 0) {
-    return this.getBucket(this.positionToZoomedAddress(voxel, zoomStep)).hasData();
+    const bucket = this.getBucket(this.positionToZoomedAddress(voxel, zoomStep));
+
+    if (!(bucket instanceof DataBucket)) {
+      // This is a NullBucket (e.g., because it's out of the bounding box or in a not-existing
+      // magnification). Don't assume fallback rendering.
+      return true;
```
daniel-wer, line 519: I can't figure out why `true` is returned, here. If the magnification is not existing, shouldn't `hasDataAtPositionAndZoomStep` return `false` for this position and zoomStep?

frontend/javascripts/test/libs/task_pool.spec.js
```diff
@@ -0,0 +1,98 @@
...
+});
+
+test("processTaskWithPool should run tasks sequentially", async t => {
+  t.plan(1);
+
+  const protocol = [];
+
+  const tasks = [
+    async () => {
+      await Utils.sleep(1);
```
daniel-wer, line 30: This makes it more explicit. Otherwise the asserted result would also likely be the result if the tasks were ran in parallel.

```suggestion
      await Utils.sleep(3);
```

frontend/javascripts/test/libs/task_pool.spec.js
```diff
@@ -0,0 +1,98 @@
...
+      protocol.push(1);
+    },
+    async () => {
+      await Utils.sleep(1);
```
daniel-wer, line 34:
```suggestion
      await Utils.sleep(2);
```

daniel-wer (commented):
Nice fixes and improvements!

You write that
> the reason I chose (and built) a pool for this is that the request can be easily aborted then.

I didn't see any code for that. Is this already implemented? So if a large mesh is loaded and I hit the delete mesh button mid-way through the loading, are the remaining requests aborted/not scheduled?

philippotto:
Thank you for your feedback!

> You write that
> 
> > the reason I chose (and built) a pool for this is that the request can be easily aborted then.
> 
> I didn't see any code for that. Is this already implemented? So if a large mesh is loaded and I hit the delete mesh button mid-way through the loading, are the remaining requests aborted/not scheduled?

The remaining tasks are started, but these would immediately exit (and not schedule a request) when the mesh object isn't in the store, anymore. See https://github.com/scalableminds/webknossos/pull/5519/files#diff-90faecf8869dada70fb6445f2b776bb9029669ec8236d867f06efd16bb674788R79.
This means the task queue is still processed, but the overhead should be marginal. The alternative would be to react to the "delete" event, but I find the current approach more appealing, since it only relies on the Store.

frontend/javascripts/libs/task_pool.js
```diff
@@ -0,0 +1,36 @@
...
+        return;
+      }
+
+      const task = tasks.shift();
+      const newPromise = task();
+      promises.push(newPromise);
+
+      // If that promise is done, process a new one (that way,
+      // the pool size stays constant until the queue is almost empty.)
+      newPromise.then(startNextTask);
```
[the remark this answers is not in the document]
philippotto: Very good catch!

frontend/javascripts/oxalis/model/bucket_data_handling/data_cube.js
philippotto: Ah, I forgot to rename this method. The semantics of it should rather be "does the given zoomstep for the given position yield 'useful' data or should we look at a fallback resolution". For nullbuckets, looking at the fallback resolutions won't have any benefit, which is why `true` is returned. I'll definitely add a comment to this method, but another name should also be used. Maybe `isVoxelAtZoomStepRenderable`?

daniel-wer:
Mostly worked very well during testing!

> move around, request a precomputed mesh, delete the mesh from the list (ideally, all of this should happen while data is loading; I did this by throttling the Internet speed)

When doing this, I end up with grayed out "Segment(s)" in the meshes list. Interacting with those results in a page crash ("Something went wrong ..."). It happens like this: Throttle, move around, load mesh, delete mesh, wait (or unthrottle), grayed out Segment appears.

delete-while-loading-mesh

frontend/javascripts/oxalis/model/bucket_data_handling/data_cube.js
philippotto: I went for the name `isZoomStepRenderableForVoxel` and improved the comments in https://github.com/scalableminds/webknossos/pull/5519/commits/b82ccf510a119d71314022110566833efe1f7866.

philippotto:
> It happens like this: Throttle, move around, load mesh, delete mesh, wait (or unthrottle), grayed out Segment appears.

Argh, great that you found that during testing! Apparently, I didn't wait long enough after deleting a mesh during the throttling :disappointed: Should be fixed now!

daniel-wer (approved):
LGTM :+1:

PR #5522, opened 2021-05-28
Improve error reporting for task creation + mesh loading

### LINKED ISSUE #5525: Improve generic error message when bulk-creating tasks
## Detailed Description
When trying to bulk-create a task for a project that is not existing, I end up with an unhelpful error toast `Couldn't find the requested resource.` without further details. It would be great if the error message would be more specific and point out that the project id is non-existent.

## Steps to reproduce
- Go to the admin -> task view
- Open the bulk task creation
- Use the following for the bulk creation (you might need to adapt the task type id):
`ROI2017_wkw,609fca97c701002702e9eb9c,sampleExp,1,0,0,0,0,0,0,1,0,0,0,1024,1024,1024,someRandomProject`

fm3:
### Steps to test:
- Try to create task for non-existing project (e.g. in bulk task creation)
- Try to load mesh for segment that is not present in the mesh file
- Both should have readable error messages

fixes #5525

- [x] Ready for review


youri-k (approved):
LGTM :shipit:

PR #5523, opened 2021-05-28
Graceful Shutdown for Local Proxy

### LINKED ISSUE #5517: Killing local-development proxy often does not release port 5005
I have to manually kill the process blocking the port before I can start wK again. Is there a way to modify the proxy to allow for this?

fm3:
Changes how the local node proxy that starts webpack, sbt and fossildb is shutting down.

The java debug process sometimes does not get terminated properly when the sbt process is terminated. Since the update of 1.4 seems to have increased the frequency of this, I had another look at how to work around that.

In this PR I…

- Added a bit more logging to the proxy.
- Run shutdown method also on SIGINT
- Added exit(0) to shutdown method
- Send SIGTERM to subprocesses instead of SIGKILL
- Also kill java debug process (anything that blocks port 5005)

### Steps to test:
- locally, run `yarn start` and then hit ctrl+c in various stages of the execution (directly after starting, somewhat later, after wk has properly started)
- logging should show shutdown
- after you get a prompt again, `lsof -i:5005` should have empty output (port 5005 should be free again)

### Issues:
- fixes #5517 

------
- [x] Ready for review


youri-k (approved):
LGTM :shipit:

PR #5526, opened 2021-05-31
Fix analytics event for dataset-upload-page

philippotto:
…and never show error for failed analytics requests

- https://___.webknossos.xyz

### Steps to test:
- open dataset upload page
- analytics request should succeed and no error toast should be shown
- break analytics route (or use ublock origin)
- request should fail but it should not raise an error toast

### Issues:
- follow-up to #5490

------
- [X] Ready for review


fm3 (approved):
Great, thanks!

I assume `{}` is flow’s type annotation for objects? Are there any other places where we should replace `Object` by `{}`? (possibly in follow-up issue)?

philippotto:
> I assume `{}` is flow’s type annotation for objects? Are there any other places where we should replace `Object` by `{}`? (possibly in follow-up issue)?

Yes, good point! I created https://github.com/scalableminds/webknossos/issues/5527 as a follow-up.

PR #5528, opened 2021-05-31
Fix needsConversion detection in dataset upload

fm3:
Fixes a regression introduced in #5490 

### Steps to test:
- set up jobsEnabled
- drop zip with tiffs in dataset upload view
- upload
- job should bestarted

------
- [x] Ready for review


philippotto (approved):
thanks for investigating this!

PR #5530, opened 2021-05-31
Add manual pass state for long-running jobs

### LINKED ISSUE #5466: Job list view: Tooltip on “Started” state has wrong text
says something went wrong, but started doesn’t mean that.

### LINKED ISSUE #5516: “Manual pass” state for long-running jobs
When certain jobs failed, show some version of “manual pass” instead of “failed”, giving the admins a chance to manually repair / run the job, and then set it to success/manually-repaired again.

In the back-end, I suggest introducing an additional field for the manually-repaired state so that no conflicts with the status as reported by celery can come up.

fm3:
Shows `convert_to_wkw` jobs that failed as “Manual“ (meaning an admin will try to repair this job), unless an admin has alreadyset the job’s `manualState` to `SUCCESS` or `FAILURE` in the database.

### Steps to test:
- set up worker, jobsEnabled
- upload datasets that work and that don’t work
- run other jobs as well
- in postgres, set a job to manually repaired or failure (i.e. wontfix)
- that state should be shown then, links should become available in success case

### Note:

I am unsure about the wording (e.g. the displayed name “Manual” for that new state), and the tooltips. Feedback welcome :)

### Issues:
- fixes #5516 
- fixes #5466 

------
- [x] Ready for review


fm3:
@youri-k I would like to add a line to the evolution setting the manualState to FAILURE for all existing jobs where the celery state is also FAILURE (because we don’t want to repair all old jobs). Since you are more familiar with json in SQL, could you give me a pointer there? No hurry.

@philippotto I incorporated your feedback :) Another question: Do you think we should make this feature opt-in? (e.g. add a `manualPassJobs` array to the features block of the config?) Currently I think we have only one instance with jobsEnabled, but that may change

frontend/javascripts/admin/job/job_list_view.js
```diff
@@ -140,15 +140,21 @@ class JobListView extends React.PureComponent<Props, State> {
...
+    const tooltipMessages = {
+      UNKNOWN:
+        "The status information for this job could not be retreived. Please try again in a few minutes, or contact us if you need assistance.",
+      SUCCESS: "This job has successfully been executed.",
+      PENDING: "This job will run as soon as a worker becomes available.",
+      STARTED: "This job is currently running.",
+      FAILURE:
+        "Something went wrong when executing this job. Feel free to contact us if you need assistance.",
+      MANUAL:
+        "The automatic job failed, but admins have been notified and will attempt to repair it manually. Please check back here soon.",
```
philippotto, line 152: Maybe:

```suggestion
        "The job will be handled by an admin shortly, since it couldn't be finished automatically. Please check back here soon.",
```

This puts emphasis on what's happening next and avoids the negative word "fail"

package.json
```diff
@@ -119,6 +119,8 @@
     "flow-coverage": "node_modules/flow-coverage-report/bin/flow-coverage-report.js -i \"./frontend/javascripts/**/*.js\" -x \"./frontend/javascripts/test/**/*.js\" -t html -t text",
     "docs": "node_modules/.bin/documentation build --shallow frontend/javascripts/oxalis/api/api_loader.js frontend/javascripts/oxalis/api/api_latest.js --github --project-name \"webKnossos Frontend API\" --format html --output public/docs/frontend-api",
     "refresh-schema": "./tools/postgres/refresh_schema.sh && rm -f target/scala-2.12/src_managed/schema/com/scalableminds/webknossos/schema/Tables.scala",
+    "enable-jobs": "sed -i -e 's/jobsEnabled = false/jobsEnabled = true/g' ./conf/application.conf",
+    "disable-jobs": "sed -i -e 's/jobsEnabled = true/jobsEnabled = false/g' ./conf/application.conf",
```
philippotto, line 123: nice :+1:

frontend/javascripts/admin/admin_rest_api.js
```diff
@@ -833,11 +833,25 @@ export async function getJobs(): Promise<Array<APIJob>> {
...
     annotationId: job.commandArgs.kwargs.annotation_id,
     annotationType: job.commandArgs.kwargs.annotation_type,
-    state: job.celeryInfo.state || "UNKNOWN",
+    state: adaptJobState(job.command, job.celeryInfo.state, job.manualState),
+    manualState: job.manualState,
     createdAt: job.created,
   }));
 }
 
+function adaptJobState(command: string, celeryState: string, manualState: ?string): string {
```
philippotto, line 842:
```suggestion
function adaptJobState(command: string, celeryState: ?string, manualState: ?string): string {
```

I'd assume this at least, since the (old and new) code explicitly handles a falsy value of `celeryState`. Yet another instance where I would have hoped that flow warns about..

frontend/javascripts/types/api_flow_types.js
```diff
@@ -562,6 +562,7 @@ export type APIJob = {
   +boundingBox: ?string,
   +type: string,
   +state: string,
+  +manualState: string,
```
philippotto, line 565:
```suggestion
  +state: "UNKNOWN" | "SUCCESS" | "PENDING" | "STARTED" | "FAILURE" | "MANUAL",
  +manualState: "SUCCESS" | "FAILURE" | null,
```

Having these precise types here in the job type would add quite a bit to type safety. I don't know whether merely adding these annotations here will make flow happy. It could very well be that it will complain about `adaptJobState` afterwards. In that case, these two types (plus a celery type probably) would need to be extracted like this:

```js
export APIJobCeleryState = ?;
export APIJobState = "UNKNOWN" | "SUCCESS" | "PENDING" | "STARTED" | "FAILURE" | "MANUAL";
export APIJobManualState = "SUCCESS" | "FAILURE" | null;
```

Then, the new signature would be:

```js
adaptJobState(command: string, celeryState: APIJobCeleryState, manualState: APIJobManualState): APIJobState
```

philippotto (changes requested):
Front-end looks great! However, I added some comments regarding the typing which we should harden a bit to make the job handling more robust. The typing was already not optimal before this PR, but I think we should use the chance now to improve it :)

fm3:
@philippotto Thanks for your feedback! :) Could you have another look?

youri-k:
> Since you are more familiar with json in SQL, could you give me a pointer there?

I hope this line from the evolution 57 is helpful:
```
UPDATE webknossos.dataSet_layers dl
SET adminViewConfiguration = dS.adminViewConfiguration->'layers'->dl.name
FROM webknossos.dataSets dS
WHERE dl._dataSet = dS._id AND dS.adminViewConfiguration ? 'layers' AND dS.adminViewConfiguration->'layers' ? dl.name;
```
This sets the layer's `adminViewConfiguration` based on the entry of the layer in the general `adminViewConfiguration`. The `WHERE` clause checks whether the `adminViewConfiguration` contains a layer object and whether that object contains the specific layer entry.

I really like this cheat sheet, where all the operators are explained: https://devhints.io/postgresql-json

Hope this helps, but you can ping me again if something does not work

philippotto (approved):
Front-end looks top notch :tada:

youri-k (approved):
Backend and SQL LGTM :tada:

PR #5532, opened 2021-05-31
Fix creating/editing a volume task type

### LINKED ISSUE #5531: Creating/editing a volume task type fails
Creating/editing a volume task type fails with "invalid JSON format".

Also see https://discuss.webknossos.org/t/unable-to-edit-task-type/1671/1.

philippotto:
This PR fixes a regression which was introduced in 61d3bee93. Even though, a volume task type doesn't really need settings, such as for soma clicking, the task type format still requires this (according to front-end as well as back-end types). In earlier versions, we simply hid the form fields but still maintained their values. I restored this behavior with this PR.

- https://___.webknossos.xyz

### Steps to test:
- create a volume task type (didn't work before)
- edit a volume task type and save (even without doing any changes, this failed)

### Issues:
- fixes #5531

------
- [X] Ready for review


frontend/javascripts/admin/tasktype/task_type_create_view.js
```diff
@@ -290,27 +290,29 @@ class TaskTypeCreateView extends React.PureComponent<Props, State> {
...
-
-                    <FormItem name={["settings", "mergerMode"]} valuePropName="checked">
-                      <Checkbox>Merger Mode</Checkbox>
-                    </FormItem>
-                  </React.Fragment>
-                ) : null
-              }
+              {({ getFieldValue }) => (
+                <div
+                  style={{
```
daniel-wer, line 295: I'd propose to add a comment so that this doesn't happen again in the future

daniel-wer (approved):
Thank you for investigating and fixing this. :+1:

frontend/javascripts/admin/tasktype/task_type_create_view.js
philippotto: Very good point :+1:

PR #5534, opened 2021-06-01
Dataset upload: Fix special case in common prefix handling

### LINKED ISSUE #5529: Dataset upload: null pointer exeption for zip with just one file
Uploading a zip with just one file as dataset yields a null exception in `PathUtils.removeSingleFileNameFromPrefix` (`prefix.subpath(0,0)`) This breaks the dataset upload in this case.

fm3:
When uploading a single file inside a zipfile, the common prefix handling ran into a bug, breaking the whole upload. This is fixed here.

### Steps to test:
- set jobsEnabled = true (no need to set up worker)
- create zip file with single tif inside
- upload that as dataset, no exception should be logged (only that the worker can’t be reached, if not set up)

### Issues:
- fixes #5529 

------
- [x] Ready for review


youri-k (approved):
Nice catch :tada:

Issue #5231, opened 2021-03-03
Reorganize content of panels in core UI


philippotto:
See #4996 on ideas about how to restructure the panels. The actual layout rework is covered in #5230.

PR #5538, opened 2021-06-01
Make centerNewNode compatible with merger mode and also influence delete behavior

### LINKED ISSUE #5398: Deactivate "Center new nodes" doesn't work in merger mode and when deleting
## Context
Not centering nodes upon creating is great.

## Expected Behavior
- Should work in Merger Mode
- Should not move to active node, when deleting a node

## Current Behavior
https://user-images.githubusercontent.com/335438/114854745-6f946580-9de5-11eb-9fbf-0d500c51095e.mp4

## Steps to Reproduce the bug
1. Open an annotation
2. Deactivate center new nodes
3. Create some nodes, delete some nodes
3. Start Merger Mode
4. Create some nodes, delete some nodes

philippotto:
- https://fixcentersettinginmergermode.webknossos.xyz

### Steps to test:
- Enable "Center new Nodes" and delete/create nodes --> active node should be centered automatically
  - test the same in merger mode
- Disable "Center new Nodes" and delete/create nodes --> active node should be NT centered automatically
  - test the same in merger mode

### Issues:
- fixes #5398 

------
- [X] Ready for review

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -73,17 +73,17 @@ jobs:
           name: Build webknossos (sbt)
           command: |
             if [ "${CIRCLE_BRANCH}" == "master" ]; then
-              docker-compose run compile sbt clean compile stage
+              docker-compose run compile sbt -no-colors clean compile stage
             else
-              docker-compose run compile sbt compile stage
+              docker-compose run compile sbt -no-colors compile stage
             fi
       - run:
           name: Build webknossos-datastore (sbt)
-          command: docker-compose run base sbt "project webknossosDatastore" copyMessages compile stage
+          command: docker-compose run base sbt -no-colors "project webknossosDatastore" copyMessages compile stage
 
       - run:
           name: Build webknossos-tracingstore (sbt)
-          command: docker-compose run base sbt "project webknossosTracingstore" copyMessages compile stage
+          command: docker-compose run base sbt -no-colors "project webknossosTracingstore" copyMessages compile stage
 
       - save_cache:
           name: Save target cache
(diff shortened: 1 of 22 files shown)


philippotto:
Also see discussion in https://github.com/scalableminds/webknossos/issues/5398. The setting should be renamed imo, but I'm not too fond of my own suggestion. I'm open for feedback!

daniel-wer (commented):
Works as described :+1: 

I think the problem with finding a good new name for this setting comes from its somewhat inconsistent handling. For example, currently even with the setting off, the active node is centered when a node/tree is activated or when a tree is deleted.
One option could be to rename the setting to "Center Active Node" and if it is disabled, really only invoke the `centerActiveNode` saga function if the user triggers the centering explicitly (using the shortcut - `CENTER_ACTIVE_NODE` action).

philippotto:
> One option could be to rename the setting to "Center Active Node" and if it is disabled, really only invoke the centerActiveNode saga function if the user triggers the centering explicitly (using the shortcut - CENTER_ACTIVE_NODE action).

I thought about this, too, but I don't think this will be pleasant to use. At least, activating a tree should always center the corresponding node in my opinion. The shortcut ("S") to center a node is quite a pro-feature, I'd guess. Maybe an additional button to center the current node would be somewhat helpful (for example, in the new statusbar after #5231 is done). However, I imagine this to still feel clunky, since one would typically click this every time when selecting a tree.

All in all, I think/hope, the current behavior in this PR is quite pleasant :thinking: What do you think about "Auto-center active node". The "auto" part leaves the impact unspecific and suggests that it affects the "right" cases. A tooltip could explain the setting in detail then. What do you think?

daniel-wer:
> All in all, I think/hope, the current behavior in this PR is quite pleasant thinking What do you think about "Auto-center active node". The "auto" part leaves the impact unspecific and suggests that it affects the "right" cases. A tooltip could explain the setting in detail then. What do you think?

I like that solution :+1:

daniel-wer (approved):
LGTM :+1:

PR #5539, opened 2021-06-01
Fix sync between simple and advanced datasource view in dataset settings.

### LINKED ISSUE #5465: Changes of dataset settings in “advanced” view are not accepted
(I have not yet investigated this further, not sure if this always happened or what the exact conditions were)

quoting from https://discuss.webknossos.org/t/changes-of-dataset-settings-in-advanced-view-are-not-accepted/1662

> I’ve tried to manually fix the issue in the “advanced” view.
>
> The green check mark in the text field indicates that the modified configuration is valid. (I’ve only removed one of the two color layers.) When clicking “save”, webKnossos shows a “Please check that all form fields are valid” toast. Based on the color coding, I assume that this is only an “info”, not an “error” (red).
> Still, it seems that the save operation failed. The segmentation layer does not show up in the tracing view despite being present in the dataset config.

MichaelBuessemeyer:
This PR tries to solve the bug of #5465 where users could not submit the updated dataset settings, when their last step was removing a dataset layer from the datasource json setting.

The reason for this bug is, that once the user removes a layer in the advanced tab and then submits the form without changing back to the simple version, the form values are updated but not the form items of the simple view. As these form items were last rendered with outdated values, there are still the form items for the removed layer. They are not rendered and thus not updated by the value change. But they are still used to validate the currently edited settings. As the form items of the removed layer still existed, the validation does not succeed.
This problem is simply tackled by forcing a re-render of the form component from the parent, directly after syncing the field between the advanced and simple tab once the submit button is pressed. Luckily, the `forceRender` function called by the parent is enough as I could not get a `forceRender` solely for the `SimpleAdvancedDataForm`working. 

- https://___.webknossos.xyz

### Steps to test:
1. First do the following on the master:
2. Prepare a dataset with more than one layer. I tested with 3 layers.
3. Got to the settings of that dataset and except any inferred changes in the simple view of the data settings.
4. Switch to the advanced settings and then remove the first layer.  Then open the console and directly submit the changes without switching back to the simple view. The submit should fail and there should be errors in the console. (If this did not happen, try to redo the steps and this time remove another layer)
5. Once you could reproduce this, try the same on this branch. The error should no longer exist and the update of the data settings should succeed smoothly.

### Issues:
- fixes  #5465

------
- [x] Ready for review

### DIFF
--- a/frontend/javascripts/dashboard/dataset/dataset_import_view.js
+++ b/frontend/javascripts/dashboard/dataset/dataset_import_view.js
@@ -511,11 +511,16 @@ class DatasetImportView extends React.PureComponent<Props, State> {
       return;
     }
 
-    // Trigger validation manually, because fields may have been updated
-    form
-      .validateFields()
-      .then(formValues => this.submit(formValues))
-      .catch(errorInfo => this.handleValidationFailed(errorInfo));
+    const afterForceUpdateCallback = () =>
+      // Trigger validation manually, because fields may have been updated
+      form
+        .validateFields()
+        .then(formValues => this.submit(formValues))
+        .catch(errorInfo => this.handleValidationFailed(errorInfo));
+    // Need to force update of the SimpleAdvancedDataForm as removing a layer in the advanced tab does not update
+    // the form items in the simple tab (only the values are updated). The form items automatically update once
+    // the simple tab renders, but this is not the case when the user directly submits the changes.
+    this.forceUpdate(afterForceUpdateCallback);
   };
 
   submit = async (formValues: FormData) => {


daniel-wer (commented):
Nice find!

Your fix works, but I'm wondering whether we can use a more targeted approach. As far as I understand antd allows to declare dependencies to other form fields to update form items if their dependencies change. See https://ant.design/components/form/#dependencies (and related https://ant.design/components/form/#shouldUpdate). I feel like that's exactly what's needed to solve this nicely, but if it doesn't work, we'll go with your current fix.

daniel-wer:
Hmm, thinking about that again, my linked approach is probably not going to work easily as the form fields need to be removed (and not only updated) it seems. One could maybe wrap the whole layer list in a Form.Item and declare the dependencies on that to force it to rerender itself (with the correct layers) once the advanced view json changes. Not sure whether that's a good solution.

MichaelBuessemeyer:
>Hmm, thinking about that again, my linked approach is probably not going to work easily as the form fields need to be removed (and not only updated) it seems.

Yeah, that was also my problem. However, I did don't try out to solve the problem with the two properties you linked. 
The `shouldUpdate` property does not work. I added this property to all fields in the simple form and also wrapped all form items with a `Form.Item` that has the property. 
The same goes for the dependency property. Encapsulating the whole list with `<FormItem name="dataSource" dependencies={["dataSourceJson"]}>` does not work :unamused:

daniel-wer (approved):
Let's go with your current solution, then - thanks for trying this out :)

PR #5550, opened 2021-06-02
Fix desync of td view camera and store state

### LINKED ISSUE #5448: Distortion in 3D viewport
## Current Behavior
The 3D viewport in fullscreen mode is distorted (too-wide)
https://user-images.githubusercontent.com/335438/116822769-0cdef000-ab81-11eb-9a24-d0c620444235.mov

The video is captured in the `dark-theme` branch, but I can also reproduce that on the current master on webknossos.org

## Steps to Reproduce the bug
1. Open a dataset in view mode
2. Open 3D viewport in fullscreen 
3. Observe distortion (does not happen always)

## Your Environment for bug

* Browser name and version: Chrome 90
* Operating System and version: OS X
* Version of WebKnossos (Release or Commit): https://github.com/scalableminds/webknossos/commit/248d936f48640950a59b592370e79b6ff63b71da

### LINKED ISSUE #5520: When zoom is the first action, the aspect ratio looks odd.
## Context
When viewing trees or meshs in the maximized 3D view port. 
Same as  #5448 but I can reproduce the bug (see below).

## Expected Behavior
The aspect ratio of the 3D view port should stay the same

## Current Behavior
The aspect ratio changes

## Steps to Reproduce the bug

1. load tree, mesh so something is visible in the 3D view port
2. maximize the 3D view port with the short cut `.`
Screenshot 2021-05-27 at 17 28 46

3. use the mouse wheel to zoom (it has to be the first action in the maximized 3D view port)
Screenshot 2021-05-27 at 17 30 00

## Your Environment for bug
Firefox, MacBook

daniel-wer:
This bug led to a distortion of the 3d viewport when moving/zooming after maximizing it (could happen after minimizing it as well). The cause of the bug was a desynchronization of the 3d view camera internal state and its store state. The 3d view camera was adjusted when maximizing the 3d viewport without updating the store state. This led to distortion errors, afterwards.

I scanned the code but didn't find other occurrences where the 3d view camera is updated without updating the store state.

- https://___.webknossos.xyz

### Steps to test:
- Create some skeletons so that you see something in the 3d view
- Maximize the 3d view using `.`
- Then, without moving the mouse out of the 3d view or clicking, either zoom or move the 3d view -> No distortion should happen.
  - The repro can be a bit tricky. I would propose you to test this on master as well, to check whether the reproduction instructions work for you! For me the distortion doesn't always happen when zooming for some reason, but moving triggers it. If you cannot observe the distortion zoom a bit when not maximized and try again.

### Issues:
- fixes #5520 
- fixes #5448 

------
- [x] Ready for review

### DIFF
--- a/CHANGELOG.unreleased.md
+++ b/CHANGELOG.unreleased.md
@@ -17,7 +17,7 @@ For upgrade instructions, please check the [migration guide](MIGRATIONS.released
 -
 
 ### Fixed
--
+- Fixed a bug that caused a distortion when moving or zooming in the maximized 3d viewport. [#5550](https://github.com/scalableminds/webknossos/pull/5550)
 
 ### Removed
 - 
--- a/frontend/javascripts/oxalis/controller/camera_controller.js
+++ b/frontend/javascripts/oxalis/controller/camera_controller.js
@@ -36,6 +36,7 @@ import api from "oxalis/api/internal_api";
 type Props = {
   cameras: OrthoViewMap<typeof THREE.OrthographicCamera>,
   onCameraPositionChanged: () => void,
+  onTDCameraChanged: () => void,
   setTargetAndFixPosition: () => void,
 };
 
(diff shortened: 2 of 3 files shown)


philippotto (approved):
Awesome! The fix makes sense to me :+1: I was on the verge of giving up reproducing this issue, but in the end the repro somehow worked all of a sudden. I don't really know what changed inbetween :shrug: On the new dev instance, I couldn't provoke this error at all (and no other error either), so, I think, this is safe to be merged :100:

PR #5551, opened 2021-06-03
In find data, sample topleft last

### LINKED ISSUE #5543: Find data: topleft should not be prepended but appended
Find data jumps to the topleft corner currently, it should jump to the middle. Regression introduced in #5487

fm3:
We added topleft to the position sampling to provide a simple solution for the histogram for tiny datasets.
However, it had the effect that the find data functionality now tends to jump to topleft rather than the center. So the topleft should be appended rather than prepended.
This is slightly slower as the list append has to copy the list, but I’d say the impact is negligible as this is not a very long list and this is not a very frequent operation.

- https://finddatatopleftlast.webknossos.xyz

### Steps to test:
- click find data on a layer that extends to the topleft of the bounding box
- should jump to center
- histogram for tiny datasets should still be computed

### Issues:
- fixes #5543 

- [x] Ready for review

### DIFF
--- a/webknossos-datastore/app/com/scalableminds/webknossos/datastore/services/FindDataService.scala
+++ b/webknossos-datastore/app/com/scalableminds/webknossos/datastore/services/FindDataService.scala
@@ -93,7 +93,7 @@ class FindDataService @Inject()(dataServicesHolder: BinaryDataServiceHolder)(imp
       }
     }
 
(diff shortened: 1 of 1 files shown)


youri-k (approved):
Yes, I noticed this, too. Very nice fix :tada:

PR #5553, opened 2021-06-03
Log to airbrake when retrying file upload many times

### LINKED ISSUE #5405: Resumable upload should log to airbrake if it retries forever
We did not notice a bug with the dataset upload for quite a few days because the failing requests were retried indefinitely. Maybe we can add a counter and log to airbrake if there are suspiciously many retries.

Also, it might make sense to introduce more analytics events on dataset upload view

grittaweisheit:
This PR logs to airbrake and send an analytics event when a file upload is retried many times.

- https://___.webknossos.xyz

### Steps to test:
- abc

### Issues:
- fixes #5405

------
- [x] Ready for review


frontend/javascripts/admin/dataset/dataset_upload_view.js
```diff
@@ -304,8 +310,19 @@ class DatasetUploadView extends React.Component<PropsWithFormAndRouter, State> {
         this.setState({ isRetrying: false, uploadProgress: resumableUpload.progress() });
       });
 
-      resumableUpload.on("fileRetry", () => {
-        this.setState({ isRetrying: true });
+      resumableUpload.on("fileRetry", file => {
+        const currentRetries = this.state.retries + 1;
+        // log to airbrake in intervals of retriesUntilAnalyticsEventSent
+        if (currentRetries >= retriesUntilAnalyticsEventSent && currentRetries % 10 === 0) {
```
philippotto, line 316: I think, the `10` is also worth extracting.

frontend/javascripts/admin/dataset/dataset_upload_view.js
```diff
@@ -304,8 +310,19 @@ class DatasetUploadView extends React.Component<PropsWithFormAndRouter, State> {
         this.setState({ isRetrying: false, uploadProgress: resumableUpload.progress() });
       });
 
-      resumableUpload.on("fileRetry", () => {
-        this.setState({ isRetrying: true });
+      resumableUpload.on("fileRetry", file => {
```
philippotto, line 313: I didn't think of it on Friday, but now I'm not sure whether it's good to trigger the request on a per-file basis. If I want to upload 100 files, and there is a problem with the upload, the events will be sent 100x.

I would suggest the following:
- create a throttled function on module level with `_.throttle` (see https://lodash.com/docs/4.17.15#throttle)
- call that function in this `fileRetry` handler à la: `sendRetryToAnalytics(datasetName)` (note that only the dataset name is used for the reporting) without any if-conditions (the function will only be executed if it's last execution was at least N seconds ago)

This has the advantage of ignoring the different files and using `throttle` saves the hassle of counting with the retry variable etc. Here, we are doing something similar which you can use as a template: https://github.com/scalableminds/webknossos/blob/9ff32431e4aaf9cf3971be054b678b79db390c41/frontend/javascripts/oxalis/view/action-bar/save_button.js#L40-L40

philippotto (changes requested):
Thanks for implementing this! I have a suggestion on how to further improve the reporting. See my one comment :)

frontend/javascripts/admin/dataset/dataset_upload_view.js
philippotto: See my other comment. When using `_.throttle`, there is only the need for one single variable in my opinion (e.g., report every 60 seconds).

frontend/javascripts/admin/dataset/dataset_upload_view.js
```diff
@@ -296,7 +303,8 @@ class DatasetUploadView extends React.Component<PropsWithFormAndRouter, State> {
         this.setState({ isRetrying: false, uploadProgress: resumableUpload.progress() });
       });
 
-      resumableUpload.on("fileRetry", () => {
+      resumableUpload.on("fileRetry", file => {
+        logRetryToAnalytics(file);
```
philippotto, line 307: This passes `file` but ideally it should pass the dataset's name. Is this easy to change?

philippotto (approved):
Great :)

PR #5565, opened 2021-06-14
Omit parent release name from wk version string

fm3:
The parent release as obtained by `git describe --tags` only yields the latest release if that tag happened in master. However, our release process means that often the tags do not land as direct ancestors to newer commits. Thus, this version string is often more confusing than helpful.

- https://___.webknossos.xyz

### Steps to test:
- visit `api/buildinfo` both locally and on dev deployment
- should show "dev" or ci tag

------
- [x] Ready for review


philippotto (approved):
:+1:

Issue #5495, opened 2021-05-18
Importing NML with nested groups messes up hierarchy when using "Create a new tree group for this file"


philippotto:
# Current State
The symptom exists in front-end and back-end (however, the bugs are caused by different conditions apparently).
- [X] fix front-end bug
- [ ] fix back-end bug

# Front-end bug

## Context
When dragging an NML file into an active tracing (i.e., front-end import is used) and checking "Create a new tree group for this file." the hierarchy is not properly wrapped.

## Expected Behavior
The entire tree hierarchy of the input NML should be wrapped by one new group `G`.

## Current Behavior
Some deeply nested trees are directly within `G`.

## Steps to Reproduce the bug
1. Ask me for the NML file I used (or try to reproduce it by downloading a "complex" hierarchy and re-uploading it).
2. Drag it into a (skeleton/hybrid) tracing.

## Your Environment for bug
chrome on linux, but this shouldn't matter.

# Back-end bug

See https://github.com/scalableminds/webknossos/issues/5495#issuecomment-858735682

jstriebel:
Something similar happens also when dragging a NML into the View-Mode, which seems to use the backend-import. I experienced that a group-id was used within it's group, creating a recursion. This only happened when `Create a new tree group for this file` was ticked. This results in rendering artifacts and crashes, and the downloaded NML can not be re-uploaded (this might lend itself nicely as a test-case). See [this slack thread](https://scm.slack.com/archives/C5AKLAV0B/p1623331839023800) for details.

philippotto:
I could fix the front-end bug, but the back-end bug seems to have a different cause as the same input didn't cause any problems when using the back-end import.

PR #5573, opened 2021-06-16
Various fixes and improvements to the tree hierarchy view and NML import

### LINKED ISSUE #5568: After drag and drop of tree, the tree component displays error: "error rendering component"
## Context
When managing trees and groups in a tracing the tree component displays an error message.

## Steps to Reproduce the bug
1. create a tracing with trees and group of trees.
2. drag a tree from a group to another group
3. The destination group is quite large in my case (~200 subgroups)
4. scroll down
4. drop the tree 
5. go to the comments tab
6. go back to the tree tab

## Your Environment for bug
Linux Chrome, Mac Firefox

### LINKED ISSUE #5569: Drag and Drop of trees in read-only annotation should not work
## Context
Read-only skeleton annotation should not allow drag and drop of trees (e.g., between different tree groups).

## Expected Behavior
Drag and drop should always use a red indicator and discard the move operation.

## Current Behavior
The tree will be moved to the target group (but saving probably fails).

## Steps to Reproduce the bug
1. create skeleton annotation with at least one tree and one tree group
2. open that annotation from another account so that it's read only
3. try to move a try into another tree gruop
4.

## Your Environment for bug
chrome 90 / linux

philippotto:
- (1) fixes `Cannot have two HTML5 backends at the same time.` error which could occur when dragging a node and then switching directly to another tab (e.g., comments) and then back again.
- (2) fixes that the UI allowed mutating trees in the tree tab (dragging/creating/deleting trees and groups)
- (3) fixes "Create a new tree group for this file" setting in front-end import when a group id of 0 was used in the NML
- (4) makes use of `canNodeHaveChildren` for `SortableTree` which greatly improves the dragging behavior of trees/groups, since the library doesn't show drag previews where a tree acts as a parent (since this isn't possible, anyway). Previously, one had to carefully move the mouse so that that preview wasn't suggested.

- https://misctreehierarchyfixes.webknossos.xyz

### Steps to test:
(1):
- create an annotation and import an NML with a high tree/group count (you may want to download the NML from here: https://webknossos.brain.mpg.de/annotations/Explorational/60c078080100007000580107#849,1199,978,0,0.424,1615)
- drag a tree from one group to another (ideally, drag the node to the bottom of the view so that the scrolling starts)
- drop the tree and quickly switch to another tab (e.g., comments tab)
- switch back to the trees view
- the tree tab should not crash. this was already a bit hard to reproduce on master, which is why I'm only 90% sure that the fix really works :see_no_evil: 

(2)
- open an annotation with trees and groups and open the share dialog for it
- switch the visibility to public and copy the url
- open the url in an incognito tab --> the annotation is read only
- ensure that trees and groups cannot be created/deleted/dragged via the trees tab (also check the dropdown menu)

(3)
- import an NML which uses the group id 0 for one of its groups
- check "Create a new tree group for this file" and click import
- the NML should properly be wrapped with a new group and have its origin hierarchy

(4)
- in general, play around with the drag and drop behavior of the trees and groups. i think, that the dnd behavior feels way better now, since invalid drag previews are less frequent.
- also pay attention to details such as the drag-icon which appears on hover, since I had to change the implementation a bit, because I didn't want the drag preview to appear in read-only tracings

### Issues:
- fixes #5569 
- fixes #5568
- part of #5495 (only the frontend part)

------
- [x] Ready for review

### DIFF
--- a/CHANGELOG.unreleased.md
+++ b/CHANGELOG.unreleased.md
@@ -19,7 +19,7 @@ For upgrade instructions, please check the [migration guide](MIGRATIONS.released
 ### Fixed
 - Fixed a bug where dataset uploads of zips with just one file inside failed. [#5534](https://github.com/scalableminds/webknossos/pull/5534)
 - Fixed crashing tree tab which could happen when dragging a node and then switching directly to another tab (e.g., comments) and then back again. [#5573](https://github.com/scalableminds/webknossos/pull/5573)
-- Fixed that the UI allowed mutating trees in the tree tab (dragging/creating/deleting trees and groups). [#5573](https://github.com/scalableminds/webknossos/pull/5573)
+- Fixed that the UI allowed mutating trees in the tree tab (dragging/creating/deleting trees and groups) in read-only tracings. [#5573](https://github.com/scalableminds/webknossos/pull/5573)
 - Fixed "Create a new tree group for this file" setting in front-end import when a group id of 0 was used in the NML. [#5573](https://github.com/scalableminds/webknossos/pull/5573)
 
 
--- a/CHANGELOG.unreleased.md
+++ b/CHANGELOG.unreleased.md
@@ -14,10 +14,14 @@ For upgrade instructions, please check the [migration guide](MIGRATIONS.released
 -
 
 ### Changed
--
+- Improved dragging behavior of trees/groups in the tree tab. [#5573](https://github.com/scalableminds/webknossos/pull/5573)
 
 ### Fixed
 - Fixed a bug where dataset uploads of zips with just one file inside failed. [#5534](https://github.com/scalableminds/webknossos/pull/5534)
+- Fixed crashing tree tab which could happen when dragging a node and then switching directly to another tab (e.g., comments) and then back again. [#5573](https://github.com/scalableminds/webknossos/pull/5573)
+- Fixed that the UI allowed mutating trees in the tree tab (dragging/creating/deleting trees and groups). [#5573](https://github.com/scalableminds/webknossos/pull/5573)
+- Fixed "Create a new tree group for this file" setting in front-end import when a group id of 0 was used in the NML. [#5573](https://github.com/scalableminds/webknossos/pull/5573)
+
 
 ### Removed
 - 
--- a/frontend/javascripts/main.js
+++ b/frontend/javascripts/main.js
@@ -19,8 +19,6 @@ import Store from "oxalis/throttled_store";
 import { DndProvider } from "react-dnd";
 import { HTML5Backend } from "react-dnd-html5-backend";
 
-const dndBackend = HTML5Backend;
-
 async function loadActiveUser() {
   // Try to retreive the currently active user if logged in
   try {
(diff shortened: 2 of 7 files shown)


CHANGELOG.unreleased.md
```diff
@@ -14,10 +14,14 @@ For upgrade instructions, please check the [migration guide](MIGRATIONS.released
 -
 
 ### Changed
--
+- Improved dragging behavior of trees/groups in the tree tab. [#5573](https://github.com/scalableminds/webknossos/pull/5573)
 
 ### Fixed
 - Fixed a bug where dataset uploads of zips with just one file inside failed. [#5534](https://github.com/scalableminds/webknossos/pull/5534)
+- Fixed crashing tree tab which could happen when dragging a node and then switching directly to another tab (e.g., comments) and then back again. [#5573](https://github.com/scalableminds/webknossos/pull/5573)
+- Fixed that the UI allowed mutating trees in the tree tab (dragging/creating/deleting trees and groups). [#5573](https://github.com/scalableminds/webknossos/pull/5573)
```
daniel-wer, line 22:
```suggestion
- Fixed that the UI allowed mutating trees in the tree tab (dragging/creating/deleting trees and groups) in read-only tracings. [#5573](https://github.com/scalableminds/webknossos/pull/5573)
```

frontend/javascripts/oxalis/model/helpers/nml_helpers.js
```diff
@@ -531,7 +531,7 @@ export function wrapInNewGroup(
   const trees = _.mapValues(originalTrees, tree => ({
     ...tree,
     // Give parentless trees the new treeGroup as parent
-    groupId: tree.groupId || unusedGroupId,
+    groupId: tree.groupId != null ? tree.groupId : unusedGroupId,
```
daniel-wer, line 534: :see_no_evil: Nice find!

frontend/javascripts/main.js
```diff
@@ -48,7 +52,9 @@ document.addEventListener("DOMContentLoaded", async () => {
   if (containerElement) {
     ReactDOM.render(
       <Provider store={Store}>
-        <Router />
+        <DndProvider backend={dndBackend}>
```
daniel-wer, line 55: Was this fix documented somewhere or how did you come up with it?

daniel-wer (approved):
Awesome, so much good stuff! Works as described, thanks for providing the detailed testing instructions :tada: The improved dragging behavior is a real treat :+1:

CHANGELOG.unreleased.md
philippotto: good point :sweat_smile:

frontend/javascripts/main.js
philippotto: Yes, the bug is somewhat known. However, the solution is usually recommended for scenarios where multiple DnD components are used in parallel. Here's a brief mention of the technique: https://github.com/frontend-collective/react-sortable-tree#if-it-doesnt-work-with-other-components-that-use-react-dnd

I also added a code comment which points to the specific template I used for the fix.

PR #5575, opened 2021-06-17
Improve error reporting for unhandled rejections

philippotto:
- https://___.webknossos.xyz

### Steps to test:
- you can check out this branch locally and apply the following changes
```diff
diff --git a/conf/application.conf b/conf/application.conf
index feb05255b..329ccdccc 100644
--- a/conf/application.conf
+++ b/conf/application.conf
@@ -228,8 +228,8 @@ braintracing {
 # Front-end analytics
 airbrake {
   environment = "dev-local"
-  projectKey = "insert-valid-projectKey-here"
-  projectID = "insert-valid-projectID-here"
+  projectID = "contact-me-if-you-want-to-test-this" // alternatively, look up these attributes in our prod config
+  projectKey = "contact-me-if-you-want-to-test-this"
 }

 # Front-end analytics
diff --git a/frontend/javascripts/admin/admin_rest_api.js b/frontend/javascripts/admin/admin_rest_api.js
index 669ee9108..c8c43e2bd 100644
--- a/frontend/javascripts/admin/admin_rest_api.js
+++ b/frontend/javascripts/admin/admin_rest_api.js
@@ -1607,6 +1607,10 @@ export function getMeshfileChunksForSegment(
   );
 }

+function extraFn() {
+  throw new Error("v1: rejection error 3");
+}
+
 export function getMeshfileChunkData(
   dataStoreUrl: string,
   datasetId: APIDatasetId,
@@ -1616,6 +1620,9 @@ export function getMeshfileChunkData(
   position: Vector3,
 ): Promise<ArrayBuffer> {
   return doWithToken(async token => {
+    console.log("h");
+    throw new Error("v1: rejection error 5");
+    extraFn();
     const data = await Request.sendJSONReceiveArraybufferWithHeaders(
       `${dataStoreUrl}/data/datasets/${datasetId.owningOrganization}/${
         datasetId.name
diff --git a/frontend/javascripts/libs/error_handling.js b/frontend/javascripts/libs/error_handling.js
index 7edfb953c..91a980ab1 100644
--- a/frontend/javascripts/libs/error_handling.js
+++ b/frontend/javascripts/libs/error_handling.js
@@ -13,7 +13,7 @@ import window, { document, location } from "libs/window";

 // Note that if you set this value to true for debugging airbrake reporting,
 // you also need to set the values for projectID and projectKey in application.conf
-const LOG_LOCAL_ERRORS = false;
+const LOG_LOCAL_ERRORS = true;

 const UNHANDLED_REJECTION_LABEL = "UnhandledRejection";
 const UNHANDLED_REJECTION_PREFIX = `${UNHANDLED_REJECTION_LABEL}: `;

```
- afterwards, clicking "load precomputed mesh" for a dataset which has a precomputed mesh, will provoke an error
- this error should be logged to airbrake like this: 
  - https://scm.airbrake.io/projects/95438/groups/3033565197044084546?tab=notice-detail
  - or https://scm.airbrake.io/projects/95438/groups/3033564689927565119?tab=overview
  - (depending on which error-throwing code was used in the above diff)
- note that the error is marked as "UnhandledRejection" which we also marked as an error class which should use "strict grouping" in airbrake. this means that the backtrace of the error will be used as a grouping criterium (previously, all rejection errors were only grouped by the line of code which emitted the error which is always the same due to the `unhandledrejection` handler)

### Issues:
- improves airbrake logging

------
- [X] Ready for review


frontend/javascripts/libs/error_handling.js
```diff
@@ -161,6 +167,13 @@ class ErrorHandling {
 
   assertExtendContext(additionalContext: Object) {
     this.airbrake.addFilter(notice => {
+      notice.errors.forEach(error => {
+        const index = error.message.indexOf(UNHANDLED_REJECTION_PREFIX);
+        if (index > -1) {
+          error.type = UNHANDLED_REJECTION_LABEL;
```
daniel-wer, line 173: This, together with configuration in airbrake to set "strict grouping" for this error type is the main contribution of this PR, right? :)

daniel-wer (approved):
Cool that you found a way to improve this! :+1:

frontend/javascripts/libs/error_handling.js
philippotto: Yes, but it only works in combination with the changes regarding the `wrappedError` since that makes sure that the stacktrace of the original error is sent to airbrake. Without the correct stacktrace, the strict grouping wouldn't work.

Issue #5567, opened 2021-06-14
Publish Datastore Jar for Releases


fm3:
The fat jar we build for the datastore (as of https://github.com/scalableminds/webknossos/pull/5558 ) should be published somewhere. Let’s discuss how this process should look like. Possibly publish to our own file server, and link to it from the docs (?)

It might also make sense to bundle it with a basic default config file to show how to overwrite the most important config fields.

Also it could be bundled with other files to handle auto-updates and https certificates, compare #5548 #5547 #5549

youri-k:
@fm3 talked about open points and gathered these tasks

- [ ] auto updater update possibility
- [ ] add assets in ci to github release
- [ ] (interactive) install script
- [ ] version from build info plugin
- [ ] ability to remove datastore
- [ ] ability to work without org folder

fm3:
Now tracked in #5468

PR #5579, opened 2021-06-22
Self-Hosted User Datastores

### LINKED ISSUE #5548: Auto-update for self-hosted datastores

youri-k:
- https://autoupdate.webknossos.xyz

### Steps to test:
- build datastore jar with `sbt "project webknossosDatastore" assembly`
- copy datastore jar into `tools/auto_update` folder and rename it to `webknossos.jar`
- execute `run.sh`
- => datastore should startup, check for updates and restart itself after downloading the update
- => after initial update occurs, no further restarts should happen
- => if `autoUpdate.enabled` is `false`, no update should happen
- to test the install script, send a `POST` request to `/api/userDatastores` with the following body fields `{url: string, port: int, name: string}`
- => should return a custom install script that guides you through the setup process 

### Issues:
- fixes #5579 

------
- [x] Ready for review


youri-k:
@fm3 This is definitely not polished, but it's working :tada: 
I used this SO answer for guidance https://stackoverflow.com/a/4002730 and chose the bash approach because it leaves the datastore JAR almost unedited.

Happy to hear what you think about it. The update checking and downloading also depends on #5567 :slightly_smiling_face:

fm3:
I think this looks pretty good already! Thanks for including the example server! I agree that a server-side file listing or some equivalent of the `/version` route is a good idea. I also think replacing the jar on disk is the right track. I wonder, though, if we really need the `run.sh` – if webKnossos moved the jar itself on disk, would that cause problems? I assume the jar is loaded into memory anyway on launch?
If so, we could possibly just have a restart-loop launching the jar. This is something people will want to have anyway – some sort of autoheal.
Also, I’m not sure if `exit` directly after the download may not be a bit too aggressive, but I also don’t really know what to do instead, prompting the user does not seem like an option. Maybe schedule the exit to 4 a.m. in the local timezone 🤷? Or even to a date specified by the server? Maybe the immediate exit is also ok, we probably won’t publish new datastore releases super often. What are your thoughts on this @normanrz ?

webknossos-datastore/app/com/scalableminds/webknossos/datastore/services/BinaryDataServiceHolder.scala
```diff
@@ -14,8 +14,11 @@ import javax.inject.Inject
...
 class BinaryDataServiceHolder @Inject()(config: DataStoreConfig, agglomerateService: AgglomerateService) {
 
-  val binaryDataService = new BinaryDataService(Paths.get(config.Datastore.baseFolder),
-                                                config.Datastore.Cache.DataCube.maxEntries,
-                                                agglomerateService)
+  val binaryDataService = new BinaryDataService(
+    Paths.get(config.Datastore.baseFolder),
+    config.Datastore.Cache.DataCube.maxEntries,
+    agglomerateService,
+    config.Datastore.SingleOrganizationDatastore.organizationName
```
fm3, line 21: It may make sense to pass the whole config object now

webknossos-datastore/app/com/scalableminds/webknossos/datastore/services/AutoUpdateService.scala
```diff
@@ -0,0 +1,99 @@
...
+import scala.concurrent.ExecutionContext.Implicits.global
+import scala.concurrent.duration._
+import scala.sys.exit
+
+trait GithubReleaseChecker {
+  def rpc: RPC
+
+  def checkForUpdate(): Fox[(Boolean, List[JsObject])] =
+    for {
+      // TODO migrate to webknossos repo once released
```
fm3, line 26: todo

webknossos-datastore/app/com/scalableminds/webknossos/datastore/rpc/RPCRequest.scala
```diff
@@ -43,6 +43,11 @@ class RPCRequest(val id: Int, val url: String, wsClient: WSClient) extends FoxIm
     this
   }
 
+  def addHeader(header: (String, String)) = {
+    request = request.addHttpHeaders(header)
+    this
+  }
```
fm3, line 49: how about `accepts(contentType: String)`?

webknossos-datastore/app/com/scalableminds/webknossos/datastore/helpers/SingleOrganizationAdapter.scala
```diff
@@ -0,0 +1,35 @@
...
+  def resolveOrganizationFolderIfExists(path: Path, organizationName: String): Path =
+    singleOrganizationName match {
+      case Some(_) => path
+      case None    => path.resolve(organizationName)
+    }
+
+  protected def isSingleOrganizationDataStore: Boolean =
+    singleOrganizationName.isDefined
+
+  // Previously check if datastore is single organization, otherwise no guarantees on name
```
fm3, line 20: Why is this necessary?

tools/auto_update/server.js
```diff
@@ -0,0 +1,25 @@
...
+          res.writeHead(404);
+          res.end(JSON.stringify(err));
+          return;
+        }
+        res.writeHead(200);
+        res.end(data);
+      });
+    }
+  })
+  .listen(9090);
```
fm3, line 25: What’s this whole server for? It serves the version or arbitrary files? Is this going to be publicly available / where should it run?

tools/auto_update/my-datastore.conf
```diff
@@ -0,0 +1,59 @@
...
+  }
+  isosurface {
+    timeout = 30 seconds
+    actorPoolSize = 1
+  }
+  agglomerateSkeleton.maxEdges = 10000
+  autoUpdate.enabled = true
+}
+
+pidfile.path = "/dev/null" # Avoid the creation of a pid file
```
fm3, line 59: Is this file needed in the repo? I thought the install script writes it?

app/models/binary/DataStore.scala
```diff
@@ -84,6 +102,39 @@ class DataStoreService @Inject()(dataStoreDAO: DataStoreDAO)(implicit ec: Execut
...
+      installScript = InstallScript.getInstallScript(config.name,
+                                                     config.url,
+                                                     config.port,
+                                                     organization.name,
+                                                     key,
+                                                     conf.Http.uri,
+                                                     updateScriptURL,
+                                                     jarURL)
+
+    } yield writeInstallScriptToFile(installScript)
```
fm3, line 135: Is there a way to remove it again? Or will the user after disabling their local datastore forever see the errors when hitting the dashboard dataset refresh button?

app/models/binary/DataStore.scala
```diff
@@ -84,6 +102,39 @@ class DataStoreService @Inject()(dataStoreDAO: DataStoreDAO)(implicit ec: Execut
...
     } yield result).getOrElse(Forbidden(Json.obj("granted" -> false, "msg" -> Messages("dataStore.notFound"))))
 
+  def createNewUserDataStore(config: UserDataStoreConfig, organizationId: ObjectId): Fox[TemporaryFile] = {
+    def writeInstallScriptToFile(fileContent: String): TemporaryFile = {
+      val file = temporaryFileCreator.create()
+      val fw = new FileWriter(file)
+      fw.write(fileContent)
+      fw.close()
+      file
+    }
```
fm3, line 112: I feel like this method could be extracted to some file utils trait somewhere?

app/models/binary/InstallScript.scala
```diff
@@ -0,0 +1,218 @@
...
+       |		sudo systemctl start webknossos-datastore.service
+       |	fi
+       |
+       |	echo "Service installation completed. To inspect the logs of the running webKnossos datastore module, you can use \"journalctl -u webknossos-datastore\" ."
+       |
+       |fi
+       |
+       |echo "Install completed! Have fun using your own webKnossos datastore."
+       |""".stripMargin
+}
```
fm3, line 218: I am not very familiar with bash, some things I noticed here are
 - We rely on things like sudo and apt being present, and the user having bash. Is this risky? At least it should be communicated in the wk UI where the user downloads the script
 - Should we add default options for the yes-no-questions so that enter gets you to a standard case?
 - What can a user do after aborting the script mid-way? Does removing some generated directories restore a clean state so that they can start over? Should we explain this somewhere?
 - I’d ask @jstriebel to have a look at the bash script please, as I’m not very familiar with the language
 - Maybe @philippotto or @normanrz could comment on the phrasing / wording / what would you expect from this script.

fm3 (changes requested):
Cool stuff, I’m looking forward to seeing this in action! I left some questions and suggestions, but nothing really major

app/models/binary/InstallScript.scala
jstriebel: I just had a rough first look, this looks really good already. For better portability rather use `#!/usr/bin/env bash` than `/bin/bash` as the shebang. Checking for sudo and apt being present upfront (e.g. using `which`) and outputting a helpful error message would be nice.
Except for that, please check it also with https://www.shellcheck.net, this is quite helpful.

jstriebel: PS: Please link to the respective documentation pages when referring to them

app/models/binary/DataStore.scala
youri-k: Yes, now there is. Have a look at https://github.com/scalableminds/webknossos/pull/5579/files#diff-602b1fba87abc9f99dcd52e9d8de29dc2d6b5b7f444c4db7b39920ffb2c11130R96-R109

app/models/binary/InstallScript.scala
philippotto: > Maybe @philippotto or @normanrz could comment on the phrasing / wording / what would you expect from this script.

No strong opinion. My 2 cents:

`Installation completed! Your own webKnossos datastore is now ready to use. In case you experience any problems, please contact hello@scm.io.`

youri-k: Thanks for your input :slightly_smiling_face: 
@jstriebel I implemented all suggestion from shellcheck. Could you have another look? Regarding the documentation, I did not find any documentation for the config files. How would you communicate to the users that they need to look into the standalone conf file for all options @fm3?

> We rely on things like sudo and apt being present, and the user having bash. Is this risky? At least it should be communicated in the wk UI where the user downloads the script.

There are only a few places, where we need either of these present and it is communicated before, so I do not think that there is need for change.

> Should we add default options for the yes-no-questions so that enter gets you to a standard case?

Interesting question :thinking: I am a little concerned that the user would rush through the installation without reading some important questions. In the standard case (auto SSL and no service), the user has to input 7 things with 4 of them single key strokes. I feel this should not be a huge burden. However, I can try to add a default path, if you want.

> What can a user do after aborting the script mid-way? Does removing some generated directories restore a clean state so that they can start over? Should we explain this somewhere?

This script should be idempotent, so there is no work required after aborting. I added a comment in the first line about this.

Feel free to have another look :slightly_smiling_face:

fm3: > How would you communicate to the users that they need to look into the standalone conf file for all options @fm3?

I think it’s fine not to explain the config format specifically. This setup is (for now) for users who don’t want particularly specific config options but a quick test.

> This script should be idempotent, so there is no work required after aborting.

Awesome!

> I am a little concerned that the user would rush through the installation without reading some important questions

Fair enough, feel free to keep it then.

> Feel free to have another look

Thanks, I’ll do that later today!

webknossos-datastore/app/com/scalableminds/webknossos/datastore/helpers/SingleOrganizationAdapter.scala
```diff
@@ -0,0 +1,46 @@
...
+  protected def replaceDataSourceOrganizationIfNeeded(dataSource: DataSource): DataSource =
+    if (isSingleOrganizationDataStore) {
+      dataSource.copy(id = dataSource.id.copy(team = ""))
+    } else {
+      dataSource
+    }
+
+  private def assertNonEmptyName(organizationName: String): Option[String] =
+    if (organizationName == "") {
+      logger.error("Config error: empty organization name. Please provide a non empty name")
```
fm3, line 34:
```suggestion
      logger.error("Config error: Please provide a non-empty organization name for this single-organization datastore")
```

fm3 (approved):
Looks good! Let’s set this up so @normanrz can test it (via install script) and give feedback on the usability.

Should we also already include a UI in webKnossos to download the install script (+register the datastore)?

app/models/binary/InstallScript.scala
jstriebel: I just had another look, LGTM :+1:

philippotto:
> Looks good! Let’s set this up so @normanrz can test it (via install script) and give feedback on the usability.

Has this already been done? Setup and/or testing? Would be great to see this past the finish line :)

app/models/binary/InstallScript.scala
```diff
@@ -0,0 +1,225 @@
...
+       |echo
+       |
+       |if [[ $$disableFileSystemWatching == "n" ]]; then
+       |	config_values[watchFileSystem]="false"
+       |else
+       |	config_values[watchFileSystem]="true"
+       |fi
+       |
+       |echo "To use your own webKnossos datastore, it must support HTTPS when communicating with the webKnossos server. There are multiple ways to implement this."
+       |echo "Option 3 will guide you to retrieve a SSL certificate and automatically renew the certificate for you. However, this option requires root permissions. Please note that you have renew the certificate by yourself when not using option 3."
```
fm3, line 97: make less verbose

app/models/binary/InstallScript.scala
```diff
@@ -0,0 +1,225 @@
...
+       |echo "2. You already setup HTTPS for your machine"
+       |echo "3. You have not setup HTTPS yet."
+       |echo "Please enter your option [1/2/3]."
+       |
+       |https_option=$$(retry "[[ !(\\$$variable == 1 || \\$$variable == 2 || \\$$variable == 3) ]]" "Unknown option")
+       |
+       |if [[ $$https_option == "1" ]]; then
+       |	echo "Please update your webKnossos datastore config accordingly. You will need to change the HTTP configuration, namely the URI and the PORT."
+       |	config_values[uri]="CHANGE_ME"
+       |	config_values[port]="CHANGE_ME"
```
fm3, line 108: prompt here directly?

fm3 (commented):
Thanks for your work on this! As discussed, I’ll take over from here :) Added some notes to self…

fm3:
We have decided to not pursue this further for the moment, since fileserver-only data hosting becomes an option with Zarr (compare #6019 ) and would be considerably simpler to set up for users who want to host their own data.

PR #5582, opened 2021-06-22
Hide unlisted users from more API calls

### LINKED ISSUE #5580: unlisted users still show up in “users with access rights to dataset” list
should be hidden

fm3:
### Steps to test:
- sign up as second user, activate
- in local database set that user to isUnlisted = true
- should no longer show in dataset access list (plus sign in dashboard dataset list)

### Issues:
- fixes #5580 

------
- [x] Ready for review


youri-k (approved):
Tested and works perfectly :tada:

PR #5583, opened 2021-06-22
Fix undefined access on not-existing segmentation layer

philippotto:
- https://fixundefinedmesh.webknossos.xyz

### Steps to test:
- open a dataset without segmentation layer in view mode and switch to the `Meshes` tab (within a hybrid annotation, an segmentation layer would always exist)
- ensure that the console doesn't print `Uncaught (in promise) TypeError: UnhandledRejection: "TypeError: Cannot read property 'fallbackLayer' of null"`

### Issues:
- fixes airbrake reports I saw with `Uncaught (in promise) TypeError: UnhandledRejection: "TypeError: Cannot read property 'fallbackLayer' of null"`

------
- [X] Ready for review


frontend/javascripts/oxalis/view/right-menu/meshes_view.js
```diff
@@ -59,7 +59,7 @@ export const stlIsosurfaceConstants = {
 
 // This file defines the component MeshesView.
 
-const mapStateToProps = (state: OxalisState) => ({
+const mapStateToProps = (state: OxalisState): * => ({
```
philippotto, line 62: Took me a while to figure out why the errors regarding the `Props` were not caught by flow. It turns out, flow simply assumed the props to be `any`. The reason for this is that the extraction of the return type for these functions here didn't work. After a bit of digging I found out that the `*` as a return type helps :facepalm: 

Here's the comment: https://github.com/facebook/flow/issues/7071#issuecomment-553896587
The issue was opened 2.5 years ago. Yet another reason to ditch flow?

I also looked for `ExtractReturn` and its usages are very rare. So, hopefully no other bad suprises.

CHANGELOG.unreleased.md
```diff
@@ -18,6 +18,7 @@ For upgrade instructions, please check the [migration guide](MIGRATIONS.released
 
 ### Fixed
 - Fixed a bug where dataset uploads of zips with just one file inside failed. [#5534](https://github.com/scalableminds/webknossos/pull/5534)
+- Fixed a benign error message when a dataset without a segmenatation layer was opened in view mode or with a skeleton-only annotation. [#5583](https://github.com/scalableminds/webknossos/pull/5583)
```
daniel-wer, line 21:
```suggestion
- Fixed a benign error message when a dataset without a segmentation layer was opened in view mode or with a skeleton-only annotation. [#5583](https://github.com/scalableminds/webknossos/pull/5583)
```

daniel-wer (approved):
Nice find and fix! Thank you for spending the extra time to get to the root of why flow didn't catch this :tada:

PR #5584, opened 2021-06-22
Update caniuse

philippotto:
When starting webpack, `caniuse` nags to do an update which I did now:

```diff
Target browser changes:
- and_chr 87
+ and_chr 91
- and_ff 83
+ and_ff 89
- android 81
+ android 91
- chrome 87
- chrome 86
- chrome 85
+ chrome 91
+ chrome 90
+ chrome 89
- edge 87
- edge 86
+ edge 91
+ edge 90
- firefox 83
- firefox 82
+ firefox 89
+ firefox 88
- ios_saf 14.0-14.2
- ios_saf 12.2-12.4
+ ios_saf 14.5-14.6
+ ios_saf 14.0-14.4
- op_mob 59
+ op_mob 62
- opera 72
- opera 71
+ opera 76
+ opera 75
- safari 13.1
+ safari 14.1
- samsung 12.0
+ samsung 14.0

```


yarn.lock
```diff
@@ -2935,9 +2935,9 @@ camelcase@^5.0.0, camelcase@^5.2.0, camelcase@^5.3.1:
   integrity sha512-L28STB170nwWS63UjtlEOE3dldQApaJXZkOI1uMFfzf3rRuPegHaHesyee+YxQ+W6SvRDQV6UrdOdRiR153wJg==
 
 caniuse-lite@^1.0.30001165:
-  version "1.0.30001165"
-  resolved "https://registry.yarnpkg.com/caniuse-lite/-/caniuse-lite-1.0.30001165.tgz#32955490d2f60290bb186bb754f2981917fa744f"
-  integrity sha512-8cEsSMwXfx7lWSUMA2s08z9dIgsnR5NAqjXP23stdsU3AUWkCr/rr4s4OFtHXn5XXr6+7kam3QFVoYyXNPdJPA==
+  version "1.0.30001239"
+  resolved "https://registry.npmjs.org/caniuse-lite/-/caniuse-lite-1.0.30001239.tgz"
```
daniel-wer, line 2939: This is now the only package in the lockfile that resolves from `registry.npmjs.org` instead of `registry.yarnpkg.com` - not sure how that happened or whether it is problematic :thinking:

philippotto: Hmm, I followed the official instructions of how to update caniuse. That instruction was probably not targeted for yarn users. However, I'd hope that it shouldn't be problematic :shrug:

philippotto: However, I can try to update it again with yarn. Shouldn't be a big deal, either. Will report back in a minute or so ^^

philippotto: I had to "fix" it manually. But yarn seems to be happy about it. If the CI succeeds, it should be alright, I guess.

daniel-wer (approved):
Great, thanks! :+1:

PR #5588, opened 2021-06-24
Fix that 3d view broke if it had no width

### LINKED ISSUE #5472: Reducing size of 3d Viewport to (0,0) crashes it.
## Context
If you resize the viewports in a way that the 3d viewport has a height and width of 0 (e.g. done by dragging the horizontal splitter to the bottom  and the vertical splitter to the right) leads to a strange / broken behaviour of the 3d viewport. The viewport doesn't display anything anymore after resizing the splitters again so that the viewport is visible.
image

## Expected Behavior
See the image above. Nothing is displayed anymore and the scalebar says NaN.

## Current Behavior
The 3d viewport should behave like expected / recover from that error sta

## Steps to Reproduce the bug
1. Reset to the default layout.
2. Move the vertical splitter as far to the right as possible.
3. Move the horizontal splitters as far to the bottom as possible.
4. Move the splitter back to where they were or reset the layout.
5. The 3d viewport is broken :boom: 

~~ Your Environment for bug~~

daniel-wer:
The condition that checked whether the 3d view was not visible, to avoid updating the camera state in that case, was too unspecific. It only checked whether the 3d view had 0 height (and not straightforward, but only by checking whether the result of a division was NaN), but not whether it had 0 width. I changed the condition to be more explicit and complete.

- https://___.webknossos.xyz

### Steps to test:
- Easier reproduction instructions than in the issue
- Move the vertical splitter (from between the viewports) all the way to the right so that the 3d view is no longer visible.
- Move it back -> the 3d view content should be visible again (on master it is not)

### Issues:
- fixes #5472 

------
- [x] Ready for review


philippotto (approved):
works great and the code looks better, too :rocket:

PR #5595, opened 2021-06-30
cross origin api sends acc message when done

### LINKED ISSUE #5552: Allow to await cross-origin API
## Detailed Description

The cross-origin API can be used to communicate with webKnossos embedded in an IFrame which works via [postMessage communication](https://developer.mozilla.org/en-US/docs/Web/API/Window/postMessage). This works. However, sometimes it needs to be possible to find out when a specific cross-origin API call has finished so that a subsequent API call can be deferred until then. Currently, this is not possible. This could be implemented by answering a cross-origin API call with a `postMessage` to the [source](https://developer.mozilla.org/en-US/docs/Web/API/MessageEvent/source) of the original message once the API call was executed, thereby allowing synchronous cross-origin API calls.

## Context

Important for projects embedding and communicating with webknossos in an IFrame (e.g. the connectome viewer).

grittaweisheit:
This PR changes the cross origin api. It now sends an acknowledgment message after finishing it's work.

Edit Daniel: I also added that the api call is indeed awaited before acknowledging it. Additionally, error messages are sent if errors occur.

- https://___.webknossos.xyz

### Steps to test:
- send request to cross origin api
- when ready, an `ack` message should be send

### Issues:
- fixes #5552 

------
- [x] Ready for review


philippotto (approved):
great! looks good to me :)

PR #5596, opened 2021-07-01
Fix Wrapping Nested Tree Groups During Upload

### LINKED ISSUE #5495: Importing NML with nested groups messes up hierarchy when using "Create a new tree group for this file"

fm3:
Fixes a bug where nested tree groups were messed up during NML upload if “Create a new tree group for this file.“ is selected.

The group structure is a list of groups which can each have a list of children. When determining the largest free id, these children have to be taken into account as well.

- https://fixwrapnestedgroups.webknossos.xyz

### Steps to test:
- Upload an NML with nested groups, or one with un-connected trees within groups, like the one posted here https://scm.slack.com/archives/C5AKLAV0B/p1623331946023900?thread_ts=1623331839.023800&cid=C5AKLAV0B
- should upload fine, have stable group hierarchy
- should then download fine and re-upload fine.

### Issues:
- fixes #5495

------
- [x] Ready for review


youri-k (approved):
LGTM :shipit:

PR #5602, opened 2021-07-06
Add readable error messages for Project/TaskType unique checks

### LINKED ISSUE #5593: TaskType name availability is not checked, produces SQL Error
When creating a new tasktype, the check whether the name is still available for your orga is not executed properly. Only the database constraint catches this. A better UX would be to check this already from the frontend in the form

fm3:
Adds a readable error message for the tasktype summary unique check, and also improves the respective messages for projects.
I’d say this solves #5593 even though there is no front-end side check. This version should already help the user a lot more than the SQL error.

- https://uniquechecks.webknossos.xyz

### Steps to test:
- Create Project with new name, should work
- Create Project with name that is already in use, should give readable error message
- Create TaskType with new name, should work
- Create TaskType with name that is already in use, should give readable error message

### Issues:
- fixes #5593

------
- [x] Ready for review


youri-k (approved):
LGTM :shipit:

PR #5603, opened 2021-07-06
Send Session ID With Back-End Analytics

### LINKED ISSUE #5280: Session IDs for analytics
The back-end should store session IDs per user for analytics events. after a pause (no analytics events for this user for at least 30min) a new session id should be created. This ID should be appended to the analytics events.

fm3:
Use session_id field for backend analytics, as described in https://developers.amplitude.com/docs/http-api-v2

### Steps to test:
- set backendAnalytics.verboseLoggingEnabled = true
- set backendAnalytics.sessionPause down to 30 seconds
- create annotation, trace some, save, should all show same session id in logging
- wait for duration “sessionPause”
- save again, new session id should show in analytics logging

### Issues:
- fixes #5280 

------
- [x] Ready for review


youri-k (approved):
LGTM :shipit:

PR #5612, opened 2021-07-08
Reduce the mouse move delta threshold for plane controls

MichaelBuessemeyer:
This PR reduces the mouse move delta threshold. Thus if the user panes only small distances with the skeleton tool no new node is created.

- https://reducemousemovethreshold.webknossos.xyz

### Steps to test:
- Open a tracing in plane mode
- Select the skeleton tool
- Do some pretty small panning distances multiple times. Only if the distance was very small (below 6px I think) a node will be created otherwise not.

### Issues:
- fixes accidental creating upon small dragging steps with skeleton tool

------
- [x] Ready for review


daniel-wer (commented):
Is there a specific reason why you didn't simply reduce `MOUSE_MOVE_DELTA_THRESHOLD`? I can't think of a good reason why it should be that large for the arbitrary and 3d view :thinking:

MichaelBuessemeyer:
> Is there a specific reason why you didn't simply reduce MOUSE_MOVE_DELTA_THRESHOLD

Well, I thought that there was a good reason why the threshold was at 30 up until now. That's why I just changed it where required / needed. But I couldn't really come up with a reason why the value was so "high".

As you cannot think of a reason either, I'll just change the constant resulting in less changes and more consisten behaviour of the `InputMouse` :)

daniel-wer (approved):
> Well, I thought that there was a good reason why the threshold was at 30 up until now. That's why I just changed it where required / needed.

That's definitely a good approach :+1: I had a look at git blame and saw that this existed basically forever and was introduced without a clear reason (at least not for why it's that high) so I feel confident that it's alright to change it :)

PR #5616, opened 2021-07-12
Dataset Upload: Improve Progress Display

### LINKED ISSUE #5615: Dataset upload: say »processing«/»unpacking«/… instead of 99.9%
In case of a large dataset zip, the unpacking takes some time. This is shown to the user as an upload progress of 99.9%. The user may feel uncertain what is going on there and assume a failure.

fm3:
- While waiting for the finishUpload request (unpacks zip etc), don’t show 99.9% anymore, but instead 100% again, with added spinner and label »Processing uploaded files …«. I think the stalling at 99.9% for long-running finishUpload requests was confusing some users.
- Move the success toast below the sleep in the control flow, so that users do not wonder why there is another pause after “success”

Do you agree that these are sensible changes?

- https://uploadfinish.webknossos.xyz

### Steps to test:
- Upload a dataset (e.g. a few 1GB wkw zip like l4_sample)
- Upload page should not make you more impatient than necessary, specifically towards the end

### Issues:
- fixes #5615

------
- [x] Ready for review


frontend/javascripts/admin/dataset/dataset_upload_view.js
```diff
@@ -287,6 +292,7 @@ class DatasetUploadView extends React.Component<PropsWithFormAndRouter, State> {
             Toast.error(messages["dataset.upload_failed"]);
             this.setState({
               isUploading: false,
+              isFinishing: true,
```
philippotto, line 295:
```suggestion
              isFinishing: false,
```

probably? otherwise, the user might think that something will change if they wait. also `isFinishing` will already be true in that case, so it's a noop currently.

fm3: Absolutely right. Didn’t show during testing since with isUploading being false, none of the uploading UI is rendered anyway. But this is certainly the cleaner state

frontend/javascripts/admin/dataset/dataset_upload_view.js
```diff
@@ -340,16 +346,14 @@ class DatasetUploadView extends React.Component<PropsWithFormAndRouter, State> {
...
-            // Round to 1 digit after the comma. Don't show 100%, since there is
-            // some additional delay until the UI shows the next screen.
-            // Otherwise, the user might think that nothing will happen after 100%
-            // was reached.
-            percent={Math.min(99.9, Math.round(uploadProgress * 1000) / 10)}
+            // Round to 1 digit after the comma.
+            percent={Math.round(uploadProgress * 1000) / 10}
             status="active"
           />
+          {isFinishing ? <Spin tip="Processing uploaded files …" /> : null}
```
philippotto, line 356: When `isFinishing` is true, there is a progress bar **and** a spinner. I would only show one of them. One of the following options probably:
(1) Always show the progress bar. due to `status="active"`, the progress bar will have a slight animation even when 100% is reached. below the progress bar, "Processing uploaded files" would still be shown (so only the spinner would be removed).
(2) Hide the progress bar and show the spinner instead when `isFinishing` becomes true.

I'd do (1). What do you think?

philippotto (approved):
Just tested it and it feels/works good/well :) I only have added a visual suggestion which might be worthwhile.

frontend/javascripts/admin/dataset/dataset_upload_view.js
fm3: I tried (1) originally, but I feld like just adding a bit of text didn’t convey the change strongly enough. I’d actually vote for showing both, but I don’t have a strong opinion on this.

philippotto: ok, then let's go with your approach :)

MichaelBuessemeyer: Thx for jumping in @philippotto. Sorry I am currently having pretty busy days.

philippotto: no worries!

PR #5622, opened 2021-07-15
Poll jobs in meshes tab only if jobs are enabled

fm3:
Since the job request can only be sent (and served) if jobs are enabled on that webKnossos instance, the polling is unnecessary otherwise as well. 
It creates a lot of error logging in the back-end if jobs are not enabled, this PR gets rid of that

- [x] Ready for review


philippotto (approved):
good catch!

PR #5626, opened 2021-07-19
Call wk-worker nuclei inferral job

### LINKED ISSUE #5555: Build job for nuclei inferral
## Detailed Description
The existing wk job infrastructure should be used to generate a nuclei segmentation of an input (color) layer. The inferral should use a pretrained model using voxelytics. This feature can be tackled quite similarly to the existing mesh-job (which also uses the wk job infrastructure and also makes use of voxelytics).
Contact @georgwiese for insights about which model (or vx workflow) should be used for this.

## Context
The goal of this issue is to provide a generic inferral solution to users for the specific use case of detecting nuclei.

## Todos
- [x] build task in webknossos-worker
- [ ] back-end glue code for the new job type
- [ ] build rudimentary UI for this feature
  - somewhere in the UI, there should be a button to kick off this nuclei inferral (could be in the dashboard, but could also be in the core dataset view; could be both) → moved to follow-up issue #5631

fm3:
- Adds a route that calls the webknossos-worker infer_nuclei job if jobsEnabled
- Shows the result of this job correctly in the job list
- Note that this PR does not yet include UI to trigger this route, that will be follow-up issue https://github.com/scalableminds/webknossos/issues/5631

### Steps to test:
- set up wk-worker locally
- visit http://localhost:9000/api/jobs/run/inferNuclei/sample_organization/l4_sample?layerName=color (exchange dataset name if necessary)
- job should show in job list
- result should be linked in job list after job is successful

### Issues:
- fixes #5555

------
- [x] Ready for review


youri-k (approved):
LGTM :shipit:

PR #5632, opened 2021-07-22
Select mag of segmentation layer for meshfile computation 

### LINKED ISSUE #5629: Compute Mesh File: Select only mags that the segmentation layer has
The front-end selected mag 4-4-1 for meshfile generation for my dataset, even though the segmentation layer only had 16-16-4

`getDatasetResolutionInfo` is used to list the available resolutions.  @philippotto Do you know if there is a good way to access only the (active) segmentation layer’s resolutions?

- [x] specific to long-running jobs ( jobsEnabled=true )

fm3:
### Steps to test:
- enable jobs
- on dataset that has more layers in color than in segmentation layer, run meshfile job
- should select resolution present in segmentation layer

### Issues:
- fixes #5629

------
- [x] Ready for review

### DIFF
--- a/CHANGELOG.unreleased.md
+++ b/CHANGELOG.unreleased.md
@@ -11,7 +11,7 @@ For upgrade instructions, please check the [migration guide](MIGRATIONS.released
 [Commits](https://github.com/scalableminds/webknossos/compare/21.07.0...HEAD)
 
 ### Added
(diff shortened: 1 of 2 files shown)


philippotto (approved):
Wonderful :sparkles:

PR #5634, opened 2021-07-23
Add mesh support to frontend api

### LINKED ISSUE #5578: Add mesh loading to frontend API
## Detailed Description
Expose a function in the Frontend API to trigger mesh loading.

## Context
For user-scripts it would be nice to be able to load precomputed meshes or on-demand isosurfaces programmatically. For example, a merger-mode-like script could display the meshes of all parts of the collected segment in 3D.

philippotto:
- https://meshapi.webknossos.xyz

### Steps to test:
- I ran the following code in the console:
```
webknossos.apiReady().then(async api => { window.api = api })

(async () => {

	const currentPosition = api.tracing.getCameraPosition()

	const segmentId = await api.data.getDataValue("segmentation", currentPosition);
	const availableMeshFiles = await api.data.getAvailableMeshFiles();
	api.data.setActiveMeshFile(availableMeshFiles[0])

	await api.data.loadPrecomputedMesh(segmentId, currentPosition);

	// api.data.setMeshVisibility(segmentId, false);
	// api.data.removeMesh(segmentId);
})()
```

### Issues:
- fixes #5578 

------
- [x] Ready for review


frontend/javascripts/oxalis/api/cross_origin_api.js
```diff
@@ -45,6 +45,23 @@ const onMessage = event => {
...
+      const segmentId = args[0];
+      const isVisible = args[1];
+      api.data.setMeshVisibility(segmentId, isVisible);
+      break;
+    }
+    case "removeMesh": {
+      const segmentId = args[0];
+      api.data.removeMesh(segmentId);
+      break;
+    }
```
philippotto, line 64: I didn't test this part, yet. I think these are the functions which are necessary for the connectome viewer, but I'm not sure. The other functions would also need to return something, but I think we haven't done something like that before :thinking: Maybe we can discuss this in person.

daniel-wer (commented):
Looking good, please see my comments and let's talk in person about the cross-origin API.

CHANGELOG.unreleased.md
```diff
@@ -12,6 +12,7 @@ For upgrade instructions, please check the [migration guide](MIGRATIONS.released
 
 ### Added
 - Fixed a bug where non-existing resolutions could be selected for wk-worker-based meshfile computations [#5631](https://github.com/scalableminds/webknossos/pull/5631)
+- Added new mesh-related functions to the front-end API: getAvailableMeshFiles, getActiveMeshFile, setActiveMeshFile, async, computeMeshOnDemand, setMeshVisibility, removeMesh. [#5634](https://github.com/scalableminds/webknossos/pull/5634)
```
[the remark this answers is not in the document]
philippotto: whoops :see_no_evil:

frontend/javascripts/oxalis/api/cross_origin_api.js
philippotto: > Wouldn't a mesh file need to be activated, first, or is that done automatically?

If there is only one mesh file, this is done automatically, but this needs to be awaited to be robust. Therefore, I also added the corresponding functions now.

> Ideally, we'd implement the connectome viewer use case at the same time to check whether everything is included.

Good point. I opened https://github.com/scalableminds/connectome-viewer/pull/47 :)

daniel-wer (approved):
LGTM, the `getSomeNodePositionByTreeName` is a little specific but I see why it is needed. Maybe at some point, meshes loaded from a mesh file will have a position (center of mass inside the segment or so) which can be used as the seed position and this function will no longer be needed.

PR #5635, opened 2021-07-23
Refactor RPC Class/Variable Naming

### LINKED ISSUE #5275: Clean up class names for RPC Clients / Controllers
The wK backend has some classes dealing with RPCs from wK to datastore/tracingstore and back. We should unify their naming. A client is a class that sends requests, a controller is a class that handles and/or answers them.

The exact names are not clear yet, let’s give this some thought. Possible ideas were RemoteWebKnossosClient, RemoteDataStoreController etc, but I’m not sure yet that this is good

fm3:
- Extracted the wk master token to separate object RpcTokenHolder
- Renamed some variables and classes concerning RPCs between webKnossos and the datastore + tracingstore modules
- New naming scheme examples: “DSRemoteWebKnossosClient” lives in the datastore, and sends RPC calls to webKnossos. “WKRemoteDataStoreController” lives in webKnossos and handles RPC requests sent by the data store.
- The variable name `webKnossosServer` was ditched, I’d rather consider the client as such, rather than think of it as a wrapper around the remote server.
- The naming scheme is of course open for debate, I’m looking forward to your opinions :)

- https://refactorrpcnames.webknossos.xyz

### Steps to test:
- Should still be able to create/save annotation, upload/download NML, edit dataset settings (e.g. bbox)

### Issues:
- fixes #5275 

----
- [x] Ready for review


youri-k (approved):
LGTM :shipit:

PR #5637, opened 2021-07-26
Improve context menu and segment id handling

philippotto:
- always read the segment id from fallback data when necessary (also respect active mapping while doing so)
- avoid empty context menus (these weren't opened previously, but this is also confusing when one expects a menu) by always showing "copy position" as a fallback.

- https://readfallbacksegidforcontext.webknossos.xyz

### Steps to test:
- open a dataset with segmentation layer in view mode and rightclick somewhere (disable classic controls): the menu should show an entry to copy the clicked position
- load a dataset with a segmentation layer (should have multiple mags)
  - right click to copy a segment id before mag 1 was loaded --> the context menu should include an item to copy the segment id. before this pr, that entry only appeared when the segment id of the active mag was already loaded (i.e., fallback rendering was disabled)
- repeat the last step, but with the cell picker tool. ensure that a segment id can be picked even when the best mag wasn't fully loaded yet

### Issues:
- follow up to #5231 

------
- [X] Ready for review


daniel-wer (approved):
LGTM :+1:

PR #5638, opened 2021-07-27
Fix condition for active segment and node id in status bar

philippotto:
- https://___.webknossos.xyz

### Steps to test:
- open skeleton-only annotation and check that active node id is displayed in status bar
- open volume-only annotation and check that active segment id is displayed in status bar

### Issues:
- follow up to #5231 

------
- [X] Ready for review


fm3 (approved):
LGTM

PR #5645, opened 2021-07-30
Allow resolution restrictions for explorative annotations

### LINKED ISSUE #5067: Relax volume annotation limits if mag1 is not present
Some tools are disabled in too-high magnifications to avoid generating huge amounts of data in mag1.
If mag1 is not present at all, those strict checks are not really required. If mag2 *also* is not present, even higher mags could be annotated.

### LINKED ISSUE #5148: Allow to select target mag for volume annotations
## Detailed Description
Allow to choose a base mag for a volume annotation when
- creating tasks
- creating an explorational annotation

## Context

When creating a volume annotation (either as explorational or via task), the base mag in which the new data is annotated will be the lowest (best quality) mag which exists in the segmentation layer. If no segmentation layer exists, this defaults to mag 1.
However, annotating large processes is usually done best in a higher mag. Therefore, it should be possible to choose which mag should be used as a base.

A current workaround for this is to create a new dataset and symlink the base mag as mag 1 in the segmentation layer.

## Approach/Implementation

If we simultaneously restrict which magnifications are visible, the implementation in the front-end should be straightforward, since we could simply adapt the magnifications of the segmentation layer as if only the selected base mag (plus descendents) exist.
The back-end would need to communicate the selected base mag and we would need UI for specifying the base mag for task creation and explorational creation.

## Related issues

The volume tool restrictions would need to be loosened too then: https://github.com/scalableminds/webknossos/issues/5067

fm3:
- Adds the option to restrict resolutions for explorational volume (and hybrid) annotations
- This can be used to annotate large structures in downsampled mags without generating large amounts of mag1 data
- Replaces the drop-down in the dashboard by this modal:
image

- https://resolutionrestrictionsexplorational.webknossos.xyz

### TODO
 - [x] create modal in front-end to select the resolution restrictions when creating new annotations
  - [x] select annotation type
  - [x] select fallback layer
  - [x] select resolution restrictions with slider
  - [x] add explaining tooltips
  - [x] fix flow errors
  - [x] Crop resolution slider to fallback layer resolutions? Backend does this, does the UI need to mirror this?
  - [x] disable resolution restrictions if skeleton-only is selected
 - [x] relax other limits designed to limit data amount if not needed (#5067) (volumetracing_accessor.js)

### Possible Follow-Up TODOs 
  - [x] Disable resolution restrictions if only one resolution is present?
  - [ ] Add more tooltips explaining each each disabled state case for the elements?
  - [ ] Use slider UI also in task type creation?

### Steps to Test:
 - create annotation directly with the New Annotation button, resolutions of volume should mirror those of the fallback layer or (if there is none) of the union of the dataset layers
 - create annotation via the new modal (dotted menu in dashboard dataset list)
  -  with varying annotation type
  -  with varying fallback layer
  -  with varying resolution restrictions
  - for different datasets (segmentation layer present or not, different resolution combinations)
 - No unexpected behavior should crop up.

### Issues:
- fixes #5148
- fixes #5067

------
- [x] Ready for review

### DIFF
--- a/frontend/javascripts/dashboard/advanced_dataset/create_explorative_modal.js
+++ b/frontend/javascripts/dashboard/advanced_dataset/create_explorative_modal.js
@@ -75,20 +75,12 @@ const CreateExplorativeModal = ({ datasetId, onClose }: Props) => {
     const highResolutionIndex = Math.min(highestResolutionIndex, userDefinedResolutionIndices[1]);
     const lowResolutionIndex = Math.max(lowestResolutionIndex, userDefinedResolutionIndices[0]);
 
-    modalContent = (
+    const resolutionSlider = (
       <React.Fragment>
-        <div style={{ marginBottom: 16 }}>
-          <Radio.Group onChange={e => setAnnotationType(e.target.value)} value={annotationType}>
-            <Radio value="hybrid">Skeleton and Volume</Radio>
-            <Radio value="skeleton">Skeleton only</Radio>
-            <Radio value="volume">Volume only</Radio>
-          </Radio.Group>
-        </div>
-        {doesSupportVolumeWithFallback(dataset) ? fallbackCheckbox : null}
         <h5 style={{ marginBottom: 0 }}>
           Volume Resolutions{" "}
           <Tooltip
-            title="Select which of the dataset resolutions the volume data should be created at. For example, resolution 1 is the most detailed, 4-4-2 is downsampled by factor 4 in x and y, and by factor 2 in z."
+            title="Select which of the dataset resolutions the volume data should be created at. Restricting the available resolutions can greatly improve the performance when annotating large structures, such as nuclei, since the volume data does not need to be stored in all quality levels. How to read: Resolution 1 is the most detailed, 4-4-2 is downsampled by factor 4 in x and y, and by factor 2 in z."
             placement="right"
           >
             <InfoCircleOutlined />
@@ -121,20 +113,36 @@ const CreateExplorativeModal = ({ datasetId, onClose }: Props) => {
             {datasetResolutionInfo.getResolutionByIndexOrThrow(highResolutionIndex).join("-")}
           </div>
         </div>
-        <Link
-          to={`/datasets/${dataset.owningOrganization}/${
-            dataset.name
-          }/createExplorative/${annotationType}/${isFallbackSegmentationSelectedString}?minRes=${Math.max(
-            ...datasetResolutionInfo.getResolutionByIndexOrThrow(lowResolutionIndex),
-          )}&maxRes=${Math.max(
-            ...datasetResolutionInfo.getResolutionByIndexOrThrow(highResolutionIndex),
-          )}`}
-          title="Create new annotation with selected properties"
-        >
-          <Button size="large" type="primary">
-            Create Annotation
-          </Button>
-        </Link>
+      </React.Fragment>
+    );
+
+    modalContent = (
+      <React.Fragment>
+        <div style={{ marginBottom: 16 }}>
+          <Radio.Group onChange={e => setAnnotationType(e.target.value)} value={annotationType}>
+            <Radio value="hybrid">Skeleton and Volume</Radio>
+            <Radio value="skeleton">Skeleton only</Radio>
+            <Radio value="volume">Volume only</Radio>
+          </Radio.Group>
+        </div>
+        {doesSupportVolumeWithFallback(dataset) ? fallbackCheckbox : null}
+        {lowestResolutionIndex < highestResolutionIndex ? resolutionSlider : null}
+        <div style={{ textAlign: "right" }}>
+          <Link
+            to={`/datasets/${dataset.owningOrganization}/${
+              dataset.name
+            }/createExplorative/${annotationType}/${isFallbackSegmentationSelectedString}?minRes=${Math.max(
+              ...datasetResolutionInfo.getResolutionByIndexOrThrow(lowResolutionIndex),
+            )}&maxRes=${Math.max(
+              ...datasetResolutionInfo.getResolutionByIndexOrThrow(highResolutionIndex),
+            )}`}
+            title="Create new annotation with selected properties"
+          >
+            <Button size="large" type="primary">
+              Create Annotation
+            </Button>
+          </Link>
+        </div>
       </React.Fragment>
     );
   }
--- a/frontend/javascripts/dashboard/advanced_dataset/create_explorative_modal.js
+++ b/frontend/javascripts/dashboard/advanced_dataset/create_explorative_modal.js
@@ -11,6 +11,7 @@ import {
   getSegmentationLayer,
   doesSupportVolumeWithFallback,
   getDatasetResolutionInfo,
+  getResolutionInfoOfSegmentationLayer,
 } from "oxalis/model/accessors/dataset_accessor";
 
 type Props = {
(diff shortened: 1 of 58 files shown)


frontend/javascripts/dashboard/advanced_dataset/create_explorative_modal.js
```diff
@@ -0,0 +1,149 @@
...
+        <br />
+        <br />
+        <Link
+          to={`/datasets/${dataset.owningOrganization}/${
+            dataset.name
+          }/createExplorative/${annotationType}/${isFallbackSegmentationSelectedString}?minRes=${Math.max(
+            ...datasetResolutionInfo.getResolutionByIndexOrThrow(lowResolutionIndex),
+          )}&maxRes=${Math.max(
+            ...datasetResolutionInfo.getResolutionByIndexOrThrow(highResolutionIndex),
+          )}`}
```
philippotto, line 125: I'm wondering whether there can be a scenario where minRes/maxRes should not be passed? I cannot think of any realistic scenarios, though. Maybe you can give it a thought, too, if you haven't already.

frontend/javascripts/dashboard/advanced_dataset/create_explorative_modal.js
```diff
@@ -0,0 +1,149 @@
...
+  const [annotationType, setAnnotationType] = useState("hybrid");
+  const [userDefinedWithFallback, setUserDefinedWithFallback] = useState(true);
+  const [userDefinedResolutionIndices, setUserDefinedResolutionIndices] = useState([0, 10000]);
+
+  let modalContent = <Spin />;
+
+  if (dataset !== null) {
+    const segmentationLayer = getSegmentationLayer(dataset);
+
+    const datasetResolutionInfo = getDatasetResolutionInfo(dataset);
```
philippotto, line 32: I think, this needs to be specific to the segmentation layer's resolutions (if one exists). For example, consider this dataset:

```
color mags: 1, 2, 4, 8
segmentation mags: 4, 8
```

the user should only be able to select a range from mag 4 to 8. however, `getDatasetResolutionInfo` will return 1 to 8.

philippotto (changes requested):
Wow! Really great stuff :+1: I think, you tamed the complexity quite well code-wise and the UX looks also very solid :1st_place_medal: I left a couple of comments of which most are of cosmetic nature (feel free to ignore if you disagree with some). However, my comment about getting the resolution info for the segmentation layer is probably important to tackle before merging.

I'll test after the next iteration :)

frontend/javascripts/dashboard/advanced_dataset/create_explorative_modal.js
fm3: The trouble here is that this depends on what the user has selected above. What you say is true in case of a fallback layer. I think it may be confusing to change the available range depending on the selections above. This may change the _selected_ range, which yields the question if that should be undone when the above selection is undone.
With the current code, the back-end will still select only those mags that are present in the fallback layer (intersection of available mags and selected range).
This is one of the check-boxes I added as “follow-up” in the PR description. If you think the current behavior is the more confusing one (mags that were selected can become unavailable in the end result), I’ll change it to the switching variant.

frontend/javascripts/dashboard/advanced_dataset/create_explorative_modal.js
fm3: I think it’s fine to pass it always. Selecting the full range will yield the same end result as passing nothing.

frontend/javascripts/dashboard/advanced_dataset/create_explorative_modal.js
philippotto: > This is one of the check-boxes I added as “follow-up” in the PR description. 

Ah, I missed that :+1: Thanks for explaining the rationale of the current approach.

> I think it may be confusing to change the available range depending on the selections above. This may change the selected range, which yields the question if that should be undone when the above selection is undone.

I understand the problem, but think that the actual negative impact is negligible. The happy (and imo also likely) workflow will be that the user decides on fallback/no-fallback first and afterwards (if at all) changes the resolution restrictions. If they go back and fiddle with the fallback property again, the automated change of the resolutions can be annoying, but at least the UI reflects that and thus it's fixable.
The downside of the current approach is that the user selects something in the modal, hits "create" and then gets an annotation which does not reflect the settings which were specified. I think, this mismatch can be very confusing and annoying (probably, the user will think they made a mistake and retry). I think, it would be better to avoid this.

> which yields the question if that should be undone when the above selection is undone.

I don't think, that this is necessary.

Btw, the other "potential follow-ups" don't need to be in this PR imo, but the above thing would be good to have here.

Do you agree? Happy to change my mind about this :)

frontend/javascripts/router.js
```diff
@@ -488,7 +488,22 @@ class ReactRouter extends React.Component<Props> {
...
-                      const annotation = await createExplorational(dataset, type, withFallback);
+
+                      const params = Utils.getUrlParamsObjectFromString(location.search);
+
+                      const resolutionRestrictions = {};
+                      if (params.minRes !== undefined) {
+                        resolutionRestrictions.min = parseInt(params.minRes);
+                      }
+                      if (params.maxRes !== undefined) {
+                        resolutionRestrictions.max = parseInt(params.maxRes);
```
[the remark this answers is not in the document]
fm3: Looks like the frontend sends null with the current code if something that is not parseable as int is passed. The backend interprets this as unlimited maxRes. Do you think this is acceptable?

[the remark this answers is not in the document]
philippotto: Hm, throwing an error here would be better in my opinion. Otherwise, potential errors (e.g., when constructing the URL) are swallowed silently. You can use something like `if (!_.isNumber(resolutionRestrictions.min)) { throw new Error("Invalid minRes parameter") } .

frontend/javascripts/dashboard/advanced_dataset/create_explorative_modal.js
fm3: Sounds fair :+1:

fm3:
@philippotto Thanks for your detailed feedback! I think I implemented all of your suggestions. Could you have another look? :)

philippotto (changes requested):
The code looks great and testing was also very pleasant :+1: During testing, I got a couple of thoughts, though:

1.
> Disable resolution restrictions if only one resolution is present?

I noticed this during testing, too. I think, the UI would feel more professional if we handled this additionally. See point 2 which has some overlap.

2. I'm wondering whether we should always show the resolution slider by default. I imagine it to be hard to understand at first and most of the time, the user won't need to touch it, anyway. What do you think about having an additional checkbox with "Restrict available resolutions for volume" which is disabled by default (in that case, the slider is not visible). Only if the user actively decides to enable the restriction, the slider appears. For the case where only one relevant resolution exists (point 1), we could make the checkbox read-only.

3. The tooltip for "Volume Resolutions" doesn't mention why one would want to use this feature. How about adding something like "Restricting the available resolutions can greatly improve the performance when annotating large structures, such as nuclei, since the volume data does not need to be stored in all quality levels."

4. The other modals usually place the confirm-buttons in the bottom right. See picture:
image For consistency, we should do this here, too, I think.

What do you think?

fm3:
Thanks for your input! I’m on board with all of your points, except number 2: I’d argue that this modal is already visible only when a user wants to go beyond the default case. I think there is little benefit in hiding the slider behind an additional click then. Do you agree?

philippotto:
> Thanks for your input! I’m on board with all of your points, except number 2: I’d argue that this modal is already visible only when a user wants to go beyond the default case. I think there is little benefit in hiding the slider behind an additional click then. Do you agree?

Yes, sounds reasonable :+1:

youri-k (approved):
Backend LGTM :tada:

fm3:
@philippotto I incorporated the remaining feedback, feel free to have another look :) I still think it’s fairly noisy that depending on the above settings the UI elements below vanish or reappear, but I can live with it and I also see the benefit of not wondering about those options if you never have datasets they apply for.

philippotto (approved):
Works great for me :) I think, this will be a big improvement :+1:

Issue #5625, opened 2021-07-19
Disable job triggers for datasets that are not reachable for wk-worker


fm3:
The webknossos-worker (features.jobsEnabled=true) cannot always access the file system of all datastores. For datasets hosted elsewhere, the job trigger UIs should be disabled.

fm3:
Also, disable job triggers that are not applicable for the open dataset. E.g. don’t trigger meshfile computation for a volume annotation (only for on-disk segmentation layers)

PR #5648, opened 2021-08-02
Disable compute-meshfile-button for volume annotations and datasets without segmentation

philippotto:
- https://___.webknossos.xyz

### Steps to test:
- open a dataset with the following properties (with enabled jobs):
  - no segmentation layer / view mode
  - no segmentation layer / with volume annotation
  - segmentation layer / view mode
  - segmentation layer / with volume annotation
  - one of the above, but without enabled jobs
- ensure that the "compute mesh file" button is enabled/disabled correctly (including tooltip)

### Issues:
- contributes to #5625

------
- [X] Ready for review

### DIFF
--- a/frontend/javascripts/oxalis/view/right-border-tabs/meshes_view.js
+++ b/frontend/javascripts/oxalis/view/right-border-tabs/meshes_view.js
@@ -84,6 +84,7 @@ const mapStateToProps = (state: OxalisState): * => ({
   mappingColors: state.temporaryConfiguration.activeMapping.mappingColors,
   flycam: state.flycam,
   activeCellId: state.tracing.volume ? state.tracing.volume.activeCellId : null,
+  hasVolume: state.tracing.volume != null,
   segmentationLayer: getSegmentationLayer(state.dataset),
   zoomStep: getRequestLogZoomStep(state),
   allowUpdate: state.tracing.restrictions.allowUpdate,
@@ -277,29 +278,43 @@ class MeshesView extends React.Component<Props, State> {
       }
     };
 
-    const getComputeMeshfileTooltip = node => (
-      <Tooltip
-        title={
-          features().jobsEnabled
-            ? "Compute a mesh file for this dataset so that meshes for segments can be loaded quickly afterwards."
-            : "The computation of mesh files is not supported by this webKnossos instance."
-        }
-      >
-        {node}
-      </Tooltip>
-    );
+    const getComputeMeshFileTooltipInfo = () => {
+      let title = "";
+      let disabled = true;
+      if (!features().jobsEnabled) {
+        title = "Computation jobs are not enabled for this webKnossos instance.";
+      } else if (this.props.hasVolume) {
+        title =
+          "Meshes cannot be computed for volume annotations. Please open this dataset in view mode to compute a mesh file";
+      } else if (this.props.segmentationLayer == null) {
+        title = "There is no segmentation layer for which a mesh file could be computed.";
+      } else {
+        title =
+          "Compute a mesh file for this dataset so that meshes for segments can be loaded quickly afterwards.";
+        disabled = false;
+      }
 
-    const getComputeMeshFileButton = () =>
-      getComputeMeshfileTooltip(
-        <Button
-          size="small"
-          loading={this.state.activeMeshJobId != null}
-          onClick={startComputingMeshfile}
-          disabled={!features().jobsEnabled}
-        >
-          Compute Mesh File
-        </Button>,
+      return {
+        disabled,
+        title,
+      };
+    };
+
+    const getComputeMeshFileButton = () => {
+      const { disabled, title } = getComputeMeshFileTooltipInfo();
+      return (
+        <Tooltip key="tooltip" title={title}>
+          <Button
+            size="small"
+            loading={this.state.activeMeshJobId != null}
+            onClick={startComputingMeshfile}
+            disabled={disabled}
+          >
+            Compute Mesh File
+          </Button>
+        </Tooltip>
       );
+    };
 
     const getDownloadButton = (segmentId: number) => (
       <Tooltip title="Download Isosurface">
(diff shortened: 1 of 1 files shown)


frontend/javascripts/oxalis/view/right-border-tabs/meshes_view.js
```diff
@@ -277,29 +278,43 @@ class MeshesView extends React.Component<Props, State> {
...
-      </Tooltip>
-    );
+    const getComputeMeshFileTooltipInfo = () => {
+      let title = "";
+      let disabled = true;
+      if (!features().jobsEnabled) {
+        title = "Computation jobs are not enabled for this webKnossos instance.";
+      } else if (this.props.hasVolume) {
+        title =
+          "Meshes cannot be computed for volume annotations. Please open this dataset in view mode to compute a mesh file";
```
hotzenklotz, line 288: Why should `view` mode help here? `View` mode will not convert the volume annotation into a segmentation, right?

Also, if the toast supports HTML links, this should be a clickable link.

frontend/javascripts/oxalis/view/right-border-tabs/meshes_view.js
```diff
@@ -458,14 +473,17 @@ class MeshesView extends React.Component<Props, State> {
...
-        </Menu.Item>
-        <Menu.Item>{getStlImportItem()}</Menu.Item>
-      </Menu>
-    );
+    const getHeaderDropdownMenu = () => {
+      const { disabled, title } = getComputeMeshFileTooltipInfo();
+      return (
+        <Menu>
+          <Menu.Item onClick={startComputingMeshfile} disabled={disabled}>
+            <Tooltip title={title}>Compute Mesh File</Tooltip>
```
hotzenklotz, line 481: I am also not a big fan of the button title. What is a `mesh file`? This is very technical. How about something more descriptive. "Computes Mesh for (whole) dataset" or "Pre-compute meshes for all segments" or "Compute all meshes". Or any such combination.

 I find it hard to find the difference between the button "Compute Mesh" and "Compute Mesh File". Likely the other button should be called something like "Compute for Mesh for (active) Segment" or "Compute single mesh" or "compute temporay mesh" ...

hotzenklotz: @normanrz Any comments?

frontend/javascripts/oxalis/view/right-border-tabs/meshes_view.js
```diff
@@ -277,29 +278,43 @@ class MeshesView extends React.Component<Props, State> {
...
-          Compute Mesh File
-        </Button>,
+      return {
+        disabled,
+        title,
+      };
+    };
+
+    const getComputeMeshFileButton = () => {
+      const { disabled, title } = getComputeMeshFileTooltipInfo();
```
hotzenklotz, line 304: Do theses till tip infos apply to the "Compute Mesh" button as well? I think they do, right? A single mesh also needs a segmentation layer and won't work with volume annotations.

frontend/javascripts/oxalis/view/right-border-tabs/meshes_view.js
normanrz: I don't have a strong opinion re `mesh file`.
I agree that meshes vs isosurfaces is quite confusing. They are essentially the same but have different names.

frontend/javascripts/oxalis/view/right-border-tabs/meshes_view.js
philippotto: > Why should view mode help here? View mode will not convert the volume annotation into a segmentation, right?

I changed the tooltip to `Meshes cannot be precomputed for volume annotations. However, you can open this dataset in view mode to precompute meshes for the dataset's segmentation layer.`. I hope this makes it clearer?

> Also, if the toast supports HTML links, this should be a clickable link.

It's a tooltip, so, no clickable link possible..

hotzenklotz: This assumes that there is a segmentation layer to begin with in `view` mode. Depending on your dataset this is not the case. I am not sure if this already considered for this conditional. If so, than I guess its fine.

frontend/javascripts/oxalis/view/right-border-tabs/meshes_view.js
philippotto: I changed the button labels to:
- Precompute Meshes (tooltip: Precompute meshes for all segments of this dataset so that meshes for segments can be loaded quickly afterwards from a mesh file.)
- Load Precomputed Mesh
- Compute Mesh (ad hoc)

I hope this makes it clearer. I also hide the "Compute Mesh (ad hoc)" button automatically, if a mesh file is already available (since there is no benefit in using the ad-hoc part).

> I agree that meshes vs isosurfaces is quite confusing. They are essentially the same but have different names.

I went through the code and changed the last user-facing "isosurface" bits to "mesh", but there were quite rare (mainly, in the download/reload buttons of the mesh list). I hope I caught them all.

frontend/javascripts/oxalis/view/right-border-tabs/meshes_view.js
philippotto: Now, the ad-hoc mesh button also has a tooltip. It needed slightly different code (e.g., it doesn't depend on `jobsEnabled`) and wording.

frontend/javascripts/oxalis/view/right-border-tabs/meshes_view.js
philippotto: Yeah, this case is already handled.

philippotto:
> On a more high-level note: Do we really want to show these button if the can not do anything, i.e. the "error" toasts? Alternatively, we could not show the buttons at all and have the "error"/help texts directly in the segmentation tab/pane. Not sure how I feel about that though...

This might feel better, but I think, the disabled buttons convey quite well what the user can expect if the conditions are met. If we decide to change this, we should do this in a different PR and focus on the polish. The current PR was only meant as a hotfix for scenarios where meshes are simply not supported.

philippotto:
@daniel-wer feel free to review now :) I'm sorry that the PR got quite big. The last commit (https://github.com/scalableminds/webknossos/pull/5648/commits/26cedef927d0e2fded3c090a57e168a8ba87bd50) is the major contributor to this. In that commit, I extracted some code from the huge `render` method to make it a bit simpler. You could review the commit individually to get a better diff experience.

daniel-wer:
@philippotto Could you annotate your testing instructions with when the button should be enabled or disabled? This would help me to review and test this PR :)

philippotto:
Sure, just updated it!

daniel-wer (approved):
Works exactly as described and the tooltip explanations are very helpful and descriptive :+1:

frontend/javascripts/oxalis/view/right-border-tabs/meshes_view.js
fm3: @hotzenklotz @philippotto I am a bit confused here, should the ad-hoc isosurface computation really be disabled for volume annotations? There exists the `/volume/:tracingId/isosurface` request in the tracingstore, which is now inaccessible?

philippotto: Oh, good catch. I'll open a PR to fix this.

PR #5658, opened 2021-08-06
Fix copy-buttons in context menu and change out-of-context-menu-click behavior

### LINKED ISSUE #5606: Disable viewport click interactions while context menu is open
When I want to close the viewport context menu without selecting one of its options, I click on an empty area near it. This currently creates a node / brushes at that spot. I’d argue that this click should instead _only_ close the context menu, and only after that, the viewport click interactions should be enabled again. That would mirror the behavior of the browser context menu

### LINKED ISSUE #5657: copy from contextmenu doesn't work: can't copy segment-id or position
## Context
I wanted to copy a segment ID from the context menu while browsing https://connectome.webknossos.org/scalable_minds/l4dense_motta_et_al_demo_v2/
Also tried it directly on webknossos.org with the same result, to make sure it's not a problem due to the iFrame.

## Expected Behavior
I should have the segment id in the clipboard. (same for the position)

## Current Behavior
The old content is still in the clipboard.

## Steps to Reproduce the bug
1. Open https://webknossos.org/annotations/Explorational/61012a4b010000aa00e17c3e?sandbox&token=OQnuk6RxaakDH3zmUYHQXA#0,0,0,0,0.826
2. Make sure the new UX is available.
3. Right-click on a position in a 2D-Viewport.
4. Click the copy-icon for the segment-id (or position).
5. The content is not copied :crying_cat_face: 

recording (5)

## Your Environment for bug
- Browser name and version: 
Google Chrome | 92.0.4515.107 (Official Build) (64-bit)
- Operating System and version: Debian 10.10
- Version of WebKnossos (Release or Commit): 15239
  (tested on wk.org, but probably not related to it specifically)

philippotto:
With this PR:
- the copy buttons in the context menu are fixed (before the PR, the `blur()` event was triggered directly and the menu was closed immediately)
- clicking outside of the context menu closes it (and doesn't do anything else, such as creating a node)
- the only downside: a repeated rightclick doesn't open the context menu again. technically, this should be possible, but the effort to implement this is several times higher than the current pr. Also, I think, it's not that important (right now, at least).

- https://___.webknossos.xyz

### Steps to test:
- open a tracing (disable classic mode)
- open a context menu via right click
- use the copy buttons (e.g., copy position) and ensure that the position is copied.
- the context menu will stay open (I find that okay)
- left click outside of the context menu --> the menu should close, but no node or similar should be created
- right click outside of the context menu --> the menu should close and nothing else will happen. ideally, the context menu would open again, but as explained above, this "feature" got dropped with this PR for now.

### Issues:
- fixes #5606
- fixes #5657

------
- [X] Ready for review

### DIFF
--- a/frontend/javascripts/oxalis/view/context_menu.js
+++ b/frontend/javascripts/oxalis/view/context_menu.js
@@ -445,6 +445,14 @@ function ContextMenu(props: Props) {
   return (
     <React.Fragment>
       <div
+        className="node-context-menu-overlay"
+        onClick={hideContextMenu}
+        onContextMenu={evt => {
+          evt.preventDefault();
+          hideContextMenu();
+        }}
+      />
+      <div
         style={{
           position: "absolute",
           left: contextMenuPosition[0],
@@ -452,7 +460,6 @@ function ContextMenu(props: Props) {
         }}
         className="node-context-menu"
         tabIndex={-1}
-        onBlur={hideContextMenu}
         ref={inputRef}
       >
         <Shortcut supportInputElements keys="escape" onTrigger={hideContextMenu} />
--- a/frontend/javascripts/oxalis/view/context_menu.js
+++ b/frontend/javascripts/oxalis/view/context_menu.js
@@ -454,7 +454,6 @@ function ContextMenu(props: Props) {
       />
       <div
         style={{
-          position: "absolute",
           left: contextMenuPosition[0],
           top: contextMenuPosition[1],
         }}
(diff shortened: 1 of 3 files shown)


frontend/javascripts/oxalis/view/context_menu.js
```diff
@@ -444,6 +444,14 @@ function ContextMenu(props: Props) {
...
+        className="node-context-menu-overlay"
+        onClick={hideContextMenu}
+        onContextMenu={evt => {
+          evt.preventDefault();
+          hideContextMenu();
+        }}
+      />
       <div
         style={{
           position: "absolute",
```
MichaelBuessemeyer, line 457:
```suggestion
```
This property can be removed as it is already included in the css.

MichaelBuessemeyer (approved):
Works very well :)

frontend/javascripts/oxalis/view/context_menu.js
philippotto: good catch!

MichaelBuessemeyer: Sorry my bad,
this was not redundant and just introduced a bug.
I am opening a quick fix pr now

Issue #5659, opened 2021-08-09
Volume tracing data download sometimes results in Connection Reset By Peer


fm3:
When using the allDataBlocking route, which is supposed to yield a data.zip for a volume tracing, sometimes the request is not completed successfully, and curl shows “Connection Reset By Peer”. We already traced this back to webKnossos, by going around all other network/cluster/docker layers.

fm3:
Looks like this is indeed closed by #5660

PR #5660, opened 2021-08-09
Update akkaHttp

fm3:
Attempt at fixing #5659 – assuming it is caused by https://github.com/akka/akka-http/issues/2093

- https://updateakka.webknossos.xyz

### Steps to test:
- create volume annotation
- download it (repeatedly), e.g. via curl
- each time the curl command should succeed and the zip should be intact

------
- [x] Ready for review


jstriebel (approved):
:+1: YOLO

PR #5661, opened 2021-08-09
Allow node selection in move tool

### LINKED ISSUE #5649: Cannot jump to tracing node with shift click in read only tracing
## Context
In a read only tracing I cannot jump to a tree node with shift click anymore.
Shift click does work in the 3D view port.

## Expected Behavior
shift click should jump to node.

## Current Behavior
shift click does not do anything outside the 3d view port.

## Steps to Reproduce the bug
1. Open read only tracing
2. shift click on node

## Your Environment for bug
- Browser name and version: Firefox
- Operating System and version: mac

philippotto:
...when modern controls are enabled OR when shift is pressed.

- https://___.webknossos.xyz

### Steps to test:
- open a skeleton tracing in read only mode
- enable classic mode
- node selection should work when shift is pressed (also see that the shortcut hint is visible in the status bar as soon as shift is pressed).
- disable classic mode
- node selection should work via a simple left click.

Rationale:
- for classic mode users, this behavior is exactly as before (shift + leftclick in read-only always used to select nodes)
- for modern mode users, this behavior should be a) intuitive (a simple click selects a node) and b) not intrusive, since accidental clicks on a node would only change the current position in the worst case (moving a node is not possible).

### Issues:
- fixes #5649

------
- [X] Ready for review


fm3 (approved):
LGTM, tested and works as described, which I think is a reasonable version :)

PR #5662, opened 2021-08-09
Auto switch organization suggestion

### LINKED ISSUE #5511: Allow to switch to dataset's organization if possible + necessary
When one is member of multiple organizations (e.g., A and B, but currently A is "active"), clicking on a link to a dataset of organization B, a 404 page is shown. Ideally, the page should show:

"The dataset belongs to organization B. Do you want to switch to that organization?"

Obviously, this should only be shown, if one has access to organization B. Ideally, the switch would happen automatically, but there are some side-effects to that (since the "active organization" changes), which is why I wouldn't do this for now.

MichaelBuessemeyer:
This PR makes it easier to view datasets of other organizations than the user's active organization. It adds a button to the "Failed loading dataset/tracing view" that lets the user directly switch to the correct organization that owns the dataset. But this only works if the user is actually part of the organization.

- https://___.webknossos.xyz

### Steps to test: Better test this locally
- Turn on `isDemoInstance` in the application.conf (and restart your local running wk instance)
- Create a new user. This user should automatically have its own organization. Upload a test dataset for this organization.
- View the newly created dataset and copy the view link.
- Switch back to scmboy and try to view the dataset via the previously copied link.
- The loading should fail and there should be a second button (left to the usual "return to dashboard button") that should automatically switch to the correct organization, so that the dataset can be viewed.
- Then open an explorative tracing with scmboy on the dataset of the new organization.
- In the tracing view switch to the sample_organization.
- The tracing should fail loading and there should be the same suggestion button to switch to the correct organization as in the step before.
- Create a task for the dataset of the new organization and open it with the scmboy. When the scmboy has the new organization active, everything should be working fine. If the sample_organization is active, there should be the suggestion to switch the new organization.
- Now log in as the second user that was previously created and try to view a dataset, an explorational tracing and a task from a dataset from the sample_organization. In all three cases there shouldn't be a suggestion to switch the active ogranization.

### Issues:
- fixes #5511
- Note: This fix only works for view mode, as I do not know how to get the owning organization of the dataset for an e.g. explorative tracing.

------
- ~~[ ] Updated [(unreleased) changelog](../blob/master/CHANGELOG.unreleased.md#unreleased) <-- This is an improvement usually just for us, right?~~
- [x] Ready for review


conf/application.conf
```diff
@@ -72,7 +72,7 @@ features {
   discussionBoardRequiresAdmin = false
   addForeignDataset = false
   hideNavbarLogin = false
-  isDemoInstance = false
+  isDemoInstance = true
```
philippotto, line 75: reminder to revert before merging

frontend/javascripts/oxalis/controller.js
```diff
@@ -281,22 +292,54 @@ class Controller extends React.PureComponent<PropsWithRouter, State> {
...
+      </AsyncButton>
+    );
     if (status === "loading") {
       return <BrainSpinner />;
     } else if (status === "failedLoading" && user != null) {
+      const message =
+        suggestedOrganization != null
+          ? `This dataset belongs to the organization ${
+              suggestedOrganization.displayName
+            } and this is currently not your active organization. Do you want to switch to that organization?`
```
philippotto, line 329:
```suggestion
            } which is currently not your active organization. Do you want to switch to that organization?`
```

frontend/javascripts/oxalis/controller.js
```diff
@@ -114,6 +118,13 @@ class Controller extends React.PureComponent<PropsWithRouter, State> {
       .catch(error => {
         this.props.setControllerStatus("failedLoading");
         const isNotFoundError = error.status === 404;
+        if (isNotFoundError) {
+          // The dataset could probably not be fetched, because the user is currently not in the correct organization.
+          // Thus fetch them to check whether the user can switch to the organization of the dataset.
```
philippotto, line 123: just nitpicking

```suggestion
          // The dataset could not be fetched. This might have happened because the user is currently not in the correct organization.
          // Thus, fetch the user's organizations to check whether they can switch to the organization of the dataset.
```

frontend/javascripts/oxalis/controller.js
```diff
@@ -281,22 +292,54 @@ class Controller extends React.PureComponent<PropsWithRouter, State> {
...
+    if (
+      user != null &&
+      initialCommandType.type === ControlModeEnum.VIEW &&
+      initialCommandType.owningOrganization !== user.organization
+    ) {
+      // Retrieving organization name from url. This only works in view mode.
+      const owningOrganizationName = location.pathname.split("/")[2];
+      suggestedOrganization = usersOrganizations.find(
+        orga => orga.name === owningOrganizationName && orga.name !== user.organization,
+      );
```
philippotto, line 307: Hm, I think, it would be very good to also support the scenario where an annotation link points to another orga. I'm wondering whether the back-end could check whether the user is member of the corresponding orga? Then, the back-end could reply with the organization id (since this is only done, if the user can switch to that orga, this is not leaking any private information).
The benefits of that would be:
- no parsing of the url to extract the organization
- support for annotation links
- no need to fetch the user's organizations async (currently, this can lead to a weird UI, where the 404 is already shown, but the "auto-switch possibility" is still fetched)

@fm3 Do you think it makes sense to do this on the back-end?

conf/application.conf
MichaelBuessemeyer: Thanks for noticing :)

frontend/javascripts/oxalis/controller.js
fm3: I started implementing an `accessibleBySwitching` route. It is not finished yet, I will continue tomorrow.

frontend/javascripts/oxalis/controller.js
```diff
@@ -120,9 +124,16 @@ class Controller extends React.PureComponent<PropsWithRouter, State> {
         ) {
           Toast.error(messages["tracing.compound_project_not_found"], { sticky: true });
         }
-        // Don't throw errors for errors already handled by the model
-        // or "Not Found" errors because they are already handled elsewhere.
+        if (
+          error.organizationToSwitchTo != null &&
+          this.props.user != null &&
+          this.props.initialCommandType.owningOrganization !== this.props.user.organization
```
philippotto, line 130: Isn't `this.props.initialCommandType` a `TraceOrViewCommand` which doesn't have `owningOrganization` for the `TRACE` case? I would have thought, that flow complains about this or am I missing something? 

```suggestion
          error.organizationToSwitchTo !== this.props.user.organization
```

would make more sense, I think?

philippotto (commented):
front-end looks very good :+1: I only added one question. After that is clarified, I'll test.

frontend/javascripts/oxalis/controller.js
MichaelBuessemeyer: Oh, thanks for catching this "problem". In the previous version where the API route did not exist I, this check was simple to avoid suggesting the user to the organization he was already in. In the `TRACE` case, the `owningOrganization` prop would simply be null and therefore the check would have still succeeded. At least this was the plan.

In the current version, your suggestion should be correct.

However, I think we can remove the whole check of `error.organizationToSwitchTo !== this.props.user.organization` in the current version, as we can assume, 
1. that the backend does not suggest to switch to the already active organization
2. The whole call to ask for the organization to switch to would only be executed once the loading fails. Thus if the user was already in the correct organization, the user should be able to see the dataset, if the user is allowed to.

fm3: I added a line to the backend that ensures it never suggests to switch to the orga the user is already in :)

philippotto: > that the backend does not suggest to switch to the already active organization

@fm3 is this the case? at least, I cannot read the behavior in the back-end code right now. if the back-end doesn't do this, the frontend can easily do this here in my opinion.
If this is not handled, some other error could send the user into an endless loop of "loading fails" -> "try this organization" -> "loading fails" -> ...

app/controllers/Application.scala
```diff
@@ -63,7 +63,7 @@ class Application @Inject()(multiUserDAO: MultiUserDAO,
     Ok(conf.raw.underlying.getConfig("features").resolve.root.render(ConfigRenderOptions.concise()))
   }
 
-  def health: Action[AnyContent] = sil.UserAwareAction { implicit request =>
+  def health: Action[AnyContent] = Action {
```
fm3, line 66: note for reviewer: this slipped in here, small fix for a compiler warning

frontend/javascripts/oxalis/controller.js
philippotto: ah, race condition during writing comments :see_no_evil: awesome! :+1:

philippotto (approved):
Great, great stuff :+1: Works very well. Please wait for @youri-k's back-end review before merging.

youri-k (approved):
Backend LGTM :tada:

PR #5663, opened 2021-08-10
Disable jobs for non-local datasets

### LINKED ISSUE #5625: Disable job triggers for datasets that are not reachable for wk-worker

MichaelBuessemeyer:
This PR adds the new property `jobsEnabled` to each dataset. Currently, only locally hosted datasets have this field enabled. This field is now used to determine, whether the running wk instance supports jobs and a dataset is hosted locally. If both properties are true, jobs are available for this dataset. For datasets that do not match this requirement jobs are now disabled with this PR. 

- https://___.webknossos.xyz

### Steps to test:
First setup a second local datastore:
- Execute the following SQL query to the postgres db of wk:
```
insert into webknossos.datastores(name, url, publicUrl, key, isForeign, isDeleted) values('standalone-9090', 'http://localhost:9090', 'http://localhost:9090', 'something-secur3', false, false);
```
- Enable jobs in the application.conf
- Start the normal wk instance
- In a second terminal in the wk dir start another wk datastore with 
```sbt "webknossosDatastore/run 9090 -Dconfig.file=webknossos-datastore/conf/standalone-datastore.conf"```

Then test this PR:
- Open a dataset hosted by the wk instance with the label "localhost" in the dataset table in view mode.
- Open the bounding box tab. Create a bounding box with the "+" button. Then click the download button. This should successfully open a modal for the download.
- Open the meshes tab. There the precompute meshes button should be enabled.
Now upload a new dataset with a segmentation layer to the none localhost instance. Therefore go to the dataset table, click add dataset and select "standalone-9090" as a datastore. After the upload finished do:
- Open the dataset in view mode
- Open the bounding box tab. Create a bounding box with the "+" button. The download button should be disabled and do nothing upon click. A tooltip should indicate, that the download is not supported for non local datasets.
- Open the meshes tab. There the precompute meshes button should be disabled with a tooltip hinting that this feature is only for local datasets.

### Issues:
- fixes #5625 

------
- [x] Ready for review


MichaelBuessemeyer:
@fm3 The checklist in the description is correct, right? Or is a datastore update or something else needed?

fm3:
Yes, those steps should work :)

philippotto (approved):
works great :) also great steps-to-test description :+1: I'm copying this to our notion.

fm3:
@philippotto note that the setup instructions are in part taken from https://github.com/scalableminds/webknossos/wiki/Set-up-a-standalone-datastore-locally maybe this should be referenced in notion as well :)

philippotto:
ah, thanks for the info. i also crosslinked that page now.

PR #5665, opened 2021-08-11
Integrate "What's New" Notifications/Updates

### LINKED ISSUE #5652: Integrate "What's New" UI
We should be able to inform our users about changes/news within webKnossos. As an example, have a look at https://announcekit.app/.

- [ ] research available options (see https://www.notion.so/scalableminds/Design-Doc-News-Channel-8062024740f34fa58b819f1f0bc8a80f)
- [ ] decide on a service
- [ ] integrate
- [ ] maintain

philippotto:
- https://olvy.webknossos.xyz

### Steps to test:
- open dev instance (isDemo is enabled currently)
- click on "What's New" to see a modal with wK news
- publish some new news on olvy (hit me up for credentials)
- reload wk and expect a red dot on the "What's New" button
- open the modal again --> dot should disappear
- reload page
- dot should be gone

### Issues:
- fixes #5652 

------
- [x] Ready for review


philippotto:
I integrated the feedback I got and now the `lastOpenedReleasesTimestamp` is stored in the database. Also, the notification indicator is integrated into the avatar. It looks like this:

image
image


@fm3 Please have a look at the back-end code I added. I'm not sure whether I adhered to the existing practices (e.g., the parameter ordering was not very clear to me).

philippotto:
@fm3 Also maybe can you help with writing a migration which adds the `lastOpenedReleasesTimestamp` column (could default to lastActivity or to the execution time of the migration if that's necessary)? Happy to have a call about it :call_me_hand:

conf/application.conf
```diff
@@ -72,7 +72,7 @@ features {
   discussionBoardRequiresAdmin = false
   addForeignDataset = false
   hideNavbarLogin = false
-  isDemoInstance = false
+  isDemoInstance = true
```
normanrz, line 75: Undo

fm3:
@philippotto I pushed a version that moves this timestamp to the multiUser’s novelUserExperienceInfos. Note that I also changed the code slightly to allow for this field to be missing. This way users that never clicked the button will see all that happened since their registration (which is similar to defaulting to the time of the evolution, as we will start publishing entries only then). This way we don’t need a migration or any other back-end changes. Do you agree that this covers the use cases?

frontend/javascripts/navbar.js
```diff
@@ -327,11 +382,32 @@ function LoggedInAvatar({ activeUser, handleLogout, ...other }) {
...
+      ? maybeUnreadReleaseCount > 0
+      : false;
+  const handleShowReleases = () => {
+    const [newUserSync] = updateNovelUserExperienceInfos(activeUser, {
+      lastViewedWhatsNewTimestamp: new Date().getTime(),
+    });
+
+    Store.dispatch(setActiveUserAction(newUserSync));
+
+    window.Olvy.show();
```
normanrz, line 397: could you add analytics here, so that we know how often den modal is shown?

normanrz (approved):

philippotto:
> @philippotto I pushed a version that moves this timestamp to the multiUser’s novelUserExperienceInfos. Note that I also changed the code slightly to allow for this field to be missing. This way users that never clicked the button will see all that happened since their registration (which is similar to defaulting to the time of the evolution, as we will start publishing entries only then). This way we don’t need a migration or any other back-end changes. Do you agree that this covers the use cases?

Thank you for doing this! Somehow I missed that this will be a front-end-only code change then :facepalm: But glad to see that it turned out that way :+1:

frontend/javascripts/navbar.js
```diff
@@ -327,11 +383,33 @@ function LoggedInAvatar({ activeUser, handleLogout, ...other }) {
 
   const isMultiMember = switchableOrganizations.length > 0;
 
+  const maybeUnreadReleaseCount = useOlvyUnreadReleasesCount(activeUser);
+  const showNotification =
+    features().isDemoInstance && maybeUnreadReleaseCount != null
+      ? maybeUnreadReleaseCount > 0
+      : false;
+  const handleShowReleases = () => {
```
fm3, line 391: what do you think about renaming this to WhatsNewView everywhere, to unify the variables?
```suggestion
  const handleShowWhatsNewView = () => {
```

frontend/javascripts/navbar.js
philippotto: good idea. done

frontend/javascripts/navbar.js
philippotto: good point. done

PR #5668, opened 2021-08-12
Fix hideUnmappedIds resetting once placing a node in merger mode

### LINKED ISSUE #5599: "Hide Unmapped IDs" resets after setting node in merger-mode
## Steps to Reproduce the bug

1. open a skeleton or hybrid tracing for a dataset with a segmentation layer
2. enable merger mode
2. enable "hide unmapped ids" in the mapping settings
3. create a new node
4. "hide unmapped ids" will disable automatically even though it should not

## Your Environment for bug
chrome / linux

MichaelBuessemeyer:
This PR fixes that the hideUnmappedIds option reset to false / was turned of each time a node was created when the merger mode was turned on.

- https://___.webknossos.xyz

### Steps to test:
1. open a skeleton or hybrid tracing for a dataset with a segmentation layer
2. enable merger mode
2. enable "hide unmapped ids" in the mapping settings
3. create a new node
4. "hide unmapped ids" should no longer disable automatically

### Issues:
- fixes #5599

------
- [x] Ready for review

Note: There were multiple ways to tackle this issue. One could have been to change the merge mode, to always pass the current setting of `hideUnmappedIds` to the API call "setMapping". Another would be to change the API "setMapping"-method to not passed undefined as  API `hideUnmappedIds`.
But I prefer the implemented version, because it is much shorter and catches the problem at the root. Additionally, I think it makes more sense when the `setMappingAction` called without `hideUnmappedIds` being set, does not change this setting.


philippotto (approved):
Perfect! :star: 

> But I prefer the implemented version, because it is much shorter and catches the problem at the root. Additionally, I think it makes more sense when the setMappingAction called without hideUnmappedIds being set, does not change this setting.

Yes, full ack :+1:

PR #5673, opened 2021-08-13
Add tooltips for settings in left sidebar

### LINKED ISSUE #5566: Add explaining tooltips for all settings in left sidebar
## Detailed Description
Some settings in the left sidebar have little "i" icons that show tooltips, while others don’t. I think some non-obvious ones are left unexplained. I suggest removing the "i" icon and instead give the text itself the tooltip functionality, as is the case already e.g. for tab titles.

## Context
New and existing users may misunderstand some settings or not find what they are looking for

fm3:
- https://moretooltips.webknossos.xyz

### Steps to test:
- open view mode, annotation mode (orthogonal, flight), with and without volume
- hover over settings labels, tooltips should be explanatory

### Issues:
- fixes #5566 

------
- [x] Ready for review


frontend/javascripts/oxalis/view/left-border-tabs/controls_and_rendering_settings_tab.js
```diff
@@ -279,31 +331,29 @@ class ControlsAndRenderingSettingsTab extends PureComponent<ControlsAndRendering
...
+              </Tooltip>
             }
             value={this.props.datasetConfiguration.fourBit}
             onChange={this.onChangeDataset.fourBit}
           />
           {Constants.MODES_ARBITRARY.includes(this.props.viewMode) ? null : (
             <SwitchSetting
-              label={settingsLabels.interpolation}
+              label={
+                <Tooltip title="Smooth the rendered data by interpolating color values, depending on zoom step">
```
philippotto, line 344:
```suggestion
                <Tooltip title="Smooth the rendered data by interpolating color values.">
```

not sure if the zoom-step-info is really helpful :thinking:

frontend/javascripts/oxalis/view/left-border-tabs/controls_and_rendering_settings_tab.js
```diff
@@ -91,20 +102,32 @@ class ControlsAndRenderingSettingsTab extends PureComponent<ControlsAndRendering
...
               roundTo={3}
               min={this.props.validZoomRange[0]}
               max={this.props.validZoomRange[1]}
               value={this.props.zoomStep}
               onChange={this.props.onChangeZoomStep}
             />
             <SwitchSetting
-              label={settingsLabels.displayCrosshair}
+              label={
+                <Tooltip title="Show crosshair lines in each viewport, with colors indicating the other viewports">
```
philippotto, line 118: some of the tooltip texts (such as this one) are duplicated. can you please extract the strings into a variable (e.g., something like `settingsLabels.displayCrosshairTooltip` in `messages.js`? alternatively, you could try to extract the entire `<SwitchSetting />` node into a method-local variable. The components look very similar and if they are equal we should remove the redundancy on that level.

philippotto (changes requested):
Awesome! I only have one nitpick :)

frontend/javascripts/oxalis/view/left-border-tabs/controls_and_rendering_settings_tab.js
fm3: I extracted the tooltip strings to messages.js. To be honest I am not sure what the difference between the case `Constants.MODE_PLANE_TRACING` and case `Constants.MODE_VOLUME` is at all. Also `default` seems to be flight mode? Maybe this could be replaced by a simple if-else?

philippotto: yeah, this seems to be the result of historic growth (from a time where the skeleton options where in the same tab). thanks for taking the time to extract the tooltip messages :+1: I think, we can leave the PR now as is :)

philippotto (approved):

frontend/javascripts/oxalis/view/left-border-tabs/controls_and_rendering_settings_tab.js
fm3: I replaced the switch by if (arbitrary or flight) {flight settings} else {orthogonal settings}, could you have another look? Or would you have preferred the old version?

philippotto: awesome! looks perfect :+1:

PR #5676, opened 2021-08-16
Fix rendering for intensity range of 0

### LINKED ISSUE #5671: Setting min and max to same value in histogram "hides" all layers
If min and max are equal, this line divides by 0 and adds the resulting NaN to the output color value.
https://github.com/scalableminds/webknossos/blob/be02fb595d949bf2e5599f8fa93e5883950d3882/frontend/javascripts/oxalis/shaders/main_data_fragment.glsl.js#L187-L187

MichaelBuessemeyer:
This PR fixes a bug where having one layer with the intensity range of 0 led to no data at all being rendered.

This problem is tackled by 
1. determining `min == max`
1. if `min == max` add +1 to the divisor of the division to avoid dividing by 0
1. Later replace the calculated color value with black in the case of `min == max`.

- https://___.webknossos.xyz

### Steps to test:
- Please do the following steps for a few datasets to ensure this fix works for "all". I tested this on the ROI and l4 sample dataset.
- Open a dataset in view mode
- Go to the layer settings and move the slider of a color layer of your choice to the same values (e.g. 100).
- The data layer should now no longer be rendered but the other active layers should still be visible.
- Additionally, there should be a warning icon stating that the layer is not rendered and that the user should prefer to disable the layer instead of reducing the intensity range to 0.

### Issues:
- fixes #5671

------
- [ ] Ready for review


frontend/javascripts/oxalis/view/left-border-tabs/layer_settings_tab.js
```diff
@@ -428,6 +431,14 @@ class DatasetSettings extends React.PureComponent<DatasetSettingsProps, State> {
             <InfoCircleOutlined style={{ marginLeft: 4 }} />
           </Tooltip>
 
+          {intensityRange[0] === intensityRange[1] && !isDisabled ? (
+            <Tooltip
+              title={`No data is being rendered for this layer as the minimum and maximum of the range have the same values. 
+            If you do not want this layer to be rendered, better disable it with the switch on the left.`}
```
philippotto, line 437: :+1: for taking the time to handle this in the UI

```suggestion
            If you want to hide this layer, you can also disable it with the switch on the left.`}
```

sounds better imo :)

philippotto (approved):
Great fix :)

philippotto:
please update the changelog before merging :)

PR #5679, opened 2021-08-17
Blur viewport maximize button after toggling it

### LINKED ISSUE #5609: SPACE key can toggle fullscreen mode
## Context
I was browsing my data and switched to full screen mode for the XY viewport by clicking on the little window icon in the upper right of the viewport. Then I wanted to continue navigating through my dataset using the SPACE hotkey. Instead of moving in z-direction the fullscreen mode was (un)-toggled (back to mini viewport). Pressing SPACE once more toggled the fullscreen mode again. Looks like the fullscreen button still has the focus after pressing it.

## Expected Behavior
Hotkey SPACE should navigate in third dimension and not toggle fullscreen mode.

## Current Behavior
Hotkey SPACE toggles fullscreen mode after pressing on fullscreen button.

## Steps to Reproduce the bug
1. Click on fullscreen mode for any viewport with mouse
2. Press SPACE
3. See it toggle between fullscreen and regular mode

## Your Environment for bug
* Browser name and version: e.g. Chrome 91
* Operating System and version: e.g. MacOS
* Version of WebKnossos (Release or Commit): 15036

MichaelBuessemeyer:
This PR solves #5609 as suggested by blurring the button to return the focus to the document body. Thus pressing space after toggling the maximize button via a mouse click no longer causes another toggle and instead triggers a movement.
Additionally, it sets the toggled viewport as active so that the third dimension is correct and pressing space moves forth / back.

- https://___.webknossos.xyz

### Steps to test:
- Open a dataset in view mode instantly move the cursor over the navbar.
- From there directly move the cursor to the maximize button of the YZ plane and maximize it.
- Then, without moving the mouse (setting the YZ viewport active via hovering with the mouse) try to move with `space` / `shift + space` back and forth. If the direction is correct, both changes should work.

- After that simply toggle the maximize on the other viewports a few times and check whether this still works fine and without any errors occurring. 

### Issues:
- fixes #5609

------
- [x] Ready for review

### DIFF
--- a/frontend/javascripts/oxalis/view/input_catcher.js
+++ b/frontend/javascripts/oxalis/view/input_catcher.js
@@ -78,10 +78,6 @@ export function recalculateInputCatcherSizes() {
   Store.dispatch(setInputCatcherRects(viewportRects));
 }
 
-export function getHTMLIdForViewportId(viewportID: Viewport) {
-  return `inputcatcher_${viewportID}`;
-}
-
 class InputCatcher extends React.PureComponent<Props, {}> {
   domElement: ?HTMLElement;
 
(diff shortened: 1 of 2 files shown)


philippotto (approved):
Excellent :partying_face:

PR #5682, opened 2021-08-23
Make not rendered layer info less verbose + more precise

### LINKED ISSUE #5650: Resolution rendering hint is too verbose
The tooltip “The layer Segmentation doesn’t exist in the current resolution 1-1-1. Adjust the zoom level to change the active resolution.” is too long. This was reported by @Albane-dlv

We’d suggest something like `“Zoom in to see layer{pluralSIn} {unrenderableLayerNamesZoomIn.map(name => "${name}").join(", ")}.` For this it would be necessary to distinguish if there are layers for which zooming in helps, and if there are layers for which zooming out helps. `unrenderableLayers` in `dataset_accessor.js` does not currently support this.

MichaelBuessemeyer:
- https://___.webknossos.xyz

### Steps to test:
- Open a dataset that has sparse layers / layers that do not have data in each mag.
- The "not rendered layers info" in the bottom left corner of a viewport should now be a bit shorter and more direct.

### Issues:
- fixes #5650 

------
- [x] Ready for review


frontend/javascripts/oxalis/model/accessors/dataset_accessor.js
```diff
@@ -591,8 +610,8 @@ function _getUnrenderableLayersForCurrentZoom(state: OxalisState) {
       resolutionInfo: getResolutionInfo(layer.resolutions),
     }))
     .filter(({ resolutionInfo }) => {
-      const isMissing = !resolutionInfo.hasIndex(zoomStep);
-      if (!isMissing) {
+      const isPresent = resolutionInfo.hasIndex(zoomStep);
+      if (isPresent) {
```
MichaelBuessemeyer, line 614: Here I got rid of the double negation which might be confusing

frontend/javascripts/oxalis/view/viewport_status_indicator.js
```diff
@@ -43,9 +49,21 @@ export default function ViewportStatusIndicator() {
...
+                  ? " "
+                  : ""
+              }`
+            : null}
+          {unrenderableLayerNamesWithHigherZoomstep.length > 0
+            ? `Zoom out to see layer${pluralSForHigherZoomstep} ${unrenderableLayerNamesWithHigherZoomstep.join(
+                ", ",
+              )}${renderMissingDataBlackHint ? " " : ""}`
+            : null}
+          {renderMissingDataBlackHint}
```
MichaelBuessemeyer, line 66: Do you have any suggestions on how to improve this @philippotto?

frontend/javascripts/oxalis/view/viewport_status_indicator.js
```diff
@@ -1,39 +1,45 @@
...
-  const pluralS = unrenderableLayerNames.length > 1 ? "s" : "";
-  const dontVerb = unrenderableLayerNames.length > 1 ? "don't" : "doesn't";
+  const unrenderableLayerNamesWithSmallerZoomstep = unrenderableLayersWithInfo
+    .filter(([, { smaller }]) => smaller)
+    .map(([layerName]) => layerName);
+  const pluralSForSmallerZoomstep = unrenderableLayerNamesWithSmallerZoomstep.length > 1 ? "s" : "";
+  const unrenderableLayerNamesWithHigherZoomstep = unrenderableLayersWithInfo
+    .filter(([, { higher }]) => higher)
+    .map(([layerName]) => layerName);
+  const pluralSForHigherZoomstep = unrenderableLayerNamesWithHigherZoomstep.length > 1 ? "s" : "";
```
MichaelBuessemeyer, line 42: I kinda dislike the naming of the variables here. The advantage is, that the names are pretty explainatory. Do you have suggestions for better names here @philippotto?

MichaelBuessemeyer (commented):
I am not so happy about some parts in this pr but could not come up with an idea how to improve them.

MichaelBuessemeyer:
A changelog entry is not needed for this pr, is it?

frontend/javascripts/oxalis/view/viewport_status_indicator.js
philippotto: How about something like this (not tested obviously):

```javascript
function getOptionalZoomMessage(directionString, unrenderableLayersWithInfo) {
  const unrenderableLayerNamesWithOtherZoomstep = unrenderableLayersWithInfo
    .filter(([, { higher, smaller }]) => directionString === "in" ? smaller : higher)
    .map(([layerName]) => layerName);
  if (unrenderableLayerNamesWithOtherZoomstep.length === 0) return null;
  
  const pluralS = unrenderableLayerNamesWithOtherZoomstep.length > 1 ? "s" : "";
  return `Zoom {directionString} to see layer${pluralS} ${unrenderableLayerNamesWithOtherZoomstep.join(
    ", ",
  )}`
}


const zoomInMessage = getOptionalZoomMessage("in", unrenderableLayersWithInfo);
const zoomOutMessage = getOptionalZoomMessage("out", unrenderableLayersWithInfo);
const missingDataHint = ...;
const sentencesString = _.flatten([zoomInMessage, zoomOutMessage, missingDataHint]).join(" ");
...
return <div>{sentencesString}</div>;
```

this would also get rid of the naming problem, since the in/out distinction is not necessary anymore (due to the helper function)

frontend/javascripts/oxalis/model/accessors/dataset_accessor.js
```diff
@@ -612,7 +631,10 @@ function _getUnrenderableLayersForCurrentZoom(state: OxalisState) {
         return resolutionInfo.hasIndex(fallbackZoomStep);
       });
     })
-    .map<DataLayerType>(({ layer }) => layer);
+    .map<[DataLayerType, SmallerOrHigherInfo]>(({ layer, resolutionInfo }) => {
+      const hasSmallOrHigher = resolutionInfo.hasSmallerAndOrHigherIndex(zoomStep);
+      return [layer, hasSmallOrHigher];
```
philippotto, line 636: I think, I would prefer an object as a return value, since this is often easier to read/use (instead of `value[0]` it would be `value.layer` for example)
```suggestion
      return { layer, hasSmallOrHigher };
```

philippotto: please also rename `_getUnrenderableLayersForCurrentZoom` to something like `_getUnrenderableLayerInfosForCurrentZoom`. also adapt the functions comment to the new return type. ideally, also add the flow return signature.

frontend/javascripts/oxalis/model/accessors/dataset_accessor.js
```diff
@@ -155,6 +161,19 @@ export class ResolutionInfo {
     return bestIndexWithDistance[0];
   }
 
+  hasSmallerAndOrHigherIndex(index: number): SmallerOrHigherInfo {
+    const indices = this.getAllIndices();
+    const hasSmallOrHigher = { smaller: false, higher: false };
+    indices.forEach(currentIndex => {
```
philippotto, line 167: I think, we usually use `for (const currentIndex of indices) { ... }` instead of the `forEach` method.

frontend/javascripts/oxalis/view/viewport_status_indicator.js
philippotto: see my other comment

philippotto (changes requested):
great work :+1: I left some suggestions :dart:

philippotto:
> A changelog entry is not needed for this pr, is it?

yeah, I guess it's alright.

frontend/javascripts/oxalis/view/viewport_status_indicator.js
MichaelBuessemeyer: Very nice suggestion :smile: Thanks a lot :t-rex:

MichaelBuessemeyer:
@philippotto I just applied your feedback. Could you please check and test this again? I did some testing for myself and it looked fine :smile:

frontend/javascripts/oxalis/model/accessors/dataset_accessor.js
```diff
@@ -612,12 +638,15 @@ function _getUnrenderableLayersForCurrentZoom(state: OxalisState) {
...
+      const smallerOrHigherInfo = resolutionInfo.hasSmallerAndOrHigherIndex(zoomStep);
+      return { layer, smallerOrHigherInfo };
+    });
   return unrenderableLayers;
 }
 
-export const getUnrenderableLayersForCurrentZoom = reuseInstanceOnEquality(
-  _getUnrenderableLayersForCurrentZoom,
+export const getUnrenderableLayersInfoForCurrentZoom = reuseInstanceOnEquality(
+  _getUnrenderableLayersInfosForCurrentZoom,
```
philippotto, line 649: two things: 1) the names are not identical (`info` vs `infos`) and 2) I would use a third variant: `_getUnrenderableLayerInfosForCurrentZoom`. The reasoning is that this implies that one gets multiple "layer info"-things. for example, the "LayersInfo` naming sounds like one would only get one "info"-thing which conveys the wrong semantics in my opinion. feel free to disagree, though :D

philippotto (approved):
awesome! see my one nitpick before merging, though

frontend/javascripts/oxalis/model/accessors/dataset_accessor.js
MichaelBuessemeyer: >1) the names are not identical (info vs infos)

:man_facepalming: 

> for example, the "LayersInfo` naming sounds like one would only get one "info"-thing which conveys the wrong semantics in my opinion.

I agree, lets go with `_getUnrenderableLayerInfosForCurrentZoom`

philippotto:
:+1:

PR #5683, opened 2021-08-23
Support multiple segmentation layers

### LINKED ISSUE #4555: Cheap support for multiple segmentation layers
It would already be useful to be able to have two segmentation layers in a datasource-properties.json and then have only one rendered in webKnossos.

The different degrees of this feature could be:
- simply pick first segmentation layer and ignore/drop the other one
- let the user pick a segmentation layer before the page loads fully (via URL parameter for example)
- let the user switch between segmentation layers in UI (could simply reload the entire page while setting the URL parameter)
- let the user switch between the layer without reloading (use `reloadBucket` functionality?)

Also: A lot of code assumes, that there is only one segmentation layer. That code could stay mostly intact by changing `getSegmentationLayer` to `getActiveSegmentationLayer` (the user would choose which seglayer is the active one).

As a very first step for this issue, verify that a segmentation layer does not need to be named "segmentation"

philippotto:
### Summary
- A dataset can have multiple segmentation layers now.
- Volume annotations are always based on ONE segmentation layer. It's not possible to brush in two different segmentation layers
- Only one segmentation layer can be rendered at all times (for now). Rendering of multiple seg layers itself is not a big problem, but there's a lot of code which simply accesses the "one visible segmentation" layer for which the adaption will take quite some effort.
- Structures, such as meshes and mappings, are maintained per segmentation layer now.

Also have a look at the follow up issue: https://github.com/scalableminds/webknossos/issues/5695

- https://multiplesegmentations.webknossos.xyz

### To Do

- [X] support opening DS with multiple segmentation layers
- [X] allow dynamic switching of visibility for segmentation layers
- [x] proper support for mappings
- [x] support for volume annotation (user should decide before hand for which layer an annotation is created)
- [x] support for (precomputed) meshes
- [X] create follow up issue: https://github.com/scalableminds/webknossos/issues/5695
- [x] the volume tools should be disabled if one has only enabled a secondary segmentation layer (i.e., not a volume layer)
- [x] fix tests
- [x] adapt "Create Annotation" and "Make Hybrid" to new `fallbackLayerName` parameter
- [x] clarify whether an annotation without fallback should still contain the original segmentation layer (maybe it's safer to do for now and later we can re-consider having both layers at the same time)
- [x] ~ensure that volume layer is always visible when creating a new annotation~
- [x] double check api modules (code examples and graceful fallback)
- [x] test manually

### Steps to test:
I suggest a lot of monkey testing, but broadly we should cover the following:
- Test that the status quo works:
  - [x] view/annotate datasets with zero segmentation layers
  - [x] view/annotate datasets with one segmentation layer
    - [x] with fallback segmentation
    - [x] without fallback segmentation
  - [x] meshes
    - [x] test precomputed mesh loading
    - [x] test ad-hoc mesh loading
  - [x] mappings
    - [x] merger mode
    - [x] JSON mappings
    - [x] agglomerate mappings
- Test the support of multiple segmentation layers
  - [x] open a dataset with two segmentation layers (simply copy a layer on disk or download a volume tracing and move it into a DS folder, be careful to set a large-enough bbox there)
  - [x] test switching between two segmentation layers
    - [x] enabling one layer should disable the layer automatically
  - [x] mappings
    - [x] test that mappings still work as usual
    - [x] enable mappings for two segmentation layers and switch between them

### Issues:
- fixes #4555 

------
- [X] Ready for review


app/models/annotation/AnnotationService.scala
```diff
@@ -139,17 +139,18 @@ class AnnotationService @Inject()(
...
-      withFallback: Boolean,
+      fallbackLayerNameOpt: Option[String],
       resolutionRestrictions: ResolutionRestrictions,
       organizationName: String,
       oldTracingId: Option[String] = None)(implicit ctx: DBAccessContext): Fox[(Option[String], Option[String])] = {
     def getFallbackLayer: Option[SegmentationLayer] =
-      if (withFallback) {
-        dataSource.dataLayers.flatMap {
+      // todo: probably should fail if the fallbackLayerName cannot be found
+      fallbackLayerNameOpt.map(fallbackLayerName =>
```
philippotto, line 148: @fm3 Can you have a look at my scala changes on this branch? You can also check out the branch and run the corresponding code by using the advanced "create explorational" modal and then picking a segmentation layer (one layer should be enough).

I think, there are at least 2 todos:
- ensuring that an error is thrown if a fallback layer name is passed which does not exist
- for the "make hybrid" route, the `fallbackLayerName` should be accepted as a GET parameter, too.

The bug which I mentioned earlier (that the fallback layer was ignored), was front-end only and is fixed now. So, no debugging necessary, I hope :)

frontend/javascripts/oxalis/model_initialization.js
```diff
@@ -472,16 +472,13 @@ function setupLayerForVolumeTracing(
...
   };
 
   if (fallbackLayer != null) {
     // Replace the original tracing layer
     layers[fallbackLayerIndex] = tracingLayer;
   } else {
-    // Remove other segmentation layers, since we are adding a new one.
-    // This is a temporary workaround. In the long term we want to support
-    // multiple segmentation layers.
-    layers = layers.filter(layer => layer.category !== "segmentation");
```
philippotto, line 484: for volume annotations, we could still kick out the other segmentation layer if there shouldn't be a fallback. might be less confusing, but I'm not sure about this.

frontend/javascripts/oxalis/model/reducers/ui_reducer.js
```diff
@@ -35,14 +35,6 @@ function UiReducer(state: OxalisState, action: Action): OxalisState {
       return updateKey(state, "uiInformation", { hasOrganizations: action.value });
     }
 
-    case "REFRESH_ISOSURFACES": {
-      return updateKey(state, "uiInformation", { isRefreshingIsosurfaces: true });
```
philippotto, line 39: was unused

frontend/javascripts/oxalis/geometries/skeleton.js
```diff
@@ -47,13 +47,13 @@ type BufferCollection = {
...
 const NodeBufferHelperType = {
-  addAttributes(geometry: typeof THREE.BufferGeometry, capacity: number): void {
-    geometry.addAttribute("position", new THREE.BufferAttribute(new Float32Array(capacity * 3), 3));
-    geometry.addAttribute("radius", new THREE.BufferAttribute(new Float32Array(capacity), 1));
-    geometry.addAttribute("type", new THREE.BufferAttribute(new Float32Array(capacity), 1));
-    geometry.addAttribute("isCommented", new THREE.BufferAttribute(new Float32Array(capacity), 1));
-    geometry.addAttribute("nodeId", new THREE.BufferAttribute(new Float32Array(capacity), 1));
-    geometry.addAttribute("treeId", new THREE.BufferAttribute(new Float32Array(capacity), 1));
+  setAttributes(geometry: typeof THREE.BufferGeometry, capacity: number): void {
+    geometry.setAttribute("position", new THREE.BufferAttribute(new Float32Array(capacity * 3), 3));
```
philippotto, line 51: this renaming is due to the threejs update

frontend/javascripts/features.js
```diff
@@ -19,6 +19,7 @@ export function getDemoDatasetUrl() {
     throw new Error("Features not yet loaded.");
   }
 
+  // todop: how to pass the fallback layer here?
```
philippotto, line 22: todo. we should probably encode the entire url in `application.conf` (so that it includes `createExplorative/hybrid?fallbackLayerName=...`).

philippotto:
@daniel-wer Welcome back :laughing: This PR got quite large, but I hope it's reviewable. There are still some todos left, but I think, I should catch a break from this and the main part should be completed. Also see the follow-up issue mentioned in the PR description, as that also sheds light on the restrictions of this first iteration. Let me know if you want to talk about this PR in person :call_me_hand:

frontend/javascripts/oxalis/model.js
```diff
@@ -148,10 +202,22 @@ export class OxalisModel {
...
+  getLayerByName(name: string): DataLayer {
     if (!this.dataLayers[name]) {
       throw new Error(`Layer with name ${name} was not found.`);
     }
-    return this.dataLayers[name].pullQueue;
+    return this.dataLayers[name];
+  }
+
+  getSegmentationLayerWithMappingSupport(): ?DataLayer {
+    const layer = getSegmentationLayerWithMappingSupport(Store.getState());
```
daniel-wer, line 216: Would it make sense to use different names indicating the difference between the two functions?

frontend/javascripts/oxalis/geometries/materials/plane_material_factory.js
```diff
@@ -678,24 +689,24 @@ class PlaneMaterialFactory {
       maximumLayerCountToRender,
     } = Store.getState().temporaryConfiguration.gpuSetup;
 
-    const segmentationLayer = Model.getSegmentationLayer();
+    // Don't compile code for segmentation in arbitrary mode
+    const hasSegmentation = this.isOrthogonal && Model.hasSegmentationLayer();
```
daniel-wer, line 693: Maybe rename to `shouldRenderSegmentation" or so to make the naming and intention more clear.

frontend/javascripts/oxalis/model.js
```diff
@@ -90,13 +95,67 @@ export class OxalisModel {
...
 
-  // Todo: Make return type optional
-  getSegmentationLayer(): DataLayer {
-    return _.find(
-      this.dataLayers,
-      dataLayer =>
-        getLayerByName(Store.getState().dataset, dataLayer.name).category === "segmentation",
+  getSegmentationLayers(): Array<DataLayer> {
+    return Object.keys(this.dataLayers)
+      .map(k => this.dataLayers[k])
```
daniel-wer, line 100: Is there a benefit of this to using `Object.values`? Flow?

frontend/javascripts/oxalis/model.js
```diff
@@ -90,13 +95,67 @@ export class OxalisModel {
...
+    return null;
+  }
+
+  hasSegmentationLayer(): boolean {
+    return this.getSegmentationLayers().length > 0;
+  }
+
+  getSegmentationTracingLayer(): ?DataLayer {
+    const { dataset } = Store.getState();
+    const layer = getSegmentationTracingLayer(dataset);
```
daniel-wer, line 144: See my comment regarding the `getSegmentationLayerWithMappingSupport` function

frontend/javascripts/oxalis/model/accessors/dataset_accessor.js
```diff
@@ -538,31 +562,111 @@ export function isColorLayer(dataset: APIDataset, layerName: string): boolean {
...
 }
 
-export function getSegmentationLayer(dataset: APIMaybeUnimportedDataset): ?APISegmentationLayer {
+export function getVisibleSegmentationLayer(state: OxalisState): ?APISegmentationLayer {
+  const { datasetConfiguration } = state;
+  const { viewMode } = state.temporaryConfiguration;
+
+  const visibleSegmentationLayers = getSegmentationLayers(state.dataset).filter(layer =>
+    isLayerVisible(state.dataset, layer.name, datasetConfiguration, viewMode),
+  );
```
daniel-wer, line 571: This can be replaced by a call to `getVisibleSegmentationLayers` I think

frontend/javascripts/oxalis/model/accessors/dataset_accessor.js
```diff
@@ -538,31 +562,111 @@ export function isColorLayer(dataset: APIDataset, layerName: string): boolean {
...
+  const layersWithEnabledMapping = visibleSegmentationLayers.filter(layer => {
+    const mappingInfo = state.temporaryConfiguration.activeMappingByLayer[layer.name];
+    return mappingInfo && mappingInfo.isMappingEnabled;
+  });
+
+  if (layersWithEnabledMapping.length > 0) {
+    return layersWithEnabledMapping[0];
+  }
+  const layersWithMappings = visibleSegmentationLayers.filter(
+    layer => layer.mappings && layer.mappings.length > 0,
```
daniel-wer, line 609: Just to make sure, does this include agglomerate file mappings?

frontend/javascripts/oxalis/model/bucket_data_handling/mappings.js
```diff
@@ -10,46 +10,42 @@ import type { APIMapping } from "types/api_flow_types";
...
-import { trackAction } from "oxalis/model/helpers/analytics";
 
 export const MAPPING_TEXTURE_WIDTH = 4096;
 export const MAPPING_COLOR_TEXTURE_WIDTH = 16;
 
 type APIMappings = { [string]: APIMapping };
 
-// For now, since we have no UI for this
-export function setupGlobalMappingsObject(segmentationLayer: DataLayer) {
+// Remove soon (e.g., October 2021)
```
daniel-wer, line 32: Out of curiosity, did you have a look whether these functions were called?

daniel-wer (commented):
First part of the review, I'll continue tomorrow. Looking very good so far :)

frontend/javascripts/oxalis/model/bucket_data_handling/mappings.js
philippotto: I just had a look and it seems to have to not been tracked at least. The overall event count was quite low. I suppose, lots of users have an adblocker :thinking:

frontend/javascripts/dashboard/advanced_dataset/dataset_action_view.js
```diff
@@ -34,7 +36,14 @@ function NewAnnotationLink({
   onShowCreateExplorativeModal,
   onCloseCreateExplorativeModal,
 }) {
-  const withFallback = doesSupportVolumeWithFallback(dataset) ? "true" : "false";
+  const firstSegmentationLayer = getFirstSegmentationLayer(dataset);
```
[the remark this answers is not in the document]
philippotto: This code should only be used within the dashboard for which there is no notion of a "visible layer". Therefore, we simply pick the first segmentation layer (if one exists). This is intended to be similar to the old behavior where there was at most one segmentation layer. If we want take any guessing out of the equation, we would need to open the modal always iff the dataset has more than 1 seg layer, but I'd rather defer the decision as described in https://github.com/scalableminds/webknossos/issues/5692. This should be less overhead for the user.

frontend/javascripts/features.js
```diff
@@ -19,10 +19,7 @@ export function getDemoDatasetUrl() {
     throw new Error("Features not yet loaded.");
   }
 
-  // Only create a tracing on the demo instance, as the user might not be logged in there
-  // otherwise.
-  const suffix = features.isDemoInstance ? "/createExplorative/hybrid/true" : "/view";
-  return features.publicDemoDatasetUrl + suffix;
+  return features.publicDemoDatasetUrl;
```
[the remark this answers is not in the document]
philippotto: Strictly speaking, yes. However, the normal URL also works without the `/view` part (luckily). Also, as you say, `publicDemoDatasetUrl` is probably only set for the demo instance anyway.

frontend/javascripts/oxalis/model.js
philippotto: Yeah, only flow :/

frontend/javascripts/oxalis/model/accessors/dataset_accessor.js
philippotto: Good point! I just checked it and agglomerates are stored separately in `layer.agglomerates` (which fits the implementation of this function well, since agglomerates aren't handled by the GPU).

frontend/javascripts/oxalis/api/api_latest.js
```diff
@@ -147,6 +150,35 @@ function assertVolume(tracing: Tracing): VolumeTracing {
...
+  throw new Error(
+    "No segmentation layer is currently visible. Pass a valid layerName (you may want to use `getSegmentationLayerName`)",
+  );
+}
+
+function getRequestedLayerNameOrVisibleLayer(state: OxalisState, layerName: ?string): ?string {
+  const layer = getRequestedSegmentationLayerOrVisibleLayer(state, layerName);
+  return layer != null ? layer.name : null;
+}
+
```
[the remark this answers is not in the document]
philippotto: Good point. The naming wasn't really consistent (nor good). I changed the function names to `getRequestedOrVisibleSegmentationLayer`, `getRequestedOrVisibleSegmentationLayerEnforced` and `getNameOfRequestedOrVisibleSegmentationLayer. I hope that's good enough.

frontend/javascripts/oxalis/model.js
philippotto: Hmm, maybe.. Not sure if this makes things easier. The difference between `DataLayer` and `APIDataLayer` is quite annoying in general (often, when having a `layer` variable, one has to find out first what exactly it is). Usually, I spot the difference between the two `getSegmentationLayerWithMappingSupport` (or similar) functions by their usage (either it's a method on `Model` or it's a function which consumes a `state`). In the long term, I want to get rid of the duplication (not sure how, though). Maybe having the same function name twice is a good indicator of this smell?

frontend/javascripts/dashboard/advanced_dataset/dataset_action_view.js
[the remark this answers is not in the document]
daniel-wer: Ah, that makes sense :+1:

frontend/javascripts/oxalis/model.js
daniel-wer: That's a good description of the issue :+1: We'll likely end up with some combination of renamings. The `APIDataLayer` could be `LayerSpec` or `LayerConfig`. The `DataLayer` could keep its name or be `LayerInstance`. However, I don't know all the details, so there might be a better solution.
I agree with leaving this as-is to keep the smell :grin:

frontend/javascripts/oxalis/view/left-border-tabs/layer_settings_tab.js
```diff
@@ -785,10 +787,9 @@ class DatasetSettings extends React.PureComponent<DatasetSettingsProps, State> {
 
   render() {
     const { layers } = this.props.datasetConfiguration;
-    const segmentationLayerName = Model.getSegmentationLayerName();
     const layerSettings = Object.entries(layers).map(entry => {
       const [layerName, layer] = entry;
-      const isColorLayer = segmentationLayerName !== layerName;
+      const isColorLayer = getLayerByName(this.props.dataset, layerName).category === "color";
```
daniel-wer, line 792: Minor: There is an `sColorLayer(dataset: APIDataset, layerName: string)` function checking exactly that in the `dataset_accessor.js`.

frontend/javascripts/oxalis/view/right-border-tabs/export_bounding_box_modal.js
```diff
@@ -119,24 +126,30 @@ const ExportBoundingBoxModal = ({ handleClose, dataset, boundingBox, tracing }:
...
+      parenthesesInfos.length > 0 ? ` (${parenthesesInfos.join(", ")})` : "";
+    return layerInfos ? (
       <p key={exportKey(layerInfos)}>
         <Button
           onClick={() => handleStartExport(layerInfos)}
           disabled={
             startedExports.includes(exportKey(layerInfos)) ||
             !layerInfos.hasMag1 ||
-            (isMergerModeEnabled && !layerInfos.isColorLayer)
+            (isMergerModeEnabled && layerInfos.tracingId != null)
```
daniel-wer, line 144: Is this the check that the layer is a volume tracing layer: `layerInfos.tracingId != null`?

frontend/javascripts/oxalis/view/right-border-tabs/export_bounding_box_modal.js
```diff
@@ -178,9 +191,7 @@ const ExportBoundingBoxModal = ({ handleClose, dataset, boundingBox, tracing }:
   let activeMappingMessage = null;
   if (isMergerModeEnabled) {
     activeMappingMessage =
-      "Exporting a volume layer does not export merger mode currently. Please disable merger mode before exporting data.";
-  } else if (isMappingEnabled) {
-    activeMappingMessage = `The active mapping ${mappingName} will be applied to the exported data.`;
```
daniel-wer, line 183: Did you remove this because it no longer happens or because the message was superfluous?

frontend/javascripts/oxalis/view/right-border-tabs/meshes_view.js
```diff
@@ -425,11 +456,12 @@ class MeshesView extends React.Component<Props, State> {
         meshfileResolutionIndex,
       );
 
-      if (this.props.segmentationLayer != null) {
+      if (this.props.visibleSegmentationLayer != null) {
         const job = await startComputeMeshFileJob(
           this.props.organization,
           this.props.datasetName,
-          this.props.segmentationLayer.fallbackLayer || this.props.segmentationLayer.name,
+          this.props.visibleSegmentationLayer.fallbackLayer ||
```
daniel-wer, line 463: This could reuse `getBaseLayerName(segmentationLayer: APIDataLayer)` from `meshes_view_helper.js`

daniel-wer (commented):
Finished the second part of the review. Very nice work :+1: 
The only thing up for debate is the shader code I'd say. Happy to discuss this in person :)

Since the testing is quite involved, I'll wait until we discussed that and all other Todos are taken care of.

frontend/javascripts/oxalis/shaders/main_data_fragment.glsl.js
```diff
@@ -139,30 +138,30 @@ void main() {
     gl_FragColor = vec4(bucketPosition, zoomStep) / 255.;
     return;
   }
+  vec3 data_color = vec3(0.0);
 
-  <% if (hasSegmentation) { %>
-    vec4 id = vec4(0.);
-    vec4 cellIdUnderMouse = vec4(0.);
+  <% _.each(segmentationLayerNames, function(segmentationName, layerIndex) { %>
```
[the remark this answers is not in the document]
philippotto: > Doesn't doing it this way with the respective variables and functions increase the shader length and complexity, leading to longer compile times.

Yes, that's correct, in general (even though, often still acceptable probably). However:

>  Since we're only ever rendering a single segmentation layer, couldn't we leave the shader code mostly as-is and only swap the textures + introduce and change some uniforms (packingDegree, layerIndex, etc.)?

The mid-term vision for this feature is to support simultaneous rendering of multiple segmentation layers (this is also listed in the follow-up issue). Therefore, I already built the shader structure that way :)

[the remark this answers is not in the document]
daniel-wer: Ah, I should have paid better attention. Thanks for clarifying, all good then :)

frontend/javascripts/oxalis/view/right-border-tabs/export_bounding_box_modal.js
philippotto: The message here was shown once (or not at all) for the entire modal. However, when there are multiple segmentation layers, the message needs to appear per layer. That's why, I moved the message into the layer-buttons (see https://github.com/scalableminds/webknossos/pull/5683/files/41a3695bc1e72a95ea8fa849b2fbb90da59d2975#diff-c3b0e8e7fff2777038951ac1d27237b7cf1fc00ea84b72eb977fde66242b7ddaR132).

frontend/javascripts/oxalis/view/right-border-tabs/export_bounding_box_modal.js
philippotto: Yes, exactly. `layerInfos.tracingId` is only defined for volume tracing layers (either with or without fallback). I added a comment to explain the three disabled-reasons. Instead of `tracingId`, I could also add another property, such as `isVolumeTracing`, but that would be somehow redundant, which is why I wasn't a big fan of that :thinking:

frontend/javascripts/oxalis/model_initialization.js
philippotto: Yeah, this is a good question. What's cool about the current state in this PR is that nothing "gets lost". If a dataset already has a segmentation, the user can always toggle it back on. However, if the user said "start with a fresh segmentation layer", that can be confusing. A trade-off might be to disable the original segmentation layer when opening the annotation. However, this would need to be done by the back-end, as the front-end cannot really distinguish between the first time and the n+1-th time a user opens an annotation.

@fm3 What do you about this? Is it feasible, to adapt the `/api/dataSetConfigurations` route so that the volume layer X is enabled and all other segmentation layers are disabled if the route is requested for the first time with a volume tracing id X?

philippotto: I discussed this with @fm3 and we agreed on kicking out the existing segmentation layer if the user chooses "without fallback". The main problem is that the visibility of the segmentation layer is stored per dataset. This means, if I create an annotation without fallback for a dataset, the "old" segmentation layer is disabled automatically and opening the dataset in view mode will be affected by that. Then, the segmentation layer will be invisible there, which is most likely not what the user wants nor expects.

I'll add this part to the follow-up issue.

philippotto:
Ok, my tests were mostly successful except for the merger mode which seems to be broken sometimes. Also, there is the question on which segmentation layer the merger mode should work. I'll have a look at this tomorrow.

philippotto:
> Ok, my tests were mostly successful except for the merger mode which seems to be broken sometimes. Also, there is the question on which segmentation layer the merger mode should work. I'll have a look at this tomorrow.

Should be fixed now.

daniel-wer (approved):
Very nice! I mostly tested that there was no regression in single-segmentation-layer-datasets and everything worked as expected :+1: 

LGTM

PR #5691, opened 2021-08-26
Allow zip download of (small) datasets

fm3:
New route `GET /data/datasets/:organizationName/:dataSetName/download?token=validDatastoreToken`

- https://downloaddataset.webknossos.xyz

### Steps to test:
- Take the token from /api/buildinfo (if you are superuser)
- Download a dataset, should result in zipfile with correct name and content
- Try downloading a dataset >1GB, should yield readable error

------
- [x] Ready for review


fm3:
As discussed in person, let’s make use of the requestRawCuboid route ( `GET /data/datasets/:organizationName/:dataSetName/layers/:dataLayerName/data`) for this

PR #5696, opened 2021-08-30
Segments lists

### LINKED ISSUE #4013: Straight cut in isosurfaces rendering
Looks like there might be an off by one error somewhere?
Could also be caused by my data...
Bildschirmaufnahme 2019-04-11 um 18 56 14

### LINKED ISSUE #4667: List of manually created segments
In volume tracings, there is no UI for seeing all manually created segments. For skeletons, there are tree and comment lists.

It would be nice to see all segments in one view and maybe also have some metadata, e.g. voxel count. This would only show manually created segments.

### LINKED ISSUE #5711: Allow configuring resolution for (ad-hoc) mesh rendering
## Detailed Description
Currently, mesh computation (ad-hoc and precomputed) will automatically pick a "suitable" magnification and resolution. Depending on the current use case, a higher resolution can be useful.

The UI could expose high/medium/low quality and should explain the trade-offs in quality and speed.

## Context

* [X] Specific to long-running jobs (set `jobsEnabled=true` in `application.conf`)

MichaelBuessemeyer:
This PR:
- changes the old "Meshes" tab to a "Segments" tab
- renders "Segments" in the new tab. The list of segments is not guaranteed to be complete or up to date. The list "grows" while creating an annotation or browsing a dataset. For example, selecting an existing segment or drawing with a new segment id will both ensure that the segment is listed. Outdated segments are not handled at all, right now. In a follow-up, we can validate existing segments ad-hoc and show an "outdated" warning etc.
- When loading a mesh, the mesh will be added as child to the segment (at least visually).
- The mesh UI (i.e., the header in the segments tab) was reworked so that the user can also choose which quality they want to have for the meshing.

- https://addsegmentslist.webknossos.xyz

### Steps to test:
- create a new annotation (with or without fallback layer) and brush --> a segment should appear in the segments tab
- open a dataset in view mode and hover existing segments --> the segments should appear, too
- open a dataset in view mode and compute meshes (precomputed or adhoc) --> the segments should appear, too
- monkey-test everything around segmentations, meshes and mappings
- please also have an eye on performance (for example, try to have more than 100 segments by shift+clicking segments in an annotation with fallback layer). overall wk performance should not be slowed down. the segment list does not use virtual rendering, but shouldn't cost anything, if the list is hidden. so, we can do further perf tuning as a follow-up.

### Issues:
- fixes #4013
- fixes #4667
- fixes #5711

------
- [x] Ready for review


MichaelBuessemeyer:
I stumbled upon a strange behavior/bug that still needs to be fixed:

This happens when a part of a segment is overdrawn and the segment's position has to be set again. 

#### Steps to reproduce:
1. In my concrete case, I draw a simple dot in one part of the dataset and then another dot in another part of the dataset. This ensures that the two parts of the segment are in different buckets.
2. Then I overdrew the second drawn dot, where currently the segment's position is at. This causes the code to find a new position for the segment. This is then exactly where the first dot was drawn but **one z-slice off**. 
3. To be able to see the first drawn dot, click on the segment in the list of the meshes tab. From there move one z slice backward and you can see the segment. 

#### Expected behavior: 
It would be ideal if it wasn't necessary to move back one slice.

MichaelBuessemeyer:
Keeping the segment list always in sync with the volume data is difficult. 
I added a section about this problem and how it might be possible to tackle it to the notion page:
https://www.notion.so/scalableminds/Design-Doc-Segment-Lists-c8ec5e23764b4a578a452a4c7290000e

MichaelBuessemeyer:
Additional comments that do not really fit into the code:
- There is a naming mismatch / duplication: `cellId` has the same meaning as `segmentId`. Sometimes the name `cellId` (in old code) is used and the name `segmentId` is used in the new code. In some code passages, it makes more sense to use one name instead of the other one. But it might not be intuitive that the names actually mean the same.

frontend/javascripts/oxalis/model/volumetracing/volume_annotation_sampling.js
```diff
@@ -333,24 +345,269 @@ export function applyVoxelMap(
...
             voxelToLabel[thirdDimensionIndex] =
               (voxelToLabel[thirdDimensionIndex] + sliceCount) % constants.BUCKET_WIDTH;
             // The voxelToLabel is already within the bucket and in the correct resolution.
             const voxelAddress = dataCube.getVoxelIndexByVoxelOffset(voxelToLabel);
-            if (shouldOverwrite || (!shouldOverwrite && data[voxelAddress] === overwritableValue)) {
+            const currentSegmentId = data[voxelAddress];
+            if (shouldOverwrite || currentSegmentId === overwritableValue) {
+              if (
+                isBucketInLowestResolution &&
+                currentSegmentId > 0 &&
```
MichaelBuessemeyer, line 371: This line should be removed, as erasing (overwriting with id 0) also overwrites the segment id at the current position (`voxelAddress`).

frontend/javascripts/oxalis/model/volumetracing/volume_annotation_sampling.js
```diff
@@ -333,24 +345,269 @@ export function applyVoxelMap(
...
+      continue;
+    }
+    const currentSegmentPosition = currentSegmentEntry.somePosition;
+    promisesToWaitFor.push(
+      checkIfSegmentPositionIsNowInvalid(
+        currentSegmentId,
+        currentSegmentPosition,
+        bucketAddressSet,
+      ).then(async (isSegmentPositionFaulty: boolean) => {
+        await checkBucketAddressSetForBucketsWithoutPartsOfSegment(
```
MichaelBuessemeyer, line 591: Consider renaming to `checkIfBucketsStillContainSegment`

philippotto:
@fm3 The back-end doesn't need any migration when deploying this, does it?

fm3:
> @fm3 The back-end doesn't need any migration when deploying this, does it?

Correct, but note that the tracingstore module needs to be upgraded with the main instance

philippotto:
> Correct, but note that the tracingstore module needs to be upgraded with the main instance

Ok, cool :)

@daniel-wer When you've checked my last commits and have given your final "Go", I'd merge this PR today :)

daniel-wer (approved):
LGTM, works beautifully :+1:

Issue #5708, opened 2021-09-06
Copying segment ID to clipboard does not work 


philippotto:
[Link to Discuss](https://discuss.webknossos.org/t/copying-segment-id-to-clipboard-does-not-work/1681)

PR #5709, opened 2021-09-06
Fix copying of segment ids

philippotto:
- the clipboard-js library we used was quite old. I removed in favor of the native navigator clipboard interface (see https://caniuse.com/mdn-api_clipboard_writetext)
- the old and new clipboard api don't accept numbers as inputs which is why I added the `toString` invocation where necessary

- https://fixcopyingnumbers.webknossos.xyz

### Steps to test:
- open a dataset with a segmentation
- rightclick on a segment and click the "copy segment id" icon
- the clipboard should update with the id (note that the toast itself is not enough as a success indicator, as the buggy version also showed the toast without copying anything)

### Issues:
- fixes https://github.com/scalableminds/webknossos/issues/5708

------
- [x] Ready for review

### DIFF
--- a/CHANGELOG.unreleased.md
+++ b/CHANGELOG.unreleased.md
@@ -18,6 +18,7 @@ For upgrade instructions, please check the [migration guide](MIGRATIONS.released
 
 ### Fixed
 - Fixed a bug where existing tasktypes with recommended configurations still had a property that is no longer valid. [#5707](https://github.com/scalableminds/webknossos/pull/5707)
+- Fixed that segment IDs could not be copied to the clipboard. [#5709](https://github.com/scalableminds/webknossos/pull/5709)
 
 ### Removed
 -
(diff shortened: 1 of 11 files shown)


philippotto:
> I removed in favor of the native navigator clipboard interface (see https://caniuse.com/mdn-api_clipboard_writetext)

@daniel-wer I thought we had a config somewhere (in webpack?) saying "the last X browser versions should be supported". However, I didn't find anything like that. Do you know what our policy on this is currently? I think the 92% which caniuse reports are sufficient, but I wanted to doublecheck this with you.

daniel-wer:
> > I removed in favor of the native navigator clipboard interface (see https://caniuse.com/mdn-api_clipboard_writetext)
> 
> @daniel-wer I thought we had a config somewhere (in webpack?) saying "the last X browser versions should be supported". However, I didn't find anything like that. Do you know what our policy on this is currently? I think the 92% which caniuse reports are sufficient, but I wanted to doublecheck this with you.

The config can be found in the `.babelrc`:

```
[
      "@babel/preset-env",
      {
        "targets": {
          "browsers": [
            "last 3 Chrome versions",
            "last 3 Firefox versions",
            "last 2 Edge versions",
            "last 1 Safari versions",
            "last 1 iOS versions"
          ]
        },
        "modules": false
      }
    ]
```

but I'm not sure whether that actually does something :see_no_evil:

daniel-wer (approved):
Nice, works like a charm, tested in newest Firefox/Chrome. :+1:

PR #5718, opened 2021-09-08
Fix volume version restore for untouched buckets

### LINKED ISSUE #5717: Volume Version Restore does nothing
In hybrid tracing, we clicked restore on an old version in the volume version history view. An update request was sent, but we did not see any changes (after refresh the data from before the reversion is loaded). We *do* see the version restore update action in the update action log. Needs investigation

fm3:
`VolumeTracingService.revertToVolumeVersion` checked against “Empty” to select which buckets need to be restored to empty. However, the fossildb client gave Failure instead of Empty for missing buckets. This is why the version restore skipped those buckets.

- https://volumerestore.webknossos.xyz

### Steps to test:
- Create volume annotation, brush some, save, then restore version 0, reload page, should still look like version 0

### Issues:
- fixes #5717 

------
- [x] Ready for review


jstriebel (approved):
code LGTM :+1:

PR #5721, opened 2021-09-09
Fix isMappingEnabled api method

daniel-wer:
I checked the api methods for other occurrences of this kind of bug but this seemed to be the only one :)

- https://___.webknossos.xyz

### Steps to test:
- Open the `ROI2017_wkw` dataset
  - Execute `webknossos.apiReady().then(async api => api.data.activateMapping("astrocyte"))` and wait for activation
  - Then execute `webknossos.apiReady().then(async api => api.data.isMappingEnabled())` which should return true

------
- [x] Ready for review


philippotto (approved):
:pray: :pray: :pray:

PR #5724, opened 2021-09-09
Fix unsafe integer usage by using string keys instead

### LINKED ISSUE #5158: 3D viewport shows orphan nodes and edges after splitting tree and deleting one component
[Link to Discuss](https://discuss.webknossos.org/t/3d-viewport-shows-orphan-nodes-and-edges-after-splitting-tree-and-deleting-one-component/1642)

daniel-wer:
In skeleton.js a recursive cantor pairing was used to combine three integers into a unique single integer which is used as a Map key. However, these numbers quickly become very large, exceeding javascript's `Number.MAX_SAFE_INTEGER`. In that case, the Map access to delete edges will fail sometimes resulting in wrongly colored or left over edges.
I switched to a simple string concatenation approach to create the unique keys. According to a quick search this shouldn't be much slower and I didn't notice any slow-downs.

- https://___.webknossos.xyz

### Steps to test:
- Open an empty tracing and drag-and-drop this NML into it
[orphan-nodes.zip](https://github.com/scalableminds/webknossos/files/7137619/orphan-nodes.zip)
  - Make sure the tracing is empty so that the IDs are not relabeled!
- Jump to position 198722, 346458, 5801 and reload the page to fix an issue with the 3D view if annotations are far outside of the dataset
- Remove the node there, then delete the smaller of the two agglomerates that were created by the split
  - On master you'll see some left over edges without nodes and warnings in the console (`[Skeleton] Unable to find buffer position for id 78519564543334160`, ...)
  - On this branch the agglomerate should be correctly deleted without any left over edges

### Issues:
- fixes #5158 

------
- [x] Ready for review

### DIFF
--- a/CHANGELOG.unreleased.md
+++ b/CHANGELOG.unreleased.md
@@ -22,6 +22,7 @@ For upgrade instructions, please check the [migration guide](MIGRATIONS.released
 - Fixed a bug where existing tasktypes with recommended configurations still had a property that is no longer valid. [#5707](https://github.com/scalableminds/webknossos/pull/5707)
 - Fixed that segment IDs could not be copied to the clipboard. [#5709](https://github.com/scalableminds/webknossos/pull/5709)
 - Fixed a bug where volume annotation version restore skipped buckets that were not yet touched in the version to be restored. [#5717](https://github.com/scalableminds/webknossos/pull/5717)
+- Fixed a rendering bug showing non-existant or wrongly-colored edges that sometimes occurred after deleting edges, nodes, or trees. [#5724](https://github.com/scalableminds/webknossos/pull/5724)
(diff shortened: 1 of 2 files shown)


philippotto (approved):
Awesome fix! Great testing steps, too :+1: Worked perfectly.

>  and reload the page to fix an issue with the 3D view if annotations are far outside of the dataset

I did not see this for master nor for this branch. Is this something we should take care of?

daniel-wer:
> I did not see this for master nor for this branch. Is this something we should take care of?

No, I don't think so. I have only ever encountered this during development where I sometimes use NMLs from much larger datasets. This is not a real-world use case, I would say.

PR #5725, opened 2021-09-09
Fix datastore deletion

youri-k:
- [x] Ready for review


jstriebel (approved):
:+1:

jstriebel:
> Is there a specific reason why admins should not be allowed to delete datastores?

According to @youri-k everybody can become admin in their own orga easily (e.g. default signup on webknossos.org), so this might be a bit to relaxed.

PR #5728, opened 2021-09-13
[WIP] Remove some dead code

philippotto:
- https://___.webknossos.xyz

### Steps to test:
I tested ortho and flight mode to check whether data loading works as before.

------
- [ ] Ready for review

### DIFF
--- a/frontend/javascripts/oxalis/api/api_latest.js
+++ b/frontend/javascripts/oxalis/api/api_latest.js
@@ -1385,7 +1385,6 @@ class DataApi {
     }
 
     segmentationLayer.cube.pushQueue.push();
-    segmentationLayer.cube.trigger("volumeLabeled");
   }
 
   /**
--- a/frontend/javascripts/oxalis/api/api_v2.js
+++ b/frontend/javascripts/oxalis/api/api_v2.js
@@ -686,7 +686,6 @@ class DataApi {
     }
 
     segmentationLayer.cube.pushQueue.push();
-    segmentationLayer.cube.trigger("volumeLabeled");
   }
 
   /**
--- a/frontend/javascripts/oxalis/controller/viewmodes/arbitrary_controller.js
+++ b/frontend/javascripts/oxalis/controller/viewmodes/arbitrary_controller.js
@@ -31,13 +31,11 @@ import {
 import ArbitraryPlane from "oxalis/geometries/arbitrary_plane";
 import ArbitraryView from "oxalis/view/arbitrary_view";
 import Crosshair from "oxalis/geometries/crosshair";
-import Model from "oxalis/model";
 import Store from "oxalis/store";
 import TDController from "oxalis/controller/td_controller";
 import Toast from "libs/toast";
 import * as Utils from "libs/utils";
 import api from "oxalis/api/internal_api";
-import app from "app";
 import constants, { ArbitraryViewport, type ViewMode, type Point2 } from "oxalis/constants";
 import getSceneController from "oxalis/controller/scene_controller_provider";
 import messages from "messages";
@@ -274,8 +272,6 @@ class ArbitraryController extends React.PureComponent<Props> {
   }
 
   bindToEvents(): void {
-    this.startListeningToBuckets();
-
     this.storePropertyUnsubscribers.push(
       listenToStoreProperty(
         state => state.userConfiguration,
@@ -307,23 +303,6 @@ class ArbitraryController extends React.PureComponent<Props> {
         },
       ),
     );
-  }
-
-  onBucketLoaded = () => {
-    this.arbitraryView.draw();
-    app.vent.trigger("rerender");
-  };
-
-  startListeningToBuckets() {
-    for (const dataLayer of Model.getAllLayers()) {
-      dataLayer.cube.on("bucketLoaded", this.onBucketLoaded);
-    }
-  }
-
-  stopListeningToBuckets() {
-    for (const dataLayer of Model.getAllLayers()) {
-      dataLayer.cube.off("bucketLoaded", this.onBucketLoaded);
-    }
   }
 
   start(): void {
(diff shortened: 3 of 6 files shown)


frontend/javascripts/test/model/binary/cube.spec.js
```diff
@@ -238,16 +238,3 @@ test("Garbage Collection should not collect buckets with shouldCollect() == fals
   const addresses = cube.buckets.map(b => b.zoomedAddress);
   t.deepEqual(addresses, [[0, 0, 0, 0], [3, 3, 3, 0], [2, 2, 2, 0]]);
 });
-
-test("Garbage Collection should not throw an exception if no bucket is collectable", t => {
```
philippotto, line 242: I removed this test, since I find my new test (which provokes the error) more important than this one. For keeping both, I didn't have a good idea, and I also think that it's not super valuable.

tools/test.sh
```diff
@@ -11,6 +11,15 @@ fi
...
 shift;
 
+# Pass params to prepare, but omit potential --timeout
+# parameters as babel does not accept these.
+additionalParamsWithoutTimeout=( )
+for arg in $@; do
+  shift
+  [ "${arg:0:9}" = "--timeout" ] && continue
+  additionalParamsWithoutTimeout+=("$arg")
+done
```
daniel-wer, line 21: Depending on the result of our discussion in https://github.com/scalableminds/webknossos/pull/5730 this might can be removed

tools/test.sh
```diff
@@ -38,14 +47,19 @@ function ensureUpToDateTests {
...
-  export NODE_PATH="$testBundlePath" && BABEL_ENV=test nyc --silent --no-clean ava $(find "$testBundlePath" -name "*.spec.js") "$@"
+  export NODE_PATH="$testBundlePath" && BABEL_ENV=test nyc --silent --no-clean --exclude binaryData ava $(find "$testBundlePath" -name "*.spec.js") "$@"
+elif [ $cmd == "test-debug" ]
+then
+  export NODE_PATH="$testBundlePath" && BABEL_ENV=test ava debug $(find "$testBundlePath" -name "*.spec.js") "$@"
 elif [ $cmd == "test-e2e" ]
 then
   ensureUpToDateTests
-  export NODE_PATH="$testBundlePath" && BABEL_ENV=test nyc --silent --no-clean ava $(find "$testBundlePath" -name "*.e2e.js") --serial -C 1 "$@"
+  # export NODE_PATH="$testBundlePath" && BABEL_ENV=test nyc --silent --no-clean ava $(find "$testBundlePath" -name "*.e2e.js") --serial -C 1 "$@"
```
daniel-wer, line 62:
```suggestion
  export NODE_PATH="$testBundlePath" && BABEL_ENV=test nyc --silent --no-clean ava $(find "$testBundlePath" -name "*.e2e.js") --serial -C 1 "$@"
```

frontend/javascripts/test/sagas/volumetracing_saga_integration.spec.js
```diff
@@ -64,14 +84,223 @@ test.serial(
...
-  Store.dispatch(startEditingAction([12, 12, 12], OrthoViews.PLANE_XY));
+  Store.dispatch(startEditingAction(paintCenter, OrthoViews.PLANE_XY));
+  Store.dispatch(addToLayerAction(paintCenter));
   Store.dispatch(finishEditingAction());
 
-  // maxCellId should be updated after brushing/tracing
-  getVolumeTracingOrFail(Store.getState().tracing).map(tracing => {
-    t.is(tracing.maxCellId, newCellId);
-  });
+  t.is(await t.context.api.data.getDataValue("segmentation", paintCenter), newCellId);
```
daniel-wer, line 105: You've probably considered this but just in case you are not aware. The `getDataFor2DBoundingBox` api function could also be used here to get a full snapshot comparison of a particular bounding box. Maybe even using ava's snapshot functionality :thinking:

frontend/javascripts/test/sagas/volumetracing_saga_integration.spec.js
```diff
@@ -64,14 +84,223 @@ test.serial(
...
+  },
+);
+
+test.serial(
+  "Brushing/Tracing should send buckets to backend and restore dirty flag afterwards",
+  async t => {
+    t.plan(1);
+    await testLabelingManyBuckets(t, true);
+    // If nothing crashed, the test was successful.
+    t.true(true);
```
daniel-wer, line 255: Wouldn't you need to test `t.false(hasRootSagaCrashed());` as the error is catched in the `root_saga`?

daniel-wer (commented):
Wow, this is awesome! Thank you for deep-diving into this and spending the time to remove left-over code as well as kick-starting the real:tm: volume tracing integration testing. The work in this PR will serve as the base for further volume tracing testing in the future, for example, the up/downsampling code, redo and the non-brushing modes.

As I wrote I feel like snapshot testing could make this even more solid. Would be interested in what you think about that or whether it's too complicated.

frontend/javascripts/test/sagas/volumetracing_saga_integration.spec.js
philippotto: Good point!

frontend/javascripts/test/sagas/volumetracing_saga_integration.spec.js
philippotto: Cool idea! Didn't think of that. I added it now.

philippotto:
After fixing a new race condition, I found yet another one. I pushed a fix, but wouldn't want to merge it today/tomorrow. Since I'm on vacation next week, it's probably a good idea to have another look afterwards to reduce the risk of any other problems :see_no_evil:

philippotto:
Hm, the tests succeed locally for me, but the CI fails for the new `Brushing/Tracing should send buckets to backend and restore dirty flag afterwards` test :shrug: Maybe I'll have an idea about this after my vacation.

daniel-wer (approved):
:+1:

philippotto:
Closing this since the changes are included in https://github.com/scalableminds/webknossos/pull/5733.

Issue #10, opened 2013-08-27
[OX-761] Time-Tracker


tmbo:
<b><i>[reporter="philipp.otto", created="Tue, 20 Aug 2013 01:59:44 +0200"]</i></b>

<a href="https://timer.scm.io/repos/537237ce1a00001a003c4a14/issues/10/create?referer=github" target="_blank">Log Time</a>

PR #5730, opened 2021-09-14
Refactor url hash to json

### LINKED ISSUE #4596: Refactor URL parameter scheme
The comma separated value list should be replaced with key-value pairs (e.g. via GET parameters). Alternatively, JSON could be used, but this might be overkill.

An important requirement for this change is the backwards compatibility with old URLs. We should add tests for this.

daniel-wer:
Refactor the url hash/fragment to alternatively use json format instead of a comma-delimited custom format. I added quite a few tests to make sure that both the csv format as well as the comment links (which also temporarily change the url hash) continue to work. Keep the csv format as the default.

- [x] Certain characters in the fragment string need to be escaped (see https://stackoverflow.com/questions/2849756/list-of-valid-characters-for-the-fragment-identifier-in-an-url). For now, I've used the built-in `encodeURIComponent` to do that. However, the function encodes quite a few characters which don't need to be encoded, resulting in a rather unreadable url hash.  We could write a custom {en,de}coding wrapper which doesn't encode `:`, `.` and `,` to make the hash more readable. On the other hand, we don't really want users to modify the url hash anyways, I'd say. What do you think?

- https://___.webknossos.xyz

### Steps to test:
- Open a url with a hash in the **csv** format and check that it opens correctly and the url state is correctly applied. E.g. append `#10,20,30,0,1.3,2` (`x,y,z,mode,zoomStep,activeNode`)
  - Modify values in the url hash json format
    - Check that the state is applied after pressing Enter
- Open a url with a hash in the **json** format and check that it opens correctly and the url state is correctly applied. E.g. append `#%7B"position"%3A%5B445%2C358%2C501%5D%2C"mode"%3A"orthogonal"%2C"zoomStep"%3A2%2C"activeNode"%3A2%7D`
  - Modify values in the url hash json format
    - Check that an error is thrown if the json format is invalid
- Check that the url hash works in flight mode or when there is no active node
- Create comments with links, for example `jump to node with id #2` and `jump to position #(1,2,3)`. Then click on the links and check that the links work as expected.

### Issues:
- fixes #4596 

------
- [x] Ready for review


tools/test.sh
```diff
@@ -20,7 +20,8 @@ function prepare {
   pbjs -t json "webknossos-datastore/proto/SkeletonTracing.proto" > "$testBundlePath/SkeletonTracing.proto.json"
   pbjs -t json "webknossos-datastore/proto/VolumeTracing.proto" > "$testBundlePath/VolumeTracing.proto.json"
   # --copy-files will copy files that are present in the source dir but are not transpiled (e.g.: json files)
-  BABEL_ENV=test babel "$jsPath" --out-dir "$testBundlePath" --copy-files
+  # "$@" inside this function refers to all arguments of the prepare function, not all arguments of this bash script
+  BABEL_ENV=test babel "$jsPath" --out-dir "$testBundlePath" --copy-files "$@"
```
philippotto, line 24: Cool that this seems to work! I also added the `"$@"` part in my other PR, since I noticed that `--watch` was broken for prepare. However, I used an older approach which we had some time ago in `test.sh` (then, we copied all the CLI arguments at the top level to another var). When doing this, `babel` would complain as it also received a `--timeout` parameter. To fix this I'm filtering that parameter away manually. See https://github.com/scalableminds/webknossos/pull/5728/commits/f3b47fa74a48e442bf5146acd23d023f8b55787a. It's quite ugly, but seems to work.

Your variant seems to work, too, because the CLI parameters are only passed when `test.sh prepare` was invoked. I'm a bit disappointed, that my miserable bash session was in vain apparently, but I guess your solution is way better :sweat_smile: I wonder whether there is another argument for always passing all arguments to `prepare`. But probably not..

daniel-wer: > I wonder whether there is another argument for always passing all arguments to prepare.

What do you mean with that? Or asked differently, what shortcomings do you see with the solution in this PR? I didn't encounter a situation yet where I wanted to pass additional parameters to `prepare` if I executed any of the `test.sh` commands other than `prepare`.

philippotto: Yeah, you are right. I just wanted to say that I'm not 100% sure whether there are use cases where another approach would be better, but I also cannot come up with any. I copied your approach to my PR :)

frontend/javascripts/oxalis/controller/url_manager.js
```diff
@@ -103,28 +131,33 @@ class UrlManager {
...
       .chain(skeletonTracing => getActiveNode(skeletonTracing))
-      .map(node => [node.id])
-      .getOrElse([]);
+      .map(node => ({ activeNode: node.id }))
+      .getOrElse({});
+
+    // $FlowIssue[exponential-spread] See https://github.com/facebook/flow/issues/8299
+    const urlState = { position, mode, zoomStep, ...rotationOptional, ...activeNodeOptional };
 
-    return [...position, viewModeIndex, zoomStep, ...rotation, ...activeNodeId].join(",");
```
philippotto, line 120: From our verbal discussion: The legacy variant (maybe we should call it `CSV` since we'll stick to it) should be still supported and also used here  for updating the URL (probably the component would need to support both ways for building the hash; the new variant would be used by the share modal).

philippotto (changes requested):
Really great stuff! Great that you also added extensive tests :+1: 

From our discussion today, there are some remaining todos (namely, sticking to the legacy URL by default), but other than that everything looks pretty good :+1: From my POV you could add the mapping-support to this PR or open a new one for that. Your call :game_die:

philippotto:
> We could write a custom {en,de}coding wrapper which doesn't encode :, . and , to make the hash more readable.

I think, this would be great, but that could be a follow-up once we are annoyed enough by it (building the new variant backwards compatible should not be a problem in my opinion).

philippotto (approved):
Code looks great and it also works well for me :)

PR #5742, opened 2021-09-23
render docs via wk-libs using mkdocs

jstriebel:
### Steps to test:
- Check the new docs as rendered with https://github.com/scalableminds/webknossos-libs/pull/422:
  https://static.webknossos.org/lib-docs/integrate_wk_docs/webknossos/index.html

This should only be merged after https://docs.webknossos.org/ is deployed via webknossos-libs


jstriebel:
The frontend linter fails with this PR, I have no clue why… Can you help @daniel-wer, @normanrz?

daniel-wer:
Fixed in https://github.com/scalableminds/webknossos/pull/5743

normanrz (approved):

PR #5743, opened 2021-09-23
Fix CI

daniel-wer:
A direct push to the master resulted in a failing prettier command. This probably wasn't noticed, because the linting and format check tasks are not run for master builds. However, builds are now failing for all branches that pulled the master.

- https://___.webknossos.xyz

### Steps to test:
- CI

------
- [x] Ready for review


jstriebel (approved):
Thanks :+1:

PR #5746, opened 2021-09-28
Fix crashes related to maximizing/minimizing

### LINKED ISSUE #5714: Maximizing 3D Viewport sometimes breaks webKnossos
When I maximize and unmaximize the 3D viewport, there's a problem: 1) tab chrashes or 2) the viewport becomes small, but ThreeJS Canvas stays maximized with the meshes.

image

https://webknossos.org/datasets/scalable_minds/l4dense_motta_et_al_demo_v2/view#2598,4693,1792,0,1.000

Chrome and Safari (Mac OS X)

daniel-wer:
- Fix a crash when maximizing in flight mode (not part of the issue, but encountered while I was at it)
  - In-depth explanation: When maximizing the flight mode, the `activeViewport` was set to `undefined` leading to an error when calculating the global position later. `activeViewport` is assumed to always be defined.
- Fix a crash when minimizing the 3D-View in datasets with lots of magnifications (10+)
  - In-depth explanation: When the 3D-View was maximized, the bucket picking still tried to select buckets. However, due to an area of 0, the results was only a single bucket for each viewport. Its position was 0, 0, 0 and its magnification was the lowest available one, for example 10. Then when the 3D-View was minimized again, the `getBaseBucketsForFallbackBucket` function is called to compute `all bucket addresses for which the fallback bucket is the provided bucket.` The fallback bucket in question is the one in mag 10 whereas the current mag is 1, leading to way too many bucket addresses being computed and usually an OOM error.

- https://___.webknossos.xyz

### Steps to test:
- To reproduce the flight mode maximizing crash: Open a tracing, switch to flight mode and then maximize the viewport. The page should not crash.
- Reproducing the other crash is only possible for datasets with lots of mags (see for example https://webknossos.org/datasets/scalable_minds/l4dense_motta_et_al_demo_v2/view#2598,4693,1792,0,1).
  - Open the dataset in view mode and maximize the 3D-View, then minimize it again (on master wK freezes and eventually OOMs).
  - I checked that the 0, 0, 0 bucket in the lowest resolution is no longer selected during bucket picking (thus avoiding the OOM) by setting a breakpoint in `layer_rendering_manager` line 238 checking the result of `consumeBucketsFromArrayBuffer`.

### Issues:
- fixes #5714 

------
- [x] Ready for review

### DIFF
--- a/CHANGELOG.unreleased.md
+++ b/CHANGELOG.unreleased.md
@@ -25,6 +25,7 @@ For upgrade instructions, please check the [migration guide](MIGRATIONS.released
 - Fixed that segment IDs could not be copied to the clipboard. [#5709](https://github.com/scalableminds/webknossos/pull/5709)
 - Fixed a bug where volume annotation version restore skipped buckets that were not yet touched in the version to be restored. [#5717](https://github.com/scalableminds/webknossos/pull/5717)
 - Fixed a rendering bug showing non-existant or wrongly-colored edges that sometimes occurred after deleting edges, nodes, or trees. [#5724](https://github.com/scalableminds/webknossos/pull/5724)
+- Fixed an error during viewport maximization in flight mode. Also, fixed a crash during minimization of the 3D-View for datasets with lots of magnnifications. [#5746](https://github.com/scalableminds/webknossos/pull/5746)
 
 ### Removed
 -
--- a/frontend/javascripts/oxalis/model/bucket_data_handling/bucket_picker_strategies/orthogonal_bucket_picker.js
+++ b/frontend/javascripts/oxalis/model/bucket_data_handling/bucket_picker_strategies/orthogonal_bucket_picker.js
@@ -89,6 +89,11 @@ function addNecessaryBucketsToPriorityQueueOrthogonal(
     bottomRightVector[v] = areas[planeId].bottom;
     bottomRightVector[u] = areas[planeId].right;
 
+    const width = bottomRightVector[u] - topLeftVector[u];
+    const height = bottomRightVector[v] - topLeftVector[v];
+    // If the viewport is not visible, no buckets need to be added
+    if (width === 0 || height === 0) continue;
+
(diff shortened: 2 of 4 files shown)


philippotto (approved):
Very cool that you tracked this down and found a fix :100: :tada: 

I think the bug came into existence because of two factors:
- the bucket picker didn't catch the zero-bucket case (which you have fixed now)
- the texture bucket manager was happy to gather thousands and thousands of buckets even though this should never be necessary

The second factor is relatively new, I think, and stems from this PR: https://github.com/scalableminds/webknossos/pull/5674/files#diff-9005d9d1722fdaae1ed8348593ca56bbb599f4733ba312e639b319908c31b819R383
Before that PR, the texture bucket manager only allowed a zoom difference of 1. The PR lifted the limitation with the reasoning that a limit of 3 (see `getMaxZoomStepDiff()`) is acceptable and will never be violated, anyway.

I'm wondering whether the texture bucket manager should get another check using `getMaxZoomStepDiff()` in `_getBaseBucketAddresses`. Something like this:

```
_getBaseBucketAddresses(bucket: DataBucket, zoomStepDifference: number): Array<Vector4> {
    if (getMaxZoomStepDiff() < zoomStepDifference) {
      return [];
    }
    const resolutions = getResolutions(Store.getState().dataset);
    return getBaseBucketsForFallbackBucket(bucket.zoomedAddress, zoomStepDifference, resolutions);
  }
```

Strictly speaking, that check should never be necessary, but the current bug showed that the scenario can slip through. To avoid the huge computational burden in that error case, the additional check seems like a good safety net. Do you agree?

daniel-wer:
> Strictly speaking, that check should never be necessary, but the current bug showed that the scenario can slip through. To avoid the huge computational burden in that error case, the additional check seems like a good safety net. Do you agree?

Yep good idea, I added the safeguard!

philippotto:
Awesome :)

PR #5756, opened 2021-10-05
Delay opening context menu to support windows again

MichaelBuessemeyer:
Windows behaves very strangely when it comes to the context menu. The same right-click to open the context menu also somehow reaches the overlay of the context menu, although the context menu isn't even rendered by React at this moment. Additionally, this occurs on all browsers on windows and on no other operating system. The browsers under windows I tested were: Firefox, Chrome, and Opera.

To prevent this weird behaviour I delayed the state change, which causes the context menu to render. This circumvents the behaviour for all tested windows browsers and the code still works on Linux + firefox for me.

- https://___.webknossos.xyz

### Steps to test:
- Open a hybrid tracing.
- Play around with the context menu, it should open and close as expected. 
- Once the dev instance is up, please also test on a windows browser.

### Issues:
- none

------
- [x] Ready for review

### DIFF
--- a/CHANGELOG.unreleased.md
+++ b/CHANGELOG.unreleased.md
@@ -17,7 +17,7 @@ For upgrade instructions, please check the [migration guide](MIGRATIONS.released
 -
 
 ### Fixed
-- 
+- Fixed a bug that windows user could not open the context menu as it instantly closed after opening. [#5756](https://github.com/scalableminds/webknossos/pull/5756).
 
 ### Removed
 -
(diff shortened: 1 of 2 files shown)


fm3:
Great find!
Quick question: would it work with timeout 0 as well, or does the bug occur again then?

MichaelBuessemeyer:
>Quick question: would it work with timeout 0 as well, or does the bug occur again then?

Yeah 0 ms works too. Thanks for the suggestion. This is a bit cleaner that 1.
I also tested it with my browsers on windows and it works fine.

Issue #5695, opened 2021-08-30
[Follow up]: Multiple segmentation layers


philippotto:
Follow up for https://github.com/scalableminds/webknossos/pull/5683.

Bugs & Weirdness
- [ ] Toggling of segmentation layer via "3" does not work properly when there are multiple segmentatin layers
	- how should we remember which layer was active before?
- [ ]  "Create tracing" when viewing DS, simply picks first segmentation layer as a base
	- maybe default to new layer? <-- should be carefully considered
- [ ] Make hybrid simply picks first segmentation layer
- [ ] `segmentationPatternOpacity` is shared across all segmentation layers
- [ ] When having multiple segmentation layers and mesh files, the UI is a bit weird. For example, switching to another seg layer will hide loaded meshes in the meshes-tab, but they will still be visible in the 3D viewport.

More Features:
- [ ] Support showing multiple segmentation layers at once
- [ ] If multiple segmentation layers can be rendered, ensure that only one mapping can be enabled in the UI right now (as mapping needs 3 textures we don't want to do this n>1 times).
  - this is a bit tricky because `isEnabled` in store is decoupled from the UI button. We need to ensure that `getSegmentationLayerWithMappingSupport` already returns the correct value quite early.

normanrz:
>  "Create tracing" when viewing DS, simply picks visible segmentation layer as a base
> - [ ] also provide dropdown / modal as in dashboard to configure how the annotation looks
maybe default to new layer? <-- should be carefully considered

I think there should be one volume annotation layer per segmentation layer, so that each segmentation layer is editable.

philippotto:
> I think there should be one volume annotation layer per segmentation layer, so that each segmentation layer is editable.

I'd rather converge on a lazy-create approach, since the existence of a volume + fallback layer means that twice as many buckets are requested for that layer when viewing it. If the ad-hoc "add volume layer" functionality is smooth (i.e., does not require a reload), both approaches should/could feel very similar for the user.

normanrz:
> > I think there should be one volume annotation layer per segmentation layer, so that each segmentation layer is editable.
> 
> I'd rather converge on a lazy-create approach, since the existence of a volume + fallback layer means that twice as many buckets are requested for that layer when viewing it. If the ad-hoc "add volume layer" functionality is smooth (i.e., does not require a reload), both approaches should/could feel very similar for the user.

Alternatively, we could create zero volume annotations and add an "edit" button to segmentation layers within an annotation.

[...]

PR #5771, opened 2021-10-12
Support multiple volume layers per annotation

### LINKED ISSUE #5759: Support having multiple volume annotation layers

fm3:
- https://multiplevolumelayers.webknossos.xyz

### TODO Backend
 - [x] adapt schema to hold multiple tracing ids per annotation
 - [x] adapt create explorational parameters
 - [x] adapt annotation json writes
 - [x] transactionally write tracing ids in postgres
 - [x] decide how precedence over conflicting per-tracing info works → lexicographically smallest tracingID. when mutating, fetch from old precedence one. after reupload? → write into all objects.
 - [x] decide how (and when) merging works → aggressively merges all layers into one, follow-up issue for name matching
 - [x] resolve remaining compiler errors
 - [x] decide which routes apply on all vs selected tracings (unlinkFallback, downsample)
 - [x] decide where to store annotation layer name (postgres vs tracingstore)
 - [x] write evolution
 - [x] do we need an sql evolution reversion? Yes, with warning that it is lossy for multi-layer annotations
 - [x] adapt test database
 - [x] add route to rename layer (editAnnotationLayer)
 - [x] do we need routes to add and delete annotation layers (other than makeHybrid)? Follow-up issue #5802
 - [x] write legacy api adapters
 - [x] document api changes
 - [x] refresh snapshots (blocked by frontend)
 - [x] test all the things (blocked by frontend)

### TODO Frontend
- [X] adapt store structure and all consumers to list-of-volume-annotations
- [x] make wk core usable with the new structure for single-volumes
  - [X] opening & annotating volume annotations
  - [x] saving
- [x] make wk core usable with multiple volume annotations
- [x] fix saving
- [x] fix undo
- [x] bugs
  - [X] save saga crashes in markBucketsAsNotDirty
  - [x] pressing "3" crashes annotation
  - [x] dataset-view is broken
  - [x] request task fails
  - [x] lasso tool is broken
  - ~[ ] create new cell button only reacts after second click? reproduce by switching between multiple volume layers beforehand~ this is a focus issue which is already present on master
  - [x] volume layer saving crashes?
- [x] adapt merger mode
- ~[ ] adapt UI for creating a new annotation~  follow-up: #5855
- [x] rename layers
- [X] front-end volume import into annotation (~should it create a new layer?~ no, not now. it will behave as before and imports the volume tracing into the currently active one)
- [x] version view/restore
- [x] bounding box export
- [x] misc
  - [x] mesh file list loading indicator doesn't stop?
  - [x] ...
- ~[ ] unlink fallback layer? should it still be supported?~ follow-up: https://github.com/scalableminds/webknossos/issues/5695
- [x] fix tests
- [x] adapt front-end api
- [x] doublecheck all selector usages for getting a volume tracing
- [x] go through PR and extract remaining todo comments

### REST API Changes (v5)
 - new format for createExplorational: expects list of layer parameters (`typ: AnnotationLayerType, fallbackLayerName: Option[String], resolutionRestrictions: Option[ResolutionRestrictions], name: Option[String]`)
 - typ is of enum AnnotationLayerType (one of 'Skeleton', 'Volume') – capitalized, not 'hybrid'
 - Annotation json no longer has an optional skeletonTracingId and volumeTracingId, but instead a list of `annotationLayers` – each has `tracingId: String, typ: AnnotationLayerType, name: Option[String]`
 - New route `PATCH /annotations/:typ/:id/editLayer/:tracingId` expects object with `name: Optional[String]` (will replace the name with None in the db if the name is missing from the request)

### Steps to test:
I think, testing should be done after the first review round, in case applying the feedback causes new bugs.
Afterwards, two testing passes would make sense in my opinion:
- test that the default functionality works for skeleton/volume-only/hybrids (including tasks) **without** multiple volume layers
- test that WK with multiple volume layers. for this phase, it is necessary to comment in some code, since the UI doesn't expose this currently (will be done in a follow-up PR).

### Issues:
- fixes #5759 

------
- [x] Ready for review


conf/webknossos.latest.routes
```diff
@@ -117,7 +118,7 @@ PATCH         /annotations/:typ/:id/transfer                                   c
 GET           /annotations/:typ/:id/info                                       controllers.AnnotationController.info(typ: String, id: String, timestamp: Long)
 PATCH         /annotations/:typ/:id/makeHybrid                                 controllers.AnnotationController.makeHybrid(typ: String, id: String, fallbackLayerName: Option[String])
 PATCH         /annotations/:typ/:id/downsample                                 controllers.AnnotationController.downsample(typ: String, id: String)
-PATCH         /annotations/:typ/:id/unlinkFallback                             controllers.AnnotationController.unlinkFallback(typ: String, id: String)
+PATCH         /annotations/:typ/:id/unlinkFallback                             controllers.AnnotationController.unlinkFallback(typ: String, id: String, tracingId: String)
```
philippotto, line 121: @fm3 Should the `tracingId` be part of the URL as it's done for `editAnnotationLayer`? Just stumbled about this. I still need to adapt the front-end here and I'm wondering where to put the id :thinking:

frontend/javascripts/oxalis/view/action-bar/tracing_actions_view.js
```diff
@@ -300,8 +300,10 @@ class TracingActionsView extends React.PureComponent<Props, State> {
       throw Error(message);
     }
 
+    // todo: does this logic make sense at all? the above condition seems to exclude
```
philippotto, line 303: todo

frontend/javascripts/oxalis/view/version_view.js
```diff
@@ -90,15 +90,15 @@ class VersionView extends React.Component<Props, State> {
                 />
               </TabPane>
             ) : null}
-            {this.props.tracing.volume != null ? (
-              <TabPane tab="Volume" key="volume">
+            {this.props.tracing.volumes.map((volumeTracing, index) => (
+              <TabPane tab={`Volume ${index + 1}`} key={volumeTracing.tracingId}>
```
philippotto, line 94: todo: use volume layer name here

frontend/javascripts/admin/admin_rest_api.js
```diff
@@ -819,6 +868,8 @@ export async function downloadNml(
 export async function unlinkFallbackSegmentation(
   annotationId: string,
   annotationType: APIAnnotationType,
+  // todo
```
philippotto, line 871: todo

frontend/javascripts/oxalis/api/api_latest.js
```diff
@@ -894,12 +903,15 @@ class TracingApi {
...
     const state = Store.getState();
-    const { annotationId, annotationType, volume } = state.tracing;
+    const { annotationId, annotationType, volumes } = state.tracing;
     if (state.task != null) {
       throw new Error("Cannot downsample segmentation for a task.");
     }
-    if (volume == null) {
-      throw new Error("Cannot downsample segmentation for annotation without volume data.");
+    if (volumes.length !== 1) {
+      // todo: support downsampling volume layers even if there are multiple ones
```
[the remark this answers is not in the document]
philippotto: the back-end doesn't support this currently.

philippotto:
@daniel-wer I think, this PR is ready for a first round of review. Unfortunately, the changed line count is very high, but most of the changes are simply due to the changed access of the (relevant) volume layer(s). Also, lots of volume-related code got additional tracing IDs passed in (such as for the saving mechanism), so that (hopefully) nothing gets mixed up. In the meantime, I will tackle my own remaining todo-comments (however, there's nothing major in there, currently).

Also a word about layer names and layer IDs: I got quite confused by these in between, since we now also have proper human-readable layer names which the user can give. My brief summary would be as follows:
- volume tracings have a unique ID and a human-readable (and changable) name
- dataset layers have a name (which is defined by the folder name on disk)
- for volume tracings, additional dataset layers are created for which the name is set to the volume tracing ID. this part can be a bit confusing, since some functions accept a layer name (since they operate on all dataset layers) and will retrieve a volume tracing ID. maybe there's a more elegant way, but for now it seems to work well.

If you think, it makes sense we can also have a call to talk about the PR :call_me_hand: Looking forward to your feedback!

Ah, and as usually: hiding whitespace changes might make this PR a tad easier to review :eagle:

frontend/javascripts/oxalis/model/volumetracing/volumelayer.js
```diff
@@ -173,8 +179,7 @@ class VolumeLayer {
   }
 
   getContourList(useGlobalCoords: boolean = false) {
-    const volumeTracing = enforceVolumeTracing(Store.getState().tracing);
-    const globalContourList = volumeTracing.contourList;
+    const globalContourList = this.volumeTracing.contourList;
```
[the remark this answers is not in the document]
philippotto: yes, the lasso tool was broken because of this. I changed it to only store the id

frontend/javascripts/oxalis/model_initialization.js
```diff
@@ -268,32 +270,36 @@ function initializeTracing(_annotation: APIAnnotation, tracing: HybridServerTrac
...
     Toast.error(messages["tracing.no_allowed_mode"]);
   } else {
-    const isHybridTracing = tracing.skeleton != null && tracing.volume != null;
-    let maybeUrlViewMode = UrlManager.initialState.mode;
-    if (isHybridTracing && UrlManager.initialState.mode === constants.MODE_VOLUME) {
-      // Here we avoid going into volume mode in hybrid tracings.
-      maybeUrlViewMode = constants.MODE_PLANE_TRACING;
-    }
+    const maybeUrlViewMode = UrlManager.initialState.mode;
+    // todo: refactor MODE_VOLUME away or make this logic compatible
```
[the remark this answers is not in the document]
philippotto: @daniel-wer I think, I would simply remove MODE_VOLUME away. I skimmed the usages of the viewMode and `MODE_VOLUME` doesn't seem to have any relevance if I didn't miss anything. I only noticed that one can weirdly switch to the "volume" mode by pressing "m" repeatedly (the view mode dropdown doesn't even contain "volume" but when switching to that mode, it will show it anyway). Do you think there is a reason which speaks against removing that mode? Backwards-compatibility in the URL needs to be preserved obviously, but other than that, the removal seems like the a good idea to me.

frontend/javascripts/oxalis/api/api_latest.js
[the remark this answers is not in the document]
philippotto: @fm3 is it easy to make the downsampling route specific to a volume tracing id? or should we simply render multiple downsample buttons (if that edge case occurs) and tell the user that all layers will be downsampled (if necessary)?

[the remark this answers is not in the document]
fm3: I added a commit that changes the downsample route to expect the additional GET parameter `?tracingId=` (analogous to the unlinkFallback route)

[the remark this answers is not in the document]
philippotto: awesome, thank you!

frontend/javascripts/oxalis/model/reducers/volumetracing_reducer.js
```diff
@@ -132,41 +133,49 @@ function handleUpdateSegment(state: OxalisState, action: UpdateSegmentAction) {
...
+    userBoundingBoxes,
+  };
+  return volumeTracing;
+}
+
 function VolumeTracingReducer(state: OxalisState, action: VolumeTracingAction): OxalisState {
   switch (action.type) {
     case "INITIALIZE_VOLUMETRACING": {
       // As the frontend doesn't know all cells, we have to keep track of the highest id
       // and cannot compute it
```
daniel-wer, line 170: This should be moved to the `serverVolumeToClientVolumeTracing` function as well

frontend/javascripts/oxalis/model/sagas/volumetracing_saga.js
```diff
@@ -919,11 +940,39 @@ export function* maintainHoveredSegmentId(): Saga<void> {
...
+function* maintainContourGeometry(): Saga<void> {
+  yield* take("SCENE_CONTROLLER_READY");
+  const SceneController = yield* call(getSceneController);
+  const { contour } = SceneController;
+
+  while (true) {
+    yield* take(["ADD_TO_LAYER", "RESET_CONTOUR"]);
+    const isTraceToolActive = yield* select(state => isTraceTool(state.uiInformation.activeTool));
+    const volumeTracing = yield* select(getActiveSegmentationTracing);
+    if (!volumeTracing) {
```
daniel-wer, line 952: You could add `|| !isTraceToolActive` here to avoid the large indented if block below.

frontend/javascripts/oxalis/model/volumetracing/volumelayer.js
```diff
@@ -173,8 +180,8 @@ class VolumeLayer {
   }
 
   getContourList(useGlobalCoords: boolean = false) {
-    const volumeTracing = enforceVolumeTracing(Store.getState().tracing);
-    const globalContourList = volumeTracing.contourList;
+    const globalContourList = getVolumeTracingById(Store.getState().tracing, this.volumeTracingId)
+      .contourList;
```
daniel-wer, line 184: This looks weirdly formatted, but maybe that's the way to do it and I'm too tired :sweat_smile:

daniel-wer (commented):
First part of the review, I'll continue tomorrow :)

frontend/javascripts/oxalis/view/action-bar/tracing_actions_view.js
daniel-wer: Yes you're right it doesn't make a lot of sense, currently. My intention was to do it right:tm: to make it easier once sandbox copying is implemented for volumes as well, but as it's essentially untested code, it's probably better to simply pass `null` as the fallbackLayer for now.

frontend/javascripts/oxalis/model_initialization.js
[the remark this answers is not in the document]
daniel-wer: Agreed, seems like a leftover to me. :+1:

frontend/javascripts/oxalis/view/right-border-tabs/skeleton_tab_view.js
```diff
@@ -219,14 +221,33 @@ export async function importTracingFiles(files: Array<File>, createGroupForEachF
...
-          const oldVolumeTracing = tracing.volume;
+          const storeState = Store.getState();
+          const { tracing, dataset } = storeState;
+          if (tracing.volumes.length === 0) {
+            throw new Error("A volume tracing must already exist when importing a volume tracing.");
+          }
+          const oldVolumeTracing = getActiveSegmentationTracing(storeState);
+          if (oldVolumeTracing == null) {
+            throw new Error(
+              "Ensure that a volume tracing layer is visible when importing a volume tracing.",
```
daniel-wer, line 232: In theory, this could be extended now by asking the user whether the volume tracing should be merged to the active volume tracing layer or be imported as a separate volume tracing layer. Maybe we can add this to a followup issue.

frontend/javascripts/oxalis/view/version_list.js
```diff
@@ -54,10 +54,24 @@ export async function previewVersion(versions?: Versions) {
   await api.tracing.restart(annotationType, annotationId, controlMode, versions);
   Store.dispatch(setAnnotationAllowUpdateAction(false));
 
-  const segmentationLayer = Model.getSegmentationTracingLayer();
-  const shouldPreviewVolumeVersion = versions != null && versions.volume != null;
-  const shouldPreviewNewestVersion = versions == null;
-  if (segmentationLayer != null && (shouldPreviewVolumeVersion || shouldPreviewNewestVersion)) {
+  if (versions == null) {
```
daniel-wer, line 57: Just a suggestion (untested):

```javascript
  const segmentationLayersToReload = []
  if (versions == null) {
    // No versions were passed which means that the newest annotation should be
    // shown. Therefore, reload all segmentation layers.
    segmentationLayersToReload.push(...Model.getSegmentationTracingLayers());
  } else if (versions.volumes != null) {
    // Since volume versions were specified, reload the volumeTracing layers
    const versionedSegmentationLayers = Object.keys(versions.volumes).map(volumeTracingId =>
      Model.getSegmentationTracingLayer(volumeTracingId)
    );
    segmentationLayersToReload.push(...versionedSegmentationLayers);
  }
  
  for (const segmentationLayer of segmentationLayersToReload) {
    segmentationLayer.cube.collectAllBuckets();
    segmentationLayer.layerRenderingManager.refresh();
  }
```

daniel-wer (commented):
Very nice :+1: I finished the code review, but would wait with testing until the CI runs through.

Seeing the SaveState type, I was wondering whether it might make sense to generalize the tracing layer handling:

```javascript
export type SaveState = {
  ...
  +queue: {
    +skeleton: Array<SaveQueueEntry>,
    +volumes: {
      [tracingId: string]: Array<SaveQueueEntry>,
    },
  },
  +lastSaveTimestamp: {|
    +skeleton: number,
    +volumes: {
      [tracingId: string]: number,
    },
  |},
}
```
could become
```
export type SaveState = {
  ...
  +queue: {
    [tracingId: string]: Array<SaveQueueEntry>,
  },
  +lastSaveTimestamp: {
    [tracingId: string]: number,
  },
}
```
but it might be that the distinction between volume and skeleton is still needed in too many places.

In general, with the generalization work being done in previous PRs and this one, I was wondering whether it would be possible to simplify some of the architecture that evolved over time. However, I think some of the work from the followup issue is needed for that. Still, at some point we should take a step back and think about opportunities where we could remove some of the complexity.

frontend/javascripts/oxalis/model/volumetracing/volumelayer.js
philippotto: I guess, we have to trust prettier on this :smile:

frontend/javascripts/oxalis/api/api_latest.js
```diff
@@ -892,18 +901,15 @@ class TracingApi {
      Note that this invoking this method will not block the UI. Thus, user actions can be performed during the
      downsampling. The caller should prohibit this (e.g., by showing a not-closable modal during the process).
    */
-  async downsampleSegmentation() {
+  async downsampleSegmentation(volumeTracingId: string) {
```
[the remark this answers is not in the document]
philippotto: Yes, this is a bit annoying. In this particular case, I'd argue that the chances for this being used are < 0.1%. In general, my take on these things is:

- try to be backwards compatible as much as possible
- small breaking changes are ok, if the chance of it being used is very low

It's cumbersome, but feels like the best trade-off to me currently.

frontend/javascripts/oxalis/view/action-bar/toolbar_view.js
```diff
@@ -224,22 +224,25 @@ function AdditionalSkeletonModesButtons() {
...
+    getMappingInfoForVolumeTracing(state, volumeTracing != null ? volumeTracing.tracingId : null),
   );
 
-  const mappingColors = useSelector(state => getMappingInfoForTracingLayer(state).mappingColors);
-  const isMappingEnabled = useSelector(
-    state => getMappingInfoForTracingLayer(state).isMappingEnabled,
-  );
+  if (!volumeTracing) {
+    return null;
+  }
```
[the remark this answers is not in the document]
philippotto: Unfortunately, this is not possible, due to the rule of hooks that hooks may not come after conditionals. That's why the null-checks are so cumbersome here. I don't really see an elegant solution for this..

frontend/javascripts/oxalis/view/action-bar/toolbar_view.js
```diff
@@ -224,22 +224,25 @@ function AdditionalSkeletonModesButtons() {
...
+const mapId = (volumeTracing, id) => {
+  const { cube } = Model.getSegmentationTracingLayer(volumeTracing.tracingId);
   return cube.mapId(id);
 };
 
 function CreateCellButton() {
-  const unmappedActiveCellId = useSelector(
-    state => enforceVolumeTracing(state.tracing).activeCellId,
+  const volumeTracing = useSelector(state => getActiveSegmentationTracing(state));
+  const unmappedActiveCellId = volumeTracing != null ? volumeTracing.activeCellId : 0;
```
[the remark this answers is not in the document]
philippotto: See my other comment: https://github.com/scalableminds/webknossos/pull/5771#discussion_r758210717

frontend/javascripts/oxalis/view/right-border-tabs/skeleton_tab_view.js
philippotto: Good point. I added it to https://github.com/scalableminds/webknossos/issues/5695.

frontend/javascripts/oxalis/view/version_list.js
philippotto: Great, thank you!

frontend/javascripts/oxalis/merger_mode.js
```diff
@@ -356,9 +383,15 @@ export async function enableMergerMode(onProgressUpdate: number => void) {
   );
   // wait for preprocessing the already existing trees before returning
   await mergeSegmentsOfAlreadyExistingTrees(0, mergerModeState, onProgressUpdate);
+
+  if (mergerModeState.segmentationLayerName) {
+    api.data.setMappingEnabled(true, mergerModeState.segmentationLayerName);
```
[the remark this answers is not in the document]
philippotto: You are right. It does not seem to be necessary. I don't remember why I added it. I removed it now :)

philippotto:
> but it might be that the distinction between volume and skeleton is still needed in too many places.

Yes, it would probably result in better code, but I don't want to do this change in this PR, as that would blow it up even further.

> In general, with the generalization work being done in previous PRs and this one, I was wondering whether it would be possible to simplify some of the architecture that evolved over time. However, I think some of the work from the followup issue is needed for that. Still, at some point we should take a step back and think about opportunities where we could remove some of the complexity.

Totally agree :+1:

philippotto:
@daniel-wer I think, I incorporated all your feedback as good as I could. The CI is green now and I also just tested a bit.

With regard to my suggested testing plan, I'm just wondering whether it would be enough to just test the "one volume layer" case, as the UI in this branch doesn't really expose anything else. Alternatively, one could comment in some code, but theoretically we could postpone the multi-volume testing until the next PR where the UI exposes this. Not sure what makes more sense. Do you have an opinion on this?

daniel-wer:
> With regard to my suggested testing plan, I'm just wondering whether it would be enough to just test the "one volume layer" case, as the UI in this branch doesn't really expose anything else. Alternatively, one could comment in some code, but theoretically we could postpone the multi-volume testing until the next PR where the UI exposes this. Not sure what makes more sense. Do you have an opinion on this?

I don't have a strong opinion on that. Since the multi-layer functionality is not exposed yet it wouldn't be a problem to postpone the testing until it's properly usable in my opinion :)

daniel-wer (approved):
Awesome, testing went very well for me :+1:

jstriebel:
Seems like there was a change in the backend responses for annotations, which breaks a single test in wk-libs but isn't used anywhere else. Since the tests are snapshotted, everything is fine for now, we should just keep this in mind for the next time. I'll adapt the client code, the snapshots and the tests soon :tm:

philippotto:
Thank you for adapting the code in wk-libs :pray:

> [...] we should just keep this in mind for the next time.

Very good point! The newest PR template includes a "Adapted wk-libs python client if relevant API parts change"-checkbox, but the PR is older than that. So, hopefully, this error class won't appear that soon again.

Issue #5653, opened 2021-08-05
Nodes are shown in wrong slice when moving back and forth with low clipping distance


fm3:
## Context
I am annotating a dataset for alignment corrections. For this it is important to set nodes in the precise z coordinate of specific image tiles. For this I turn the clipping distance very low. I then move back and forth a lot by using the scroll wheel or d/f.

## Expected Behavior
Nodes with z=10 should be visible exactly when position’s z is shown as 10 in the toolbar.

## Current Behavior
Sometimes the node is shown in the neighboring slice and *not* in slice 10.

## Steps to Reproduce the bug
- Open https://webknossos.brain.mpg.de/annotations/Explorational/610a44ee0100008d118a58ad#1386,19353,2326,0,15.863,10
- Set clipping distance to 1 (what’s the unit here? Related issue #5566  )
- Select a node
- move around with d/f
- depending on the move direction the node is visible in different slices
- 
## Your Environment for bug
Firefox Linux

## Related issue
I suspect this is closely related to #5233 and the old PR https://github.com/scalableminds/webknossos/pull/5201, which is mentioned there

philippotto:
I think, we should try to fix #5233 (see my newest comment) first. I assume, this issue here will be either solved automatically or should be easier to grasp.

philippotto:
I implemented the approach I suggested in #5233 (PR: #5801). I still think that the change should soften the problem which is described in the current issue here. However, the actual problem is probably as follows:

- The clipping distance goes in both directions (so, if the clipping distance is large enough you can see nodes before *and* after the current slice). This is by design and should probably not change (even if we changed it, the actual problem wouldn't be solved either).
- Due to the floating point position, a fixed clipping distance cannot ensure that only nodes of the current slice are rendered. For example, if I'm at z=0.5, the clipping distance would need to be 0.5 to only see the node in the current slice. If I'm at z=0.6, it would need to be 0.4 in one direction and 0.6 in the other --> doesn't work.

For this particular use case ("I only want to see nodes in the current slice"), I would consequently suggest to:
- have a dedicated option à la "Only show nodes in the current slice"
- or implement the related issue #5766

For now, I suggest to wait a bit to see how #5801 will improve things (hopefully). Afterwards, we can decide how to proceed.

fm3:
I just noticed #4366 is a duplicate of this issue. closing it now, but to preserve the info, I’m quoting from the description there:

My scenario is quite rare I guess:
[...]

PR #5801, opened 2021-10-29
Improve handling of movements in 3rd dimension

### LINKED ISSUE #5233: Toggling between two slices does not visibly move other viewports
## Context
Viewing a dataset, toggling between two adjacent slices (either by alternating d/f or mousewheel), zoomed in to mag1

## Expected Behavior
The yz, xz viewports and 3d viewport should show in which direction the move was taking place

## Current Behavior
Those viewports do not visibly change

## Steps to Reproduce the bug
Viewing a dataset, toggling between two adjacent slices (either by alternating d/f or mousewheel), zoomed in to mag1

## Your Environment for bug
Firefox 86 Windows 10

## Note
https://github.com/scalableminds/webknossos/pull/5201 probably causes this. Philipp already explained that to solve this while keeping the desired behavior from there, the code would have to differentiate between holding keys pressed and hitting them singly. Let’s see if other users complain about this. If it’s just me, the priority is not high.

philippotto:
This PR removes the old "clamp to edge" behavior on initial third-dimension movements, since this had undesired side effects (such as #5233). As a replacement, the keyboard delay is adapted according to the current position. See code comments for more details.

https://fixvisiblemove.webknossos.xyz

### Steps to test:
- open a dataset (ideally with a real-world scale, such as 11,11,24)
- turn down the move value to something unbearably slow (e.g., 5 nm/s which I allowed temporarily for this PR)
- play around with the position field (which now supports floats in this PR <-- I will undo this)
  - set the z-value to x.99 and move forward with f (keep the key pressed)
  - set the z-value to x.00 and move forward with f (keep the key pressed)
  - the experienced keyboard delay for the first scenario should be higher than for the second. you can also check the console output to see the calculated delays.

### Issues:
- fixes #5233 
- hopefully softens #5653 

------
- [X] Ready for review


philippotto:
@fm3 Feel free to check out https://fixvisiblemove.webknossos.xyz to see whether it's an improvement with regard to https://github.com/scalableminds/webknossos/issues/5233.

fm3:
Thanks! Yes, my use case is handled better now :)

frontend/javascripts/oxalis/controller/viewmodes/plane_controller.js
```diff
@@ -172,31 +172,27 @@ function createDelayAwareMoveHandler(multiplier: number) {
     const state = Store.getState();
     let direction = Math.sign(multiplier);
 
-    const moveDistanceIn1Second =
-      state.userConfiguration.moveValue / getBaseVoxel(state.dataset.dataSource.scale);
     const { activeViewport } = state.viewModeData.plane;
+    const thirdDim = dimensions.thirdDimensionForPlane(activeViewport);
+    const voxelPerSecond =
+      state.userConfiguration.moveValue / state.dataset.dataSource.scale[thirdDim];
```
philippotto, line 178: @daniel-wer Using `getBaseVoxelFactors` didn't produce the correct result (the timing was still off). It turned out that it was even easier: just use the nm-based moveValue and divide it by the correct component of the dataset scale. Now, the delay should be right :)

daniel-wer (approved):
Very nice, feels exactly right now :)

LGTM once the temporary changes are reversed :+1:

PR #5824, opened 2021-11-03
Correctly propagate project tracing time as Long instead of Int

fm3:
Some project queries summed their task’s tracingTime (ms) values, which is too much for int. In the db, everything is Bigint already, but the scala code previously stored the result in Ints.

In the same query, I also changed sum(openInstances) from int to long, to match the db format, even though we don’t really expect that many instances.

- https://longtracingtime.webknossos.xyz

### Steps to test:
- Create at least one task
- Trace for 1200 hours (or manipulate the db)
- Project list should still be rendered with no error

------
- [x] Ready for review


youri-k (approved):
Nice catch!
LGTM :shipit:

PR #5831, opened 2021-11-04
Uploads survive backend restarts

### LINKED ISSUE #5056: Enable parallel upload of dataset chunks during import
might improve speed a bit. low-pri in my opinion. the following needs to be fixed then:

> The name-reserving logic of the wk backend leads to errors if upload chunk requests come in parallel. Since that fix looks nontrivial I suggest we disable parallel requests for the moment, to fix the uploads.

see https://github.com/scalableminds/webknossos/pull/5055 for context

### LINKED ISSUE #5150: Dataset Upload should Resume after Backend Restart
## Detailed Description
Currently, the datastore backend stores the information which chunks have already been received in a local variable. Thus, the information is lost when the datastore is restarted. Since this happens every time we upgrade the instance, it could potentially break long uploads.

If the information could be stored elsewhere (e.g. redis?), it could be persisted across restarts.

Currently the access of this information is made thread-safe via scalas `synchronize` primitive. We should check that no concurrency issues arise if we change this.

youri-k:
- https://robusteruploads.webknossos.xyz

### Steps to test:
- upload a dataset from the dashboard (try with single zip file, and with unpacked multi-file dataset)
- kill backend while upload is going
- restart backend
- => upload should continue and finish correctly 

### Follow-up
created follow up issue  #5849 for orphan cleanup

### Issues:
- fixes #5150 
- fixes #5056 

------
- [x] Ready for review


youri-k:
Open ToDo:
- [ ] Ensure redis is running (if upload enabled), so datastore can access it

fm3 (approved):
I took over this PR from @youri-k who is out of office atm.
I changed some naming, fixed a bug where cleanup code deleted still-alive items, added redis to the docker-compose file.
I also tested this with different datasets and scenarios. I noticed what I wrote down as follow-up issue #5849 but I think it is not going to block this.
@philippotto could you have a quick look at the front-end changes? The upload now starts with a “pre-flight” request reserving the dataset name.

frontend/javascripts/admin/admin_rest_api.js
```diff
@@ -1099,14 +1097,26 @@ export function createResumableUpload(
         chunkSize: 10 * 1024 * 1024, // set chunk size to 10MB
         permanentErrors: [400, 403, 404, 409, 415, 500, 501],
         // Only increase this value when https://github.com/scalableminds/webknossos/issues/5056 is fixed
```
philippotto, line 1099: this comment can be removed then, I assume :)

frontend/javascripts/admin/admin_rest_api.js
```diff
@@ -1099,14 +1097,26 @@ export function createResumableUpload(
...
         chunkRetryInterval: 2000,
         maxChunkRetries: undefined,
         generateUniqueIdentifier,
       }),
   );
 }
 
+export function reserveDatasetUpload(
+  datastoreHost: string,
+  reserveUploadInformation: {},
```
philippotto, line 1110: please add a type à la

`reserveUploadInformation: ReserveUploadInformation`

with

```
type ReserveUploadInformation = {
  uploadId: string,
  organization: string,
  name: string,
  totalFileCount: number,
  initialTeams: Array<string>,
}

export function reserveDatasetUpload...
```

philippotto (commented):
very cool :) i only left 2 nitpicks

jstriebel:
> Should I attempt to do that or do you want to do it?

It'd be great if you could give this a try :+1:

fm3:
https://github.com/scalableminds/webknossos-libs/pull/484

PR #5832, opened 2021-11-05
Add tags for datasets

### LINKED ISSUE #5829: Organization of datasets with tags
## Detailed Description
Similar to how tags can be assigned to annotations, tag-supported could be added to facilitate the management of datasets. Predominantly, this would allow:

- adding arbitrary meta data to datasets
- tag-based search and filtering

## Context

fm3:
Adding a tag system to Datasets, analogous to what Annotations already have.

- https://datasettags.webknossos.xyz

### Backend Changes
- Dataset json now contains additional field `tags`  (list of string)
- Dataset update route now expects `tags` as well (it being missing will be interpreted as empty list), changes tags in database

### Steps to test:
- Open the dataset table. There should be a new row for tags.
- Try adding some tags for a few datasets.
- Clicking on one of the tags should add the tag to the search and hide all datasets that do not have this tag.
- The clicked tag should appear on the top left of the dataset table.
- Clicking on close on the list of tags the datasets are filtered for, should remove that tag from the filtering
- Reloading the page should preserve the filtered tags.
- Adding or removing a tag from a dataset, then waiting a second for the backend call and then reloading the whole page, should show the dataset with the added/removed tag. 
- Reloading a dataset should also not change the displayed tags of a dataset.

### Issues:
- fixes #5829 

------
- [x] Ready for review


fm3:
@MichaelBuessemeyer The backend part should be done. I could only test it in a limited way so far, as there is no frontend yet. If you encounter problems, feel free to ping me :)

tools/postgres/schema.sql
```diff
@@ -98,6 +98,7 @@ CREATE TABLE webknossos.dataSets(
   logoUrl VARCHAR(2048),
   sortingKey TIMESTAMPTZ NOT NULL,
   details JSONB,
+  tags VARCHAR(256)[] NOT NULL DEFAULT '{}',
```
jstriebel, line 101: We could also use the `json` or `jsonb` type (seems like it's available from pg 9.6). It's also possible to create indices on jsonb columns and query them efficiently (e.g. when searching/selecting by tag):

https://www.postgresql.org/docs/9.4/datatype-json.html#JSON-INDEXING

jstriebel (commented):
Backend mostly LGTM, please see my two comments @fm3:

tools/postgres/schema.sql
jstriebel: But also totally fine as-is for now I guess.

app/models/binary/DataSetService.scala
```diff
@@ -374,6 +374,7 @@ class DataSetService @Inject()(organizationDAO: OrganizationDAO,
         "isUnreported" -> Json.toJson(isUnreported(dataSet)),
         "isForeign" -> dataStore.isForeign,
         "jobsEnabled" -> jobsEnabled,
+        "tags" -> dataSet.tags
```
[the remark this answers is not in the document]
fm3: So far we have not considered additional fields a breaking change. This would increase the number of new versions by a lot.
I agree that we should unify which routes are versioned and what constitutes a new version. I created a separate issue for this at https://github.com/scalableminds/webknossos/issues/5840

philippotto (changes requested):
The front-end code looks great :+1: However, I think that I've found a potential race condition (see my comment).

Other than that, I have some UI suggestions:
- turn the cursor into a pointer when hovering the tags (via CSS)
- show a tooltip which explains the tag interactions (e.g., "Click to only show datasets with this tag", "Add a new tag", "Remove this tag from this dataset")

Would be great if you could generalize these UI improvements so that the tags for the explorative annotations also profit from these :)

conf/evolutions/reversions/077-add-dataset-tags.sql
```diff
@@ -0,0 +1,11 @@
+START TRANSACTION;
+
+DROP VIEW webknossos.dataSets_;
+
+ALTER TABLE webknossos.dataSets DROP COLUMN tags;
+
+CREATE VIEW webknossos.dataSets_ AS SELECT * FROM webknossos.dataSets WHERE NOT isDeleted;
+
+UPDATE webknossos.releaseInformation SET schemaVersion = 76;
```
philippotto, line 9: the version number in the file name and in this line also need to be incremented by one (I think, this is why the CI fails)

fm3: Correct, since on master the version has now been incremented to 77, the verison in schema.sql needs to be 78 here now. Also, both evolution files should be renamed to 078*, the links in migrations.md updated, as well as the numbers incremented as you said (this was done for the forward evolution already above). The CI does not check the reversion, though. It fails because schema.sql Line 24 was not incremented to 78.

frontend/javascripts/dashboard/dataset/dataset_cache_provider.js
```diff
@@ -155,9 +168,41 @@ export default function DatasetCacheProvider({ children }: { children: Node }) {
     }
   }
 
+  async function updateCachedDataset(dataset: APIDataset) {
+    if (isLoading) return;
+    setIsLoading(true);
```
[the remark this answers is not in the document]
MichaelBuessemeyer: Thanks for spotting this. I implemented a queueing like behaviour for all updates on the same dataset. This queuing ensures that each previous request is finished before the current request starts. Therefore it is ensured that the backend always gets the updates in the correct order. Thus no update call that (by network magic) was faster at the server gets lost because a slower update request that is much slower overwrites its changes afterwards.
Here is a screenshot of a few dataset changes made on a single dataset in a row by changing the tags of a single dataset. The connection is throttled very strongly to test this corner case. The graph on the right illustrates that one call only starts after the previous updates finished.
image

frontend/javascripts/dashboard/dataset/dataset_cache_provider.js
```diff
@@ -155,9 +171,48 @@ export default function DatasetCacheProvider({ children }: { children: Node }) {
...
+        return dataset;
+      } else {
+        return currentDataset;
+      }
+    });
+    setDatasets(updatedDatasets);
+    datasetCache.set(updatedDatasets);
+    try {
+      const previousDatasetUpdatePromise = pendingDatasetUpdates[dataset.name];
+      const newDatasetUpdatePromise = Promise.resolve(previousDatasetUpdatePromise).then(async () =>
```
philippotto, line 191: While the behavior seems to be correct, I stumbled upon this, because it confused me that no null-check is done and I personally only use `Promise.resolve` to create a fulfilled Promise with a dedicated value. I would either add a comment which explains why the code works or rewrite it like this:

```suggestion
      const previousDatasetUpdatePromise = pendingDatasetUpdates[dataset.name] || Promise.resolve();
      const newDatasetUpdatePromise = previousDatasetUpdatePromise.then(() =>
```

Also note that I removed the `async` keyword as it wasn't necessary.

CHANGELOG.unreleased.md
```diff
@@ -11,7 +11,14 @@ For upgrade instructions, please check the [migration guide](MIGRATIONS.released
 [Commits](https://github.com/scalableminds/webknossos/compare/21.11.0...HEAD)
 
 ### Added
--
+- Enhanced the volume fill tool to so that it operates beyond the dimensions of the current viewport. Additionally, the fill tool can also be changed to perform in 3D instead of 2D. [#5733](https://github.com/scalableminds/webknossos/pull/5733)
```
philippotto, line 14: this is probably a merge error?

frontend/javascripts/dashboard/dataset/dataset_cache_provider.js
MichaelBuessemeyer: I am fine with your version :)

I just learned about this behaviour of `Promise.resolve` while reading on how to create a "promise-queue" :)

CHANGELOG.unreleased.md
MichaelBuessemeyer: Yeah i just found them and tried to fix it :/

philippotto (approved):
Great, looks merge-ready to me :)

MichaelBuessemeyer:
Is it correct, that I had to adjust the version of the `tools/postgres/schema.sql` manually? 

If yes, hopefully the CI now works 😅

frontend/javascripts/dashboard/dataset/dataset_cache_provider.js
```diff
@@ -44,12 +56,16 @@ export const DatasetCacheContext = createContext<Context>({
...
-  updateDataset: async () => {},
+  reloadDataset: async () => {},
+  updateCachedDataset: async () => {},
 });
 
 export default function DatasetCacheProvider({ children }: { children: Node }) {
   const [datasets, setDatasets] = useState(datasetCache.get());
   const [isLoading, setIsLoading] = useState(false);
+  const [pendingDatasetUpdates, setPendingDatasetUpdates] = useState<{
+    [string]: Array<APIDataset>,
```
philippotto, line 67: This should probably be

```suggestion
    [string]: Promise<APIDataset>,
```

to fix the flow error.

PR #5843, opened 2021-11-12
move /srv/webknossos to /webknossos in the dockerfiles

normanrz:
Moves the webKnossos files in the docker container from `/srv/webknossos` to `/webknossos`. This will make deployment on our infrastructure much easier, because we can mount `/srv` wholly and all symlinks etc will work internally.

Unfortunately, this brings a migration burden, where all mount-configs need to be changed. Would love to hear your feedback on that.

This PR also cleans up some unused and probably out-of-date scripts.

------
- [x] Ready for review


philippotto:
Sounds reasonable to me :+1:

fm3 (commented):
I can’t really estimate how much effort the migration is, but seems reasonable to me too.

Dockerfile
```diff
@@ -17,6 +17,9 @@ RUN addgroup --system --gid 999 webknossos \
   && chmod go+x bin/webknossos \
   && chmod go+w .
 
+RUN echo '#!/bin/bash\numask 002\n/bin/bash "$@"\n' > /docker-entrypoint.sh \
+  && chmod +x /docker-entrypoint.sh
```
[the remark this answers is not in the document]
jstriebel: PS: You could also bake the `bin/webknossos` part into the entrypoint to keep it more backwards-compatible. Also you don't need another bash call, e.g.
```suggestion
RUN echo '#!/bin/bash\numask 002\nbin/webknossos "$@"\n' > /docker-entrypoint.sh \
  && chmod +x /docker-entrypoint.sh
```

webknossos-tracingstore/Dockerfile
```diff
@@ -1,20 +1,23 @@
...
-VOLUME /srv/webknossos-tracingstore/tracingData /tmp
+VOLUME /webknossos-tracingstore/tracingData /tmp
 
 COPY target/universal/stage .
 
 RUN chown -R webknossos . \
   && chmod go+x bin/webknossos-tracingstore \
   && chmod go+w .
 
+RUN echo '#!/bin/bash\numask 002\n/bin/bash "$@"\n' > /docker-entrypoint.sh \
```
jstriebel, line 18: Same as above.

webknossos-datastore/Dockerfile
```diff
@@ -15,6 +15,9 @@ RUN chown -R webknossos . \
   && chmod go+x bin/webknossos-datastore \
   && chmod go+w .
 
+RUN echo '#!/bin/bash\numask 002\n/bin/bash "$@"\n' > /docker-entrypoint.sh \
+  && chmod +x /docker-entrypoint.sh
+
```
jstriebel, line 20: Same as above.

jstriebel (commented):
Nice! I mostly added one suggestion to keep the `bin/webknossos` call part of the entrypoint, which allows to keep the current `command`s as they are.

> Unfortunately, this brings a migration burden, where all mount-configs need to be changed. Would love to hear your feedback on that.

I think that's fine :+1:

jstriebel (approved):
Nice, LGTM :+1:

PR #5848, opened 2021-11-15
Admins can share annotations with all teams of organization

fm3:
The assumption that organization admins are implicitly members of all the organization’s teams was not respected in the updateSharedTeams method for annotations. This PR fixes that.

Instead of intersecting the passed teams with the user’s own memberships, we now check if all passed teams can be accessed by the user. This uses the correct semantic of either being a member of the team or an admin of the team’s organization.

Another change is that this now fails with an error instead of silently omitting the teams that could not be set.

- https://fixadminannotationsharing.webknossos.xyz

### Steps to test:
- Create second team in Administration→Teams
- Create an annotation
- Share annotation with that second team as admin
- Reload the page, that team should still be listed in the share modal
- Register a second user, activate them, make them member of the second team
- The second user should see the annotation in the Shared Annotations tab.

------
- [x] Ready for review


jstriebel (commented):
LGTM :+1:

jstriebel (approved):

PR #5851, opened 2021-11-15
Add wk libs client checkbox to PR template

fm3:
We need to take care that the webKnossos API and the python client using it do not diverge.


jstriebel (approved):
:+1:

PR #5853, opened 2021-11-15
Fix initial min/max values of a histogram of a color layer

### LINKED ISSUE #5694: Initial min/max values need to be changed once to get active
## Context
When opening a dataset link, the pre-set min/max values of a color layer may end up not being used for rendering until the min/max slider was changed once.

## Expected Behavior
The displayed min/max values in the sidebar should be used for rendering without any user interaction.

## Current Behavior
The user needs to change min/max for the initial values to take effect.

## Steps to Reproduce the bug

1. open https://webknossos.org/datasets/scalable_minds/l4dense_motta_et_al_types/view#2867,4160,1770,0,0.563 in incognito tab
2. change min/max slider for color layer a tiny bit --> contrast changes dramatically (which means that the values weren't properly used before).

## Your Environment for bug

- Browser name and version: Chrome 90
- Operating System and version: Ubuntu 20.04
- Version of WebKnossos (Release or Commit): f471130fc9d580c54b74c01f472d9a2ae353a692

MichaelBuessemeyer:
It is possible to configure only the `min` and `max` for a color layer in the view configurations of a dataset. This causes that the backend to send the default values for the `intensityRange` for the color layer. 
IIn case the `min` is higher than the sent intensity minimum or the max is lower than the sent intensity maximum the frontend got into an inconsistent state where the `intensityRange` is out of bounds considering the `min` and `max`.

A short solution to this is to force the intensityRange to be within the range given by `min` and `max`.

- https://___.webknossos.xyz

### Steps to test:
 - Go to the settings of a dataset and then to the "View Configurations"-tab.
 - There choose a (e.g. uint8) color layer and set `min` to 100 and `max` to 150.
 - Then make the dataset publically available under the "Sharing & Permissions"-tab, copy the sharing link and save the settings.
 - Then open the dataset with the public/sharing link in a private browser tab.
 - The rendered color layer should render data like the histogram slider is at `[100,150]`.   

### Issues:
- fixes #5694

------
- [x] Ready for review


frontend/javascripts/oxalis/model/sagas/load_histogram_data_saga.js
```diff
@@ -46,21 +46,28 @@ export default function* loadHistogramData(): Saga<void> {
...
-    if (layerConfigurations[layerName]) {
+    const currentLayerConfig = layerConfigurations[layerName];
+    if (currentLayerConfig) {
       newIntensityRange = [
-        Math.max(layerConfigurations[layerName].intensityRange[0], minimumInHistogramData),
-        Math.min(layerConfigurations[layerName].intensityRange[1], maximumInHistogramData),
+        Math.max(
+          currentLayerConfig.intensityRange[0],
+          currentLayerConfig.min != null ? currentLayerConfig.min : minimumInHistogramData,
+        ),
```
philippotto, line 55: Can you add a comment which explains the intuition here? I think, I kind of got now, but I had think this through for a few minutes and I'm still not 100% sure. My take would be:
`intensityRange` is the range for which the histogram is plotted, right? By default, it's defined by the data type (e.g., uint16), but it can be configured per dataset layer.
That range serves as a boundary (i.e., the range cannot be increased by something else), but can still be narrowed down. If nothing else is provided, that down-narrowing is done by the sampled histogram. If the user specified a default min/max value, that is used instead.

Also interesting: intensityRange is more like a "theoretical" range which is not exceeded by the data (e.g., the microscope  did only emit values between x and y), whereas `min`/`max` is more like a recommended cut-off to get a good rendering?

Did I get this right? :laughing:

philippotto (approved):
Great find! Only see my one comment before merging :)

frontend/javascripts/oxalis/model/sagas/load_histogram_data_saga.js
MichaelBuessemeyer: > `intensityRange` is the range for which the histogram is plotted, right?

It is the exact opposite. The min/max is the range of the histogram (slider) while the `intensityRange` is the range to which the color values are clamped.

Is my new comment explaining the code good enough? I tried keeping it short.

frontend/javascripts/oxalis/model/sagas/load_histogram_data_saga.js
```diff
@@ -46,21 +46,32 @@ export default function* loadHistogramData(): Saga<void> {
     const minimumInHistogramData = Math.min(...allMinValues);
     const maximumInHistogramData = Math.max(...allMaxValues);
     let newIntensityRange = [];
-    if (layerConfigurations[layerName]) {
+    const currentLayerConfig = layerConfigurations[layerName];
+    if (currentLayerConfig) {
+      // The intensityRange is the range where the color value will be clamped to
+      // while min/max are the value range of the histogram slider.
+      // The following keeps the intensityRange within the min and max bounds if they exist.
+      // If the do not exist the tracing / dataset has never been opened and thus the histogram data is used as a default range delimiter.
```
philippotto, line 54:
```suggestion
      // If they do not exist the tracing / dataset has never been opened and thus the histogram data is used as a default range delimiter.
```

frontend/javascripts/oxalis/model/sagas/load_histogram_data_saga.js
philippotto: Ah, makes sense! Yes, should be good to go now :)

philippotto (approved):
Feel free to merge :ship:

PR #5860, opened 2021-11-23
Send Slack notification if bucket loading requests are too slow

### LINKED ISSUE #5856: Send Slack notifications on very slow-loading buckets
This would help noticing when there’s something wrong on the datastore server, e.g. i/o overload.

fm3:
### Steps to test that I used:
- Enter valid slack token (or set verboseLoggingEnabled=true in application.conf and monitor backend logging)
- Add sleep to bucket loading
- slack notification should be sent/logged

### Issues:
- fixes #5856 

------
- [x] Ready for review


philippotto (approved):
looks good to me :+1:

PR #5863, opened 2021-11-23
In dataset upload, enable symlinking layers of existing datasets

### LINKED ISSUE #5850: Allow uploading segmentation layer for public color layer
The webknossos dataset upload API should support uploading a segmentation layer that should then be linked together with a specified existing public color layer of another dataset into a new combined dataset.
This functionality should be usable from the webknossos api client

fm3:
Extend the back-end API for dataset upload such that layers of existing public datasets can be symlinked into the new uploaded dataset.

Both routes /datasets/reserveUpload and /datasets/finishUpload now expect the additional field `layersToLink`, which is a list of objects each consisting of `organizationName: String, dataSetName: String, layerName: String` and optional `newLayerName: String`. In the upload protocol, the layersToLink must match in both requests.

The webKnossos front-end always sends an empty list there for the moment.

### Steps to test:
- in dataset_upload_view.js, change both occurences of layersToLink to something like 
```
            [{
              organizationName: "sample_organization",
              dataSetName: "dsA+",
              layerName: "color",
              newLayerName: "colorLinked"
            }]
```
- assuming that you have this dataset.
- Uploading a segmentation-layer-only dataset should then have the color layer from dsA+ linked in.
- In case of errors (dataset not public, duplicate layer name, invalid datasource-properties.json), readable error messages should be available.

### Todo:
 - [x] pass information about layers to symlink
 - [x] create the symlinks
 - [x] permission check (only if link target dataset is public)
 - [x] update datasource-properties.json
 - [x] error handling
 - [x] change back default to [] in js
 - [x] add rename layer functionality

### Issues:
- fixes #5850

------
- [x] Adapted [wk-libs python client](https://github.com/scalableminds/webknossos-libs/tree/master/webknossos/webknossos/client) if relevant API parts change → https://github.com/scalableminds/webknossos-libs/pull/508
- [x] Ready for review


webknossos-datastore/app/com/scalableminds/webknossos/datastore/services/DSRemoteWebKnossosClient.scala
```diff
@@ -80,14 +80,14 @@ class DSRemoteWebKnossosClient @Inject()(
...
     for {
-      userToken <- option2Fox(userTokenOpt) ?~> "initialTeams.noUserToken"
-    } yield
-      rpc(s"$webKnossosUri/api/datastores/$dataStoreName/verifyUpload")
+      userToken <- option2Fox(userTokenOpt) ?~> "validateDataSourceUpload.noUserToken"
+      _ <- rpc(s"$webKnossosUri/api/datastores/$dataStoreName/verifyUpload")
         .addQueryString("key" -> dataStoreKey)
         .addQueryString("token" -> userToken)
         .post(info)
+    } yield ()
```
fm3, line 90: This is a bug introduced in #5831 – the result of the rpc was not unpacked (and thus a Fox[Fox[WSResponse]] returned). Now I unpack it, so if the outer *or* the inner fox fails, the return value is a Failure as well

webknossos-datastore/app/com/scalableminds/webknossos/datastore/services/UploadService.scala
```diff
@@ -142,29 +160,54 @@ class UploadService @Inject()(dataSourceRepository: DataSourceRepository,
...
+  private def cleanUpOnFailure[T](result: Box[T],
+                                  dataSourceId: DataSourceId,
+                                  dataSetNeedsConversion: Boolean,
+                                  label: String): Fox[Unit] =
+    result match {
+      case Full(_) =>
+        Fox.successful(())
+      case Empty =>
+        Fox.failure(s"Unknown error $label")
+        deleteOnDisk(dataSourceId.team, dataSourceId.name, dataSetNeedsConversion, Some("the upload failed"))
```
jstriebel, line 202: Doesn't the Fox need to be at the end of this block to be returned?

webknossos-datastore/app/com/scalableminds/webknossos/datastore/services/UploadService.scala
```diff
@@ -22,14 +23,31 @@ case class ReserveUploadInformation(uploadId: String,
                                     name: String,
                                     organization: String,
                                     totalFileCount: Long,
+                                    layersToLink: List[LinkedLayerIdentifier],
```
jstriebel, line 26: Could we make this field optional? This might be an easy change to make the API compatible with old clients.

jstriebel (commented):
Nice, mostly LGTM :+1: Just see my two comments:

webknossos-datastore/app/com/scalableminds/webknossos/datastore/services/UploadService.scala
fm3: Fair enough, since the client incompatibilities are not checked properly right now, see https://github.com/scalableminds/webknossos-libs/issues/515
Long-term I’d rather not make this optional, since the empty list is the same semantically, and with Option I have to use .getOrElse(List.empty) everywhere.

jstriebel (approved):
Thanks, LGTM :+1:

PR #5865, opened 2021-11-24
Make w shortcut non-looping

daniel-wer:
See the linked issues/reports.
While I was at it, I removed an old keyboard shortcut which allowed to cycle between tools using `1` but only if a volume tool was active. Very confusing and inferior to the `w` shortcut imo.

The toggling is a little bit weird, because the "Eraser" appears twice (once for brushing, once for tracing) and the one for tracing is hidden by default unless the "Trace" tool is active. However, that has been the case before as well.

- https://___.webknossos.xyz

### Steps to test:
- Use w in an annotation to toggle through the tools. This should work reliably without jumping

### Issues:
- fixes https://forum.image.sc/t/bug-w-shortcut-debouncing-in-volume-annotation-mode-not-working-correctly/59787
- fixes https://discuss.webknossos.org/t/toggling-modes-with-w-does-not-work-reliably/1676/2

------
- [x] Ready for review


philippotto (approved):
Great! Works well :+1:

PR #5834, opened 2021-11-05
Rework for webknossos-worker as client-only

fm3:
The communication with webknossos-worker for long-running jobs no longer uses flower/celery, but instead webKnossos itself assigns jobs to polling workers.

### TODO
 - [x] save workers in postgres
 - [x] structure job state in postgres
 - [x] give out work to worker
 - [x] ensure no duplicates
 - [x] let worker post job status
 - [x] select datastore for job when submitting
 - [x] jobs status report
 - [x] worker status report
 - [x] insert/delete local worker in yarn enable-jobs
 - [x] detect worker death + resurrection
 - [x] in worker, retry posting job status
 - [x] slack notifications
 - [x] remove unused code
 - [x] write postgres evolutions
   - [x] forward
   - [x] reversion
 - [x] remove redundant jobsEnabled field in datastore sql table (should stay in dataSet publicWrites, though)
 - [x] re-build smoke test in worker
 - [x] make CI happy
 - [x] write follow-up issues
   - [x] construct result link for slack #5857
   - [x] propagate / post errors from logfile #5858

### Steps to test:
- enable jobs by calling `yarn enable-jobs`
- start a webknossos-worker (version from https://github.com/scalableminds/webknossos-worker/pull/70/ )
- export a bounding box from a dataset (bbox tab in view mode in right side bar)
- job should be executed, result should be downloadable from the wk jobs overview page
- kill worker during job, restart, job should be restarted
- submit multiple jobs, see that no combination breaks the setup
- api/jobs/status should show the worker and the jobs with status

### Issues:
- fixes #5687 

------
- [x] Updated [(unreleased) changelog](../blob/master/CHANGELOG.unreleased.md#unreleased)
- [x] Updated [(unreleased) migration guide](../blob/master/MIGRATIONS.unreleased.md#unreleased) if applicable
- [x] Ready for review

PR #5868, opened 2021-11-25
Do not use datastore actor system in WorkerLivenessService

fm3:
Follos-up fix for #5834

I tested this locally with "sbt run" and the datastore module disabled


jstriebel (approved):
:+1:

PR #5876, opened 2021-11-29
Fix flaky bucket eviction tests by splitting/isolating them

philippotto:
I basically split the two relevant tests into separate modules so that ava's process isolation kicks in. I also tried to debug what exactly interferes with each other, but at some point I gave up :shrug: 

- https://___.webknossos.xyz

### Steps to test:
- CI

### Issues:
- fixes flaky CI runs

------
- [X] Ready for review


daniel-wer (approved):
Thank you :heart:

PR #5877, opened 2021-11-29
Fix: Add path to schema diffing

MichaelBuessemeyer:
The schema diffing uses `execSync` and for some commands sets the environment variables. As this replaces the environment variables the `$PATH` is also reset which caused it no not find psql anymore on my machine. Therefore I added the envs of the process (including the PATH variable) to the envs of the process. This fixed the issue for me.

Note that the spread operator is not supported because the node version that is used by the backend is `0.12`.

Additionally, I propose a yarn command as a short command to kill all running nodejs and java processes if some of the ports are still occupied when the server did not get shutdown properly. But this is only optionally.

- https://___.webknossos.xyz

### Steps to test:
- Starting the server locally and opening a first webpage causes the backend to perform the schema diffing. If this works out fine (which can be seen in the console via the absence of an error) the fix should work.

### Issues:
- The is no issue for that

------
- [x] Ready for review

### DIFF
--- a/package.json
+++ b/package.json
@@ -91,7 +91,6 @@
   },
   "scripts": {
     "start": "node tools/proxy/proxy.js",
-    "stop": "killall java && killall node",
     "build": "node --max-old-space-size=4096 node_modules/.bin/webpack --env.production",
     "build-dev": "node_modules/.bin/webpack",
     "build-watch": "node_modules/.bin/webpack -w",
--- a/package.json
+++ b/package.json
@@ -91,6 +91,7 @@
   },
   "scripts": {
     "start": "node tools/proxy/proxy.js",
+    "stop": "killall java && killall node",
     "build": "node --max-old-space-size=4096 node_modules/.bin/webpack --env.production",
     "build-dev": "node_modules/.bin/webpack",
     "build-watch": "node_modules/.bin/webpack -w",
--- a/tools/postgres/diff_schema.js
+++ b/tools/postgres/diff_schema.js
@@ -14,6 +14,9 @@ const POSTGRES_URL =
   typeof process.env.POSTGRES_URL !== "undefined"
     ? process.env.POSTGRES_URL
     : "jdbc:postgresql://localhost/webknossos";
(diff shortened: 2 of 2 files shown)


MichaelBuessemeyer:
@philippotto I do not know whether you are the correct person to review this pr. If you know someone better (e.g. fm3?), please reassign the review :)

Thanks for looking over this who every will be doing that.

package.json
```diff
@@ -91,6 +91,7 @@
   },
   "scripts": {
     "start": "node tools/proxy/proxy.js",
+    "stop": "killall java && killall node",
```
philippotto, line 94: This seems to be too much of a bazooka. If the developer has other software running via java or node, those will all get killed (for example, a java-based IDE ^^). You could try to do  get the PID which runs a specific port via `lsof -i :9000 -Fp`, but the output still needs to be parsed. I suggest to remove this bit from the PR for now, so that it can get merged.

philippotto (commented):
Thanks for debugging and fixing this :+1: Does this also work for @Dagobert42? I remember that he checked out this branch and there was some trouble? Not sure whether the actual trouble came from the code diff, though.

package.json
MichaelBuessemeyer: Alrighty, I get what you mean . Then I keep this bazooka to myself 💥

Dagobert42: Trouble on my end got fixed with 5253e65. The ... operator seems to have been the cause :)

MichaelBuessemeyer: The explaination for this is that the backend starts the node script with a library that can start a own nodejs process in a defined environment. But the nodejs support seems to be not supported / maintained since a quite some time and therefore the newest nodejs version seems to be `0.12.<something>` and this is far too old 🧓. Nodejs only supports the spread operator (...) since version `8.*`. Thus I took merged the objects to form the env vars the complicated way 🤷‍♂️ . This seems to work and fixes the diffing for MacOS.

philippotto (approved):

PR #5885, opened 2021-12-02
Split cells via Min Cut

### LINKED ISSUE #5934: Split cell via min-cut
## Detailed Description
Given one segment and two seeds, I want to split the segment into two so that the seeds are in one segment each. The splitting strategy could work by computing a min-cut. See this https://github.com/scalableminds/webknossos/pull/5885 for a draft which shows how this approach could work. Currently, the performance is very slow, though.

Possible ways to improve performance:
- compute on mag 2 or mag 4 (I would suggest to do this first to get an intuition about whether the algorithm produces nice looking results)
- improve the algorithm (e.g., do two breadth-first searchs in parallel from both sides or use [incremental breadth-first algorithms](https://citeseerx.ist.psu.edu/viewdoc/download?doi=10.1.1.880.6530&rep=rep1&type=pdf))
- use wasm?

## Context

philippotto:
This PR:
- implements a min-cut approach to cut a cell given two seeds
- fixes a rendering bug where fallback rendering did not use the best available mag sometimes

- https://mincut.webknossos.xyz

### Steps to test:
- open a volume tracing with fallback layer
- use adhoc meshing for a cell to find a good spot to cut (e.g., a spinehead)
- create a tree place two connected nodes in the cell which represent the seeds (e.g., one in the dendrite and one in the spinehead)
- rightclick the tree and select "perform min-cut"
- a min cut should be performed which erases some voxels between the two seeds
- you can verify that the split was successful by using the floodfill with a new segment id on one half of the segment

### To do:
- [X] implement working version
- [X] clean up UI
- [X] clean up code
- [x] pull master
- [x] generalize mag handling (currently, the existence of 2-2-1 is assumed)
- [x] try to refine min-cut on better mags
- [x] make operation undo-able
- ~[ ] fix flicker bug when zooming from mag 4 to mag 8.~ extracted to https://github.com/scalableminds/webknossos/pull/6029

### Issues:
- fixes #5934 

------
- [x] Ready for review


.circleci/not-on-master.sh
```diff
@@ -1,8 +1,8 @@
 #!/usr/bin/env bash
 set -Eeuo pipefail
 
-if [ "${CIRCLE_BRANCH}" == "master" ]; then
+# if [ "${CIRCLE_BRANCH}" == "master" ]; then
```
philippotto, line 4: todo

frontend/javascripts/oxalis/api/api_latest.js
```diff
@@ -1233,16 +1233,16 @@ class DataApi {
...
     let zoomStep;
     if (_zoomStep != null) {
       zoomStep = _zoomStep;
     } else {
-      const resolutionInfo = getResolutionInfo(layer.resolutions);
       zoomStep = resolutionInfo.getClosestExistingIndex(0);
     }
 
-    const { resolutions } = layer;
+    const resolutions = resolutionInfo.getDenseResolutions();
```
philippotto, line 1245: this is a bug fix.

frontend/javascripts/oxalis/default_state.js
```diff
@@ -95,7 +95,7 @@ const defaultState: OxalisState = {
       maximumLayerCountToRender: 32,
     },
     preferredQualityForMeshPrecomputation: 2,
-    preferredQualityForMeshAdHocComputation: 2,
+    preferredQualityForMeshAdHocComputation: 1,
```
philippotto, line 98: todo: revert

frontend/javascripts/oxalis/controller/scene_controller.js
```diff
@@ -136,6 +136,22 @@ class SceneController {
       return cube;
     };
 
+    window.addVoxelMesh = (position: Vector3, cubeLength: Vector3, optColor?: string) => {
```
philippotto, line 139: I used this for debugging purposes, but there seems to be a slight offset in the position. Not sure why, though.

frontend/javascripts/oxalis/model/bucket_data_handling/texture_bucket_manager.js
```diff
@@ -388,9 +388,16 @@ export default class TextureBucketManager {
...
+          if (lookUpIdx === -1) {
+            // The lookUpIdx is invalid. Skip the entire loop.
+            break;
+          } else if (
+            this.lookUpBuffer[posInBuffer] !== -2 &&
+            this.lookUpBuffer[posInBuffer + 1] < bucketZoomStep
+          ) {
+            // Another bucket was already placed here with a better zoomstep.
+            // Skip the entire loop.
+            // TODO: clarify flicker when zooming from mag 4 to mag 8.
```
philippotto, line 400: todo: I might need a rubber duck here :see_no_evil:

philippotto:
@daniel-wer I think, this PR is ready for review. Let me know whether you want a high-level overview of this PR before you dive into it :) I tried to add many comments, but I'm not sure how well the big picture is explained.

frontend/javascripts/oxalis/model/sagas/min_cut_saga.js
```diff
@@ -0,0 +1,694 @@
...
+  volumeTracingLayer,
+): Saga<void> {
+  console.log("resolutionIndex, targetMag", { resolutionIndex, targetMag });
+
+  const boundingBoxTarget = boundingBoxMag1.fromMag1ToMag(targetMag);
+
+  const globalSeedA = V3.fromMag1ToMag(nodes[0].position, targetMag);
+  const globalSeedB = V3.fromMag1ToMag(nodes[1].position, targetMag);
+
+  let minDistToSeed = MIN_DIST_TO_SEED / targetMag[0];
```
philippotto, line 321: this is not correct, since the three dimensions have to be taken into account each.

frontend/javascripts/oxalis/model/sagas/min_cut_saga.js
```diff
@@ -0,0 +1,694 @@
...
+
+  // For the 3D volume flat arrays are constructed
+  // into which can be accessed with the helper methods
+  // l(x, y, z) and l([x, y, z]).
+  const size = boundingBoxTarget.getSize();
+  const l = (x, y, z) => z * size[1] * size[0] + y * size[0] + x;
+  const ll = ([x, y, z]) => z * size[1] * size[0] + y * size[0] + x;
+
+  if (inputData[ll(seedA)] !== inputData[ll(seedB)]) {
+    console.warn(
```
philippotto, line 346: toast

frontend/javascripts/oxalis/model/sagas/min_cut_saga.js
```diff
@@ -0,0 +1,694 @@
...
+    position[2] >= size[2]
+  );
+}
+
+function buildGraph(inputData, segmentId, size, length, l, ll, timeoutThreshold) {
+  const edgeBuffer = new Uint16Array(length);
+  for (let x = 0; x < size[0]; x++) {
+    if (x % Math.floor(size[0] / 10) === 0 && performance.now() > timeoutThreshold) {
+      // After each 10% chunk, we yield to redux-saga to allow
+      // cancellation.
```
philippotto, line 448:
```suggestion
      // After each 10% chunk, we check whether there's a timeout
```

frontend/javascripts/oxalis/model/sagas/min_cut_saga.js
```diff
@@ -0,0 +1,694 @@
...
+      throw TimeoutError;
+    }
+    for (let y = 0; y < size[1]; y++) {
+      for (let z = 0; z < size[2]; z++) {
```
philippotto, line 452: should I reorder the x-y-z loops to z-y-x for better cache locality? probably no harm in changing it.

frontend/javascripts/oxalis/model/sagas/min_cut_saga.js
```diff
@@ -0,0 +1,694 @@
...
+  const voxelStack = [seedB];
+  const maxDistance = distanceField[ll(seedB)];
+  let removedEdgeCount = 0;
+
+  while (voxelStack.length > 0) {
+    const currentVoxel = voxelStack.pop();
+    const currentDistance = distanceField[ll(currentVoxel)];
+
+    if (V3.equals(currentVoxel, seedA)) {
+      // console.log("Finished removing shortest path. Deleted edges:", removedEdgeCount);
```
philippotto, line 571:
```suggestion
```

frontend/javascripts/oxalis/model/sagas/min_cut_saga.js
```diff
@@ -0,0 +1,694 @@
...
+    }
+
+    if (!(distanceField[ll(neighborPos)] < currentDistance && distanceField[ll(neighborPos)] > 0)) {
+      throw new Error("Direction points towards a higher distance?");
+    }
+
+    const currDist = distanceField[ll(neighborPos)];
+    const distToSeed = Math.min(currDist, maxDistance - currDist);
+
+    if (distToSeed > minDistToSeed) {
```
philippotto, line 596:
```suggestion
    // We don't want to remove voxels which are close to the seeds.
    // See explanation for MIN_DIST_TO_SEED for details.
    if (distToSeed > minDistToSeed) {
```

frontend/javascripts/oxalis/model/bucket_data_handling/texture_bucket_manager.js
philippotto: I reverted the changes and extracted them to a separate PR: https://github.com/scalableminds/webknossos/pull/6029. Since this proved to be a bit tricky, a separate PR is easier to revert if it turns out to cause trouble in production. Also, it doesn't block the current PR then.

frontend/javascripts/oxalis/model/sagas/min_cut_saga.js
```diff
@@ -0,0 +1,722 @@
...
+    const neighbor = NEIGHBOR_LOOKUP[neighborIdx];
+
+    const neighborPos = V3.add(currentVoxel, neighbor);
+
+    if (isPositionOutside(neighborPos, size)) {
+      throw new Error("Neighbor is outside of volume?");
+    }
+
+    if (!(distanceField[ll(neighborPos)] < currentDistance && distanceField[ll(neighborPos)] > 0)) {
+      throw new Error("Direction points towards a higher distance?");
```
daniel-wer, line 616:
```suggestion
      throw new Error("Direction points towards an equal/higher/zero distance?");
```

frontend/javascripts/oxalis/model/sagas/min_cut_saga.js
```diff
@@ -0,0 +1,722 @@
...
+function traverseResidualsField(
+  boundingBoxTarget: BoundingBox,
+  seedA: Vector3,
+  ll,
+  edgeBuffer: Uint16Array,
+) {
+  // Perform a breadth-first search from seedA and return
+  // which voxels were visited (visitedField). This represents
+  // the graph component for seedA.
+  const visitedField = new Uint16Array(boundingBoxTarget.getVolume());
```
daniel-wer, line 651: Since the field only holds 0s and 1s, a `Uint8Array` is sufficient.
```suggestion
  const visitedField = new Uint8Array(boundingBoxTarget.getVolume());
```

frontend/javascripts/oxalis/model/sagas/min_cut_saga.js
```diff
@@ -0,0 +1,722 @@
...
+  const queue: Array<{ voxel: Vector3 }> = [{ voxel: seedA }];
```
daniel-wer, line 652: Could also simply be `Array<Vector3>`

frontend/javascripts/oxalis/model/sagas/min_cut_saga.js
```diff
@@ -0,0 +1,722 @@
...
+                V3.add(boundingBoxTarget.min, neighborPos),
+                targetMag,
+              );
+              for (let dx = 0; dx < targetMag[0]; dx++) {
+                for (let dy = 0; dy < targetMag[1]; dy++) {
+                  for (let dz = 0; dz < targetMag[2]; dz++) {
+                    api.data.labelVoxels([V3.add(position, [dx, dy, dz])], 0);
+                  }
+                }
+              }
```
daniel-wer, line 704: Wouldn't it be equivalent to label the single voxel in the target mag and let the upsampling take care of the rest. Or is it needed that the voxels in mag 1 are immediately labeled?

frontend/javascripts/oxalis/model/sagas/min_cut_saga.js
```diff
@@ -0,0 +1,722 @@
...
+      }
+    }
+  }
+}
+
+export default function* listenToMinCut(): Saga<void> {
+  yield* takeEveryUnlessBusy("PERFORM_MIN_CUT", performMinCut, "Min-cut is being computed.");
+}
+
+// Todo: remove before merging
```
daniel-wer, line 721: TODO

frontend/javascripts/oxalis/model/sagas/min_cut_saga.js
```diff
@@ -0,0 +1,722 @@
...
+// From the perspective of A, B can be accessed by using the outgoing edge
+// along vector [0, 1, 0] (idx=4). From the perspective of B, this is an
+// ingoing edge (referenced by idx=1 == [0, -1, 0]).
+const invertNeighborIdx = (neighborIdx: number) =>
+  (neighborIdx + NEIGHBOR_LOOKUP.length / 2) % NEIGHBOR_LOOKUP.length;
+
+function _getNeighborsFromBitMask(bitMask) {
+  // Note: Use the memoized version of this: getNeighborsFromBitMask.
+  // Ingoing and outgoing edges are stored as a bitmask. The first half
+  // of the bitmask holds the ingoing edges as bits. The second half the
```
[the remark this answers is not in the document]
philippotto: Probably depends on how one interpets "first half" :D I meant first half when viewing the binary representation (going from high-value to low-value). However, I'll simply rephrase it to be more precise.

frontend/javascripts/oxalis/view/context_menu.js
```diff
@@ -295,18 +309,20 @@ function NodeContextMenuOptions({
...
-        disabled={activeNodeId == null || !areInSameTree || isTheSameNode}
-        onClick={() =>
-          activeNodeId != null
-            ? measureAndShowLengthBetweenNodes(activeNodeId, clickedNodeId)
-            : null
-        }
-      >
-        Path Length to this Node
-      </Menu.Item>
+      {isTheSameNode ? null : (
```
philippotto, line 312: It bugged me that rightclicking a selected node always showed "Path length to this node: 0 vx".

frontend/javascripts/oxalis/controller/scene_controller.js
philippotto: Ah, now the penny has dropped :smile: However, the `0.9` is not fully responsible because I saw the offset already before I added the 0.9 factor. As you correctly point out, the flooring is messing things up. Since I used a cube length of `[1, 1, 1]`, the division by two (which is necessary since threejs expects the position of the cube to be in the center) produced fractional numbers which were important for rendering at the correct position, but were removed again by the flooring. I don't know why I even added the flooring in the first place. Maybe I thought, that there would be problems with fractionals :shrug:
I removed the flooring and also explained the 0.9 factor (which I added so that one can see the border between voxels in form of a gap) which wasn't used when instantiating the box geometry, by the way. I guess, it was late when I added this code :sweat_smile:

frontend/javascripts/oxalis/model/sagas/saga_helpers.js
```diff
@@ -0,0 +1,41 @@
...
+    if (busyBlockingInfo.isBusy) {
+      console.warn(
+        `Ignoring ${action.type} request (reason: ${busyBlockingInfo.reason || "null"})`,
+      );
+      return;
+    }
+
+    yield* put(setBusyBlockingInfoAction(true, reason));
+    yield* call(saga, action);
+    yield* put(setBusyBlockingInfoAction(false));
```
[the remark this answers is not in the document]
philippotto: The busy-info stuff is meant as a sort of mutex. I just checked all occurrences (there are not many) and all use the pattern correctly. As a safe guard, I added an assertion to the corresponding reducer so that an error is thrown in case the mutex is not handled properly (now or in the future). See 24acd467b.

frontend/javascripts/oxalis/model/sagas/min_cut_saga.js
```diff
@@ -0,0 +1,722 @@
...
+        color: Utils.getRandomColor(),
+        isVisible: true,
+      }),
+    );
+
+    boundingBoxObj = newBBox;
+  } else if (boundingBoxes.length === 1) {
+    boundingBoxObj = boundingBoxes[0].boundingBox;
+  } else {
+    console.log(
```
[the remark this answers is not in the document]
philippotto: > My gut tells me that this bounding box business is a little finicky and opaque. I'm not sure whether users will understand the effect of having bounding boxes visible, since it doesn't seem to be communicated. I can also imagine that it is somewhat annoying having to disable all (but one) bounding boxes every time a min-cut should be performed OR users forgetting they have an unrelated bounding box visible.

You are absolutely right. This was bugging me, too. I now changed the code so that one can choose which bounding box should be used from the context menu. Only the bounding boxes which contain the seed nodes are shown in the list. Additionally, one can select "Create new bounding box". I'm quite happy with the UX now :)
The sad news is that this change required to use antd submenus within the context menu which turned out to be very painful. I had to dig quite deep and tried several approaches to get this working. The final solution can be seen in the following commits. The link points to a comment where I explained the rationale behind this. I suggest to review these commits separately. Also note that I had to move quite a lot of code within the context menu module. Most of it stayed as is, but I added a new `<Dropdown />` component and separated it from the actual container. The upside is that I could remove quite a lot of custom CSS classes which we passed to antd menus :tada: 

https://github.com/scalableminds/webknossos/pull/5885/files/600dc941fe5e5a251cba60112b23fdb3fc846f78..58562dd90b1c73bc35b35cafe54215ee04fdf56d#diff-989748590830a0e73eef9fd9e98df428dab5559c0a88e4a656e9bd2814f3cb65R669

> Was the customizable bounding box a requirement for this feature? Otherwise, we could start by always using the automatically sized bounding box and think about adding a padding size setting or a bounding box selector once it is requested thinking

This would have been alright for the first iteration probably, but when developing the feature I quickly noticed that a poor bounding box can lead to poor min cut results. Ideally, it should cover exactly the merger (not too much and not too little).  That's why I wanted to make it easy to select the bbox. If I would have known that this was going to be so tricky, I might have decided differently :sweat_smile: However, I'm confident that this context menu re-architecturing will be helpful in the future, too.

frontend/javascripts/oxalis/model/sagas/min_cut_saga.js
```diff
@@ -0,0 +1,722 @@
...
+// in the selected mag, the next mag is tried.
+const MIN_CUT_TIMEOUT = 10 * 1000;
+// During the refinement phase, the timeout is very forgiving.
+// Even if the refinement is slow, we typically don't want to
+// abort it, since the initial min-cut has already been performed.
+const MIN_CUT_TIMEOUT_REFINEMENT = 10 * 60 * 1000;
+
+// To choose the initial mag, a voxel threshold is defined
+// as a heuristic. This avoids that an unrealistic mag
+// is tried in the first place.
```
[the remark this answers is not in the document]
philippotto: > Or does this result in worse cuts?

Yes, I think, when taken to the extreme (think of mag 1024), the "fat erasing" will definitely cause harm as it can easily erase too much. Still, there are probably ways to improve the heuristic, but I think for the first version, this should be fine.

frontend/javascripts/libs/mjs.js
```diff
@@ -209,6 +209,10 @@ V3.divide3 = function divide3(a, k, r) {
   return r;
 };
 
+V3.fromMag1ToMag = (vec, targetMag) => V3.floor(V3.divide3(vec, targetMag));
+
+V3.fromMagToMag1 = (vec, sourceMag) => V3.floor(V3.scale3(vec, sourceMag));
```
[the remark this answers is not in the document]
philippotto: I found it appealing to put it here, since the V3 module contains all operations which we perform on Vector3's. I know, that we haven't really used the domain knowledge of mags/resolutions here, but I think, it makes sense to do so. Overall, the entire mag/resolution handling (starting with the naming) could use a good cup of refactoring, obviously.. :see_no_evil:

philippotto:
@daniel-wer I think, I covered all your feedback and the remaining to dos :) Feel free to test again and have a look at the code. Also, see my one comment about the rationale behind the context menu changes: https://github.com/scalableminds/webknossos/pull/5885#discussion_r802808843

daniel-wer (commented):
Nice improvements! Looking forward to testing the reworked bounding box selection, I'll report back, afterwards :)

daniel-wer:
My notes from testing:

- [x] The bounding box selection and "happy path" (meaning nodes with a somewhat thin connection between them) work really well :partying_face: 
- [ ] Since the min-cut can take a while (seconds to minutes), some kind of progress indicator is needed in my opinion. Could simply be antd `Message`s showing the group descriptions that you are already using in the console (via `progress_callback.js`).
- [ ] I had a case that took a little longer and the console showed that "Find & delete paths" took more than 20s. I would've expected it to switch to the next-higher mag after 10 seconds as indicated in the code. The dataset had higher resolutions.
  ```
  Trying min-cut computation at 2-2-1
    Setting minDistToSeed to  12.5
    Loading data... (for 439350 vx)
    Starting min-cut...
    Build Graph: 65.99365234375 ms
    Segmentation is partitioned
    Find & delete paths: 21472.55126953125 ms
    traverseResidualsField: 38.225830078125 ms
    labelDeletedEdges: 137.98681640625 ms
    Total min-cut: 21714.920166015625 ms
  Refining min-cut at 1-1-1
    ...
  ```
- [ ] I had a min-cut reaching the bounding box border, because the cell was not entirely contained, leading to a spherical "min-cut" of one of the nodes. I understand why this happens and that it's not easy to solve, but maybe we could notify the user that the bounding box was likely too small if the edge(s) is/are reached. What do you think?
- [ ] [Low-Prio] I tried to min-cut a thicker cell which triggered the switch to the next-higher mag (which is good :+1:). In that case, some of the timers are not reset correctly, it seems
  ```
  min_cut_saga.js:404 Timer 'Total min-cut' already exists
  min_cut_saga.js:422 Timer 'Find & delete paths' already exists
  ```

philippotto:
Thanks for your feedback, @daniel-wer! I think, I incorporated almost all of it. Would you mind having another look at my newest commits?

> I had a case that took a little longer and the console showed that "Find & delete paths" took more than 20s. I would've expected it to switch to the next-higher mag after 10 seconds as indicated in the code.

I could reproduce this :+1: One modulo operation used a quite high operand which I changed from 100000 to 10000. Now, it should work better.

> I had a min-cut reaching the bounding box border, because the cell was not entirely contained, leading to a spherical "min-cut" of one of the nodes. I understand why this happens and that it's not easy to solve, but maybe we could notify the user that the bounding box was likely too small if the edge(s) is/are reached. What do you think?

This is the only thing I skipped now. I'd like to get some feedback from real-world usages before I fune-tune this feature further. The feature itself is marked as "experimental" in the UI, anyway, and there might be other more impactful levers later which we need to turn. I hope this is alright with you? :)

daniel-wer (approved):
Code LGTM :partying_face: 

> This is the only thing I skipped now. I'd like to get some feedback from real-world usages before I fune-tune this feature further. The feature itself is marked as "experimental" in the UI, anyway, and there might be other more impactful levers later which we need to turn. I hope this is alright with you? :)

Absolutely, makes sense :)

PR #5886, opened 2021-12-02
Refactor and unify restrictions handling in skeletontracing reducer

daniel-wer:
This has been nagging me for quite a while. The restrictions checking in the skeleton tracing reducer was not done in one spot but in (every) single reducer, leading to the following issues:
- Restrictions were not checked in some reducers, for example, group creation, group modification, tree naming
- Duplicated code
- Increased indentation depth

Description of the changes:

- For `skeletontracing_reducer.js` I only moved around code, removed restrictions arguments and added the early out after the restrictions-independent actions.
  - I did a small change change for the `ADD_TREES_AND_GROUPS` action, decoupling the `ensureTreeNames` functionality from the `addTreesAndGroups` reducer so that it no longer needs the state and can be reused for the connectome functionality.
  - The following actions did not check `allowUpdate` before, but do so now: `SET_NODE_POSITION`, `CREATE_TREE`, `RESET_SKELETON_TRACING`, `SET_TREE_NAME`, `SET_TREE_GROUPS`, `SET_TREE_GROUP` so this functionality should be tested
- `skeletontracing_reducer_helpers.js` should be easily reviewable with `Hide Whitespace`
- I checked that no other file imports reducers from `skeletontracing_reducer_helpers.js`

- https://betterrestrictionshandling.webknossos.xyz

### Steps to test:
- Open a read-only tracing, for example, by making an annotation public, copying the sharing link and logging out
- Check that actions like shuffling the tree color, enabling merger mode etc. still work
- Check that actions like creating trees using `c`, removing nodes or creating groups don't work

------
- [x] Ready for review


philippotto (approved):
Sweet! Works well, too :candy:

PR #5890, opened 2021-12-06
Make access token check explicit in data/tracing stores

### LINKED ISSUE #5864: Avoid accessing GET parameters implicitly
The routes file is not very readable in that some routes (e.g. datastore/tracingstore/worker access check) use `request.getQueryString` rather than unpacking the query strings directly. This should be avoided.

### LINKED ISSUE #5875: Swagger: add token param to data download route
Data download needs an optional `token` param, otherwise download of private datasets is not possible

fm3:
The implicit accessing of GET parameters via getQueryString led to confusion when reading the code and to them missing from the swagger docs. This PR changes them to explicit GET parameters (usually tokens identifying the user, or keys identifying datastores/tracingstores).

- https://explicittokencheck.webknossos.xyz

### Steps to test:
- Tracing saving, upload/download, data viewing should still work both if logged in and (viewing public datasets) if logged out

### Issues:
- fixes #5875
- fixes #5864 

Follow-up for Filters queries (the one remaining instance of `getQueryString` after this): https://github.com/scalableminds/webknossos/issues/5891

------
- [x] Ready for review


fm3:
@jstriebel This adds the token field to swagger, as requested in #5875 – do you know if I can consider the “Adapted wk-libs python client if relevant API parts change” checkbox done? Will the change break the current client?

jstriebel:
> Will the change break the current client?

No, adding new fields should be fine :+1:

webknossos-datastore/app/com/scalableminds/webknossos/datastore/controllers/BinaryDataController.scala
```diff
@@ -100,6 +102,7 @@ class BinaryDataController @Inject()(
       new ApiResponse(code = 400, message = "Operation could not be performed. See JSON body for more information.")
     ))
   def requestRawCuboid(
+      @ApiParam(value = "Datastore token identifying the requesting user", required = false) token: Option[String],
```
jstriebel, line 105: I'm wondering if the `Option` part makes the parameter optional in swagger, too, or if the `required = False` is strictly necessary. Totally fine to leave as-is in any way.

jstriebel (approved):
Nice stuff! Code LGTM and during testing everything worked fine :+1: :rocket:

webknossos-datastore/app/com/scalableminds/webknossos/datastore/controllers/BinaryDataController.scala
fm3: Turns out the default of required is False. I’ll remove it then.

PR #5894, opened 2021-12-07
Connectome Tab

### LINKED ISSUE #5828: Connectome Viewing in webKnossos
## Detailed Description

See internal work package D.1

daniel-wer:
This PR implements the internal work package D.1 Connectome Viewing in webKnossos (see internal communication). Refer to the work package for the full feature set and a mockup.

This PR adds a new "Connectome Tab" and I tried to encapsulate the new functionality fairly well so that most of it is contained in a couple of files. Still, I had to adapt the code in a couple of spots to support some new use cases which I'll describe in the following:
- Allow to set the `enableRenderOnDemand` prop for layout tabs. Setting the prop to false will disable the lazy-loading of the tab and instead always render it. For the connectome tab, this is important to allow hotlinking to agglomerates/synapses without the tab being visible when loading the annotation. Otherwise, the code encapsulation would not work as well. I don't think this is a problem, though, since the tab doesn't do much if it's not actively being used.
- Allow to add multiple Skeletons (skeleton.js) to the scene. This is used to separate the connectome skeleton data from the "normal" skeleton data so the two don't interfere with each other and the connectome skeleton data is effectively read-only. Also allow to remove a Skeleton from the scene. This is done during cleanup or when the segmentation layer is switched.
- Introduce the `lastVisibleSegmentationLayerName` to allow to continue to use the connectome tab even if the segmentation was temporarily disabled. This might be useful in other spots as well.

### ToDo:
- [x] Separate connectome_view.js into separate logical units
- [x] Adapt docs/sharing.md#sharing-link-format to new link format
- [x] Check all check boxes that represent the same segment id (not only the clicked one). ~~No idea what would be a good way to do that yet~~
- [x] ~~Show the segment color next to the entry in the tree (if the agglomerate skeleton is loaded)~~ _Wait for feedback, maybe followup_
- [x] The connectome tab should work in view mode
- [x] Do we need a layout migration or is it sufficient to increase the layout version number?
- [x] Hovering over a Segment entry in the tab could highlight the segment in the data viewport. The same feature exists for the segment tab, so it might be easy to copy.
- [x] The context menu in the data viewport could have an entry for loading a segment into the connectome tab
- [x] If an agglomerate could not be loaded, because the node limit is exceeded, a nicer error message could be rendered. Currently, one has to expand the details of the error message to find out that this is not a bug.

- https://connectomeviewing.webknossos.xyz

### Steps to test:
- Open a dataset without a connectome. The connectome tab should state there is no connectome.
- Use the  `l4dense_motta_et_al_dev` dataset that has a linked connectome file on the deployed dev instance
- The connectome tab should work in view mode as well as in annotations
- Right-Click on a segment in the viewport and select "Import Agglomerate and Synapses". The segment should be shown in the connectome tab, its agglomerate skeleton loaded and all synapses shown as single node trees as well.
  - The same should happen if you enter the segment ID by hand
- Entering two segment IDs (comma-separated) should show both skeletons and only the synapses between the two agglomerates
- It should be possible to toggle the segmentation layer while a segment is shown in the connectome tab without the info in the connectome tab to disappear
- For datasets with multiple segmentation layers, if the segmentation layer is switched the tab should be reset
- The filters by synapse direction or synapse type should work and hide the filtered synapses/agglomerates from the tree list as well as the viewports
- Right-click on a 2nd-level segment in the tree list and select "Show All Synapses of This Segment" to see the synapses of that segment, instead
- The "Reset" buttons to reset the filters as well as the connectome tab should work
- It should be possible to link to a specific agglomerate and its synapses (tested in the screenshot tests), for example, [this link](https://connectomeviewing.webknossos.xyz/datasets/sample_organization/l4dense_motta_et_al_dev/view#{%22position%22:[2816,4543,1770],%22mode%22:%22orthogonal%22,%22zoomStep%22:0.734,%22stateByLayer%22:{%22segmentation%22:{%22connectomeInfo%22:{%22connectomeName%22:%22paper_l4_full_connectome%22,%22agglomerateIdsToImport%22:[1421,2717]}}}})

### Issues:
- fixes #5828 

------
- [x] Adapted [wk-connect](https://github.com/scalableminds/webknossos-connect) if datastore API changes
- [x] Ready for review


daniel-wer:
@philippotto As discussed, it would be great if you could have a first look at this PR. I'd suggest to have a look at all files except for the large `connectome_view.js`. I'll spend some time on Monday, trying to refactor it into more digestible portions :)

frontend/stylesheets/trace_view/_right_menu.less
```diff
@@ -69,6 +69,12 @@
   flex-direction: column;
 }
 
+#connectome {
```
philippotto, line 72: maybe change to connectome-view/list/tab or something more specific

frontend/javascripts/oxalis/model/reducers/connectome_reducer.js
```diff
@@ -0,0 +1,193 @@
...
+  addTreesAndGroups,
+  getMaximumNodeId,
+} from "oxalis/model/reducers/skeletontracing_reducer_helpers";
+import { updateKey3 } from "oxalis/model/helpers/deep_update";
+import Constants from "oxalis/constants";
+
+function getSkeletonTracingForConnectome(
+  state: OxalisState,
+  layerName: string,
+): Maybe<SkeletonTracing> {
```
philippotto, line 19: Since flow handles nullable types quite well, I'm not too convinced of the value of `Maybe`, anymore. This is why I haven't really adopted the class for new stuff. If you think, it's useful, feel free to keep it, but if you agree with my take, then I'd suggest to not add further usages of it (and in the long run, we could try to remove it along the way).

/edit: Ah, you probably did this to maximize the compatibility with the existing skeleton reducer (helpers). Then, it probably makes sense to keep it as is (at least for the scope of this PR).

philippotto (commented):
Nice stuff! I reviewed everything except for `connectome_view.js` and the code looks very clean and polished already :+1:

frontend/javascripts/oxalis/model/reducers/connectome_reducer.js
daniel-wer: That makes sense :+1: As you say, in this case it was done to maintain compatibility :)

frontend/javascripts/oxalis/throttled_store.js
```diff
@@ -17,14 +17,14 @@ Store.subscribe(() => {
...
     waitForUpdate.resolve();
-    waitForUpdate = new Deferred();
   }
 });
 
 async function go() {
   // eslint-disable-next-line no-constant-condition
   while (true) {
     await waitForUpdate.promise();
+    waitForUpdate = new Deferred();
```
daniel-wer, line 27: This fixes a bug where store updates could be swallowed if they were triggered by a store listener, themselves. In this case, this `go` loop would still be in line 30 executing the listeners (triggering a store change) which would cause the promise in line 19 to be resolved and then immediately reassigned. After executing the listeners, the code would then wait for a store change in line 27, although the change already happened.
I tested that my change fixes this scenario.

daniel-wer:
@philippotto Thank you for the first review!
I squashed a couple more bugs and refactored the large `connectome_view.js` into four smaller components which are all contained in the `oxalis/view/right-border-tabs/connectome_tab/` folder. I'd say you can go on with reviewing my last changes as well as the unreviewed files. Let me know if it's hard to follow the code in some spots or if you need some more background information, happy to jump into a call with you :)

frontend/javascripts/oxalis/view/right-border-tabs/connectome_tab/connectome_filters.js
```diff
@@ -0,0 +1,191 @@
...
+    );
+  };
+
+  render() {
+    const { availableSynapseTypes, disabled } = this.props;
+    const { filters } = this.state;
+    const isSynapseTypeFilterAvailable = availableSynapseTypes.length;
+
+    const isSynapseTypeFiltered = filters.synapseTypes.length !== availableSynapseTypes.length;
+    const isSynapseDirectionFiltered = filters.synapseDirections.length !== 2;
```
philippotto, line 176:
```suggestion
    const isSynapseDirectionFiltered = filters.synapseDirections.length !== defaultFilters.synapseDirections.length;
```

frontend/javascripts/oxalis/view/right-border-tabs/connectome_tab/connectome_view.js
```diff
@@ -0,0 +1,784 @@
...
+  // filters the synapses object, the two effectively need to be unioned and only the synapse ids
+  // that occur in one of the agglomerates' in/out arrays as well as in the synapses object need to be returned.
+  const { synapses, agglomerates } = connectomeData;
+  return unique(
+    Object.values(agglomerates).flatMap(
+      // $FlowIssue[incompatible-call] remove once https://github.com/facebook/flow/issues/2221 is fixed
+      ({ in: inSynapses = [], out: outSynapses = [] }: Agglomerate) => [
+        ...inSynapses,
+        ...outSynapses,
+      ],
```
philippotto, line 116: maybe a concat is better here? not sure if this is better perf-wise (and whether it's necessary at all), but somehow the two spreads caught my eye.

```suggestion
      ({ in: inSynapses = [], out: outSynapses = [] }: Agglomerate) => inSynapses.concat(outSynapses),
```

frontend/javascripts/oxalis/view/right-border-tabs/connectome_tab/connectome_view.js
```diff
@@ -0,0 +1,784 @@
...
+  const topLevelAgglomerateIds = Object.keys(agglomerates).map(agglomerateId => +agglomerateId);
+  const filteredSynapseIds = getSynapseIdsFromConnectomeData(connectomeData);
+  const partnerAgglomerateIds = filteredSynapseIds.map(
+    (synapseId): number => {
+      const synapse = synapses[synapseId];
+      // $FlowIssue[incompatible-return] Flow doesn't understand that if src == null -> dst != null
+      return synapse.src != null ? synapse.src : synapse.dst;
+    },
+  );
+  return unique([...topLevelAgglomerateIds, ...partnerAgglomerateIds]);
```
philippotto, line 137: same here

frontend/javascripts/oxalis/view/right-border-tabs/connectome_tab/connectome_view.js
```diff
@@ -0,0 +1,784 @@
...
+      expandedKeys: [],
+    });
+  }
+
+  initializeSkeleton() {
+    const { segmentationLayer } = this.props;
+    if (segmentationLayer == null) return;
+
+    Store.dispatch(initializeConnectomeTracingAction(segmentationLayer.name));
+    this.skeletonId = getSceneController().addSkeleton(
```
philippotto, line 275: Maybe assert that `this.skeletonId == null` to ensure that `removeSkeleton` was called before?

frontend/javascripts/oxalis/view/right-border-tabs/connectome_tab/connectome_view.js
```diff
@@ -0,0 +1,784 @@
...
+      synapseDestinations,
+      synapsePositions,
+      synapseTypesAndNames,
+    ] = await Promise.all([
+      getSynapseSources(...fetchProperties, allInSynapseIds),
+      getSynapseDestinations(...fetchProperties, allOutSynapseIds),
+      getSynapsePositions(...fetchProperties, allSynapseIds),
+      getSynapseTypes(...fetchProperties, allSynapseIds),
+    ]);
+    // TODO: Remove once the backend sends the typeToString mapping from the hdf5 file
```
philippotto, line 367: todo

philippotto (commented):
Beautiful stuff! I haven't finished connectome_view.js yet, though. Will review the rest later today or tomorrow if that's alright :)

frontend/javascripts/oxalis/view/right-border-tabs/connectome_tab/connectome_view.js
```diff
@@ -0,0 +1,784 @@
...
+    if (skeleton == null) return;
+
+    const { trees } = skeleton;
+
+    if (deletedSynapseIds.length > 0) {
+      const treeIdsToDelete = [];
+      deletedSynapseIds.forEach(synapseId =>
+        findTreeByName(trees, getTreeNameForSynapse(synapseId)).map(tree =>
+          treeIdsToDelete.push(tree.treeId),
+        ),
```
philippotto, line 462: Since is executed in bulk, one could theoretically compute a look up data structure from tree name to tree first. Right now, this performs a linear search for each synapse ID. I don't know how many synapse ids here are typically, so maybe it's completely unnecessary. Just crossed my mind when reading this..

frontend/javascripts/oxalis/view/right-border-tabs/connectome_tab/connectome_view.js
```diff
@@ -0,0 +1,784 @@
...
+
+    const checkedAgglomerateIds = getAgglomerateIdsFromKeys(checkedKeys);
+    const visibleAgglomerateIds = checkedAgglomerateIds.filter(agglomerateId =>
+      filteredAgglomerateIds.includes(agglomerateId),
+    );
+
+    const reset =
+      prevConnectomeData != null &&
+      connectomeData != null &&
+      prevConnectomeData.connectomeFile.mappingName !== connectomeData.connectomeFile.mappingName;
```
philippotto, line 517: I'm likely missing the big picture here, but for example, wouldn't this need the following `||` case? `prevConnectomeData.connectomeFile !== connectomeData.connectomeFile`? So, if the connectomeFile changed, the agglomerate trees would also need to be reset? Or is this handled somewhere else? Or do you assume that this would automatically mean that the `mappingName` also changed? But then, the condition could be more loose.

In general, I find the `componentDidUpdate` logic a bit hard to grasp, since the individual methods which can be called from there also do this kind of `reset` logic. Maybe it's easier to read if the entire "what needs to be updated?"-logic is handled within `componentDidUpdate` and parameters, such as `reset`, are then simply passed to the helper methods?

frontend/javascripts/oxalis/view/right-border-tabs/connectome_tab/connectome_view.js
```diff
@@ -0,0 +1,784 @@
...
+    const disabled = currentConnectomeFile == null;
+
+    return (
+      <>
+        <Input.Group compact className="compact-icons" style={{ marginBottom: 10 }}>
+          <Tooltip title="Show Synaptic Connections for Segment ID(s)">
+            <InputComponent
+              value={activeAgglomerateIdString}
+              onPressEnter={this.handleChangeActiveSegment}
+              placeholder="Segment ID 1[, Segment ID 2, ...]"
```
philippotto, line 714: I'm not sure whether the placeholder could confuse people so that they try to input "Segment ID 1". Maybe "Enter one or multiple segment ids here (separated by commas)" would be clearer?

frontend/javascripts/oxalis/view/right-border-tabs/connectome_tab/connectome_view.js
```diff
@@ -0,0 +1,784 @@
...
+    const { activeAgglomerateIds, currentConnectomeFile } = this.props;
+    const { filteredConnectomeData, checkedKeys, expandedKeys } = this.state;
+
+    if (currentConnectomeFile == null) {
+      return <Empty image={Empty.PRESENTED_IMAGE_SIMPLE} description="No connectome available." />;
+    } else if (activeAgglomerateIds.length === 0 || filteredConnectomeData == null) {
+      return (
+        <Empty
+          image={Empty.PRESENTED_IMAGE_SIMPLE}
+          description="No segment selected. Use the input field above to enter a segment ID."
```
philippotto, line 745:
```suggestion
          description="No segment selected. Use the input field above to enter segment IDs."
```

frontend/javascripts/oxalis/view/right-border-tabs/connectome_tab/connectome_view.js
```diff
@@ -0,0 +1,784 @@
...
+          connectomeData={filteredConnectomeData}
+        />
+      );
+    }
+  }
+
+  render() {
+    const { segmentationLayer } = this.props;
+
+    // Render the tab in the background to avoid rebuilding the tree when switching tabs.
```
philippotto, line 765: Does this comment refer to a particular line of code here? I thought, this is already covered by https://github.com/scalableminds/webknossos/pull/5894/files#diff-8062b2d313f68c910f76cce564fa2a4af33d48232c4ea9e79ff3c707da0e33caR148. If so, I'd remove the comment here.

philippotto (commented):
Whew, great (and a lot of) stuff :) I looked through all of the code and it looked very good to me. I only left a couple of comments.

I've got some personal feature requests, which are obviously only suggestions:
-  Hovering over a Segment entry in the tab could highlight the segment in the data viewport. The same feature exists for the segment tab, so it might be easy to copy.
- The context menu in the data viewport could have an entry for loading a segment into the connectome tab
- If an agglomerate could not be loaded, because the node limit is exceeded, a nicer error message could be rendered. Currently, one has to expand the details of the error message to find out that this is not a bug.

Regarding your questions:

>  Should this work in view mode? (Currently it partly works, but agglomerate skeletons are not loaded)

Ideally, this would work. Then, not registered users could fiddle with this, too. However, the sandbox mode might also be a solution here. Nevertheless, supporting view-only would be nice, but not necessary in my opinion.

On a related note: If I edit the volume layer in my annotation, the connectome file would ignore these changes, right? This is the same problem class which we already have for the agglomerate file, right?

> Do we need a layout migration or is it sufficient to increase the layout version number?

In the past, we simply increased the layout version. I think, it would be fine.

------------

You will probably do another iteration now, right?

frontend/javascripts/oxalis/model/reducers/connectome_reducer.js
```diff
@@ -0,0 +1,200 @@
...
+
+    case "UPDATE_CURRENT_CONNECTOME_FILE": {
+      const { layerName, connectomeFileName } = action;
+      const availableConnectomeFiles =
+        state.localSegmentationData[layerName].connectomeData.availableConnectomeFiles;
+      if (availableConnectomeFiles == null) {
+        // Connectome files have not been fetched yet, temporarily save the connectome file name
+        // so it can be activated once the files have been fetched
+        return updateKey3(state, "localSegmentationData", layerName, "connectomeData", {
+          prendingConnectomeFileName: connectomeFileName,
```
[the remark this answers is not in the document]
daniel-wer: I also wasn't sure about this. This pattern is not used anywhere else as far as I know.

I thought about using a saga as you suggested, but I was worried about race conditions between the saga and the ConnectomeSettings component. If both of them fetch the available connectome files at the same time, one would need to make sure that the activation of the connectome file from the URL wins over the activation of the connectome file from the ConnectomeSettings component (which would activate the first connectome file if none was activated before).
Having the `prendingConnectomeFileName` property in the store was my approach to "synchronize" the two and to make sure the correct one "wins".

Still, I think a saga would be the expected way to accomplish this. I'll try that :)

frontend/javascripts/oxalis/view/right-border-tabs/connectome_tab/connectome_view.js
daniel-wer: Seems to make a difference for large array sizes since concat is less generic and can be better optimized (https://stackoverflow.com/a/57191593). I'll switch :+1:

frontend/javascripts/oxalis/view/right-border-tabs/connectome_tab/connectome_view.js
daniel-wer: This cannot be done yet, since the backend jhdf5 library has a bug when reading array attributes. The library author was contacted, but didn't comment yet. I'll extend the TODO comment.

frontend/javascripts/oxalis/view/right-border-tabs/connectome_tab/connectome_view.js
daniel-wer: I had this in the back of my mind as a possible performance optimization, but didn't expect gains this large!

I loaded a moderately large segment (1421) with ~600 synapses and reset/filtered randomly. I measured the time to identify which trees to delete, given the IDs of the deleted synapses.

Before:

51.032958984375 ms
11.222900390625 ms
30.21484375 ms
36.15087890625 ms
0.9521484375 ms
38.079833984375 ms

After:

1.378662109375 ms
0.9638671875 ms
0.55908203125 ms
0.52685546875 ms
0.863037109375 ms
0.447998046875 ms

Thank you for bringing this up :partying_face:

frontend/javascripts/oxalis/view/right-border-tabs/connectome_tab/connectome_view.js
daniel-wer: Didn't think of that, but you're likely right. Since the placeholder would get quite long, I'd try it with "Enter Segment ID(s)" and expect users to figure out (or understand intuitively) that multiple values need to be comma-separated :thinking:

daniel-wer: Related, I thought about using a multi-select with a "badge-style" instead of the standard input field. This would be configurable to auto-create badges when entering "," or " ". Also, it would be easier to remove single segment IDs and look better imo. See this sandbox for an example: https://codesandbox.io/s/automatic-tokenization-antd-4-18-3-forked-b13ig
Enter a segment ID and press Enter or type "," or " ".

frontend/javascripts/oxalis/view/right-border-tabs/connectome_tab/connectome_view.js
daniel-wer: > prevConnectomeData.connectomeFile !== connectomeData.connectomeFile? So, if the connectomeFile changed, the agglomerate trees would also need to be reset?

Since the agglomerate trees are loaded from the agglomerate file, they only need to be reset if the mapping changes (which is indicated by the `mappingName`). The connectome file only holds the agglomerate IDs which are then used to look up the agglomerate skeletons in the agglomerate file.

> In general, I find the componentDidUpdate logic a bit hard to grasp, since the individual methods which can be called from there also do this kind of reset logic.

That's good feedback, I'll think about how to make this easier to grasp!

daniel-wer:
@philippotto Thank you for the thorough review and apologies again for the large amount of code!

> On a related note: If I edit the volume layer in my annotation, the connectome file would ignore these changes, right? This is the same problem class which we already have for the agglomerate file, right?

Yes, that is exactly right, the connectome file is static, same as the agglomerate file.

> You will probably do another iteration now, right?

Yes, I'll integrate the remainder of your feedback, take care of (or postpone some of) the remaining TODOs and then ping you again :)

frontend/javascripts/oxalis/view/right-border-tabs/connectome_tab/connectome_view.js
philippotto: Wow, that's great news :)

frontend/javascripts/oxalis/view/right-border-tabs/connectome_tab/connectome_view.js
philippotto: Interesting! I agree with the advantages you mention. I only want to bring up one disadvantage: One cannot easily copy all active segment ids, anymore (pasting obviously works). If one wants to make notes in an external text file or something like that, this could be a bit annoying. However, this scenario is a bit far fetched and could easily be fixed with a dedicated "copy segment ids" button. Just wanted to mention it.

frontend/javascripts/oxalis/model/reducers/connectome_reducer.js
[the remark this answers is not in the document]
daniel-wer: I thought about and had a look at this a bit more. What further complicates the issue is that the connectome file activation needs to be tracked on a per-layer basis. The saga would need to track pending connectome file activations in some kind of object or be executed per layer. The first option is not that far away from the current solution and the second option doesn't feel right :tm: 
I would opt to leave this as-is for now.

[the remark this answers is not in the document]
philippotto: Sure, fine by me :)

daniel-wer:
@philippotto I should've addressed all of your feedback and also took care of a couple of bugs and remarks from alpha-testing.

I think, this version is very close to being mergeable. The open issues I see are:
- [ ] Currently the connectome tab is always visible. Since it is kind of a niche functionality, should it maybe only be visible for specific organizations (that want to test this functionality and/or we know to be in the connectomics field)? And if so, does that have any implications for the viewport configurations (if the tab is sometimes there and sometimes not)?
- [ ] Open checkboxes in PR description. @fm3 could you have a look at the three checkboxes and delete them if they are not needed?
- [ ] Final testing round

More notes:
- The Connectome File specification is not (yet?) open source. Therefore, I didn't add much information to the docs yet. The change log notes where the connectome file needs to be placed (mostly for us or for beta users that have a connectome file). Once the connectome tab functionality stabilized a bit, we should add a section to the docs that explains how to use it and what a connectome file needs to look like.
- I'm still not sure whether the decision to keep the connectome trees separate from the "normal" trees was the right one. For future reference, these are the pros and cons I see:
  - Pro: Trees are easily read-only, easy to avoid saving trees, easier housekeeping of trees (since users cannot add trees with interfering names), works in view mode
  - Contra: Nodes cannot be selected (to center them for example or to access the context menu), tree color cannot be changed, some settings are only available if there is a "normal" skeleton layer (particle size, etc.), copy to account functionality cannot be used to create tracing from sandbox/view mode, undo/redo doesn't work

fm3:
@daniel-wer I removed the two checkboxes I was comfortable removing. I’m not sure about the wkconnect case. What does the frontend do if a wkconnect dataset is opened, where the new routes do not exist? Should we add the list connectomes route there (always returning empty array)?

daniel-wer:
> What does the frontend do if a wkconnect dataset is opened, where the new routes do not exist? Should we add the list connectomes route there (always returning empty array)?

Yes, good call, that's the only route that should be needed :+1:

fm3:
I created the PR https://github.com/scalableminds/webknossos-connect/pull/127 to restore compatibility. That one should be probably merged first.
During testing I noticed that this branch is quite out of date with the master, you may want to reserve some time to check if everything still works after master merge :)

daniel-wer:
> During testing I noticed that this branch is quite out of date with the master, you may want to reserve some time to check if everything still works after master merge :)

Do you mean this `connectome-viewing` branch? If so, I merged the current master, yesterday, so it shouldn't be that out of date :thinking: Still, we'll definitely test the new functionality before merging :+1:

fm3:
You are of course right, I just did not realize I did not pull before testing.

.circleci/config.yml
```diff
@@ -278,8 +278,9 @@ jobs:
           command: curl -X POST --fail https://master.webknossos.xyz/data/triggers/checkInboxBlocking?token=$WK_AUTH_TOKEN
       - run:
           name: Run screenshot-tests
+          # TODO: Replace with master before merging
           command: |
-            URL=https://master.webknossos.xyz/ \
+            URL=https://connectomeviewing.webknossos.xyz/ \
```
philippotto, line 283: todo

.circleci/config.yml
```diff
@@ -293,6 +294,8 @@ workflows:
           filters:
             tags:
               only: /.*/
+      # TODO: Remove before merging
```
philippotto, line 297: todo

philippotto:
> * Currently the connectome tab is always visible. Since it is kind of a niche functionality, should it maybe only be visible for specific organizations (that want to test this functionality and/or we know to be in the connectomics field)? And if so, does that have any implications for the viewport configurations (if the tab is sometimes there and sometimes not)?

Yeah, I think, it would currently be best if the tab could be enabled/disabled via a feature flag so that we can control it per organization. Otherwise, the tab might cause a lot of irritation, as no one will know how to use it. I think, it should "just work" :tm: to adapt `default_layout_configs.js` so that it uses the features. Hopefully, this will work. The layout version needs to be bumped then, too, I assume..

philippotto (commented):
Code looks great :) Didn't test, yet, though, since there are some empty checkboxes still.

philippotto:
> > Yeah, I think, it would currently be best if the tab could be enabled/disabled via a feature flag so that we can control it per organization. Otherwise, the tab might cause a lot of irritation, as no one will know how to use it. I think, it should "just work" tm to adapt default_layout_configs.js so that it uses the features. Hopefully, this will work. The layout version needs to be bumped then, too, I assume.
> 
> Hmm, yeah, I think this method would work, but I really dislike that each time another organization is added, (one needs to remember to increase the layout version and) all layout configs become invalidated :/ What I could try to write would be a kind of evolution that runs after loading the persisted layout (or loading the default one) and adds or removes the Connectome Tab, depending on the feature flag thinking This would also allow us to add new tabs in the future without having to reset all layout configs for all users.

You are totally right. Dynamically adding/removing the tab would be best probably.

conf/application.conf
```diff
@@ -87,6 +87,8 @@ features {
   exportTiffMaxVolumeMVx = 1024
   exportTiffMaxEdgeLengthVx = 8192
   defaultToLegacyBindings = false
+  # The Only valid item value is currently "ConnectomeView":
+  optInTabs = ["ConnectomeView"]
```
philippotto, line 91: remove before merging
```suggestion
  optInTabs = []
```

philippotto (approved):

PR #5904, opened 2021-12-13
Fix job run error message for foreign organization

fm3:
When a user attempts to run a job on a dataset of a foreign organization, the organization should be found (GlobalAccessContext permits that), but the next line should give them a readable error message (“Currently nuclei inferral is only allowed for datasets of your own organization.”).
The old code just replied with organization couldn’t be found


jstriebel (approved):
LGTM :+1:

PR #5913, opened 2021-12-20
Upgrade to log4j 2.17.0

jstriebel:
Upgrades log4j to the newest version, mitigating more vulnerabilities. Not sure if wk-wrap must also be upgraded so that this has an effect, going to check this in the built container.

- [x] Ready for review

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -4,7 +4,7 @@ import sbt._
 object Dependencies {
   private val akkaVersion = "2.6.14"
   private val akkaHttpVersion = "10.2.6"
-  private val log4jVersion = "2.15.0"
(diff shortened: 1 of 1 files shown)


normanrz (approved):

jstriebel:
Unfortunately, the dependencies still include the old log4j version, we need to bump it in wk-wrap as well: https://github.com/scalableminds/webknossos-wrap/pull/72

```
docker run --rm -ti --entrypoint ls scalableminds/webknossos:log4j_217 /webknossos/lib | grep log4j
 org.apache.logging.log4j.log4j-api-2.15.0.jar
 org.apache.logging.log4j.log4j-core-2.15.0.jar
```

PR #5914, opened 2021-12-20
Upgrade webknossos-wrap dependency to 1.1.15

fm3:
Upgrades wk-wrap to 1.1.15 which drops the log4j dependency.

------
- [x] Ready for review


jstriebel (approved):
The log4j dependency is not part of `/webknossos/lib` now anymore, not sure where webknossos' dependency is listed now, but seems fine :+1:

PR #5918, opened 2021-12-21
fix dataset edit button shown to users who cant edit

### LINKED ISSUE #5791: Dataset edit button is shown even to users who cannot edit the datast settings
In the view mode (or annotation mode), in the info tab, next to the dataset name, an icon is shown that links to the dataset edit page. This should only be shown if the user can actually visit that edit page (i.e. is in the dataset’s organization and is dataset manager or admin)

Also, could we use a different icon (e.g. cogwheel)? The edit stylus icon is the same as for editing the annotation name, which is confusing to me

Dagobert42:
### Steps to test:
1. open dataset and use the changed edit icon in the info tab, next to the dataset name
2. share the current dataset to a secondary account
3. verify that the icon does not appear in read-only mode

### Issues:
- fixes #5791 

------
- [✓] Ready for review


philippotto (approved):
Excellent :+1: Feel free to merge :ship:

PR #5921, opened 2021-12-22
Add agglomerate skeleton to context menu

### LINKED ISSUE #5762: Make agglomerate skeleton loading feature available via context menu
## Detailed Description

Make the "Shift + Middleclick" feature to load agglomerate skeletons also available via the context menu.

## Context

Would be a great improvement for the usability of the feature as the `Shift + Middleclick` shortcut is not discoverable and the button in the segmentation layer setting can only be used to load the centered agglomerate.

Dagobert42:
### Steps to test:
1. open [dataset](https://www.notion.so/scalableminds/Test-Data-c0563be9c4a4499dae4e16d9b2497cfb#4f27c86bc8fc485e9d04b0dc1b1488c9) with ID mapping 
2. right click context menu on cell and check tooltip on Create Agglomerate Skeleton (should be disabled)
3. activate ID mapping and try Create Agglomerate Skeleton (should work)

### Issues:
- fixes #5762 

------
- [✓] Ready for review

### DIFF
--- a/frontend/javascripts/oxalis/controller/combinations/segmentation_handlers.js
+++ b/frontend/javascripts/oxalis/controller/combinations/segmentation_handlers.js
@@ -4,32 +4,42 @@ import Model from "oxalis/model";
 import { calculateGlobalPos } from "oxalis/model/accessors/view_mode_accessor";
 import { getMappingInfo } from "oxalis/model/accessors/dataset_accessor";
 import { loadAgglomerateSkeletonAction } from "oxalis/model/actions/skeletontracing_actions";
-import Store from "oxalis/store";
+import Store, { type OxalisState } from "oxalis/store";
 import Toast from "libs/toast";
 import messages from "messages";
 import { clickSegmentAction } from "oxalis/model/actions/volumetracing_actions";
 import api from "oxalis/api/internal_api";
 import { getSegmentIdForPosition } from "oxalis/controller/combinations/volume_handlers";
 
-export async function hasAgglomerateMapping() {
-  const state = Store.getState();
+export function hasAgglomerateMapping(state: OxalisState) {
   const segmentation = Model.getVisibleSegmentationLayer();
   if (!segmentation) {
-    return false;
+    return {
+      value: false,
+      reason: "",
+    };
   }
   const { mappingName, mappingType, mappingStatus } = getMappingInfo(
     state.temporaryConfiguration.activeMappingByLayer,
     segmentation.name,
   );
   if (mappingName == null || mappingStatus !== MappingStatusEnum.ENABLED) {
-    Toast.error(messages["tracing.agglomerate_skeleton.no_mapping"]);
-    return false;
+    return {
+      value: false,
+      reason: messages["tracing.agglomerate_skeleton.no_mapping"],
+    };
   }
+
   if (mappingType !== "HDF5") {
-    Toast.error(messages["tracing.agglomerate_skeleton.no_agglomerate_file"]);
-    return false;
+    return {
+      value: false,
+      reason: messages["tracing.agglomerate_skeleton.no_agglomerate_file"],
+    };
   }
-  return true;
+  return {
+    value: true,
+    reason: "",
+  };
 }
 
 export async function handleAgglomerateSkeletonAtClick(clickPosition: Point2) {
@@ -49,12 +59,16 @@ export async function loadAgglomerateSkeletonAtPosition(position: Vector3) {
     segmentation.name,
   );
 
-  if (mappingName && hasAgglomerateMapping()) {
+  const isAgglomerateMappingEnabled = hasAgglomerateMapping(state);
+
+  if (mappingName && isAgglomerateMappingEnabled.value) {
     const renderedZoomStep = api.data.getRenderedZoomStepAtPosition(segmentation.name, position);
 
     const cellId = segmentation.cube.getMappedDataValue(position, renderedZoomStep);
 
     Store.dispatch(loadAgglomerateSkeletonAction(segmentation.name, mappingName, cellId));
+  } else {
+    Toast.error(isAgglomerateMappingEnabled.reason);
   }
 }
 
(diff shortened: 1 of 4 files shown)


frontend/javascripts/oxalis/controller/combinations/segmentation_handlers.js
```diff
@@ -4,45 +4,72 @@ import Model from "oxalis/model";
...
 import { clickSegmentAction } from "oxalis/model/actions/volumetracing_actions";
 import api from "oxalis/api/internal_api";
 import { getSegmentIdForPosition } from "oxalis/controller/combinations/volume_handlers";
 
+export function hasAgglomerateMapping(state: OxalisState) {
+  const segmentation = Model.getVisibleSegmentationLayer();
+  if (!segmentation) {
+    return {
+      value: false,
+      reason: "",
```
philippotto, line 19:
```suggestion
      reason: "A segmentation layer needs to be visible to load an agglomerate skeleton.",
```

frontend/javascripts/oxalis/controller/combinations/segmentation_handlers.js
```diff
@@ -4,45 +4,72 @@ import Model from "oxalis/model";
...
+      reason: messages["tracing.agglomerate_skeleton.no_agglomerate_file"],
+    };
+  }
+  return {
+    value: true,
+    reason: "",
+  };
+}
+
 export async function handleAgglomerateSkeletonAtClick(clickPosition: Point2) {
```
philippotto, line 45: Can you check whether this function is still used? If not, it can be removed.

frontend/javascripts/oxalis/view/context_menu.js
```diff
@@ -510,6 +515,19 @@ function NoNodeContextMenuOptions(props: NoNodeContextMenuProps) {
...
           </Menu.Item>,
+
+          <Menu.Item
+            className="node-context-menu-item"
+            key="load-agglomerate-skeleton"
+            disabled={!isAgglomerateMappingEnabled.value}
+            onClick={() => loadAgglomerateSkeletonAtPosition(globalPosition)}
+          >
+            {isAgglomerateMappingEnabled.value ? (
+              "Create Agglomerate Skeleton"
```
philippotto, line 526: @daniel-wer Do you think this wording is good? The other button (in the left sidebar) uses "Import Skeleton for Centered Segment" which is too unspecific, as the mapping context is missing in the context menu. So, I like the suggested wording more. Not sure whether "create" should be changed to "import" or "load"..

And another question for you, @daniel-wer: Do you agree that the "Import Skeleton for Centered Segment" button should be removed, since the new context menu entry covers this?

philippotto (changes requested):
Didn't test, yet, but already looks pretty neat :necktie: Only left some slight remarks and asked input from @daniel-wer :smile: :)

frontend/javascripts/oxalis/controller/combinations/segmentation_handlers.js
Dagobert42: Seems like the middle click was the only place this was used, so unless that functionality is supposed to make a return I will delete the method

philippotto: Exactly, it can be removed :+1:

frontend/javascripts/oxalis/view/context_menu.js
daniel-wer: > Do you think this wording is good? The other button (in the left sidebar) uses "Import Skeleton for Centered Segment" which is too unspecific, as the mapping context is missing in the context menu. So, I like the suggested wording more. Not sure whether "create" should be changed to "import" or "load"..

I'd go with `Import Agglomerate Skeleton` as that most closely resembles what's actually happening.

> And another question for you, @daniel-wer: Do you agree that the "Import Skeleton for Centered Segment" button should be removed, since the new context menu entry covers this?

Yes, I think it can be removed. However, do we have the possibility to tell users where the functionality moved to? Otherwise, I imagine, this could be a bit frustrating. Maybe, leave a disabled button with a tooltip explaining to use the context menu for a couple of weeks, then remove it?

Also, on another note, I didn't expect the shortcut to be removed. At least for me (and I know from at least one other "power user"), it's a use case to quickly load a bunch of agglomerate skeletons which would be quite a bit faster if a shortcut was available. Would it be easy to add that back and mention it in the context menu as is done for other shortcuts? This is not a deal-breaker for me and we could also merge this without the shortcut, but I would expect a discuss issue regarding that, rather quickly :see_no_evil:

philippotto: Sounds very reasonable to me :+1: @Dagobert42 could you then please do the following?

- rename menu item to "Import Agglomerate Skeleton"
- "leave a disabled button with a tooltip explaining to use the context menu" + open a follow-up issue to remove the button again in a couple of weeks
- re-add the shift+middleclick shortcut
- mention the shortcut in the context menu (for the middleclick you can probably reuse the `icon-statusbar-mouse-wheel.svg` icon. just search the codebase for its usage).

Let me know in case you have questions or get stuck somewhere! Thank you!

Dagobert42:
Here's how the last changes turned out:

Screenshot 2022-01-05 at 11 13 08

Screenshot 2022-01-05 at 10 11 03

@philippotto Could you please verify that the middle click behaviour works on your machine?

frontend/javascripts/oxalis/view/left-border-tabs/mapping_settings_view.js
```diff
@@ -165,13 +165,14 @@ class MappingSettingsView extends React.Component<Props, State> {
 
     // Only show the option to import a skeleton from an agglomerate file if an agglomerate file mapping is activated.
     const shouldRender = this.props.isMappingEnabled && mappingName != null && isAgglomerateMapping;
-    const isDisabled = !this.props.allowUpdate;
-    const disabledMessage = "Skeletons cannot be imported in view mode or read-only tracings.";
+    const isDisabled = true;
+    const disabledMessage =
+      "Agglomerate Skeletons can now be imported through the context menu or with Shift + Middle Click.";
```
philippotto, line 170:
```suggestion
      "Agglomerate Skeletons should now be imported through the context menu or with Shift + Middle Click.";
```

philippotto (approved):
Excellent work :+1: Just see my one wording suggestions. Afterwards, this PR can be merged from my point of view :1st_place_medal:

philippotto:
Ah, the dangers of "auto-merge" ^^ I opened https://github.com/scalableminds/webknossos/pull/5941.

Dagobert42:
Ah thanks for the quick fix :)

PR #5922, opened 2021-12-22
Add neuron reconstruction job backend and frontend part

MichaelBuessemeyer:
This PR adds the necessary parts to the frontend and backend to make the new nuclei reconstruction job of the worker callable.

Open TODOs:
- [x] The job needs a description
- [x] The job needs an example image

- https://___.webknossos.xyz

### Steps to test:
- Enable jobs via `yarn enable-jobs`.
- Open a dataset in view mode.
- In the dataset info tab open the processing jobs menu and select the nuclei reconstruction menu entry.
- Go to the jobs list and wait for you job to finish.

### Issues:
- No issue exists in the repo for this

------
    - I think this is not needed as this feature is still experimental, right?
- [x] Ready for review


MichaelBuessemeyer:
@philippotto I am currently working on making the bounding box of the volume that should be predicted for neurons configurable in the frontend.

The approach I went for is having another select (just like it is right now when the dataset has multiple color layers).
Instead of giving the user the possibility to enter the bounding box via numbers, which I think is quite bothersome, I went for the users bounding boxes. The UI now lets the user select one of the bounding boxes of this annotation. Do you think this is alrighty this way? 

I think the advantages are that the user can / must explicitly but quite easy via the bounding box tool, select the volume for the prediction. Therefore the selected volume is first visually displayed to the user so the user can check again whether thats exactly the volume the user wants.
On the other hand, a simple number input that parses 6 numbers is better for automation.
My argument against the simple number input is that this is a UI and it should be more "visual" instead of simply entering numbers. If automation is needed, then this might be a good candidate for wk-libs?

Do you think my arguments make sense and that I chose the right approach?

Thanks for you feedback. 🐱

philippotto:
@MichaelBuessemeyer Sounds great! In the rare (?) scenario that a user happens to have 6 digits in their clipboard, they can always go to the bounding box tab to paste it there. We want to encourage the use of the bounding box tools and therefore it makes sense to reuse that. Good thinking!

MichaelBuessemeyer:
> In the rare (?) scenario that a user happens to have 6 digits in their clipboard, they can always go to the bounding box tab to paste it there.

Thanks , thats a very smart solution to that problem 👍

frontend/javascripts/oxalis/view/right-border-tabs/starting_job_modals.js
```diff
@@ -0,0 +1,292 @@
...
+      onCancel={handleClose}
+      visible
+      width={700}
+      footer={null}
+    >
+      {description}
+      <br />
+      <div style={{ textAlign: "center" }}>
+        <img
+          //TODO: get an image
```
MichaelBuessemeyer, line 189: This is still an open TODO for me :)

frontend/javascripts/oxalis/view/right-border-tabs/starting_job_modals.js
```diff
@@ -0,0 +1,292 @@
...
+    userBoundingBoxes: userBB,
+  };
+}
+
+export const NeuronInferralModal = connect<Props, OwnProps, _, _, _, _>(mapStateToProps)(
+  _NeuronInferralModal,
+);
+export const NucleiInferralModal = connect<Props, OwnProps, _, _, _, _>(mapStateToProps)(
+  _NucleiInferralModal,
+);
```
MichaelBuessemeyer, line 292: Do you have a better Idea to get the dataset and bounding boxes to the modals without code duplication @philippotto?

Currently, it is a mixture of hooks and none-hooks.

MichaelBuessemeyer (commented):
@philippotto I just checked the frontend part of the neuron inferral again. It is now ready for your review. Could you please take over that review? 👀 📝 

Random emoji of this post: 🐧

frontend/javascripts/admin/job/job_list_view.js
```diff
@@ -204,6 +204,20 @@ class JobListView extends React.PureComponent<Props, State> {
...
+            <Link
+              to={`/datasets/${job.organizationName}/${job.result}/view`}
+              title="View Segmentation"
+            >
+              <EyeOutlined />
+              View
+            </Link>
+          )}
+        </span>
+      );
```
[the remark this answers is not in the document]
MichaelBuessemeyer: If you want, I could also change it in this pr.

MichaelBuessemeyer:
@fm3 could you please check the backend changes?

fm3 (commented):
Backend looks good 🎉

frontend/javascripts/oxalis/view/right-border-tabs/starting_job_modals.js
```diff
@@ -0,0 +1,292 @@
...
+      handleClose();
+    }
+  };
+
+  const ColorLayerSelection = (): Node =>
+    colorLayerNames.length > 1 ? (
+      <React.Fragment>
+        <p>
+          The detection approach uses a single color layer. Please select the layer that should be
+          used for detection.
```
philippotto, line 95:
```suggestion
          Please select the layer that should be used for the inferral.
```

frontend/javascripts/oxalis/view/right-border-tabs/starting_job_modals.js
```diff
@@ -0,0 +1,292 @@
...
+  const BoundingBoxSelection = (): Node =>
+    isBoundingBoxConfigurable ? (
+      <React.Fragment>
+        <p>
+          The detection requires you to set a bounding box for the volume where the detection shall
+          be done. As the detection takes a long time, it is suggested to only take a small volume
+          and not the whole dataset. To select a volume please create a bounding box of the desired
+          volume with the bounding box tool in the toolbar at the top. The created bounding boxes
+          will be available below for selection. The format of the Bounding Boxes is &quot;minX,
+          minY, minZ, width, height, depth&quot;.
```
philippotto, line 149: I tried to make it a bit shorter:
```suggestion
          Please select the bounding box for which the inferral should be computed. Note that large
          bounding boxes, can take very long.
          You can create a new bounding box for the desired volume with the bounding box tool in the toolbar at the top. 
          The created bounding boxes will be listed below. The format of the Bounding Boxes is &quot;minX,
          minY, minZ, width, height, depth&quot;.
```

frontend/javascripts/oxalis/view/right-border-tabs/starting_job_modals.js
```diff
@@ -0,0 +1,292 @@
...
+  return (
+    <StartingJobModal
+      dataset={dataset}
+      handleClose={handleClose}
+      userBoundingBoxes={userBoundingBoxes}
+      jobName="neuron inferral"
+      isBoundingBoxConfigurable
+      jobApiCall={async (colorLayerName, boundingBox) => {
+        if (!boundingBox) {
+          return Promise.resolve();
```
philippotto, line 248:
```suggestion
          return;
```

should also work afaik, since the function is already async and thus always returns a promise.

frontend/javascripts/oxalis/view/right-border-tabs/starting_job_modals.js
philippotto: I'd just switch to hooks all the way. There will still be a tad of duplication probably, but it will be negligible as far I as I see. Just use `useSelector((state: OxalisState) => ...);` for the two store properties each (`dataset` will need to be on the outer components, but `userBoundingBoxes` can be on the inner one, I think). This should make lines 16 to 27 and 279 to 292 obsolete.

frontend/javascripts/oxalis/view/right-border-tabs/starting_job_modals.js
```diff
@@ -0,0 +1,292 @@
...
+|};
+type OwnProps = {|
+  handleClose: () => void,
+|};
+
+type Props = {
+  ...StateProps,
+  ...OwnProps,
+};
+type StartingJoblModalProps = {
```
philippotto, line 28: Jobl is probably a typo :)

frontend/javascripts/oxalis/view/right-border-tabs/starting_job_modals.js
fm3: I think it would be cool allow inputting a bounding box in this modal directly. This is what I initially tried before reading the wall of text. But that would be more coding so let’s defer that maybe.

Philipp’s suggestion has a superfluous comma in `bounding boxes, can take very long`.
I’d suggest omitting the format hint at the end of the text here. If a user managed to create a bounding box, they just need to select it now. In fact, the format hint was the first thing I saw here so I immediately started typing.

frontend/javascripts/oxalis/view/right-border-tabs/starting_job_modals.js
MichaelBuessemeyer: Sadly, this does work for eslint.
> Async arrow function expected no return value. eslint(consistent-return)

 I think disabling this rule here is not work the effort. Do you agree?

frontend/javascripts/oxalis/view/components/setting_input_views.js
```diff
@@ -507,15 +507,9 @@ export class ColorSetting extends React.PureComponent<ColorSettingPropTypes> {
     style = style || {};
     return (
       <div
-        id="color-picker-wrapper"
```
MichaelBuessemeyer, line 510: As each of the bounding boxes rendered in the bounding box tab has its own diff, the name `color-picker-wrapper` is not unique and therefore a class

MichaelBuessemeyer (commented):
@philippotto I just finished applying your feedback. I also added the example screenshot for the neuron inferral. Could you please another look on this pr 👀 🏃‍♂️ 

Random emoji of this post: 🏎️

frontend/javascripts/oxalis/view/right-border-tabs/starting_job_modals.js
philippotto: Ok, then let's keep it as is :+1:

frontend/javascripts/oxalis/view/right-border-tabs/starting_job_modals.js
```diff
@@ -186,8 +187,7 @@ function StartingJobModal(props: StartingJoblModalProps) {
       <br />
       <div style={{ textAlign: "center" }}>
         <img
-          //TODO: get an image
-          src="/assets/images/nuclei_inferral_example.jpg"
+          src={`/assets/images/${_.snakeCase(jobName)}_example.jpg`}
```
philippotto, line 184: this will make it very hard to search where the file is used. I'd suggest to create a mapping object from jobName to path to make it easier to find. something like this:

```
const jobNameToImagePath = {
  "neuron_inferral": "neuron_inferral_example.jpg",
  ...
}
```

philippotto: also: the screenshot quality of the new image is quite poor. maybe set the zoom value to 1 to avoid resizing issues? also, I'd hide the "crosshair" in the image.

MichaelBuessemeyer: Thanks for the feedback. I did notice that the image had such a bad resolution. It should be better blazingly sharp 🗡️  now.

frontend/javascripts/oxalis/view/right-border-tabs/starting_job_modals.js
```diff
@@ -0,0 +1,270 @@
...
+      description={
+        <>
+          <p>
+            Start a job that automatically detects the neurons for this dataset. This job creates a
+            copy of this dataset once it has finished. The new dataset will contain the new
+            segmentation which segments the neurons of the dataset.
+          </p>
+          <p>
+            <b>
+              Note that this feature is still experimental and takes quite a long time. Thus we
```
philippotto, line 260:
```suggestion
              Note that this feature is still experimental and can take a long time. Thus we
```

philippotto (commented):
Awesome! I could test it and after fixing an issue in vx, it worked very well :) I only left some cosmetic feedback for the UI. Merging this PR is blocked by the worker PR, anyway.

fm3:
Sorry I didn’t read through all of the code and comments just now, but quick question: what mechanism are you using to limit this feature to scm team members? the isSuperUser property is not currently exposed in the user json, is it?

philippotto:
> Sorry I didn’t read through all of the code and comments just now, but quick question: what mechanism are you using to limit this feature to scm team members? the isSuperUser property is not currently exposed in the user json, is it?

Oh, this is a good point. It's only using `activeUser.isAdmin`. Could you expose the property?

frontend/javascripts/oxalis/view/right-border-tabs/starting_job_modals.js
```diff
@@ -0,0 +1,270 @@
...
+      <>
+        {bbox.name}{" "}
+        <div
+          className="color-display-wrapper"
+          style={{
+            backgroundColor: colorAsHexString,
+            marginTop: -2,
+          }}
+        />{" "}
+        - {computeArrayFromBoundingBox(bbox.boundingBox).join(", ")}
```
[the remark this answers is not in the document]
MichaelBuessemeyer: Thanks for the suggestion 🎉 . This looks definitely better ✨

MichaelBuessemeyer:
@philippotto Could you please check the change that makes the neuron inferral only available to super users? 
I think after that this pr is ready to 🚢 .

philippotto (approved):
LGTM :)

Issue #5684, opened 2021-08-23
Volume data download sometimes contains truncated data.zip


fm3:
Downloaded a volume annotation. It worked via the allDataBlocking route directly, but when using the wk-side download, the inner zip seems to be incomplete. I expect there is some race condition when waiting for the data.zip. Note that I cannot say with certainty whether the problem is on wk side or tracingstore side.

fm3:
Unfortunately I cannot reproduce the issue anymore, neither in production nor test setup :/

jstriebel:
There seems to be the same or a similar problem with the annotation mentioned in https://github.com/scalableminds/webknossos-libs/issues/534, the annotation download contains a corrupted data.zip, and during the download the following logs occur:

```
2021-12-20 11:22:20,427 [ERROR] akka.dispatch.Dispatcher - Timeout on waiting for new data
java.io.IOException: Timeout on waiting for new data
at akka.stream.impl.io.InputStreamAdapter.$anonfun$read$5(InputStreamSinkStage.scala:171)
at scala.runtime.java8.JFunction0$mcI$sp.apply(JFunction0$mcI$sp.java:23)
at akka.stream.impl.io.InputStreamAdapter.executeIfNotClosed(InputStreamSinkStage.scala:131)
at akka.stream.impl.io.InputStreamAdapter.read(InputStreamSinkStage.scala:156)
at akka.stream.impl.io.InputStreamAdapter.read(InputStreamSinkStage.scala:145)
at play.api.libs.iteratee.Enumerator$.$anonfun$fromStream$2(Enumerator.scala:563)
at scala.runtime.java8.JFunction0$mcI$sp.apply(JFunction0$mcI$sp.java:23)
at akka.dispatch.MonitorableThreadFactory$AkkaForkJoinWorkerThread$$anon$3.block(ThreadPoolBuilder.scala:174)
at java.util.concurrent.ForkJoinPool.managedBlock(ForkJoinPool.java:3313)
at akka.dispatch.MonitorableThreadFactory$AkkaForkJoinWorkerThread.blockOn(ThreadPoolBuilder.scala:172)
```

When using the allDataBlocking route the download aborted ("network error") after roughly 30s, but nothing suspicious in the logs. The downloaded length varies across retries.

Internal details can be found [in this thread](https://scm.slack.com/archives/C5AKLAV0B/p1639996444001400).

fm3:
[...]

PR #5927, opened 2022-01-03
Fix race condition in volume data download

fm3:
Fixing a bug I presumably introduced in 2018, when I did not understand that the onComplete block of a Future is not guaranteed to be performed immediately after the Future is complete. This lead to the zip writing not being flushed at the time when sendFile started. Because of this, sendFile determined (and sent) a wrong contentLength, which broke the transfer, as the file was flushed and closed later, creating a file size mismatch.

This means we can remove the Thread.sleep from allDataBlocking. 

It _may_ be a fix for #5684 though I am not super confident there (since that one also sometimes happened when allDataBlocking itself worked ok)

### Steps to test:
- Create a farly large volume annotation (i.e. brush around the dataset in mag2, creating lots of mag1 data)
- Download that volume annotation
- Should contain an intact data.zip
- To ensure the error is fixed, check out master _remove the Thread.sleep_ , see that that the data.zip is broken
- To make testing easier, I queried the allDataBlocking route directly via curl

------
- [x] Ready for review


fm3:
Yes, that’s exactly right :)

philippotto:
Is this blocked by something?

fm3:
Nope, merging now, thanks for the reminder!

PR #5931, opened 2022-01-03
Rename Sample User

fm3:
Renames development mode default user SCM Boy to Sample User < sample @ scm.io >. Hoping that the shorter email address leads to less typing and gender neutrality leads to less exclusion.

- https://sampleuser.webknossos.xyz

### Steps to test:
- Start normally, log in as sample @ scm.io default password »secret« (or the internal dev deployment password)

------
- [x] Ready for review

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -272,7 +272,7 @@ jobs:
             wait
       - run:
           name: Refresh datasets
-          command: curl https://master.webknossos.xyz/data/triggers/checkInboxBlocking?token=secretScmBoyToken
+          command: curl https://master.webknossos.xyz/data/triggers/checkInboxBlocking?token=secretSampleUserToken
       - run:
           name: Run screenshot-tests
           command: |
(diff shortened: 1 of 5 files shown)


fm3:
As english is not my first language I am unsure about this, but the word Sample may be a bit unusual here. We could also consider changing all the »Sample« usages to »Example«. What do you think? Anybody more familiar with the language?

philippotto:
Kudos for picking this up :+1: 

> As english is not my first language I am unsure about this, but the word Sample may be a bit unusual here. We could also consider changing all the »Sample« usages to »Example«. What do you think?

I'm not sure either, but from experience with other software, I would say that the following are at least more common for the use case:
- test user
- dummy user
- default user

However, in the end, it's not too important in my opinion.

normanrz:
I think "Sample" is fine

philippotto (approved):

jstriebel:
Nice! We also need to change this in wk-libs with the new release, see https://github.com/scalableminds/webknossos-libs/search?q=secretScmBoyToken

PR #5933, opened 2022-01-04
Release 22.01.0

philippotto:
I also added a make-changelog.py which I adapted from https://github.com/scalableminds/webknossos-libs/blob/master/make_changelog.py to work with the unreleased/released dichotomy. Reviewing the python script is not strictly necessary in my opinion. I will battle test it in the next releases. For now, it produced the current PRs diff.

------
- [X] Updated [(unreleased) migration guide](../blob/master/MIGRATIONS.unreleased.md#unreleased) if applicable
- [X] Ready for review


CHANGELOG.released.md
```diff
@@ -7,6 +7,25 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Calendar Versioning](http://calver.org/) `0Y.0M.MICRO`.
 For upgrade instructions, please check the [migration guide](MIGRATIONS.released.md).
 
+## [22.01.0](https://github.com/scalableminds/webknossos/releases/tag/22.01.0) - 2022-01-04
+[Commits](https://github.com/scalableminds/webknossos/compare/21.11.2...22.01.0)
+
+### Added
+- Added tagging support for datasets. [#5832](https://github.com/scalableminds/webknossos/pull/5832)
```
fm3, line 14: This one could be a highlight maybe?

fm3 (approved):
LGTM, as you suggested I did not review the python script in detail.
I would suggest to move it into the tools directory if possible.

CHANGELOG.released.md
philippotto: Very good point! immediately forgot about the highlights-section after automating the changelog update process -.- I added a print-statement to the update script to remind about this.

philippotto:
> Note that after the move the paths in the script may have to be adapted :)

I thought so to, at first, but the paths are relative to the current working dir. So, if one executes the script from the repo's root, it works :)

PR #5946, opened 2022-01-07
Allow Cancelling Jobs

### LINKED ISSUE #5932: Add cancel functionality to jobs
## Detailed Description
Add a cancel feature to jobs that also cleans up any intermediate/unfinished data that has been produced.
This feature should be exposed through a button in the jobs list.

## Context
Some jobs need to be cancelled for user reasons (eg. re-run because of wrong parameter selected) or for technical reasons (eg. job is stuck or takes up too much resources).

* [x] Specific to long-running jobs (set `jobsEnabled=true` in `application.conf`)

fm3:
Corresponding PR in worker https://github.com/scalableminds/webknossos-worker/pull/82

New Protocol: 
 - user cancel sets manualState in postgres
 - worker polling gets those with manualState = Cancelled but State != Cancelled (to_cancel)
 - worker terminates those jobs if running
 - worker sends state = Cancelled to wk for those jobs (if it was still running at the time or not)

TODO:
- [x] make latestRunId optional in JobStatus, also in sql
- [x] evolution to add CANCELLED state

### Steps to test:
- set up worker locally, connect to wk via `yarn enable-jobs`
- schedule a job
- in jobs page, click cancel
- should be cancelled (should not be scheduled if it hasn’t already been, should be killed if running)

### Issues:
- fixes #5932 

------
- [x] Ready for review


app/controllers/JobsController.scala
```diff
@@ -32,12 +32,12 @@ class JobsController @Inject()(jobDAO: JobDAO,
   def status: Action[AnyContent] = sil.SecuredAction.async { implicit request =>
     for {
       _ <- Fox.successful(())
-      jobCountsByStatus <- jobDAO.countByStatus
+      jobCountsByState <- jobDAO.countByState
```
fm3, line 35: just renamed this here to match  the class names

app/models/job/Job.scala
```diff
@@ -217,10 +233,11 @@ class JobDAO @Inject()(sqlClient: SQLClient)(implicit ec: ExecutionContext)
...
 
-  def countByStatus: Fox[Map[String, Int]] =
+  def countByState: Fox[Map[String, Int]] =
     for {
       result <- run(sql"""select state, count(_id)
                            from webknossos.jobs_
+                           where manualState is null
                            group by state
                            order by state
                            """.as[(String, Int)])
```
jstriebel, line 243: Not sure what this is used for, but shouldn't we also count the manualState fields and add those, since they have precedence?

jstriebel (commented):
Nice work, I'm just find the protocol very confusing:

> New Protocol: 
> - user cancel sets manualState in postgres
> - worker polling gets those with manualState = Cancelled but State != Cancelled (to_cancel)
> - worker terminates those jobs if running
> - worker sends state = Cancelled to wk for those jobs (if it was still running at the time or not)

To me it seems like we should rather use an additional state for cancelled but not cleaned up jobs (which might also be shown as `cancelled` to the frontend), or a new field to indicate any state between webknossos and wk-worker (this one could have a state `cancelling`). Using the manualState to encode this seems like a hack to me, and one that is quite complicated to understand from the code if you don't know the protocol exactly. What do you think, @fm3?

jstriebel (approved):
As discussed in person, we'll leave the state management as-is. It changes the semantics slightly and needs some more comments. One can thing of the basic `state` as the wk-worker state, and the `manualState` as anything that might be set from the outside (admin or user).

PR #5949, opened 2022-01-10
Prevent organization creation with existing email address

### LINKED ISSUE #5861: wk.org signup with existing email creates invalid state
Should check first that both organization and email are free, if any is not present, nothing should be created.

Currently, an organization is created, but then no user. This reserves the name and cannot be reused except by fixing it in the database.

## Your Environment for bug
- [x] Specific to webKnossos.org (set `isDemoInstance=true` in `application.conf`)

youri-k:
- https://___.webknossos.xyz

### Steps to test:
- set up demo instance of wk
- register user
- try to register another user with the same email => should not work
- register user with different email => should work

### Issues:
- fixes #5861 

------
- [x] Ready for review


fm3 (approved):
:tada:

PR #5950, opened 2022-01-11
Validate file paths on dataset upload

fm3:
### Steps to test:
- Attempt to upload a dataset where a file contains paths that lead it outside of its parent directory (`b/../../a`) (This cannot currently be done via the frontend, custom requests will have to be constructed) → Error should be thrown

------
- [x] Ready for review


youri-k (approved):
LGTM :shipit:

PR #5953, opened 2022-01-12
Install jhdf5 via maven

fm3:
The jhdf5 dependency is now published via maven, so we don’t have to keep its binary files in the wk repository :tada: This will also make upgrading to a newer version easier.

Also, I removed the unused bintray dependency resolver.

I also in this PR: A few more comments I added to the wk backend when discussing some comples code with Youri

- https://jhdf5maven.webknossos.xyz

### Steps to test:
- Open dataset with agglomerate file, activate it, ID mapping should work and look plausible

----
- [x] Ready for review


youri-k (approved):
LGTM :shipit:

Issue #5715, opened 2021-09-08
Erasing in mag 2 can accidentally restore data from fallback layer in mag 1


philippotto:
## Context
Volume annotation with fallback layer in higher mag.

## Expected Behavior
Changes in mag 2 should propagate to the other mags in a consistent manner.

## Current Behavior
Mag 1 can have inconsistent data (not always; probably a race condition).

## Steps to Reproduce the bug

- [ ] Cannot reproduce the bug anymore / needs deeper investigation.

1. Create an annotation with a fallback layer and zoom to mag 2
2. Refresh the page (this should ensure that buckets from mag 1 are not loaded yet)
3. Erase data in mag 2
4. Save & Reload
5. Zoom in to mag 1 and check whether the data is consistent

## Your Environment for bug

- Browser name and version: ?
- Operating System and version: ?
- Version of WebKnossos (Release or Commit):
- [ ] Specific to long-running jobs (set `jobsEnabled=true` in `application.conf`)
- [ ] Specific to webKnossos.org (set `isDemoInstance=true` in `application.conf`)

philippotto:
Could be related to this line:

https://github.com/scalableminds/webknossos/blob/cda0fe94c078acf260511412c71d47d5199828f5/frontend/javascripts/oxalis/model/bucket_data_handling/bucket.js#L418

Due to the `||` operator the erased data does not survive. Fixing this is not super straightforward since we need to distinguish between "the data is 0 because the user erased this" and "the data is 0 because it wasn't loaded yet".

I'm quite certain that this explains a major part of the issue, but I don't think that it explains everything, since we also saw that entire buckets were restored to the original fallback layer even though these buckets were not erased, but relabelled.

philippotto:
@daniel-wer and I discussed this issue and agreed on a solution which would look something like this:

- for each bucket that the TemporalBucketManager holds, a binary "LabeledMask" needs to be maintained (0 = not labeled, 1 = labeled)
- when merging local bucket data with data from the backend, the LabeledMask can be used to decide which voxel should be used (0 => backend data, 1 => frontend data)
- a good spot to update the LabeledMask, would probably be in `applyVoxelMap`.
- note that not all volume mutating code uses `applyVoxelMap`. therefore,
  - either: refactor the code so that `applyVoxelMap` is used everywhere
  - or: adapt the other code mechanisms to update the LabeledMask, as well

Estimated work: 4d

PR #5955, opened 2022-01-12
Fix async volume bugs

### LINKED ISSUE #5796: "Only overwrite empty areas" is not respected when brushing in higher mag
> If in a volume annotation with fallback layer, "Only overwrite empty areas" is activated and I brush in mag 4, it looks like nothing is annotated (as would be correct, since there are segments), but after zooming in, parts of the data were overwritten (see screenshot)
  partially_overwritten_data

_Originally posted by @daniel-wer in https://github.com/scalableminds/webknossos/issues/5696#issuecomment-953108198_

### LINKED ISSUE #5936: Volume Annotation Inconsistency Bugs
Annotating volume data (without fallback) can lead to buggy / inconsistent data where cells are missing suddenly.

It's not very clear how the bug can be reproduced and what the culprit is, but roughly the following causes problems:

1. Annotate volume data in mag 4 (e.g., three different large cells).
2. Zoom in/out in-between.
3. Use undo
4. Use copy previous slice (`v` shortcut)
5. At some point (after reload?), some segments are missing

Example bug:
Before reload:
image

After reload:
image

philippotto:
Fix multiple bugs and conceptual problems with the volume annotation code. The biggest, conceptual change is that the back-end data is not merged with via a simple `||` operation anymore, as this does not work for erasure. Instead, "pending operations" are persisted which can be applied on the fetched back-end data. This also helps to respect the proper "overwrite mode" when propagating brushed data to other magnifications which were not loaded yet.

- https://fixvolumemerges.webknossos.xyz

### To do:
- [x] don't merge back-end data with `||` anymore but use an array with `pendingOperations` which rely on `LabeledVoxelMap`s
  - [X] use this for brushing/tracing
  - [x] ~use this for flood-fill~ the floodfill always operates on loaded data which is why it never can cause a merge operation. I refactored the code a bit, but unfortunately there is no way to make this really sturdy as there are no read-only typed arrays in JS / flow
  - [x] use this for "copy from previous slice"
  - [x] `label` in API (deprecate the API and ensure that it loads the data first before mutating to avoid merge operations)
- [x] find and fix bugs
  - [X] initialize `maybeUnmergedBucketLoadedPromise` at correct times (use `needsBackendData()` instead of `triggeredBucketFetch`)
  - [X] corruption of data types during compression/decompression
  - [X] don't merge already merged data 1 (due to using `bucketLoaded` event)
  - [X] don't merge already merged data 2 (within `applyAndGetRevertingVolumeBatch` after `getData`)
  - [X] race condition for adding a bucket to the undo set
  - [X] race condition if undo was executed before compression of backend data finished
- [X] add tests (especially for mutating buckets which are not loaded, yet)
- [X] misc
  - [X] mocked workers in node context should return promises instead of the raw values
  - [X] properly alert user about permanent errors and don't mention retry if none is happening
  - [X] clean up, add comments, rename variables for clarity
  - [X] don't ignore potential errors in `markBucketsAsNotDirty`
  - [X] don't allow brushing when wk is busy with undo/redo/floodfill

### Steps to test:
- see the new tests I added
- additionally, monkey-testing brushing/erasing/undo/redoing (potentially with slow internet connections) should be done
  - [X] performed by me
  - [X] performed by Tom

### Issues:
- fixes https://github.com/scalableminds/webknossos/issues/5715
- fixes #5796
- hopefully fixes #5936

------
- [x] Ready for review


frontend/javascripts/oxalis/model/bucket_data_handling/bucket.js
```diff
@@ -466,13 +552,21 @@ export class DataBucket {
...
-      this.data[i] = this.data[i] || newData[i];
+
+    if (this.pendingOperations.length > 0) {
+      for (const op of this.pendingOperations) {
+        op(fetchedData);
+      }
+      this.data = fetchedData;
+      this.pendingOperations = [];
+    } else {
+      // todo: ensure that everything is handled via pendingOperations.
```
philippotto, line 566: todo

frontend/javascripts/oxalis/model/bucket_data_handling/data_cube.js
```diff
@@ -527,6 +527,8 @@ class DataCube {
       // Add the bucket to the current volume undo batch, if it isn't already part of it.
       currentBucket.markAndAddBucketForUndo();
       // Mark the initial voxel.
+      // Todo: This should probably act on a LabeledVoxelMap? Or is it alright, since
```
philippotto, line 530: todo

frontend/javascripts/oxalis/model/sagas/save_saga.js
```diff
@@ -100,17 +101,39 @@ import window, { alert, document, location } from "libs/window";
...
 
-const byteArrayToLz4Array = createWorker(compressLz4Block);
+const _byteArrayToLz4Array = createWorker(compressLz4Block);
+const decompressToTypedArray = (
+  bucket: DataBucket,
+  compressedData: Uint8Array,
+): ?BucketDataArray => {
+  const decompressedBackendData = _byteArrayToLz4Array(compressedData, false);
+  if (decompressedBackendData == null) {
+    // Can this happen?
```
philippotto, line 111: @MichaelBuessemeyer Do you know whether lz4 decompression can return null? Old code seemed to handle this for some reason? See https://github.com/scalableminds/webknossos/pull/5955/files#diff-acc1d929544bdde3687f0e79f303c4f23b61ca97fafe5e4579202972d2e37acaL413

frontend/javascripts/oxalis/model/sagas/save_saga.js
```diff
@@ -100,17 +101,39 @@ import window, { alert, document, location } from "libs/window";
...
 const UndoRedoRelevantBoundingBoxActions = AllUserBoundingBoxActions.filter(
   action => action !== "SET_USER_BOUNDING_BOXES",
 );
 
 type UndoBucket = {
   zoomedBucketAddress: Vector4,
-  data: Uint8Array,
-  backendData?: Uint8Array,
+  data: Uint8Array, // it's Uint8Array because it's compressed
+  backendData?: Uint8Array, // it's Uint8Array because it's compressed
```
philippotto, line 134: consider renaming to `compressedData` etc

frontend/javascripts/oxalis/model/sagas/save_saga.js
```diff
@@ -522,10 +539,44 @@ function* applyStateOfStack(
 function mergeDataWithBackendDataInPlace(
   originalData: BucketDataArray,
   backendData: BucketDataArray,
+  pendingOperations: Array<(BucketDataArray) => void>,
+  log: boolean,
 ) {
-  for (let i = 0; i < originalData.length; ++i) {
-    originalData[i] = originalData[i] || backendData[i];
+  if (originalData.length !== backendData.length) {
```
philippotto, line 544: The old code occasionally merged uint8 with uint32 data since the decompression always returned uint8.

frontend/javascripts/oxalis/model/sagas/save_saga.js
```diff
@@ -555,20 +606,14 @@ function* applyAndGetRevertingVolumeBatch(
     // Prepare a snapshot of the bucket's current data so that it can be
     // saved in an VolumeUndoState.
     let bucketData = null;
+    const currentPendingOperations = bucket.pendingOperations.slice();
     if (bucket.hasData()) {
       // The bucket's data is currently available.
       bucketData = bucket.getData();
       if (compressedBackendData != null) {
         // If the backend data for the bucket has been fetched in the meantime,
-        // we can first merge the data with the current data and then add this to the undo batch.
```
philippotto, line 563: I think this code was simply wrong and I also managed to produce a failing test which now succeeds with the new change.

philippotto:
@daniel-wer @MichaelBuessemeyer Could you both have a look at this PR? I added several comments (code-wise and github-wise) which should explain the changes. Would be great to already get your feedback. Right now, two things are probably left to do:

- hope that the CI passes (I already saw in an older run that a test failed which passes locally for me -.-)
- ensure that the `pendingOperations` is always used (which means that `applyVoxelMap` has to be used everywhere). This is not really the case for the "copy from previous slice" and the floodfill feature.

However, I'll pause this PR (at least, for today). So, feel free to already have a look at the PR.

philippotto:
Hm, first monkey-tests went poorly, unfortunately. I annotated three cells in mag 4 for some slices and then used the undo button several times. The main problems are:

- [ ] the undo doesn't immediately change the rendered data
- [ ] when reloading the volume layer, the backend partly replies with 500 errors. @fm3 Do you have an idea what is going on? Below is an excerpt from the logs. I changed some stuff in the frontend regarding the lz4 compression for the undo stack, but this should be independent of the lz4 compression which is used for sending it to the backend (`byteArrayToLz4Base64` vs. `compressLz4Block`).

```
2022-01-18 15:10:49,738 [ERROR] com.scalableminds.webknossos.tracingstore.tracings.volume.VolumeTracingBucketProvider - Failed to LZ4-decompress volume bucket (key: a6581026-d0f1-48ef-9790-b6e215a54276/4/164-[0,4,3], 0 bytes, version 244, expected uncompressed size 131072)
2022-01-18 15:10:49,738 [ERROR] com.scalableminds.webknossos.tracingstore.tracings.volume.VolumeTracingBucketProvider - Failed to LZ4-decompress volume bucket (key: a6581026-d0f1-48ef-9790-b6e215a54276/4/111-[7,1,3], 0 bytes, version 230, expected uncompressed size 131072)
2022-01-18 15:10:49,738 [ERROR] ErrorHandler -
! @7md13c84j - Internal server error, for (POST) [/tracings/volume/a6581026-d0f1-48ef-9790-b6e215a54276/data?token=xD4Ykfv5RLwcyT6cUennAw] ->
play.api.UnexpectedException: Unexpected exception[ArrayIndexOutOfBoundsException: Array index out of range: 0]
at play.api.http.HttpErrorHandlerExceptions$.throwableToUsefulException(HttpErrorHandler.scala:358)
at play.api.http.DefaultHttpErrorHandler.onServerError(HttpErrorHandler.scala:264)
at play.core.server.AkkaHttpServer$$anonfun$2.applyOrElse(AkkaHttpServer.scala:430)
at play.core.server.AkkaHttpServer$$anonfun$2.applyOrElse(AkkaHttpServer.scala:422)
at scala.concurrent.Future.$anonfun$recoverWith$1(Future.scala:417)
at akka.dispatch.BatchingExecutor$AbstractBatch.processBatch(BatchingExecutor.scala:63)
at akka.dispatch.BatchingExecutor$BlockableBatch.$anonfun$run$1(BatchingExecutor.scala:100)
at scala.runtime.java8.JFunction0$mcV$sp.apply(JFunction0$mcV$sp.java:23)
Caused by: java.lang.ArrayIndexOutOfBoundsException: Array index out of range: 0
```

philippotto:
>  when reloading the volume layer, the backend partly replies with 500 errors. @fm3 Do you have an idea what is going on? Below is an excerpt from the logs. I changed some stuff in the frontend regarding the lz4 compression for the undo stack, but this should be independent of the lz4 compression which is used for sending it to the backend (byteArrayToLz4Base64 vs. compressLz4Block).

Hm, it's probably a frontend bug after all (both symptoms stem from the same bug, as far as I can see). Will dig deeper. Sorry, for pinging prematurely.

philippotto:
> Hm, first monkey-tests went poorly, unfortunately. I annotated three cells in mag 4 for some slices and then used the undo button several times. The main problems are:
> 
> * [x]  the undo doesn't immediately change the rendered data
> * [x]  when reloading the volume layer, the backend partly replies with 500 errors. @fm3 Do you have an idea what is going on? Below is an excerpt from the logs. I changed some stuff in the frontend regarding the lz4 compression for the undo stack, but this should be independent of the lz4 compression which is used for sending it to the backend (`byteArrayToLz4Base64` vs. `compressLz4Block`).
> 
> ```
> 2022-01-18 15:10:49,738 [ERROR] com.scalableminds.webknossos.tracingstore.tracings.volume.VolumeTracingBucketProvider - Failed to LZ4-decompress volume bucket (key: a6581026-d0f1-48ef-9790-b6e215a54276/4/164-[0,4,3], 0 bytes, version 244, expected uncompressed size 131072)
> 2022-01-18 15:10:49,738 [ERROR] com.scalableminds.webknossos.tracingstore.tracings.volume.VolumeTracingBucketProvider - Failed to LZ4-decompress volume bucket (key: a6581026-d0f1-48ef-9790-b6e215a54276/4/111-[7,1,3], 0 bytes, version 230, expected uncompressed size 131072)
> 2022-01-18 15:10:49,738 [ERROR] ErrorHandler -
> ! @7md13c84j - Internal server error, for (POST) [/tracings/volume/a6581026-d0f1-48ef-9790-b6e215a54276/data?token=xD4Ykfv5RLwcyT6cUennAw] ->
> play.api.UnexpectedException: Unexpected exception[ArrayIndexOutOfBoundsException: Array index out of range: 0]
> at play.api.http.HttpErrorHandlerExceptions$.throwableToUsefulException(HttpErrorHandler.scala:358)
> at play.api.http.DefaultHttpErrorHandler.onServerError(HttpErrorHandler.scala:264)
> at play.core.server.AkkaHttpServer$$anonfun$2.applyOrElse(AkkaHttpServer.scala:430)
> at play.core.server.AkkaHttpServer$$anonfun$2.applyOrElse(AkkaHttpServer.scala:422)
> at scala.concurrent.Future.$anonfun$recoverWith$1(Future.scala:417)
> at scala.concurrent.impl.Promise.$anonfun$transformWith$1(Promise.scala:41)
> at scala.concurrent.impl.CallbackRunnable.run(Promise.scala:64)
> at akka.dispatch.BatchingExecutor$AbstractBatch.processBatch(BatchingExecutor.scala:63)
> at akka.dispatch.BatchingExecutor$BlockableBatch.$anonfun$run$1(BatchingExecutor.scala:100)
> at scala.runtime.java8.JFunction0$mcV$sp.apply(JFunction0$mcV$sp.java:23)
> Caused by: java.lang.ArrayIndexOutOfBoundsException: Array index out of range: 0
> ```

Should be fixed with 161f81d. The tests passed, since there are no workers in the node context (the functions are simply called which makes them synchronous).

frontend/javascripts/test/sagas/volumetracing/volumetracing_saga_integration.spec.js
```diff
@@ -556,3 +705,164 @@ test.serial("Brushing/Tracing with undo (II)", async t => {
...
+  const oldCellId = 11;
+  t.context.mocks.Request.sendJSONReceiveArraybufferWithHeaders = createBucketResponseFunction(
+    Uint16Array,
+    oldCellId,
+    500,
+  );
+
+  console.log("zoom step:", getRequestLogZoomStep(Store.getState()));
+  Store.dispatch(setZoomStepAction(4));
+  console.log("zoom step:", getRequestLogZoomStep(Store.getState()));
```
daniel-wer, line 719: console.logs probably can be (partly) deleted

frontend/javascripts/test/sagas/volumetracing/volumetracing_saga_integration.spec.js
```diff
@@ -556,3 +705,164 @@ test.serial("Brushing/Tracing with undo (II)", async t => {
...
+  for (let zoomStep = 0; zoomStep <= 5; zoomStep++) {
+    console.log("request at zoomStep", zoomStep);
+    t.is(
+      await t.context.api.data.getDataValue(volumeTracingLayerName, [0, 0, 0], zoomStep),
+      oldCellId,
+      `Center should have old value at zoomstep=${zoomStep}`,
+    );
+  }
+
+  console.log("Set overwriteMode to OVERWRITE_EMPTY");
```
daniel-wer, line 782:
```suggestion
  console.log("Set overwriteMode to OVERWRITE_ALL");
```

frontend/javascripts/test/sagas/volumetracing/volumetracing_saga_integration.spec.js
```diff
@@ -556,3 +705,164 @@ test.serial("Brushing/Tracing with undo (II)", async t => {
...
+  Store.dispatch(setToolAction(AnnotationToolEnum.ERASE_BRUSH));
+
+  Store.dispatch(startEditingAction(paintCenter, OrthoViews.PLANE_XY));
+  Store.dispatch(addToLayerAction(paintCenter));
+  Store.dispatch(finishEditingAction());
+
+  console.log("Save");
+  await t.context.api.tracing.save();
+
+  console.log("Load");
```
daniel-wer, line 797: Some of the console.logs can probably be removed

frontend/javascripts/test/sagas/volumetracing/volumetracing_saga_integration.spec.js
```diff
@@ -556,3 +705,164 @@ test.serial("Brushing/Tracing with undo (II)", async t => {
...
+  // Reload buckets which might have already been loaded before swapping the sendJSONReceiveArraybufferWithHeaders
+  // function.
+  await t.context.api.data.reloadAllBuckets();
+  const volumeTracingLayerName = t.context.api.data.getVolumeTracingLayerIds()[0];
+
+  const paintCenter = [0, 0, 0];
+  // This particular brushSize used to trigger a bug. It should not be changed.
+  const brushSize = 263;
+
+  for (let zoomStep = 0; zoomStep <= 5; zoomStep++) {
```
daniel-wer, line 773: The test says "... where mag 1 is unloaded", but don't these lines load the data in all mags, before erasing?

daniel-wer (commented):
Awesome, thank you for digging deep and eradicating more and more volume annotation bugs! The tests are paying off and it's great to see that you're adding more.

The `pendingOperations` make a lot of sense and I'm much more confident that volume annotations with pending backend data download are working correctly now :)

From our last discussion, I remember that there was an error class where multiple annotations, before data from the server was loaded, interfered with each other and lead to errors. As far as I can see, this should be fixed with this PR as well. If it's not too complicated, I think this would be a great addition to the tests (ideally even combined with two undos, afterwards).

philippotto:
> From our last discussion, I remember that there was an error class where multiple annotations, before data from the server was loaded, interfered with each other and lead to errors. As far as I can see, this should be fixed with this PR as well. If it's not too complicated, I think this would be a great addition to the tests (ideally even combined with two undos, afterwards).

I'm not sure whether I understand the exact case you are referring to, but here's a test which has "interfering" annotations (the same position is brushed twice with two different cell ids and afterwards erased). See https://github.com/scalableminds/webknossos/pull/5955/files#diff-0a3ebf9c931222beceb5285c8805351bf2731be81a98a617772822d9c03fa497R572. Or do you mean another scenario?

frontend/javascripts/oxalis/model/bucket_data_handling/bucket.js
```diff
@@ -235,6 +236,16 @@ export class DataBucket {
     return this.state === BucketStateEnum.MISSING;
   }
 
+  isUnsynced(): boolean {
```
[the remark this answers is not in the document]
philippotto: good ideas and inspirations! I'll dwell on that a bit.

another idea: `isPartial` in the sense of "it's still missing something (i.e., the potentially existing data from the backend)"

frontend/javascripts/oxalis/model/sagas/save_saga.js
MichaelBuessemeyer: Nope, I don't know. 
According to the [docs](https://www.npmjs.com/package/lz4js) the compression and decompression always return an array and therefore not null.
I could only guess that this maybe is a flow thing because we compress / decompress using a worker with comlink 🤷 . Maybe the worker and comlink usage introduce the option that it might be null.

So I think as long as flow does not what a if-"catch" for a null we can leave it out.

daniel-wer:
> Or do you mean another scenario?

That's exactly it, nevermind then :see_no_evil: :+1:

frontend/javascripts/oxalis/model/sagas/save_saga.js
```diff
@@ -555,65 +579,67 @@ function* applyAndGetRevertingVolumeBatch(
     // Prepare a snapshot of the bucket's current data so that it can be
     // saved in an VolumeUndoState.
     let bucketData = null;
+    const currentPendingOperations = bucket.pendingOperations.slice();
     if (bucket.hasData()) {
       // The bucket's data is currently available.
       bucketData = bucket.getData();
+      // todo: clarify whether getData could already return the merged data
```
philippotto, line 586: todo

frontend/javascripts/test/sagas/volumetracing/volumetracing_saga_integration.spec.js
```diff
@@ -506,13 +536,130 @@ test.serial("Brushing/Tracing with undo (I)", async t => {
   Store.dispatch(addToLayerAction(paintCenter));
   Store.dispatch(finishEditingAction());
 
+  if (assertBeforeUndo) {
```
MichaelBuessemeyer, line 539: Could you please add a comment similar to line 528 and 533 as you did it in line 594?
> `// Brush with ${newCellId}`

This makes the test easier / faster to understand.

MichaelBuessemeyer (commented):
Thanks a lot for the fixes, deep investigations and new tests 🟢 . 
Props for finding the flaws in volume undo code 🎉  🕺 .

I only left a few thoughts, nothing big. Once the rest of the todos are done, I'll have another  look 👀

frontend/javascripts/oxalis/model/bucket_data_handling/bucket.js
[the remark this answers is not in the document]
philippotto: > If the backend is the onliest data source of "missing data" then I would prefer isMissingBackendData as it is the most straight forward imo. It does directly tell in what condition the bucket is in: It misses backend data.

Yes, this is true, but the caveat is that this can be confused with "isMissing" (i.e., the backend doesn't have any data for this bucket). This is why, I'm going for `needsBackendData` now. I hope this alright.

frontend/javascripts/oxalis/model/bucket_data_handling/bucket.js
```diff
@@ -340,17 +372,90 @@ export class DataBucket {
...
+          _data,
+          voxelMap,
+          cellId,
+          get3DAddress,
+          sliceCount,
+          thirdDimensionIndex,
+          shouldOverwrite,
+          overwritableValue,
+        ),
+      );
```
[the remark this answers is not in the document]
philippotto: This is a good idea. I just throttled the Internet connection and did some testing. Everything seemed to work and the total heap size in JS also stayed alright :+1: Luckily, each voxel map slice is only a few kB big (1/32 of a bucket). If one annotates through 32 slices while the backend data was not received, yet, the RAM consumption for buckets would double, but I think, this is very edge case-y (and still tolerable).

frontend/javascripts/oxalis/model/sagas/save_saga.js
philippotto: of course, this turned out to be another race condition. when the bucket data already arrived from the back-end, but wasn't yet compressed, this logic was incorrect. I turned the `backendData` property into a promise which is set immediately when the bucket was loaded. If the promise is set, then it's awaited in this saga. I also added a regression test for this which provokes the error case by introducing an artificial delay to the compression.

philippotto:
Whew, I think, this PR is ready for a final review now. I found two new bugs which I also covered with new tests:

- a race condition regarding the bucket compression (see https://github.com/scalableminds/webknossos/pull/5955#discussion_r791911918)
- I assumed that `bucket.dirty === true` would always mean that `bucket.pendingOperations.length > 0`. however, this is not the case, when one brushes an unloaded bucket and then undos this (afterwards, no pending operations need to be done, but the bucket is still dirty, since it still needs saving). this is a quite unfortunate scenario which should be fixed properly by refactoring, but I consider this to be out-of-scope for this PR. for this scenario, wk will send warnings to airbrake so we can see how we should prioritize this. I don't think, that this scenario would produce a huge havoc, but we should keep an eye on it, nevertheless.

@daniel-wer (and @MichaelBuessemeyer if you want), please have a final look at this PR. [Here's](https://github.com/scalableminds/webknossos/pull/5955/files/ea960d31f4f0cb4990009e56cc70b2e885de4b25..744c893f0213b0c189f259dff0b9b92e9bb03d26) the diff of what's changed since the last review (feel free to double-check the commit range).

A summary:
- fixed the two bugs I mentioned above
- converted "copy from previous slice" to use "labeled voxel maps"-only
- marked the old `label` method as deprecated and changed it to be async. also see the new docstrings. by awaiting the buckets before mutating them, no pending operations have to be maintained. I also searched through all existing user scripts and found that `labelVoxels` is not used by anyone.
- the floodfill didn't need big adjustments since it already awaited the buckets for the algorithm to work. however, I refactored the corresponding bucket methods a bit to make this clearer.
- ensured that the user cannot brush when wk is busy with undo/redo/floodfill
- further clean up & incorporated pr feedback

frontend/javascripts/oxalis/model/sagas/save_saga.js
```diff
@@ -615,27 +602,28 @@ function* applyAndGetRevertingVolumeBatch(
     // Decompress the bucket data which should be applied.
     let decompressedBucketData = null;
     let newPendingOperations = volumeUndoBucket.pendingOperations;
-    if (compressedBackendData != null) {
+
+    if (compressedBackendDataPromise != null) {
+      const compressedBackendData = yield* call(unpackPromise, compressedBackendDataPromise);
```
daniel-wer, line 607: This should simply work (unless you did the other way on purpose to satisfy flow or so):

```suggestion
      const compressedBackendData = yield compressedBackendDataPromise;
```

frontend/javascripts/oxalis/workers/slow_byte_array_lz4_compression.worker.js
```diff
@@ -0,0 +1,28 @@
...
+
+import lz4 from "lz4js";
+
+import { sleep } from "libs/utils";
+
+import { expose } from "./comlink_wrapper";
+
+let isSleepEnabled = false;
+
+export function enableSlowCompression(isEnabled: boolean) {
```
daniel-wer, line 14: Maybe `setSlowCompression` would be better since it can be enabled or disabled

frontend/javascripts/oxalis/workers/slow_byte_array_lz4_compression.worker.js
```diff
@@ -0,0 +1,28 @@
...
+  isSleepEnabled = isEnabled;
+}
+
+async function slowCompressLz4Block(data: Uint8Array, compress: boolean): Promise<Uint8Array> {
+  if (isSleepEnabled) {
+    await sleep(400);
+  }
+  if (compress) {
```
daniel-wer, line 22: Would it work to export `compressLz4Block` in the original file to import it here and use it directly? This would get rid of the duplication and safeguard against desynchronization of the two files. Or would you be worried that someone imports `compressLz4Block` and uses it directly in a production setting, accidentally then?

frontend/javascripts/test/model/binary/temporal_bucket_manager.spec.js
```diff
@@ -16,78 +16,96 @@ const TemporalBucketManager = mockRequire.reRequire(
...
+function prepareBuckets(manager, cube) {
   // Insert two buckets into manager
-  const bucket1 = new DataBucket("uint8", [0, 0, 0, 0], manager, mockedCube);
-  const bucket2 = new DataBucket("uint8", [1, 0, 0, 0], manager, mockedCube);
+  const bucket1 = new DataBucket("uint8", [0, 0, 0, 0], manager, cube);
+  const bucket2 = new DataBucket("uint8", [1, 0, 0, 0], manager, cube);
   for (const bucket of [bucket1, bucket2]) {
-    bucket.label(_.noop);
+    bucket.startDataMutation(_.noop);
+    bucket.endDataMutation(_.noop);
```
daniel-wer, line 100:
```suggestion
    bucket.startDataMutation();
    bucket.endDataMutation();
```

daniel-wer (commented):
Thanks for addressing the feedback and squashing even more bugs! Mostly LGTM, only some minor suggestions :+1:

frontend/javascripts/oxalis/model/sagas/save_saga.js
philippotto: Unfortunately, flow doesn't like this :cry: But I'll add a comment which explains this.

frontend/javascripts/oxalis/workers/slow_byte_array_lz4_compression.worker.js
philippotto: > Or would you be worried that someone imports compressLz4Block and uses it directly in a production setting, accidentally then?

Yes, that's what I was worried about. However, I now defined a `__compressLz4BlockHelper` function within the actual compression worker which is then imported. This reduces the duplication of the actual compress calls and the name (plus docstring) should make clear that it should not be imported somewhere else. I hope that's better now!

philippotto:
@daniel-wer Thanks for your feedback! I hope the PR is ready now from your point of view?

daniel-wer (approved):
Awesome, LGTM! I did not test this again. Should we do a round of monkey testing, what's your feeling?

PR #5958, opened 2022-01-13
Allow cancelling uploads

### LINKED ISSUE #5040: un-reserve dataset names after dataset is deleted
Related to dataset management

### LINKED ISSUE #5433: Un-reserve dataset names on failed conversion job
If my understnading is correct, uploading datasets that need conversion reserves the dataset name. It should be released when the job fails.

We need to take care that the paths on disk don’t collide despite that (e.g. add timestamp)

### LINKED ISSUE #5446: Add possibility to cancel dataset upload
A user noted that the dataset upload cannot currently be cancelled. 
Sometimes they would like to double-check that everything was set correctly.

youri-k:
https://cancelupload.webknossos.xyz/

#### upload cancel 
- upload a dataset
- press the `x` on the top right corner of the upload progress bar
- confirm the cancellation => should cancel the upload
- abort the cancellation => upload should finish normally
- after cancellation => next upload in same window should work normally
#### unreserve dataset names
- let a conversion job fail => name should still work
- delete an existing dataset => name should be valid again
- delete the new version of a dataset (i.e. same name) => should work and also free name  

### Issues:
- fixes #5446 
- fixes #5040 
- fixes #5433 

------
- [x] Ready for review


app/models/job/Job.scala
```diff
@@ -286,6 +287,18 @@ class JobService @Inject()(wkConf: WkConf,
     ()
   }
 
+  def cleanUpIfFailed(job: Job): Fox[Unit] =
+    if (job.state == JobState.FAILURE && job.command == "convert_to_wkw") {
```
youri-k, line 291: I could have created a more nested version that is easier to extend, but I remembered the XP practice of code for today, not tomorrow and this should not be complicated to change in the future if more jobs need such a functionality.

frontend/javascripts/dashboard/dataset/import_delete_component.js
```diff
@@ -55,6 +55,7 @@ const ImportDeleteComponent = ({ datasetId, history }: Props) => {
     setIsDeleting(false);
     // Trigger dataset check to make sure the dataset list is up-to-date
     // but also make sure that the toast can be read
+    // TODO it is no longer necessary to recheck the datasets since the delete call now removes the db from its cache
```
youri-k, line 58: Up until now, the datastore did not remove a deleted dataset from its cache directly after deleting, so it was necessary to check the disk. This is not necessary anymore and could be replaced by the less expensive `fetchDatasets` call. However, since we sleep for 2 seconds anyway, this would not improve the performance here. What are your opinions on this?

webknossos-datastore/app/com/scalableminds/webknossos/datastore/helpers/DataSetDeleter.scala
```diff
@@ -14,6 +15,22 @@ trait DataSetDeleter extends LazyLogging {
                    dataSetName: String,
                    isInConversion: Boolean = false,
                    reason: Option[String] = None)(implicit ec: ExecutionContext): Fox[Unit] = {
+    @tailrec
+    def deleter(sourcePath: Path, targetPath: Path, retryCount: Int = 0): Fox[Unit] =
```
youri-k, line 19: Since deleting a dataset now frees its name, we could delete two datasets with the same name. Therefore, we append a number to the end of the dataset names in case the name already exists in the trash folder.

youri-k (commented):
For now, you can only cancel the upload from within the upload view because we do not know about the `uploadId` outside of it. I created a follow-up issue to also cancel running uploads from the dashboard.

frontend/javascripts/dashboard/dataset/import_delete_component.js
philippotto: I'm not sure why the sleep exists. Does the `checkDatasets` call block until the backend is done with checking the disk? If yes, the sleep is unnecessary. If not, the sleep might be insufficient as the checking can easily take more than 2 seconds? As you suggest, I'd replace `checkDatasets` with `fetchDatasets` and then remove the sleep altogether.

frontend/javascripts/admin/dataset/dataset_upload_view.js
```diff
@@ -230,14 +239,18 @@ class DatasetUploadView extends React.Component<PropsWithFormAndRouter, State> {
         initialTeams: formValues.initialTeams.map(team => team.id),
       };
 
-      await reserveDatasetUpload(formValues.datastore.url, reserveUploadInformation);
+      const datastoreUrl = formValues.datastore.url;
```
philippotto, line 242: I changed this in another PR which should be merged soon. After pulling the master in, this needs to be turned into:

```suggestion
      const datastoreUrl = formValues.datastoreUrl;
```

frontend/javascripts/admin/dataset/dataset_upload_view.js
```diff
@@ -336,6 +349,34 @@ class DatasetUploadView extends React.Component<PropsWithFormAndRouter, State> {
     }
   };
 
+  cancelUpload = async () => {
+    const { uploadId, resumableUpload, datasetId, datastoreUrl } = this.state;
+    resumableUpload.pause();
+    const shouldCancel = await confirmAsync({
+      title:
+        "Cancelling the running upload will delete all already uploaded files on the server and cannot be undone. Are you certain to cancel the upload?",
```
philippotto, line 357:
```suggestion
        "Cancelling the running upload will delete all already uploaded files on the server and cannot be undone. Are you sure you want to cancel the upload?",
```

sounds more natural in my opinion

philippotto (commented):
Front-end code looks good :100: See my few comments. Let's wait for the back-end review now :)

app/controllers/WKRemoteDataStoreController.scala
```diff
@@ -145,7 +145,7 @@ class WKRemoteDataStoreController @Inject()(
     }
   }
 
-  def deleteErroneous(name: String, key: String): Action[JsValue] = Action.async(parse.json) { implicit request =>
+  def deleteDatasource(name: String, key: String): Action[JsValue] = Action.async(parse.json) { implicit request =>
```
fm3, line 148:
```suggestion
  def deleteDataset(name: String, key: String): Action[JsValue] = Action.async(parse.json) { implicit request =>
```

I’d go for dataset here as this is on the wk side and specifically operates on the datasets table

frontend/javascripts/admin/dataset/dataset_upload_view.js
```diff
@@ -346,6 +354,34 @@ class DatasetUploadView extends React.Component<PropsWithFormAndRouter, State> {
     }
   };
 
+  cancelUpload = async () => {
+    const { uploadId, resumableUpload, datasetId, datastoreUrl } = this.state;
+    resumableUpload.pause();
+    const shouldCancel = await confirmAsync({
+      title:
+        "Cancelling the running upload will delete already uploaded files on the server and cannot be undone. Are you sure you want to cancel the upload?",
+      okText: "Yes, Cancel the upload.",
```
fm3, line 363: I’d suggest overwriting cancelText as well (default is cancel, I believe, which is confusing here)
```suggestion
      okText: "Yes, Cancel the upload",
      cancelText: "No, keep it running",
```

webknossos-datastore/app/com/scalableminds/webknossos/datastore/controllers/DataSourceController.scala
```diff
@@ -463,14 +499,15 @@ Expects:
...
-                                        token) {
+      val dataSourceId = DataSourceId(dataSetName, organizationName)
+      accessTokenService.validateAccess(UserAccessRequest.deleteDataSource(dataSourceId), token) {
         AllowRemoteOrigin {
           for {
             _ <- binaryDataServiceHolder.binaryDataService.deleteOnDisk(organizationName,
                                                                         dataSetName,
                                                                         reason =
                                                                           Some("the user wants to delete the dataset"))
+            _ <- dataSourceRepository.cleanUpDataSource(dataSourceId) // also removes the name of the wK DB
```
fm3, line 510:
```suggestion
            _ <- dataSourceRepository.cleanUpDataSource(dataSourceId) // also frees the name in the wk-side database
```

webknossos-datastore/app/com/scalableminds/webknossos/datastore/helpers/DataSetDeleter.scala
fm3, line 19:
```suggestion
    def deleteWithRetry(sourcePath: Path, targetPath: Path, retryCount: Int = 0): Fox[Unit] =
```

webknossos-datastore/app/com/scalableminds/webknossos/datastore/helpers/DataSetDeleter.scala
```diff
@@ -14,6 +15,22 @@ trait DataSetDeleter extends LazyLogging {
...
+        }
+        logger.info(s"Successfully moved dataset from $sourcePath to $targetPath...")
+        Fox.successful(())
+      } catch {
+        case _: java.nio.file.FileAlreadyExistsException => deleter(sourcePath, targetPath, retryCount + 1)
+        case e: Exception                                => Fox.failure(s"Deleting dataset failed: ${e.toString}", Full(e))
+      }
+
     val dataSourcePath =
       if (isInConversion) dataBaseDir.resolve(organizationName).resolve(".forConversion").resolve(dataSetName)
```
fm3, line 35: I feel like the directory names ".forConversion" and ".trash" are already defined as constants somewhere? If so, it would be cool if they could be reused here.

webknossos-datastore/app/com/scalableminds/webknossos/datastore/services/UploadService.scala
```diff
@@ -353,6 +370,10 @@ class UploadService @Inject()(dataSourceRepository: DataSourceRepository,
     this.synchronized {
       PathUtils.deleteDirectoryRecursively(uploadDir)
     }
+    cleanUpRedis(uploadId)
```
fm3, line 373:
```suggestion
    removeFromRedis(uploadId)
```

webknossos-datastore/app/com/scalableminds/webknossos/datastore/storage/RedisTemporaryStore.scala
```diff
@@ -59,9 +59,9 @@ trait RedisTemporaryStore extends LazyLogging {
       r.exists(id)
     }
 
-  def remove(id: String): Fox[Unit] =
+  def remove(id: String): Fox[Boolean] =
```
fm3, line 62: Is this returned boolean used anywhere? I didn’t find a usage, and if not, I’d rather have it be Unit again

app/models/job/Job.scala
```diff
@@ -286,6 +287,18 @@ class JobService @Inject()(wkConf: WkConf,
...
+      logger.info(s"Job ${job._id} failed. Deleting Dataset...")
```
fm3, line 292:
```suggestion
      logger.info(s"WKW conversion job ${job._id} failed. Deleting dataset from the database, freeing the name...")
```

app/models/job/Job.scala
```diff
@@ -286,6 +287,18 @@ class JobService @Inject()(wkConf: WkConf,
...
+      val commandArgs = job.commandArgs.value
+      for {
+        datasetName <- commandArgs.get("dataset_name").map(_.as[String]).toFox
+        organizationName <- commandArgs.get("organization_name").map(_.as[String]).toFox
+        dataset <- dataSetDAO.findOneByNameAndOrganizationName(datasetName, organizationName)(GlobalAccessContext)
+        _ <- dataSetDAO.deleteDataset(dataset._id)
```
fm3, line 298: After a failed job, what exactly is removed?
 - dataset entry from postgres (should definitely be removed, I think that is what you do here already)
 - is anything deleted on the datastore side?
   - I think the target directory of the dataset (if it has already been created by the conversion job) should be deleted as well
   - I’d say the directory in `.converting` should be kept
   - I’d say the directory in `.forConversion` should also be kept
   - those two have the job id or some timestamp in their path anyway, right? :thinking: so that a retry with the same name will not cause problems

fm3 (commented):
This already looks very good, thanks for looking into this! I added a few suggestions and questions.

What is the progress on the second iteration with the new redis structure for upload metadata? 
Would you say it makes sense to merge this first and then work on finalizing the second iteration? Or merge both at once?

app/models/job/Job.scala
youri-k: - The dataset entry is removed from the progres.
- I think the target directory cannot exist if the job fails, because the directory in `.converting`  gets moved to the target directory after a successful conversion.
- The directory in `.converting`  is deduplicated by the request id.
- The directory in `.forConversion` is not deduplicated, so it would need to be deleted, or renamed. Do you have any preferences on this?

webknossos-datastore/app/com/scalableminds/webknossos/datastore/helpers/DataSetDeleter.scala
youri-k: They are not. Should I define it somewhere? If so, where?

youri-k:
> What is the progress on the second iteration with the new redis structure for upload metadata?
Would you say it makes sense to merge this first and then work on finalizing the second iteration? Or merge both at once?

For the existing master, I have implemented this already on the `save-upload-info` branch. I think the change would fit here well, so I'll merge these branches and merge it at once.

app/models/job/Job.scala
fm3: Thanks for the explanations! You are right, the target directory should not exist yet, so that’ll be fine.

I think it would be cool to deduplicate the contents of .forConversion somehow. Possibly with the upload id /timestamp? :thinking: will have to be passed to the worker job somehow, though. I am not sure how this works currently. If you can find an easy solution there, feel free to implement it, otherwise we can talk about this again :)

webknossos-datastore/app/com/scalableminds/webknossos/datastore/helpers/DataSetDeleter.scala
fm3: Yes, maybe a DirectoryConstants (not sure about the name) trait can be created that all relevant services/controllers/traits can extend so the constants become available

app/models/job/Job.scala
youri-k: I have implemented the deduplication step in the worker, which should do the trick :slightly_smiling_face:

philippotto:
@youri-k What's the status here? Can we re-review it since you merged the `save-upload-info` branch? Or was that branch reviewed somewhere else?

frontend/javascripts/admin/admin_rest_api.js
```diff
@@ -1218,6 +1209,18 @@ export function finishDatasetUpload(datastoreHost: string, uploadInformation: {}
   );
 }
 
+export function cancelDatasetUpload(
+  datastoreHost: string,
+  cancelUploadInformation: {},
```
philippotto, line 1214:
```suggestion
  cancelUploadInformation: {uploadId: string},
```

youri-k:
> Can we re-review it since you merged the save-upload-info branch?

Yes that would be great 🙂

webknossos-datastore/app/com/scalableminds/webknossos/datastore/services/UploadService.scala
```diff
@@ -138,29 +154,43 @@ class UploadService @Inject()(dataSourceRepository: DataSourceRepository,
...
+  def cancelUpload(cancelUploadInformation: CancelUploadInformation): Fox[Unit] = {
+    val uploadId = cancelUploadInformation.uploadId
+    for {
+      dataSourceId <- getDataSourceIdByUploadId(uploadId)
+      knownUpload <- isKnownUpload(uploadId)
+    } yield
+      if (knownUpload) {
+        logger.info(f"Canceling dataset upload of ${dataSourceId.team}/${dataSourceId.name} with id ${uploadId}...")
+        PathUtils.deleteDirectoryRecursively(uploadDirectory(dataSourceId.team, uploadId))
+        removeFromRedis(uploadId)
```
fm3, line 178: please make sure that removeFromRedis happens first, so that a new incoming chunk does not try to write to a deleted file

fm3 (commented):
Very nice, the handling of ids in redis is a huge improvement :)
Please see my two comments.
Also, could you add some error chain messages so that we are not in the dark if something fails? (I guess not every for needs one, but those that seem likely to fail)
I’ll test it as well now

frontend/javascripts/admin/dataset/dataset_upload_view.js
```diff
@@ -267,23 +273,20 @@ class DatasetUploadView extends React.Component<PropsWithFormAndRouter, State> {
 
         const uploadInfo = {
           uploadId,
-          organization: datasetId.owningOrganization,
-          name: datasetId.name,
-          layersToLink: [],
```
philippotto, line 272: When `layersToLink` is omitted, it simply defaults to `[]`, right?

philippotto (approved):
Front-end looks good :) Only see my one commit suggestion for improving the type coverage.

frontend/javascripts/admin/dataset/dataset_upload_view.js
youri-k: I removed the field here because the backend now saves the information from the `reserveUpload` request. The frontend always sends an empty list there https://github.com/scalableminds/webknossos/pull/5958/files/57f7c21ff54f1a25cd443404e2a796fa50681095#diff-2060cc6fbf0debf8ec6f63bfdd187937f7b1ded7d5c031522616362cd9f17149L250

PR #5961, opened 2022-01-16
Add auto scaling for color layer

### LINKED ISSUE #5887: Auto-tune min/max (brightness/contrast) to current view
## Detailed Description
There could be a button next to the histogram which adapts the min/max slider to the currently visible data.

Originally suggested here: https://forum.image.sc/t/feature-proposal-automatic-contrast-selection-in-dataset-viewer/59639

## Context

Dagobert42:
### Steps to test:
1. load dataset with color layer
2. use histogram clip button
3. activate histogram edit mode and 2.

### Issues:
- fixes #5887 

------
- [✓] Ready for review
Screenshot 2022-01-16 at 22 18 16


Dagobert42:
Also needs feedback on whether the adjustment should be more/less strong as well as testing with different datasets.

Dagobert42:
Seems to take considerable time before the button functions properly upon (re)loading the dataset. Any ideas on what might be the cause of that?

daniel-wer:
@Dagobert42 
@philippotto asked me to extend the `getDataFor2DBoundingBox` to be able to return data given a specific zoomstep to avoid having to possibly load a lot of data for the histogram auto-clipping. I implemented this, modified the clipping code to supply the current zoom step when obtaining the data cuboids, and tested it as well as I could. To me, it looks like it works as expected, no new buckets are requested from the server when using the "Clip" button (since all visible data is already loaded in the current zoom step), while the clipping still seems to work as well. Please have a look whether I broke something unintentionally :see_no_evil: 
Let me know if you have any questions or feedback :)

philippotto:
@daniel-wer Awesome, thank you for adding the zoomstep support :pray:

frontend/javascripts/oxalis/view/left-border-tabs/histogram_view.js
```diff
@@ -185,6 +193,80 @@ class Histogram extends React.PureComponent<HistogramProps, HistogramState> {
...
+  };
+
+  clipHistogram = async (isInEditMode: boolean, layerName: string) => {
+    const [lowClip, highClip] = await this.getClippingValues(layerName);
+    if (!isInEditMode) {
+      this.onThresholdChange([lowClip, highClip]);
+    } else {
+      this.onThresholdChange([lowClip, highClip]);
+      this.setState({ currentMin: lowClip, currentMax: highClip });
+      this.updateMinimumDebounced(lowClip, layerName);
```
philippotto, line 265: The debounced version is meant to collapse multiple function calls if they are shortly fired after each other (useful for mouse events which can be triggered frequently in a short time). Since it always introduces an artificial delay (in this case 500 ms), I'd use `this.props.onChangeLayer(layerName, "min", value)` instead (analoguous for the maximum).

frontend/javascripts/oxalis/view/left-border-tabs/histogram_view.js
```diff
@@ -40,7 +48,7 @@ type HistogramState = {
 
 const uint24Colors = [[255, 65, 54], [46, 204, 64], [24, 144, 255]];
 const canvasHeight = 100;
-const canvasWidth = 300;
+const canvasWidth = 318;
```
philippotto, line 51: Can you explain the reason for this change?

frontend/javascripts/oxalis/view/left-border-tabs/histogram_view.js
```diff
@@ -185,6 +193,80 @@ class Histogram extends React.PureComponent<HistogramProps, HistogramState> {
     500,
   );
 
+  getCuboidForViewport = async (viewport: OrthoView, layerName: string) => {
```
philippotto, line 196: This method should probably moved into api_latest into the "Data" class. I'd name it `getDataForViewport`.

frontend/javascripts/oxalis/view/left-border-tabs/histogram_view.js
```diff
@@ -185,6 +193,80 @@ class Histogram extends React.PureComponent<HistogramProps, HistogramState> {
...
+
+    const resolutionIndex = getRequestLogZoomStep(state);
+    const cuboid = await api.data.getDataFor2DBoundingBox(layerName, { min, max }, resolutionIndex);
+    return cuboid;
+  };
+
+  getClippingValues = async (layerName: string, threshold: number = 0.05) => {
+    const cuboidXY = await this.getCuboidForViewport(OrthoViews.PLANE_XY, layerName);
+    const cuboidXZ = await this.getCuboidForViewport(OrthoViews.PLANE_XZ, layerName);
+    const cuboidYZ = await this.getCuboidForViewport(OrthoViews.PLANE_YZ, layerName);
```
philippotto, line 225: Promises which don't depend on each other can be parallelized like this:

```suggestion
    const [cuboidXY, cuboidXZ, cuboidYZ] = Promise.all([
      this.getCuboidForViewport(OrthoViews.PLANE_XY, layerName),
      this.getCuboidForViewport(OrthoViews.PLANE_XZ, layerName),
      this.getCuboidForViewport(OrthoViews.PLANE_YZ, layerName),
    ]);
```

frontend/javascripts/oxalis/view/left-border-tabs/histogram_view.js
```diff
@@ -185,6 +193,80 @@ class Histogram extends React.PureComponent<HistogramProps, HistogramState> {
...
+
+    const { elementClass } = getLayerByName(Store.getState().dataset, layerName);
+    const [TypedArrayClass] = getConstructorForElementClass(elementClass);
+    const cuboidData = new TypedArrayClass(cuboidXY.length + cuboidXZ.length + cuboidYZ.length);
```
philippotto, line 229: `dataForAllViewports` would be a clearer name than `cuboidData` in my opinion.

frontend/javascripts/oxalis/view/left-border-tabs/histogram_view.js
```diff
@@ -185,6 +193,80 @@ class Histogram extends React.PureComponent<HistogramProps, HistogramState> {
...
+    cuboidData.set(cuboidXZ, cuboidXY.length);
+    cuboidData.set(cuboidYZ, cuboidXY.length + cuboidXZ.length);
+
+    const valueCounts = {};
+    let maxCount = 0;
+    for (let i = 0; i < cuboidData.length; i++) {
+      if (cuboidData[i] !== 0 && cuboidData[i] in valueCounts) {
+        valueCounts[cuboidData[i]] += 1;
+        if (maxCount < valueCounts[cuboidData[i]]) maxCount = valueCounts[cuboidData[i]];
+      } else valueCounts[cuboidData[i]] = 1;
```
philippotto, line 241: We typically avoid brace-less if/else bodies. Therefore, I'd recommend to change the code to:
```suggestion
        if (maxCount < valueCounts[cuboidData[i]]) {
          maxCount = valueCounts[cuboidData[i]];
        }
      } else {
        valueCounts[cuboidData[i]] = 1;
      }
```

frontend/javascripts/oxalis/view/left-border-tabs/histogram_view.js
```diff
@@ -185,6 +193,80 @@ class Histogram extends React.PureComponent<HistogramProps, HistogramState> {
...
+    }
+
+    let lowClip = 255;
```
philippotto, line 244: There are multiple data types which a layer can have. 255 is only valid for uint8. For uint16 you would need 2**16 - 1. However, I think, it's easier to initialize both values (`lowClip` and `highClip`) with `-1` and check explicitly against `-1` (i.e., `cuboidData[i] !== 0 && (cuboidData[i] < lowClip || lowClip === -1)`.

If the entire cube data is `0` then, low and max would still be -1. In that case, you wouldn't have nice default values (which is the advantage of your current approach), but I would argue that the histogram should not be changed at all in that case, since no useful data exists at this moment. You could simply `return null` in that case à  la:

```
// the following would replace the old "if (lowClip > highClip) return [highClip, lowClip];" line probably
if (lowClip === -1 || highClip === -1) {
  return null;
}
```

The caller would need to check whether the returned value is null and then simply do nothing (or even better: show a toast message that the clipping didn't succeed).

frontend/javascripts/oxalis/view/left-border-tabs/histogram_view.js
```diff
@@ -185,6 +193,80 @@ class Histogram extends React.PureComponent<HistogramProps, HistogramState> {
...
+    let highClip = 1;
+    const threshValue = (threshold * maxCount).toFixed();
+    for (let i = 0; i < cuboidData.length; i++) {
+      if (valueCounts[cuboidData[i]] > threshValue) {
+        if (cuboidData[i] !== 0 && cuboidData[i] < lowClip) lowClip = cuboidData[i];
```
philippotto, line 249: same here with the if braces

frontend/javascripts/oxalis/view/left-border-tabs/histogram_view.js
```diff
@@ -209,6 +291,22 @@ class Histogram extends React.PureComponent<HistogramProps, HistogramState> {
           width={canvasWidth}
           height={canvasHeight}
         />
+        <Tooltip title="Clip the histogram to enhance contrast. In Edit Mode this also adjusts the histogram's range.">
+          <Button
+            id="mine"
```
philippotto, line 296: id can probably be removed

philippotto (changes requested):
Very cool stuff, @Dagobert42! The feature works well for me :tada: However, I noticed that it doesn't work for non-uint8 data (e.g., I tested it on multi channel uint24 data here: https://addautoscalingforcolorlayer.webknossos.xyz/datasets/sample_organization/Multi-Channel-Test/view#1219,468,7,0,0.621).

I left some feedback which should also help to fix the non-uint8 support. We can talk about it tomorrow :)

frontend/javascripts/oxalis/view/left-border-tabs/histogram_view.js
Dagobert42: I just noticed that the right edges weren't aligned at 300 pixels (which felt like an arbitrary number).
Screenshot 2022-01-18 at 20 39 01

frontend/javascripts/oxalis/view/left-border-tabs/histogram_view.js
Dagobert42: I get : an index signature declaring the expected key / value type is missing in  `Promise` when I try to implement this.

frontend/javascripts/oxalis/view/left-border-tabs/histogram_view.js
Dagobert42: woops 😅 thanks for the good eyes

frontend/javascripts/oxalis/view/left-border-tabs/histogram_view.js
Dagobert42: Could there also be a problem with uint24 not being part of `TypedArrayClass`?
```
type t =
  | typeof Uint8Array
  | typeof Uint16Array
  | typeof Uint32Array
  | typeof Float32Array
```

Dagobert42: Screenshot 2022-01-18 at 21 31 24

frontend/javascripts/oxalis/view/left-border-tabs/histogram_view.js
philippotto: Oh, sorry! :see_no_evil: Glad you got it working :)

frontend/javascripts/oxalis/view/left-border-tabs/histogram_view.js
philippotto: > Could there also be a problem with uint24 not being part of `TypedArrayClass`?
> 
> ```
> type t =
>   | typeof Uint8Array
>   | typeof Uint16Array
>   | typeof Uint32Array
>   | typeof Float32Array
> ```

This should still work, since `uint24` is basically only an alias for "uint8 with 3 channels". The `getConstructorForElementClass` method which you are using will return `uint8` and `3` when the element class is "uint24". The channel count could be used to calculate the total length of the flattened array, but since you are using the `length` property of the other arrays, you don't respect the channel count, as far as I see.



> Screenshot 2022-01-18 at 21 31 24

Cool toast!

frontend/javascripts/oxalis/view/left-border-tabs/histogram_view.js
philippotto: Ah, cool, thanks for fixing this :) :100:

frontend/javascripts/oxalis/view/left-border-tabs/histogram_view.js
```diff
@@ -185,6 +194,98 @@ class Histogram extends React.PureComponent<HistogramProps, HistogramState> {
...
+      }
+    }
+    return [lowClip, highClip];
+  };
+
+  clipHistogram = async (isInEditMode: boolean, layerName: string) => {
+    const [lowClip, highClip] = await this.getClippingValues(layerName);
+    if (lowClip === -1 || highClip === -1) {
+      Toast.warning(
+        "Clipping the histogram failed. The data did not contain any brightness values greater than 0.",
```
philippotto, line 275: I'd try to avoid negative words, such as "failed" in the UI, when possible. My suggestion:

```suggestion
        "The histogram could not be clipped, because the data did not contain any brightness values greater than 0.",
```

frontend/javascripts/oxalis/view/left-border-tabs/histogram_view.js
```diff
@@ -185,6 +194,98 @@ class Histogram extends React.PureComponent<HistogramProps, HistogramState> {
...
+          maxCount = valueCounts[dataForAllViewports[i]];
+        }
+      } else {
+        valueCounts[dataForAllViewports[i]] = 1;
+      }
+    }
+
+    let lowClip = -1;
+    let highClip = -1;
+    const threshValue = (threshold * maxCount).toFixed();
```
philippotto, line 255: I think, this threshold is a good starting point, but I noticed that it doesn't work too well, when there are large areas with a single intensity. Maybe this is contrived, but the one RGB DS I tested it with, had a large blue sky which dominated the histogram, even when the sky only covered half of the picture. Then, the `maxCount` is very high (in comparison to other other frequencies which are spread over the remaining 254 values) and practically defines the entire clipping range.
I researched a bit and found a very similar approach, which uses the cumulative sum which sounds like it could help to soften this problem scenario very well.

Have a look at this [SO answer](https://stackoverflow.com/a/56909036/896760). The relevant code is this:

```python
    # Calculate cumulative distribution from the histogram
    accumulator = []
    accumulator.append(float(hist[0]))
    for index in range(1, hist_size):
        accumulator.append(accumulator[index -1] + float(hist[index]))

    # Locate points to clip
    maximum = accumulator[-1]
    clip_hist_percent *= (maximum/100.0)
    clip_hist_percent /= 2.0

    # Locate left cut
    minimum_gray = 0
    while accumulator[minimum_gray] < clip_hist_percent:
        minimum_gray += 1

    # Locate right cut
    maximum_gray = hist_size -1
    while accumulator[maximum_gray] >= (maximum - clip_hist_percent):
        maximum_gray -= 1
```

It basically clips a certain percentage of the entire image's intensity starting from left and right in the histogram using the cumulative/accumulated sum. I think, this is closer to what we need. What do you think? Happy to chat about this today :)

Dagobert42: I implemented this to the best of my knowledge. @philippotto could you please make sure I'm iterating the accumulator and localHist objects correctly. I'm also open for suggestions on how to do it more efficiently :)

frontend/javascripts/oxalis/view/left-border-tabs/histogram_view.js
philippotto: I think, you forgot to apply my suggestion or didn't push it yet? :see_no_evil:

frontend/javascripts/oxalis/view/left-border-tabs/histogram_view.js
```diff
@@ -199,6 +278,9 @@ class Histogram extends React.PureComponent<HistogramProps, HistogramState> {
       `Enter the ${minimumOrMaximum} possible value for layer ${layerName}. Scientific (e.g. 9e+10) notation is supported.`;
 
     const minMaxInputStyle = { width: "100%" };
+    const tooltipText = isInEditMode
+      ? "Clipping in Edit Mode will also adjust the histogram's range."
+      : "Automatically clip the histogram to enhance contrast.";
```
philippotto, line 283: When being in edit mode, the tooltip doesn't explain the high-level feature. How about:

```suggestion
    const editModeAddendum = isInEditMode ?  "In Edit Mode, the histogram's range will be adjusted, too." : "";
    const tooltipText = `Automatically clip the histogram to enhance contrast.${editModeAddendum}`;
```

frontend/javascripts/oxalis/view/left-border-tabs/histogram_view.js
```diff
@@ -209,6 +291,18 @@ class Histogram extends React.PureComponent<HistogramProps, HistogramState> {
           width={canvasWidth}
           height={canvasHeight}
         />
+        <Tooltip title={tooltipText}>
+          <VerticalAlignMiddleOutlined
+            style={{
+              position: "absolute",
+              transform: "rotate(90deg)",
+              top: 14,
+              right: 82,
```
philippotto, line 300: It's a bit unusual to position the icon that way, but I see, that there's not really a super straightforward way to do it differently, since the button would need to be defined higher in the component hierarchy while still interacting with the histogram. This is a symptom of our UI being not super consistent (the button is only specific to the histogram view, but exists outside of it). Tldr: Let's keep it as is, but we should probably change it at some point in the future :tm: :)

frontend/javascripts/oxalis/view/left-border-tabs/histogram_view.js
```diff
@@ -185,6 +191,79 @@ class Histogram extends React.PureComponent<HistogramProps, HistogramState> {
     500,
   );
 
+  getClippingValues = async (layerName: string, threshold: number = 0.05) => {
```
philippotto, line 194: I'd rename `threshold` to `percentageThreshold`, as the later `thresholdValue` is confusing, otherwise.

frontend/javascripts/oxalis/view/left-border-tabs/histogram_view.js
```diff
@@ -185,6 +191,79 @@ class Histogram extends React.PureComponent<HistogramProps, HistogramState> {
...
+    ]);
+    const dataForAllViewports = new TypedArrayClass(
+      cuboidXY.length + cuboidXZ.length + cuboidYZ.length,
+    );
+
+    dataForAllViewports.set(cuboidXY);
+    dataForAllViewports.set(cuboidXZ, cuboidXY.length);
+    dataForAllViewports.set(cuboidYZ, cuboidXY.length + cuboidXZ.length);
+
+    const localHist = {};
```
philippotto, line 211: Using a mapping structure here (and for `accumulator`) is a good idea to avoid large arrays. However, JS objects have the annoying property that numeric keys are automatically converted to strings. This is why you needed the `Number` casting below in the code. This can lead to subtle bugs which is why I'd recommend to use `new Map()` here. The interface looks like this:

```
const localHist = new Map();
localHist.set(3, 10);
localHist.get(3) == 10;
```

Then, you can also remove the `Number(...)` conversions below.

frontend/javascripts/oxalis/view/left-border-tabs/histogram_view.js
```diff
@@ -185,6 +191,79 @@ class Histogram extends React.PureComponent<HistogramProps, HistogramState> {
...
+        localHist[dataForAllViewports[i]] += 1;
+      } else {
+        localHist[dataForAllViewports[i]] = 1;
+      }
+    }
+
+    const accumulator = {};
+    let area = 0;
+    let lastKey = 0;
+    for (const key of Object.keys(localHist)) {
```
philippotto, line 223: For the accumulation it's important to iterate in the correct order over the data structure. Enumeration order is a complex topic in JS, unfortunately. I think, in your use case, it's working correctly with modern browsers, but I'd still want to keep it safe by doing a simple sort beforehand. Something like this:

```
// assuming that localHist is a Map()
const sortedHistKeys = Array.from(localHist.keys()).sort((a, b) => a - b)
for (const key of sortedHistKeys) {
...
```

frontend/javascripts/oxalis/view/left-border-tabs/histogram_view.js
```diff
@@ -185,6 +191,79 @@ class Histogram extends React.PureComponent<HistogramProps, HistogramState> {
...
+      accumulator[key] = localHist[key] + area;
+      area += localHist[key];
```
philippotto, line 225: this should be equivalent and is easier to read in my opinion:

```suggestion
      area += localHist[key];
      accumulator[key] = area;
```

frontend/javascripts/oxalis/view/left-border-tabs/histogram_view.js
```diff
@@ -185,6 +191,79 @@ class Histogram extends React.PureComponent<HistogramProps, HistogramState> {
...
+      lastKey = key;
+    }
+    const maximum = accumulator[lastKey];
+    const thresholdValue = (threshold * maximum) / 2.0;
+
+    let lowClip = -1;
+    for (const key of Object.keys(accumulator)) {
```
philippotto, line 232: the iteration method should be similar to my other comment when switching to Map().

frontend/javascripts/oxalis/api/api_latest.js
```diff
@@ -1219,20 +1227,70 @@ class DataApi {
...
+    }
     const buckets = await Promise.all(
       bucketAddresses.map(addr => this.getLoadedBucket(layerName, addr)),
     );
     const { elementClass } = getLayerByName(Store.getState().dataset, layerName);
-    return this.cutOutCuboid(buckets, bbox, elementClass);
+    return this.cutOutCuboid(buckets, bbox, elementClass, resolutions, zoomStep);
+  }
+
+  async getDataForViewport(viewport: OrthoView, layerName: string) {
```
philippotto, line 1259: I think, naming-wise `getViewportData` would be a bit better :) Sorry, I think, the name was even my suggestion, but it just crossed my mind..

philippotto (changes requested):
Really, really awesome stuff :100: It's a lot of fun to use this feature and it works quite well for the datasets I tested it with. This will be a great addition!

The UI looks good & the code feels fast :+1: I only left a couple of comments regarding the implementation, since you ran into some "specialties" of JavaScript which I'd ideally avoid. I think, the implementation itself is correct, but can be a bit modernized with newer JS features. See my few comments and let me know in case you have questions :)

frontend/javascripts/oxalis/view/left-border-tabs/histogram_view.js
Dagobert42: Indeed 😅 I pushed it now!

frontend/javascripts/oxalis/view/left-border-tabs/histogram_view.js
Dagobert42: I ended up calling the handed-in variable `thresholdRatio`

frontend/javascripts/oxalis/view/left-border-tabs/histogram_view.js
Dagobert42: Awesome! Now I can properly expect <number, number> when iterating over the map. I think the code looks quite a bit better now 👍 Thanks!

frontend/javascripts/oxalis/view/left-border-tabs/histogram_view.js
Dagobert42: I also realized I can use this instead of maximum right away 🧐

philippotto (approved):
Awesome :) Great work!

PR #5967, opened 2022-01-18
Store layer visibility with annotation

### LINKED ISSUE #5910: Store visibility of layers per annotation (instead of per dataset)
Currently, we have to remove existing segmentation layers when creating a volume annotation without fallback data. The main problem with that is that the visibility of the segmentation layer is stored per dataset. This means, if I create an annotation without fallback for a dataset, the "old" segmentation layer is disabled automatically and opening the dataset in view mode will be affected by that. Then, the segmentation layer will be invisible there, which is most likely not what the user wants nor expects.

fm3:
- https://layervisibilityinannotation.webknossos.xyz

### Steps to test:
- Create an annotation with multiple annotation layers
- toggle layer visibilities
- create second account, open the same annotation
- should show the layers the first user selected
- toggling by the second user should not save into the annotation

### Issues:
- fixes #5910

------
- [x] Ready for review


philippotto:
@fm3 I implemented the front-end side for this PR. Do you think it's ready now? Regarding the "keep old segmentation layer" topic, I'd suggest to tackle it in separate PR.

fm3:
@philippotto Thanks for looking into this! I just tried to test the functionality, but when I toggled layers in an annotation, I did not (in the browser’s network tab) see any requests to the annotation update route. When opening the annotation as a second user, I still see the layers that I disabled as the annotation owner. Maybe something is still missing here?

philippotto:
> @philippotto Thanks for looking into this! I just tried to test the functionality, but when I toggled layers in an annotation, I did not (in the browser’s network tab) see any requests to the annotation update route. When opening the annotation as a second user, I still see the layers that I disabled as the annotation owner. Maybe something is still missing here?

My bad! I introduced a bug when I tried to ignore read-only tracings. Now, it should work :)

fm3:
Yes, works now :) (except for the Skeleton layer but I assume that is intentional?)

fm3:
@jstriebel the annotation json now contains a new field `viewConfiguration` and the annotation edit request can now also set that field. Do you know if we have to adapt anything in the python client for this in order to not break things? I assume the edit request was not yet used by the libs?

philippotto:
> Yes, works now :) (except for the Skeleton layer but I assume that is intentional?)

Great! Yes, it's intention or rather known. The skeleton visibility is not saved at all, right now. I'd keep it that way for now. Do you agree?

fm3:
by the way: how does this interact with name changes of annotation layers? is the configuration updated then to match that? :)

philippotto:
> by the way: how does this interact with name changes of annotation layers? is the configuration updated then to match that? :)

The new settings are keyed by layer name for dataset layers and by ID for volume annotation layers. That way, this is robust against changes to the volume annotation layer names. if the dataset layers are renamed on disk, the settings won't survive, but I think, this is to be expected.

fm3:
Sounds good, thanks!

jstriebel:
> @jstriebel the annotation json now contains a new field viewConfiguration and the annotation edit request can now also set that field. Do you know if we have to adapt anything in the python client for this in order to not break things?

I think new fields in responses are not a problem so far. However, I guess the `viewConfiguration` has varying entries, right? Then we should add it [here](https://github.com/scalableminds/webknossos-libs/blob/master/webknossos/__generate_client.py#L124), so that the client will not assume specific content when we re-generate it.

> I assume the edit request was not yet used by the libs?

Yes, it wasn't.

daniel-wer (approved):
Code LGTM.

Testing mostly worked well, but there is one thing that could be improved imo (no blocker). When creating an annotation and then sharing it without toggling any layers, the view configuration is never persisted to the annotation. Hence, the second user might see different layers (if the first user adapted the dataset's layers at any time before, for example, in view mode).
One possible solution could be to initialize the annotation view configuration with the user's dataset view configuration upon annotation creation. Not sure if that'd be easy for the backend.

fm3:
The backend does not currently have any understanding of the content of this JSON field. I assume the format is not guaranteed to be identical to the one of the dataset view configuration? If not, it would be a nontrivial change. If you say it has no high priority, I’d skip it for the moment, unless this problem crops up again during normal use

youri-k (approved):
Backend LGTM :+1:

PR #5974, opened 2022-01-20
Clean up frontend deprecations and warnings

daniel-wer:
Lots of react deprecation warnings and html structure warnings were clogging up the console, making it hard to see relevant information.
Most of the `componentWillMount`, `componentWillUpdate`, and `componentWillReceiveProps` occurrences could be replaced with the `didMount`/`didUpdate` alternatives fairly easily. However, for some of them I didn't see a good way.
The `// eslint-disable-next-line react/no-did-update-set-state` are a little annoying since having the `setState` behind an if-comparison-clause seems to be the officially recommended way :shrug: 

### URL of deployed dev instance (used for testing):
- https://cleanupdeprecationsandwarnings.webknossos.xyz

### Steps to test:
- Test that the persistence still works by entering a search term in one of the admin views and then reloading the page
- Test the other components that were changed non-trivially: permissions_and_teams_modal_view, team_selection_component, vector_input, merge_modal_view, editable_text_label, histogram_view
- While clicking around (almost) no deprecation warnings should turn up in the console (except for Remarkable which already is the newest version, but doesn't seem to be maintained any longer :/)

### Issues:
- fixes clogged up console

------
(Please delete unneeded items, merge only when none are left open)
- [x] Ready for review

PR #5976, opened 2022-01-20
Avoid scroll bars caused by context menu positioning

### LINKED ISSUE #5383: Position the node context menu dependent on the position of the page
## Context

Currently, the node context menu is always on the right side of the clicked position and grows to the bottom. This leads to a "content" overflow if the mouse is at the bottom or at the right border of the page. These cases lead to scroll bars and that not all menu options are visible all the time.

## Expected Behavior

The node context menu should be positioned according to its position on the page just like the browser build in context menu. If the clicked position is near a border (top, left, right, bottom) the context menu should stretch into the opposite direction to not lead to an overflow and to make all options visible at a time.

## Current Behavior

Just as described above. The node context menu is always on the right side of the clicked position and grows downwards.

## Steps to Reproduce the bug

 - Open the context menu near the right border. This should lead to an overflow.  (Needs Tab rearrangement)
 - Open the context menu near the bottom border. This should lead to an overflow.

daniel-wer:
CSS for the win.

Review with "Hide Whitespace" for minimum code changes and maximum pleasure :)

- https://___.webknossos.xyz

### Steps to test:
- Test using firefox, chrome
- Open the context menu next to the right or bottom edge. The menu should no longer stick out of the viewport and shouldn't cause any scrollbars to appear.
- The menu items and copy buttons should still work.

### Issues:
- fixes #5383 

------
- [x] Ready for review

### DIFF
--- a/frontend/javascripts/oxalis/view/context_menu.js
+++ b/frontend/javascripts/oxalis/view/context_menu.js
@@ -733,13 +733,19 @@ function ContextMenu(props: Props) {
           evt.preventDefault();
           hideContextMenu();
         }}
-      >
+      />
+      {/* This div serves as a "prison" for the sticky context menu. The above div
+      cannot be used since the context menu would then be closed on every click.
+      Since both divs are absolutely positioned and cover the whole page,
+      avoid blocking click events by temporarily disabling them for this "prison" div. */}
+      <div className="node-context-menu-overlay" style={{ pointerEvents: "none" }}>
         <div
           style={{
             position: "sticky",
             left: contextMenuPosition[0],
             top: contextMenuPosition[1],
             width: "fit-content",
+            pointerEvents: "all",
           }}
           className="node-context-menu"
           tabIndex={-1}
--- a/frontend/javascripts/oxalis/view/context_menu.js
+++ b/frontend/javascripts/oxalis/view/context_menu.js
@@ -733,24 +733,26 @@ function ContextMenu(props: Props) {
           evt.preventDefault();
           hideContextMenu();
         }}
-      />
-      <div
-        style={{
-          position: "absolute",
-          left: contextMenuPosition[0],
-          top: contextMenuPosition[1],
-        }}
-        className="node-context-menu"
-        tabIndex={-1}
-        ref={inputRef}
       >
-        <Shortcut supportInputElements keys="escape" onTrigger={hideContextMenu} />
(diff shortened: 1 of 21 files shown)


frontend/javascripts/oxalis/view/context_menu.js
```diff
@@ -734,23 +734,31 @@ function ContextMenu(props: Props) {
...
-        <Divider className="hide-if-first hide-if-last" style={{ margin: "4px 0px" }} />
-        {infoRows}
+      {/* This div serves as a "prison" for the sticky context menu. The above div
+      cannot be used since the context menu would then be closed on every click.
+      Since both divs are absolutely positioned and cover the whole page,
+      avoid blocking click events by temporarily disabling them for this "prison" div. */}
+      <div className="node-context-menu-overlay" style={{ pointerEvents: "none" }}>
+        <div
+          style={{
+            position: "sticky",
```
MichaelBuessemeyer, line 744: Using Sticky is a very very nice idea 🎉  🚀

MichaelBuessemeyer (approved):
Nice fix 🎉 , the context menu seems to still work for me (MacOS + Chrome).

During testing I came across two other 🐛 s 😅 . But they are only minor ones.
1. When I open the context menu with the move tool active in a hybrid tracing I get a `Warning: Each child in a list should have a unique "key" prop.`error in the console.
2. It looks like the titles of the viewports have a higher z value than the context menu. This looks a bit weird and funny. It should be easy to fix.
image

As the two bugs mentioned above are not part of the issue behind this pr, feel free to merge :)
But if you want to fix one or both of them in this PR feel free.

frontend/javascripts/oxalis/view/context_menu.js
philippotto: :exploding_head: :+1: Didn't know that `sticky` can be used for this scenario. Grand stuff!

daniel-wer:
@MichaelBuessemeyer Thanks for the review and the thorough testing!

> When I open the context menu with the move tool active in a hybrid tracing I get a Warning: Each child in a list should have a unique "key" prop.error in the console.

I already fixed that in https://github.com/scalableminds/webknossos/pull/5974 :)

> It looks like the titles of the viewports have a higher z value than the context menu. This looks a bit weird and funny. It should be easy to fix.

I'll have a quick look.

PR #5982, opened 2022-01-24
Honor mappingStatus when requesting to compute meshes ad-hoc

### LINKED ISSUE #5966: Ad-hoc mesh computation doesn't honor mapping disabled status
## Context

I tested computing ad-hoc meshes with enabled and disabled mappings.

## Expected Behavior

After enabling and disabling a mapping, the ad-hoc mesh is still computed for the mapping, although it is disabled.

## Current Behavior

The ad-hoc mesh should be computed for the segment without a mapping.

## Steps to Reproduce the bug

1. Open a dataset with a segmentation layer and mappings.
2. Enable a mapping.
3. Disable the mapping again.
4. Select "Compute Mesh (ad-hoc)" from the context menu -> The mesh is computed as if the mapping was still enabled.

## Your Environment for bug

- Version of WebKnossos (Release or Commit): master

daniel-wer:
The `mappingName` and `mappingType` remain in the store if a mapping is disabled. Therefore, the `mappingStatus` needs to be taken into account to check whether a mapping is actually enabled.

- https://___.webknossos.xyz

### Steps to test:
- Open a dataset with a segmentation layer and a mapping
- Compute an ad-hoc mesh for a segment (using the context-menu or the "Segments" list - segments appear in the segment list after clicking on them)
- Enable a mapping and compute an ad-hoc mesh for the mapped segment
- Disable the mapping again and delete both meshes in the "Segments" list
- Compute an ad-hoc mesh for the first segment -> The result should be the same as in the beginning

### Issues:
- fixes #5966 

------
- [x] Ready for review


Dagobert42:
The fix works for me 👍 (only thing left seems to be fixing the CHANGELOG conflict)

Dagobert42 (approved):
The fix works for me 👍 only thing left seems to be fixing the CHANGELOG conflict :)

daniel-wer:
Done, I've also added a change log entry for https://github.com/scalableminds/webknossos/pull/5961 while I was at it since it's a great new feature and users should know about it :)

Dagobert42:
Awesome thanks! :)

PR #5983, opened 2022-01-24
Fix placeholder resolutions for anisotropic layers with missing base resolutions

### LINKED ISSUE #5972: Incorrect placeholder magnifications are calculated in certain scenarios
If there are multiple layers which only have resolutions poorer than mag 8, webKnossos calculates placeholder magnifications which can be incorrect.

For example, consider the following datasource-properties.json:

```
{
  "scale": [
    10,
    10,
    50
  ],
  "id": {
    "name": "...",
    "team": "..."
  },
  "dataLayers": [
    {
      "name": "segmentation",
      "boundingBox": {
        "topLeft": [
          320,
          320,
          80
        ],
        "width": 42368,
        "height": 34160,
        "depth": 2392
      },
      "wkwResolutions": [
        {
          "resolution": [
            16,
            16,
            4
          ],
          "cubeLength": 256
        }
      ],
      "elementClass": "uint32",
      "largestSegmentId": 1823,
      "category": "segmentation",
      "dataFormat": "wkw"
    },
    {
      "name": "color",
      "category": "color",
      "boundingBox": {
        "topLeft": [
          0,
          0,
          0
        ],
        "width": 43008,
        "height": 34810,
        "depth": 2555
      },
      "wkwResolutions": [
        {
          "resolution": [
            16,
            16,
            4
          ],
          "cubeLength": 1024
        },
        {
          "resolution": [
            32,
            32,
            4
          ],
          "cubeLength": 1024
        },
        {
          "resolution": [
            64,
            64,
            8
          ],
          "cubeLength": 1024
        },
        {
          "resolution": [
            128,
            128,
            16
          ],
          "cubeLength": 1024
        },
        {
          "resolution": [
            256,
            256,
            32
          ],
          "cubeLength": 1024
        },
        {
          "resolution": [
            512,
            512,
            64
          ],
          "cubeLength": 1024
        }
      ],
      "elementClass": "uint8",
      "dataFormat": "wkw"
    },
    {
      "name": "prediction",
      "category": "color",
      "boundingBox": {
        "topLeft": [
          320,
          320,
          80
        ],
        "width": 42368,
        "height": 34160,
        "depth": 2392
      },
      "wkwResolutions": [
        {
          "resolution": [
            16,
            16,
            4
          ],
          "cubeLength": 256
        }
      ],
      "elementClass": "uint8",
      "dataFormat": "wkw"
    }
  ]
}
```

wK will produce the placeholder mags: 1, 2, 4, 8 which is wrong since it violates the constraint that the components have to increase monotonously.

Relevant code:
- `getResolutionUnion` in `dataset_accessor.js`
- tests in `model_resolutions.spec.js`

daniel-wer:
- https://fixdenseanisotropicresolutions.webknossos.xyz

### Steps to test:
- I adjusted the expected unit test output (not sure why that didn't raise any red flags earlier) so that should be enough, I think

### Issues:
- fixes #5972 

------
- [x] Ready for review


daniel-wer:
On a side note, one thing that I noticed while I was looking at this issue. Since the densification of the layer resolutions is done for each layer separately, the computed resolutions might differ between layers (as can be seen in the tests).
In practice, we avoid this by using `getResolutions` which does `convertToDenseResolution(getResolutionUnion(dataset))` **almost** everywhere.

However, there is one spot where `convertToDenseResolution` is called directly (through  `getDenseResolutions`) which is in `save_saga:markBucketsAsNotDirty` ([link to code](https://github.com/scalableminds/webknossos/blob/master/frontend/javascripts/oxalis/model/sagas/save_saga.js#L777)). I was wondering whether this might be a potential bug, where some buckets cannot be marked as notDirty for datasets where the resolutions between layers might differ. It could also be that such buckets for resolutions that don't exist in the segmentation layer will never be part of the save queue, but I think it could actually happen that buckets in "placeholder" resolutions are annotated.

frontend/javascripts/test/model/model_resolutions.spec.js
```diff
@@ -32,7 +32,7 @@ test("Complex convertToDenseResolution", t => {
   ensureMatchingLayerResolutions(dataset);
   const expectedResolutions = {
     "0": [[1, 1, 1], [2, 2, 1], [4, 4, 1], [8, 8, 1], [16, 16, 2], [32, 32, 4]],
-    "1": [[1, 1, 1], [2, 2, 2], [4, 4, 4], [8, 8, 8], [16, 16, 16], [32, 32, 4]],
+    "1": [[1, 1, 1], [2, 2, 2], [4, 4, 4], [8, 8, 4], [16, 16, 4], [32, 32, 4]],
```
philippotto, line 35: Whoops :see_no_evil: Great that you spot this :+1:

philippotto (approved):
Great fix :+1:

daniel-wer:
@philippotto I forgot to mention you in the comment I made above. Could you have a look at whether my observation indeed points at a possible bug? If so, I could try to fix it in this PR as well.

philippotto:
> However, there is one spot where `convertToDenseResolution` is called directly (through `getDenseResolutions`) which is in `save_saga:markBucketsAsNotDirty` ([link to code](https://github.com/scalableminds/webknossos/blob/master/frontend/javascripts/oxalis/model/sagas/save_saga.js#L777)). I was wondering whether this might be a potential bug, where some buckets cannot be marked as notDirty for datasets where the resolutions between layers might differ. It could also be that such buckets for resolutions that don't exist in the segmentation layer will never be part of the save queue, but I think it could actually happen that buckets in "placeholder" resolutions are annotated.

Interesting point. Thanks for bringing this up :+1:  I had to dig into this a bit, but I now think, that this should not expose a bug (only a code smell). Here's why I think this: "Placeholder" resolutions should (must) never be labeled (otherwise, that would be a bug). This is why, the volumetracing_saga uses the `ResolutionInfo` class to get "usable" resolutions from potentially sparse resolutions. That helper class is more powerful because it can correctly handle resolutions without having to pad them (this is the reason why you only found one `convertToDenseResolution` call for volume-layers). Ideally, `markBucketsAsNotDirty` would also use that helper class, but the current way should work, too, as the densified resolution list is always a super set of the actually existing resolution list.

> [...] where some buckets cannot be marked as notDirty for datasets where the resolutions between layers might differ

Just to clarify this: the notDirty code will only be called for volume layers and there shouldn't be any mixing between layers (I just doublechecked that the volumetracing saga does not use `getResolutions`, but only `getResolutionInfo`).

daniel-wer:
Thanks a lot for digging into this. Your explanation makes sense and in that case I don't see any need for immediate action :)

philippotto:
@daniel-wer I guess this PR can be merged then? :)

PR #5987, opened 2022-01-24
Fix nml upload by providing process browser shim

daniel-wer:
The saxophone library depends on a library that accesses node's process dependency even when in the browser. Related to the webpack upgrade, since webpack no longer auto-provides these node polyfills.

Unfortunately, the nml parsing tests didn't catch this since they are executed in node. Ideally we could configure ava (or babel?) to simulate a browser environment and not include the node globals :thinking: 

- https://fixprocessundefined.webknossos.xyz

### Steps to test:
- Upload an NML in the frontend

------
- [x] Ready for review

### DIFF
--- a/package.json
+++ b/package.json
@@ -166,6 +166,7 @@
     "nanoevents": "^5.1.8",
     "pako": "^0.2.8",
     "pretty-bytes": "^5.1.0",
+    "process": "^0.11.10",
     "protobufjs": "^6.8.6",
     "react": "^16.12.0",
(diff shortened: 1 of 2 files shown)


philippotto (approved):
Great fixing :+1: (I didn't test it, since I assume you did a before/after test)

philippotto:
> Ideally we could configure ava (or babel?) to simulate a browser environment and not include the node globals thinking

Will be tricky probably, since we would also need to run the webpack output in node then to validate the webpack configuration..

PR #5993, opened 2022-01-26
[WIP] Add loaded meshes to sharing link

daniel-wer:
- Includes some cleanup (review first commit separately to have an isolated look)
  - I was annoyed that because flow returns mixed for Object.values, we often had to add $FlowIssue comments, or worse, adapted the code to work around that by using Object.keys, converting the key to int and then mapping to get the values. I introduced `Utils.values` which returns the correct type and should be the only spot where the $FlowIssue comment is needed.
  - A couple of times I specified the type if the values were mapped, afterwards, which is a separate issue, since mapping `Array<number>` will lead to flow assuming a type of `number | any` for the mapped value argument :shrug: I did not systematically look for these occurrences, but I assume there are quite a few in our code, leading to less type security.
- 

- https://___.webknossos.xyz

### Steps to test:
- abc

### Issues:
- fixes #

------
- [ ] Ready for review

### DIFF
--- a/frontend/javascripts/oxalis/api/api_latest.js
+++ b/frontend/javascripts/oxalis/api/api_latest.js
@@ -140,6 +140,7 @@ import * as Utils from "libs/utils";
 import dimensions from "oxalis/model/dimensions";
 import messages from "messages";
 import window, { location } from "libs/window";
+import DataLayer from "oxalis/model/data_layer";
 
 type OutdatedDatasetConfigurationKeys = "segmentationOpacity" | "isSegmentationDisabled";
 
@@ -1009,9 +1010,7 @@ class DataApi {
    */
   async reloadBuckets(layerName: string): Promise<void> {
     await Promise.all(
-      Object.keys(this.model.dataLayers).map(async currentLayerName => {
-        const dataLayer = this.model.dataLayers[currentLayerName];
-
+      Utils.values(this.model.dataLayers).map(async (dataLayer: DataLayer) => {
         if (dataLayer.name === layerName) {
           if (dataLayer.cube.isSegmentation) {
             await Model.ensureSavedState();
@@ -1030,7 +1029,7 @@ class DataApi {
     if (hasVolumeTracings(Store.getState().tracing)) {
       await Model.ensureSavedState();
     }
-    _.forEach(this.model.dataLayers, dataLayer => {
+    _.forEach(this.model.dataLayers, (dataLayer: DataLayer) => {
       dataLayer.cube.collectAllBuckets();
       dataLayer.layerRenderingManager.refresh();
     });
--- a/frontend/javascripts/oxalis/view/action-bar/save_button.js
+++ b/frontend/javascripts/oxalis/view/action-bar/save_button.js
@@ -17,6 +17,7 @@ import {
   LoadingOutlined,
 } from "@ant-design/icons";
 import ErrorHandling from "libs/error_handling";
+import { values } from "libs/utils";
 
 type OwnProps = {|
   onClick: (SyntheticInputEvent<HTMLButtonElement>) => Promise<*>,
@@ -133,8 +134,7 @@ function getOldestUnsavedTimestamp(saveQueue): ?number {
   if (saveQueue.skeleton.length > 0) {
     oldestUnsavedTimestamp = saveQueue.skeleton[0].timestamp;
   }
-  for (const volumeQueueKey of Object.keys(saveQueue.volumes)) {
-    const volumeQueue = saveQueue.volumes[volumeQueueKey];
+  for (const volumeQueue of values(saveQueue.volumes)) {
     if (volumeQueue.length > 0) {
       const oldestVolumeTimestamp = volumeQueue[0].timestamp;
       oldestUnsavedTimestamp = Math.min(
--- a/frontend/javascripts/admin/user/permissions_and_teams_modal_view.js
+++ b/frontend/javascripts/admin/user/permissions_and_teams_modal_view.js
@@ -8,6 +8,7 @@ import update from "immutability-helper";
 import type { APIUser, APITeam, APITeamMembership } from "types/api_flow_types";
 import { updateUser, getEditableTeams } from "admin/admin_rest_api";
 import messages from "messages";
+import * as Utils from "libs/utils";
 
 const RadioButton = Radio.Button;
 const RadioGroup = Radio.Group;
@@ -129,7 +130,7 @@ class PermissionsAndTeamsModalView extends React.PureComponent<TeamRoleModalProp
   setPermissionsAndTeams = () => {
     const newUserPromises = this.props.users.map(user => {
       if (this.props.selectedUserIds.includes(user.id)) {
-        const newTeams = ((Object.values(this.state.selectedTeams): any): Array<APITeamMembership>);
+        const newTeams = Utils.values(this.state.selectedTeams);
         const newUser = Object.assign({}, user, { teams: newTeams });
         if (this.props.activeUser.isAdmin && this.props.selectedUserIds.length === 1) {
           // If the current user is admin and only one user is edited we also update the permissions.
--- a/frontend/javascripts/libs/utils.js
+++ b/frontend/javascripts/libs/utils.js
@@ -24,6 +24,11 @@ export function mod(x: number, n: number) {
   return ((x % n) + n) % n;
 }
 
+export function values<K, V>(o: { [K]: V }): Array<V> {
+  // $FlowIssue[incompatible-return] remove once https://github.com/facebook/flow/issues/2221 is fixed
+  return Object.values(o);
+}
+
 export function map2<A, B>(fn: (A, number) => B, tuple: [A, A]): [B, B] {
   const [x, y] = tuple;
   return [fn(x, 0), fn(y, 1)];
@@ -514,8 +519,8 @@ export function filterWithSearchQueryOR<T: { +[string]: mixed }, P: $Keys<T>>(
       _.some(properties, fieldName => {
         const value = typeof fieldName === "function" ? fieldName(model) : model[fieldName];
         if (value != null && (typeof value === "string" || value instanceof Object)) {
-          const values = getRecursiveValues(value);
-          return _.some(values, v => v != null && v.toString().match(regexp));
+          const recursiveValues = getRecursiveValues(value);
+          return _.some(recursiveValues, v => v != null && v.toString().match(regexp));
         } else {
           return false;
         }
@@ -545,8 +550,8 @@ export function filterWithSearchQueryAND<T: { +[string]: mixed }, P: $Keys<T>>(
         _.some(properties, fieldName => {
           const value = typeof fieldName === "function" ? fieldName(model) : model[fieldName];
           if (value !== null && (typeof value === "string" || value instanceof Object)) {
-            const values = getRecursiveValues(value);
-            return _.some(values, v => v != null && v.toString().match(pattern));
+            const recursiveValues = getRecursiveValues(value);
+            return _.some(recursiveValues, v => v != null && v.toString().match(pattern));
           } else {
             return false;
           }
--- a/frontend/javascripts/oxalis/merger_mode.js
+++ b/frontend/javascripts/oxalis/merger_mode.js
@@ -13,9 +13,9 @@ import api from "oxalis/api/internal_api";
 import messages from "messages";
 
 type MergerModeState = {
-  treeColors: Object,
-  colorMapping: Object,
-  nodesPerSegment: Object,
+  treeColors: { [number]: ?number },
+  colorMapping: { [number]: number },
+  nodesPerSegment: { [number]: number },
   nodes: Array<NodeWithTreeId>,
   // A properly initialized merger mode should always
   // have a segmentationLayerName. However, some edge cases
@@ -41,7 +41,7 @@ function getTreeColor(treeId: number, mergerModeState: MergerModeState) {
   const { treeColors } = mergerModeState;
   let color = treeColors[treeId];
   // Generate a new color if tree was never seen before
-  if (color === undefined) {
+  if (color == null) {
     color = Math.ceil(127 * Math.random());
     treeColors[treeId] = color;
   }
@@ -252,8 +252,8 @@ function shuffleColorOfCurrentTree(mergerModeState: MergerModeState) {
     treeColors[activeTreeId] = undefined;
     // Applies the change of the color to all connected segments
     Object.keys(colorMapping).forEach(key => {
-      if (colorMapping[key] === oldColor) {
-        colorMapping[key] = getTreeColor(activeTreeId, mergerModeState);
+      if (colorMapping[+key] === oldColor) {
+        colorMapping[+key] = getTreeColor(activeTreeId, mergerModeState);
       }
     });
     // Update the segmentation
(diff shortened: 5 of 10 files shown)


frontend/javascripts/oxalis/model/sagas/isosurface_saga.js
```diff
@@ -369,6 +358,164 @@ function* maybeLoadIsosurface(
...
+    cancel: _take(
+      otherAction =>
+        otherAction.type === "REMOVE_ISOSURFACE" &&
+        otherAction.cellId === cellId &&
+        otherAction.layerName === layerName,
+    ),
+  });
+}
+
+function* loadPrecomputedMeshForSegmentId(
```
daniel-wer, line 460: This function was moved from `segments_view_helper.js` and converted to a saga. Because of the race- and cancel-ability of sagas, the numerous checks whether the mesh was deleted in the mean time could be removed.

daniel-wer:
@philippotto I have mostly tested the happy paths for this PR, but will do some more monkey testing next week. Still, this should be ready for a first review :)

frontend/javascripts/oxalis/model/sagas/isosurface_saga.js
```diff
@@ -369,6 +374,168 @@ function* maybeLoadIsosurface(
...
+      id,
+      chunkPosition,
+    );
+
+    const geometry = parseStlBuffer(stlData);
+    getSceneController().addIsosurfaceFromGeometry(geometry, id);
+  });
+
+  try {
+    yield* call(processTaskWithPool, tasks, PARALLEL_PRECOMPUTED_MESH_LOADING_COUNT);
```
philippotto, line 525: Are you sure that this function is cancelled once it was kicked off? I would have assumed that once an element from `tasks` has been started, that it's not cancelable by the redux sagas (a symptom of this is that the tasks are `async`s and not generator functions). The proper way to fix this would probably be to convert `processTaskWithPool` to be saga-based. It's not used anywhere else, so adapting callers wouldn't be a problem. However, adapting the tests could get a bit annoying..

A simple alternative would obviously be to re-introduce some of the "is canceled"-checks :see_no_evil:

philippotto (changes requested):
Great work & refactoring :+1: It's a joy to read through these changes :man_dancing: 

I only have some concerns with `processTaskWithPool`. Maybe they are unfounded (I didn't test my assumption), but my current understanding is that this needs some changes.. Happy to call about this, if you want :call_me_hand:

frontend/javascripts/oxalis/controller/url_manager.js
```diff
@@ -181,6 +195,20 @@ class UrlManager {
         stateByLayer[layerName] = { mappingInfo: { mappingName, mappingType } };
       }
     }
+    for (const layerName of Object.keys(state.localSegmentationData)) {
+      const { isosurfaces, currentMeshFile } = state.localSegmentationData[layerName];
+      if (currentMeshFile != null) {
+        const { meshFileName } = currentMeshFile;
+        const meshes = Utils.values(isosurfaces)
+          .filter(({ isVisible }: IsosurfaceInformation) => isVisible)
+          .map(({ segmentId, seedPosition, isPrecomputed }: IsosurfaceInformation) => ({
```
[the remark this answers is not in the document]
daniel-wer: I added these types because of this observation:

> A couple of times I specified the type if the values were mapped, afterwards, which is a separate issue, since mapping `Array<number>` will lead to flow assuming a type of `number | any` for the mapped value argument :shrug: I did not systematically look for these occurrences, but I assume there are quite a few in our code, leading to less type security.

However, I revisited the type annotations and found out that the `any`s were not caused by the map, but instead because `stateByLayer` was initialized with `{}` without a type annotation. This led to flow inferring less-restrictive types. Adding the correct type annotation `UrlStateByLayer` to the initialization fixes this. So, as you say, these types can be removed and mapping doesn't produce `any`s :+1:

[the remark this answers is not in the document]
philippotto: Ah, makes sense. Thanks for the explanation!

frontend/javascripts/oxalis/view/action-bar/save_button.js
```diff
@@ -17,6 +17,7 @@ import {
   LoadingOutlined,
 } from "@ant-design/icons";
 import ErrorHandling from "libs/error_handling";
+import { values } from "libs/utils";
```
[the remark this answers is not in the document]
daniel-wer: Seems like a good idea :)

frontend/javascripts/oxalis/model_initialization.js
```diff
@@ -650,5 +655,24 @@ function applyLayerState(stateByLayer: UrlStateByLayer) {
         }
       }
     }
+
+    if (layerState.meshInfo) {
+      const { meshFileName, meshes } = layerState.meshInfo;
+
+      Store.dispatch(updateCurrentMeshFileAction(effectiveLayerName, meshFileName));
```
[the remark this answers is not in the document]
daniel-wer: Good that you mention it!
Currently, there is no way to find out which mesh file was used to load a precomputed mesh, after the fact. There is a `  +fileName?: string,` property in the `IsosurfaceInformation` in the store, but it doesn't seem to be populated, ever.

A related problem is which mapping was activated when an ad-hoc mesh was generated. Currently, this information is also not saved in any way.

I see two main directions we could take
(1) Save the `meshFileName` and `mappingName` information on a per-mesh basis and include it in the url. This would further increase the url length (maximum length for Edge [seems to be](https://stackoverflow.com/a/417184) 2047 characters).
(2) Revisit whether meshes should be invalidated/removed/hidden in certain situations (e.g.: if the mesh file is changed or the mapping is changed). Currently, the UI in the Segments tab can be pretty misleading. For example it is not possible to compute a mesh for segment 4 in the over segmentation, change the mapping to agglomerate_view_80 and compute a mesh for segment 4, there. I feel like we should clarify what should happen in these situations and then derive the url scheme from that, maybe?

Happy to jump on a call

frontend/javascripts/oxalis/model/sagas/isosurface_saga.js
daniel-wer: Good point, I'll have a look and test what happens, but you're very likely correct.

> However, adapting the tests could get a bit annoying..

I don't think there are any tests for the mesh functionality yet, at least I haven't come across them :sweat_smile:
My plan is to add a screenshot test for this functionality. However, I'll also try to add a couple "higher-level" saga tests and will see whether it is too much "pain".

philippotto: > I don't think there are any tests for the mesh functionality yet, at least I haven't come across them sweat_smile

I meant the unit tests which exist for `processTaskWithPool` :)

frontend/javascripts/oxalis/model_initialization.js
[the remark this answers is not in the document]
philippotto: Good points!

I think, we should explore direction (2) and see how we can handle these situations elegantly. However, I'd say this is better suited for a follow-up issue. Do you agree?

Approach (1) would probably work (and would be relatively simple, when ignoring URL limits), but the actual complexity (highlighted by (2) wouldn't be solved. So, it doesn't feel like the right way.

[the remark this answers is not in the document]
daniel-wer: > I think, we should explore direction (2) and see how we can handle these situations elegantly. However, I'd say this is better suited for a follow-up issue. Do you agree?

Yes, I agree. So would you say, we keep the format as proposed in this PR for now (although it does not correctly work in some situations)?

Since we don't have a method to evolve the json url format to support older formats, yet, I would like to try to keep changes to a minimum, but it also doesn't feel right to merge this with known bugs :thinking:

frontend/javascripts/oxalis/api/api_latest.js
```diff
@@ -1654,7 +1653,9 @@ class DataApi {
       }
     }
 
-    await loadMeshFromFile(segmentId, seedPosition, meshFileName, segmentationLayer, dataset);
+    Store.dispatch(
```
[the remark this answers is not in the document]
daniel-wer: It is used and awaited in the connectome viewer (via the cross origin api). However, as far as I remember this stems from a time, where deleting a mesh while it was loading led to errors. Therefore, it is no longer relevant to await the mesh loading and might even speed up the connectome viewer, because both meshes will be loaded in parallel instead of sequentially.
Since the mesh loading can take quite long, I'm also not sure whether the blocking nature of this function would add much value. I'll adapt the usages and doc-string.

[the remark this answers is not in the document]
philippotto: > It is used and awaited in the connectome viewer (via the cross origin api). However, as far as I remember this stems from a time, where deleting a mesh while it was loading led to errors. Therefore, it is no longer relevant to await the mesh loading and might even speed up the connectome viewer, because both meshes will be loaded in parallel instead of sequentially.

Two birds with one stone, I guess :tada: 

> Since the mesh loading can take quite long, I'm also not sure whether the blocking nature of this function would add much value. I'll adapt the usages and doc-string.

Sounds very plausible :+1:

frontend/javascripts/libs/task_pool.js
```diff
@@ -1,36 +1,50 @@
...
+  }
 
-    for (let i = 0; i < poolSize; i++) {
-      startNextTask();
-    }
-  });
+  for (let i = 0; i < poolSize; i++) {
+    yield* fork(startNextTask);
+  }
+  // The saga will wait for all forked tasks to terminate before returning.
```
philippotto, line 49: Once again, I'm noticing that redux saga is really neat :) I didn't know (or forgot?) the detail about attached vs unattached forks and was a bit surprised that this works that way (and that `spawn` is the counterpart to `fork`). I think, it wouldn't hurt to mention this detail in the comment, too. Then, the next reader might save one doc lookup :)

```suggestion
  // The saga will wait for all forked tasks to terminate before returning, because
  // fork() creates attached forks (in contrast to spawn()).
```

frontend/javascripts/oxalis/model/sagas/isosurface_saga.js
```diff
@@ -369,6 +374,172 @@ function* maybeLoadIsosurface(
...
+          dataset.dataStore.url,
+          dataset,
+          getBaseSegmentationName(segmentationLayer),
+          fileName,
+          id,
+          chunkPosition,
+        );
+
+        const geometry = parseStlBuffer(stlData);
+        getSceneController().addIsosurfaceFromGeometry(geometry, id);
```
philippotto, line 524: It probably doesn't make a real difference, since the parsing is synchronous, but using `call` for the two lines, would be more consistent, I guess.

philippotto (commented):
Great adaptation of the pool "saga" :1st_place_medal: I couldn't get the sharing link to work, yet, though. I tried it with adhoc meshes and the URL doesn't seem to contain anything mesh related :thinking:

frontend/javascripts/oxalis/model/sagas/isosurface_saga.js
```diff
@@ -369,6 +416,192 @@ function* maybeLoadIsosurface(
...
+  // By that we avoid to remove cells that got annotated during reloading from the modifiedCells set.
+  const currentlyModifiedCells = new Set(modifiedCells);
+  modifiedCells.clear();
+
+  const segmentationLayer = Model.getVisibleSegmentationLayer();
+  if (!segmentationLayer) {
+    return;
+  }
+  // First create an array containing information about all loaded isosurfaces as the map is manipulated within the loop.
+  const isosurfacesMapForLayer = isosurfacesMapByLayer[segmentationLayer.name] || new Map();
```
philippotto, line 439: If the `new Map` branch is hit, it probably should be stored in `isosurfacesMapByLayer`. Maybe something like this: 

```suggestion
  isosurfacesMapByLayer[segmentationLayer.name] = isosurfacesMapByLayer[segmentationLayer.name] || new Map();
  const isosurfacesMapForLayer = isosurfacesMapByLayer[segmentationLayer.name];
```

frontend/javascripts/oxalis/model_initialization.js
```diff
@@ -650,5 +655,36 @@ function applyLayerState(stateByLayer: UrlStateByLayer) {
...
+            loadPrecomputedMeshAction(segmentId, seedPosition, meshFileName, effectiveLayerName),
+          );
+        } else {
+          const { mappingName, mappingType } = mesh;
+          Store.dispatch(
+            loadAdHocMeshAction(
+              segmentId,
+              seedPosition,
+              // `undefined` values will be interpreted to mean the currently active mapping which
+              // is not desired for meshes loaded via url.
```
philippotto, line 680: This seems a bit dangerous. Maybe the action creator should accept an optional `mappingInfo` object? If `mappingInfo == null`, the currently active mapping (if it exists) is used. If `mappingInfo` is defined, its content is used (which also allows `{mappingName: undefined, mappingName: undefined}` to enforce that no mapping is used). What do you think?

philippotto (commented):
I wasn't sure, whether the new round is ready for review again, but especially the last commit seemed like a bigger chunk ready to review :) Great work refactoring this! I only left two suggestions. Let me know when/whether I should also test again :)

frontend/javascripts/oxalis/model_initialization.js
daniel-wer: That's a very good idea and the code looks much better now :)

daniel-wer:
> I wasn't sure, whether the new round is ready for review again, but especially the last commit seemed like a bigger chunk ready to review :)

Thanks a lot!

I addressed your feedback, would be great if you could do another test round.
I'll also test some more and will add a screenshot test as well.

philippotto (approved):
Sorry, this PR slipped through on my side! I reviewed the latest commits and also did another testing round. Looks very solid :+1: Looking forward to see this live :tada:

daniel-wer:
Nightly test ran through well: https://app.circleci.com/pipelines/github/scalableminds/webknossos/5717/workflows/22bbe067-6ff4-4a38-a907-c37a4132ee19/jobs/17004

frontend/javascripts/test/puppeteer/dataset_rendering_helpers.js
```diff
@@ -124,17 +127,29 @@ async function waitForTracingViewLoad(page: Page) {
...
+  while (currentShot == null || !isPixelEquivalent(changedPixels, width, height)) {
+    console.log(`Waiting for rendering to finish. Changed pixels: ${changedPixels}`);
     await page.waitForTimeout(10000);
     currentShot = await page.screenshot({ fullPage: true });
     if (lastShot != null) {
-      changedPixels = pixelmatch(lastShot, currentShot, {}, 1920, 1080, {
-        threshold: 0.0,
-      });
+      changedPixels = pixelmatch(
+        // The buffers need to be converted to png before comparing them
```
philippotto, line 142: Has this become necessary due to the update to `pixelmatch`? Or has this always been an issue? Good catch, either way :+1:

philippotto (approved):
Wonderful! Thank you so much for taking care of the screenshot tests :pray: I reviewed the latest commits and everything looks good!

frontend/javascripts/test/puppeteer/dataset_rendering_helpers.js
daniel-wer: From what I read and experienced, `pixelmatch` started to assert this in the new version but the problem existed before. Usually the difference was fairly small (which is why it mostly worked before), but I also encountered garbled diffs during my testing with other `--use-gl` flags.

PR #5999, opened 2022-01-31
Don't enforce save state when saving is triggered by a timeout

### LINKED ISSUE #5681: Moving while saving initiates too many save requests
## Context
When saving an annotation, one can provoke countless save requests by simply moving continuously through the dataset.

## Expected Behavior
Batching subsequent move actions should still be batched.

## Current Behavior
For each move action, there will be a save request.

## Steps to Reproduce the bug

1. move through an annotation with "f" for 60s
2. after 30s the saving should start and with that, the countless save requests will be triggered

## Your Environment for bug
chrome/linux

## Details
The culprit is at: https://github.com/scalableminds/webknossos/blob/a5fafe017ba9de00f8d9bf0fd8b8ab2830587b8c/frontend/javascripts/oxalis/model/sagas/save_saga.js#L404-L404

The logic here is intuitive, but leads to the behavior explained above. I think, we should only drain the save queue if the user explicitly clicked "Save" (i.e., the `SAVE_NOW` action was dispatched). In that case, the user expects the spinner to step in the next seconds. For the auto-save, it's fine to simply empty the save-queue and ignore new items added to it afterwards.

MichaelBuessemeyer:
- https://___.webknossos.xyz

### Steps to test:
- Open a tracing.
- Open the network tab of the browser and clear it.
- Move through the dataset for at least 30 seconds.
- Search in the network tab with the string "update"
- There should only be 2 update calls triggered aber 30 seconds of moving
- Do the same thing on the master. There should be much more update request. They should continue being sent until you stop moving.

### Issues:
- fixes #5681

------
- [x] Ready for review

### DIFF
--- a/frontend/javascripts/oxalis/model/sagas/save_saga.js
+++ b/frontend/javascripts/oxalis/model/sagas/save_saga.js
@@ -664,19 +664,25 @@ export function* pushTracingTypeAsync(
       // Save queue is empty, wait for push event
       yield* take("PUSH_SAVE_QUEUE_TRANSACTION");
     }
-    yield* race({
+    const { forcePush } = yield* race({
       timeout: _delay(PUSH_THROTTLE_TIME),
       forcePush: _take("SAVE_NOW"),
     });
     yield* put(setSaveBusyAction(true, tracingType));
-    while (true) {
-      // Send batches to the server until the save queue is empty
-      saveQueue = yield* select(state => selectQueue(state, tracingType, tracingId));
-      if (saveQueue.length > 0) {
-        yield* call(sendRequestToServer, tracingType, tracingId);
-      } else {
-        break;
+    if (forcePush) {
+      while (true) {
+        // Send batches to the server until the save queue is empty
+        saveQueue = yield* select(state => selectQueue(state, tracingType, tracingId));
+        if (saveQueue.length > 0) {
+          yield* call(sendRequestToServer, tracingType, tracingId);
+        } else {
+          break;
+        }
       }
+    } else {
+      // Saving the tracing a via timeout only saves the current state.
+      // It does not require to hit a save state without changes.
+      yield* call(sendRequestToServer, tracingType, tracingId);
     }
     yield* put(setSaveBusyAction(false, tracingType));
   }


MichaelBuessemeyer:
@philippotto Do you want to / could you please take a look at this?

frontend/javascripts/oxalis/view/layouting/flex_layout_wrapper.js
```diff
@@ -313,7 +315,6 @@ class FlexLayoutWrapper extends React.PureComponent<Props, State> {
   }
 
   onLayoutChange = () => {
-    sendAnalyticsEvent("change_tracing_layout", { viewMode: this.props.layoutKey });
```
MichaelBuessemeyer, line 316: I removed the analytics here and moved it into `onAction` as each change to the layout, first triggers an action. This method gets information about what action was done and therefore can better decide whether to send an analytics event.

However, the `onAction` method is not called when the layout is completely rebuild. For that case the event is also send in `rebuildLayout`.

frontend/javascripts/test/sagas/save_saga.spec.js
```diff
@@ -103,7 +103,7 @@ test("SaveSaga should send update actions", t => {
   saga.next(); // select state
   expectValueDeepEqual(t, saga.next([]), take("PUSH_SAVE_QUEUE_TRANSACTION"));
   saga.next(); // race
-  saga.next(SaveActions.pushSaveQueueTransaction(updateActions));
```
MichaelBuessemeyer, line 106: The previous argument here was wrong as this `saga.next` corresponds to the `race` of the `save_saga.js` in lines 667ff. This race only takes a timeout or a `saveNowAction`. In order for the test to work with the changes of this pr, I had to correct this.

frontend/javascripts/oxalis/view/layouting/flex_layout_wrapper.js
```diff
@@ -68,6 +68,8 @@ type State = {
   model: Model,
 };
 
+const ignoredLayoutChangesByAnalytics = ["FlexLayout_SetActiveTabset"];
```
philippotto, line 71: I think, the following scenarios shouldn't send an event, either:

- toggling the side bars
- switching tabs (in the sidebar, the event is already ignored. but when moving XY and XZ next to each other, switching the tabs sends events)

philippotto (commented):
Great! Works very well :+1: Regarding the events, I noticed two further things which should probably be ignored.

frontend/javascripts/oxalis/view/layouting/flex_layout_wrapper.js
MichaelBuessemeyer: But moving the viewport around and resizing them should still send the event? Or should the event only be send when switching between layouts?

philippotto: > I think it depends on what exactly we want to track / use this tracking event for.

Yes, definitely. I don't have a strong opinion about what we really want to track, but I'm leaning towards tracking only stuff which is not "obvious" (e.g., it seems obvious that users will click on tabs to switch between them). With regard to maximizing, resizing and switching layouts, I'm not too sure that this is obvious for all users. Therefore, these events seem more interesting to me. However, in the end, this is also not very high-pri / not worth delving into how to filter the events best. So, maybe opt for the easier route if it makes sense (e.g., only track switches between layouts).

philippotto (approved):
Excellent!

frontend/javascripts/oxalis/view/layouting/flex_layout_wrapper.js
```diff
@@ -89,6 +91,13 @@ class FlexLayoutWrapper extends React.PureComponent<Props, State> {
...
+      sendAnalyticsEvent("switched_layout", {
+        from: { viewMode: prevProps.layoutKey, layoutName: prevProps.layoutName },
+        to: {
+          viewMode: this.props.layoutKey,
+          layoutName: this.props.layoutName,
+        },
+      });
       this.rebuildLayout();
     }
   }
```
MichaelBuessemeyer, line 103: Here is the change 👀

philippotto:
> I just added a commit that adds an additional analytics event that tracks when the user changed between any two layouts. This allows tracking whether multiple layouts are actively used / switched between or whether a user usually stays at only one active layout changing this one over and over again.
> Do you think this makes sense @philippotto? If not, I can just revert this pr.

Great! Makes total sense :+1:

philippotto:
@MichaelBuessemeyer What's the status here? I think, the PR can be merged?

MichaelBuessemeyer:
> @MichaelBuessemeyer What's the status here? I think, the PR can be merged?

Whoopsie, I forgot this pr.

Issue #5973, opened 2022-01-20
Refactor/harden volume annotation architecture/code


philippotto:
## Motivation
wK's volume annotation feature bears a lot of complexity due to the combination of the following aspects:
- dealing with potentially not-yet downloaded data (this is relevant for pure labeling but also for all other actions, such as undo)
- automatic down-/upsampling of volume data into other magnifications
- supporting front-end based undo/redo (using async compression)
- in-background, throttled persistence to server (note that versions on server are mostly not identical to versions in undo/redo stack)

The current code which backs these features is not straightforward to grasp. https://github.com/scalableminds/webknossos/pull/5955 fixes several bugs (mostly race conditions), but I'm sure we could improve the architecture + code so that we can be more confident that the volume annotation feature is robust.

## Todos
- [ ] Visualize current high-level architecture for the above features. We should clarify responsibilities of components, (time) dependencies of async operations (including the actual UI actions initiated by the user), (implicit) state machines, and potential problems in general
- [ ] Brainstorm how we could simplify the current architecture. For example:
  - [ ] lots of bucket-related code could also happen in sagas
  - [ ] more explicit state machines
  - [ ] ...
- [ ] Implement/refactor architecture
- [ ] Think about more edge cases
  - [ ] e.g., very slow save operations which might interfere with undo operations which were scheduled so early that they assume they get an older version from the backend (but in the meantime, a new version might have been stored)
- [ ] Write tests for these (the front-end test would probably need a mocked key-value store for the bucket persistence)

MichaelBuessemeyer:
Maybe using the [command pattern](https://en.wikipedia.org/wiki/Command_pattern) might make the code more readable. [This](https://betterprogramming.pub/utilizing-the-command-pattern-to-support-undo-redo-and-history-of-operations-b28fa9d58910) looks like a kinda useful code example.

PR #6000, opened 2022-01-31
Increase maximum bucket count limit for segmentation layers

philippotto:
This is a bandaid fix to lower the chance that annotating large amounts of data can cause a `A bucket was forcefully garbage-collected. This indicates that too many buckets are currently in RAM.` message. Especially, when annotating large structures in mag 1 (e.g., somata) a high amount of buckets can be labeled very quickly. Currently, wK does not cope very well with this. Increasing the threshold should help a bit (temporarily).

- https://___.webknossos.xyz

### Steps to test:
- I tested that the session start time is sent to airbrake correctly.
- I opened a dataset with a segmentation layer to check that it loads+renders correctly

### Issues:
- related to #5973

------
- [X] Ready for review


daniel-wer (approved):
LGTM as a temporary measure.

Please correct my math if it's wrong, but this seems to be quite a large amount of data:
With the 5000 bucket limit: 4 bytes per 32-bit ID * 32^3 voxels * 5000 buckets = ~655 MB
Now: ~1.3 GB
Since there are multiple layers, this adds up quick. Maybe we should have a look at whether we're loading unnecessary buckets again. This could be an alternative to increasing the limit.

For now, this solution is ok :)

philippotto:
> Please correct my math if it's wrong, but this seems to be quite a large amount of data:
With the 5000 bucket limit: 4 bytes per 32-bit ID * 32^3 voxels * 5000 buckets = ~655 MB
Now: ~1.3 GB
Since there are multiple layers, this adds up quick. Maybe we should have a look at whether we're loading unnecessary buckets again. This could be an alternative to increasing the limit.

Your math looks correct :) However, there's typically only one segmentation layer. Therefore, this change is comparable to having one additional volume layer per dataset (without the GPU-related resources luckily) which seems okay to me. We should definitely keep an eye on that and also follow-up with a proper fix (or at least undo this change). I'll make this todo more clear in #5973.

PR #6001, opened 2022-01-31
Allow Download of Public Annotations While Logged Out

fm3:
To support wk-libs example using public annotation

### Steps to test:
- set a dataset to public in dataset sharing settings
- create an annotation on that dataset, set to public as well
- open that annotation while logged out (e.g. private window)
- download annotation, should yield zip instead of error

------
- [x] Ready for review


jstriebel (approved):
Thanks for the fast help! :heart:

PR #6006, opened 2022-02-01
Add option to change active segement

### LINKED ISSUE #5935: Allow to make segment id active via rightclick in segments tab
## Detailed Description
Rightclicking a segment item in the "Segments" tab should expose a "Activate Segment ID" menu item which makes the selected segment id "active" (i.e., use `setActiveCellAction`).

## Context

Dagobert42:
- https://___.webknossos.xyz

### Steps to test:
- use brush tool
- create multiple segments
- open the segments tab
- the right-click context menu for each segment now gives an option to make the segment ID active
- the context menu item should be disabled if the segment ID is already acive (has the brush icon)

### Issues:
- fixes #5935 

------
- [✓] Updated [(unreleased) changelog](../blob/master/CHANGELOG.unreleased.md#unreleased)
- [✓] Ready for review

### DIFF
--- a/CHANGELOG.unreleased.md
+++ b/CHANGELOG.unreleased.md
@@ -11,6 +11,7 @@ For upgrade instructions, please check the [migration guide](MIGRATIONS.released
 [Commits](https://github.com/scalableminds/webknossos/compare/22.02.0...HEAD)
 
 ### Added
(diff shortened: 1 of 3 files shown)


philippotto (approved):
Flawless :+1:

PR #6009, opened 2022-02-02
use allDataBlocking and temporary files for explorative download

fm3:
Changing the wk-backend-side handling of downloading single annotations (tasks + exploratives, not compound): ditching the stream forwarding in favor of what has proven stable for compound annotations already (see #3959). A negative side effect could be slightly longer response times, since the file is only served after it is completely written. Though if prevents the dataloss we sometimes saw previously, I think it is worth it.

- https://volumedownloadblocking.webknossos.xyz

### Steps to test:
- create annotations + download them, check for expected content
  - skeleton-only
  - hybrid with one volume layer
  - hybrid with multiple volume layers

### Issues:
- possible fix for #5684 

------
- [x] Ready for review


jstriebel (approved):
Nice, LGTM! :+1: Thanks for the walk-through

PR #6022, opened 2022-02-03
For empty volume tracings, serve header-only zip

### LINKED ISSUE #6015: Empty volume layers should not yield zero-bytes zip
Options include
 - no zip at all, skip the location attribute in nml `<volume id="1" name="myVolume" />`
 - write empty zip with intact header
 - write zip with `.empty` file, with content something like "no volume data present"

fm3:
Before this, empty volume tracings (i.e. never brushed) resulted in zero-bytes data.zip files.
Now we create a 22-byte header-only “empty zip” file.

- https://emptyzip.webknossos.xyz

### Steps to test:
- click “create annotation” for a dataset
- do not brush, select“download”
- resulting zip should contain a data_Volume.zip file that can be opened as a zipfile (but does not have any contents)

### Issues:
- fixes #6015 

------
- [x] Ready for review

### DIFF
--- a/util/src/main/scala/com/scalableminds/util/io/ZipIO.scala
+++ b/util/src/main/scala/com/scalableminds/util/io/ZipIO.scala
@@ -93,17 +93,20 @@ object ZipIO extends LazyLogging {
   def zip(sources: List[NamedStream], out: OutputStream)(implicit ec: ExecutionContext): Future[Unit] =
     zip(sources.toIterator, out)
 
-  def zip(sources: Iterator[NamedStream], out: OutputStream)(implicit ec: ExecutionContext): Future[Unit] =
+  def zip(sources: Iterator[NamedStream], out: OutputStream)(implicit ec: ExecutionContext): Future[Unit] = {
+    val zip = startZip(out)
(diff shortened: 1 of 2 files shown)


util/src/main/scala/com/scalableminds/util/io/ZipIO.scala
```diff
@@ -93,17 +93,20 @@ object ZipIO extends LazyLogging {
...
 
-  def zip(sources: Iterator[NamedStream], out: OutputStream)(implicit ec: ExecutionContext): Future[Unit] =
+  def zip(sources: Iterator[NamedStream], out: OutputStream)(implicit ec: ExecutionContext): Future[Unit] = {
+    val zip = startZip(out)
     if (sources.nonEmpty) {
-      val zip = startZip(out)
       for {
         _ <- zipIterator(sources, zip)
         _ = zip.close()
+        _ = out.close()
```
fm3, line 102: I _tend_ to think that zip.close also does out.close but I’d rather be safe than sorry here

jstriebel (approved):
LGTM :+1:

PR #6024, opened 2022-02-07
Log legacy route usages

### LINKED ISSUE #5986: log usage of old versioned routes
We should find out if people are still using old versions of the API, as we develop the new versioning concept.

fm3:
Adds logging to legacy routes usages, including user id. This should give us an idea how much support we need to give the legacy routes in the future.

### Steps to test:
- visit http://localhost:9000/api/v3/projects/sampleProject
- backend logging should contain the route

### Issues:
- fixes #5986 

------
- [x] Ready for review


app/controllers/LegacyApiController.scala
```diff
@@ -452,4 +489,13 @@ class LegacyApiController @Inject()(annotationController: AnnotationController,
       } yield Ok(Json.toJson(newJson)).copy(header = result.header)
     } else Fox.successful(result)
 
+  private def logVersioned(request: SecuredRequest[WkEnv, _]): Unit =
+    logVersioned(request.identity, request.uri)
```
jstriebel, line 493: What exactly is `request.identity`?

jstriebel (approved):
LGTM :+1:

app/controllers/LegacyApiController.scala
fm3: `SecuredRequest[WkEnv]` is the request class constructed by Silhouette that enforces user authentication. It makes the requesting `User` (our own model, fetched from the db upon successful authentication) available as `request.identity`. The `UserAwareRequest[WkEnv]` is similar, but there the `request.identity` is an `Option[User]`

PR #6028, opened 2022-02-08
Allow uploading multi-layer volume annotations

### LINKED ISSUE #6020: Annotation Upload should read all volume layers.
Currently when uploading an annotation with multiple volume layers, only one will be respected. Instead, all should be.
We should discuss how the merging behavior should be (or if such annotations can only be uploaded one at a time)

fm3:
- https://uploadmultilayervolumes.webknossos.xyz

### Steps to test:
- Create volume annotation, add second layer in left sidebar, name the layers, brush different content
- download the annotation, reupload, should look the same
- with/without fallback layer information should also still be propagated
- uploading multiple such annotations is not supported
- uploading multiple volumes with *at most one* volume layer each is still supported in this case the volume content should all be merged into a single layer
- uploading multiple skeletons should yield one merged skeleton (with or without additional group hierarchy depending on checkbox)

### Issues:
- fixes #6020

------
- [x] Ready for review

### DIFF
--- a/app/controllers/AnnotationIOController.scala
+++ b/app/controllers/AnnotationIOController.scala
@@ -13,6 +13,7 @@ import com.scalableminds.webknossos.datastore.VolumeTracing.{VolumeTracing, Volu
 import com.scalableminds.webknossos.datastore.helpers.ProtoGeometryImplicits
 import com.scalableminds.webknossos.datastore.models.datasource.{
   AbstractSegmentationLayer,
+  DataLayerLike,
   GenericDataSource,
   SegmentationLayer
 }
@@ -138,7 +139,6 @@ Expects:
       return Fox.failure("Cannot merge multiple annotations that each have multiple volume layers.")
     if (volumeLayersGrouped.length == 1) { // Just one annotation was uploaded, keep its layers separate
       Fox.serialCombined(volumeLayersGrouped.toList.flatten) { uploadedVolumeLayer =>
-        val dataZipFile = otherFiles.get(uploadedVolumeLayer.dataZipLocation).map(_.path.toFile)
         for {
           savedTracingId <- client.saveVolumeTracing(uploadedVolumeLayer.tracing,
                                                      uploadedVolumeLayer.getDataZipFrom(otherFiles))
@@ -247,8 +247,8 @@ Expects:
       }
     } yield allAdapted
 
-  private def adaptPropertiesToFallbackLayer[T](volumeTracing: VolumeTracing,
-                                                dataSource: GenericDataSource[T]): Fox[VolumeTracing] =
+  private def adaptPropertiesToFallbackLayer[T <: DataLayerLike](volumeTracing: VolumeTracing,
+                                                                 dataSource: GenericDataSource[T]): Fox[VolumeTracing] =
     for {
       _ <- Fox.successful(())
       fallbackLayer = dataSource.dataLayers.flatMap {
@@ -371,7 +371,7 @@ Expects:
         _ = fetchedVolumeLayers.zipWithIndex.map {
           case (volumeLayer, index) =>
             volumeLayer.volumeDataOpt.foreach { volumeData =>
-              val dataZipName = volumeLayer.volumeDataZipName(index, fetchedSkeletonLayers.length == 1)
+              val dataZipName = volumeLayer.volumeDataZipName(index, fetchedVolumeLayers.length == 1)
               zipper.addFileFromBytes(dataZipName, volumeData)
             }
         }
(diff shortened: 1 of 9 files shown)


app/controllers/AnnotationIOController.scala
```diff
@@ -320,7 +371,7 @@ Expects:
         _ = fetchedVolumeLayers.zipWithIndex.map {
           case (volumeLayer, index) =>
             volumeLayer.volumeDataOpt.foreach { volumeData =>
-              val dataZipName = volumeLayer.volumeDataZipName(index, fetchedSkeletonLayers.length == 1)
+              val dataZipName = volumeLayer.volumeDataZipName(index, fetchedVolumeLayers.length == 1)
```
fm3, line 374: This was a bug: the data.zips should get their index in the file name if there are multiple *volume* layers, not *skeleton* layers

app/controllers/AnnotationIOController.scala
```diff
@@ -64,7 +69,7 @@ class AnnotationIOController @Inject()(
     value =
       """Upload NML(s) or ZIP(s) of NML(s) to create a new explorative annotation.
 Expects:
- - As file attachment: any number of NML files or ZIP files containing NMLs, optionally with at most one volume data ZIP referenced from an NML in a ZIP
+ - As file attachment: any number of NML files or ZIP files containing NMLs, optionally with volume data ZIPs referenced from an NML in a ZIP
```
jstriebel, line 72: So here also a single zip (or multiple) with multiple data-zips can be given, which is merged then, right? Could you please expand this a bit, e.g. list all the different possibilities, e.g.:
```
Each file creates an annotation. If a file is a zipfile, the following content combinations are allowed:
* 1+ nml files --> Skeleton Tracing, nmls either merged or as groups depending on createGroupForEachFile
* 1+ zip files, each containing an nml and optionally a single volume layer (is the name important here, does it need to be referenced from the nml?) --> Hybrid Tracing, nmls either merged or as groups depending on createGroupForEachFile, the volume annotations are merged (how? bucket-wise? which one wins?)
* a combination of the two above, where the nmls are merged
* 1 zip file, containing an nml and multiple volume layers --> Hybrid tracing with multiple volume annotation layers
```
That's at least the rough understanding of the intended usage I have from that route without testing it in detail^^ Is this correct? It seems that everything is actually unpacked recursively, so merging multiple layers is only prevented if they have distinct names, right?

app/controllers/AnnotationIOController.scala
```diff
@@ -135,6 +131,55 @@ Expects:
       }
   }
 
+  private def mergeAndSaveVolumeLayers(volumeLayersGrouped: Seq[List[UploadedVolumeLayer]],
+                                       client: WKRemoteTracingStoreClient,
+                                       otherFiles: Map[String, TemporaryFile]): Fox[List[AnnotationLayer]] = {
+    if (volumeLayersGrouped.isEmpty) return Fox.successful(List())
+    if (volumeLayersGrouped.length > 1 && volumeLayersGrouped.exists(_.length > 1))
+      return Fox.failure("Cannot merge multiple annotations that each have multiple volume layers.")
+    if (volumeLayersGrouped.length == 1) { // Just one annotation was uploaded, keep its layers separate
```
jstriebel, line 140: I'm not sure if I'm understanding `Just one annotation was uploaded, keep its layers separate`. There can still be multiple annotations, but each of those has a different layer, right?

jstriebel (commented):
Nice stuff! I'm not yet seeing though :100:%, but mostly LGTM. I already added some comments, but let's maybe have a call about this?

app/controllers/AnnotationIOController.scala
```diff
@@ -86,42 +91,33 @@ Expects:
...
-
           for {
-            _ <- bool2Fox(skeletonTracings.nonEmpty || volumeTracingsWithDataLocations.nonEmpty) ?~> "nml.file.noFile"
-            dataSet <- findDataSetForUploadedAnnotations(skeletonTracings, volumeTracingsWithDataLocations.map(_._1))
+            parseSuccesses <- Fox.serialCombined(parseResultsFiltered)(r => r.toSuccessBox)
+            name = nameForUploaded(parseResultsFiltered.map(_.fileName))
+            description = descriptionForNMLs(parseResultsFiltered.map(_.description))
+            _ <- assertNonEmpty(parseSuccesses)
+            skeletonTracings = parseSuccesses.flatMap(_.skeletonTracing)
+            volumeLayersGroupedRaw = parseSuccesses.map(_.volumeLayers).filter(_.nonEmpty)
```
fm3, line 109: add comment how this grouping works

jstriebel (approved):
Thanks a lot for guiding my through the code! With the added comments this LGTM :+1: :shipit:

app/controllers/AnnotationIOController.scala
fm3: As discussed in person it is slightly different. This route always results in a single annotation. I added comments in code describing the different cases.

PR #6029, opened 2022-02-08
Fix occasional incorrect fallback rendering

### LINKED ISSUE #6138: Annotating an anisotropic dataset with volume resolution restrictions behaves incorrectly
## Context

This bug occured during volume annotation on an anisotropic dataset with resolutions 1-1-1 and 2-2-1. Volume annotation was restricted to resolution 2-2-1.

## Expected Behavior

Only the z-slice that was annotated should be annotated, both in mag 1-1-1 and 2-2-1.

## Current Behavior

If brushing in mag 1-1-1, the next slice is mistakenly annotated as well. In mag 2-2-1, the next slice is not annotated (as expected).
If brushing in mag 2-2-1, only a single slice is annotated in mag 2-2-1 (as expected). However, in mag 1-1-1, the previous slice is annotated as well.

## Steps to Reproduce the bug

1. Create a hybrid annotation on an anisotropic dataset (with mags 1-1-1 and 2-2-1) and restrict the volume resolution to disallow annotation in mag 1-1-1
2. Use the brush to annotate in mag 1-1-1 -> The next slice is mistakenly annotated as well. In mag 2-2-1, the next slice is not annotated.
3. Also, use the brush to annotate in mag 2-2-1 -> In mag 1-1-1, the previous slice is annotated as well.

## Your Environment for bug

- Version of WebKnossos (Release or Commit): 22.04.0

philippotto:
This PR fixes how the look up buffer for the GPU rendering was written to. The new code only writes a bucket into a slot of the look up texture if there is not already another bucket there with a higher quality. The old code did not check this which is why sometimes data from a coarser mag was rendered even though a finer quality was available.

### Todo:
- [x] fix flicker when zooming

- https://ensurecorrectfallbackrendering.webknossos.xyz

### Steps to test:
- See the reproduction steps in https://github.com/scalableminds/webknossos/issues/6138
- There should be no flickering when zooming in or out (and switching zoomsteps)

### Issues:
- fixes #6138

------
- [x] Ready for review


daniel-wer:
@philippotto I think I found the culprit for the flickering and fixed it :)

I was also able to roughly observe what was happening. See this log from my debugging session (The first number indicates the bucket's zoomstep and the second number `posInBuffer`. I've focused on a single bucket, the one with `posInBuffer` 66134):

```
Zoomed from zoomstep 0 to zoomstep 1 ...
###### Refresh lookup buffer for zoomstep 1
Base bucket fallback written 1 66134
Fallback written 0 66134
###### Refresh lookup buffer for zoomstep 1
Base bucket fallback written 1 66134
Fallback written 0 66134
###### Refresh lookup buffer for zoomstep 1
Base bucket fallback written 1 66134
###### Refresh lookup buffer for zoomstep 1
Base bucket fallback written 1 66134
...
```

Basically, during the first two refreshes of the lookup buffer, after zooming out from zoomstep 0 to zoomstep 1, the `activeBucketToIndexMap` still contained buckets in zoomstep 0. These buckets in the finer resolution were treated as fallback buckets (although they are anti-fallback buckets ^^), because `isBaseBucket` was false for them (`zoomStepDifference === 0`). As far as I can see, all of the remaining code expected `zoomStepDifference` to be positive, but it was negative in that case, leading to weird effects which I didn't fully debug.
This bug existed before, but apparently had no effect, because the buckets with zoomstep 0 were never written to the lookup buffer, because the ones with zoomstep 1 were written before (the bug you fixed). Your fix to write buckets to the lookup buffer if they have a better zoomstep brought this bug to light.

philippotto:
Ah, awesome debugging & find! Focusing on a single bucket is a great idea (which I wish to have had myself back then ^^). I'm very happy that this turned out to be such a plausible bug and not some weird GPU problem :)

https://media.giphy.com/media/3ohzdIuqJoo8QdKlnW/giphy.gif

I'll make the PR merge-ready next.

daniel-wer (approved):
Thanks for making this PR merge-ready! LGTM :+1:

PR #6030, opened 2022-02-08
[Draft/deferred] Cuckoo hashing for data rendering

### LINKED ISSUE #3915: Refactor how buckets are addressed for GPU rendering
At the moment, buckets are addressed relative to an "anchor point", which means the look up buffer has to be rewritten every time the anchor point changes. Additionally, the code is quite cumbersome at the moment for various reason.
As an alternative, we should explore to address buckets via cuckoo hashing or a similar method. I'd also expect a performance gain from this.

philippotto:
This branch was created in March 2021 and deprioritzed shortly after. I opened the PR for it to avoid it being completely "gone".

- https://___.webknossos.xyz

### Steps to test:
- abc

### Issues:
- fixes #3915 

------
- [ ] Ready for review


philippotto:
Short update:
- cuckoo hashing is implemented (in JS) and should work
- the integration of the new hashing is blocked because the murmur hash is not compatible with webgl 1
- switching to webgl 2 is prepared, but does not work for some reason with the look up buffer (which is a 2-channel float texture)
- I updated threejs a bit, hoping that this would solve some problems (or at least reduce technical debt). however, I'm afraid this causes more work than it solves problems for now.
- Details for the float texture problem:
  - the console is spammed with `GL_INVALID_OPERATION: Level of detail outside of range.` when the look up texture is updated
  - I tried to switch to signed integer textures, but the shader still only finds zeros in the  texture apparently
  - I also tried to use another format than `LuminanceAlpha` for the two-channel format (I tried `RGFormat` which seems to be new to webgl2). not much luck, either, though.
  - I suspect that the implementation of UpdatableTexture could be at fault

philippotto:
Closing in favor of #6748 which implements the basic idea of this PR, too.

PR #6034, opened 2022-02-09
Sensible scalebar values

### LINKED ISSUE #5418: Useful scalebar values
## Detailed Description
Make the scale bar display useful sizes. This implies that the scale bar will get a variable width depending on the exact zoom level. However, it should have a max width. For inspiration: https://github.com/google/neuroglancer/blob/master/src/neuroglancer/widget/scale_bar.ts

## Context
The scale bar has a fixed width which results in fractional values. That makes it hard to use the scale bar for roughly measuring objects in the data.

Dagobert42:
- https://___.webknossos.xyz

### Steps to test:
- open a dataset and zoom around
- minimize window/viewports width until scalebar collapses

### Issues:
- fixes #5418 

------
- [✓] Updated [(unreleased) changelog](../blob/master/CHANGELOG.unreleased.md#unreleased)
- [✓] Ready for review


Dagobert42:
One open issue I could think of is that the scalebar will theoretically start to explode with scales over 10 km. Since we are google earth I assume this is fine for now but I will add a switch to stop the scalebar from expanding infinitely in that case.

frontend/javascripts/oxalis/view/scalebar.js
```diff
@@ -22,34 +22,62 @@ type OwnProps = {|
...
+    if (Math.abs(anchor - mantissa) < Math.abs(bestAnchor - mantissa)) {
+      bestAnchor = anchor;
+    } else {
+      break;
+    }
+  }
+  return bestAnchor * closestPowerOfTen;
+};
+
+const scalebarWidthFactor = 0.3;
```
philippotto, line 50:
```suggestion
// This factor describes how wide the scalebar would ideally be.
// However, this is only a rough guideline, as the actual width is changed
// so that round length values are represented.
const idealScalebarWidthFactor = 0.3;
```

frontend/javascripts/oxalis/view/scalebar.js
```diff
@@ -22,34 +22,62 @@ type OwnProps = {|
...
+const maxScaleBarWidthFactor = 0.45;
+const minWidthToFillScalebar = 130;
+
+function Scalebar({ zoomValue, dataset, viewportWidthInPixels, viewportHeightInPixels }: Props) {
+  const viewportWidthInNm = convertPixelsToNm(viewportWidthInPixels, zoomValue, dataset);
+  const viewportHeightInNm = convertPixelsToNm(viewportHeightInPixels, zoomValue, dataset);
+  const scaledWidthInNm = viewportWidthInNm * scalebarWidthFactor;
+  const scalebarWidthInNm = getBestScalebarAnchorInNm(scaledWidthInNm);
```
philippotto, line 58:
```suggestion
  const idealWidthInNm = viewportWidthInNm * idealScalebarWidthFactor;
  const scalebarWidthInNm = getBestScalebarAnchorInNm(idealWidthInNm);
```

frontend/javascripts/oxalis/view/scalebar.js
```diff
@@ -58,10 +86,11 @@ function Scalebar({ zoomValue, dataset, widthInPixels, heightInPixels }: Props)
...
           right: "1%",
-          // The scalebar should have a width of 25% from the actual viewport (without the borders)
-          width: `calc(25% - ${Math.round(
-            ((2 * OUTER_CSS_BORDER) / constants.VIEWPORT_WIDTH) * 100,
-          )}%)`,
+          width: collapseScalebar
+            ? 16
+            : `calc(${scaleBarWidthFactor * 100}% - ${Math.round(
+                ((2 * OUTER_CSS_BORDER) / constants.VIEWPORT_WIDTH) * 100,
+              )}%)`,
```
philippotto, line 93: I created a "horizontal" tree and tried to make it as large as the scalebar. When using the context menu on the tree and choosing "Path length ...", the length differed slightly. It looks like the written length represents the black box and not the the distance between the white lines.

image

In line 101, a padding of `4` is defined for this div. I suspect, that that padding has to be taken into account here. Can you extract the padding into a variable and add it to the calc expression? I think, the new expression would be:

```suggestion
          width: collapseScalebar
            ? 16
            : `calc(${scaleBarWidthFactor * 100}% - ${Math.round(
                ((2 * OUTER_CSS_BORDER) / constants.VIEWPORT_WIDTH) * 100,
              )}% + ${2 * PADDING}px)`,
```

philippotto (changes requested):
Great stuff! Works well and the code looks solid, too :+1: I left some suggestions and I assume there is one small bug in the calculation of the width (which was probably already present in earlier versions). See my comments :)

frontend/javascripts/libs/format_utils.js
```diff
@@ -59,14 +59,35 @@ export function formatScale(scaleArr: ?Vector3, roundTo?: number = 2): string {
...
-    return `${lengthInNm.toFixed(0)}${ThinSpace}nm`;
-  } else if (lengthInNm < 1000000) {
-    return `${(lengthInNm / 1000).toFixed(1)}${ThinSpace}µm`;
-  } else {
-    return `${(lengthInNm / 1000000).toFixed(1)}${ThinSpace}mm`;
+  let unit = nmFactorToUnit.get(sortedNmFactors[0]);
+  const closestFactor = findClosestLengthUnitFactor(lengthInNm);
+  unit = nmFactorToUnit.get(closestFactor);
+  if (unit == null) {
+    return "error";
```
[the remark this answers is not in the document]
Dagobert42: Ah nice. I thought about Toasting, but we would really only run into this if the sorted units and unit map were out of sync somehow. Which your next suggestion also nicely prevents :)

frontend/javascripts/oxalis/view/scalebar.js
Dagobert42: How much should I use comments in general? I'm trying to express as much as possible in code (via naming, etc.). Maybe we could talk about any conventions regarding this.

frontend/javascripts/oxalis/view/scalebar.js
Dagobert42: Nice catch 👀 Thanks!

philippotto: just fyi: pulling the padding into the term with the outer css border is not equivalent to my suggestion above. Since I still saw the imprecision problem, I just checked it again and the other version indeed works. so, everything should be fine now :)

philippotto (approved):
cool :) looking forward to see this live!

PR #6045, opened 2022-02-14
Include ParamFailure values in error chains

fm3:
ParamFailure is a subclass of Box/Fox/Failure that contains an additional typed value with information about the error. This information should also be included in the Fox error chain.

Also, it should not be erased in `cleanUpOnFailure` in the UploadService.

This way, uploads with linkedLayers but invalid DataSources should yield readable error messages. We use ParamFailures only very rarely. DataSource validation is one, the other is Task creation (where the error is already appended to the response json manually). This is why this never came up before.

- https://paramfailureformat.webknossos.xyz

- [x] Ready for review


jstriebel (approved):
LGTM :+1:

PR #6047, opened 2022-02-14
Fix non-32-aligned bucket requests

fm3:
`toHighestResBoundingBox` method of `BucketPosition` needs to use the bucket-length aligned topleft as its topleft.

This error resulted in buckets not being selected for a cuboid request if they touched the bottomright faces of the layer bounding box.

- https://rawcuboidfix.webknossos.xyz

### Steps to test:
- request non-32-aligned cuboid that borders the boundingbox (bottomright)
- should not skip buckets

----
- [x] Ready for review


jstriebel (approved):
:+1:

Issue #6008, opened 2022-02-02
Ensure layer headers look good even with long names


Dagobert42:
## Context
Layer names exhibit different behaviours when breaking words

![image](https://user-images.githubusercontent.com/46814136/152139359-bbf84df9-2b05-4728-bfa4-b8479284b29f.png)

![Screenshot 2022-02-02 at 10 52 39](https://user-images.githubusercontent.com/46814136/152138266-d7d5a856-7ab5-48d0-bc75-b39f33bcbd41.png)

## Expected Behavior
- layer name is displayed such that it can be copy pasted
- break-word behaviour is consistent across layers
- edit button and info/brush icons are positioned consistently

## Steps to Reproduce the bug
1. open/create a volume annotation and/or color layer with very long names

## Your Environment for bug
- Firefox 96.0.3

Issue #5627, opened 2021-07-20
Long layer names don't wrap properly in sidebar


philippotto:
A long layer name will collide with the reload buttons etc. See screenshot:

![image](https://user-images.githubusercontent.com/2486553/126281155-362674f8-2888-434f-a331-def3918df94f.png)

Long layer name for reproduction: `a-b-c-d-e-f-g-h-i-j-a-b-c-d-e-f-g-h-i-j-a-b-c-d-e-f-g-h-i-j-a-b-c-d-e-f-g-h-i-j`

PR #6050, opened 2022-02-15
Fix clip button changing position

### LINKED ISSUE #5990: Follow-up to auto contrast/brightness

Dagobert42:
- https://___.webknossos.xyz

### Steps to test:
- create a color layer with a very long name
- clip button stays in layer settings header
- use clip button in and out of edit mode
- feedback on optional wiggle room in edit mode appreciated

### Issues:
- fixes #5990 2/5 bug: prevent clip button from moving with very long layer names

------
- [✓] Ready for review


philippotto (changes requested):
Great! The code looks good to me :+1: Could you please also fine-tune the layout of all layer-icons (except for the "info" icon which sticks to the layer name, anyway) so that it looks okay when the layer name overflows?

I created a rough layout which looks like this:
image

So, the name may overflow, but the icons are always on the right-side in a separate box.

The layout code looks like this:

```js
Example 1:
<div>
  <div class="flex-container">
    <div class="flex-item">long name 1234 1234 1234 1234 1234 1234 1234 1234 1234 1234 1234 1234 1234 1234 1234 1234 1234 1234 1234 1234 1234 1234 1234 1234 1234 1234 1234 1234 1234 1234 1234 1234 1234 1234 1234 1234 1234 1234 1234 1234 1234 1234 1234 1234 1234 1234 xyz</div>
    <div class="flex-item icons">
      <span>icon 1</span>
      <span>icon 2</span> 
    </div>
  </ul>
</div>

Example 2:
<div>
  <div class="flex-container">
    <div class="flex-item">not so long name </div>
    <div class="flex-item icons">
      <span>icon 1</span>
      <span>icon 2</span> 
    </div>
  </ul>
</div>
```

And the corresponding CSS is:
```css
.flex-container {
  justify-content: space-between;
  display: flex;

  padding: 0;
  margin: 0;
  border: 1px solid silver;
}

.flex-item {
  display: inline-block;
  flex-grow: 0;
  word-break: break-all;

  background: tomato;
  padding: 5px;
  margin: 10px;
  line-height: 100px;
  color: white;
  font-size: 2em;
}

.icons {
  flex-shrink: 0;
}
```

Thank you!

Dagobert42:
Sure :) I had actually opened #6008 for that and I think #5627 would be fixed with that as well so I'd open a new PR for this.

philippotto (approved):
Ok, sure, then let's tackle this in another PR :)

PR #6051, opened 2022-02-15
Time tracking: Do not fail on empty timestamps list

fm3:
Hotfix for head-of-empty-list runtime esception in timetracking.
My assumption is that the recent frontend changes cause empty tracing save requests, resulting in an update report with an empty list of timestamps. This should not cause an exception on the wk side though.

- [x] Ready for review


philippotto (approved):
👍

PR #6052, opened 2022-02-16
Guard against empty saves

philippotto:
This fixes a regression introduced by https://github.com/scalableminds/webknossos/pull/5999.
See these line changes: https://github.com/scalableminds/webknossos/commit/3831806188f218aea876a1e76181b462bc652a44#diff-acc1d929544bdde3687f0e79f303c4f23b61ca97fafe5e4579202972d2e37acaR673

- https://___.webknossos.xyz

### Steps to test:
- abc

### Issues:
- follow-up for #5999 (also see https://github.com/scalableminds/webknossos/pull/6051 which fixes the symptom in the back-end)

------
- [X] Ready for review

### DIFF
--- a/frontend/javascripts/oxalis/model/sagas/save_saga.js
+++ b/frontend/javascripts/oxalis/model/sagas/save_saga.js
@@ -678,12 +678,15 @@ export function* pushTracingTypeAsync(
         }
       }
     } else {
-      // Saving the tracing automatically (via timeout) only saves the current state.
-      // It does not require to reach an empty saveQueue. This is especially
-      // important when the auto-saving happens during continuous movements.
-      // Always draining the save queue completely would mean that save
-      // requests are sent as long as the user moves.
-      yield* call(sendRequestToServer, tracingType, tracingId);
+      saveQueue = yield* select(state => selectQueue(state, tracingType, tracingId));
+      if (saveQueue.length > 0) {
+        // Saving the tracing automatically (via timeout) only saves the current state.
+        // It does not require to reach an empty saveQueue. This is especially
+        // important when the auto-saving happens during continuous movements.
+        // Always draining the save queue completely would mean that save
+        // requests are sent as long as the user moves.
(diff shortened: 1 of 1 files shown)


frontend/javascripts/oxalis/model/sagas/save_saga.js
```diff
@@ -678,12 +678,15 @@ export function* pushTracingTypeAsync(
...
       }
     } else {
-      // Saving the tracing automatically (via timeout) only saves the current state.
-      // It does not require to reach an empty saveQueue. This is especially
-      // important when the auto-saving happens during continuous movements.
-      // Always draining the save queue completely would mean that save
-      // requests are sent as long as the user moves.
-      yield* call(sendRequestToServer, tracingType, tracingId);
+      saveQueue = yield* select(state => selectQueue(state, tracingType, tracingId));
+      if (saveQueue.length > 0) {
```
philippotto, line 682: I don't really know how this scenario could be triggered with a single volume layer, since line 663 explicitly waits for a push-action which should increase the saveQueue. However, the old code (before https://github.com/scalableminds/webknossos/commit/3831806188f218aea876a1e76181b462bc652a44#diff-acc1d929544bdde3687f0e79f303c4f23b61ca97fafe5e4579202972d2e37acaR673) also always re-checked for length > 0, so this should be equivalent now.

In case of multiple volume layers, I see how the bug could be easily triggered, since line 663 doesn't listen to the specific tracing id.

philippotto: I temporarily added

```
else {
  console.log("superfluous save attempt");
}
```

which was indeed printed. So, the code change seems to be necessary indeed.

daniel-wer (approved):
LGTM :+1:

frontend/javascripts/oxalis/model/sagas/save_saga.js
philippotto: Ah, yes, that makes sense :+1:

PR #6055, opened 2022-02-17
Dynamic bucket allocation & miscellaneous volume improvements

philippotto:
- increase speed of compression for volume buckets by using two workers instead of one for that job (adds a new `WorkerPool` abstraction for that)
- adds a tooltip (only visible on hover) to the save button in case there are currently buckets waiting for compression (when annotating large amounts the compression can take some seconds which might irritate the user since the progress was not really visible)
- adapts the bucket allocation logic so that there is no hard limit anymore. instead, there is an initial limit which can be exceeded if necessary. if the limit is exceeded by a factor of 2 (this corresponds to the hotfix of bumping the limit to 10000 for segmentation layers, the user is warned about that and advised to save&reload). I didn't built a shrinking logic, since I don't consider the advantages to be worth it. Happy to discuss this, though!
- fixes that `PushQueue.push` could starve when the user never does a break longer than 1 second (unrealistic, but when testing, one can easily run into this)
- refactor: rename some important variables (such as `queue`) and add some comments

- https://dynamicbucketallocation.webknossos.xyz

### Steps to test:
- I adapted the code to output some periodic logging about the bucket counts and use a lower maximum bucket limit, so that the critical cases can be tested easier
- open a new annotation and open the devtools
- brush large areas over multiple viewports to quickly build up a load of changed buckets
- shortly before it reaches 500, stop brushing and hit save
- hovering over the save-button should show some information about how many buckets need to be compressed
- when everything is saved, brush again (in moderate speed)
- ideally, the allocated bucket limit should not exceed 500
- now, brush very fast across all viewports and jump into new regions to force wk to allocate new buckets
- --> the bucket count should grow over 500
- continue so that the bucket count grows over 1000 --> a toast should appear
- save & wait
- reload the page --> everything should look as before

image

image

### Issues:
- contributes to #5973 

------
- [x] Ready for review


frontend/javascripts/oxalis/workers/fetch_buffer_with_headers.worker.js
```diff
@@ -17,10 +17,13 @@ function fetchBufferWithHeaders(
       for (const [key, value] of headers.entries()) {
         headerObject[key] = value;
       }
-      return {
-        buffer,
-        headers: headerObject,
-      };
+      return transfer(
```
philippotto, line 20: Strictly speaking this change should enable the browser to not copy the data between worker and main loop, but instead move it. However, I couldn't really measure a difference (I measured the e2e loading time of opening a dataset and waiting for all buckets, but maybe there are better ways). Anyway, using transferables is more of a separate issue on its own I think.

frontend/javascripts/oxalis/view/action-bar/save_button.js
```diff
@@ -107,11 +125,19 @@ class SaveButton extends React.PureComponent<Props, State> {
...
       >
-        {this.shouldShowProgress() ? (
-          <span style={{ marginLeft: 8 }}>{Math.floor((progressFraction || 0) * 100)} %</span>
-        ) : (
-          <span className="hide-on-small-screen">Save</span>
-        )}
+        <Tooltip
+          title={
+            totalBucketsToCompress > 0
+              ? `${totalBucketsToCompress} buckets remaining to compress...`
```
philippotto, line 131: Ideally, this tooltip wouldn't use the word "bucket", since we cannot expect the user to know this term. However, I don't have a better idea. One could convert to `MB`, but I'm not really fond of this idea.

Also cc @fm3 :)

frontend/javascripts/oxalis/model/bucket_data_handling/data_cube.js
```diff
@@ -108,13 +120,42 @@ class DataCube {
...
+          }
+        }
+        if (weirdDirtyState > 0) {
+          console.log("weirdDirtyState", weirdDirtyState);
+        }
+
+        console.log("dirtyBucketsCount", dirtyBucketsCount);
+        console.log("this.buckets.length", this.buckets.length);
+      }, 2000);
+    }
```
philippotto, line 149: can be removed after testing.

frontend/javascripts/oxalis/model/bucket_data_handling/data_cube.js
```diff
@@ -108,13 +120,42 @@ class DataCube {
...
+
     _.extend(this, BackboneEvents);
 
     this.cubes = [];
-    if (isSegmentation) {
-      this.MAXIMUM_BUCKET_COUNT *= 2;
+    if (isSegmentation && process.env.BABEL_ENV !== "test") {
+      // undo
+      this.BUCKET_COUNT_SOFT_LIMIT = Math.floor(this.BUCKET_COUNT_SOFT_LIMIT / 10);
```
philippotto, line 156: can be removed after testing.

frontend/javascripts/oxalis/model/bucket_data_handling/data_cube.js
```diff
@@ -648,10 +693,7 @@ class DataCube {
     if (bucket.type === "null") {
       return;
     }
-    bucket.setData(data);
-    bucket.pendingOperations = newPendingOperations;
-
-    this.pushQueue.insert(bucket);
```
philippotto, line 654: the changes here are merely refactoring.

frontend/javascripts/oxalis/workers/fetch_buffer_with_headers.worker.js
philippotto: I opened https://github.com/scalableminds/webknossos/issues/6064.

frontend/javascripts/oxalis/model/bucket_data_handling/data_cube.js
```diff
@@ -280,44 +321,50 @@ class DataCube {
...
-          elementClass: this.elementClass,
-          isSegmentation: this.isSegmentation,
-          resolutionInfo: this.resolutionInfo,
-        });
-      }
 
-      this.collectBucket(this.buckets[this.bucketIterator]);
-      this.bucketCount--;
-    }
+        if (this.buckets.length > (2 * constants.MAXIMUM_BUCKET_COUNT_PER_LAYER) / 10) {
```
daniel-wer, line 354:
```suggestion
        if (this.buckets.length > 2 * this.BUCKET_COUNT_SOFT_LIMIT) {
```

frontend/javascripts/oxalis/view/action-bar/save_button.js
daniel-wer: You could simply refer to them as items. This would be rather vague but still get the message across that things are happening (or need to happen) :)

frontend/javascripts/oxalis/workers/fetch_buffer_with_headers.worker.js
daniel-wer: It's probably not immediately faster, because moving takes the same time as copying, but it might pay out in reduced GC time, instead (which is harder to measure and correlate). Just a thought, I haven't read up on it.

daniel-wer (changes requested):
Great changes, code LGTM :+1: 

During testing I encountered some issues. I'm not sure whether they are related to the temporary changes to ease testing, but I would think not. Maybe buckets get collected, although they shouldn't be?

- Something seems to be fishy with the data loading. When in mag 1 some buckets are not loaded/shown in mag 1, see for example the right side of the XY viewport here (loading had finished):
loaded_but_not_mag_1

- The loading pattern seems spotty and slower than before.
Before:
loading before

  With this PR:
loading after

- Also, I wasn't able to get a tooltip for the save button yet. I tried hovering before clicking the save button or afterwards. Is this very timing-dependent?

philippotto:
Thanks for testing and raising these oddities :+1: I guess, I got a little bit betriebsblind and didn't notice this myself.

> * Something seems to be fishy with the data loading. When in mag 1 some buckets are not loaded/shown in mag 1, see for example the right side of the XY viewport here (loading had finished):

I finally found the issue for this and fixed it in https://github.com/scalableminds/webknossos/pull/6055/commits/5f2f7bf2cfacad09d2a91b04c3ed99d7ae67cfb0. The code which marked a bucket as active happened to late which allowed the bucket collector to collect a bucket while allocating another one in `consumeBucketsFromArrayBuffer`. Through the artificial low bucket limit, this issue became present + frequent, but I assume this fixes also the rare disappearing buckets I saw from time to time in prod :tada: 

> * The loading pattern seems spotty and slower than before.

The spottiness should be solved with the above bug fix, too. The slowness is probably a consequence of the low bucket limit which makes GC more frequent (and this slows wk down). Due to the fix from above, wk recognizes that it's necessary to exceed the bucket limit, anyway (otherwise, not everything would be rendered). Therefore, garbage-collections and repeated re-downloads are now avoided.

Could you please test again to see whether both issues are resolved now? Afterwards, I'll lift the artificial limits and we can do another testing round (however, I'm assuming that to be way more robust). I just deployed the newest version: https://dynamicbucketallocation.webknossos.xyz/

> * Also, I wasn't able to get a tooltip for the save button yet. I tried hovering before clicking the save button or afterwards. Is this very timing-dependent?

Yes, kind of. One needs a really big brush size (i.e., 300) and draw quickly (2 seconds should be enough). Then, hovering over the save button will briefly show the tooltip until all buckets are compressed. Depending on how many buckets were labeled, the tooltip might show progress for several seconds.

frontend/javascripts/oxalis/workers/fetch_buffer_with_headers.worker.js
philippotto: I think, the main reason for `transfer` to exist is performance. However, there seems to be dependence on the size of the transferred data. [This article](https://joji.me/en-us/blog/performance-issue-of-using-massive-transferable-objects-in-web-worker/) suggests that small elements don't really benefit from being transferred (even the opposite, it can get worse). This sparked my curiosity and I did a benchmark with a dummy webworker which consumes an array buffer (size of one bucket in uint32) and returns it again. For 1000 runs, this shows:

```
no transfers: 173.39208984375 ms
only with input transfer: 112.991943359375 ms
only with output transfer: 142.373046875 ms
with all transfers: 94.93017578125 ms
```

So, the transfer seems to a have a positive effect when micro-benchmarking. So, I'd keep the change in and hope that this also has some positive effects (even when subtle).

daniel-wer (commented):
> I finally found the issue for this and fixed it in https://github.com/scalableminds/webknossos/commit/5f2f7bf2cfacad09d2a91b04c3ed99d7ae67cfb0. The code which marked a bucket as active happened to late which allowed the bucket collector to collect a bucket while allocating another one in consumeBucketsFromArrayBuffer. Through the artificial low bucket limit, this issue became present + frequent, but I assume this fixes also the rare disappearing buckets I saw from time to time in prod tada

Ah nice find! Good thing you introduced these strict limits for testing which made the issue much more apparent :+1:

> Could you please test again to see whether both issues are resolved now? (for me it looks good, but I didn't really notice a big speed difference before, either).

Both issues are solved for me and also was able to get a quick glance at the save button tooltip :+1:

frontend/javascripts/oxalis/workers/fetch_buffer_with_headers.worker.js
daniel-wer: Sounds good! #MeasureToMakeDecisions

daniel-wer:
One thing that happened during testing when I was brushing with a large brush size was that the sagas crashed with "Maximum call stack size exceeded". Not sure if that's known. I don't think it has high priority since I brushed unrealistically fast using the maximum brush size. Still I wanted to mention it, stack trace attached.

[call_stack_exceeded_stack_trace.txt](https://github.com/scalableminds/webknossos/files/8119011/call_stack_exceeded_stack_trace.txt)

philippotto:
Thanks for testing! I now cleaned up the debugging code and think this PR is ready to merge if you agree :)

> One thing that happened during testing when I was brushing with a large brush size was that the sagas crashed with "Maximum call stack size exceeded". Not sure if that's known. I don't think it has high priority since I brushed unrealistically fast using the maximum brush size. Still I wanted to mention it, stack trace attached.
> 
> [call_stack_exceeded_stack_trace.txt](https://github.com/scalableminds/webknossos/files/8119011/call_stack_exceeded_stack_trace.txt)

Yes, I also noticed this error from time to time (only in unrealistic annotation scenarios and already before this PR) and also spent some time debugging it some time ago (unfortunately, without much success). Interestingly, your stack trace looks way bigger than mine (I "only" had 200 function calls on the stack) whereas you have more than 10000. This gave me the idea that redux-saga might do some sort of tail recursion which leads to this problem. I think I've got a fix for this, but I'd like to do this in a separate PR to unblock the current one.

frontend/javascripts/oxalis/model/bucket_data_handling/data_cube.js
```diff
@@ -351,7 +317,7 @@ class DataCube {
           });
         }
 
-        if (this.buckets.length > (2 * constants.MAXIMUM_BUCKET_COUNT_PER_LAYER) / 10) {
+        if (this.buckets.length > 2 * constants.MAXIMUM_BUCKET_COUNT_PER_LAYER) {
```
daniel-wer, line 320:
```suggestion
        if (this.buckets.length > 2 * this.BUCKET_COUNT_SOFT_LIMIT) {
```

Did you deliberately decide against this suggestion? I know the two are the same, but the documentation at the top of the file only talks about `BUCKET_COUNT_SOFT_LIMIT` and it is also used a few lines above. Using it here as well is less confusing in my opinion :)

daniel-wer (approved):
LGTM, see my one small suggestion.

Let's do a quick round of testing once the CI ran through and then this should be ready to be merged :+1:

frontend/javascripts/oxalis/model/bucket_data_handling/data_cube.js
philippotto: Oh, sorry, no, that wasn't my intention. I thought the suggestion only aimed to remove the `/ 10`. When I cleaned up the debugging code, I did this manually. Changed it now :)

philippotto:
> LGTM, see my one small suggestion.
> 
> Let's do a quick round of testing once the CI ran through and then this should be ready to be merged +1

Sounds great! I just tested a bit on https://dynamicbucketallocation.webknossos.xyz. Feel free to do so, too, and let me know whether I should merge :)

PR #6056, opened 2022-02-17
Prevent errors thrown by keyboard shortcuts used int the 3D viewport

### LINKED ISSUE #5975: Some shortcuts trigger an error toast in the 3D viewport
For example, pressing d or f in the 3D viewport will cause a toast to appear. The relevant code should check whether the 3D viewport is active an then do an early-out (i.e., do nothing instead of raising an error).

Potential relevant shortcuts:
- d
- f
- space
- v

Note that other shortcuts, such as i, o or `.` should still work. Have a look at the keyboard documentation for reference.

MichaelBuessemeyer:
This PR should fix all error currently thrown by shortcuts triggered in the 3D viewport that are not meant for the 3D viewport.
Examples for these short cuts are: d, f, space, v.

- https://lesserrorsindviewport.webknossos.xyz

### Steps to test:
- Open a hybrid tracing in wk.
- Use the shortcuts
  - d
  - f
  - space
  - space + shift
In the 3d viewport. They should do nothing and should not throw an error.
- Test the listed shortcuts in one of the other viewports and make sure they do work.
- Please also test whether other shortcuts trigger an error or unwanted side effect when used in the 3d viewport. I tried to catch every keyboard shortcut.

### Issues:
- fixes #5975

------
- [x] Ready for review


frontend/javascripts/oxalis/controller/viewmodes/plane_controller.js
```diff
@@ -133,12 +133,12 @@ class VolumeKeybindings {
...
-      v: () => {
+      v: guardNotTDViewport(() => {
         Store.dispatch(copySegmentationLayerAction());
-      },
-      "shift + v": () => {
+      }),
+      "shift + v": guardNotTDViewport(() => {
         Store.dispatch(copySegmentationLayerAction(true));
-      },
+      }),
```
daniel-wer, line 141: Why is this needed? As far as I can see, the corresponding `copySegmentationLayer` saga has

```
  const activeViewport = yield* select(state => state.viewModeData.plane.activeViewport);
  if (activeViewport === "TDView") {
    // Cannot copy labels from 3D view
    return;
  }
```

in its first lines.
Also, I cannot provoke any error triggering this functionality in the 3D viewport on master.

MichaelBuessemeyer: > Also, I cannot provoke any error triggering this functionality in the 3D viewport on master.

Thanks for noticing :D. Looks like a messed up the testing I did on my own 🙈

daniel-wer (approved):
Works well, thanks for fixing! :+1:

PR #6065, opened 2022-02-21
Compile tests to different dir to avoid unnecessary recompilations

philippotto:
- https://___.webknossos.xyz

### Steps to test:
- Before this PR, I had the problem that the sbt output emited `[info] Slick SQL schema already up to date.` multiple times when I ran `yarn test-prepare-watch` (which meant that it recompiled).
- now, the issue does not exist, anymore
- CI should be enough

### Issues:
- fixes unnecessary recompilations

------
- [X] Ready for review


daniel-wer (approved):
Sweet, thanks for taking the time to fix this! :+1:

PR #6076, opened 2022-02-23
Fix rare callstack overflow when annotating large areas

### LINKED ISSUE #6074: Maximum stack size exceeded when brushing large areas very fast
## Context
Volume Annotation.

## Current Behavior
Crashes with maximum call stack size exceeded error.

Also see https://github.com/scalableminds/webknossos/pull/6055#issuecomment-1048057756

## Steps to Reproduce the bug

Unfortunately, the repro is not 100% reliable:
1. Open new volume annotation
2. Brush large areas very fast and click save in between

## Your Environment for bug

philippotto:
I suspect that redux-saga uses too much recursion as suggested here https://github.com/redux-saga/redux-saga/issues/605. Let's say whether the workaround really helps.

- https://fixstackoverflow.webknossos.xyz

### Steps to test:
- brush a lot very fast
- saga shouldn't crash

### Issues:
- fixes #6074

------
- [X] Ready for review


frontend/javascripts/oxalis/model/sagas/save_saga.js
```diff
@@ -224,7 +224,12 @@ export function* collectUndoStates(): Saga<void> {
     "REDO",
   ]);
 
+  window.counter = 0;
   while (true) {
+    window.counter++;
+    if (window.counter % 100 === 0) {
+      yield _delay(0);
+    }
```
philippotto, line 232:
```suggestion
  let loopCounter = 0;
  while (true) {
    loopCounter++;
    if (loopCounter % 100 === 0) {
      yield _delay(0);
    }
```

daniel-wer (commented):
I wasn't able to provoke a call stack overflow on this branch. However, I also wasn't able to produce one on master :see_no_evil: 
Are you still able to reproduce the original error on the deployed master or was this possibly fixed in an unrelated PR in the mean time?

PR #6079, opened 2022-02-24
Make clip button indicate loading

### LINKED ISSUE #5990: Follow-up to auto contrast/brightness

Dagobert42:
- https://___.webknossos.xyz

### Steps to test:
- use reload button on color layer
- use clip histogram button
- may only be visible at slow connection speed

### Issues:
- fixes Follow-up to auto contrast/brightness 3/5

------
- [✓] Ready for review


Dagobert42:
Unfortunately I'm experiecing problems with my localhost again, so I pushed the changes we discussed without actually testing. I hope everything works as expected.

philippotto (approved):
Great! Looks good to me :)

PR #6086, opened 2022-02-28
Add frontend and backend part for job to materialize a volume annotation

### LINKED ISSUE #4603: Allow applying result of merger mode tracing
There's already a script to apply the skeleton tracing on the segmentation. This could be integrated into webknossos-worker.

MichaelBuessemeyer:
This PR adds the frontend and backend part for the job that applies the results of a merger mode tracing to a segmentation layer.

Open TODOs:
- [x] Properly support and test the support for tracings with multiple segmentation layers.
- [x] Add a preview image for the apply merger mode job
- [x] Test selecting from multiple segmentation layers and whether the worker also chooses the correct segmentation layer and applies the segment merging to it.
- [x] See my mentions below :)
 
- https://___.webknossos.xyz

### Steps to test:
- As this pr is closely connected to the [counter part pr](https://github.com/scalableminds/webknossos-worker/pull/95) in the worker's repo, the testing instructions can be taken from the wk-worker pr as they would be the same for this pr.

### Issues:
- fixes #4603

------
- [x] Updated [documentation](../blob/master/docs) if applicable -> Add a small sentence + image on where to trigger the job
- ~~[ ] Adapted [wk-libs python client](https://github.com/scalableminds/webknossos-libs/tree/master/webknossos/webknossos/client) if relevant API parts change~~
- [X] Ready for review


app/models/job/Job.scala
```diff
@@ -72,7 +72,7 @@ case class Job(
           }
         case "export_tiff" =>
           Some(s"$dataStorePublicUrl/data/exports/${_id.id}/download")
-        case "infer_nuclei" | "infer_neurons" =>
+        case "infer_nuclei" | "infer_neurons" | "apply_merger_mode" =>
```
MichaelBuessemeyer, line 75: I am not sure, whether this is everything I need to do here for the new job. I did not test the jobs list view and therefore this is likely untested.

frontend/javascripts/oxalis/view/action-bar/toolbar_view.js
```diff
@@ -222,6 +224,18 @@ function AdditionalSkeletonModesButtons() {
...
+          <ButtonComponent
+            style={narrowButtonStyle}
+            onClick={() => setShowApplyMergerModeModal(true)}
+            faIcon="fa fa-cogs"
+          />
+        </Tooltip>
+      )}
+      {showApplyMergerModeModal && (
+        <ApplyMergerModeModal handleClose={() => setShowApplyMergerModeModal(false)} />
+      )}
```
MichaelBuessemeyer, line 238: I am not totally convinced that this is the best place where to put this modal. As I did not put much thought into it there might be a better place somewhere in the code :)

frontend/javascripts/oxalis/view/right-border-tabs/starting_job_modals.js
```diff
@@ -28,23 +32,26 @@ type StartingJobModalProps = {
   jobName: string,
   description: Node,
   isBoundingBoxConfigurable?: boolean,
+  chooseSegmentationLayer?: boolean,
 };
 
 function StartingJobModal(props: StartingJobModalProps) {
   const isBoundingBoxConfigurable = props.isBoundingBoxConfigurable || false;
+  const chooseSegmentationLayer = props.chooseSegmentationLayer || false;
```
MichaelBuessemeyer, line 40: The main changes here and below are that I enabled `StartingJobModal` to not only select color layers but also segmentation layers when `chooseSegmentationLayer=true` is passed to the component. The changes are mainly removing "color" from that variable names as the layers the component now works with might also be of kind segmentation.

frontend/javascripts/oxalis/view/right-border-tabs/starting_job_modals.js
```diff
@@ -198,7 +205,7 @@ function StartingJobModal(props: StartingJobModalProps) {
 }
 
 export function NucleiInferralModal({ handleClose }: Props) {
-  const dataset = useSelector((state: OxalisState) => state.dataset);
+  const dataset = useSelector(state => state.dataset);
```
MichaelBuessemeyer, line 208: These type annotations are not needed. At least according to my local flow checker

MichaelBuessemeyer (commented):
These are all the comments that seemed helpful to me. If you got questions about the code, just let me know.

frontend/javascripts/oxalis/view/action-bar/toolbar_view.js
philippotto: I think, it's fine. Being close to the button which opens it, creates some coherence :+1:

frontend/javascripts/oxalis/view/action-bar/toolbar_view.js
```diff
@@ -114,15 +115,24 @@ const RadioButtonWithTooltip = ({
...
+  onClick?: Function,
 }) => (
   <Tooltip title={disabled ? disabledTitle : title}>
     {/* $FlowIgnore[cannot-spread-inexact] */}
-    <Radio.Button disabled={disabled} {...props} />
+    <Radio.Button
+      disabled={disabled}
+      {...props}
+      onClick={evt => {
+        if (document.activeElement) document.activeElement.blur();
```
philippotto, line 132: :+1:

frontend/javascripts/oxalis/view/right-border-tabs/starting_job_modals.js
```diff
@@ -85,34 +91,43 @@ function StartingJobModal(props: StartingJobModalProps) {
...
+  const LayerSelectionFromItem = (): Node => {
+    const layerType = chooseSegmentationLayer ? "segmentation layer" : "color layer";
+    return (
+      <Form.Item
+        label={layerType}
+        name="layerName"
+        rules={[
+          {
+            required: true,
+            message: `Please the ${layerType} that should be used for this job.`,
```
philippotto, line 103:
```suggestion
            message: `Please select the ${layerType} that should be used for this job.`,
```

probably?

frontend/javascripts/oxalis/view/right-border-tabs/starting_job_modals.js
```diff
@@ -268,3 +310,78 @@ export function NeuronInferralModal({ handleClose }: Props) {
...
+          baseSegmentationName,
+          volumeLayerName,
+          newDatasetName,
+          tracing.annotationId,
+          tracing.annotationType,
+        );
+      }}
+      description={
+        <p>
+          Start a job that take the current state of this merger mode tracing and apply it to the
```
philippotto, line 341:
```suggestion
          Start a job that takes the current state of this merger mode tracing and applies it to the
```

frontend/javascripts/oxalis/view/right-border-tabs/starting_job_modals.js
```diff
@@ -268,3 +310,78 @@ export function NeuronInferralModal({ handleClose }: Props) {
...
+          segmentation layer. This will create a new dataset which contains the merged segmentation
+          layer. If this dataset has more than one segmentation layer, please select the
+          segmentation layer to the merging should be applied to.
```
philippotto, line 344:
```suggestion
          segmentation layer to which the merging should be applied to.
```

philippotto: on a second thought: when the merger mode is active, there can only be one visible segmentation layer. wouldn't simply using that layer make sense? one could still show the layer name in the modal (but as a disabled field) to inform the user. but making the user choose the layer again seems unnecessary to me?

philippotto (commented):
Great stuff! Didn't test yet, but the code looks good overall :+1: Mainly left feedback for naming and maybe usability.

frontend/javascripts/oxalis/view/right-border-tabs/starting_job_modals.js
MichaelBuessemeyer: > when the merger mode is active, there can only be one visible segmentation layer.

In general, there con only be a single visible segmentation layer at a time. But it is possible to switch between segmentation layers while the merger mode is on.

Nevertheless, I applied you feedback. But I am still unsure whether to keep the value for the selected segmentation layer fixed. Because what if the user changes the visible segmentation layer before trying to trigger the merger mode job. Then the user would be unable to select the annotation layer they actually want the merging to be applied to. I know this is a pretty hypothetical scenario, but it is currently still possible and at least worth considering imo.

frontend/javascripts/admin/admin_rest_api.js
```diff
@@ -1060,30 +1061,79 @@ export function startNeuronInferralJob(
...
+  fallbackLayerName: string,
+  volumeLayerName: ?string,
+  newDatasetName: string,
   annotationId: string,
   annotationType: APIAnnotationType,
 ): Promise<APIJob> {
+  const volumeLayerArg = volumeLayerName != null ? `&volumeLayerName=${volumeLayerName}` : "";
   return Request.receiveJSON(
-    `/api/jobs/run/globalizeFloodfills/${organizationName}/${datasetName}?newDataSetName=${newDataSetName}&layerName=${layerName}&annotationId=${annotationId}&annotationType=${annotationType}`,
+    `/api/jobs/run/${jobURLPath}/${organizationName}/${datasetName}?fallbackLayerName=${fallbackLayerName}${volumeLayerArg}&annotationId=${annotationId}&annotationType=${annotationType}&newDatasetName=${newDatasetName}`,
```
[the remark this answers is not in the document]
philippotto: > The URL API is not supported by Internet Explorer. But I think we do not support IE anyway, right?

Yeah, that ship has sailed luckily :)

frontend/javascripts/oxalis/view/right-border-tabs/starting_job_modals.js
philippotto: > Because what if the user changes the visible segmentation layer before trying to trigger the merger mode job. Then the user would be unable to select the annotation layer they actually want the merging to be applied to. I know this is a pretty hypothetical scenario, but it is currently still possible and at least worth considering imo.

As long as it's transparent, on which layer the job will work, I think this is fine. The user can simply close the modal, change the visibility and open it again.

MichaelBuessemeyer: Alrighty :D
In that case, all your requested changes should be applied

philippotto:
Great stuff! Testing went mostly smooth, but I noticed a few things:

- unfortunately, the downsampled segmentation layer did not match the existing color layer mags. this caused the `This dataset contains multiple layers which differ in their resolution. Please convert the layers to make their resolutions match.` error for me.
  - I ran the job on l4 sample where the color layer has `[1, 1, 1], [2, 2, 1], [4, 4, 2], [8, 8, 4], [16, 16, 8]`, but the new seg layer got `[1, 1, 1], [2, 2, 1], [4, 4, 1], ...`.
  - I see two options for fixing this:
    - call `downsample_mag_list` and pass the mags which existed for the old segmentation layer
    - or use this code from vx: https://github.com/scalableminds/voxelytics/blob/024839faa2d564f48d3dbdabf1a648389e77486c/voxelytics/commons/downsampling_helper.py#L152-L152 (ideally, this should be moved into wk-libs)
  - the first option is easier. the second would be nicer for the long-term. but I'd rather recommend to use the first one.
- no "View" link in the job page (I think you need to adapt the `renderActions` method)
- the color layer could not be rendered correctly (see https://github.com/scalableminds/webknossos-worker/pull/95/files/742c9273deb45836e0709f16f81a33138d16e911#r843730822)
- I'd change the icon for the "apply merger mode" button to `<ExportOutlined />`

MichaelBuessemeyer:
@fm3 I think I need some help in this pr because I changed the long-running jobs routes a little. Do you think bumping the API version is needed here or can we somehow circumvent this? 

Also: wklibs does not contain any methods that work with the backend long-running jobs routes, or does it? IF it does, wklibs might also need some slight adjustments.
> - [ ] Adapted [wk-libs python client](https://github.com/scalableminds/webknossos-libs/tree/master/webknossos/webknossos/client) if relevant API parts change

fm3:
As far as I know the `/jobs/run` routes are currently only called by the javascript front-end. Because of this, I’d say we do not need to provide backwards compatibility / bump the api version or adapt the python client

docs/volume_annotation.md
```diff
@@ -48,6 +48,11 @@ To fix split errors see section on flood filling below.
 
 ![Video: Proof-Reading Volume Annotations](https://www.youtube.com/watch?v=Sq4AuWanK14)
 
+After finishing the proof-reading a long-running job can be started to apply the combining of segments into a new dataset with the same layers. The job can be started via a modal. The modal can be open by clicking on the button next to the merger mode button.
```
philippotto, line 51:
```suggestion
After finishing the proof-reading, a long-running job can be started to apply the merging of segments into a new dataset with the same layers. The job can be started via a modal. The modal can be open by clicking on the button next to the merger mode button.
```

frontend/javascripts/admin/job/job_list_view.js
```diff
@@ -180,7 +180,7 @@ class JobListView extends React.PureComponent<Props, State> {
     ) {
       return (
         <span>
-          Applied merger mode to layer {job.layerName} of{" "}
+          Applying merger mode to layer {job.layerName} of{" "}
```
philippotto: why "applying"? because it's misleading if it's still processing? I find "Applied" better since it definitely works in the "done" state and also kinda works in the "in-progress" state.

philippotto (changes requested):
The front-end looks good to me now :) Only see my two minor comments. Also note:

- [ ] the "apply merger mode" button is visible even when jobs are disabled

frontend/javascripts/admin/job/job_list_view.js
MichaelBuessemeyer: > because it's misleading if it's still processing?

Jep that was my thought.
> I find "Applied" better since it definitely works in the "done" state and also kinda works in the "in-progress" state.

sure, I got no strong opinion on this.

frontend/javascripts/oxalis/view/right-border-tabs/starting_job_modals.tsx
```diff
@@ -1,78 +1,268 @@
...
+          },
+        },
+      ]}
+      hidden={!hasOutputSegmentationLayer}
+    >
+      <Input />
+    </Form.Item>
+  );
+}
+
```
MichaelBuessemeyer, line 222: This new `Form.Item` is new and wasn't reviewed by @philippotto. Could you please do that @daniel-wer?

frontend/javascripts/oxalis/view/right-border-tabs/starting_job_modals.tsx
```diff
@@ -97,148 +287,69 @@ function StartingJobModal(props: StartingJobModalProps) {
...
-      </React.Fragment>
-    ) : null;
-  }
-
-  const hasUnselectedOptions =
-    selectedColorLayerName == null || (isBoundingBoxConfigurable && selectedBoundingBox == null);
+  initialOutputSegmentationLayerName = `${
+    initialOutputSegmentationLayerName || "segmentation"
+  }_corrected`;
+  // TODO: Other jobs also have an output segmentation layer. The names for these jobs should also be configurable.
```
MichaelBuessemeyer, line 299: This TODO needs to be kept, as the new `OutputSegmentationLayerNameFormItem` should also be used for the inferral jobs. I did not do it in this pr to avoid making this pr even bigger.

frontend/javascripts/oxalis/view/right-border-tabs/starting_job_modals.tsx
```diff
@@ -315,3 +433,111 @@ export function NeuronInferralModal({ handleClose }: Props) {
...
+  );
+  if (tracing.volumes.length === 0) {
+    description = (
+      <p>
+        Start a job that takes the current state of this merger mode tracing and materializes it
+        into a new dataset. In the new output dataset the segments connected via skeleton nodes will
+        be merged. Please enter the name of the output dataset and the output segmentation layer.
+      </p>
+    );
+  }
```
MichaelBuessemeyer, line 472: It would be nice if you could check whether this text makes sense to you @daniel-wer.

The reassignment of `description` in the if only happens when there is no volume annotation (and therefore the annotation is skeleton-only).

> `` "${fixedSelectedLayer && getReadableNameOfVolumeLayer(fixedSelectedLayer, tracing)}" ``

Here the `fixedSelectedLayer` is needed because `fixedSelectedLayer` is potentially undefined if there is no volume annotation layer. That's why I added the reassignment of `description` in the if statement to cover this case and give a better text to the user that matches the context.

MichaelBuessemeyer (commented):
@daniel-wer I just added comments to the code parts I'd like some feedback on before merging. Could you please do that 🙏 🥨 ?

frontend/javascripts/oxalis/view/left-border-tabs/layer_settings_tab.tsx
```diff
@@ -457,6 +473,9 @@ class DatasetSettings extends React.PureComponent<DatasetSettingsProps, State> {
           <div className="flex-item">
             {hasHistogram && !isDisabled ? this.getEditMinMaxButton(layerName, isInEditMode) : null}
           </div>
+          <div className="flex-item">
+            {isVolumeTracing && !isDisabled ? this.getMergeWithFallbackLayerButton(layer) : null}
```
daniel-wer, line 477: Not every volume annotation has a fallback layer. Shouldn't this be checked before showing this button?

frontend/javascripts/oxalis/view/right-border-tabs/starting_job_modals.tsx
```diff
@@ -292,17 +293,13 @@ function StartingJobModal(props: StartingJobModalProps) {
...
     initialOutputSegmentationLayerName = getReadableNameOfVolumeLayer(fixedSelectedLayer, tracing);
   }
-  initialOutputSegmentationLayerName = `${initialOutputSegmentationLayerName}_corrected`;
-  // TODO: Other jobs also have an output segmentation layer. The names for them should also be configurable.
-  const hasOutputSegmentationLayer = jobName === "apply merger mode";
+  initialOutputSegmentationLayerName = `${
+    initialOutputSegmentationLayerName || "segmentation"
+  }_corrected`;
+  // TODO: Other jobs also have an output segmentation layer. The names for these jobs should also be configurable.
+  const hasOutputSegmentationLayer = jobName === "materializing this volume annotation";
```
daniel-wer, line 300: This comparison seems rather brittle, I would at least extract the string into a constant and use the constant in both places. Also, I think `materialize volume annotation` would be a more natural job name.

frontend/javascripts/oxalis/view/right-border-tabs/starting_job_modals.tsx
```diff
@@ -435,17 +434,51 @@ export function NeuronInferralModal({ handleClose }: Props) {
...
+  const fixedSelectedLayer = selectedVolumeLayer || activeSegmentationTracingLayer;
+  const isMergerModeEnabled = useSelector(
+    (state: OxalisState) => state.temporaryConfiguration.isMergerModeEnabled,
+  );
+  let description = (
+    <p>
+      Start a job that takes the current state of this volume annotation and materializes it into a
+      new dataset. All annotations done on the
+      {` "${fixedSelectedLayer && getReadableNameOfVolumeLayer(fixedSelectedLayer, tracing)}" `}
+      volume layer will be merged with the data of the fallback layer.
```
daniel-wer, line 457: Again, I would expect that not every annotation has a fallback layer or am I wrong about that?

frontend/javascripts/oxalis/view/right-border-tabs/starting_job_modals.tsx
daniel-wer: Instead of having two different ways to describe merger mode you could replace

> Additionally, the skeletons will be used to merge segments as merger mode is active.

and this part of the description

> In the new output dataset the segments connected via skeleton nodes will be merged.

both with 

"Since merger mode is currently active, in the new output dataset the segments connected via skeleton nodes will be merged."

but it's fine like this as well.

frontend/javascripts/oxalis/view/right-border-tabs/starting_job_modals.tsx
```diff
@@ -435,17 +434,51 @@ export function NeuronInferralModal({ handleClose }: Props) {
...
+      </p>
+    );
+  }
+
   return (
     <StartingJobModal
       handleClose={handleClose}
-      jobName="apply merger mode"
+      title="Start Materializing this Volume Annotation"
+      jobName="materializing this volume annotation"
```
daniel-wer, line 478: See my other comment about this string.

daniel-wer (commented):
Please check my comments. The biggest issue in my opinion is that the code assumes that there always is a fallback layer. In my understanding, that's not the case, annotations can also be created without a fallback layer :thinking:

frontend/javascripts/oxalis/view/right-border-tabs/starting_job_modals.tsx
daniel-wer: Could you explain the rationale behind `additionalAllowedNames`? I didn't understand why this is needed from looking at the code.

MichaelBuessemeyer: `additionalAllowedNames` is a list of names that are allowed to enter although the dataset already has a layer with that name. In the modal this is used to allow the user to let the output segmentation layer have the exact same name as the input segmentation layer. This is possible because the output segmentation layer replaces the input segmentation layer in the output dataset. Therefore non `_corrected` or something like this at the end of the output layer name is needed.

Does this behaviour make sense to you?

daniel-wer: Ah yes that makes sense. One could also remove the extra parameter `additionalAllowedNames` and filter the `layers` before handing them as a prop to `OutputSegmentationLayerNameFormItem` (to filter out the segmentation layer that is replaced). A small comment could be added at the filtering spot to explain why this is done.
I feel like this would be cleaner and easier to understand, what do you think?

frontend/javascripts/oxalis/view/right-border-tabs/starting_job_modals.tsx
MichaelBuessemeyer: I added an enum for the job names that is now used everywhere where jobname strings were used previously

daniel-wer: Awesome :1st_place_medal:

frontend/javascripts/oxalis/view/right-border-tabs/starting_job_modals.tsx
MichaelBuessemeyer: I changed the code so that 
> All annotations done on the
      {` "${fixedSelectedLayer && getReadableNameOfVolumeLayer(fixedSelectedLayer, tracing)}" `}
      volume layer will be merged with the data of the fallback layer.

will only be shown when there is a fallback layer.

MichaelBuessemeyer:
@daniel-wer I incorporated your feedback, could you please check the changes 💇 ?

docs/volume_annotation.md
```diff
@@ -36,6 +36,15 @@ In the `Segmentation` tab on the right-hand sidepanel, you can see the segment I
 
 The active segment ID under the cursor can be found in the status bar at the bottom of the screen or through the context-sensitive menu on right-click.
 
+### Merging volume annotation with fallback data
+
+Once finished annotating a volume layer with a fallback layer, the state of this volume layer can be materialized into a new dataset. for this go to the layer settings in the left border. on the top right of the volume layer is the following button:
```
daniel-wer, line 41:
```suggestion
After finishing the annotation of a volume layer with a fallback layer, the combined state of these layers can be materialized into a new dataset. For this go to the layer settings in the left border tab. On the top right of the volume layer is the following button:
```

docs/volume_annotation.md
```diff
@@ -48,13 +57,18 @@ To fix split errors see section on flood filling below.
 
 ![Video: Proof-Reading Volume Annotations](https://www.youtube.com/watch?v=Sq4AuWanK14)
 
+After finishing the proof-reading, a long-running job can be started to apply the merging of segments into a new dataset with the same layers. The job can be started via a modal. The modal can be open by clicking on the button next to the merger mode button.
```
daniel-wer, line 60:
```suggestion
After finishing the proof-reading, a long-running job can be started to apply the merging of segments into a new dataset with the same layers. The job can be started via a modal. The modal can be opened by clicking on the button next to the merger mode button.
```

daniel-wer (commented):
Please also see my comment regarding `additionalAllowedNames`.

frontend/javascripts/oxalis/view/right-border-tabs/starting_job_modals.tsx
MichaelBuessemeyer: > what do you think?

Sounds good to me :)

MichaelBuessemeyer:
@daniel-wer I finished your last feedback, could you please have a look again? Maybe Monday so this pr can be merged on that day?

docs/volume_annotation.md
```diff
@@ -36,6 +36,15 @@ In the `Segmentation` tab on the right-hand sidepanel, you can see the segment I
 
 The active segment ID under the cursor can be found in the status bar at the bottom of the screen or through the context-sensitive menu on right-click.
 
+### Merging volume annotation with fallback data
+
+After finishing the annotation of a volume layer with a fallback layer, the combined state of these layers can be materialized into a new dataset. For this go to the layer settings in the left border tab. On the top right of the volume layer is the following button:
```
philippotto, line 41:
```suggestion
After finishing the annotation of a volume layer with a fallback layer, the combined state of these layers can be materialized into a new dataset. For this, go to the layer settings in the left border tab. On the top right of the volume layer is the following button:
```

frontend/javascripts/oxalis/view/right-border-tabs/starting_job_modals.tsx
```diff
@@ -232,6 +237,7 @@ function StartingJobModal(props: StartingJobModalProps) {
   const tracing = useSelector((state: OxalisState) => state.tracing);
   const activeUser = useSelector((state: OxalisState) => state.activeUser);
   const layers = chooseSegmentationLayer ? getSegmentationLayers(dataset) : getColorLayers(dataset);
+  const allLayer = getDataLayers(dataset);
```
philippotto, line 240:
```suggestion
  const allLayers = getDataLayers(dataset);
```

usages need to be changed then, too.

frontend/javascripts/oxalis/view/right-border-tabs/starting_job_modals.tsx
philippotto: please open a follow-up issue for this :)

frontend/javascripts/oxalis/view/right-border-tabs/starting_job_modals.tsx
```diff
@@ -297,7 +303,19 @@ function StartingJobModal(props: StartingJobModalProps) {
     initialOutputSegmentationLayerName || "segmentation"
   }_corrected`;
   // TODO: Other jobs also have an output segmentation layer. The names for these jobs should also be configurable.
-  const hasOutputSegmentationLayer = jobName === "materializing this volume annotation";
+  const hasOutputSegmentationLayer = jobName === JobNames.MATERIALIZE_VOLUME_ANNOTATION;
+  const notAllowedOutputLayerNames = allLayer
+    .filter((layer) => {
+      // Filtering out the layer that is currently selected as this name is allowed
+      // because the output layer overwrites the selected layer.
```
philippotto, line 310:
```suggestion
      // Filtering out the layer that is currently selected as this name is not allowed
      // because the output layer would overwrite the selected layer.
```

if I understood it correctly

frontend/javascripts/oxalis/view/right-border-tabs/starting_job_modals.tsx
```diff
@@ -446,17 +463,24 @@ export function MaterializeVolumeAnnotationModal({
...
-      new dataset. All annotations done on the
-      {` "${fixedSelectedLayer && getReadableNameOfVolumeLayer(fixedSelectedLayer, tracing)}" `}
-      volume layer will be merged with the data of the fallback layer.
+      new dataset.
+      {hasFallbackLayer
+        ? ` All annotations done on the "${readableVolumeLayerName}" volume layer will be merged with the data of the fallback layer. `
+        : null}
       {isMergerModeEnabled
-        ? " Additionally, the skeletons will be used to merge segments as merger mode is active. "
+        ? " Since merger mode is currently active, in the new output dataset the segments connected via skeleton nodes will be merged. "
```
philippotto, line 483:
```suggestion
        ? " Since the merger mode is currently active, the segments connected via skeleton nodes will be merged within the new output dataset."
```

philippotto (approved):
See my few nitpicks, but other than that everything looks great :)

frontend/javascripts/oxalis/view/right-border-tabs/starting_job_modals.tsx
MichaelBuessemeyer: No sorry, that is exactly not the case here.
All layer names are forbidden !except! the name of the selected volume layer, because this volume layer will don't exist in the output dataset, but it will be replaced with the modified version of this layer. The modified version then will have the name given by the `OutputSegmentationLayerNameFormItem`.

frontend/javascripts/oxalis/view/right-border-tabs/starting_job_modals.tsx
MichaelBuessemeyer: issue: #6197

frontend/javascripts/oxalis/view/right-border-tabs/starting_job_modals.tsx
MichaelBuessemeyer: As you misunderstood this comment, do you have a different way to express this, that might be easier to understand / not misunderstand?

philippotto, line 310:
```suggestion
      // Existing layer names may not be used for the output layer. The only exception is the name of the currently selected layer.
      // If that layer name is chosen, the original layer data won't exist in the output dataset, since it will be replaced by the new output.
```

Maybe like this? Needs prettier-formatting, though. You could also keep the comment as is. My interpretation didn't really make sense, which is why I'd hope I won't question the veracity of the comment in the future :laughing:

MichaelBuessemeyer: > // If that layer name is chosen, the original layer data won't exist in the output dataset, since it will be replaced by the new output.

Technically speaking, this is not correct. The selected layer will never directly exist in the output dataset, if the job has an output layer. Therefore, this is even the case if the layer name is not chosen. But your 3rd sentence reads like it is only the case when the layer name is chosen.

I rephrased your suggestion a little and made it more expressive:

       // is the name of the currently selected layer. This layer is the only one not
       // copied over from the original dataset to the output dataset.
       // Therefore, this name is available as the name for the output layer name.
       // That is why that layer is filtered out here.

philippotto: Makes sense :+1:

PR #6087, opened 2022-03-01
Fix layer headers layout

### LINKED ISSUE #6008: Ensure layer headers look good even with long names

Dagobert42:
- https://___.webknossos.xyz

### Steps to test:
- create layers with longer names

### Issues:
- fixes #6008 and #5627 

------
- [✓] Ready for review

### DIFF
--- a/frontend/stylesheets/flex_layout_overwrites.less
+++ b/frontend/stylesheets/flex_layout_overwrites.less
@@ -121,14 +121,3 @@
 .hide_first_splitter_border ~ .hide_first_splitter_border {
   pointer-events: all;
 }
-
-.flex-container {
-  justify-content: space-between;
-  display: flex;
-}
-
-.flex-item {
-  display: inline-block;
-  flex-grow: 1;
-  word-break: break-all;
-}
--- a/frontend/stylesheets/flex_layout_overwrites.less
+++ b/frontend/stylesheets/flex_layout_overwrites.less
@@ -121,3 +121,14 @@
 .hide_first_splitter_border ~ .hide_first_splitter_border {
   pointer-events: all;
 }
+
+.flex-container {
+  justify-content: space-between;
+  display: flex;
+}
+
+.flex-item {
+  display: inline-block;
+  flex-grow: 1;
+  word-break: break-all;
+}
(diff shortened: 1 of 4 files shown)


Dagobert42:
Two questions regarding these first two commits: 

1. Will the change to the EditableTextLabel have any unintended effects anywhere else? 
2. Is there a better place to add the css rules to?

Dagobert42:
Here's what it currently looks like:
Screenshot 2022-03-01 at 15 11 19

philippotto:
Awesome! Major improvement :+1: 

> The EditableTextLabel made for awkward break behaviour so I moved the edit button into the same span as the label. Could this break the UI in other places?

I think, the description here looks weird now:
image

> Nonetheless, the edit button can still appear out of place as can be seen above so maybe there's a better solution.

You might have success with the approach (a) from here: https://codepen.io/snookca/pen/JBdBVZ
But if it takes longer than 15 minutes, I'd skip it for now :)

frontend/stylesheets/flex_layout_overwrites.less
```diff
@@ -121,3 +121,14 @@
...
+.flex-container {
+  justify-content: space-between;
+  display: flex;
+}
+
+.flex-item {
+  display: inline-block;
+  flex-grow: 1;
+  word-break: break-all;
+}
```
philippotto, line 134: This file is actually meant to tune the FlexLayout library (has nothing to do with flexboxes, even though I see why it could seem like it :)). I'd suggest to move it into `_utils.less` as these are generic helper classes.

Dagobert42:
The latest changes to EditableTextLabel also have an impact on TextWithDescription. This leads to changes in the annotations dashboard. However, I'd say that the break-word behaviour there wasn't optimal in the first place:

Before:
Screenshot 2022-03-04 at 11 39 36

After:
Screenshot 2022-03-04 at 11 41 11

I'd open a new issue to deal with that.

philippotto (approved):
Great, I think, this can be merged :) Preferably, on Monday, as we usually try to not push to master on Fridays.

Dagobert42:
Thanks! Will merge on monday :)

PR #6098, opened 2022-03-07
Add remote origin headers also in error case

### LINKED ISSUE #6078: Set CORS headers also in error case
Routes that have the AllowRemoteOrigin block set the CORS headers only in success case. They should also be set in the error case so that the actual error message is propagated.

fm3:
This allows the CORS headers to be set also in error case by moving the adding of the header directly to the function that creates Result objects from Success or Failures etc (BoxToResultHelpers).
 - Moves the mechanism from the individual routes up to the whole Controller (override allowRemoteOrigin = true)
 - Only controller that still uses the old mechanism (no longer implemented as block wrapper, though) is DataSetController because it has some routes where CORS should be set and some where not. Note that for those two routes the error case will still be missing the headers (would have to extract them to separate controllers to reuse the mechanism. let’s do that only if the need arises).
 - Moves some code from the datastore-specific Controller to the generic ExtendedController trait to reduce import redundancy
 - When reviewing, hide whitespace in github diff. Note that some line breaks may be changed by the formatter.

### Steps to test:
- observe result headers in network tab of /api/datasets (should have cors headers) and /api/teams (should not)
- break a datastore route (e.g. insert line `_ <- Fox.failure("Nope")` after line 454 of the BinaryDataController) – subsequent histogram requests should fail, but still have the header.

### Issues:
- fixes #6078 

------
- [x] Ready for review


philippotto (approved):
Awesome :+1: LGTM!

PR #6102, opened 2022-03-09
Add "hide/show other trees" to tree context-menu

hotzenklotz:
There already is a functionality in webknossos to "Hide all inactive trees", i.e. toggle the visibility of all other trees but the currently selected one. This feature is bound to keyboard hotkey `2`. 
In this PR, I exposed this functionality through the right-click context menu for each tree/group.

- https://___.webknossos.xyz

### Steps to test:
- Create a skeleton annotation with a few trees and groups.
- Right-click on a tree/group and select "Hide/Show all other trees"
- The selected tree/group should now: 
  - be selected (blue background behind name)
  - the visibility of all tree/groups should be toggled
 - Right-click again to undo

### Issues:
- no issue 

------
- [x] Ready for review


philippotto (approved):
Works mostly well :+1: I think, it's worth an entry in the changelog.

Some observations which are not blocking in my opinion, though:
- the context menu does not close when using the new item. it's a bit weird, but in line with most of the other items, since the actions are often used in succession (e.g., shuffle the tree color until it's good). so, it can probably stay as is. however, for the context menu on the group item, the behavior is different (it will close directly after using the entry).
- when using this action on a tree which is invisible, it will be turned visible, but all the other trees won't turn invisible. maybe acceptable, since one can just click again. can be reproduced with this state of trees:

image

hotzenklotz:
@philippotto Updated the changeling :-)

>the context menu does not close when using the new item. it's a bit weird, but in line with most of the other items, since the actions are often used in succession (e.g., shuffle the tree color until it's good). so, it can probably stay as is. however, for the context menu on the group item, the behavior is different (it will close directly after using the entry).

Good catch. I fixed this. Now the context menu closes after clicking on the "Hide/Show" menu entry.

> when using this action on a tree which is invisible, it will be turned visible, but all the other trees won't turn invisible. maybe acceptable, since one can just click again. can be reproduced with this state of trees:

Yes, this is odd. But since I am relying on an existing functionality I will not changes that as part of this PR.

CHANGELOG.unreleased.md
```diff
@@ -13,6 +13,7 @@ For upgrade instructions, please check the [migration guide](MIGRATIONS.released
 ### Added
 - The visible meshes are now included in the link copied from the "Share" modal or the "Share" button next to the dataset position. They are automatically loaded for users that open the shared link. [#5993](https://github.com/scalableminds/webknossos/pull/5993)
 - Added a new "Connectome Tab" which can be used to explore connectomes by visualizing neurites and their synaptic connections. Connectome files need to be placed in a `connectomes` folder inside of the respective segmentation layer. It is possible to craft links that automatically load specific agglomerates and their synapses when openend. For more information refer to https://docs.webknossos.org/webknossos/sharing.html#sharing-link-format. [#5894](https://github.com/scalableminds/webknossos/pull/5894)
+- Added a context-menu option when right-clicking on skeleton trees to hide/show all other trees but the selected one. Great for inspecting a single tree oin isolation. Identical to keyboard shortcut "2". [#6102](https://github.com/scalableminds/webknossos/pull/6102)
```
```suggestion
- Added a context-menu option when right-clicking on skeleton trees to hide/show all other trees but the selected one. Great for inspecting a single tree in isolation. Identical to keyboard shortcut "2". [#6102](https://github.com/scalableminds/webknossos/pull/6102)
```

philippotto (approved):
Good stuff!

PR #6103, opened 2022-03-09
Unify long-running RPC timeouts

fm3:
Remote calls between wk and datastore/tracingstore had different kind of timeouts (some the default of 2 min, some 30 min, some infinite). This PR introduces the withLongTimeout method for the RPC wrapper, and adds it to the RPCs that are potentially slow (basically everything that handles the content of potentially large skeletons or volume data).

### Steps to test:
- (will be hard to test without very large data)
- downsample a large mag1 volume annotation, should not timeout

------
- [x] Ready for review


leowe (approved):
LGTM!

PR #6104, opened 2022-03-09
Fix resolution datasource-properties suggestion bug

### LINKED ISSUE #6057: Don’t show datasource-properties.json suggestions if diff is trivial
In the dataset settings view, the backend sends a “suggested” datasource-properties.json. If there is a diff to the existing one, the user is notified about this. A kind of diff that happened a few times is isotropic wkw resolutions expressed as `1` vs `[1, 1, 1]`. This diff is trivial and should not be displayed to the user.
Possible solution:
 - in frontend, detect trivial diffs
 - in backend, always write isotropic (could even send this for “old” ones that have just `1` when reserializing)
   - this may just be changing the `Left` case to  `case Left(r)  => Vec3Int.Vec3IntWrites.writes(Vec3Int(r, r, r))` in `ResolutionFormatHelper`
 - in backend, take previous existing datasource as input and take the format present there for the newly inferred datasource.

Compare discussion in https://scm.slack.com/archives/C5AKLAV0B/p1645109446016519

leowe:
- Fix preview of recommended config differed in isotropic resolutions (`1` vs `[1,1,1]`)
- Removed internal representation of single int, instead are only using `Vec3Int`
- Also changed the proxy addresses to use `127.0.0.1` instead of `localhost` to make it work on macOS

- https://___.webknossos.xyz

### Steps to test:
- change `datasource-config.json` on disk to have `1` as resolution and check in the datasource settings in webKnossos whether there are suggested changes regarding the resolution

### Issues:
- fixes #6057

------
- [x] Ready for review


fm3 (approved):
Awesome :rocket: 
image This I like :D

Please wait with merging until https://github.com/scalableminds/webknossos/pull/6103 is through, since that one was requested for today

PR #6111, opened 2022-03-12
adapt readme for m1-mac setups

normanrz:
Makes changes to the readme in order to better support installations on M1-based Macs.
Also, gets rid of the `webpack.pid` file, that doesn't seem to be used anywhere and creates problems, because the `target` folder is not always available when running `yarn start` for the first time (race condition with sbt).

- [x] Ready for review


philippotto (approved):
:+1: can't say anything about the m1 stuff, though.

leowe (approved):
Nice! To be sure that this installation actually works, we'd probably need a fresh install or a VM setup or something. For now, this looks exactly like the things we did at my laptop.

Issue #5163, opened 2021-02-13
Histogram values are mag-dependent. Is mag1 the best choice?


fm3:
We currently compute the histogram on the highest resolution data available (usually mag1). This has the effect that noise in the image data heavily influences the histogram values. High-Res SEM data can be quite noisy, some datasets much more so than others.

The noise can get averaged out during downsampling, resulting in a much narrower histogram. We should discuss if sampling the histogram from downsampled data may yield a better representation of the “effective“ brightness distribution.

It may even be useful to make this an option available to the user so that they can inspect in this way how noisy their dataset is.

Example taken from a nucleus in the P5 dataset published in 10.1126/science.abb4534 : 

![image](https://user-images.githubusercontent.com/1390035/107846390-9d5c2e80-6de3-11eb-9ce1-fb607e809c8e.png)

PR #6113, opened 2022-03-15
lower histogram clip threshold a bit

### LINKED ISSUE #5990: Follow-up to auto contrast/brightness

Dagobert42:
- https://___.webknossos.xyz

### Steps to test:
- use histogram clip button

### Issues:
- fixes #5990 

------
- [x] Ready for review

### DIFF
--- a/frontend/javascripts/oxalis/model/sagas/clip_histogram_saga.js
+++ b/frontend/javascripts/oxalis/model/sagas/clip_histogram_saga.js
@@ -20,7 +20,7 @@ function onThresholdChange(layerName: string, [firstVal, secVal]: [number, numbe
   }
 }
 
-async function getClippingValues(layerName: string, thresholdRatio: number = 0.03) {
+async function getClippingValues(layerName: string, thresholdRatio: number = 0.005) {
   const { elementClass } = getLayerByName(Store.getState().dataset, layerName);
   const [TypedArrayClass] = getConstructorForElementClass(elementClass);
 
--- a/frontend/javascripts/oxalis/model/sagas/clip_histogram_saga.js
+++ b/frontend/javascripts/oxalis/model/sagas/clip_histogram_saga.js
@@ -20,7 +20,7 @@ function onThresholdChange(layerName: string, [firstVal, secVal]: [number, numbe
   }
 }
 
-async function getClippingValues(layerName: string, thresholdRatio: number = 0.05) {
+async function getClippingValues(layerName: string, thresholdRatio: number = 0.03) {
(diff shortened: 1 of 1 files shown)


fm3 (commented):
Thanks a lot for looking into this!

For the datasets I used this for, I believe 3% is still clipping away too much.

The way I understand this, the clip values are only there because of noise in the image data. Is this correct?
If the image data has no noise, I would want to see all present brightness values, because anything that is mapped above 1 or below 0 is undesired loss of information – so in case of no noise I would want this clip value to be 0.
Noise means brighter and darker pixels are present than the “effective” brightness range of the “real” data. This is why clip value 0 would fit those in, making the “effective” data bland.
This corresponds to what I described in #5163 
So what we really want is to de-noise the image (e.g. by slightly blurring it – or just take the histogram from higher mags, which are already kind of blurred) and then use clip value 0. The effect of using different mags can already be tested by zooming out. Try for example this (very noisy) public EM dataset https://webknossos.org/datasets/MPI_Brain_Research/2020-06-12-p5-align2/view#3447,5165,2560,0,30.867 – there, in mag1 the strong 5% already look ok because of the heavy noise. In higher mags (denoised), 5% is way too aggressive.
Until we can have that, I would rather err on the side of displaying bland data instead of accidentally clipping away interesting data. So I would argue for something way lower, like 0.5 %.

Do you agree @Dagobert42  @philippotto ?

Dagobert42:
Thanks for the detailed feedback :) 
I see! I've lowered the clipping value even more for a start. 

As I understood, the intention was to overexaggerate the contrast a bit. But then we would actually want a sort of noise estimation or noise level function? I will see if I can find any simple algortihms for this.

fm3 (approved):
Thanks a lot! :) I see, but I’d think overexaggerating the contrast is usually not the user’s wish. I may be wrong though :shrug:. Let’s see if we get external feedback on this.

I think you don’t need to look further into the denoising for now. This is a good follow-up issue (I just wrote https://github.com/scalableminds/webknossos/issues/6115) that we may schedule at a later time.

philippotto:
> As I understood, the intention was to overexaggerate the contrast a bit. But then we would actually want a sort of noise estimation or noise level function? I will see if I can find any simple algortihms for this.

If I understood @fm3 correctly, a simple blur pass before creating the histogram could be enough? See [here](https://img.ly/blog/how-to-apply-filters-in-javascript/) the code beginning with `if (filter === "blur") {` as an example.

I agree with @fm3 and think this is something for another PR for which we should figure out the priority later :)

PR #6116, opened 2022-03-16
allow all resolutions if there is no volume annotation in the nml

### LINKED ISSUE #6063: Uploading annotation zip with empty volume layers leads to error
## Context
When uploading an annotation zip-file which includes empty volume tracings (which is produced e.g. when downloading an annotation without brushing, or via wk-libs), then I get the following error:
```
Unsuccessful WS request to https://webknossos.org/tracings/volume/5aa9f1e2-fa19-467c-8d1b-85d776548aa1/initialData (ID: 551).Status: 400. Response: {"messages":[{"error":"Resolution restrictions result in zero resolutions"}]}
```

This should not happen.

## Steps to Reproduce the bug
1. Drag'n'Drop the following annotation zip into webknossos
  [l4dense_motta_et_al_nuclei__explorational__jstriebel__b3eddd.zip](https://github.com/scalableminds/webknossos/files/8108776/l4dense_motta_et_al_nuclei__explorational__jstriebel__b3eddd.zip)
2. The error mentioned above is shown, no annotation opens :crying_cat_face:

leowe:
- When uploading a NML check whether there are any volume annotations, if there are none, allow all resolutions
- add `leowe` to slack notifications
- fix small language mistake
- change `logger.error` to `logger.warn` for filesystem checking for the data set indexing

- https://emptytracing.webknossos.xyz/

### Steps to test:
- Open any dataset (e.g. `l4_sample`)
- Create a empty annotation, download it, and then drag and drop it in the view (or use [this](https://github.com/scalableminds/webknossos/files/8262647/l4_sample__explorational__suser__bb74f7.zip) empty annotation)
- A new annotation should open without an error message

### Issues:
- fixes #6063 

------
- [x] Ready for review

### DIFF
--- a/webknossos-tracingstore/app/com/scalableminds/webknossos/tracingstore/tracings/volume/VolumeTracingService.scala
+++ b/webknossos-tracingstore/app/com/scalableminds/webknossos/tracingstore/tracings/volume/VolumeTracingService.scala
@@ -202,7 +202,7 @@ class VolumeTracingService @Inject()(
         saveBucket(dataLayer, bucketPosition, bytes, tracing.version)
       }
     }
-    // if none of the tracings contained any volume data. do not save buckets, use full resolution list
+    // if none of the tracings contained any volume data, use the dataset’s full resolution list
     if (savedResolutions.isEmpty) return getRequiredMags(tracing).map(_.toSet)
 
     unzipResult.map(_ => savedResolutions.toSet)
--- a/webknossos-tracingstore/app/com/scalableminds/webknossos/tracingstore/tracings/volume/VolumeTracingService.scala
+++ b/webknossos-tracingstore/app/com/scalableminds/webknossos/tracingstore/tracings/volume/VolumeTracingService.scala
@@ -186,10 +186,14 @@ class VolumeTracingService @Inject()(
   def initializeWithData(tracingId: String,
                          tracing: VolumeTracing,
                          initialData: File,
-                         resolutionRestrictions: ResolutionRestrictions): Box[Set[Vec3Int]] = {
+                         resolutionRestrictions: ResolutionRestrictions): Fox[Set[Vec3Int]] = {
     if (tracing.version != 0L) {
       return Failure("Tracing has already been edited.")
     }
+
+    val resolutionSet = resolutionSetFromZipfile(initialData)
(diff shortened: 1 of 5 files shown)


webknossos-tracingstore/app/com/scalableminds/webknossos/tracingstore/tracings/volume/VolumeTracingService.scala
```diff
@@ -186,11 +186,15 @@ class VolumeTracingService @Inject()(
   def initializeWithData(tracingId: String,
                          tracing: VolumeTracing,
                          initialData: File,
-                         resolutionRestrictions: ResolutionRestrictions): Box[Set[Vec3Int]] = {
+                         resolutionRestrictions: ResolutionRestrictions): Fox[Set[Vec3Int]] = {
     if (tracing.version != 0L) {
       return Failure("Tracing has already been edited.")
     }
 
+    val resolutionSet = resolutionSetFromZipfile(initialData)
```
fm3, line 194: This unpacks the zipfile an additional time for each upload. Could the empty check also happen on the result from below?

fm3 (commented):
Cool stuff, I like that you also fixed a few small things along the way! :D  
I added a suggestion for performance optimization.

webknossos-tracingstore/app/com/scalableminds/webknossos/tracingstore/tracings/volume/VolumeTracingService.scala
leowe: Good point, I changed it.

webknossos-tracingstore/app/com/scalableminds/webknossos/tracingstore/tracings/volume/VolumeTracingService.scala
```diff
@@ -202,6 +202,9 @@ class VolumeTracingService @Inject()(
         saveBucket(dataLayer, bucketPosition, bytes, tracing.version)
       }
     }
+    // if none of the tracings contained any volume data. do not save buckets, use full resolution list
```
fm3, line 205:
```suggestion
    // if none of the tracings contained any volume data, use the dataset’s full resolution list
```

fm3 (approved):
LGTM 🚀

PR #6130, opened 2022-03-30
Fix copy slice tool in mags other than mag 1

### LINKED ISSUE #5938: Allow copy slice in higher mag
## Detailed Description
Copying the previous slice (`v` shortcut) is currently only possible in mag 1. Higher mags should be supported as long as the viewport area can also be labeled by the volume tools (e.g., you cannot brush in mag 16 --> you shouldn't be able to copy the previous slice in mag 16 to avoid huge resource consumptions).

## Context

daniel-wer:
Turns out the copy slice tool was already enabled for the second finest mag, but it was buggy.

I fixed that the z-position was wrong when applying the voxel map, because it was not adapted to the resolution that was labeled in. I also fixed that the copying sometimes didn't work at all, because when annotation in z resolution 2, the previous/next slice might be 2 slices away instead of one (depending on the exact current voxel position in z).

I understand the text in the issue to mean that the copy slice tool should work in the resolutions where brushing works (finest four resolutions). However, for now it's only enabled in the resolutions where the trace tool works (finest two resolutions) since I found it to be conceptually more similar regarding the number of voxels that can be annotated at once. Happy to change that if you feel differently.

- https://___.webknossos.xyz

### Steps to test:
- Test the copy slice tool (`v`, `shift + v`) in mag 1 and 2
- Test that in mag 2 it doesn't matter whether the z coordinate is even or odd

### Issues:
- fixes #5938 

------
- [x] Ready for review


philippotto (commented):
Great fixes! However, I'm still observing buggy (or at least confusing) behavior. I think, it would be great to show a toast when no voxels were copied. E.g., `No voxels with cell id 7 were found to copy from slice x/y/z=3`. For example, when using d/f switching, this could help a bit.

Now, to the buggy scenario I experienced:

- Open this annotation https://fixcopysliceinothermags.webknossos.xyz/annotations/Explorational/624578440100009c00f6ef49#3478,3490,1068,0,0.467
- Directly press v -> nothing happens. However, d/f switching is disabled and switching to the previous slice with `d` shows that there is a blue cell.

For me, wk is zooomed to mag 1. I restricted the volume annotation itself to go from mag 2 to mag 16, since I wanted to test the restriction case, too. Not sure whether this has something to do with my experience, though.

daniel-wer:
Thanks for your review :) I'll add the toast you describe :+1: 

Interesting test scenario, I looked into it a bit.

Looks like the segmentation does indeed exist in slice z 1067 in mag 1-1-1, but it actually doesn't exist in slice z 1067 in mag 2-2-1 (which has the same z-resolution). I would've assumed it wouldn't be possible to do that from within webKnossos.

Further investigation showed that there might be a bug in the frontend up/downsampling for anisotropic resolutions. I'm seeing weird behavior if I brush a slice in mag 2-2-1 and then switch to mag 1-1-1, then the previous slice is brushed as well. The other way around if I brush a slice in mag 1-1-1, then the next slice is brushed as well, but only in mag 1-1-1, not in mag 2-2-1 (which is what seems to have happened in your case which then led to the buggy behavior with the copy slice tool).

If one switches to a slice that actually has a segmentation in the slice next to it (in the mag that one is allowed to annotate in), the copy slice tool works as expected to me.

In any case, I'm wondering whether it should be possible at all to brush/copy-slice/... in mag 1-1-1 if volume annotation is restricted to mag 2 to mag 16? Wouldn't this lead to other weird effects? For example, if the resolutions were 1-1-1 and 2-2-2 (restricted to mag 2), brushing in mag 1 would always brush the next slice as well (since the annotation is restricted to happen in mag 2)?

daniel-wer:
> Hm, very weird. Also because it's mag 2-2-1 and not mag 2-2-2. Could you open another issue for this? Do you think this only happens because of the mag restrictions?

I tested this with an unrestricted annotation and there the bug did not occur. I opened issue https://github.com/scalableminds/webknossos/issues/6138

PR #6132, opened 2022-03-31
change cursor according to annotation tool

### LINKED ISSUE #4879: Change cursor according to active tool
@fm3 suggested that we could switch the cursor depending on the active tool. Would be a a follow-up to #4875.

Dagobert42:
- https://___.webknossos.xyz

### Steps to test:
- hover on input catchers
- cycle through active tools

### Issues:
- fixes #4879 

------
- [ ] Ready for review


Dagobert42:
The lasso and new-bounding-box icons are problematic as they cant be filled without further modification. I replaced the lasso with a crosshair and made new-bounding-box dark but its probably not quite what we want.
Also the transposition of the actual clicking position still needs to be fine-tuned.
Regarding this, would you expect A or B here?

Screenshot 2022-03-31 at 18 42 51

normanrz:
Not sure it is a good idea to replace the built-in cursors, because it makes it harder to know what exact pixel will be targetted.
Also, we need to be careful with copying icons from FontAwesome into our repository. Not sure, if the license permits that.

Dagobert42:
> Not sure it is a good idea to replace the built-in cursors, because it makes it harder to know what exact pixel will be targetted.
Agreed, but I think we found a good compromise with most icons. For now I went through all the SVGs and tried to lign up their click position as best I could.

Regarding the Font Awesome license, I've checked it and it allows for modifications.

frontend/javascripts/oxalis/view/input_catcher.js
```diff
@@ -79,48 +76,77 @@ export function recalculateInputCatcherSizes() {
...
+    MOVE: "move",
+    SKELETON: "crosshair",
+    BRUSH: "url(/assets/images/paint-brush-solid-border.svg) 0 10,auto",
+    ERASE_BRUSH: "url(/assets/images/eraser-solid-border.svg) 0 8,auto",
+    TRACE: "url(/assets/images/lasso-pointed-solid-border.svg) 0 14,auto",
+    ERASE_TRACE: "url(/assets/images/eraser-pointed-solid-border.svg) 0 16,auto",
+    FILL_CELL: "url(/assets/images/fill-pointed-solid-border.svg) 0 16,auto",
+    PICK_CELL: "url(/assets/images/eye-dropper-solid-border.svg) 0 12,auto",
+    BOUNDING_BOX: "move",
+  };
```
philippotto, line 125: You can extract this constant to the module-level :weight_lifting:

philippotto (approved):
Awesome :) Please update the changelog before merging, though!

PR #6140, opened 2022-04-06
Disable editing of tracing dependent options in right tabs in read only mode

### LINKED ISSUE #6136: In read-only mode, user input is accepted but silently discarded
[Link to Discuss](https://discuss.webknossos.org/t/in-read-only-mode-user-input-is-accepted-but-silently-discarded/1717/2)

Ideally, in read-only, nothing should be editable (UI should be disabled). Can we enforce this somehow?

Doublecheck:
- comments
- trees
- bounding boxes
- segments

MichaelBuessemeyer:
This PR disables all editing options of the right border tabs which are tracing dependent when the tracing is in read-only mode. This includes the skeleton tab, the comments tab, and the bounding box tab. Each disabled option gets a tooltip telling the user this option is disabled due to read-only mode. 

Note that I could not find any setting in the segments tab that seems to take effect in opening the same tracing in read-only mode. 

- https://disableeditinginreadonlymode.webknossos.xyz
- Due to making a dataset public, it might be better to test this locally.

### Steps to test:
- Create a hybrid tracing with some trees and group them. Create at least one bounding box, and comment a few skeleton nodes.
- Then make the dataset publicly available (or choose an already public available dataset). Next also make the tracing public and open it via the sharing link in a private / incognito tab.
- Try to edit the skeletons, commends, and bounding boxes in the read-only version of the tracing in the private tab. It should not be possible except for a few settings like visibility and sorting.

### Issues:
- fixes #6136

------
- ~~[ ] Adapted [wk-libs python client](https://github.com/scalableminds/webknossos-libs/tree/master/webknossos/webknossos/client) if relevant API parts change~~
- [x] Ready for review


frontend/javascripts/messages.js
```diff
@@ -160,6 +160,7 @@ instead. Only enable this option if you understand its effect. All layers will n
   "tracing.compound_project_not_found":
     "It looks like this project does not have a single task completed. Make sure that at least one task of this project is finished to view it.",
   "tracing.no_allowed_mode": "There was no valid allowed annotation mode specified.",
+  "tracing.read_only_mode_notification": "This tracing is in read-only mode and cannot be updated.",
```
daniel-wer, line 163:
```suggestion
  "tracing.read_only_mode_notification": "This annotation is in read-only mode and cannot be updated.",
```

frontend/javascripts/oxalis/view/right-border-tabs/comment_tab/comment_tab_view.js
```diff
@@ -87,6 +87,7 @@ function getCommentSorter({ sortBy, isSortedAscending }: SortOptions): Comparato
 
 type StateProps = {|
   skeletonTracing: ?SkeletonTracing,
+  allowUpdates: boolean,
```
daniel-wer, line 90: I'd rename this to `allowUpdate` without the "s" to stay consistent

frontend/javascripts/oxalis/view/right-border-tabs/comment_tab/comment_tab_view.js
```diff
@@ -426,18 +431,27 @@ class CommentTabView extends React.Component<PropsWithSkeleton, CommentTabState>
                   />
                   <InputComponent
                     value={activeCommentContent}
-                    disabled={activeNodeMaybe.isNothing}
+                    disabled={isEditingDisabled}
+                    title={
+                      !this.props.allowUpdates
```
daniel-wer, line 436: Minor, feel free to ignore, but usually it's easier for a reader to parse this kind of condition if the "!" is dropped and the if/else is swapped. Same for the condition in line 449.

frontend/javascripts/oxalis/view/right-border-tabs/skeleton_tab_view.js
```diff
@@ -649,7 +656,7 @@ class SkeletonTabView extends React.PureComponent<Props, State> {
   }
 
   getSelectedTreesAlert = () =>
-    this.state.selectedTrees.length > 0 ? (
+    this.state.selectedTrees.length > 0 && this.props.allowUpdate ? (
```
daniel-wer, line 659: Why should this alert be disabled, it only helps to deselect the selection, doesn't it? If anything, shouldn't the tree selection itself be disabled, instead? However, I don't think that's important since it cannot be used to modify anything as far as I can see.

frontend/javascripts/oxalis/view/right-border-tabs/skeleton_tab_view.js
```diff
@@ -729,27 +737,45 @@ class SkeletonTabView extends React.PureComponent<Props, State> {
...
+                        : "Delete Selected Trees"
+                    }
+                    disabled={isEditingDisabled}
                   >
                     <i className="far fa-trash-alt" />
                   </ButtonComponent>
                   <ButtonComponent
                     onClick={this.toggleAllTrees}
-                    title="Toggle Visibility of All Trees (1)"
+                    title={
```
daniel-wer, line 762: Since the shortcuts and check boxes to toggle tree visibility still work, these two buttons could remain enabled, I would say. Toggling tree visibility is one of the things that should remain possible in read-only mode in my opinion.

daniel-wer (commented):
Nice work! I didn't spot any issues regarding the right tabs during my testing.

However, the context menu still offers a lot of functionality to edit the annotation. I know this was not specifically mentioned in the issue description, but it would be worthwhile to tackle this as well. For example, it's possible to create bounding boxes (which works - I think to make sure this never works in read-only the reducer should be adjusted), and it is offered to delete nodes, merge trees, delete edges, floodfill segments, etc...

Of course this could also be done as a followup PR.

MichaelBuessemeyer:
> However, the context menu still offers a lot of functionality to edit the annotation. I know this was not specifically mentioned in the issue description, but it would be worthwhile to tackle this as well. For example, it's possible to create bounding boxes (which works - I think to make sure this never works in read-only the reducer should be adjusted), and it is offered to delete nodes, merge trees, delete edges, floodfill segments, etc...

@daniel-wer Would you suggest to disabled the options and add a tooltip to them (just as it is done in the tabs on the right) or should I just not render the options? 🤔 

I am leaning towards not rendering the options that try to modify a tracing are not rendered at all. The disabled options on the right and the hint in the navbar should be enough for the user do understand why nothing can be edited, right?


Another question: I thought that the bounding boxes being editable in a read-only tracing is/was intended. Is it not?

frontend/javascripts/oxalis/view/components/input_component.tsx
```diff
@@ -103,20 +95,34 @@ class InputComponent extends React.PureComponent<InputComponentProps, InputCompo
...
-        onChange={this.handleChange}
-        onFocus={this.handleFocus}
-        onBlur={this.handleBlur}
-        value={this.state.currentValue}
-        onPressEnter={onPressEnter != null ? onPressEnter : defaultOnPressEnter}
-        onKeyDown={this.blurOnEscape}
-      />
-    );
+    let input;
+    if (isTextArea) {
```
daniel-wer, line 99: Why was it needed to split this up and duplicate the instantiation code?

daniel-wer (commented):
Code LGTM :+1: 

Do you plan to address the context menu in this PR or will you create a follow up issue? In any case, I would at least re-add the possibility to add bounding boxes in the bbox tab since we agree that this should be possible in read-only mode.

MichaelBuessemeyer:
> I'm also leaning towards not rendering to keep the context menu slim in read-only.

Context menu options that manipulate the annotation are no longer present in the context menu, in case the annotation is read only. 🎉 


> Whichever way, the two UIs should be consistent, I think :) Thinking about it, I would prefer to allow to create and edit bounding boxes in read-only. I've used this feature a lot in the past in view mode to find the largest bounding box of a dataset that doesn't contain black borders, for example.

I just read this, after making my last commit, which disables manipulating the bounding boxes in read-only mode (except hiding / showing them). It should be easy to reenable them. My question is: How did you manipulate a bounding box in read-only mode? The bounding box tool is not enabled in read-only mode 🤔. By only changing the number input in the bounding box tab?

frontend/javascripts/oxalis/view/components/input_component.tsx
MichaelBuessemeyer: I think I introduced this if/else in the typescript pr because typescript was complaining about something and this if/else fixed the error.

After a closer look, I could fix the problem now :)

daniel-wer: Awesome, thanks for having another look :+1:

daniel-wer (approved):
Looks very good to me :+1: 

> I just read this, after making my last commit, which disables manipulating the bounding boxes in read-only mode (except hiding / showing them). It should be easy to reenable them. My question is: How did you manipulate a bounding box in read-only mode? The bounding box tool is not enabled in read-only mode thinking. By only changing the number input in the bounding box tab?

I think, I just created a bounding box using the context menu, but wasn't able to modify it, since you disabled the inputs in the bounding box tab in this PR.

Personally, I would be in favor of allowing to create and manipulate bounding boxes in read-only annotations since I use this feature from time to time in "View" mode. However, I can see that it's not really consistent with disallowing all other modifications, especially for read-only annotations and could potentially lead to confusion. I would also be fine with having to create an annotation to create and modify bounding boxes in the future. So let's go with the way it is currently implemented in this PR (no bbox creation/modification allowed in read-only) unless @philippotto has another opinion on this?

philippotto:
> Personally, I would be in favor of allowing to create and manipulate bounding boxes in read-only annotations since I use this feature from time to time in "View" mode. However, I can see that it's not really consistent with disallowing all other modifications, especially for read-only annotations and could potentially lead to confusion. I would also be fine with having to create an annotation to create and modify bounding boxes in the future. So let's go with the way it is currently implemented in this PR (no bbox creation/modification allowed in read-only) unless @philippotto has another opinion on this?

Exactly my thoughts. It can be very practical to create/edit bboxes in view-mode, but creating a new annotation in that scenario is acceptable from the UX. The upside is the consistency. So, let's take that route for now :+1:

PR #6144, opened 2022-04-08
Zarr streaming

### LINKED ISSUE #6106: Expose dataset through a Zarr-compatible interface
We're already implementing support for reading/viewing Zarr datasets #6019 in webKnossos. Since, Zarr is not only a file format, but actually an API, we can expose all datasets on webKnossos as Zarr images. This will make it easy for users to access/stream data with any Zarr-compatible library (most commonly [zarr-python](https://github.com/zarr-developers/zarr-python/)).

I think we'd need to expose 2 new endpoints and maybe a third:
```
GET /data/zarr/:organization/:dataset/:layer/:mag/.zarray
GET /data/zarr/:organization/:dataset/:layer/:mag/:chunkdata
GET /data/zarr/:organization/:dataset/datasource-properties.json
```
- The contents `.zarray` endpoint would be derived from the actual available data (e.g. reading from the WKW header).
- The contents of the chunk data would ideally be streamed as-is. That means without decompression. That should work, because Zarr and WKW both support lz4 compression.
- The last endpoint would list the available layers and mags to be consumed in wklibs. In a future state, we could replace the proprietary datasource-properties.json with an OME-NGFF compatible metadata format.
- Authentication could be token-based. Maybe we can use the `Authorization: Bearer ...` or `Authorization: Basic ...` HTTP headers, because I think that would be easier to use from client libraries than an addition query param.

leowe:
- https://zarrstreaming.webknossos.xyz/

### Steps to test:
- Use this script to test:
```
def _assert_check_equality(args) -> None:
    view_a, view_b, _ = args
    assert np.array_equal(view_a.read(absolute_offset=(3072, 3072, 512), size=(32, 32, 32)),
                          view_b.read(absolute_offset=(3072, 3072, 512), size=(32, 32, 32)))
    print("They are equal!!")


if __name__ == "__main__":
    p = UPath('http://localhost:9000/data/zarr/sample_organization/l4_sample/', **{"headers": {"X-Auth-Token": "secretSampleUserToken"}})
    assert p.exists()

    disk = wk.Dataset.open("/Users/leo/Documents/work/scm/webknossos/binaryData/sample_organization/l4_sample/")
    http = wk.Dataset.open(p)

    mag_disk = disk.get_layer("color").get_mag("1-1-1")
    mag_http = http.get_layer("color").get_mag("1-1-1")
    mag_disk.for_zipped_chunks(_assert_check_equality, mag_http)
```

There are currently 4 issues on the client side: 1) `is_dir` is wrongly implemented and mostly returns false in fsspec (should be fixed in [UPath PR](https://github.com/fsspec/universal_pathlib/pull/58/files#)), 2) `encoding=utf-8` caused a problem, 3) `iterdir` in UPath uses a false implementation of `_sub_path` (should be fixed in [UPath](https://github.com/fsspec/universal_pathlib/pull/58/files#)), and 4) `is_file` check, doesn't check first whether file exists which leads to an exception (change first line of  `wk.WKWArray.open` to `if (path / "header.wkw").exists() and (path / "header.wkw").is_file():`

### Issues:
- fixes #6106 

------
- [x] Ready for review

------
**ToDo**
- [x] check with Zarr files
- [x] check with different mags
- [x] check with different layer names


webknossos-datastore/app/com/scalableminds/webknossos/datastore/controllers/BinaryDataController.scala
```diff
@@ -191,6 +191,142 @@ class BinaryDataController @Inject()(
...
+            "fill_value" -> 0,
+            "zarr_format" -> 2,
+            "order" -> "F",
+            "chunks" -> List(channels, cubeLength, cubeLength, cubeLength),
+            "compressor" -> Json
+              .obj("id" -> "lz4"), // TODO außer es ist uncompressed, lz4 hat vllt noch andere parameter
+            "filters" -> None,
+            "shape" -> List(
+              channels,
+              dataLayer.boundingBox.width + dataLayer.boundingBox.topLeft.x,
```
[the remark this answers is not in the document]
leowe: yes you are right

leowe:
@fm3 Can you have a look already? 

What doesn't seem to work yet is working with different mags... Somehow the request always leads to empty responses.

When substituting the VoxelPosition calculation in `requestRawZarr` with one that doesn't multiply by `cubeLength` and `mag`, it worked for higher mags, but not for Mag-1.

webknossos-datastore/app/com/scalableminds/webknossos/datastore/controllers/BinaryDataController.scala
```diff
@@ -191,6 +200,263 @@ class BinaryDataController @Inject()(
...
+                                 dataSource: DataSource,
+                                 dataLayer: DataLayer,
+                                 dataLayerName: String): Fox[Boolean] =
+    dataLayer.bucketProvider match {
+      case provider: WKWBucketProvider =>
+        WKWHeader(
+          provider
+            .wkwHeaderFilePath(mag, Some(dataSource.id), Some(dataLayerName), binaryDataService.dataBaseDir)
+            .toFile).flatMap((h) => Full(h.blockType == BlockType.LZ4))
+      case _: ZarrBucketProvider => ???
```
fm3, line 357: @normanrz Do you think the first version of this should already support both data backends? (Currently only wkw data can be exposed as zarr)

webknossos-datastore/app/com/scalableminds/webknossos/datastore/controllers/BinaryDataController.scala
```diff
@@ -191,6 +200,263 @@ class BinaryDataController @Inject()(
...
+    }
+
+  def requestRawZarrWithDot(
```
fm3, line 360: Why do we need to implement both routes (dot and slash)? I’d say we are in control which separator we want (since we generate the .zarray) – so we should be allowed to pick one, right?

fm3 (commented):
This is good stuff, thank you for pushing this so far already :)

I added some comments about what I noticed while first reading the code.

Regarding different mags, I did not immediately see an issue, though I think I would need a concrete test case to debug this. Maybe we can look at this together this week.
One thing that could be a culprit is the supplied coordinates. We should double check whether the requested coordinates need to still be in mag1 or not, and what the client that you tested, does.

webknossos-datastore/app/com/scalableminds/webknossos/datastore/controllers/BinaryDataController.scala
normanrz: WKW is must have, Zarr is nice to have for first version. But should be in there eventually.

webknossos-datastore/app/com/scalableminds/webknossos/datastore/controllers/BinaryDataController.scala
```diff
@@ -191,6 +200,263 @@ class BinaryDataController @Inject()(
     }
   }
 
+  def requestDatasourceFolderContents(token: Option[String],
```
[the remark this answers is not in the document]
normanrz: Eventually, we'll also need a directory listing for the mag folder that contains the "Zarr chunks". That should be derived from the available WKW files present on-disk.

webknossos-datastore/app/com/scalableminds/webknossos/datastore/controllers/BinaryDataController.scala
leowe: This was a hacky fix, because the client ignored the specified slash. Since the dot route is working fine, I will remove the slash route

leowe:
Since currently all buckets are uncompressed when serving them, we don't need to check the compression.
For reference for later:
```
  private def isWkwCompressed(mag: Vec3Int,
                              dataSource: DataSource,
                              dataLayer: DataLayer,
                              dataLayerName: String): Fox[Boolean] = {
    val result = dataLayer.bucketProvider match {
      case provider: WKWBucketProvider =>
        WKWHeader(
          provider
            .wkwHeaderFilePath(mag, Some(dataSource.id), Some(dataLayerName), binaryDataService.dataBaseDir)
            .toFile)
      case provider: ZarrBucketProvider => ???
    }

    result.flatMap((h) => Full(h.isCompressed))
  }
```

webknossos-datastore/app/views/datastoreZarrDatasourceDir.scala.html
```diff
@@ -0,0 +1,52 @@
...
+  <body>
+    <a href="." id="wkLink">
+      <img src="https://static.webknossos.org/images/oxalis.svg" alt="" />
+      <span>Go to webKnossos</span>
+    </a>
+    <p id="hint">This is the webKnossos @(kind) “@(name)” folder.</p>
+    <p>The following are the contents of the folder:</p>
+    <ul>
+      @folderContents.map { case (name, link) =>
+        <li><a href=@link>@name</a>: @link</li>
```
fm3, line 48: Is this some kind of standard format? I find it a bit hard to read for a human, (also when you click on the href in the browser, it cannot be displayed) but if this is what it should look like for the client, fine by me. This is not designed for humans anyway.

fm3 (commented):
This is getting somewhere :) I left some minor comments, but I’d say this will soon be good to go!
One more thing: please write follow-up issues for the things that were discussed as future work (e.g. sending 404 for chunks outside the real bounding box instead of empty 200 replies; handing through already-compressed buckets)

webknossos-datastore/app/views/datastoreZarrDatasourceDir.scala.html
leowe: The only thing that's strictly necessary is the <a href=@link> part, the rest I only added for debugging purposes.

fm3: fair enough :)

fm3 (approved):
:rocket:

Issue #6119, opened 2022-03-22
Zarr backend follow-up: tech


fm3:
Follow-up for reading zarr-datasets in wk backend:

  - [ ] cache clear for datavaults (when password for an account changes or the data vault breaks for some reason)
  - [ ] double-check that fill value is used if chunk does not exist
  - [ ] support string-encoded fill values
  - [x] perf: async cache, for file systems, handles – use akka-caching (caution: Futures all around, ~needs actorSystem~)
  - [x] do not depend on jzarr, instead on its dependencies itself (only those we use) https://github.com/scalableminds/webknossos/pull/6152
  - [x] where to store file systems / file system provider ref
  - [x] perf optimization: skip allocating buffer if single-bucket request

fm3:
I’d say with #6152 and #6165 the two most important parts of this issue are taken care of. Let’s schedule it again when we feel that the other checkboxes are relevant enough :)

fm3:
I split the remaining parts of this issue into  #6989 and #6990

PR #6152, opened 2022-04-19
Do not depend on jzarr, use jblosc, ucarCdm directly

fm3:
Since our code duplicates the functionality of jzarr, we do not need it as a dependency. We had kept that because we relied on *its* dependencies. Now we use those directly (only two are left, jblosc and ucar cdm)

I tested that zarr datasets can still be correctly viewed

contributes to https://github.com/scalableminds/webknossos/issues/6119

- [x] Ready for review


jstriebel (approved):
:+1:

Issue #6053, opened 2022-02-16
Consistent Terminology


jstriebel:
Since we agreed on a specific terminology (see comment below), the following steps need to be done to adapt the user-facing parts:
* [x] supply terminology overview in the docs (#6154)
* [x] adapt terminology usage throughout existing docs
* [x] adapt terminology usage in the frontend
* [ ] adapt terminology usage in wk-libs (part 1: https://github.com/scalableminds/webknossos-libs/pull/704, part 2: https://github.com/scalableminds/webknossos-libs/pull/706)

--------------------
## Detailed Description
The (user-facing) terminology throughout webKnossos could need some structuring and standardization. Atm there are multiple terms to describe the same thing, and also single terms describing various things (e.g. tracing). To get some consistency, at least in user-facing text (e.g. webknossos frontent, documentation, APIs), I'd propose to start an terminology overview, where terms are collected an either relevant docs are referenced or a short explanation is added. This overview could be part of the documentation, basically as an index for those terms.

Some examples that confused me at some point:
- volume-annotation, volume-layer, volume-annotation-layer, annotation-layer (which of those is what)
- annotation, tracing, skeleton(s), graphs, trees, groups, NML, Task (what is what)
- what is a branchpoint?
- data set vs dataset or datasource or data source
- chunk, file, bucket, shard, block
- explorative vs explorational
- mag vs magnification vs mip vs resolution vs zoomstep
- mag exponent vs index vs ?
- higher/lower mag vs finer/coarser?

Besides clarifying "default" names (and maybe abbreviations) internally, I think this would just be a great documentation for users by itself. Also dependencies between them might be useful, e.g. as an ER diagram:
```mermaid
erDiagram
    Project ||--|{ Task : has
    Project }o--|| Team : has
    Project ||--|| User : "owned by"
    Team }o--|{ User : "consists of"
    Task }o--o{ Annotation : has
    User ||--|| Annotation : annotates
```

fm3:
I agree that converging on canonical names (with explanations) would be extremely valuable 👍 
I’d say the dependency visualization is nice-to-have (but hard to get right)

jstriebel:
Notes from the discussion today:
* volume-annotation, volume-layer, volume-annotation-layer, annotation-layer (which of those is what)
* annotation, tracing, skeleton(s), graphs, trees, groups, NML, Task (what is what)
  -> long term goal: editable segmentation layer
  -> annotation: n annotation layers
[...]

PR #6159, opened 2022-04-20
Use mag in data requests, no more zoomStep/exponent

fm3:
- use mag instead of zoomStep in data requests
- start renaming some resolution and zoomStep to mag in the backend internally (some places still use resolution)
- remove unused datastore routes
- clean up thumbnail management (no more json with base 64 encoding)
- fixed the snapshot tests, see discussion below

- https://datarequestmag.webknossos.xyz

### TODO
 - [x] frontend should send mag when requesting 
   - [x] data (normal bucket request) (also in volume annotations)
   - [x]  data (rawCuboid)
   - [x]  isosurfaces
 - [x] wkconnect should also be adapted https://github.com/scalableminds/webknossos-connect/pull/129
 - [x] bump datastore api version

### Follow-Up
 - wk python client should also be adapted (uses rawCuboid GET) https://github.com/scalableminds/webknossos-libs/issues/719 – this is currently backwards compatible, but I want to remove that optional parameter here eventually, tracked in https://github.com/scalableminds/webknossos/issues/6189

### Steps to test:
- Load some data, isosurfaces
- use rawCuboid via frontend API, request dataset thumbnail

### Issues:
- contributes to #6053 
- fixes https://github.com/scalableminds/webknossos/issues/4893

------
- [x] Adapted [wk-connect](https://github.com/scalableminds/webknossos-connect) if datastore API changes
- [x] Ready for review


daniel-wer:
@fm3 I'm not sure whether I got everything, but I switched the bucket request, the ad-hoc isosurface computation request, and the data cuboid request code. In my testing all of these work as expected, again. However, I'm not sure whether there's more interfaces that have changed.

daniel-wer:
I also just noticed that the snapshot tests were not doing anything, because the snapshots were missed to move when the test directory was changed in https://github.com/scalableminds/webknossos/pull/6065 /cc @philippotto. I corrected that, let's see whether any errors pop up.

jstriebel (commented):
Awesome stuff, huge fan of the restructured datastore API :heart:
The backend changes look very good to me, props for using mags consistently, this helps readability in several places!

frontend/javascripts/oxalis/api/api_latest.ts
```diff
@@ -1459,6 +1460,33 @@ class DataApi {
     return result;
   }
 
+  /**
+   * Helper method to build the download URL for a raw data cuboid.
+   *
+   * @ignore
+   */
+  _getDownloadUrlForRawDataCuboid(
```
MichaelBuessemeyer, line 1468: Nice refactoring 👍

frontend/javascripts/oxalis/model/accessors/dataset_accessor.ts
```diff
@@ -167,6 +181,16 @@ export class ResolutionInfo {
     return Math.log2(this.getLowestResolutionPowerOf2());
   }
 
+  getHighestResolution(): Vector3 {
+    // @ts-ignore
+    return this.getResolutionByPowerOf2(this.getHighestResolutionPowerOf2());
+  }
```
MichaelBuessemeyer, line 187: I just noticed that this method isn't used anywhere. Only `getLowestResolution` is being used. 

But I think it is fine to keep this method in case someone wants the highest resolution in the future.

MichaelBuessemeyer (approved):
The frontend part looks fine to me 👍  🔍  🚀

webknossos-datastore/app/com/scalableminds/webknossos/datastore/controllers/BinaryDataController.scala
```diff
@@ -112,15 +101,21 @@ class BinaryDataController @Inject()(
       @ApiParam(value = "Target-mag width of the bounding box", required = true) width: Int,
       @ApiParam(value = "Target-mag height of the bounding box", required = true) height: Int,
       @ApiParam(value = "Target-mag depth of the bounding box", required = true) depth: Int,
-      @ApiParam(value = "Exponent of the dataset mag (e.g. 4 for mag 16-16-8)", required = true) resolution: Int,
+      @ApiParam(value = "Mag in three-component format (e.g. 1-1-1 or 16-16-8)", required = true) mag: Option[String],
+      @ApiParam(value = "[DEPRECATED: use mag instead] exponent of largest dimension of mag.", required = true) zoomStep: Option[
```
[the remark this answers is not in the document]
fm3: Good point :) Just noticed I had accidentally renamed the resolution parameter here to zoomStep, thus not providing any backwards compatibility at all :laughing: undid that now and removed the annotation as you suggested

PR #6162, opened 2022-04-22
Allow to interpolate segments for faster annotation

### LINKED ISSUE #5434: Allow volume interpolation between sections
## Detailed Description
Interpolation between two volume-annotated sections can enable a user to only annotate each n-th section. Therefore, increasing volume annotation speed by a factor of n (the trade-off is of course less precision which makes an additional proofreading pass necessary).

## Context
This would strengthen webKnossos' position as a fast annotation tool. Often high-precision is crucial, but there are also lot of cases where a trade-off in quality is perfectly fine.

## Behavior/Implementation
We already have the "copy segmentation from previous layer" functionality (mapped on `v`). We could leverage the same feature and extend it so that it also performs interpolation. An example workflow could be:
- user annotates slice n
- user moves forward (f) twice
- user annotates slice n + 2
- ...
- user annotates slice n + 20
- user goes back to slice n + 1
- user hits V to copy and interpolate the segmentation from the adjacent slices
- user moves forward to n + 3
- ...

This workflow might look cumbersome at first, but I think, that it's easy to grasp what happens. Also, the second pass should be made, anyway, by the user to proof-read the annotation. Further improvements could be made in later iterations.

## Effort
`v` would need to be adapted to take into account both sides of the current slice and interpolate between them. For now, we could limit this to only the immediate neighbors (so, only annotating every 2nd slice is supported). Therefore, this should be quite easy to implement. One would probably need to construct distance-maps for the two slices and interpolate between those.
Maybe 1d - 2d.

philippotto:
- implements annotating every 2nd slice so that the missing slices are interpolated
  - only works in XY (I think, this is also fine for now since this is the highest quality, anyway)
  - only when labeling (not erasing) <-- also intended
  - details:
    - a signed distance transform is computed for slice 1 and 3
    - the mean of these distances is computed and values below 0 are labeled then
- types the mjs / V2 / V3 module "properly" (ignore whitespace when reviewing)
- refactors the volume saga a bit (the helper module is new and only consists of moved code)
- misc (type) fixes
- Changes default of `dynamicSpaceDirection` property to false to avoid confusion.
- add `volumeInterpolationAllowed` to task settings which is disabled by default to avoid that tasks are accidentally done with interpolation

- https://interpolatesegments.webknossos.xyz

### Steps to test:
- create a new annotation
- switch to brush or trace tool
- enable the interpolation feature (see new button in toolbar)
- label every 2nd slice
- every (2nd+1) slice should be filled out automatically
- task handling
  - create a task type with volume enabled and volume interpolation allowed (default should be disallowed)
  - also enable the recommended settings and set volume interpolation to true (should be listed in the default recommended settings)
  - create a new task with the type from above
  - get task in dashboard
  - ensure that interpolation feature is enabled due to the recommendation and that it can be used
  - edit task type to allow interpolation
  - reload task annotation from above
  - ensure that interpolation feature cannot be used

### Todo:
- [x] make it work with other directions (i.e., decreasing z)
- Potentially as a follow-up
  - [x] multi-mag support
    - [X] make zoomed out interpolation work in 2-2-1
    - [x] make interpolation in 2-2-1 work when annotation is restricted to 2-2-1 (still buggy, could be related to https://github.com/scalableminds/webknossos/issues/6138)
    - [x] make interpolation work in 4-4-2 (i.e., z is not 1)
  - [x] interpolation depth > 2
    - [X] implement support
    - ~[ ] enable in UI (I think, we need to think about the details here; for example: what happens if I set the interpolation depth to 4 and then only go 3 slices forward? the interpolation will then look 4 slices back which is suboptimal for multiple reasons)~ I created https://github.com/scalableminds/webknossos/issues/6183 for this
  
### Issues:
- fixes #5434 

------
- [x] Ready for review


frontend/javascripts/test/model/model_resolutions.spec.ts
```diff
@@ -28,10 +28,10 @@ test("Complex convertToDenseResolution", (t) => {
       dataLayers: [
         {
           resolutions: [
+            [16, 16, 2],
```
philippotto, line 31: this test covers now unsorted resolutions, too.

frontend/javascripts/oxalis/model/volumetracing/volume_annotation_sampling.ts
```diff
@@ -344,16 +344,11 @@ export function applyVoxelMap(
...
+        const nextBucketZoomedAddress: Vector4 = [...labeledBucketZoomedAddress];
         nextBucketZoomedAddress[thirdDimensionIndex]++;
         postprocessBucket(bucket);
-        // @ts-expect-error ts-migrate(2345) FIXME: Argument of type 'number[]' is not assignable to p... Remove this comment to see the full error message
         bucket = dataCube.getOrCreateBucket(nextBucketZoomedAddress);
         preprocessBucket(bucket);
-
-        if (bucket.type === "null") {
-          continue;
-        }
```
philippotto, line 356: this already happens directly after the if.

philippotto:
@daniel-wer Maybe you already want to have a look at this PR? Also from a usability point of view..
I'm still having some trouble with the multi-mag support (see PR description), but maybe I'll simply restrict this feature to mag 1 for now. Then the PR is not blocked, anymore. What do you think?

The only bit missing is the "make it work with other directions (i.e., decreasing z)". However, the changes shouldn't be that big.

frontend/javascripts/oxalis/model/sagas/volume/volume_interpolation_saga.ts
```diff
@@ -0,0 +1,257 @@
...
+  const isVolumeInterpolationEnabled = yield* select(
+    (state) => state.userConfiguration.isVolumeInterpolationEnabled,
+  );
+
+  if (!isVolumeInterpolationEnabled) {
+    return;
+  }
+
+  if (activeViewport !== "PLANE_XY") {
+    // Interpolation is only done/supported in XY
```
daniel-wer, line 116: One open question is how this could be communicated to the user. Since you're showing a toast if the resolution is too low, the same could be done here, maybe

frontend/javascripts/oxalis/model/bucket_data_handling/bounding_box.ts
```diff
@@ -131,23 +130,56 @@ class BoundingBox {
...
+    return new BoundingBox({
+      min: V3.sub(this.min, marginsLeft),
+      max: V3.add(this.max, marginsRight),
+    });
+  }
+
+  /*
+   * Each component of margins is used as
+   *   - a left margin IF the value is positive
+   *   - a right margin IF the value is negative (the absolute value will be
```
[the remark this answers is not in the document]
philippotto: I thought about the semantics of negative values for a while when writing this and wasn't really sure. In the end, I picked the variant which didn't require a `-1` on the calling side. But if you expected it to be the other way round, that's a good indicator to do so. I'm changing it :)

daniel-wer (commented):
Great job completing all those potential follow-ups in this PR already! Code LGTM, I'll test again on Monday, once the newest version is deployed.

CHANGELOG.unreleased.md
```diff
@@ -11,8 +11,11 @@ For upgrade instructions, please check the [migration guide](MIGRATIONS.released
 [Commits](https://github.com/scalableminds/webknossos/compare/22.05.0...HEAD)
 
 ### Added
+- Added Volume Interpolation feature. When enabled, it suffices to only label every 2nd slice. The skipped slices will be filled automatically by interpolating between the labeled slices. The little arrow indicates whether you are currently labeling with increasing or decreasing X/Y/Z. This feature is disabled by default (and also forbidden by default for tasks). [#6162](https://github.com/scalableminds/webknossos/pull/6162)
```
```suggestion
- Added Volume Interpolation feature. When enabled, it suffices to only label every 2nd slice. The skipped slices will be filled automatically by interpolating between the labeled slices. This feature is disabled by default. Note that the feature is even forbidden for tasks by default, but can be enabled/recommended. [#6162](https://github.com/scalableminds/webknossos/pull/6162)
```

CHANGELOG.unreleased.md
```diff
@@ -11,8 +11,11 @@ For upgrade instructions, please check the [migration guide](MIGRATIONS.released
...
 
 ### Changed
+- The mouse cursor now changes appearance to indicate the currently selected annotation tool  [#6132](https://github.com/scalableminds/webknossos/pull/6132)
```
philippotto, line 17:
```suggestion
```

merge error

frontend/javascripts/oxalis/model/bucket_data_handling/bounding_box.ts
```diff
@@ -67,20 +67,16 @@ class BoundingBox {
...
-      Math.min(this.max[1], other.max[1]),
-      Math.min(this.max[2], other.max[2]),
-    ];
+    const newMin = V3.max(this.min, other.min);
+    const uncheckedMax = V3.min(this.max, other.max);
+
+    // Ensure the bounding box does not get a negative
+    // extent.
+    const extent = V3.max(V3.sub(uncheckedMax, newMin), [0, 0, 0]);
+    const newMax = V3.add(newMin, extent);
```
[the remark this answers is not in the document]
philippotto: Ah, very good :+1: It felt a bit weird to me when writing it :laughing:

conf/evolutions/082-annotationsettings-volumeInterpolationAllowed.sql
```diff
@@ -0,0 +1,12 @@
+-- https://github.com/scalableminds/webknossos/pull/???
+
```
fm3, line 2:
```suggestion
```
I have come to believe that git blame should actually be enough to link evolutions back to PRs :) Let’s skip looking up and inserting the PR number here from now on. (No need to modify old evolution files, though)

fm3 (commented):
Backend LGTM! 💪

daniel-wer (commented):
Changes LGTM.

Feedback from testing:
- [x] It's not possible to create a task type where volume interpolation is allowed. The frontend always seems to send `volumeInterpolationAllowed: false` to the server when creating/editing a task type, regardless of the actual state of the setting.
- [x] Volume interpolation also cannot be enabled in an explorative annotation any longer since the server sends `volumeInterpolationAllowed: false` in the info request which disables the button. I think this was caused by a change in the defaults in one of the latest changes.

fm3 (approved):
Works for me 🚀

PR #6165, opened 2022-04-25
Asynchronous cache for zarr chunk loading

fm3:
Switches the chunk cache inside zarr array implementation from a synchronous to asynchronous one. This means that concurrent incoming requests that are all cache misses do not each trigger chunk fetching. Only one does and the other requests hook into that.

I use the caching library from akka for this. Side effects include:
 - The new cache is no longer LRU but LFU (frequency based). I do not think this will have much noticeable impact.
 - The new cache has expiry times, which I think is a nice addition made simple by the new akka cache
 - The signatures of many functions had to becomes asynchronous, which makes the code a bit less readable but should be acceptable. It also paves the way for other asynchronous action
 - The exception handling inside ZarrBucketProvider could no longer use the `tryo` shortcut, instead I use `Fox.recover` to achieve the same effect.

- https://zarralru.webknossos.xyz

### Steps to test:
- Load remote zarr dataset, ideally one with very different chunk shape from the requested 32² (talk to me about a datasource-properties.json)
- The first few requests should cause a lot less load, be answered more quickly.
- wkw datasets and local zarr files should still load

### Issues:
- contributes to #6119 

------
- [x] Ready for review

### DIFF
--- a/CHANGELOG.unreleased.md
+++ b/CHANGELOG.unreleased.md
@@ -19,6 +19,7 @@ For upgrade instructions, please check the [migration guide](MIGRATIONS.released
   - Add filter functionality to "Access Permissions" column to filter for public datasets.
   - Removed `isActive` and `isPublic` columns to save screen space.
   - Changed data layer entries to display layer names instead of categories, e.g. "color" --> "axons".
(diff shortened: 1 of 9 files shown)


jstriebel (approved):
LGTM :+1:

PR #6167, opened 2022-04-26
In dataset upload, allow un-sorted anisotropic mags in properties.json

fm3:
In datasource validation there is a check asserting that the mags are consistent (meaning that all dimensions increase monotonically, you cannot have 1-1-2 AND 2-2-1). This check previously relied on the mags also being IN ORDER in the properties.json. This PR removes this latter requirement (I think it was there only accidentally). We double-checked that there are no important code paths that rely on this order.

### Steps to test:
- Edit a datasource-properties.json from the dataset settings page, try saving it (disable frontend validation?)
- inconsistent mags should still be forbidden but any ordering should be fine

------
- [x] Ready for review


jstriebel (approved):
:+1:

PR #6171, opened 2022-04-27
Create new export modal

### LINKED ISSUE #5672: Build export dialog
## Detailed Description
Currently, clicking on "Download" within the dropdown menu for an open annotation simply outputs a file (either NML or zip) and the user has to figure out how to deal with it. We could build an export modal which (implicitly) explains some stuff and provides options:

- always show export/download modal (or only when there are indeed multiple options? I'd argue for always showing the modal and adding tooltips for disabled stuff)
- modal should provide options/checkboxes, such as:
  - 1) Download Volume as wkw
  - 2) Download Volume as Tiff (if available)
  - 3) Skeleton as NML
  - 5) Other Layers as wkw/tiff
- additionally, available bounding boxes could be selectable to restrict the volume export (skeleton will be untouched; the UI should reflect this)
- link to wk (lib) docs to guide the user for further processing of the exported data

We already have a similar modal for exporting bounding boxes. We could re-use and overhaul that modal so that it can be used from both places.

## Context

For (new) users, the question of "how do I get my data?" always comes up sooner or later. Currently, the UI doesn't really provide a big support for answering this. Improving the UX on this matter could go a long way.

### LINKED ISSUE #6059: Allow to restrict magnifications when adding new volume layer
## Detailed Description
When using the "Add Volume Layer" button one cannot choose whether/how the magnifications should be restricted (as it can be done when using the ... menu in the dashboard's explorative list).

The front-end could re-use the component from the dashboard. Not sure whether the back-end already accepts the payload in the request.

Dagobert42:
https://createnewexportmodal.webknossos.xyz/

### Steps to test:
- open an annotation and select Download from the tracing actions
- the new modal should show with tabs for Download, Export and Python Client

### Issues:
- fixes #5672 

------
- [ ] Ready for review


frontend/javascripts/messages.ts
```diff
@@ -290,6 +290,12 @@ instead. Only enable this option if you understand its effect. All layers will n
   "annotation.delete": "Do you really want to reset and cancel this annotation?",
   "annotation.was_edited": "Successfully updated annotation",
   "annotation.shared_teams_edited": "Successfully updated the sharing options for the annotation",
+  "annotation.download":
+    "The options in this tab are available for download immediately. If you would like to prepare data for donwload in TIFF format, please switch to the export tab.",
+  "annotation.export":
+    "Exporting will trigger a job to prepare TIFF data for download. This may take a while depending on the data you select. When the job has finished you can download your selected data in the download list.",
+  "annotation.python":
+    "The following code snippets are suggestions to get you started quickly with the wklibs python client. To download and use the dataset in your python project just copy and paste the code snippets to your script.",
```
hotzenklotz, line 298:
```suggestion
    "The following code snippets are suggestions to get you started quickly with the webKnossos Python client. To download and use the dataset in your Python project just copy and paste the code snippets to your script.",
```

frontend/javascripts/oxalis/view/action-bar/download_modal_view.tsx
```diff
@@ -0,0 +1,332 @@
...
+          </Row>
+          <Divider
+            style={{
+              margin: "18px 0",
+            }}
+          />
+          {moreInfoHint}
+        </TabPane>
+
+        <TabPane tab="Export" key="2">
```
hotzenklotz, line 190: Why not name it `TIFF Export` to make this clear without having to read the small print first.

frontend/javascripts/oxalis/view/action-bar/download_modal_view.tsx
```diff
@@ -0,0 +1,332 @@
...
+                lineHeight: "20px",
+                padding: "5px 12px",
+              }}
+            >
+              Select the data you would like to download.
+            </Col>
+            <Col span={15}>
+              <CheckboxGroup onChange={handleCheckboxChange} defaultValue={selection}>
+                <Checkbox style={checkboxStyle} value="Volume">
+                  Volume as WKW
```
hotzenklotz, line 145:
```suggestion
                  Volume Annotation as WKW
```

frontend/javascripts/oxalis/view/action-bar/download_modal_view.tsx
```diff
@@ -0,0 +1,332 @@
...
+                  style={{
+                    marginLeft: 24,
+                    marginBottom: 12,
+                  }}
+                >
+                  Download a zip folder containing WKW files.
+                </Hint>
+
+                <Checkbox style={checkboxStyle} value="Skeleton">
+                  Skeleton as NML
```
hotzenklotz, line 157:
```suggestion
                  Skeleton Annotations as NML
```

frontend/javascripts/oxalis/view/action-bar/download_modal_view.tsx
```diff
@@ -0,0 +1,332 @@
...
+                  style={{
+                    marginLeft: 24,
+                    marginBottom: 12,
+                  }}
+                >
+                  Download a zip folder containing NML files.
+                </Hint>
+
+                <Checkbox style={checkboxStyle} value="Fallback">
+                  Fallback Layers
```
hotzenklotz, line 169: This should only be selectable / enabled if a fallback layer is part of the DS.

hotzenklotz: Also I am not quite sure if I understand this feature. This would likely download a whole segmentation layer, right?

frontend/javascripts/oxalis/view/action-bar/download_modal_view.tsx
```diff
@@ -0,0 +1,332 @@
...
+
+  const moreInfoHint = (
+    <Hint
+      style={{
+        margin: "0px 12px 0px 12px",
+      }}
+    >
+      For more information on how to process downloaded layers visit the{" "}
+      <a href="https://docs.webknossos.org" target="_blank" rel="noreferrer">
+        user documentation...
```
hotzenklotz, line 87:
```suggestion
        user documentation.
```

frontend/javascripts/oxalis/view/action-bar/download_modal_view.tsx
Dagobert42: This was a suggestion from me (not fully understanding the extent that fallback layers could have) and will be removed.

Dagobert42:
Thanks for proof-reading and the suggestions @hotzenklotz ! I will adapt them asap.

fm3:
Note that the back-end already supports an optional boolean `skipVolumeData` in the annotation download route. 
TODO:
 - [ ] also omit the `<volume>` tags in the NML file in this case, replace them by explaining comment

frontend/javascripts/oxalis/view/action-bar/download_modal_view.tsx
fm3: The exception being the tiff export job (webknossos-worker feature). That one does actually merge the volume annotation with the fallback layer. The trick here is that the user has to select a small-ish bounding box for the tiff export so that they do not create a huge load for job and download. This limit is currently set to 1024 Mvx. (can be configured in application.conf and queried in api/features)

frontend/javascripts/oxalis/view/action-bar/download_modal_view.tsx
```diff
@@ -0,0 +1,463 @@
...
+export default function DownloadModalView(props: Props): JSX.Element {
+  const { isVisible, onClose, annotationType, annotationId, hasVolumeFallback } = props;
+  const userBoundingBoxes = useSelector((state: OxalisState) =>
+    getUserBoundingBoxesFromState(state),
+  );
+  const [activeTabKey, setActiveTabKey] = useState("download");
+  const [includeVolumeData, setIncludeVolumeData] = useState(true);
+  const [keepWindowOpen, setKeepWindowOpen] = useState(false);
+
+  // const layers = chooseSegmentationLayer ? getSegmentationLayers(dataset) : getColorLayers(dataset);
```
philippotto, line 116: remove probably

frontend/javascripts/oxalis/view/action-bar/download_modal_view.tsx
```diff
@@ -0,0 +1,463 @@
...
+  const isMergerModeEnabled = useSelector(
+    // @ts-expect-error ts-migrate(2339) FIXME: Property 'temporaryConfiguration' does not exist o... Remove this comment to see the full error message
+    (state) => state.temporaryConfiguration.isMergerModeEnabled,
+  );
+  const activeMappingInfos = useSelector(
+    // @ts-expect-error ts-migrate(2339) FIXME: Property 'temporaryConfiguration' does not exist o... Remove this comment to see the full error message
+    (state) => state.temporaryConfiguration.activeMappingByLayer,
+  );
+
+  const [selectedLayerName, setSelectedLayerName] = useState("");
```
philippotto: better than defaulting to an empty string would be defaulting to null. then, the typechecker will complain if the null case is not handled somewhere. I'll change this in a commit, since I also saw some other small type issues which are easier to fix than to explain ^^

frontend/javascripts/oxalis/view/action-bar/download_modal_view.tsx
```diff
@@ -0,0 +1,463 @@
...
+    )
+`;
+
+  const alertTokenIsPrivate = () => {
+    Toast.warning(
+      "The clipboard contains private data. Do not share this information with anyone you do not trust!",
+    );
+  };
+
+  const selection = ["Volume", "Skeleton"];
```
philippotto, line 258: I think, you can inline this variable as it's only used once.

frontend/javascripts/oxalis/view/action-bar/download_modal_view.tsx
```diff
@@ -0,0 +1,463 @@
...
+      >
+        {messages["annotation.export_no_worker"]}
+        <a href="mailto:hello@webknossos.com">hello@webknossos.com.</a>
+      </Text>
+    </Row>
+  );
+
+  const checkboxStyle = {
+    height: "30px",
+    lineHeight: "30px",
```
philippotto, line 238:
```suggestion
    height: 30,
    lineHeight: 30,
```

frontend/javascripts/oxalis/view/action-bar/download_modal_view.tsx
```diff
@@ -0,0 +1,463 @@
...
+                  style={{
+                    lineHeight: "20px",
+                    padding: "5px 12px",
+                  }}
+                >
+                  Select the layer you would like to prepare for export.
+                </Col>
+                <Col span={15}>
+                  {/* // if hasColorLayer(s) 
+                  from props.chooseSegmentationLayer*/}
```
philippotto, line 371: can be removed?

frontend/javascripts/oxalis/view/action-bar/download_modal_view.tsx
```diff
@@ -0,0 +1,463 @@
...
+                  style={{
+                    lineHeight: "20px",
+                    padding: "5px 12px",
+                  }}
+                >
+                  Select a bounding box to constrain the data for export.
+                </Col>
+                <Col span={15}>
+                  <BoundingBoxSelection
+                    // from props.isBoundingBoxConfigurable
```
philippotto, line 399: remove?

frontend/javascripts/oxalis/view/action-bar/download_modal_view.tsx
```diff
@@ -0,0 +1,463 @@
...
+                padding: "5px 12px",
+              }}
+            >
+              Select the data you would like to download.
+            </Col>
+            <Col span={15}>
+              <CheckboxGroup onChange={handleCheckboxChange} defaultValue={selection}>
+                <Checkbox style={checkboxStyle} value="Volume" checked={includeVolumeData}>
+                  Volume Annotations as WKW
+                </Checkbox>
```
philippotto, line 302: This checkbox shouldn't be there in case of skeleton-only annotations (you can use `hasVolumeTracings(state.tracing)` for that).

frontend/javascripts/oxalis/view/action-bar/download_modal_view.tsx
```diff
@@ -0,0 +1,463 @@
...
+                  style={{
+                    marginLeft: 24,
+                    marginBottom: 12,
+                  }}
+                >
+                  Download a zip folder containing WKW files.
+                </Hint>
+
+                <Checkbox style={checkboxStyle} value="Skeleton" checked disabled>
+                  Skeleton Annotations as NML
```
philippotto, line 313: In case of volume-only annotations this should say `Meta data as NML` (check for `state.tracing.skeleton == null`).

philippotto (changes requested):
Cool stuff! Already looks pretty good. I left a couple of comments which should be tackled before merging, but nothing major.

I pushed a small commit to improve some typing, since I saw some `never` annotations which didn't seem quite right (and I thought it would be quicker to fix than to explain it :)). You can have a look at it here: 1016d85ec

frontend/javascripts/messages.ts
```diff
@@ -105,7 +105,7 @@ Editing should be done in a single window only.
 
 In order to restore the current window, a reload is necessary.`,
   "react.rendering_error":
-    "Unfortunately, we encountered an error during rendering. We cannot guarantee that your work is persisted. Please reload the page and try again.",
+    "Unfortunately, we encountered an error during rendering. We cannot guarantee that your work endured. Please reload the page and try again.",
```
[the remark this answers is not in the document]
fm3: How about `Unfortunately, webKnossos encountered an error during rendering. Your latest changes may not have been saved. Please reload the page to try again.`

frontend/javascripts/oxalis/view/action-bar/download_modal_view.tsx
```diff
@@ -0,0 +1,461 @@
...
+  const authToken = useFetch(getAuthToken, "loading...", []);
+  const wkInitSnippet = `import webknossos as wk
+
+with wk.webknossos_context(token="${authToken}"):
+    annotation = wk.Annotation.download(
+        "${annotationId}",
+        annotation_type="${annotationType}",
+        webknossos_url="${window.location.origin}"
+    )
+`;
```
jstriebel, line 250: Just double-checked the code, we should move the url in the context, then everything works :tada:
```suggestion
with wk.webknossos_context(
    token="${authToken}",
    url="${window.location.origin}"
):
    annotation = wk.Annotation.download(
        "${annotationId}",
        annotation_type="${annotationType}",
    )
`;
```

frontend/javascripts/oxalis/view/action-bar/download_modal_view.tsx
Dagobert42: I've left this as is, since it leads to the Checkbox suddenly spacing weirdly.
Screenshot 2022-05-18 at 15 57 28

Dagobert42:
@MichaelBuessemeyer thanks for the help with  [b5bb9c8](https://github.com/scalableminds/webknossos/pull/6171/commits/b5bb9c8ce4caeb8cd7f6cf52c6d3e22ca881b6dc) :)

jstriebel (commented):
Just tested the Python snipplet for annotation download, works like a charm :sparkles:

Screenshot from 2022-05-18 16-40-09

philippotto (approved):
Great work :) Please update the changelog before merging, though. I also pushed a few changes, since I noticed that the size of the bounding boxes is not validated in the modals form.

It's not super pretty, but gets the job done:

image

PR #6177, opened 2022-04-28
More swagger annotations for dataset routes

jstriebel:
This PR adds more swagger annotations for the update, updateTeams and sharingToken routes on datasets for usage in wk-libs.

### Steps to test:
- Mostly double-checking that the descriptions and params look right.

### Issues:
- fixes #

------
- [x] Ready for review


fm3 (approved):
LGTM. When implementing the client, note that this particular update method will set attributes to None in the database if they are null or absent in the request body. I believe this is not currently consistent across different update routes in the API.

PR #6186, opened 2022-04-29
Don't remove other volume layers when opening volume annotation

### LINKED ISSUE #6143: Don't remove other volume layers when opening volume annotation
See #5695  for context.

philippotto:
When creating a new annotation with a volume layer (without fallback) for a dataset which has an existing segmentation layer:
- [X] keep on-disk segmentation layers in the annotation so that they can be activated
- [X] hide on-disk segmentation layer on first view and make new volume layer visible
- [x] ensure that changing the layer visibility of a layer in an annotation, does not change the layer visibility for the dataset settings. otherwise, opening the dataset would likely show no segmentation layer.

- https://keepseglayers.webknossos.xyz

### Steps to test:
- Open a dataset with a segmentation layer (keep that dataset in an own tab)
- Hide the segmentation layer & refresh -> layer should still be hidden
- Unhide the segmentation layer & refresh -> layer should still be shown
- Open a second tab to create an annotation for that dataset via the dashboard (use the advanced menu) and select that a new segmentation should be created
- The empty segmentation layer should be visible by default. Also, the original segmentation layer should still be listed in the left sidebar (but hidden by default).
- Reload the annotation -> visibilities shouldn't have changed (i.e., the original segmentation layer should be invisible)
- Reload the dataset in tab 1 -> visibilities shouldn't have changed there (i.e., the original segmentation layer should still be visible).
- Make the original segmentation layer visible in tab 2 and reload --> original segmentation layer should still be visible.

### Issues:
- fixes #6143

------
- [X] Ready for review


frontend/javascripts/oxalis/model/sagas/settings_saga.ts
```diff
@@ -36,6 +36,7 @@ function* pushDatasetSettingsAsync(): Saga<void> {
       SETTINGS_RETRY_DELAY,
       updateDatasetConfiguration,
       dataset,
+      // mask updates to layer visibility here if this is an annotation?
```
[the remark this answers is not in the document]
fm3: I agree that both options are less than beautiful 🤔 I’d say front-end masking/merging that into the request makes sense. The back-end api request is about the dataset configuration, which should stay constant. 

We could also implement the logic in the back-end that fields missing from the request are just not overwritten in the back-end, but this does not work if there are any fields which are actually supposed to be settable to null/none. Also it would require a bit of work in the back-end since it currently just stores the json in the db, just as it is sent by the frontend.

So I’d say if it is feasible to patch this into the request in the front-end, that’s probably the best option.

frontend/javascripts/oxalis/model_initialization.ts
```diff
@@ -479,7 +497,7 @@ function setupLayerForVolumeTracing(
   // layers, we cannot render both at the same time. Hiding the existing segmentation
   // layer would be good, but this information is stored per dataset and not per annotation
   // currently. Also, see https://github.com/scalableminds/webknossos/issues/5695
-  const newLayers = originalLayers.filter((layer) => layer.category !== "segmentation");
+  const newLayers = originalLayers.slice(); // .filter((layer) => layer.category !== "segmentation");
```
daniel-wer, line 500: The comments behind and above this line (also the one in line 490) can be removed now, I think.

philippotto:
Thanks for your review!

> Since this functionality is somewhat brittle, would it be a good candidate for a snapshot test using the e2e test setup? One could create a new annotation without a fallback layer, and would test that the visibility and number of the layers is correct.

Do you mean the e2e tests or the screenshot tests? The e2e tests mainly test the REST API between JS and Scala which doesn't really lend itself to test the behavior introduced in this PR (since the main logic happens in wk JS core). However, I just adapted the screenshot tests to also create a hybrid annotation for a dataset (once with a fallback layer and once without). This tests the basic setup of the default visibilities quite well, I think.

Testing that the visibilities are changed correctly would be nice, too. However, I don't see an easy way to do so, since this logic happens in the sagas. Interacting with the page via puppeteer sounds too messy/brittle and calling the sagas manually without the surrounding page context is also not really lucrative in my opinion. This is why I hope that the two new screenshot tests are a good trade-off for now. Do you agree?

.circleci/config.yml
```diff
@@ -292,6 +292,10 @@ workflows:
   version: 2
   circleci_build:
     jobs:
+      - nightly:
```
daniel-wer, line 295: Reminder to remove this before merging

frontend/javascripts/test/puppeteer/dataset_rendering.screenshot.ts
```diff
@@ -182,6 +191,51 @@ datasetNames.map(async (datasetName) => {
...
+            height,
+            BASE_PATH,
+            `annotation_${datasetName}_${fallbackLayerName}`,
+          );
+          return isPixelEquivalent(changedPixels, width, height);
+        },
+        (condition) => {
+          t.true(
+            condition,
+            `Dataset with name: "${datasetName}" does not look the same, see ${datasetName}.diff.png for the difference and ${datasetName}.new.png for the new screenshot.`,
```
daniel-wer, line 231:
```suggestion
            `Annotation for dataset with name: "${datasetName}" does not look the same, see annotation_${datasetName}_${fallbackLayerName}.diff.png for the difference and annotation_${datasetName}_${fallbackLayerName}.new.png for the new screenshot.`,
```

daniel-wer (approved):
LGTM :+1: 

> Do you mean the e2e tests or the screenshot tests?

I meant the e2e tests, but I overlooked that the frontend doesn't run in that case. Thanks for pointing that out. In that case, I think the screenshot testing is the best we can do right now.

PR #6191, opened 2022-05-04
Replace “Trace” by ”Open” in annotations list

fm3:
Replace the word “Trace” by “Open” in annotations list. I’d say trace is non-standard (we mostly use annotate now), but additionally, one may want to open an annotation to look at it but not actually work on it.

- https://tracevsopen.webknossos.xyz

### Steps to test:
- Open dashboard, statistics view, task type create view, texts should make sense (and avoid the word trace where not specifically talking about the trace tool as opposed to the brush tool)

------
- [x] Checked that docs do not contain references to this
- [x] Ready for review


philippotto (approved):
:+1:

PR #6195, opened 2022-05-09
Editable Mappings aka Supervoxel Proofreading

### LINKED ISSUE #6139: Easier proofreading of volume data
https://www.notion.so/scalableminds/Design-Doc-Supervoxel-Proof-Reading-e012fb444d1242fdacaaa1c9d498f285

### LINKED ISSUE #6228: Unify async cache usages, refactor
The async akka cache  can be used in a number of places in the backend. Since it requires quite a bit of boilerplate to set up, we should unify the usages and abstract that away. Also, could we provide convenience wrappers that deal with Fox-wrapped results?

fm3:
Work-in-Progress Prototype for Editable Mappings a.k.a Supervoxel-Based Proofreading

Compare notes at https://www.notion.so/scalableminds/Design-Doc-Supervoxel-Proof-Reading-e012fb444d1242fdacaaa1c9d498f285

### Assumptions for Prototype
 - no undo/redo support for now
 - no mutating volume buckets in editable-mapping annotations
 - ~no evolutions / backwards compatibility~
 - ~only one datastore per tracingstore~
 - no support in webknossos-connect
 - no uint64 segmentations (can’t operate directly on Long then. Also, those are currently forbidden for volume annotations anyway.)
 - a valid hdf5 agglomerate file exists for the fallback layer
 - no export/download/copy to my account
 
### TODO Frontend
 - [x] Hide proofread button if no agglomerate view exists
 - [x] Make proofreading work for non-uint32 segmentations
 - [x] Repair isosurface computation for non-editable mapping annotations with fallback layer
 - [x] Split up proofreading saga functions into smaller parts
 - [x] Decide whether the meshes from the oversegmentation are helpful or not (_disabled by default for now, enable using `window.__proofreadProximityNm = 2000;`_)
 - [x] Gracefully fail if proofreading is not allowed, because the user brushed before (maybe find out in advance when activating the proofread mode?) (_not possible to find out in advance yet, but the code fails gracefully_)

### TODO Backend
 - [x] find largest agglomerate id from hdf5 agglomerate mapping
 - [x] how to let the front-end know that a volume layer has an editable mapping?
 - [x] merge master
 - [x] clean up isosurface code (remove EditableMappingsIsosurfaceService?)
 - [x] perf
    - [x] Caching for infos fetched from datastore
      - [x] unify caching code
      - [x] add caching for largest agglomerate id
    - [x] Store as JSON vs Protobuf?
 - [x] check that no brushing has happened? (key-scan on make editable, then later refuse update bucket actions if editable mapping exists)

### Follow-Up

Moved to follow up issue https://github.com/scalableminds/webknossos/issues/6288

### Note on FossilDB Data

This branch introduces two new column families in fossilDB. After running that and then switching to a different wk branch, you need to run `rm -rf fossildb/data; git checkout fossildb/data` to re-initialize the fossildb content, otherwise you’ll get `RocksDBException: You have to open all column families`

- https://editablemappings.webknossos.xyz

### Steps to test:
- abc

### Issues:
- fixes #6139
- fixes #6228
- contributes to #6241
- contributes to #6237

------
- [x] Ready for review


fm3:
> I merged the current master

:+1: 

> I implemented a rudimentary proofreading tool

:tada: 

> agglomerate ids are included as well. Are those needed/helpful?

Yes, it would save the back-end some code and a round-trip between tracingstore and datastore to look those up. If the front-end already has them, it would be great if they could be included in the update actions. This will also make reading the update actions easier for some version log later.

> Could you document the new routes

Sure! Note that not everything is working yet in the backend :)

There is one new VolumeUpdateAction: `{"name:" "updateMappingName", value: {"mappingName": "asdf"}}` (use this to pin the mapping to an existing on-disk agglomerate file or to an editable mapping id, see below.). The VolumeTracing proto object has a new field "mappingName" containing the same value.

POST        /volume/:tracingId/createEditableMapping?token=asdf

…will initialize an empty editable mapping and return its ID. Note: This will assert that the mappingName of the volume tracing is already pinned (so pin it to an on-disk one first). The front-end should then send the updateMappingName volume update action again, this time with the id returned by the createEditableMapping call.

GET         /volume/:tracingId/agglomerateSkeleton?token=asdf

…will reply with an agglomerate skeleton (asserts that the mapping is pinned. no matter if to an on-disk or editable one).

POST        /volume/:tracingId/updateEditableMapping?token=asdf

… takes one EditableMappingUpdateAction as json body. The ones you defined above are these update actions. Note: current implementation is not expecting two positions yet. I will change that. Will fail if the mapping is not already set to an editable one. Note that it takes the tracingId (not the mapping id).


While writing this I noticed that the front-end probably has no way to know if the mapping that the VolumeTracing references is indeed editable. I see three options here:

 (1) Since this is a prototype, the front-end just has to remember that it sent such an update action. Breaks on reload, though
 (2) Additional boolean field in the VolumeTracing proto definition
 (3) Every mapping is considered editable, create the id lazily.

We also need to consider how to assert that nothing has been brushed etc. Let’s talk about that issue again in the next days.

I hope this already helps you :)

fm3:
@daniel-wer As discussed, I adapted the update action to always take positions (and mag) instead of the segment IDs. I also implemented some of the missing methods. From my current understanding, only the fetching of the `largestAgglomerateId` is missing (needed to assign a new id during a split action). And the above question of how to give the front-end the information that the current annotation is an editable-mapping annotation. Everything else should be complete, but I did not test any of it yet.

fm3:
As discussed, 
 - the route, now called `/volume/:tracingId/makeMappingEditable` directly inserts the update action setting the editable mapping in the volume tracing.
 - the volume tracing proto now includes an `mappingIsEditable` boolean (it is optional for backwards compatibility. missing should be interpreted as false)
 - I added the new route `/volume/:tracingId/editableMappingUpdateActionLog`
 - The agglomerate skeleton route now includes the agglomerate id in the URI (`/volume/:tracingId/agglomerateSkeleton/:agglomerateId`)

daniel-wer:
Thanks @fm3!

I implemented that the mapping name is persisted with the volume annotation and that a mapping is ensured to be editable before executing the first proofreading action. Additionally, I started to work on saving the proofreading update actions separately from skeleton/volume ones and sending them to another end point.

A couple of other minor things/problems I noticed:
- Since only the mapping name is persisted in the volume annotation, one cannot distinguish between json or hdf5 mappings. Since the proofreading is supposed to work with hdf5 mappings, only those are saved in the mapping name if they are activated. Consequently, when retrieving a mapping from an annotation's `mappingName` property, it is assumed to be an hdf5 mapping. I would say that's fine.
- [ ] It's not possible to set `mappingName` to null. The code tries to do that once an hdf5 mapping is disabled or if another non-hdf5 mapping is enabled.
- [ ] After making a mapping editable, the `mappingName` in the annotation changes to an ID. However, the ID doesn't appear in the list of available mappings which is why the frontend complains that the mapping is invalid. Would it be complicated to add that?
- [ ] Where is the version of the editable mapping supposed to be saved (and retrieved from)?

  - Relatedly, I'm wondering whether it would be complicated for the backend to handle the editable mapping save end point more similarly to the `skeleton` and `volume` end points. The way it is designed, currently, makes it rather hard to reuse existing functionality to save update actions and to retrieve a version log (also see the issue with the version of the editable mapping). Maybe the `mappingName` attached to a volume annotation should be an object instead (contrary to what we agreed upon earlier :see_no_evil:). This object could have a name (which serves as an ID), a version, and the `isEditable` flag :thinking: In line with the other routes the save route could then be `mapping/<mappingId>/update` and the update action log route could be `mapping/<mappingId>/updateActionLog`. These are just ideas, I have no idea whether that would be hard to do in the backend. Let's discuss tomorrow, maybe :)

fm3:
new route format:
```
POST        /mapping/:tracingId/update         
GET         /mapping/:tracingId/updateActionLog
GET         /mapping/:tracingId                
```
update now expects an update action group with exactly one action in this format:
```
version: Long,
timestamp: Long,
actions: List[EditableMappingUpdateAction]
```

the get route returns info json:
```
 "mappingName" string,
 "version" long,
 "tracingId" string
```

the same info json is now also returned in makeMappingEditable

fm3:
I added the creation timestamp :) I do not yet understand where the null pointer exception in the isosurface service comes from. I will investigate that further

frontend/javascripts/oxalis/model/sagas/isosurface_saga.ts
```diff
@@ -366,8 +368,11 @@ function* maybeLoadIsosurface(
   }`;
   const tracingStoreUrl = `${tracingStoreHost}/tracings/volume/${layer.name}`;
   const volumeTracing = yield* select((state) => getActiveSegmentationTracing(state));
-  // Fetch from datastore if no volumetracing exists or if the tracing has a fallback layer.
-  const useDataStore = volumeTracing == null || volumeTracing.fallbackLayer != null;
```
daniel-wer, line 370: @fm3 This condition was adapted so that even for volume tracings with a fallback layer, the isosurface is requested from the tracing store (to make isosurfaces for editable mappings work). Does this break functionality for tracings without an editable mapping or is the isosurface request proxied to the data store in that case?
If this breaks stuff, I need to adapt the condition and check whether an editable mapping is active as well.

fm3: I’d say if your request an isosurface from a normal volume annotation with fallback layer *from the tracingstore* it will not do proxying but instead run marching cubes only on the modified buckets. if there are no modified buckets, there will be no isosurfaces. So yes, I’d say this breaks this.
In the long run, I guess we will want this proxying to happen in the backend also in this case, but maybe not as part of this PR

fm3:
@daniel-wer I managed to fix the null pointer exception, and I also now fixed the volume-buckets-are-empty assertion in makeMappingEditable. I’d consider the back-end ready for review now if you don’t think something else is missing.
Note that the front-end does not yet gracefully handle the failing makeMappingEditable call, maybe you can add something there.

MIGRATIONS.unreleased.md
```diff
@@ -8,4 +8,6 @@ User-facing changes are documented in the [changelog](CHANGELOG.released.md).
 ## Unreleased
 [Commits](https://github.com/scalableminds/webknossos/compare/22.06.0...HEAD)
 
+ - Note that this upgrade can not be trivially rolled back, since new rocksDB column families are added and it is not easy to remove them again from an existing database. [#6195](https://github.com/scalableminds/webknossos/pull/6195)
```
fm3, line 11: add note on fossildb arguments, new column families

webknossos-datastore/app/com/scalableminds/webknossos/datastore/dataformats/wkw/WKWBucketProvider.scala
```diff
@@ -14,9 +14,9 @@ class WKWCubeHandle(wkwFile: WKWFile) extends DataCubeHandle with FoxImplicits {
 
   def cutOutBucket(bucket: BucketPosition)(implicit ec: ExecutionContext): Fox[Array[Byte]] = {
     val numBlocksPerCubeDimension = wkwFile.header.numBlocksPerCubeDimension
-    val blockOffsetX = bucket.x % numBlocksPerCubeDimension
-    val blockOffsetY = bucket.y % numBlocksPerCubeDimension
-    val blockOffsetZ = bucket.z % numBlocksPerCubeDimension
+    val blockOffsetX = bucket.bucketX % numBlocksPerCubeDimension
```
fm3, line 17:
```suggestion
    val blockOffsetX = bucket.bucketIndexX % numBlocksPerCubeDimension
```
?

webknossos-datastore/app/com/scalableminds/webknossos/datastore/services/IsosurfaceService.scala
```diff
@@ -54,9 +56,13 @@ class IsosurfaceService(binaryDataService: BinaryDataService,
...
                         isosurfaceActorPoolSize: Int)(implicit ec: ExecutionContext)
-    extends FoxImplicits {
+    extends FoxImplicits
+    with LazyLogging {
 
-  private val agglomerateService: AgglomerateService = binaryDataService.agglomerateService
+  private val agglomerateService
+    : Option[AgglomerateService] = try { Some(binaryDataService.agglomerateService) } catch {
+    case _: NullPointerException => None
+  }
```
fm3, line 65: see follow-up issue https://github.com/scalableminds/webknossos/issues/6234

webknossos-datastore/app/com/scalableminds/webknossos/datastore/services/IsosurfaceService.scala
```diff
@@ -184,19 +190,23 @@ class IsosurfaceService(binaryDataService: BinaryDataService,
...
     val vertexBuffer = mutable.ArrayBuffer[Vec3Double]()
 
     for {
+      before <- Fox.successful(System.currentTimeMillis())
       data <- binaryDataService.handleDataRequest(dataRequest)
+      afterLoading = System.currentTimeMillis()
       agglomerateMappedData = applyAgglomerate(data)
       typedData = convertData(agglomerateMappedData)
       mappedData <- applyMapping(typedData)
+      agglomerateIds: Set[T] = mappedData.toSet
```
fm3, line 206: remove with logging before merge

webknossos-tracingstore/app/com/scalableminds/webknossos/tracingstore/controllers/VolumeTracingController.scala
```diff
@@ -440,4 +434,94 @@ class VolumeTracingController @Inject()(
...
   }
 
+  def agglomerateSkeleton(token: Option[String], tracingId: String, agglomerateId: Long): Action[AnyContent] =
+    Action.async { implicit request =>
+      accessTokenService.validateAccess(UserAccessRequest.readTracing(tracingId), token) {
+        for {
+          tracing <- tracingService.find(tracingId)
+          mappingName <- tracing.mappingName ?~> "annotation.agglomerateSkeleton.noMappingSet"
+          remoteFallbackLayer <- editableMappingService.remoteFallbackLayer(tracing)
+          _ <- Fox.assertTrue(editableMappingService.exists(mappingName)) ?~> "Cannot query agglomerate skeleton for volume annotation"
```
fm3, line 444: use tracing.ismappingeditable

webknossos-tracingstore/app/com/scalableminds/webknossos/tracingstore/controllers/VolumeTracingController.scala
```diff
@@ -440,4 +434,94 @@ class VolumeTracingController @Inject()(
...
+      }
+    }
+
+  def updateEditableMapping(token: Option[String], tracingId: String): Action[List[EditableMappingUpdateActionGroup]] =
+    Action.async(validateJson[List[EditableMappingUpdateActionGroup]]) { implicit request =>
+      accessTokenService.validateAccess(UserAccessRequest.writeTracing(tracingId), token) {
+        for {
+          tracing <- tracingService.find(tracingId)
+          mappingName <- tracing.mappingName.toFox
+          _ <- Fox.assertTrue(editableMappingService.exists(mappingName))
```
fm3, line 489: add isEditable assertion

webknossos-datastore/app/com/scalableminds/webknossos/datastore/services/AgglomerateService.scala
```diff
@@ -189,4 +191,78 @@ class AgglomerateService @Inject()(config: DataStoreConfig) extends DataConverte
...
+        reader.uint64().readArrayBlockWithOffset("/agglomerate_to_edges_offsets", 2, agglomerateId)
+
+      logger.info(s"positionsRange: ${positionsRange(0)} to ${positionsRange(1)}, agglomerateId: $agglomerateId")
+
+      val nodeCount = positionsRange(1) - positionsRange(0)
+      val edgeCount = edgesRange(1) - edgesRange(0)
+      val edgeLimit = config.Datastore.AgglomerateSkeleton.maxEdges
+      if (nodeCount > edgeLimit) {
+        throw new Exception(s"Agglomerate has too many nodes ($nodeCount > $edgeLimit)")
+      }
```
jstriebel, line 237: We cannot assume that there is at least an edge per node (e.g. thinking about the current use-case with the sparse agglomerates and adding all other segments as their own agglomerates). So I'd find it confusing to change maxEdges because of the number of nodes. Maybe just rename the config entry to `nodeAndEdgeLimit` or `maxNodesAndEdges`?

webknossos-tracingstore/app/com/scalableminds/webknossos/tracingstore/tracings/volume/VolumeTracingService.scala
```diff
@@ -81,12 +83,16 @@ class VolumeTracingService @Inject()(
                         updateGroup: UpdateActionGroup[VolumeTracing],
                         previousVersion: Long): Fox[Unit] =
     for {
-      updatedTracing: VolumeTracing <- updateGroup.actions.foldLeft(find(tracingId)) { (tracingFox, action) =>
+      tracing <- find(tracingId)
```
fm3, line 86: skip, use from below

webknossos-datastore/app/com/scalableminds/webknossos/datastore/dataformats/wkw/WKWBucketProvider.scala
fm3: deferred to refactoring issue #6237

webknossos-datastore/app/com/scalableminds/webknossos/datastore/services/AgglomerateService.scala
```diff
@@ -189,4 +191,78 @@ class AgglomerateService @Inject()(config: DataStoreConfig) extends DataConverte
...
+      val hdfFile = agglomerateFileKey.path(dataBaseDir, agglomerateDir, agglomerateFileExtension).toFile
+
+      val reader = HDF5FactoryProvider.get.openForReading(hdfFile)
+
+      val positionsRange: Array[Long] =
+        reader.uint64().readArrayBlockWithOffset("/agglomerate_to_segments_offsets", 2, agglomerateId)
+      val edgesRange: Array[Long] =
+        reader.uint64().readArrayBlockWithOffset("/agglomerate_to_edges_offsets", 2, agglomerateId)
+
+      logger.info(s"positionsRange: ${positionsRange(0)} to ${positionsRange(1)}, agglomerateId: $agglomerateId")
```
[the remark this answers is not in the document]
fm3: Nope, I’ll remove this line

webknossos-datastore/app/com/scalableminds/webknossos/datastore/services/AgglomerateService.scala
fm3: Makes sense but I’d vote to defer this, in order not to make the migrations needed for deploying this PR more complicated, especially since this problem was not introduced here but already existed.

jstriebel: Fair :+1:

fm3:
I implemented the PR feedback from @jstriebel that came up when we spoke in person :)

jstriebel (commented):
Awesome stuff :rocket: Backend LGTM :+1:

frontend/javascripts/oxalis/model/sagas/proofread_saga.ts
```diff
@@ -0,0 +1,451 @@
...
+  const targetTree = findTreeByNodeId(trees, targetNodeId).getOrElse(null);
+
+  if (sourceTree == null || targetTree == null) {
+    return;
+  }
+
+  const sourceNodePosition = sourceTree.nodes.get(sourceNodeId).position;
+  const targetNodePosition = targetTree.nodes.get(targetNodeId).position;
+  const sourceNodeAgglomerateId = yield* call(
+    [api.data, api.data.getDataValue],
```
philippotto, line 310: this returns the mapped ids, right?

frontend/javascripts/oxalis/model/sagas/proofread_saga.ts
```diff
@@ -0,0 +1,451 @@
...
+  const layerName = volumeTracingId;
+  const serverEditableMapping = yield* call(makeMappingEditable, tracingStoreUrl, volumeTracingId);
+  // The server increments the volume tracing's version by 1 when switching the mapping to an editable one
+  yield* put(setVersionNumberAction(upToDateVolumeTracing.version + 1, "volume", volumeTracingId));
+  yield* put(setMappingNameAction(layerName, serverEditableMapping.mappingName));
+  yield* put(setMappingisEditableAction());
+  yield* put(initializeEditableMappingAction(serverEditableMapping));
+}
+
+function* splitOrMergeAgglomerate(action: MergeTreesAction | DeleteEdgeAction) {
```
philippotto, line 257: this saga is quite long. would it be possible to split it up? or is this too tedious since a lot of variables would need to be passed?

frontend/javascripts/oxalis/model/sagas/save_saga.ts
```diff
@@ -713,20 +714,22 @@ function* applyAndGetRevertingVolumeBatch(
...
-export function* pushTracingTypeAsync(
-  tracingType: "skeleton" | "volume",
+export function* pushSaveQueueAsync(
+  saveQueueType: SaveQueueType,
   tracingId: string,
+  isWkReady: boolean = false,
 ): Saga<void> {
-  yield* take("WK_READY");
-  yield* put(setLastSaveTimestampAction(tracingType, tracingId));
+  if (!isWkReady) yield* take("WK_READY");
```
philippotto, line 722: it seems a bit brittle to require from the caller to pass the correct `isWkReady` info. especially, since the parameter is optional. couldn't the saga check itself whether wk_ready was already dispatched?

frontend/javascripts/oxalis/model/sagas/save_saga.ts
```diff
@@ -988,33 +990,54 @@ export function performDiffTracing(
 
   return actions;
 }
+
+let isWkReady = false;
+
+function setWkReady() {
+  isWkReady = true;
+}
```
philippotto, line 998: adding to my previous comment: how about moving the wk-ready related code into a separate module and hooking it directly up in the root saga (no dependence on the save_saga). then, there could be a separate saga like this:

```
export function* ensureWkReady() {
  if (isWkReady) return;
  yield* take("WK_READY");
}
```

then, `ensureWkReady` would be an easy way to ensure that wk is ready. might make sense to use it for other sagas, too, but that's obviously not necessary for this PR.

frontend/javascripts/oxalis/model/sagas/save_saga.ts
philippotto: also see my next comment in line 994.

frontend/javascripts/oxalis/model/sagas/save_saga.ts
```diff
@@ -988,33 +990,54 @@ export function performDiffTracing(
...
 export function* saveTracingAsync(): Saga<void> {
+  yield* takeEvery("WK_READY", setWkReady);
   yield* takeEvery("INITIALIZE_SKELETONTRACING", saveTracingTypeAsync);
   yield* takeEvery("INITIALIZE_VOLUMETRACING", saveTracingTypeAsync);
+  yield* takeEvery("INITIALIZE_EDITABLE_MAPPING", saveEditableMappingAsync);
+}
+
+export function* saveEditableMappingAsync(
```
philippotto, line 1006: Not new due to this PR, but what do you think about renaming `saveEditableMappingAsync` and `saveTracingTypeAsync` to something like `setupSavingFor...`? I think, it would fit better to what is happening there. The current `save` sounds like it would happen once when called.

philippotto (changes requested):
awesome stuff! I'm already leaving some feedback, but have to review some additional files, though :)

MIGRATIONS.unreleased.md
```diff
@@ -8,4 +8,6 @@ User-facing changes are documented in the [changelog](CHANGELOG.released.md).
 ## Unreleased
 [Commits](https://github.com/scalableminds/webknossos/compare/22.06.0...HEAD)
 
+ - FossilDB now has to be started with two new additional column families: editableMappings,editableMappingUpdates. Note that this upgrade can not be trivially rolled back, since new rocksDB column families are added and it is not easy to remove them again from an existing database. In case webKnossos needs to be rolled back, it is recommended to still keep the new column families in FossilDB. [#6195](https://github.com/scalableminds/webknossos/pull/6195)
```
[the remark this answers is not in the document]
fm3: That depends on how the fossildb is launched in a concrete environment. When using docker-compose with the prepared files, that should already happen automatically with the new files. However, in a kubernetes environment, the columnFamilies key will have to be added to the release values

frontend/javascripts/oxalis/model/sagas/update_actions.ts
```diff
@@ -183,30 +209,40 @@ export type UpdateAction =
...
+};
 type AddServerValuesFn<T extends { value: any }> = (arg0: T) => T & {
   value: T["value"] & {
     actionTimestamp: number;
   };
 };
 
 type AsServerAction<A extends { value: any }> = ReturnType<AddServerValuesFn<A>>;
-// Since flow does not provide ways to perform type transformations on the
+// Since typescript does not provide ways to perform type transformations on the
```
philippotto, line 245: I was curious and checked whether TS might be smarter than flow and in fact it is :tada: The entire ServerUpdateAction can be slimmed down to:

```ts
export type ServerUpdateAction = AsServerAction<
  UpdateAction | ImportVolumeTracingUpdateAction | CreateTracingUpdateAction
>;
```

frontend/javascripts/oxalis/model/sagas/update_actions.ts
```diff
@@ -220,18 +256,23 @@ export type ServerUpdateAction =
...
   | AsServerAction<UpdateTreeGroupsUpdateAction>
-  | AsServerAction<CreateTracingUpdateAction>
-  | AsServerAction<RemoveFallbackLayerAction>
-  | AsServerAction<UpdateTdCameraAction>;
+  | AsServerAction<RemoveFallbackLayerUpdateAction>
+  | AsServerAction<UpdateTdCameraUpdateAction>
+  | AsServerAction<UpdateMappingNameUpdateAction>
+  | AsServerAction<SplitAgglomerateUpdateAction>
+  | AsServerAction<MergeAgglomerateUpdateAction>
+  // These two actions are never sent by the frontend and, therefore, don't exist in the UpdateAction type
```
philippotto, line 273: what would happen if they are added to `UpdateAction`? Would this force us to handle the actions during diffing etc? maybe that would be a bit more consistent, since the current way seems a bit confusing (e.g., only looking at the `UpdateAction` definition might raise the question why not all actions are listed there). I don't have a strong opinion about this, though.

frontend/javascripts/oxalis/view/action-bar/toolbar_view.tsx
```diff
@@ -503,8 +504,9 @@ function ChangeBrushSizeButton() {
 export default function ToolbarView() {
   const hasVolume = useSelector((state: OxalisState) => state.tracing.volumes.length > 0);
   const hasSkeleton = useSelector((state: OxalisState) => state.tracing.skeleton != null);
-  const viewMode = useSelector((state: OxalisState) => state.temporaryConfiguration.viewMode);
```
philippotto, line 506: so, the view mode is not taken into account anymore? doesn't this have implications on the arbitrary mode?

frontend/javascripts/oxalis/view/context_menu.tsx
```diff
@@ -812,7 +819,8 @@ function NoNodeContextMenuOptions(props: NoNodeContextMenuProps) {
       mode="vertical"
     >
       {allActions}
-      {infoRows}
+      {/* In the TD viewport the info rows are not usable since the click position cannot be computed */}
+      {isTdViewport ? empty : infoRows}
```
philippotto, line 823: I think, it would be better if the caller of this function would pass an empty `infoRows` if the TD viewport is active. This would save some computations and allocations of JSX nodes and also the caller should have better knowledge over whether the info rows are irrelevant for the 3D case (e.g., in the future, there might be info rows which make sense for the 3D viewport).

frontend/javascripts/types/api_flow_types.ts
```diff
@@ -610,8 +610,17 @@ export type ServerVolumeTracing = ServerTracingBase & {
...
+  mappingIsEditable?: boolean;
 };
 export type ServerTracing = ServerSkeletonTracing | ServerVolumeTracing;
+export type ServerEditableMapping = {
+  createdTimestamp: number;
+  version: number;
+  mappingName: string;
+  // The id of the volume tracing the editable mapping belongs to
+  tracingId: string;
+};
```
philippotto, line 623: what do you think about defining `EditableMapping` in store.ts using this definition? Sth like

```
type EditableMapping = ReadOnly<ServerEditableMapping> & {
    readonly type: "mapping";
}
```

philippotto (commented):
I'm through with the front-end review. Excellent work! :100: I'll test now :)

philippotto:
Usability-wise:
- [ ] It feels a bit unusual to split and merge trees with the proofreading tool. I could imagine that users will quickly switch back to the skeleton tool for that (which wouldn't do any editing on the mapping then). Not sure how to solve this, though.. Maybe one could detect that the user edited an agglomerate skeleton in the wrong mode? I've made the mistake myself during testing multiple times..
- [ ] I haven't found the meshing of super-voxels super-helpful. I rather used the agglomerate skeleton itself. However, this is only my personal impression. Disabling this via window

(Potential) Bugs:
- [ ] Maybe this has already been like this, but I noticed these artifacts in the mesh when looking from a certain direction:
image
This is a normal ad-hoc mesh (I think I used mapping 50). Segment position: 1687, 4456, 1740
image
(ok, I'm pretty sure this hasn't been like this before. could this have something to do with the rendering order and the passive meshes?).

- [ ] Are the meshes more transparent? While reading the code, I thought, this would only be the case for the automatic meshing during proofreading, but I think normal meshing also got more transparent? Could this be?
- [ ] Even though, the brush tool (and co) are not shown when an editable mapping is present, I can press "W" to switch to these tools.
- [ ] At one point, I merged two nodes and then the active position jumped to some other location. Not sure why this was happening.
- [ ]

philippotto:
Ok, I just finished my first testing round and overall most worked pretty well :+1: The usability is not ultra smooth, but we can iterate in later PRs and with other user feedback :)

See my list below of things I noticed:

(Potential) Bugs:
- [ ] Maybe this has already been like this, but I noticed these artifacts in the mesh when looking from a certain direction:
image
This is a normal ad-hoc mesh (I think I used mapping 50). Segment position: 1687, 4456, 1740
image
(ok, I'm pretty sure this hasn't been like this before. could this have something to do with the rendering order and the passive meshes? or is this only because of the agglomeration?).

- [ ] Are the meshes more transparent? While reading the code, I thought, this would only be the case for the automatic meshing during proofreading, but I think normal meshing also got more transparent? Could this be?
- [ ] Even though, the brush tool (and co) are not shown when an editable mapping is present, I can press "W" to switch to these tools.
- [ ] At one point, I merged two nodes and then the active position jumped to some other location. Not sure why this was happening. In fact, it happened twice.
- [ ] Somehow, I managed to create various redundant agglomerate skeletons (see screenshot). I didn't notice this at first, but wondered why I couldn't delete an edge (the edge existed multiple times which is why I thought the deletion didn't happen).
image
- [ ] When clicking on a agglomerate node in the 3D viewport, sometimes this toast comes up "Clicked on the background. Please click on a segment to load a skeleton."
- [ ] After I split a tree, I hit "undo" (this is probably not supported, right now?). Then, both tree components were gone. Hitting "undo" again, restored the initial tree. Redoing twice took me back to the two tree components. Not sure why there is a superfluous (?) entry in the undo stack.
 
Usability comments:
- [ ] It feels a bit unusual to split and merge trees with the proofreading tool. I could imagine that users will quickly switch back to the skeleton tool for that (which wouldn't do any editing on the mapping then). Not sure how to solve this, though.. Maybe one could detect that the user edited an agglomerate skeleton in the wrong mode? I've made the mistake myself during testing multiple times..
- [ ] I haven't found the meshing of super-voxels super-helpful. I rather used the agglomerate skeleton itself. However, this is only my personal impression. Disabling this via window is fine.
- [ ] The UI doesn't explain anything about the proofreading right now (and also not why this disables some UI). For the first iteration, it's probably alright. Just wanted to mention it. 
- [ ] Should the merger-mode be disabled when an editable mapping exists? I don't think, the combination would work well.


Let me know if you want to discuss things in person!

philippotto:
Ah, and some other thoughts I forgot above:

- I think, I'd move the proofreading tool in the toolbar to the right (either before or after the bbox tool). Right now, it feels to prominent to me.
- When using the version restore view and clicking on an older entry in the mappings tab, it says "Version preview for mappings is not supported yet.". I'd change that to `Version restoring for mappings...`. Since that's the actual culprit. At least, I couldn't figure out how to restore an older version..

frontend/javascripts/oxalis/controller/combinations/tool_controls.ts
```diff
@@ -617,10 +623,62 @@ export class BoundingBoxTool {
...
+    if (plane === OrthoViews.TDView) {
+      if (didSelectNode) {
+        getSkeletonTracing(Store.getState().tracing).map((skeletonTracing: SkeletonTracing) =>
+          getNodeAndTree(skeletonTracing).map(([_activeTree, activeNode]) => {
+            globalPosition = activeNode.position;
+          }),
+        );
+      }
+    } else {
+      globalPosition = calculateGlobalPos(Store.getState(), pos);
```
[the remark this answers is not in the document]
daniel-wer: The `pos` argument to `calculateGlobalPos` is the _click position_ in 2d viewport coordinates. `calculateGlobalPos` converts this to the global position in 3d. This works great for the orthogonal viewports (and is used for other tools and the context menu). However, this doesn't work for the 3D viewport, because the click position in 3D cannot be determined uniquely. If the user clicked _on a node_ in the 3d viewport, though, we can determine the click position by looking up the node position. That's what's happening here.
I'll add a comment to make this more clear!

frontend/javascripts/oxalis/model/accessors/volumetracing_accessor.ts
```diff
@@ -416,6 +416,34 @@ export function getMappingInfoForVolumeTracing(
...
+
+  if (isEditableMappingActive) {
+    const volumeTracing = getActiveSegmentationTracing(state);
+
+    // This should never be the case, since editable mappings can only be active for volume tracings
+    if (volumeTracing == null) return false;
+
+    // Only allow mapping activations of the editable mapping itself if an editable mapping is saved
+    // in the volume tracing. Editable mappings cannot be disabled or switched for now.
+    return mappingName === volumeTracing.mappingName;
```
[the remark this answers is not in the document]
daniel-wer: I agree with the early-out :+1:
However, I'm not sure about the `mappingName == null`. What exactly would the condition look like?
In my view, the current condition is testing the right thing. If `mappingName` is null, but `volumeTracing.mappingName` is not null, the mapping (de)activation is not allowed (correct). If `mappingName` is null, and `volumeTracing.mappingName` also is null (far fetched, because `mappingIsEditable` should not be true then), the mapping (de)activation is allowed (also correct imo).

frontend/javascripts/oxalis/model/actions/volumetracing_actions.ts
```diff
@@ -261,3 +279,6 @@ export const dispatchFloodfillAsync = async (
   dispatch(action);
   await readyDeferred.promise();
 };
+export const setMappingisEditableAction = (): SetMappingIsEditableAction => ({
```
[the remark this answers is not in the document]
daniel-wer: :eagle: :eyes: 

I would say it's rather likely that in the near future it will be possible to switch from an editable mapping to another (non-editable) one (and back again). The boolean parameter can be introduced then. So I'd vote to keep this as-is.

frontend/javascripts/oxalis/model/reducers/save_reducer.ts
```diff
@@ -74,6 +102,14 @@ function SaveReducer(state: OxalisState, action: Action): OxalisState {
       });
     }
 
+    case "INITIALIZE_EDITABLE_MAPPING": {
+      // Set up empty save queue array for editable mapping
+      const newMappingsQueue = { ...state.save.queue.volumes, [action.mapping.tracingId]: [] };
```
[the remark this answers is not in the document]
daniel-wer: Uff, great catch :pray:

frontend/javascripts/oxalis/model/reducers/skeletontracing_reducer_helpers.ts
```diff
@@ -534,10 +539,12 @@ export function addTreesAndGroups(
   treeGroups: Array<MutableTreeGroup>,
 ): Maybe<[MutableTreeMap, Array<TreeGroup>, number]> {
   const treeIds = Object.keys(trees).map((treeId) => Number(treeId));
-  // TreeIds > 1024^2 break webKnossos, see https://github.com/scalableminds/webknossos/issues/5009
-  const hasTreeIdsLargerThanMaximum = treeIds.some((treeId) => treeId > 1048576);
+  const hasInvalidTreeIds = treeIds.some(
+    (treeId) => treeId < Constants.MIN_TREE_ID || treeId > Constants.MAX_TREE_ID,
+  );
+  const hasInvalidNodeIds = getMinimumNodeId(trees) < Constants.MIN_NODE_ID;
```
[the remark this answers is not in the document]
daniel-wer: During development, the backend sent nodes with id 0 mistakenly (which our code cannot handle). This led to weird issues which were not that easy to debug. Also, an uploaded NML could contain node IDs smaller than 1. Same for the tree IDs.

frontend/javascripts/oxalis/model/reducers/skeletontracing_reducer_helpers.ts
```diff
@@ -59,8 +59,13 @@ export function generateTreeName(state: OxalisState, timestamp: number, treeId:
 
   return `${prefix}${Utils.zeroPad(treeId, 3)}`;
 }
+function getMinimumNodeId(trees: TreeMap | MutableTreeMap): number {
+  const minNodeId = _.min(_.flatMap(trees, (tree) => tree.nodes.map((n) => n.id)));
+
+  return minNodeId != null ? minNodeId : Constants.MIN_NODE_ID;
```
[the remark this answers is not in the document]
daniel-wer: You're right, the two functions have different behavior in the fallback case. The `getMaximum{Node,Tree}Id` functions return the largest ID encountered. If there are no objects they opt to return the minimal ID - 1 so that the next highest ID would be the minimal ID.

To make `getMinimumNodeId` work uniformly, the fallback case should return Constants.MAX_NODE_ID + 1. However, there is no maximum node id as far as I know :sweat_smile: Since `getMinimumNodeId` is only used for validity checking, returning any number within the valid range is fine, I would argue (as is done currently).

frontend/javascripts/oxalis/model/sagas/proofread_saga.ts
```diff
@@ -0,0 +1,451 @@
...
+  const { mappingName, mappingType } = mappingInfo;
+
+  // Load the whole agglomerate mesh in a coarse resolution for performance reasons
+  const oldPreferredQuality = yield* select(
+    (state) => state.temporaryConfiguration.preferredQualityForMeshAdHocComputation,
+  );
+
+  const resolutionIndices = resolutionInfo.getAllIndices();
+  const coarseResolutionIndex =
+    resolutionIndices[Math.min(proofreadCoarseResolutionIndex(), resolutionIndices.length - 1)];
```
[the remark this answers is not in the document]
daniel-wer: Nice, that's much better! I've also noticed that this "safety" procedure was missing for the fine resolution index. I added it there as well :+1:

frontend/javascripts/oxalis/model/sagas/proofread_saga.ts
daniel-wer: It is ensured that an HDF5 mapping is active, so yes, the mapped data values are returned since the frontend doesn't even know the unmapped ones.

frontend/javascripts/oxalis/model/sagas/update_actions.ts
daniel-wer: A quick test showed that nothing would happen, the code still type-checks and we wouldn't need to handle the actions during diffing, since the diffing is where update actions are instantiated in the first place.
I see this more as a distinction to _help_ readers of the code. I wanted to avoid that they wonder why there was no function to create such an update action (at least for the `ImportVolumeTracingUpdateAction` - the `CreateTracingUpdateAction` is created on demand for the version view).
I would argue that the current definitions more closely resemble the reality, but if it's confusing I'm also open to adapt it the way you described.

frontend/javascripts/oxalis/model/sagas/update_actions.ts
daniel-wer: That's awesome! Thanks for testing!

frontend/javascripts/oxalis/view/action-bar/toolbar_view.tsx
daniel-wer: Ah I wanted to add a comment in the PR for this, but forgot. This is a clean up, because this check is not needed any more. The ToolbarView instantiation looks like this:
`{!isReadOnly && constants.MODES_PLANE.indexOf(viewMode) > -1 ? <ToolbarView /> : null}`

At some point, we decided to not render the toolbar in arbitrary modes. This check wasn't removed at the time.

frontend/javascripts/oxalis/model/sagas/save_saga.ts
daniel-wer: That's a great idea :+1:

daniel-wer:
@fm3 During testing of the newest version I noticed that in a hybrid annotation (no editable mapping), I encounter the following error when requesting data from the tracing store, BUT only after I brushed for the first time and if an agglomerate view mapping is active :thinking: Can you reproduce this behavior?

```
Execution exception
[NullPointerException: null]

In /home/daniel/projects/oxalis/webknossos-datastore/app/com/scalableminds/webknossos/datastore/services/BinaryDataService.scala:59
55          data <- handleDataRequest(request)
56          mappedData = convertIfNecessary(
57            request.settings.appliedAgglomerate.isDefined && request.dataLayer.category == Category.segmentation && request.cuboid.mag.maxDim <= MaxMagForAgglomerateMapping,
58            data,
59            agglomerateService.applyAgglomerate(request)
60          )
61          convertedData = convertIfNecessary(
62            request.dataLayer.elementClass == ElementClass.uint64 && request.dataLayer.category == Category.segmentation,
63            mappedData,
```

frontend/javascripts/oxalis/model/sagas/isosurface_saga.ts
daniel-wer: @fm3 Is this exposed somehow (whether there are brush strokes in a volume annotation)? In any case, I'd rather not call a route to get this information before each isosurface request. One option could be to request this info once during annotation loading and then keep track of volume modifications in the frontend :shrug: Alternatively, the backend could take care of the proxying, as discussed.
For now, the simplest solution is probably to restore this condition as it was before. Then, adjust the condition to request from the tracing store if and editable mapping is active.

daniel-wer: > What downsides do you see for your last suggestion?

Good question - For the purposes of this PR, I don't see any actually :)

daniel-wer: One upside of an exposed route would be that the proofreading tool could be disabled from the beginning if an annotation with volume modifications is opened. However, not that important for the first iteration.

daniel-wer:
@philippotto Thank you so much for the helpful feedback and extensive testing!

I should have addressed most of your feedback by now, except for some of the usability issues you noticed during testing. It would be great if you could have another look at the code.

I've disabled the rendering of the fine meshes from the oversegmentation by default for now. They can be enabled by setting `window.__proofreadProximityNm` to a value larger than 0.

> Maybe this has already been like this, but I noticed these artifacts in the mesh when looking from a certain direction:

This seems to be related to the transparent meshes. Researching online, rendering double sided transparent objects seems to be rather hard. I'm still fiddling around to find a good solution, maybe rendering the passive (very transparent) meshes only one sided would be an option. I've reset the transparency for the "normal" meshes, so this shouldn't be an issue for those.

>  Somehow, I managed to create various redundant agglomerate skeletons (see screenshot). I didn't notice this at first, but wondered why I couldn't delete an edge (the edge existed multiple times which is why I thought the deletion didn't happen).

I would be interested whether you can still reproduce this in the newly deployed version. Not sure what happened there or whether an older version was deployed - I made sure duplicate agglomerate skeletons are not loaded quite a while ago :thinking:

> When clicking on a agglomerate node in the 3D viewport, sometimes this toast comes up "Clicked on the background. Please click on a segment to load a skeleton."

I wasn't able to reproduce this :/ What I've noticed is that if agglomerate skeleton node positions are exactly at the lower (or left) edge of a segment, the segment ID that is read seems to be that of the lower (or left) segment instead of the actual segment the node is in. I'll have to look into that, but it could be related.

> After I split a tree, I hit "undo" (this is probably not supported, right now?). Then, both tree components were gone. Hitting "undo" again, restored the initial tree. Redoing twice took me back to the two tree components. Not sure why there is a superfluous (?) entry in the undo stack.

Yes, Undo/Redo is not properly supported for now. The disappearing trees happened, because I used to delete and reload the agglomerate skeletons after a proofreading action to make sure the names were updated correctly. I've switched to simply renaming the skeletons in the frontend. Still, the renaming causes separate undo stack entries which is not ideal. Not sure what to do about this, to be honest. A fairly involved solution could be to actually reverse the merge/split actions so that the editable mapping actions is undone as well.

> It feels a bit unusual to split and merge trees with the proofreading tool. I could imagine that users will quickly switch back to the skeleton tool for that (which wouldn't do any editing on the mapping then). Not sure how to solve this, though.. Maybe one could detect that the user edited an agglomerate skeleton in the wrong mode? I've made the mistake myself during testing multiple times..

We could try that (only if an editable mapping is active maybe to avoid false positive warnings) and wait for feedback whether this is annoying or helpful :)

daniel-wer:
> > Somehow, I managed to create various redundant agglomerate skeletons (see screenshot). I didn't notice this at first, but wondered why I couldn't delete an edge (the edge existed multiple times which is why I thought the deletion didn't happen).

> I would be interested whether you can still reproduce this in the newly deployed version. Not sure what happened there or whether an older version was deployed - I made sure duplicate agglomerate skeletons are not loaded quite a while ago thinking

Nevermind, I was able to reproduce this. This happens if an editable mapping X is activated and an agglomerate Y that was not modified before is selected. The frontend checks whether a tree with name "agglomerate Y (X)" exists. If it doesn't exist, the agglomerate skeleton is loaded. However, the loaded agglomerate skeleton has the name "agglomerate Y (agglomerate_view_80)", despite mapping X being active (probably because agglomerate Y was not modified before by mapping X). Therefore, on every click a new skeleton is loaded, because the expected and actual agglomerate skeleton tree names differ.
@fm3 Would this be possible to fix in the backend? If not, I don't have a good idea how to handle this behavior :thinking:

frontend/javascripts/oxalis/model/sagas/update_actions.ts
philippotto: Ok, I understand. Then, let's keep it as is :)

frontend/javascripts/oxalis/view/action-bar/toolbar_view.tsx
```diff
@@ -504,6 +511,14 @@ function ChangeBrushSizeButton() {
 export default function ToolbarView() {
   const hasVolume = useSelector((state: OxalisState) => state.tracing.volumes.length > 0);
   const hasSkeleton = useSelector((state: OxalisState) => state.tracing.skeleton != null);
+  const hasAgglomerateMappings = useSelector((state: OxalisState) => {
+    const visibleSegmentationLayer = getVisibleSegmentationLayer(state);
+    return (
+      visibleSegmentationLayer != null &&
+      visibleSegmentationLayer.agglomerates != null &&
+      visibleSegmentationLayer.agglomerates.length > 0
```
philippotto, line 519:
```suggestion
      visibleSegmentationLayer?.agglomerates?.length > 0
```

I hope I'm not making any false claims with my suggestions, but I'm starting to dig the `?.` operator :smile:

frontend/javascripts/oxalis/model/sagas/proofread_saga.ts
```diff
@@ -268,22 +265,18 @@ function* splitOrMergeAgglomerate(action: MergeTreesAction | DeleteEdgeAction) {
...
+  const isHdf5MappingEnabled = yield* call(ensureHdf5MappingIsEnabled, layerName);
+  if (!isHdf5MappingEnabled) return;
 
   if (!volumeTracing.mappingIsEditable) {
-    yield* call(createEditableMapping);
+    try {
+      yield* call(createEditableMapping);
+    } catch (e) {
+      console.error(e);
+      // Undo the last splitting/merging action since the proofreading action did not succeed
```
philippotto, line 276: what would be an example reason for this to fail?

philippotto (commented):
code looks great :+1: just left some cosmetics :nail_care: will test again now!

frontend/javascripts/oxalis/model/sagas/proofread_saga.ts
daniel-wer: If the user brushed before, the creation of the editable mapping will fail. That's the main case.

philippotto:
> Yes, Undo/Redo is not properly supported for now. The disappearing trees happened, because I used to delete and reload the agglomerate skeletons after a proofreading action to make sure the names were updated correctly. I've switched to simply renaming the skeletons in the frontend. Still, the renaming causes separate undo stack entries which is not ideal. Not sure what to do about this, to be honest. A fairly involved solution could be to actually reverse the merge/split actions so that the editable mapping actions is undone as well.

That's probably something for the next iteration. Is there a follow-up issue already?
Also: Would it be easy to warn the user if they try to use undo/split in combination with the proofreading? I feel like this could easily produce inconsistent states otherwise :thinking:

frontend/javascripts/oxalis/model/sagas/skeletontracing_saga.ts
```diff
@@ -265,20 +265,35 @@ function* getAgglomerateSkeletonTracing(
...
     }
 
+    if (parsedTracing.trees.length !== 1) {
+      throw new Error(
+        `Agglomerate skeleton response does not contain exactly one tree, but ${parsedTracing.trees.length} instead.`,
+      );
+    }
+
+    // Make sure the tree is named as expected
+    parsedTracing.trees[0].name = getTreeNameForAgglomerateSkeleton(agglomerateId, mappingName);
```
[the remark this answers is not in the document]
daniel-wer: Yes exactly, I discussed this with @fm3 internally. This solution fixes two issues at once.
For one, the mentioned bug where agglomerate skeletons didn't have the expected name sometimes if an editable mapping was active.
Additionally, the frontend implicitly assumed that the agglomerate skeleton naming by the backend was deterministic - if it ever were to be changed it would have had to be adapted in both back- and frontend which is a source of potential bugs.

This way, no more assumption by the frontend is needed since the `getTreeNameForAgglomerateSkeleton` function is the definitive source of truth for agglomerate skeleton names.

frontend/javascripts/oxalis/view/action-bar/toolbar_view.tsx
daniel-wer: Indeed, very helpful! I'll have to remember to use it :)

daniel-wer:
@philippotto I should have addressed all of your feedback and also moved the followup tasks to a new issue https://github.com/scalableminds/webknossos/issues/6288

The newest version is deployed on the dev branch, would be great if you could have another look :eagle:

PR #6196, opened 2022-05-09
Fix resolution name in data request for backwards-compatibility

jstriebel:
There was still a left-over from the [zoomStep/resolution renaming back-and-forth](https://github.com/scalableminds/webknossos/pull/6159#discussion_r865801752), leading to failures with wk-libs. Let's release this soon.

Follow-up from #6159


fm3 (approved):

jstriebel:
PS: @fm3, do you have an idea if we could fail the build if an argument in the routes file is not accepted in the function? I guess the other way around is a normal case, but this way should lead to an error IMO.

fm3:
I would have expected the scala compiler to fail in this case, which is also why I didn’t catch the error in the first place. Unfortunately, I don’t think there is an easy way to fix this. We could write our own validation logic, but it would not be simple. Since this is the first time this has happened, I’d say it is not high-enoughp riority to build that.

Issue #6112, opened 2022-03-14
New Mailer tags for active/inactive after x weeks


fm3:
New tags for `3 weeks after they subscribed -> true if active during the last 2 weeks` were requested

Compare https://scm.slack.com/archives/C02A9H8CU/p1647252083251459

PR #6213, opened 2022-05-17
Update lastUserActivity on typical requests

fm3:
In preparation for the mailchimp tag for active or inactive users, we need to store the last user activity.

Two weeks after this is deployed, we can then deploy a follow-up that tags users based on this data.

The lastActivity column in the users table was already in place, but unused. This PR revives it by setting it to the current time when the user sends a request. I inserted this for some of the most frequent requests to catch most users without polluting too much code.

I had to refresh the snapshots, with the lastActivity field masked out of the user json, as it differs from run to run.

### Steps to test:
- Browser webKnossos some
- in the db, check that your user’s lastActivity timestamp is recent

### Issues:
- contributes to #6112

------
- [x] Ready for review


leowe:
How does the session expiry work? If the user keeps the tab open, does ever need to re-authenticate? If that were the case it should be sufficient to only update the lastActive entry with each authentication?

leowe (approved):
Looks good and also works for me 👍

fm3:
Auth expiry is set to quite long (like, months, not sure about the precise value; the token lasts even longer). That will not be fine-grained enough for what we want to do here, which is why I added it to a couple more requests.

PR #6215, opened 2022-05-17
Fix disappearing "ID mapping" dropdown

### LINKED ISSUE #6210: Mapping UI disappears when changing opacity of layer to 0
## Context
Dataset with segmentation layer + mappings.

## Expected Behavior
mapping UI should not disappear when changing opacity to 0.

## Current Behavior
mapping UI disappears when changing opacity to 0.

## Steps to Reproduce the bug

1. Open Dataset with segmentation layer + mapping
2. enable mapping
3. change opacity to 0 --> "id mapping" UI disappears
4.

## Your Environment for bug

## Further details:
The code currently probably looks for the visible layer. Depending on where this code happens, a "preferred fallback" could be passed around.
Or a alternative "dumb" solution: prohibit that the slider can reach a 0 value. Instead, the user should disable the layer (also an opacity of 1% is virtually imperceptible, anyway, so that might be alright).

hotzenklotz:
PR fixes a bug where the ÌD Mapping` dropdown of a segmentation layer was hidden when the layer opacity is set to 0. 

Possible cause of the bug:
- Previous versions of the wK UI had the "ID Mapping" dropdown/UI separate of the `Segmentation Layer` settings. I.e. the ID mapping dropdown had to "disable"/hide itself, when the `Segmentation` layer was disabled. 
- Since moving the ID mapping setting into the `Segmentation` layer setting, this behavior is no longer required.

- https://___.webknossos.xyz

### Steps to test:
- Set `Segmentation` layer opacity to `0`
- Toggle the segmentation layer to refresh React rendering --> ID mapping dropdown should still be visible
- Toggle the ID mapping toggle to refresh React rendering --> ID mapping dropdown should still be visible

### Issues:
- fixes #6210 

------
- [x] Ready for review

### DIFF
--- a/CHANGELOG.unreleased.md
+++ b/CHANGELOG.unreleased.md
@@ -27,6 +27,7 @@ For upgrade instructions, please check the [migration guide](MIGRATIONS.released
 - Fixed a bug in the task creation, where creation for some tasks with initial volume data would fail. [#6178](https://github.com/scalableminds/webknossos/pull/6178)
 - Fixed a bug in the task cration, where creation for some tasks with initial volume data would fail. [#6178](https://github.com/scalableminds/webknossos/pull/6178)
 - Fixed a rendering bug which could cause an incorrect magnification to be rendered in rare scenarios. [#6029](https://github.com/scalableminds/webknossos/pull/6029)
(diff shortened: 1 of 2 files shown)


philippotto (approved):
Nice :+1:

PR #6216, opened 2022-05-17
Batch nml task upload to avoid uploading too many nmls with one request

### LINKED ISSUE #6194: Task creation should batch requests not only in bulk mode, but also in from-files mode
## Detailed Description
In the task creation, there is a limit for how many tasks can be created in one request. This is 1000 for skeleton, only, 100 otherwise. The front-end automatically chunks larger amounts of tasks, but only in task_create_bulk_view. A user now tried to create tasks from 289 files, and the backend rejected the request. task_create_form_view should also implement this batching logic if the user appends multiple files.

Side node: If the user does not append the files individually but instead zips them all together, the batching cannot run. I’d say this is acceptable for the moment, though.

MichaelBuessemeyer:
This PR batches the task creation via nml to a maximum of 100 per request as this is the limit set by the backend. Previously, trying to upload more than 100 nmls at once resulted in an error. Now the batching prevents this error and triggers one request after another until all nmls are uploaded.

- https://batchnmltaskupload.webknossos.xyz

### Steps to test:
- This test requires a lot of nml files. Therefore you can use the nml file contained in this zip
[tracing.nml.zip](https://github.com/scalableminds/webknossos/files/8709818/tracing.nml.zip). Extract this nml into an empty folder. In that folder create a `.sh` script and past the following code in that file: `for i in {1..150}; do cp tracing.nml "tracing$i.nml"; done`. This code will duplicate the nml file 150 times. Then execute the script.
- Create a new task type for hybrid annotations (with skeleton and volume annotation). If there exists already one, this step can be skipped.
- Open the task admin view, click on Àdd Task` in the top right corner and use the `Create Task` Tab. There fill out the form with the newly created hybrid task type.
- Under `Task Specification` in the form select the `Upload NML File` option.
- Then click on the nml upload zone and select all the nml files created during the first step of this list. You might need to wait a little because of the high count of files.
- Next try to upload all those files as tasks via the `Create Task` button. This should succeed.
- If you try to upload the same amount of tasks on the master, an error should occur, preventing the successful task creation.

### Issues:
- fixes #6194

------
- [x] Ready for review

### DIFF
--- a/CHANGELOG.unreleased.md
+++ b/CHANGELOG.unreleased.md
@@ -12,7 +12,7 @@ For upgrade instructions, please check the [migration guide](MIGRATIONS.released
 
 ### Added
 - Added Volume Interpolation feature. When enabled, it suffices to only label every 2nd slice. The skipped slices will be filled automatically by interpolating between the labeled slices. This feature is disabled by default. Note that the feature is even forbidden for tasks by default, but can be enabled/recommended. [#6162](https://github.com/scalableminds/webknossos/pull/6162)
-- Added a batching mechanism to task creation via nmls to support uploading more that 100 nmls at a time. [#6216](https://github.com/scalableminds/webknossos/pull/6216)
+- Added a batching mechanism to the task creation via NML to support uploading more than 100 NMLs at a time. [#6216](https://github.com/scalableminds/webknossos/pull/6216)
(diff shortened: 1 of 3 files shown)


daniel-wer (approved):
Works like a charm :tada: 

I agree that avoiding the bit of code duplication would be rather difficult and is not worth it.

PR #6217, opened 2022-05-18
Fix swagger list-of-string type annotation

fm3:
Using the workaround suggested in https://github.com/swagger-api/swagger-play/issues/104#issuecomment-295737502

- https://swaggerlistofstring.webknossos.xyz

### Steps to test:
- Compile, and launch should not log swagger exception
- swagger.json should contain valid definition for update teams route:
```
{
  "in" : "body",
  "name" : "datasetUpdateTeamsInformation",
  "required" : true,
  "schema" : {
    "type" : "array",
    "items" : {
      "type" : "string"
    }
  }
}
```

I also removed an unused import and an unused injection that the compiler warned about.

- [x] Ready for review

### DIFF
--- a/app/controllers/DataSetController.scala
+++ b/app/controllers/DataSetController.scala
@@ -6,28 +6,20 @@ import com.scalableminds.util.geometry.Vec3Int
 import com.scalableminds.util.mvc.Filter
 import com.scalableminds.util.tools.DefaultConverters._
 import com.scalableminds.util.tools.{Fox, JsonHelper, Math}
-import io.swagger.annotations.{
-  Api,
-  ApiImplicitParam,
-  ApiImplicitParams,
-  ApiOperation,
-  ApiParam,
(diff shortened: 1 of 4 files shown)


jstriebel (approved):
Nice, thanks :heart:

PS: Can we do something similar for the other body parameters? Atm they all have `"type" : "string"` in their annotation, which we [fix in the wk-libs later](https://github.com/scalableminds/webknossos-libs/blob/master/webknossos/__generate_client.py#L310-L324). You might search for `"in" : "body"` in [the swagger.json](https://webknossos.org/swagger.json) to find those. I think they should all have `type: object` instead (except for `updateTeams` you fixed here). Feel free to ignore this for this PR though.

fm3:
Not sure how to make it output type: object :shrug: Let’s defer this until the larger issue dealing with swagger inconsistencies

PR #6219, opened 2022-05-18
Allow tiff export via worker for all datasets a user can see

### LINKED ISSUE #5254: Tiff export permission system
Currently, the user can export tiff of any dataset of *their own organization*, if they can open it.
Should it be possible to restrict or open this further? Should it be strictly opt-in? We could save per layer or per dataset if it should be exportable.

fm3:
### Steps to test:
- `yarn enable-jobs`
- create two organizations (use isDemoInstance in application.conf)
- you should be able to tiff-export on all datasets you are allowed to see, even if they belong to a different organization
- you should not be allowed to send tiff-export requests for datasets not shared with you (via team or public)

### Issues:
- fixes #5254 

------
- [x] Ready for review


jstriebel (approved):
Nice, the code looks even simpler now :+1:  LGTM, didn't test it myself, though.

PR #6220, opened 2022-05-18
Add hint on bounding boxes in read-only tracing/view mode

fm3:
Dataset view mode:

image

Annotation without bboxes:
image

Annotation with bboxes:

image

Foreign annotation without bboxes:
image

Foreign annotation with bboxes:
image


philippotto:
I fine-tuned the logic a bit. Still not the most beautiful UI, but I hope the message is clear.

daniel-wer (approved):
LGTM, Thanks to the two of you for taking care of this :1st_place_medal:

PR #6221, opened 2022-05-18
Make far plane of td viewport adaptive to dataset

### LINKED ISSUE #3849: 3D viewport for low res datasets is broken
bildschirmfoto 2019-02-28 um 18 18 02
See https://webknossos.org/datasets/Demo_Organization/MPRAGE_250um/view#320,440,440,0,1.30
Reproduce locally with https://static.webknossos.org/data/MPRAGE_250um.zip

MichaelBuessemeyer:
This PR makes the far clipping plane of the 3d port adaptive to the dataset extent in nm. This should prevent errors of like parts or all planes being clipped away in the 3d viewport. An example of such a problem can be found in #3849.

- https://adaptivefarplanefortdviewport.webknossos.xyz

### Steps to test:
- Download this [MPRAGE_250um dataset](https://static.webknossos.org/data/MPRAGE_250um.zip) and import it into wk (locally or testing instance).
- If you view it on the current master like running on wk.org (see [link](https://webknossos.org/datasets/Demo_Organization/MPRAGE_250um/view#320,440,440,0,1.30)) the 3d Viewport should be broken.
- Now view the dataset with the changes of this branch (locally or testing instance). The 3d viewport should now actually render all planes as expected.
- Play around with the rotation and the rotation buttons of the 3d viewport and try to find any position in which clipping still kicks in and the planes are not fully rendered. There should not be a case like this. At least I could not find any case like this during my testing.
- Try the same things on a "normal" dataset with a resolution in the nanometer range. The 3d viewport should also work here.

### Issues:
- fixes #3849

------
- [x] Ready for review


frontend/javascripts/oxalis/controller/camera_controller.ts
```diff
@@ -241,6 +243,10 @@ export function rotate3DViewTo(id: OrthoView, animate: boolean = true): void {
     const paddingFactor = 1.1;
     width = Math.sqrt(datasetExtent.width ** 2 + datasetExtent.height ** 2) * paddingFactor;
     height = width / aspectRatio;
+    // Take the whole maximum diagonal width of the dataset to get the possible maximum extent of the dataset.
+    // This is used as an indication to set the far plane.
+    // A padding factor of 2 is used to ensure the adaptive far value is big enough in any case.
+    far = Math.sqrt(width ** 2 + datasetExtent.depth ** 2) * 2;
```
MichaelBuessemeyer, line 249: That is the approximation I came up with. Does it make sense? The padding factor `1.1` does not seem sufficient here. Instead, something bigger is needed. I do not exactly know why `1.1` is not big enough and did not want to spend too much time thinking about this. Therefore I tried out 2 and it looks to be sufficient.

daniel-wer: > I do not exactly know why 1.1 is not big enough and did not want to spend too much time thinking about this.

Reading the code, I would assume this is due to the `clippingOffsetFactor` which is the maximum of width, height, and depth. The camera position is offset by that value in all 3 dimensions to make sure the whole dataset is visible. Therefore, a factor of 2 of the dataset diagonal makes sense to me. However, the code is new to me and I could have very well overlooked something.

frontend/javascripts/oxalis/controller/camera_controller.ts
```diff
@@ -225,6 +226,7 @@ export function rotate3DViewTo(id: OrthoView, animate: boolean = true): void {
   // Use width and height to keep the same zoom.
   let width = tdCamera.right - tdCamera.left;
   let height = tdCamera.top - tdCamera.bottom;
+  let far = tdCamera.far;
```
daniel-wer, line 229: I'm wondering why this value is recomputed every time the `rotate3DViewTo` function is called. Isn't this value static and can be computed once when initializing the cameras (or once the dataset extent is known)?

daniel-wer (commented):
Thank you for looking into this and finding a fix :tada:

I have one main suggestion apart from small fixes.

frontend/javascripts/oxalis/controller/camera_controller.ts
philippotto: Also note that there is already the following line in `componentDidMount` of the `CameraController`:

```
const far = 8000000;
```

This line should be removed and probably replaced with the new calculation.

On another note: Could `far` simply be set to `Number.MAX_VALUE`? Or does something speak against this?

MichaelBuessemeyer: Yeah, moving the code to `componentDidMount` in `CameraController` sounds good 👍 .

> Could far simply be set to Number.MAX_VALUE? Or does something speak against this?

Honestly, I cannot think of an argument against this. As far as I know, we want to render everything within the scene of the 3d viewport, right? (At least we do not want to clip scene elements that are far in the back; only those who are cut off on the left, right, upper, down side)
What do you think @daniel-wer?

Should I do the `const far = Number.MAX_VALUE;` change for every ortho viewport or only for the 3d viewport. I think it should not affect the other viewports as well.

daniel-wer: > Honestly, I cannot think of an argument against this. As far as I know, we want to render everything within the scene of the 3d viewport, right? (At least we do not want to clip scene elements that are far in the back; only those who are cut off on the left, right, upper, down side)
What do you think @daniel-wer?

There could be issues with number inaccuracies or performance implications, but that should be testable.

> Should I do the const far = Number.MAX_VALUE; change for every ortho viewport or only for the 3d viewport. I think it should not affect the other viewports as well.

I would say that only makes sense for the 3d viewport since in the other viewports the plane is offset by the clipping distance, to make nodes that would be behind the plane visible, and everything further away is blocked by the plane. I'm not sure how the maximum value for the clipping distance is chosen and/or how the far value for the orthogonal viewports is chosen, though. I don't think it makes sense to set far to a number larger than the maximum current clipping distance (+ a potential offset due to the camera position, not sure).

MichaelBuessemeyer: > There could be issues with number inaccuracies or performance implications, but that should be testable.

How would you test for this? Create an extremely large dataset?

MichaelBuessemeyer: Ok, using ``Number.MAX_VALUE` is not an option. The gif below shows the rendering bugs I got by using that for the 3d viewport as a value for the far plane.

output

Thus I'll use the heuristic as done previously but now setting it only once in the `componentDidMount`.

philippotto: Thanks for giving it a try!

daniel-wer (approved):
Code LGTM, thank you for testing out the alternative possible solutions! :+1:

PR #6224, opened 2022-05-18
Fix zarray character for neutral byte order

### LINKED ISSUE #6222: uint8 needs to be specified as `|u1`
The current `.zarray` writes out `dtype` properties for each mag. For `uint8` datasets it prints `<u1`, which is incorrect because 1-byte values don't have a specified byte order. Instead it should output `|u1`. Also for `int8` => `|i1`.
See https://data-humerus.webknossos.org/data/zarr/scalable_minds/l4dense_motta_et_al_demo/color/1-1-1/.zarray

fm3:
Fix zarray character for neutral byte order

### Issues:
- fixes #6222


webknossos-datastore/app/com/scalableminds/webknossos/datastore/controllers/ZarrStreamingController.scala
```diff
@@ -268,14 +268,14 @@ class ZarrStreamingController @Inject()(
     token.orElse(request.headers.get("X-Auth-Token"))
 
   private def zarrDtypeFromElementClass(elementClass: ElementClass.Value): (Int, String) = elementClass match {
-    case ElementClass.uint8  => (1, "<u1")
+    case ElementClass.uint8  => (1, "|u1")
     case ElementClass.uint16 => (1, "<u2")
     case ElementClass.uint24 => (3, "<u1")
```
fm3, line 273:
```suggestion
    case ElementClass.uint24 => (3, "|u1")
```

normanrz (approved):

leowe (approved):

PR #6229, opened 2022-05-20
Allow restricting mags for new volume layer

### LINKED ISSUE #6059: Allow to restrict magnifications when adding new volume layer

Dagobert42:
https://allowrestrictingmagsfornewvolumelayer.webknossos.xyz/

### Steps to test:
- create a new layer using the create explorative modal from dashboard
- and using the add volume annotation layer button

### Issues:
- fixes #6059 

------
- [x] Ready for review


fm3:
Maybe this is a good opportunity to adapt (parts of) the front-end to the new terminology as described in https://docs.webknossos.org/webknossos/terminology.html (in this case mostly magnification vs resolution)?

Dagobert42:
I would like that. Working with mags for the first time also had me a bit confused.

frontend/javascripts/oxalis/view/left-border-tabs/modals/add_volume_layer_modal.tsx
```diff
@@ -25,31 +31,48 @@ export default function AddVolumeLayerModal({
...
-
+  let selectedSegmentationLayer = null;
   const handleAddVolumeLayer = async () => {
     await api.tracing.save();
 
+    const minResolutionAllowed = Math.max(
+      ...datasetResolutionInfo.getResolutionByIndexOrThrow(resolutionIndices[0]),
+    );
+    const maxResolutionAllowed = Math.max(
+      ...datasetResolutionInfo.getResolutionByIndexOrThrow(resolutionIndices[1]),
```
philippotto, line 52: `resolutionIndices` are initialized to dummy values which would cause an error here if the user doesn't change the slider. I'll add another comment with a code snippet how to fix this.

frontend/javascripts/dashboard/advanced_dataset/create_explorative_modal.tsx
```diff
@@ -63,6 +70,82 @@ export function NewVolumeLayerSelection({
...
+    const datasetFallbackLayerResolutionInfo = getResolutionInfo(
+      selectedSegmentationLayer.resolutions,
+    );
+    highestResolutionIndex = datasetFallbackLayerResolutionInfo.getHighestResolutionIndex();
+    lowestResolutionIndex = datasetFallbackLayerResolutionInfo.getClosestExistingIndex(0);
+  }
+
+  const highResolutionIndex = Math.min(highestResolutionIndex, resolutionIndices[1]);
+  const lowResolutionIndex = Math.max(lowestResolutionIndex, resolutionIndices[0]);
+
```
philippotto, line 92: See my other comment. This component should ideally ensure that the low/high values are always valid. When changing these values, this is already ensured by antd since you provide the correct min/max limits. However, if the user never changes the slider, the initial values are likely invalid.

Therefore, I suggest to set the values correctly on initial render (or in case the valid range changes which probably can never happen). This can be done with `useEffect` which will execute the passed function, if one of the elements of the second parameter (the "dependencies") change. Adding the following after line 91 should do the trick:

```
  useEffect(() => {
    setResolutionIndices([lowestResolutionIndex, highestResolutionIndex]);
  }, [lowestResolutionIndex, highestResolutionIndex]);
```

(`useEffect` needs to be imported)

philippotto (changes requested):
Looks great :) However, I found a small bug. See my comments :)

philippotto:
> Maybe this is a good opportunity to adapt (parts of) the front-end to the new terminology as described in https://docs.webknossos.org/webknossos/terminology.html (in this case mostly magnification vs resolution)?

Good point :+1: However, I discussed with @Dagobert42 that a separate PR would be better for this.

frontend/javascripts/oxalis/view/left-border-tabs/modals/add_volume_layer_modal.tsx
```diff
@@ -34,6 +34,13 @@ export default function AddVolumeLayerModal({
   const datasetResolutionInfo = getDatasetResolutionInfo(dataset);
   const [resolutionIndices, setResolutionIndices] = useState([0, 10000]);
 
+  useEffect(() => {
```
philippotto, line 37: I think this is redundant here, as the same code does already exist within `RestrictResolutionSlider`. Can you remove it here and test that the UI still works when not changing the slider?

Dagobert42: Got it. Yes, works as expected :)

philippotto (approved):
Awesome :)

PR #6230, opened 2022-05-20
Merge "Shared Annotations" with "My annotations"

### LINKED ISSUE #6142: Rework annotations table towards new collaboration features
See screenshot in:
https://www.notion.so/scalableminds/2022-03-08-Collaboration-65cdfa42e1a6486f8d28e8a29b699d72

philippotto:
- [X] @fm3 Could you have a look at my backend changes? I'm not sure whether this is the right approach (the "compact writes" got a bit more expensive due to the additional look ups). Also, I'm not sure how the pagination should be built when merging "own" with "shared" annotations.

- https://dashboardcollab.webknossos.xyz

### Steps to test:
- Set up two users (one should already exist; for the second one, register in an incognito browser tab and enable the account from the first user)
- Create multiple annotations from both accounts
- Assign a common team to both users and share some annotations with that team
- The dashboard "annotation" tab should always show the own annotations *and* the ones which are shared with that user
- Shared annotations should not be editable via the "Archive", "Archive All" or "Edit Name" buttons.
- Test the filtering UI in the "Owners & Teams" column

### Todo:
- [x] unify `user` and `owner` attribute in annotations

### Issues:
- fixes #6142

------
- [x] Ready for review


frontend/javascripts/dashboard/explorative_annotations_view.tsx
```diff
@@ -145,13 +159,15 @@ class ExplorativeAnnotationsView extends React.PureComponent<Props, State> {
       this.setState({
         isLoading: true,
       });
+
       const tracings =
         this.props.userId != null
-          ? await getCompactAnnotationsForUser(this.props.userId, showArchivedTracings, pageNumber)
-          : await getCompactAnnotations(showArchivedTracings, pageNumber);
+          ? // todo: also implement pendant for getCompactAnnotationsForUser ?
```
philippotto, line 165: todo

docs/dashboard.md
```diff
@@ -39,11 +39,12 @@ Each annotation can be renamed to reflect its content.
...
 
 
-## Shared Annotations
-This screen shows all annotations that were shared by other collaborators of your organization. Only annotations shared through the "Team Sharing" mechanic will be listed provided your user account is part of the respective team. Read more about sharing your own annotations in the [Sharing guide](./sharing.md#annotation-sharing).
+### Sharing Annotations
+The annotations tab also shows all annotations that were shared by other collaborators of your organization. Only annotations shared through the "Team Sharing" mechanic will be listed provided your user account is part of the respective team. Read more about sharing your own annotations in the [Sharing guide](./sharing.md#annotation-sharing).
 
 You can view the linked annotations (read-only) or copy them to your account for modification.
 
+todo: update image
```
philippotto, line 47: todo

frontend/javascripts/types/api_flow_types.ts
```diff
@@ -332,8 +332,53 @@ export type APIAnnotationCompact = {
...
+    visibility,
+    name,
+    state,
+    stats,
+    tags,
+    tracingTime,
+    typ,
+    // todo: clean up when user and owner are unified
+    // @ts-ignore
+    owner: owner || user,
```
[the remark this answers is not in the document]
philippotto: blocked by https://github.com/scalableminds/webknossos/pull/6250

philippotto:
@daniel-wer I think this PR is ready for review :)

daniel-wer (commented):
Works really well! Thanks for drive-by-improving the types!

Some smaller issues I noticed during testing:
- [x] If I select the annotations tab, the URL switches to `https://dashboardcollab.webknossos.xyz/dashboard/shared` - I would have expected `/annotations`, but maybe this is by design
- [x] The Owner/Teams filter in the table header works great, but doesn't seem to be taken into consideration by the "Archive All" button. Although only one annotation (which is read-only and cannot be archived) is visible due to the filter, the message reads "Are you sure you want to archive all 2 explorative annotations matching the ..."
- [x] If I archive or reopen an annotation and then switch to the respective view where it can be found, it turns up twice. A page reload "repairs" that.

frontend/javascripts/dashboard/explorative_annotations_view.tsx
```diff
@@ -193,35 +215,33 @@ class ExplorativeAnnotationsView extends React.PureComponent<Props, State> {
...
+      shouldShowArchivedTracings,
+    );
   };
 
+  _updateAnnotationWithArchiveAction = (
+    annotation: APIAnnotationCompact,
+    type: "finish" | "reopen",
+  ): APIAnnotationCompact => ({
+    ...annotation,
+    state: type == "reopen" ? "Active" : "Finished",
```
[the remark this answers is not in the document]
philippotto: this was not intentional. I escalated the linter so that it will throw an error for this. I think, this rule got softened during the TS conversion.

philippotto:
Thanks for catching these bugs! I fixed them all now (plus some additional potential culprits) and also opened https://github.com/scalableminds/webknossos/issues/6275, since the entire tab could use some rethinking. Please have another look :)

frontend/javascripts/dashboard/explorative_annotations_view.tsx
```diff
@@ -545,6 +560,9 @@ class ExplorativeAnnotationsView extends React.PureComponent<Props, State> {
         scroll={{
           x: "max-content",
         }}
+        onChange={(_pagination, _filters, _sorter, extra) => {
+          this.setState({ filteredAnnotations: extra.currentDataSource });
```
daniel-wer, line 564: I think `filteredAnnotations` needs to be initialized when rendering the table, since this line will only be triggered onChange. The current symptome is that when pressing "Archive All" as the first action after loading the page, there is a toast stating "No annotations available to archive.". After filtering once, the archiving works as expected (searching doesn't make it work, only changing the Owner & Teams filter).

daniel-wer (commented):
There is one issue left with the "Archive All" functionality - apart from that LGTM :+1:

philippotto:
Great catch, @daniel-wer :+1: It took me a bit to find a good workaround, since I found the `onChange` prop to be unsuitable to fix all issues. I got a solution which works with the currently visible items of the active page. It's not the same as before, since all other pages are ignored, but I think this even has benefits (the user will only mutate the items which are really visible). Please have a look at my latest commit (and the comments it contains) :)

daniel-wer (approved):
Nice that you found a workaround! I agree that only modifying the visible results from the active page is even better than the old behavior :+1:

PR #6235, opened 2022-05-24
Rework volume interpolation feature to be shortcut-bound and support depths > 2

### LINKED ISSUE #6183: Follow-up: Enable volume Interpolation with depth > 2 and clarify usability
## Detailed Description
The interpolation itself already supports higher depths, but the usability might need some tuning. We should clarify the following questions:

- Is a depth of 4 meant as the exact depth or as the maximum depth? In other words: what happens if the depth is set to 4, but I only advance by 3 slices and brush. Will this
  - use the volume data from 4 slices ago for interpolation (which is likely interpolated itself and therefore not a good source)
  - use the volume data from 3 slices ago for interpolation
  - don't do any interpolation?
- In general, can/should we let the user know from which slice the next brush action would use data for the interpolation?
- Can we let the user know which slices where drawn by interpolation so that proof-reading of interpolated slices is easy to do after annotating a bunch of slices?

## Context
See https://github.com/scalableminds/webknossos/pull/6162/

philippotto:
From changelog:
- Added support for segment interpolation with depths > 2. Also, the feature was changed to work on an explicit trigger (either via the button in the toolbar or via the shortcut V). When triggering the interpolation, the current segment id is interpolated between the current slice and the least-recently annotated slice.
- Removed the feature to copy a segment from the previous/next slice with the V shortcut. Use the new volume interpolation feature instead (also bound to V and available via the toolbar).

- https://deepinterpolation.webknossos.xyz/

### Steps to test:
- create a new annotation
- label a cell in slice X
- move forward to slice X + 3
- press "V" shortcut --> X + 1 and X + 2 should be filled automatically
- repeat the same with varying constellations (e.g., z distance == 1 which should disable the button and show a toast when pressing V; distance > 8; interpolating for a slice where the segment does not exist)

### Issues:
- fixes #6183 

------
- [X] Updated [documentation](../blob/master/docs) if applicable
- [x] Ready for review


MichaelBuessemeyer (changes requested):
Awesome PR 🎉 . Thanks for making the interpolation available for higher distances 📐 .

Codewise I only found some minor things.

Bugs I noticed.  I tested on the dev instance in [this tracing](https://deepinterpolation.webknossos.xyz/annotations/Explorational/628dfa440100007f006a56a3#3276,3214,1091,0,3.221):
- [ ]  In resolution 8-8-2 I drew on one slice in the xy viewport, and moved two slices forward by pressing space two times. Then I annotated a little again. Now the interpolation button is disabled (which imo shouldn't be the case) and when I press `v` I get the `Could not interpolate segment because last labeled slice should be at least 2 slices away` error message. In the described scenario the two annotated slices should be far enough away from each other for the interpolation to work, right? So I think this is a bug. While reading the code I could not spot why this might be the case.

Usability nice to haves:
- [x] When the interpolation button is disabled, the button is only disabled. There is no tooltip telling why the button is disabled. I would like to have an explanatory tooltip.
- [x] I don't know whether it is by design, but the volume interpolation button is wider than the other tool buttons. Maybe it would look nicer if it would have the same length as the other tool buttons. 
image
- [ ] There is no visual feedback on whether the interpolation is still being computed or already finished. Something like a little yellow dot for "I am still computing" and green for "I finished interpolating" would be nice imo.

philippotto:
@MichaelBuessemeyer Thanks for your feedback! I think, I fixed all bugs (the wide button and missing toolbar was also a bug, caused by switching from `Button` to `ButtonComponent`). Please have another look :)

> There is no visual feedback on whether the interpolation is still being computed or already finished. Something like a little yellow dot for "I am still computing" and green for "I finished interpolating" would be nice imo.

The mouse cursor should turn into a waiting cursor when wk is busy. I think, this generalizes better to a yellow dot or similar. However, I don't see that cursor (maybe the workload is not large enough?). `takeEveryUnlessBusy` should take care of this actually... If I got another second, I'll try to investigate, but I'd hope that this doesn't block the PR.

MichaelBuessemeyer (approved):
Thanks for fixing my address issues. I couldn't reproduce them anymore. 🚀 

> The mouse cursor should turn into a waiting cursor when wk is busy. I think, this generalizes better to a yellow dot or similar.

Ah, using the cursor as an indication sounds good.

> However, I don't see that cursor (maybe the workload is not large enough?). takeEveryUnlessBusy should take care of this actually... If I got another second, I'll try to investigate, 

Same here, for my setup (MacOS + Chrome). The cursor does not turn into a waiting-like symbol :/. It might actually be that the computation is too fast 🚀 .

> but I'd hope that this doesn't block the PR.

Agreed, this is be something for a follow-up pr.

PR #6236, opened 2022-05-24
Allow multiple contributors per annotation

### LINKED ISSUE #6227: Allow multiple contributors per annotation
- [x] include user id in update actions, show user in update action log
- [ ] aggregate list of contributors in report, store in postgres
- [x] adapt conflict error message
- [x] heuristic to notify in advance that this is currently being modified by someone else. e.g. check last update action timestamp + author
- [x] sharing: permission bool, make editable by others, enforce, tell front-end
- [ ] reduce save delay 30s to lower value?

part of https://github.com/scalableminds/webknossos/issues/6010

fm3:
It is now possible to allow other users, who may see your annotation, to also edit it.

- https://annotationcontributors.webknossos.xyz

### Backend API Changes
 - annotation json now includes new field `contributors: List[User]`, listing the users who pushed update actions, but who are not the owner
 - update actions in `updateActionLog` now contain new optional field `authorId: String` containing a user ID. Legacy update actions without this field should be considered to be authored by the annotation owner.
 - new route `PATCH /api/annotations/:typ/:id/othersMayEdit?othersMayEdit=true` allows to set this permission
 - users who can see an annotation with `othersMayEdit==true` will get `allowUpdate: true` in the annotation json
 - annotation json now includes new field `othersMayEdit: boolean` indicating whether the above option has been set for the annotation

### Steps to test:
- ensure you have two separate users set up (one logged-in in incognito tab)
- create a new annotation
- open the share modal and assign a orga team to the "allowed teams"
- save
- open the annotation link for user B and check that the annotation is in read-only mode
- open the share modal again for user A and change the "allow others" setting to true
- reload the annotation for user B --> annotation should be editable
- trace a bit in window B and save
- wait a bit in window A --> a toast should appear warning about conflicts

### Issues:
- fixes #6227 

------
- [x] Ready for review


app/controllers/AnnotationController.scala
```diff
@@ -436,7 +436,7 @@ class AnnotationController @Inject()(
     sil.SecuredAction.async { implicit request =>
       for {
         annotation <- provider.provideAnnotation(typ, id, request.identity)
-        _ <- bool2Fox(annotation._user == request.identity._id && annotation.visibility != AnnotationVisibility.Private) ?~> "notAllowed" ~> FORBIDDEN
+        _ <- bool2Fox(annotation._user == request.identity._id) ?~> "notAllowed" ~> FORBIDDEN
```
philippotto, line 439: I removed the `&& annotation.visibility != AnnotationVisibility.Private` bit, since I found it weird that I could not change the may-others-edit boolean (even if it was true), if the visibility was set to true. also the user might want to enable its team manager to edit the annotation, for example.

frontend/javascripts/oxalis/view/right-border-tabs/dataset_info_tab_view.tsx
```diff
@@ -525,24 +526,22 @@ class DatasetInfoTabView extends React.PureComponent<Props, State> {
...
-      );
-    }
-
-    // Active user is owner
-    return null;
+    return (
+      <div>
+        <div>Owner: {formatUserName(activeUser, owner)}</div>
+        <div>
+          Contributors: {contributors.map((user) => formatUserName(activeUser, user)).join(", ")}
```
philippotto, line 541: when there are no contributors, this should read a bit nicer than `contributors:`.

frontend/javascripts/oxalis/model_initialization.ts
```diff
@@ -281,6 +281,7 @@ function initializeTracing(_annotation: APIAnnotation, serverTracings: Array<Ser
         ...annotation,
         restrictions: {
           ...annotation.restrictions,
+          // todo: is this just an alias? can this be removed?
           allowSave: annotation.restrictions.allowUpdate,
```
philippotto, line 285: @daniel-wer Do you know why the front-end unpacks the `allowUpdate` restriction to `allowSave`? Is there a difference between these two?

philippotto:
I think, the PR is ready for review. @fm3 I fiddled a bit in the back-end code (see my one comment and also the `newestVersion` route I added). Would be great if you could have a look :)

fm3:
Sure, looks good :) removing the private check makes sense to me. On the latestVersion route, I added the token access check that was still missing. Not that it would have given away anything critical :shrug:  but for the sake of completeness

frontend/javascripts/oxalis/model_initialization.ts
daniel-wer: Yes there are differences. In line 277 (a couple of lines above) the sandbox case is handled, where `allowUpdate` is true, but `allowSave` is false.
`allowSave` is also used to disable saving when the `Disable Saving` button from the dropdown next to the save button is pressed. In that case, updates should still be allowed which also requires this distinction.

philippotto: Ah, great. Thanks for clearing up my confusion!

philippotto:
@daniel-wer the CI is now finally green and the dev instance is ready to be tested :)

daniel-wer (changes requested):
Nice and clean implementation :+1: I like the notifications if the annotation was modified somewhere else. These will certainly be helpful!

Overall, everything worked pretty well. I encountered a couple of minor issues. Especially the conflict-case should be polished to work well imo:

- [x] When opening a shared annotation from a user account that is not the owner, the info tab didn't contain any "Owner" or "Contributors" information. Is this by design?
  read_only_no_owner
- [x] When modifying a shared annotation from another user while the annotation is open by the owner, saving will trigger the version conflict owner. However, (especially) if the modifications consist of multiple versions, a lot of error messages are shown one after the other, without the page reloading. See the gif, where it was actually impossible to confirm all errors:
infinite_errors
- [x] In the shared annotations tab, the Creator column is empty. This is probably fixed by #6230.
shared_annotations_empty_creator

frontend/javascripts/test/helpers/apiHelpers.ts
```diff
@@ -193,6 +194,10 @@ export function __setupOxalis(
     );
   Request.receiveJSON.returns(Promise.resolve({}));
   Request.sendJSONReceiveJSON.returns(Promise.resolve({}));
+  Request.sendJSONReceiveJSON
+    .withArgs(sinon.match((arg) => arg == `/api/users/${dummyUser.id}/taskTypeId`))
```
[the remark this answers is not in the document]
philippotto: This mock relates to the `updateLastTaskTypeIdOfUser` api. There's a saga which updates the last active task type that wasn't run before, because there was no active user. That's why I had to mock it, too. I added a comment to explain to what the route relates.

daniel-wer (commented):
Almost LGTM!

I encountered one more weirdness. If an annotation is shared and made editable, but the user that got access opens the annotations via the sharing link, the annotation appears editable, but save requests are denied. The error message is `Token may be expired, consider reloading. Access forbidden: No write access to tracing`. Removing the token from the URL (or opening the annotation from the "Shared Annotations" tab [which also doesn't have a token]) makes it work as expected.

After fixing this and the owner/contributors, this should be ready to be merged.

daniel-wer:
> Interesting, I think, you ran into https://github.com/scalableminds/webknossos/issues/6278 which was reported last week by another user. This problematic scenario probably becomes way more likely when this PR is merged. Therefore, I would implement the solution suggestion (2) from the issue. There is a caveat to it, but at least, it's better than the current behavior.

Agreed, that solution would be better. One option I'm not seeing in the list is for the backend to ignore the sent token at the tracingstore if the user has access to the annotation, anyways. This would mitigate the caveat of suggestion (2) in my opinion. Have you thought about or discussed that?

philippotto:
> Agreed, that solution would be better. One option I'm not seeing in the list is for the backend to ignore the sent token at the tracingstore if the user has access to the annotation, anyways. This would mitigate the caveat of suggestion (2) in my opinion. Have you thought about or discussed that?

I think, the tracing store doesn't know whether a user has access to the annotation. The entire access is just done via tokens. But maybe I'm wrong. Maybe @fm3 can confirm or correct this :)

fm3:
Exactly. The tracingstore does not get the user cookies. So the front-end needs to send a token identifying the user (as generated by userToken/generate). This is exactly what is missing (because instead the front-end sends the sharingToken)

frontend/javascripts/oxalis/model/sagas/save_saga.ts
```diff
@@ -1132,7 +1151,13 @@ function* watchForSaveConflicts() {
   while (true) {
     const interval = yield* call(getPollInterval);
     yield* call(sleep, interval);
-    yield* call(checkForNewVersion);
+    try {
+      yield* call(checkForNewVersion);
+    } catch (exception) {
```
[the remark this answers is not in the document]
philippotto: Yes-ish. I'm not sure how exactly this escalated (I tried to debug the other issue which is why this might have been only a secondary effect). Then, I noticed that this could be a problem, if the client is offline for example. But you are right: Sending this to airbrake would be cleaner. I'll do so.

philippotto:
> I fixed the compiler error and the schema version number.

Awesome, thanks!

> One thing I noticed: The MakeMappingEditable update action is inserted by the backend. At the time, the user id is not known in the tracingstore so it won’t have an author id. If we want to add that, we need to send the author id with that specific request.
>
> And a related, more general concern: there is currently no way to validate that the sent author id is correct. Theoretically, by changing the request content, a user with write access could forge their update actions to look like they are authored by someone else. I’m not sure how much of a concern this is, but maybe it would be worth to take steps against this. What do you think?

Could the backend infer the `authorId` via cookie? That would probably be the clean approach, right?
I don't think that forging the identity is a relevant attack vector, since the user was granted trust, anyway, by inviting him/her to edit the annotation. But still, it's not ideal.. If this would solve the `MakeMappingEditable` issue, too, that would be great. I don't think, sending the author id there, too, is the right way to solve this..

----------------
I wanted to tackle the last two checkboxes from this PR, but I ran into the following problems:
- [ ] if user A creates an annotation and shares it with B so that B can modify it, B cannot see this annotation in its dashboard. I would have assumed that it should pop up there, since it is "readable" for B? maybe we didn't think about this case before, since the shared annotation table still existed when this PR was opened, but I think it would make sense to adapt this.
- [ ] if user B is a regular user without any special rights, `owner` is set to `null` when fetching the annotation info. I vaguely remember that we discussed this before and that we agreed to fill in the owner value, since one can expect that this info is shared with anyone who can access the annotation? otherwise, one wouldn't see who is the owner of the annotation which was sent to you.

@fm3 What do you think about these two points?

fm3:
> Could the backend infer the authorId via cookie?
Unfortunately that is not easy, as this request goes to the tracingstore, where no cookies are sent and there is no concept there yet for identifying the user. We would have to add an RPC to webKnossos to look up the user id by the token. While I agree that this would be the safest approach (both for sent update actions and for the makeMappingEditable call), it would be some effort. I agree that the attack vector is not super critical, so maybe let’s make that a follow-up issue.

> B cannot see this annotation in their dashboard
That’s weird, how exactly did A share it with B? I’d have thought there were no changes to this mechanism, and it still uses the team-based sharing?

> owner is set to null
Fair enough, I’ll change that

philippotto:
> That’s weird, how exactly did A share it with B? I’d have thought there were no changes to this mechanism, and it still uses the team-based sharing?

Hmm, good point. I think I shared it via a url+token which is not really the classic use case we imagined. I didn't think much about it when testing, but it somehow makes sense that it works. However, then, the annotation table probably needs to be populated by also checking whether the active user appears in the contributor list of an annotation? Do you think this is possible / makes sense?
Alternatively, we could limit the "may others edit" checkbox so that it can only be checked if a team was added for sharing. But I think this can still lead to weird cases (e.g., you added a team, but shared the link with someone outside of the team).

fm3:
> also checking whether the active user appears in the contributor list of an annotation

Ah, I see. Yes, I think that makes sense. I’ll try implementing that :)

philippotto:
> > also checking whether the active user appears in the contributor list of an annotation
> 
> Ah, I see. Yes, I think that makes sense. I’ll try implementing that :)

Great!

Another related issue could be the following scenario:
- I share an annotation with someone and allow them to edit it
- they edit the annotation and it appears in their dashboard
- I disable the ability to edit the annotation for them
- should the annotation still appear in their dashboard? if so, it probably needs to be marked as read-only

Maybe this will work out-of-the-box, but I'll check that after the other stuff works.

fm3:
I’d say it should still be listed then, yes. I’d guess it would already be marked as read-only in this case, for the othersMayEdit boolean would be false and the owner not the person who is viewing the list?

philippotto:
@daniel-wer This PR is ready for another round of feedback/testing :)

daniel-wer (approved):
Nice, code LGTM and works as expected :+1:

PR #6240, opened 2022-05-25
Upgrade to scala 2.12.15, sbt 1.6.2, remove silencer compiler plugin

fm3:
scala 2.12.15 comes with its own warning-silencing mechanism, no need for the silencer plugin anymore.

CI run should be enough test, I tested that I can still compile, run, view and annotate data.


leowe (approved):
Nice👍

PR #6245, opened 2022-05-27
Update volume interpolation video in docs

philippotto:
plus minor improvements which caught my eye

- [X] Ready for review


frontend/javascripts/oxalis/view/right-border-tabs/skeleton_tab_view.tsx
```diff
@@ -537,7 +537,7 @@ class SkeletonTabView extends React.PureComponent<Props, State> {
       return;
     }
 
-    await this.setState({
```
philippotto, line 540: the await does not have an effect here

hotzenklotz (approved):

Issue #6241, opened 2022-05-26
Refactor backend: dataclass static function naming: from vs createFrom


fm3:
Some of our dataClasses have static methods `from(x)`, some have `createFrom(x)`, some have `parse(x)`. 
Also, if there are multiple sources a class instance can be created from, the overloading is sometimes done with actual overloading (all named the same) and sometimes the names differ (`fromArray`, `fromMagLiteral`).

I’d say we should define a unified convention and re-name functions we come across accordingly.

Personally, I’d vote to use the last mentioned variant, i.e. overloading explicitly in names, and no “create” – The [Vec3Int class ](https://github.com/scalableminds/webknossos/blob/master/util/src/main/scala/com/scalableminds/util/geometry/Vec3Int.scala#L56) already uses this convention.

@leowe @jstriebel any preferences?

fm3:
Thanks for your input :) Then I’ll set the new convention as I suggested above. Yes, the `create` prefix should be removed. Ideally it should say `fromSomething(x)`.

Note that this is only about dataclass static methods. More complex creation operations like `taskCreationService.createTracingsFromBaseAnnotations(…)` should still have the create prefix to indicate what is happening.

jstriebel:
Sorry, I missed to answer here. I fully agree, using `from()` and `fromThisThing()` sounds coherent and seems to be a more clear change :+1:

PR #6249, opened 2022-05-31
Change parameter precedence when creating task from base

### LINKED ISSUE #6200: Respect "Starting Position" in task when using an Annotation ID as Base
## Context
Tasks with a base annotation will always use the edit position from that base annotation instead of the one which is provided when creating a task.

## Steps to Reproduce the bug

1. Create a new task type with volume enabled
2. Create an annotation, go to position `100, 100, 100`, save and copy its id
3. Create a new task, select the task type from (1) and select "Use Annotation ID as Base" where the id from (2) should be pasted in
4. Enter a Starting Position (e.g., `1, 2, 3`)
5. Go to dashboard, get a new task and open it
6. The active position should be the one from (4). Instead, it is `100, 100, 100`.

## Your Environment for bug

fm3:
When creating a task from a base annotation, the starting position/rotation and bounding box as specified during task creation are now used and overwrite the ones from the original base annotation.

- https://taskbaseparameters.webknossos.xyz

### Steps to test:
- Create an explorative annotation, trace some, copy id
- Create a task, select base annotation, insert above id
- When requesting the task, you should start at the edit position + rotation as specified during task creation, not the ones from said explorative annotation.

### Issues:
- fixes #6200 
- contributes to #6241

------
- [x] Ready for review


jstriebel (approved):
Nice, I like the new refactorings! LGTM :ship:

PR #6250, opened 2022-06-01
Unify user vs owner field in annotation (it is now called owner)

fm3:
- https://annotationowner.webknossos.xyz

### Steps to test:
- Create annotation, trace some, go to dashboard
- everything should still work as expected

------
- [ ] Adapted [wk-libs python client](https://github.com/scalableminds/webknossos-libs/tree/master/webknossos/webknossos/client) – API change is backwards compatible, but https://github.com/scalableminds/webknossos-libs/issues/743 should be done soon.
- [x] Ready for review


frontend/javascripts/admin/task/task_annotation_view.tsx
```diff
@@ -227,8 +227,6 @@ class TaskAnnotationView extends React.PureComponent<Props, State> {
               })
             }
             onChange={this.updateAnnotationState}
-            // @ts-expect-error ts-migrate(2769) FIXME: No overload matches this call.
-            userId={this.state.currentAnnotation.user.id}
```
philippotto, line 231: that prop wasn't used by the component

CHANGELOG.unreleased.md
```diff
@@ -13,6 +13,8 @@ For upgrade instructions, please check the [migration guide](MIGRATIONS.released
 ### Added
 
 ### Changed
+- An the api routes that return annotation info objects, the user field was renamed to owner. User still exists as an alias, but will be removed in a future release. [#6250](https://github.com/scalableminds/webknossos/pull/6250)
```
```suggestion
- For the api routes that return annotation info objects, the user field was renamed to owner. User still exists as an alias, but will be removed in a future release. [#6250](https://github.com/scalableminds/webknossos/pull/6250)
```

maybe?

frontend/javascripts/types/api_flow_types.ts
```diff
@@ -360,6 +360,8 @@ type APIAnnotationBase = APIAnnotationCompact & {
   readonly restrictions: APIRestrictions;
   readonly viewConfiguration?: AnnotationViewConfiguration | null | undefined;
   readonly settings: APISettings;
+  readonly owner?: APIUserBase;
```
daniel-wer, line 363: Is the owner field really optional?

daniel-wer (approved):
LGTM with one minor question :+1:

frontend/javascripts/types/api_flow_types.ts
philippotto: Yes, I think, the owner is not set if the current user does not have the permissions to view the user details (e.g., when it's public). However, this might change with #6230. @fm3 correct me if I'm wrong :)

fm3: I think that’s right, yes. We should double-check again as part of #6230 that every case gets a valid response, and see how much info about the user is required where.

Issue #104, opened 2013-12-11
data set problems


deployboy:
Reported by: Raphael Kneissl (raphael-kneissl@web.de) on master

<a href="https://timer.scm.io/repos/537237ce1a00001a003c4a14/issues/104/create?referer=github" target="_blank">Log Time</a>

PR #6254, opened 2022-06-01
Pass annotation layer name to export job

MichaelBuessemeyer:
This pr is complementing a wk-worker pr to fix the export tiff job. The main change is that the volume annotation layer name is now included in the call that starts the job. As `volume_tracing_id` and `volume_tracing_version` aren't needed anymore, they will be removed.

Accompanying this change is a "fix" that name of volume annotation layers were static instead of the readable label shown in the layer settings. This led to possible duplication of names displayed to the user in the bounding box export modal opened by clicking the download icon of a bounding box in the bounding box tab. 
Previous:
image
With the fix:
image

- https://___.webknossos.xyz

### Steps to test:
- Testing the fix listed above with the images:
  - Create a volume annotation, add multiple additional volume layers to the annotation and give them a different label.
  - The Export modal opened via the download icon on a bounding box in the bounding box tab should now contain the assigned labels and not the generic name "Volume annotation". This should make the buttons distinctive.
- Testing the export job fix: For that do the same steps as should be done to test the wk worker pr part. Make sure there is a local worker instance running on branch `fix-export-tiff`.
> - This pr needs to be tested with wk on branch `fix-export-tiff`. The corresponding pr is [here](https://github.com/scalableminds/webknossos/pull/6254).
> - A dataset with at least one color layer and one segmentation layer with mapping is needed.
> - Create an empty hybrid annotation and create a small bounding box. Start an export job for the color layer and one for the segment layer. Both should run through smoothly.
> - Activate a mapping for the segmentation layer and start the job again. This should also succeed.
> - Make some brush strokes in the annotation and then start another export job, followed by one where the mapping is disabled again.
> - Create a new volume annotation layer. Note that the name must not be the same as for the already existing volume annotation layer. Then start an export job for the empty layer. This job should crash with the error that the annotation layer is empty.
> - Brush something in the new annotation layer and then start a job again. This should run through smoothly.

### Issues:
- fixes #

### Connected PRs:
- [#104](https://github.com/scalableminds/webknossos-worker/pull/104) from the worker repo

------
- [x] Ready for review


philippotto (approved):
Ok, since the worker PR is merged, this can also be merged :) Call me when you've got time for the deployment.

PR #6255, opened 2022-06-01
Display warning when resolution gets too low during zooming

### LINKED ISSUE #6110: Show explaining hint when zooming out on z=1-downsampled datasets
Zooming out on datasets that are not downsampled in z is limited while viewports YZ or XZ are visible because of the amount of data that can be loaded.
In this case a hint should be shown recommending to maximize the XY viewport.

This is a common use case when inspecting alignments

Dagobert42:
- https://___.webknossos.xyz

### Steps to test:
- load a z=1 downsampled dataset and zoom out
- minimizing the XY viewport enough should show an explanatory warning once per session

### Issues:
- fixes #6110 

------
- [x] Ready for review


frontend/javascripts/oxalis/model/sagas/dataset_saga.ts
```diff
@@ -31,3 +32,29 @@ export function* watchMaximumRenderableLayers(): Saga<void> {
...
+    const currentRes = yield* select((state) => getCurrentResolution(state));
+    const currentZoomStep = yield* select((state) => state.flycam.zoomStep);
+    const minVoxelPerPixel = 0.5;
+
+    if (!downsamplingWarningDidShow) {
+      // checking the downsampled dimensions x and y
+      for (let i = 0; i < 2; i++) {
+        const voxelPerPixelXY = currentZoomStep / currentRes[i];
+        if (voxelPerPixelXY < minVoxelPerPixel) {
+          Toast.warning(messages["dataset.z1_downsampling_hint"]);
```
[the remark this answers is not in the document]
philippotto: Additionally, I'd suggest to pass an `onClose` handler to `Toast.warning` like this:

```
Toast.warning(..., {
  key: ...
  onClose: () => {
    userClosedWarning = true
  }
}
```

`userClosedWarning` would replace `downsamplingWarningDidShow`. That way, the behavior should be as follows:

- the warning shows and hides automatically depending on how the current quality condition is (it can appear multiple times)
- if the user explicitly closed the notification via the X button, the onClose handler will be triggered which sets the boolean. afterwards, the toast won't show again due to the boolean.

I think, this behavior would be quite cool :)

frontend/javascripts/oxalis/model/sagas/dataset_saga.ts
```diff
@@ -31,3 +32,29 @@ export function* watchMaximumRenderableLayers(): Saga<void> {
...
+          downsamplingWarningDidShow = false;
```
philippotto, line 49: see my other suggestion for `userClosedWarning`. if you follow that, the line here can be removed completely as the `onClose` handler will set the boolean to true.

```suggestion
```

frontend/javascripts/oxalis/model/sagas/dataset_saga.ts
```diff
@@ -31,3 +32,29 @@ export function* watchMaximumRenderableLayers(): Saga<void> {
 
   yield* takeEvery(["WK_READY", "UPDATE_LAYER_SETTING", "UPDATE_DATASET_SETTING"], warnMaybe);
 }
+
+let downsamplingWarningDidShow = false;
+export function* watchZ1Downsampling(): Saga<void> {
+  function* maybeShowWarning(): Saga<void> {
+    const currentRes = yield* select((state) => getCurrentResolution(state));
+    const currentZoomStep = yield* select((state) => state.flycam.zoomStep);
```
philippotto, line 40:
```suggestion
    const currentZoomStep = yield* select((state) => state.flycam.zoomStep);
    if (currentZoomStep < 1) {
      // If the user has zoomed into the data, the rendering quality is expected to be relative low.
      // Don't show any warnings in that case.
      return;
    }
```


otherwise, zooming in very far (which also results in minecraft-quality) would show the warning even though there's nothing to do about it.

philippotto (changes requested):
Very cool stuff! I left some comments to tweak the feature a bit :)

frontend/javascripts/oxalis/model/sagas/dataset_saga.ts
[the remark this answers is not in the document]
Dagobert42: Agreed! Thanks for the idea :)

philippotto:
@Dagobert42 awesome stuff :tada: I just tested it out and tweaked some small things, since they caught my eye:

- there was a `takeLatest` usage for `WK_READY` which should be a `take`. The difference is that only the latter one actually blocks the control flow. This is why wk crashed for me in the dashboard (the rest of the saga executed even though `WK_READY` is not dispatched in the dashboard).
- I tweaked the quality threshold even further because the old value still triggered the warning for the z1 dataset even when my viewport was maximized (at certain zoom levels).
- I optimized the saga invocation a bit to avoid that it's called unnecessarily often. For that, I used `takeLatest` which aborts already running invocations of `maybeShowWarning` if a newer action arrives (e.g., when zooming out in a fluent motion).

philippotto (changes requested):
really hyped about this :) see my last two suggestions, though :nerd_face:

frontend/javascripts/oxalis/model/sagas/dataset_saga.ts
```diff
@@ -31,3 +32,43 @@ export function* watchMaximumRenderableLayers(): Saga<void> {
...
+        if (voxelPerPixelXY < minVoxelPerPixel) {
+          Toast.warning(messages["dataset.z1_downsampling_hint"], {
+            sticky: true,
+            key: "DOWNSAMPLING_CAUSES_BAD_QUALITY",
+            onClose: setUserClosedWarningTrue,
+          });
+        } else {
+          Toast.close("DOWNSAMPLING_CAUSES_BAD_QUALITY");
+        }
+      }
```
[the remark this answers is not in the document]
Dagobert42: True, I believe I was originally thinking about iterating over the z dimension as well, but I guess there's really no use-case for that 😅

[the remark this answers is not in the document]
Dagobert42: Actually, this also solves the linter complaining about multiple instantiations of the arrow function (which was the reason for setUserClosedWarningTrue) 👍

frontend/javascripts/messages.ts
```diff
@@ -284,6 +284,8 @@ instead. Only enable this option if you understand its effect. All layers will n
     "This dataset location is marked as 'scratch' and meant for testing only. Please move this dataset to a permanent storage location and reimport it.",
   "dataset.resolution_mismatch":
     "This dataset contains multiple layers which differ in their resolution. Please convert the layers to make their resolutions match. Otherwise, rendering errors cannot be avoided.",
+  "dataset.z1_downsampling_hint":
+    "The currently rendered quality is not optimal due to the available magnifications and the viewport arrangement. To improve the quality try to increase the size of the XY viewport (e.g. by maximizing it).",
```
philippotto, line 288: @fm3 do you want to have a look at the wording? :) this is currently what @Dagobert42 and I came up with, but I know that you have a good eye for improving messages towards conciseness :)

philippotto (approved):
Awesome :tada: Works great :+1: I'm already accepting it, but let's wait for @fm3's opinion on the wording before merging.

frontend/javascripts/messages.ts
fm3: I think the message is already very good, but I notice it is very general.
I’d say that the constant_z downsampling is a very special case and could be explicitly mentioned? “For datasets that were not downsampled in z, it is recommended to view only the XY viewport.” – what do you think?

philippotto: The current code simply checks the rendered quality (ratio of #data_voxels to #rendered_pixels) and does not check whether the dataset was z1 downsampled. Therefore, the message is a bit generic. However, it would probably be alright to use your suggestion, anyway, since the quality should not drop for another reason :shrug: I don't have a strong opinion on this..

fm3: Yeah, I already guessed that this may be the reason. Let’s stick with the original message then, not risking false positives there :)

Dagobert42: "The currently rendered quality of the XY viewport is not optimal as the dataset is not downsampled along the Z-axis. To improve the quality try to increase the size of the XY viewport (e.g. by maximizing it)."

PR #6269, opened 2022-06-10
Bucket loading meter

MichaelBuessemeyer:
This PR adds the current average download speed of buckets to the statusbar.

- https://bucketloadingmeter.webknossos.xyz/

### Steps to test:
- Open a dataset in view mode or any tracing.
- While the buckets are being loaded the download indicator in the statusbar should show the current download speed. Note that the past downloaded buckets of up to 5 seconds are considered for the calculation. This means that up to five seconds long after nothing is downloaded, the status indication still shows that buckets are being downloaded.
- Hover the download indication in the status bar. A tooltip should show the accumulated downloaded data amount in the form of buckets. To test this do the following:
  - Wait until no buckets are being downloaded anymore. The download speed should be 0.
  - Open the dev tools, open the network tab and start recording. Write down the currently accumulated amount of downloaded bucket data from the tooltip.
  - Reload a visible layer in the layer settings by clicking on the reload icon. The network tab should now list the requests of the layer data being prefetched. 
  - Once the download speed is at 0 again the reloading of the layer should be complete. Below the list, in the network tab, there is the data amount of resources. Calculate the difference between the written down value prior to the layer reloading and what is now shown in the download speed tooltip as the total amount. The difference should match about the same amount as what is given by the network tab.

- I do not know how to test, whether the shown download speed is actually accurate 🤔 ....

### Issues:
- No issue exists for this pr

------
- [x] Ready for review


hotzenklotz:
I wonder if we should move the static information - download speed and magnifications - all the way to the right to prevent then from moving around so much, when position/segment id update. What you you think?

image

philippotto:
> I wonder if we should move the static information - download speed and magnifications - all the way to the right to prevent then from moving around so much, when position/segment id update. What you you think?
> 
> image

Good idea :+1: Strictly speaking, all of the infos are dynamic (i.e., they can change), but some change faster than others. I propose the following order:

- Segment Id (fast changes due to mouse movement)
- Position (fast changes due to mouse movement)
- Download Speed (can change without the user doing anything)
- the other infos should only change due to user actions and they won't change as rapidly. I don't have a strong opinion here, but semantic grouping would be nice (e.g., skeleton-related, volume-related, general...)

hotzenklotz:
>I propose the following order:
> 
> * Segment Id (fast changes due to mouse movement)
> * Position (fast changes due to mouse movement)
> * Download Speed (can change without the user doing anything)
> * the other infos should only change due to user actions and they won't change as rapidly. I don't have a strong opinion here, but semantic grouping would be nice (e.g., skeleton-related, volume-related, general...)

Sounds good to me.

frontend/javascripts/libs/window.ts
```diff
@@ -72,6 +74,7 @@ const _window =
         addEventListener,
         removeEventListener,
         open: (_url: string) => {},
+        performance: { now: () => ++performanceCounterForMocking },
```
MichaelBuessemeyer, line 77: `performance.now()` needs to be mocked for the tests. By increasing a counter I want to simulate "time moving on" and therefore avoid that a calculation like `roundTripTime = endTime - startTime` results in a value of 0. A value of 0 might lead to a `division by 0` error and that's what I want to avoid.

frontend/javascripts/oxalis/model/data_connection_info.ts
```diff
@@ -1,31 +1,115 @@
...
+  }
+
+  filterOutOldEntries() {
+    const currentTime = window.performance.now();
+    this.data = this.data.filter((entry) => currentTime - entry.endTime < 5000);
+  }
+
+  log(startTime: number, endTime: number, loadedBytes: number): void {
+    // TODO: stalling of requests is included in the round trip time :/. try not to measure this. maybe overkill.
+    // TODO: consider saving all failed buckets for ever and not only those within the last 5 secs.
```
philippotto, line 72: I think, that idea can also be discarded for now.

frontend/javascripts/oxalis/view/statusbar.tsx
```diff
@@ -314,21 +328,22 @@ function Infos() {
...
           Pos [{globalMousePosition ? getPosString(globalMousePosition) : "-,-,-"}]
         </span>
       ) : null}
-      {isPlaneMode && hasVisibleSegmentation ? getCellInfo(globalMousePosition) : null}
-
+      <span className="info-element">
+        <Tooltip
+          title={`Downloaded ${formatCountToDataAmountUnit(
+            totalDownloadedByteCount,
+          )} Bytes of Shard Data`}
```
philippotto, line 341:
```suggestion
          )} Bytes of Image Data (after decompression)`}
```

frontend/javascripts/oxalis/model/data_connection_info.ts
```diff
@@ -1,31 +1,115 @@
...
+
+  onStatisticUpdates(notifyFunction: CallableStatsUpdate): () => void {
+    this.subscriber.push(notifyFunction);
+    const unsubscribe = () => {
+      this.subscriber = this.subscriber.filter(
+        (currentFunc: CallableStatsUpdate) => currentFunc !== notifyFunction,
+      );
+    };
+    return unsubscribe;
+  }
```
philippotto, line 63: I'd remove this listener logic here and instead do that on the caller side (i.e., the status bar should do this). In the react component, you can simply use `useInterval` to poll the newest stats. That should be less code and also has the advantage that the polling stops, when the component is unmounted (I know, you are currently unsubscribing but the interval itself is never cleared).

philippotto (changes requested):
Good stuff! I left a couple of comments, but overall it already looks pretty good :)

CHANGELOG.unreleased.md
```diff
@@ -8,17 +8,21 @@ and this project adheres to [Calendar Versioning](http://calver.org/) `0Y.0M.MIC
 For upgrade instructions, please check the [migration guide](MIGRATIONS.released.md).
 
 ## Unreleased
+
 [Commits](https://github.com/scalableminds/webknossos/compare/22.06.0...HEAD)
 
 ### Added
-Added a warning for when the resolution in the XY viewport on z=1-downsampled datasets becomes too low, explaining the problem and how to mitigate it. [#6255](https://github.com/scalableminds/webknossos/pull/6255)
+
+- Added a shard download speed indicator to the statusbar. On hover a tooltip is shown that show the total amount of downloaded shard data. [#6269](https://github.com/scalableminds/webknossos/pull/6269)
```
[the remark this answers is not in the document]
MichaelBuessemeyer: >ah, it's probably meant as a replacement for bucket?

Jep 🎉. I thought that `shard` is the new name according to the new terminology.

> i'd simply use `image data`

okay 👯, sounds more user friendly 🍤 🚀

frontend/javascripts/libs/format_utils.ts
```diff
@@ -80,28 +98,27 @@ const nmFactorToUnit = new Map([
...
+export function formatCountToDataAmountUnit(count: number): string {
+  return formatNumberToUnit(count, byteFactorToUnit);
 }
-export function findClosestLengthUnitFactor(lengthInNm: number): number {
-  let closestFactor = sortedNmFactors[0];
 
-  for (const factor of sortedNmFactors) {
-    if (lengthInNm >= factor) {
+export function findClosestToUnitFactor(number: number, unitMap: Map<number, string>): number {
+  const sortedFactors = Array.from(unitMap.keys()).sort((a, b) => a - b);
```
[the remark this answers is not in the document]
MichaelBuessemeyer: Yeah, that was also my thought, but I went for the "the sorting probably not significant and can be redone every time".
As you also think that memorizing the sorted array is smart, let's do that 🚀

frontend/javascripts/oxalis/model/bucket_data_handling/pullqueue.ts
```diff
@@ -26,19 +25,19 @@ class PullQueue {
...
-  connectionInfo: ConnectionInfo;
+  connectionInfo: DataConnectionInfo;
   datastoreInfo: DataStoreInfo;
   abortController: AbortController;
 
   constructor(
     cube: DataCube,
     layerName: string,
-    connectionInfo: ConnectionInfo,
+    connectionInfo: DataConnectionInfo,
```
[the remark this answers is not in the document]
MichaelBuessemeyer: You are right. The connection was present but unused in many different files. I removed it everywhere, where it was not needed.

frontend/javascripts/oxalis/model/data_connection_info.ts
philippotto: I meant the hook `useInterval` which is defined in react_helpers.ts :) I hope this just works :tm: If not, please let me know!

MichaelBuessemeyer: Sadly, the `setInterval` hook does not fix the problem. But this seems to be related to Chrome + MacOS. With firefox + MacOS the statusbar happily updates its download speed all the time without any user interaction .... :/

MichaelBuessemeyer: > Sadly, the setInterval hook does not fix the problem. But this seems to be related to Chrome + MacOS. With firefox + MacOS the statusbar happily updates its download speed all the time without any user interaction .... :/

Looks like I made some mistakes while testing with Chrome yesterday. Today the issue did not occur anymore. Therefore this pr is ready for the 2nd review round 🔔

philippotto (approved):
Awesome, looks good to me :+1:

PR #6271, opened 2022-06-10
Fix benign error toast for public annotations and other improvements

philippotto:
- Don't show error toast when fetching token or teams failed
- Fix padding of share button
- Make some modals lazy

- https://fixtoast.webknossos.xyz

### Steps to test:
- create a public sharing link for an annotation and open that in incognito tab --> no error toast should be visible
- the share button shouldn't look off, anymore

### Issues:
- fixes https://github.com/scalableminds/webknossos/issues/6262

------
- [X] Ready for review

### DIFF
--- a/frontend/javascripts/admin/admin_rest_api.ts
+++ b/frontend/javascripts/admin/admin_rest_api.ts
@@ -339,13 +339,17 @@ export function updateTaskType(taskTypeId: string, taskType: APITaskType): Promi
 
 // ### Teams
 export async function getTeams(): Promise<Array<APITeam>> {
-  const teams = await Request.receiveJSON("/api/teams");
+  const teams = await Request.receiveJSON("/api/teams", {
+    doNotInvestigate: true,
+  });
   assertResponseLimit(teams);
   return teams;
 }
 
 export async function getEditableTeams(): Promise<Array<APITeam>> {
-  const teams = await Request.receiveJSON("/api/teams?isEditable=true");
+  const teams = await Request.receiveJSON("/api/teams?isEditable=true", {
+    doNotInvestigate: true,
+  });
   assertResponseLimit(teams);
   return teams;
 }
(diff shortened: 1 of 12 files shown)


frontend/javascripts/admin/admin_rest_api.ts
```diff
@@ -339,13 +339,17 @@ export function updateTaskType(taskTypeId: string, taskType: APITaskType): Promi
 
 // ### Teams
 export async function getTeams(): Promise<Array<APITeam>> {
-  const teams = await Request.receiveJSON("/api/teams");
+  const teams = await Request.receiveJSON("/api/teams", {
+    doNotInvestigate: true,
```
daniel-wer, line 343: In `share_modal_view.tsx`, `showErrorToast: false` is used twice for this use case. However, I think your solution makes more sense, maybe you can unify the behavior for the share_model code as well :)

frontend/javascripts/libs/react_helpers.tsx
```diff
@@ -67,4 +67,22 @@ export function usePolledState(callback: (arg0: OxalisState) => void, interval:
     callback(state);
   }, interval);
 }
+
+export function makeModalLazy<T extends { isVisible: boolean }>(
```
daniel-wer, line 71: There's no reason this wouldn't work for any React element, right? Maybe rename to `makeComponentLazy`?

daniel-wer (approved):
Nice fixes, works as advertised! I made two suggestions, but nothing blocking :+1:

frontend/javascripts/libs/react_helpers.tsx
philippotto: The component needs to have a `isVisible` prop, but you are right. It's not special to modals. I'll rename it :)

frontend/javascripts/admin/admin_rest_api.ts
philippotto: To follow up with our in-person convo: I could remove the request parameters from `share_modal_view.tsx` by checking whether the user is logged in. However, the `doNotInvestigate` params in this module are quite handy for the team related selection UIs (`PermissionsAndTeamsModalView` and `TeamSelectionComponent`) which I didn't want to adapt for this hot-fix PR.

PR #6274, opened 2022-06-13
Allow version history view in annotations not owned by you

### LINKED ISSUE #6163: Allow to see version history of not-owned annotation
The "Restore Older Version" feature is currently only shown when updating the annotation is allowed. This should be changed so that one can always open the history view. However, restoring the actual version should be disabled if saving is disallowed.

MichaelBuessemeyer:
This PR enables the version history view for view only annotations / annotations where updates are not allowed. 

- https://allowversionhistoryinnotownedannotation.webknossos.xyz

### Steps to test:
- Create a new user (if not already existing) and activate that user.
- Log in as the new user and create a new annotation. Annotate around, whatever you like as long as it creates new versions of the annotation.
- Log out and log in again with the admin user. View the annotation of the newly created user. The `Restore Older Version` entry should now be available in the action bar dropdown menu. Prior to this pr, this item would not be enabled.
- Click this menu entry and play around with the version history view on the right. The elements in the version view should allow you to download the annotation of that version.

### Issues:
- fixes #6163

------
- [x] Ready for review

### DIFF
--- a/CHANGELOG.unreleased.md
+++ b/CHANGELOG.unreleased.md
@@ -12,7 +12,7 @@ For upgrade instructions, please check the [migration guide](MIGRATIONS.released
 [Commits](https://github.com/scalableminds/webknossos/compare/22.06.1...HEAD)
 
 ### Added
-
+- Added a image data download speed indicator to the statusbar. On hover a tooltip is shown that show the total amount of downloaded shard data. [#6269](https://github.com/scalableminds/webknossos/pull/6269)
 - Added a warning for when the resolution in the XY viewport on z=1-downsampled datasets becomes too low, explaining the problem and how to mitigate it. [#6255](https://github.com/scalableminds/webknossos/pull/6255)
 - Added the possibility to view and download older versions of read-only annotations. [#6274](https://github.com/scalableminds/webknossos/pull/6274)
 
@@ -23,7 +23,6 @@ For upgrade instructions, please check the [migration guide](MIGRATIONS.released
 - When creating a task from a base annotation, the starting position/rotation and bounding box as specified during task creation are now used and overwrite the ones from the original base annotation. [#6249](https://github.com/scalableminds/webknossos/pull/6249)
 
 ### Fixed
-
 - Fixed that bounding boxes were deletable in read-only tracings although the delete button was disabled. [#6273](https://github.com/scalableminds/webknossos/pull/6273)
 - Fixed that (old) sharing links with tokens did not work, because the token was removed during a redirection. [#6281](https://github.com/scalableminds/webknossos/pull/6281)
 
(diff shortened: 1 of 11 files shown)


frontend/javascripts/oxalis/view/version_entry_group.tsx
```diff
@@ -18,6 +18,39 @@ type State = {
...
+            <FormattedDate timestamp={lastTimestamp} format="HH:mm" />)
+          </React.Fragment>
+        }
+        avatar={
+          <Avatar size="small" icon={expanded ? <CaretDownOutlined /> : <CaretRightOutlined />} />
+        }
+      />
+    </List.Item>
+  );
+}
```
MichaelBuessemeyer, line 53: I moved `GroupHeader` out of `VersionEntryGroup` because of some typescript warning. While doing so I got rid of 2 ignored typescript warnings.

frontend/javascripts/oxalis/view/action-bar/tracing_actions_view.tsx
```diff
@@ -614,14 +614,12 @@ class TracingActionsView extends React.PureComponent<Props, State> {
       );
     }
 
-    if (restrictions.allowSave) {
```
daniel-wer, line 617: I was curious why this condition was introduced (since the version view seems to work just fine for read-only annotations) and checked using git blame. Turns out, I introduced this when sandbox annotations were implemented. For sandbox annotations, the version view doesn't work, it cannot be closed again, for example :see_no_evil: You can check using a sandbox URL, like https://allowversionhistoryinnotownedannotation.webknossos.xyz/datasets/sample_organization/l4_sample/sandbox/skeleton
However, I think this condition was too broad and instead should only check that that no sandbox annotation is active. You should be able to do something like this `state.temporaryConfiguration.controlMode !== ControlModeEnum.SANDBOX`

daniel-wer (changes requested):
LGTM once the sandbox case is fixed :+1:

frontend/javascripts/oxalis/view/action-bar/tracing_actions_view.tsx
MichaelBuessemeyer: > I was curious why this condition was introduced (since the version view seems to work just fine for read-only annotations) and checked using git blame. Turns out, I introduced this when sandbox annotations were implemented. For sandbox annotations, the version view doesn't work, it cannot be closed again, for example 🙈 You can check using a sandbox URL, like https://allowversionhistoryinnotownedannotation.webknossos.xyz/datasets/sample_organization/l4_sample/sandbox/skeleton


Thanks for paying attention to this 🚀

daniel-wer (approved):
LGTM :+1:

PR #6276, opened 2022-06-14
DRY type definitions of redux actions

philippotto:
It has bothered me for years that we defined redux actions and their action creators in an almost redundant manner. The PR suggests a more concise way of doing this by simply using `ReturnType` (I think, we also tried this with flow, but it didn't work for some reason).

@daniel-wer What do you think about this? I adapted only two action modules, but if you agree, I'd do the rest, too.

- https://___.webknossos.xyz

### Steps to test:
- CI should be enough

------
- [x] Ready for review


frontend/javascripts/oxalis/model/actions/user_actions.ts
```diff
@@ -1,16 +1,17 @@
...
-};
-type LogoutUser = {
-  type: "LOGOUT_USER";
-};
+
+export const setActiveUserAction = (user: APIUser) =>
+  ({
+    type: "SET_ACTIVE_USER",
+    user,
+  } as const);
```
philippotto, line 12: `as const` is necessary so that the type is not interpreted as a string, but as a specific literal.

philippotto:
(hide whitespace during review makes things a bit better)

daniel-wer (commented):
Nice, this has bothered me for a while as well!

I think this looks great. My only fear would be to forget about the `as const` leading to less strict type checking. Would this provoke an error during type checking (or could we provoke one somehow)?

philippotto:
@daniel-wer I finally got around to DRY the other action modules :) I don't think that testing is necessary, since I only shuffled the types around for which the type checker should be enough.

philippotto:
(a thorough code review is also not really realistic nor necessary in my opinion)

daniel-wer (approved):
Thanks a lot :tada:

PR #6283, opened 2022-06-16
Add download modal to dataset view actions

### LINKED ISSUE #6148: Add "Download/Connect via Python Client" Modal to Dataset View
## Detailed Description
To make the Python client and functionality more discoverable, it would be great to have the download code next to the normal download functionality in the dataset view (see #5672). The specific code to download the current dataset (or a subpart) could be shown as a text-block with a copy-button next to it. The content could be something like:
```bash
pip install webknossos
```

```python
import webknossos as wk

with wk.webknossos_context(token="MY_TOKEN"):
    dataset = wk.Dataset.download(
        dataset_name_or_url="dataset_name",
        organization_id="my_organization",
    )
```

With links to the generic documentation and the exact download methods:
https://docs.webknossos.org/api/webknossos/dataset/dataset.html#Dataset.download
https://docs.webknossos.org/webknossos-py

We might want to adapt the code, depending on if it's a public dataset (no token necessary) or if it uses a sharing token. We might also add UI to specify a subset (bbox, layers, mags).

Dagobert42:
- https://addpythonmodaltodatasetview.webknossos.xyz/

### Steps to test:
- open a dataset in view-only
- select python client from the dataset view actions drop-down

### Issues:
- fixes #6148 

------
- [x] Ready for review


frontend/javascripts/oxalis/view/action-bar/python_client_modal_view.tsx
```diff
@@ -0,0 +1,107 @@
...
+import { getAuthToken } from "admin/admin_rest_api";
+import { useSelector } from "react-redux";
+import type { OxalisState } from "oxalis/store";
+import { CopyableCodeSnippet, MoreInfoHint } from "./download_modal_view";
+const { Paragraph, Text } = Typography;
+type Props = {
+  onClose: () => void;
+};
+
+export default function PythonClientModalView(props: Props): JSX.Element {
```
philippotto, line 15: In a recent commit, I changed the modals to become lazy so that they are only rendered after they were opened at least once. You can have a look at how this can be done by searching for `const DownloadModalView = makeComponentLazy(_DownloadModalView);` and adopting that.

philippotto: ah, yeah, good point :+1: apparently, I missed the dataset-view-mode related modals in my PR. thank you!

jstriebel:
We might also consider to propose[`open_remote`](https://docs.webknossos.org/api/webknossos/dataset/dataset.html#Dataset.open_remote) instead of download.

frontend/javascripts/oxalis/view/action-bar/view_dataset_actions_view.tsx
```diff
@@ -30,6 +48,13 @@ export default function ViewDatasetActionsView(props: Props) {
         Share
       </Menu.Item>
       {screenshotMenuItem}
+      <Menu.Item
+        key="python-client-button"
+        onClick={() => Store.dispatch(setPythonClientModalVisibilityAction(true))}
+      >
+        <CodeOutlined />
+        Python Client
```
philippotto, line 56:
```suggestion
        Download
```

philippotto (approved):
Great :+1:

PR #6284, opened 2022-06-16
Make highlight color darker in light theme

### LINKED ISSUE #5998: Highlighted bounding box is hard to see in 3D viewport + light theme
When using the bounding box tool to resize an existing bounding box, the lines are highlighted in white. When using the light theme, the background in the 3D viewport is white, too. This makes it very hard to see the current extent of the box. An easy fix for now would be to set another highlighting color for the 3D viewport when the light theme is used.

Dagobert42:
### Steps to test:
- check if Bboxes in 3D viewport are visible enough now

### Issues:
- fixes #5998 

------
- [ ] Ready for review

### DIFF
--- a/frontend/javascripts/oxalis/geometries/cube.ts
+++ b/frontend/javascripts/oxalis/geometries/cube.ts
@@ -62,7 +62,7 @@ class Cube {
   getLineMaterial() {
     return this.isHighlighted
       ? new THREE.LineBasicMaterial({
-          color: 0xffffff,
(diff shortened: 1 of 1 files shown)


Dagobert42:
The darker we go here the more visible the Bbox becomes on the white background but the more it seems to merge with the grey areas in the data. We could invert the color of the Bbox to use as highlight color but I think that looks a little too disco...

philippotto (approved):
Great! I think, we can simply go with this solution for now :)

PR #6287, opened 2022-06-20
2nd attempt: Merge "Shared Annotations" with "My annotations"

philippotto:
Second attempt to merge https://github.com/scalableminds/webknossos/pull/6230. However, there seems to be a bug with the `annotations/readable` route which fails sometimes with a 404 status.

https://webknossos.org/api/annotations/readable?isFinished=false&pageNumber=0 failed for me in prod (isFinished=true worked).


fm3:
@jstriebel could you have a look here at the latest commits? after broadening the access context, which solved the 404 error, the second bug was that annotations that were public were also listed on this page. however, the semantics of “public” is really “unlisted” (since such a list did not exist before and we cannot just change that semantics as it was communicated like this to the users). so now I added a separate list access query. i also added a slight performance improvement for fetching shared teams from the db

jstriebel (approved):
LGTM :+1:

Issue #6251, opened 2022-06-01
Make annotation layer names unique per annotation


fm3:
Users need a way to uniquely identify annotation layers. The current "id" field in the NML is fragile as it relies on the ordering as returned by the database.
Also, it would be more user-friendly to use the name. Dataset layer names are already enforced to be unique.

An evolution needs to
- name NULL-named layers ('Skeleton' or 'Volume' to match front-end default)
- resolve conflicts for annotations with multiple layers of the same name
- add the NOT-NULL constraint
- add the UNIQUE constraint

The frontend needs to stop the user giving a layer the same name as another layer

fm3:
Query to find annotations with multiple volume layers

```
select _annotation, name from webknossos.annotation_layers where _annotation in (select s._annotation from 
(select _annotation, count(_annotation) from webknossos.annotation_layers where typ = 'Volume' group by _annotation order by count(_annotation) desc limit 1000) as s
where count > 1) and typ = 'Volume' order by _annotation
```

fm3:
note that we should also in some form restrict the layer names themselves, i.e. forbid forward slashes, to avoid path ambiguity

fm3:
Reopening this as the front-end pr #6289 will not fix this in the db for existing annotations.

fm3:
```
-- change null names to the old front-end-side defaults

update webknossos.annotation_layers set name = 'Volume' where name is null and typ = 'Volume'
update webknossos.annotation_layers set name = 'Skeletons' where name is null and typ = 'Skeleton'


-- find annotations with multiple layers

select _annotation, name from webknossos.annotation_layers where _annotation in (select s._annotation from 
(select _annotation, count(_annotation) from webknossos.annotation_layers where typ = 'Volume' group by _annotation order by count(_annotation) desc limit 1000) as s
where count > 1) and typ = 'Volume' order by _annotation


-- find layers with interesting names

select * from webknossos.annotation_layers where name is not null and name not in ('Volume', 'Volume1', 'Mitochondria','Volume2', 'Segmentation', 'segmentation', 'Skeleton')
```

PR #6289, opened 2022-06-20
Ensure unique volume layer names when changing names in frontend

MichaelBuessemeyer:
This PR makes it impossible for the user to set layer names of volume annotation layers that already exist. The reason is to prevent duplicate names. The prevention is done when changing a name of a layer and when adding a new volume annotation layer. This pr does not change the names of existing volume layers but it marks them red if there are duplicates and asks the user to rename the layer with duplicated names.

- https://uniqueannotationlayernames.webknossos.xyz

### Steps to test:
I suggest testing this locally as the first step required is not (easily) possible on a dev branch.
- Create a volume annotation with multiple layers. Give the layers the same name. Add another layer that includes a `/` in its name.
- Now checkout the new branch, there layer names in the layer settings should now be red and a tooltip should say that these layers should be renamed to avoid duplicated names. The layer with the `/` should say that this character is not allowed and it should also be marked red.
- Try renaming the layers. This should work. Trying to enter a name that already exists should not be possible. Instead, an error toast should pop up. Entering a name that contains a `/` should also result in an error toast.
- Rename the layer so no names are duplicated and no name contains a `/`. This should then work.
- Try adding a new annotation layer. The suggested name should not match any of the existing names. Entering an existing name or one containing a `/` and then trying to confirm should result in an error toast.

### Issues:
- fixes [#6251](https://github.com/scalableminds/webknossos/issues/6251)

------
- [x] Ready for review


frontend/javascripts/oxalis/model/accessors/volumetracing_accessor.ts
```diff
@@ -94,6 +97,56 @@ export function getReadableNameByVolumeTracingId(
...
+  const hasDuplicatedName = checkForLayerNameDuplication(
+    readableLayerName,
+    dataset,
+    tracing,
+    dontCountGivenName ? readableLayerName : null,
+  );
+  const nameContainsSlash = checkForInvalidCharacters(readableLayerName);
+  return hasDuplicatedName === true ? nameContainsSlash : hasDuplicatedName;
+}
+
```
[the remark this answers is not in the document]
daniel-wer: You could put the `check...` and `validate...` methods into the `add_volume_layer_modal.tsx`.

frontend/javascripts/oxalis/view/components/editable_text_label.tsx
```diff
@@ -8,11 +8,12 @@ import Toast from "libs/toast";
 type Rule = {
   message?: string;
   type?: string;
+  validator?: (arg0: number | string) => boolean;
```
daniel-wer, line 11: Why is this `number | string`? As far as I can see, validator is only called with `this.state.value` which is `string`.

frontend/javascripts/oxalis/view/left-border-tabs/layer_settings_tab.tsx
```diff
@@ -357,17 +365,46 @@ class DatasetSettings extends React.PureComponent<DatasetSettingsProps, State> {
...
+                  onChange={(newName) => {
+                    this.props.onEditAnnotationLayer(volumeDescriptor.tracingId, {
+                      name: newName,
+                    });
+                  }}
+                  rules={[
+                    {
+                      message: messages["tracing.volume_layer_name_duplication"],
+                      validator: (newLayerName) =>
+                        typeof newLayerName === "string"
```
daniel-wer, line 387: When is `newLayerName` not a string and if so what is it instead?

frontend/javascripts/oxalis/view/components/editable_text_label.tsx
```diff
@@ -21,6 +22,9 @@ export type EditableTextLabelProp = {
   disableEditing?: boolean;
   onContextMenu?: () => void;
   width?: string | number;
+  updateOnAllChanges?: boolean;
```
daniel-wer, line 25: This appears to be unused

MichaelBuessemeyer (commented):
Hi Daniel,
thanks for the review. I applied all your feedback and adjusted the validation of invalid characters so that URI encoding does not change them.
I had two ideas on how to approach this. The first is the "brute force" method: I remove invalid chars from the new layer name until the URI encoded version matches the original version. While doing so I track the invalid chars that are then shown to the user as invalid. This method should be 100% correct.
The second method uses a regex that removes all save uri characters that will not be encoded in uris according to (this post)[https://perishablepress.com/stop-using-unsafe-characters-in-urls/#character-encoding-chart]. The left over chars are invalid ones. 
The second version is shorter and easier to understand, while I am not 100% sure whether the uri safe chars from the post are correct. But I would personally prefer the second version. 

You can compare both versions in commit [e281c4a](https://github.com/scalableminds/webknossos/pull/6289/commits/e281c4aea515e3a911479e214c3d96d601d44189)

frontend/javascripts/oxalis/view/left-border-tabs/modals/add_volume_layer_modal.tsx
```diff
@@ -17,11 +17,67 @@ import {
...
+
+export function validateReadableLayerName(
+  readableLayerName: string,
+  allReadableLayerNames: string[],
+  nameNotToCount?: string,
+): ValidationResult {
+  if (nameNotToCount) {
+    // If the given nameNotToCount is already once within the allReadableLayerNames. But as this name is now being renamed and having
+    // the same name as the previous name given by nameNotToCount should be still allowed, this name is removed from the array once.
+    // Nevertheless, additional duplicated of nameNotToCount are counted as a duplication.
```
daniel-wer, line 63: There seem to be a couple of missing words. Could you re-check this paragraph? :)

frontend/javascripts/oxalis/view/left-border-tabs/modals/add_volume_layer_modal.tsx
```diff
@@ -17,11 +17,67 @@ import {
...
+): ValidationResult {
+  const layerNameDoesNotExist = _.countBy(allReadableLayerNames)[readableLayerName] <= 0;
+  return {
+    isValid: layerNameDoesNotExist,
+    message: layerNameDoesNotExist ? "" : messages["tracing.volume_layer_name_duplication"],
+  };
+}
+
+export function checkLayerNameForInvalidCharacters(readableLayerName: string): ValidationResult {
+  const uriSaveCharactersRegex = /[0-9a-zA-Z-._~]+/g;
```
daniel-wer, line 39: I also like this version much more than the brute-force one. As you said, it's easier to understand and more elegant. I'm certain that the list of safe characters is correct, see [this post](https://stackoverflow.com/a/695469) which references the corresponding RFC section.

frontend/javascripts/oxalis/view/left-border-tabs/modals/add_volume_layer_modal.tsx
```diff
@@ -17,11 +17,67 @@ import {
...
 import InputComponent from "oxalis/view/components/input_component";
 import api from "oxalis/api/internal_api";
 import Toast from "libs/toast";
+
+export type ValidationResult = { isValid: boolean; message: string };
+export function checkForLayerNameDuplication(
+  readableLayerName: string,
+  allReadableLayerNames: string[],
+): ValidationResult {
+  const layerNameDoesNotExist = _.countBy(allReadableLayerNames)[readableLayerName] <= 0;
```
daniel-wer, line 31: For layer names that do not exist, `_.countBy(allReadableLayerNames)[readableLayerName]` will be `undefined`, resulting in `undefined <= 0` which is `false`. So I think in its current state this method will never return that the layer name is valid.

Something like
```suggestion
  const layerNameDoesNotExist = !allReadableLayerNames.includes(readableLayerName);
```

should do the job.

frontend/javascripts/oxalis/view/components/editable_text_label.tsx
```diff
@@ -91,18 +90,22 @@ class EditableTextLabel extends React.PureComponent<EditableTextLabelProp, State
...
+        const isValid = re.test(this.state.value);
+        if (!isValid) {
+          Toast.error(rule.message);
+          return false;
+        }
       } else if (rule.validator != null) {
-        isValid = rule.validator(this.state.value);
+        const validationResult = rule.validator(this.state.value);
+        if (!validationResult.isValid) {
+          Toast.error(rule.message);
```
daniel-wer, line 104:
```suggestion
          Toast.error(validationResult.message);
```

frontend/javascripts/oxalis/view/left-border-tabs/modals/add_volume_layer_modal.tsx
```diff
@@ -17,11 +17,67 @@ import {
...
+    isValid: layerNameDoesNotExist,
+    message: layerNameDoesNotExist ? "" : messages["tracing.volume_layer_name_duplication"],
+  };
+}
+
+export function checkLayerNameForInvalidCharacters(readableLayerName: string): ValidationResult {
+  const uriSaveCharactersRegex = /[0-9a-zA-Z-._~]+/g;
+  // Removing all URISaveCharacters from readableLayerName. The left over chars are all invalid.
+  const allInvalidChars = readableLayerName.replace(uriSaveCharactersRegex, "");
+  const allUniqueInvalidCharsAsSet = new Set(...allInvalidChars);
```
daniel-wer, line 42:
```suggestion
  const allUniqueInvalidCharsAsSet = new Set(allInvalidChars);
```

otherwise only the first character will end up in the Set, compare the [MDN page](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set/Set).

daniel-wer (changes requested):
Thank you for applying the feedback, the code already looks better and is easier to understand, agreed! :+1: There are a couple of functional issues left, please see my comments.

frontend/javascripts/oxalis/view/left-border-tabs/layer_settings_tab.tsx
```diff
@@ -357,17 +365,46 @@ class DatasetSettings extends React.PureComponent<DatasetSettingsProps, State> {
...
+                              newLayerName,
+                              dataset,
+                              tracing,
+                              readableName,
+                            ) === true
+                          : false,
+                    },
+                    {
+                      message: messages["tracing.volume_layer_name_includes_slash"],
+                      validator: (newLayerName) =>
```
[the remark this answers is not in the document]
daniel-wer: My idea was to do the filtering at the caller site, so before `validateReadableLayerName` is called. This way, the `nameNotToCount` argument wouldn't be needed and the function interface would be smaller and easier to understand. However, I see that this would need to be done at two spots, which is not ideal - so it's also fine as is in my opinion.

frontend/javascripts/oxalis/view/left-border-tabs/modals/add_volume_layer_modal.tsx
MichaelBuessemeyer: Oh, you are right... I thought I did a quick tryout of this in the browser console. 🙈

MichaelBuessemeyer (commented):
@daniel-wer Thanks a lot for the quick review yesterday. And thanks for already finding the bugs in the code and providing a solution 🙏. 
Sorry, for not testing before pinging you. I did not have the time for it.

frontend/javascripts/oxalis/view/left-border-tabs/modals/add_volume_layer_modal.tsx
```diff
@@ -14,9 +14,72 @@ import {
...
+  };
+}
+
+export function validateReadableLayerName(
+  readableLayerName: string,
+  allReadableLayerNames: string[],
+  nameNotToCount?: string,
+): ValidationResult {
+  if (nameNotToCount) {
+    // If the given nameNotToCount is already once within the allReadableLayerNames we need to remove this name once from the allReadableLayerNames.
```
daniel-wer, line 61:
```suggestion
    // nameNotToCount needs to be removed once if it is included in allReadableLayerNames.
```

Feel free to ignore, but sounds less clunky to me.

frontend/javascripts/oxalis/view/left-border-tabs/modals/add_volume_layer_modal.tsx
```diff
@@ -14,9 +14,72 @@ import {
...
+    // This is needed in case of saving an existing volume layer's name when the name was not modified.
+    // In this scenario nameNotToCount should be the previous name the volume layer which will then be remove once from the allReadableLayerNames.
```
daniel-wer, line 63:
```suggestion
    // In this scenario nameNotToCount should be the previous name of the volume layer which will then be removed once from the allReadableLayerNames.
```

daniel-wer (approved):
Great fixes, everything works as advertised :+1: 

> Sorry, for not testing before pinging you. I did not have the time for it.

No worries :)

PR #6294, opened 2022-06-22
fix error when loading agglomerate skeleton for single-segment agglomerate

fm3:
In newer hdf handling code we already added zero-length checks, but this old code did not have them. So when loading single-segment agglomerate skeletons, the edges array was empty, which cannot be passed to jhdf5 as is.

- [x] Ready for review


daniel-wer:
I wanted to test this, but I'm not able to reproduce the error locally on master. I'm in a hybrid annotation without an editable mapping, rightclick - "Import Agglomerate Skeleton" for an agglomerate skeleton with a single node. Loads just fine :thinking:

fm3:
you are right :thinking: That ay mean that this problem does not happen for multi-dimensional arrays like edges. I wonder what the original bug report was about, then :monocle_face:

PR #6304, opened 2022-06-23
Fix zooming out for datasets with very large scale

### LINKED ISSUE #6061: Datasets with a large scale might not allow zooming to coarsest mags
## Problem
Given a dataset with a large scale / voxel size (e.g.,  `5860 × 5860 × 1`), might not allow zooming out to all mags.

The maximum zoom value is capped to something like 1.1**100 (== 13780.6) by the following line:

https://github.com/scalableminds/webknossos/blob/6c6a23b4668ba5cfa5c34f05581fc8a5c871ed31/frontend/javascripts/oxalis/model/accessors/flycam_accessor.js#L135-L135

## Your Environment for bug

philippotto:
Respect the dataset scale when calculating the maximum iteration limit for the max zoom value determination.

- https://___.webknossos.xyz

### Steps to test:
- find a dataset which goes until mag 1024
- set a dataset scale to 600, 600, 1
- open the dataset and maximize the XY viewport
- zoom out (without this PR, zooming out was not possible until mag 1024)

### Issues:
- fixes #6061

------
- [X] Ready for review

### DIFF
--- a/frontend/javascripts/oxalis/model/accessors/flycam_accessor.ts
+++ b/frontend/javascripts/oxalis/model/accessors/flycam_accessor.ts
@@ -143,12 +143,14 @@ export function _getMaximumZoomForAllResolutions(
   // From that, we calculate the theoretical maximum zoom value. The dataset scale is taken into account,
   // because the entire scene is scaled with that.
   const maxSupportedZoomValue = 2 ** MAX_SUPPORTED_MAGNIFICATION_COUNT * Math.max(...datasetScale);
-  const maximumIterationCount = Math.log(maxSupportedZoomValue) / Math.log(ZOOM_STEP_INTERVAL);
-
   // Since the viewports can be quite large, it can happen that even a zoom value of 1 is not feasible.
   // That's why we start the search with a smaller value than 1. We use the ZOOM_STEP_INTERVAL factor
   // to ensure that the calculated thresholds correspond to the normal zoom behavior.
-  let currentMaxZoomValue = 1 / ZOOM_STEP_INTERVAL ** 20;
+  const ZOOM_IN_START_EXPONENT = 20;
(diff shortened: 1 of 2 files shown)


frontend/javascripts/oxalis/model/accessors/flycam_accessor.ts
```diff
@@ -128,27 +128,36 @@ export function _getMaximumZoomForAllResolutions(
...
+  // However, to avoid an infinite loop in case of violated assumptions (e.g., because the calculated
+  // bucket size isn't strictly increasing anymore), we calculate an iteration limit.
+
+  // For that limit, we specify how many magnifications we want to support at least.
+  // 15 magnifications means that the highest mag would be something like [32768, 32768, 1024].
+  const MAX_SUPPORTED_MAGNIFICATION_COUNT = 15;
+  // From that, we calculate the theoretical maximum zoom value. The dataset scale is taken into account,
+  // because the entire scene is scaled with that.
+  const maxSupportedZoomValue = 2 ** MAX_SUPPORTED_MAGNIFICATION_COUNT * Math.max(...datasetScale);
+  const maximumIterationCount = Math.log(maxSupportedZoomValue) / Math.log(ZOOM_STEP_INTERVAL);
```
daniel-wer, line 146: I think this would be the correct iteration count if the start zoom value were 1. However, it is `1 / ZOOM_STEP_INTERVAL ** 20` so 20 would need to be added.

daniel-wer (commented):
Nice fix, but I'm not sure whether the maximum iteration count is computed correctly, see my comment :)

frontend/javascripts/oxalis/model/accessors/flycam_accessor.ts
philippotto: Very good point :+1: I adapted the code.

daniel-wer (approved):
Nice :+1:

Issue #6193, opened 2022-05-04
mapping and other parameters for zarr streaming


fm3:
Currently, applying back-end side mappings is only possible for the wk data request. The other two ways of requesting raw data (GET raw and zarr) should also support this.

fm3:
It is unclear how the zarr URI schema should be adapted for this. Adding discussion label.

normanrz:
> It is unclear how the zarr URI schema should be adapted for this. Adding discussion label.

I was thinking of using query params for that. However, I discussed that in the zarr gitter. The strong recommendation was to use separate URLs (i.e. datasets) for that. 

I don't know how that could fit our URL scheme. A few ideas:
```
GET /data/zarr/scalable_minds/l4_sample/segmentation/with_mapping_agglomerate_60
GET /data/zarr/scalable_minds/l4_sample/segmentation__mapping=agglomerate_60
GET /data/zarr/scalable_minds/l4_sample/segmentation:mapping=agglomerate_60
GET /data/zarr/scalable_minds/l4_sample/segmentation:mapping=agglomerate_60&only_segment=12312
```

fm3:
Using special characters like : and = means they can not then be used in a normal dataset name. Is there a reason not to use / (which is already a special character)?

`GET /data/zarr/scalable_minds/l4_sample/segmentation/withMapping/agglomerate_60/2-2-1/.zarray`

normanrz:
On a similar note, I think we also need to think about a way to input the token as part of the path. This will allow it to view private datasets and annotations in other tools. I haven't seen one that would implement an authentication UI to input anything.

leowe:
@normanrz On what does the decision how `zarr` handles the parameters depend? Is this still an open question in the spec?

normanrz:
The decision depends on us sitting down and thinking it through properly :-) I think the implementation needs to wait until then.

normanrz:
Summarizing today's discussion:
- Focus on making private (token-protected) annotations available through zarr. Private datasets can easily be shared, by creating an empty annotation. The mapping can also be set in the annotation.
- Add new database-backed links ("access links") for providing authenticated access to these resources. The database items contain annotation id, expiration date, and maybe additional fields.
- Access to these annotations will be available through the datastores, because they will likely contain the heaviest data (e.g. color data, full segmentation)
- Routes could be in the form of `https://datastore/data/zarr_private/:accessId/:dataLayerName/...`.
[...]

PR #6311, opened 2022-06-29
add mapping name param for rawcuboid

leowe:
### Steps to test:
- request an URL like `GET http://localhost:9000/data/datasets/sample_organization/test-agglomerate-file/layers/segmentation/data?token=secretSampleUserToken&x=64&y=128&z=32&width=32&height=32&depth=32&mag=1-1-1&mappingName=agglomerate_view_70` and compare to response without mappingName

### Issues:
- contributes to #6193 

------
- [x] Ready for review


fm3 (approved):
LGTM :)

leowe:
@fm3 do I need to do any of these: 
Updated [documentation](https://github.com/scalableminds/webknossos/blob/master/docs) if applicable
  Adapted [wk-connect](https://github.com/scalableminds/webknossos-connect) if datastore API changes
  Adapted [wk-libs python client](https://github.com/scalableminds/webknossos-libs/tree/master/webknossos/webknossos/client) if relevant API parts change ?

fm3:
No I don’t think so. this request is not in the documentation, and not used nor served by wkconnect. the python client can now use the new paramter, but since it is optional and backwards compatible, there is nothing blocking this PR :)

PR #6312, opened 2022-06-30
X-Auth Tokens everywhere (in datastore & tracingstore)

### LINKED ISSUE #6192: All datastore and tracingstore routes should support X-Auth-Token
Currently, the routes of tracingstore and datastore require the token as GET parameter. They should also read it from the header if present. This was already implemented for the zarr routes in #6144

leowe:
- add urlTokenOrHeader to all relevant endpoints by adding `urlOrHeaderToken`

[https://xautheverywhere.webknossos.xyz](https://xautheverywhere.webknossos.xyz/)

### Steps to test:
- use token in header instead of as parameter in any datastore or tracingstore route

### Issues:
- fixes #6192

------
- [x] Ready for review


webknossos-datastore/app/com/scalableminds/webknossos/datastore/controllers/DataSourceController.scala
```diff
@@ -102,7 +103,9 @@ Expects:
                            paramType = "body")))
   def reserveUpload(token: String): Action[ReserveUploadInformation] =
     Action.async(validateJson[ReserveUploadInformation]) { implicit request =>
-      accessTokenService.validateAccess(UserAccessRequest.administrateDataSources, Some(token)) {
+      // TODO token is not optional and urlOrHeaderToken always takes token from URL if provided
+      accessTokenService.validateAccess(UserAccessRequest.administrateDataSources,
+                                        urlOrHeaderToken(Some(token), request)) {
```
leowe, line 108: @fm3 There are a few endpoints like this one, where the token is a non-optional query parameter. What do you think is the best way to handle these? The method `urlOrHeaderToken` is currently implemented in a way that it looks only in the header when the query parameter is not set.

The way I see it, there are four options:
- We leave it the way it is now, the method does nothing at the moment in this case, but if we would want to change the behaviour, it is easier now, since it's more centralised
- We don't use the method here to make it clear that the token is needed as part of the query
- We require only that a token is provided as part of the URL or in the header and make an explicit check here
- We don't require that a token is provided as with all other endpoints and let it fail in the access validation when none is given

I would be in favour of 1 or 4. What do you think?

fm3: Well spotted! I’d say option 4 makes the most sense, then these routes are no longer special cases for this.

fm3 (approved):
Good stuff, thank you!

Issue #5074, opened 2021-01-21
Remove Save Button from some Modals?


fm3:
We should go through modals like the annotation sharing modal and see which of them should auto-save on every click, and which need an explicit save button

normanrz:
Auto-save:
- [ ] Sharing modal in annotation

hotzenklotz:
> ein problem ist, dass man bei dem share-modal auf “save” klicken muss. das hab ich schon ein paar mal vergessen

PR #6314, opened 2022-07-05
Auto save changes in sharing modal

MichaelBuessemeyer:
This PR automatically saves settings / content of a modal. Therefore it is no longer required to explicitly hit "Save". As already stated in [#5074](https://github.com/scalableminds/webknossos/issues/5074), this makes sense for the sharing modal. Another modal is not mentioned, and I could not find another one where it makes sense to me to save the current input automatically.

The original issue #5074 mentioned checking all modals to whether auto-saving is useful in that case. On a rough look, I could not find a modal that would benefit from auto-saving (search the codebase for all files having modal in their name).
If you can think of another modal, just tell me and I'll add it here.

- https://___.webknossos.xyz

### Steps to test:
- Open an explorative tracing
- Open the sharing modal via the drop-down menu in the navbar
- Update the sharing settings as you want (switch from private to public and set some teams and more)
- The changes in the modal should automatically be send to the backend and therefore automatically updated.

### Issues:
- fixes [#5074](https://github.com/scalableminds/webknossos/issues/5074)

------
- [x] Ready for review


philippotto (changes requested):
The code looks good, but I noticed a couple of things during testing:

- [x] When the annotation is set to "internal", opening the modal will also immediately show a toast (only when opening the modal for the first time, as the modal is rendered lazily and won't be destroyed when closing it).
- [x] There can only be one instance of the success toast. When switching fast between the settings, no new toast will appear (--> therefore, one could wonder if it was really saved). I suggest to decrease the time for how long the toast will be there and ensure somehow that the toast can be there multiple times (maybe assign a random key to the toast?).
- [x] Switching to "Private" won't show a toast (I think the toast comes from updating the teams, but this isn't done when setting to "Private").
- [ ] There is a race condition when changing the visibility very fast. I somehow managed to get an error that the teams cannot be updated. Probably, the frontend thought that the annotation is "public" and tried to update the teams, but the back-end still had the annotation set to "private".

MichaelBuessemeyer:
HI @philippotto,

I am currently working on your feedback. 

>  There is a race condition when changing the visibility very fast. I somehow managed to get an error that the teams cannot be updated. Probably, the frontend thought that the annotation is "public" and tried to update the teams, but the back-end still had the annotation set to "private".

What would you suggest to circumvent this race condition? freeze / disable the modal until the changes are applied in the backend? I think this would only be necessary for the visibility setting and not for the team setting, correct? Changing selected teams should not lead to this race condition. Therefore I would suggest disabling the visibility setting while a backend update call is still in process 🤔.

What do you think?

philippotto:
> What would you suggest to circumvent this race condition? freeze / disable the modal until the changes are applied in the backend?

Yes, I think that's fair.

> I think this would only be necessary for the visibility setting and not for the team setting, correct? 

Could very well be, but I think it's easier to just always block the view to avoid any problems if these dependencies change in the future. In most cases, the blocking shouldn't even be noticeable to the user.

MichaelBuessemeyer:
> Could very well be, but I think it's easier to just always block the view to avoid any problems if these dependencies change in the future. In most cases, the blocking shouldn't even be noticeable to the user.

It is only slightly noticeable 👍.

frontend/javascripts/oxalis/view/action-bar/share_modal_view.tsx
```diff
@@ -153,29 +156,49 @@ function _ShareModalView(props: Props) {
...
+    reportSuccessfulChange(newVisibility);
+  };
 
+  const handleSharedTeamsChange = async (value: APITeam | APITeam[]) => {
+    const newTeams = _.flatten([value]);
+    if (_.isEqual(newTeams, sharedTeams) || !hasUpdatePermissions) {
+      return;
+    }
+    setIsChangingInProgess(true);
+    setSharedTeams(_.flatten([value]));
```
philippotto, line 192:
```suggestion
    setSharedTeams(newTeams);
```

frontend/javascripts/oxalis/view/action-bar/share_modal_view.tsx
```diff
@@ -153,29 +156,49 @@ function _ShareModalView(props: Props) {
     fetchAndSetSharedTeams();
   }, [annotationType, annotationId, activeUser]);
 
-  const handleCheckboxChange = (event: RadioChangeEvent) => {
-    setVisibility(event.target.value as any as APIAnnotationVisibility);
+  const reportSuccessfulChange = (newVisibility: APIAnnotationVisibility) => {
+    const randomKeyToAllowDuplicates = Math.random().toString(36).substring(0, 5);
```
philippotto, line 160: I think you can simply remove the `key` property to get the behavior you want.

philippotto (changes requested):
awesome, works great :+1: however, the share modal changed on the master branch (a new boolean setting was added), which needs to be incorporated now :)

frontend/javascripts/oxalis/view/action-bar/share_modal_view.tsx
MichaelBuessemeyer: The key is required as per default the content string of the toast is chosen as a key. Therefore there are no duplicates of the Toast without a manual key. Note that I reduced the timeout time of the toast to 3,5 seconds as 4 secs felt too long for me. Feel free to argue with that :D

philippotto: > The key is required as per default the content string of the toast is chosen as a key.

Are you sure about this? Maybe it changed in the latest antd versions, but the current documentation + example seems to allow duplicate messages if no key is provided. The very first demo here suggests this at least: https://ant.design/components/notification/#header
However, if it doesn't work without the key, the code can obviously stay as is.

> Note that I reduced the timeout time of the toast to 3,5 seconds as 4 secs felt too long for me. Feel free to argue with that :D

Sounds fair :)

MichaelBuessemeyer:
@philippotto I just applied your feedback 🌮.

Thanks for informing me that another setting was added to the sharing modal on the master. This made it easier to understand how to resolve the merge conflicts.
Note that the current update to this pr also fixes a bug: Previously the old value of the visibility and the shared teams setting was used when sending updates about changes to the backend server. This was actually a but and I found this one by setting some settings, reloading, and then opening the sharing modal again. Prior to the fix, the settings were not what I left before reloading. But now this bug should be resolved and the settings should be correctly persisted. I checked this myself but please do so too to avoid merging bugs 🙈.

This pr is ready for your second review 🏁

philippotto:
> Note that the current update to this pr also fixes a bug: Previously the old value of the visibility and the shared teams setting was used when sending updates about changes to the backend server. This was actually a but and I found this one by setting some settings, reloading, and then opening the sharing modal again. Prior to the fix, the settings were not what I left before reloading. But now this bug should be resolved and the settings should be correctly persisted. I checked this myself but please do so too to avoid merging bugs .

Thanks for fixing this! Was the bug introduced recently (e.g, with the allow-contributors-to-modify change) or has it been like this for a while? Could you point to the line changes where you fixed this?

MichaelBuessemeyer:
> Thanks for fixing this! Was the bug introduced recently (e.g, with the allow-contributors-to-modify change) or has it been like this for a while? Could you point to the line changes where you fixed this?

Oh, maybe I misformualted my previous message. The truth is: I introduced this bug with this pr and noticed it only pretty late 😅. That's why I wanted to make sure that this bug does not make it to production by asking you to also specifically checking the fix.

frontend/javascripts/oxalis/view/action-bar/share_modal_view.tsx
MichaelBuessemeyer: > Are you sure about this? Maybe it changed in the latest antd versions, but the current documentation + example seems to allow duplicate messages if no key is provided. The very first demo here suggests this at least: https://ant.design/components/notification/#header

Oh, that might be right but we changed this behaviour in our `toast.tsx` lib. Toast code sets a key per default as long as the message of the toast is a plain string.
See: 
https://github.com/scalableminds/webknossos/blob/4e1d4d644e137ee4d06859d43303685ea2211787/frontend/javascripts/libs/toast.tsx#L86

Therefore the explicit key is needed and the key needs to be a random string.

philippotto: ah, I see, thank you for looking this up :+1:

philippotto:
Nice! Seems to work great :) I noticed two minor things which should be changed:

- When a save requests fails (e.g., because the client is offline), the dialog is disabled completely, no error is shown and closing + reopening the dialog does not change anything. Effectively, the user has to reload the page to change the settings. An error message should be shown and maybe offer a retry possibility?
- the "Cancel" button is superfluous (and potentially irritating) now -> can be removed.

MichaelBuessemeyer:
> When a save requests fails (e.g., because the client is offline), the dialog is disabled completely, no error is shown and closing + reopening the dialog does not change anything. Effectively, the user has to reload the page to change the settings. An error message should be shown and maybe offer a retry possibility?

Thanks for finding this. I now ensure that upon a failed update request an error is shown to the user, the UI resets its state prior to the failure so that it matches what's currently saved in the backend and it is always reenabled. For additional information, the error is logged to the console.

> the "Cancel" button is superfluous (and potentially irritating) now -> can be removed.

👍  Done 🍡

philippotto (approved):
Excellent :100:

PR #6315, opened 2022-07-05
Enable publications to link to annotations

### LINKED ISSUE #3937: Show case annotations in publications view
Right now, featured publications link datasets (including segmentations in case of volume annotations). However, skeleton tracings cannot be linked currently. It would be cool to be able to also link concrete annotations instead of only datasets.

leowe:
- https://___.webknossos.xyz

### Steps to test:
- request `http://webknossos/api/publications`

### Issues:
- closes #3937

------
- [x] Ready for review

------
Still todo
- [x] Check with frontend whether provided information is enough


app/models/binary/Publication.scala
```diff
@@ -30,6 +37,46 @@ class PublicationService @Inject()()(implicit ec: ExecutionContext) {
...
+                                   annotationDAO: AnnotationDAO)(implicit ec: ExecutionContext) {
+
+  def publicWritesWithDatasetsAndAnnotations(publication: Publication): Fox[JsObject] = {
+    implicit val ctx: DBAccessContext = GlobalAccessContext
+    for {
+      dataSets <- dataSetDAO.findAllByPublication(publication._id) ?~> "not found" ~> NOT_FOUND
+      annotations <- annotationDAO.findAllByPublication(publication._id) ?~> "not found" ~> NOT_FOUND
+      dataSetsJson <- Fox.serialCombined(dataSets)(d => dataSetService.publicWrites(d, None, None, None))
+      annotationsJson <- Fox.serialCombined(annotations) { annotation =>
+        for {
```
fm3, line 56: I’d suggest to extract this into a separate function, either here, or in the annotation service. maybe `annotationWritesWithDataset`?

tools/postgres/schema.sql
```diff
@@ -34,6 +34,7 @@ CREATE TABLE webknossos.annotations(
   _task CHAR(24),
   _team CHAR(24) NOT NULL,
   _user CHAR(24) NOT NULL,
+  _publication CHAR(24),
```
fm3, line 37: there is a foreign key section at the bottom of schema.sql. this reference should be added there (also to the evolution)

fm3 (commented):
Looks good already :) I added two small comments. Otherwise, Let’s wait for the frontend

normanrz (commented):
2 things I noticed in the `GET /api/publications` API:
- A dataset contains the information about a publication, which is redundant.
- An annotation has the id in the form of `annotations: [{ "id": { "id": "...." }, ... }, ...]`, which seems to be unnecessarily nested.

normanrz:
I would argue, we can remove the `publication` field from the `GET /api/datasets` route

fm3:
> I would argue, we can remove the publication field from the GET /api/datasets route

That’s good news, we weren’t sure if the front-end needs this in some other place (i.e. to link back to the publication from the dataset page or something). But if that is not needed, that will simplify the json and the backend code :)

> unnecessarily nested

That’s probably just the objectid wrapper, @leowe you can un-nest it by passing annotation._id.id to the json

normanrz:
I fixed the annotation `id` and added `name` and `description` to the annotation object.

So, what is left is removing the `publication` from the datasets.

normanrz:
@philippotto please review the frontend part :-)

normanrz:
@fm3 I now also removed the `publication` field in https://github.com/scalableminds/webknossos/pull/6315/commits/22bfe14becce8eceeba7636aa22a277e080ec93a. Can you please check that I did that correctly?

frontend/javascripts/dashboard/publication_details_view.tsx
```diff
@@ -1,116 +1,72 @@
...
-        isLoading: false,
-      });
-    }
-  }
-
-  render() {
-    const { isLoading, datasets } = this.state;
-    const datasetsOfPublication = datasets.filter(
-      (ds) =>
-        ds.isActive && ds.publication != null && ds.publication.id === this.props.publicationId,
```
philippotto, line 64: the filtering by `isActive` isn't done anywhere, anymore, right? isn't it still necessary?

app/models/binary/Publication.scala
```diff
@@ -18,7 +22,8 @@ case class Publication(_id: ObjectId,
                        created: Long = System.currentTimeMillis(),
                        isDeleted: Boolean = false)
 
-class PublicationService @Inject()()(implicit ec: ExecutionContext) {
+class CompactPublicationService @Inject()()(implicit ec: ExecutionContext) {
```
fm3, line 25: I’d say this can be merged into one service, now that the dataset publicWrites no longer includes the publication :) cc @leowe

philippotto (commented):
the front-end looks great :+1:  nice clean up! didn't test, though, since I don't know how to do that. maybe you could prepare a dev instance where some publications are available?

frontend/javascripts/dashboard/publication_details_view.tsx
normanrz: Good question. @leowe can non-active datasets be part of a publication?

frontend/javascripts/dashboard/publication_card.tsx
```diff
@@ -21,26 +26,50 @@ type ExtendedDatasetDetails = APIDatasetDetails & {
 };
 const thumbnailDimension = 500;
 const miniThumbnailDimension = 75;
-const blacklistedSegmentationNames = ["2012-09-28_ex145_07x2_ROI2017"];
```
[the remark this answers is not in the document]
normanrz: The problem was that this dataset only had an oversegmentation and was at the top of the list. Since the dataset is now much lower in the list and we have a version with applied mapping, this hotfix can go.

frontend/javascripts/dashboard/publication_card.tsx
```diff
@@ -21,26 +26,50 @@ type ExtendedDatasetDetails = APIDatasetDetails & {
...
 
-function getDisplayName(dataset: APIDataset): string {
-  return dataset.displayName != null && dataset.displayName !== ""
-    ? dataset.displayName
-    : dataset.name;
+enum AnnotationOrDataset {
+  ANNOTATION = "ANNOTATION",
+  DATASET = "DATASET",
 }
+type AnnotationOrDatasetItem =
```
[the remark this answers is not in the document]
normanrz: Do we use the `-able` prefix for types somewhere already?

frontend/javascripts/dashboard/publication_details_view.tsx
leowe: The publication simply has a foreign key on the annotations which itself has a foreign key on the datasets. There are no checks in the code on active/non-active datasets except that we don't show deleted datasets.

app/models/binary/Publication.scala
```diff
@@ -18,17 +22,32 @@ case class Publication(_id: ObjectId,
...
 
-class PublicationService @Inject()()(implicit ec: ExecutionContext) {
-  def publicWrites(p: Publication): Fox[JsObject] =
-    Fox.successful(
+class PublicationService @Inject()(dataSetService: DataSetService,
+                                   dataSetDAO: DataSetDAO,
+                                   annotationService: AnnotationService,
+                                   annotationDAO: AnnotationDAO)(implicit ec: ExecutionContext) {
+
+  def publicWritesWithDatasetsAndAnnotations(publication: Publication): Fox[JsObject] = {
```
fm3, line 30:
```suggestion
  def publicWrites(publication: Publication): Fox[JsObject] = {
```

app/controllers/PublicationController.scala
```diff
@@ -0,0 +1,47 @@
...
+
+  @ApiOperation(value = "Information about a publication", nickname = "publicationInfo")
+  @ApiResponses(
+    Array(new ApiResponse(code = 200, message = "JSON object containing information about this publication."),
+          new ApiResponse(code = 400, message = badRequestLabel)))
+  def read(@ApiParam(value = "The id of the publication") publicationId: String): Action[AnyContent] =
+    sil.UserAwareAction.async { implicit request =>
+      for {
+        publication <- publicationDAO.findOne(ObjectId(publicationId)) ?~> "publication.notFound" ~> NOT_FOUND
+        js <- publicationService.publicWritesWithDatasetsAndAnnotations(publication)
```
fm3, line 35:
```suggestion
        js <- publicationService.publicWrites(publication)
```

app/controllers/PublicationController.scala
```diff
@@ -0,0 +1,47 @@
...
+      } yield Ok(js)
+    }
+
+  def listPublications: Action[AnyContent] = sil.UserAwareAction.async { implicit request =>
+    {
+      for {
+        publications <- publicationDAO.findAll ?~> "publication.notFound" ~> NOT_FOUND
+        jsResult <- Fox.serialCombined(publications)(publicationService.publicWritesWithDatasetsAndAnnotations)
```
fm3, line 43:
```suggestion
        jsResult <- Fox.serialCombined(publications)(publicationService.publicWrites)
```

frontend/javascripts/dashboard/publication_details_view.tsx
normanrz: I added an `isActive` filter.

frontend/javascripts/dashboard/publication_card.tsx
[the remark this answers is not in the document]
philippotto: > Do we use the `-able` prefix for types somewhere already?

probably not, but in the end it's an adverb which (in combination with a noun) would be perfectly acceptable to me. I don't really like the standalone version of it (`Publishable`), but wanted to avoid the `type` suffix, because we decided at some point that it's redundant in most cases. But maybe, this time it has a right to exist (especially, since you also used `-type` in your new commit).

What I don't like about the new `PublicationItemType` and `PublicationItem` is that it sounds like it could be an "item of an publication array". But `Publication != PublicationItem`. That's why I came up with `PublishableItem`. However, it's not a huge deal. Naming is hard and this discussion is probably not worth to carry out more :sweat_smile: Do what you deem is ok :)

fm3 (approved):
Backend LGTM

normanrz:
> maybe you could prepare a dev instance where some publications are available?

Here you go: https://publicationsannotations.webknossos.xyz/ or https://publicationsannotations.webknossos.xyz/publications/5c766bec6c01006c018c7459

frontend/javascripts/dashboard/publication_card.tsx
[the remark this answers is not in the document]
normanrz: I think I'll keep it as is now :-)

philippotto (approved):
awesome :man_dancing:

PR #6317, opened 2022-07-06
Support uint64 segmentations (up to 2^53 - 1)

### LINKED ISSUE #4852: Fully support uint64 segmentations and volume annotations
As far as I know this was/is blocked by javascript not supporting uint64 arrays. Let’s track here that we should add this feature once js supports this.

fm3:
- note that the backend will support only segment ids up to (2^63)-1 (scala/java long type)
- note that the frontend will only support segment ids up to (2^53)-1 (javascript number type)
- the back-end asserts the smaller (2^53)-1 constraint
- JS-side mappings (aka JSON mappings) are only supported for 32 bits right now. If we want to improve this, cuckoo hashing could become useful again, but I'm not sure about the priority here which is why I'd defer this.

### Todo:
- [X] warn user about merger mode support being limited to 32-bits
- [x] warn user about mapping support being limited to 32-bits

- https://uint.webknossos.xyz/datasets/sample_organization/64bit_segmentation/view#12,16,0,0,0.164

### Steps to test:
- I created a test dataset and uploaded it to the dev instances (see [here](https://www.notion.so/scalableminds/Test-Data-c0563be9c4a4499dae4e16d9b2497cfb#0412f70b2dfc4b1988e3ef4cea05c738) for details; including how this was produced to see where to expect which voxels)
- test normal viewing of the dataset
  - differing segment ids should have different colors and should be isolated on hover
- create an annotation with the existing layer
  - test brushing
  - test both overwrite modes
  - test the segment list view
  - test isosurface rendering
  - test the merger mode (it should warn that only 32-bit values are valid)
  - activate a json mapping (it should warn that only 32-bit values are valid)
  - test the volume interpolation feature
  - test the flood fill

### Issues:
- fixes #4852 

------
- [x] Ready for review


philippotto:
@fm3 could you resolve the merge conflicts if you have a spare second? :)

fm3:
done :)

frontend/javascripts/libs/utils.ts
```diff
@@ -746,9 +747,33 @@ export function convertDecToBase256(num: number): Vector4 {
 
   [tmp, a] = divMod(tmp); // eslint-disable-line prefer-const
 
-  // Big endian
-  return [a, b, g, r];
+  // Little endian
+  return [r, g, b, a];
```
philippotto, line 751: The order used to be reversed by the caller which is why I changed it here. It wasn't a bug.

philippotto:
> the back-end asserts the smaller (2^53)-1 constraint

@fm3 how is this done and how is it communicated to the user? I created a segmentation which violates this constraint and I don't see any error message or warning :thinking:

fm3:
> @fm3 how is this done and how is it communicated to the user? I created a segmentation which violates this constraint and I don't see any error message or warning thinking

Yes, the checks are not run on every dataset load. There are a couple of events that trigger validation
 - Uploading a dataset
 - The save request after editing the datasource in the dataset settings view
 - Creating volume annotation with fallback layer
 - Uploading volume annotation with fallback layer
 - Merging volume annotations (where the remapping of ids can cross the threshold)

All of this assumes correct largestSegmentId values. Just inserting larger ids somewhere in the data will presumably always fail silently. Do you think we should change that? If I remember correctly, so far, we never let bucket requests fail (instead return the missing buckets header for any failed loading)

philippotto:
> All of this assumes correct largestSegmentId values. Just inserting larger ids somewhere in the data will presumably always fail silently. Do you think we should change that?

I think that makes sense :+1: Dealing with incorrect largest segment ids is out of scope for this PR in my opinion and I also think it's fine to keep it as is in general.

> Also please notify me if the above checks don’t appear to work for you

It worked for me when editing the datasource-properties.json :)

philippotto:
@daniel-wer I think this PR is ready for review. Let me know in case you want to have a call, but I hope most of it is relatively straightforward. The PR turned out simpler than thought, since we already had the `packingDegree` logic which is able to pack multiple voxels into one texel. By defining the packing degree as `0.5` for 64-bit segmentations, this logic automatically handles the inverted case (having multiple texel for one voxel). This was quite useful :)

frontend/javascripts/oxalis/model/sagas/mapping_saga.ts
```diff
@@ -175,6 +188,15 @@ function* maybeFetchMapping(
     mappingColors,
     hideUnmappedIds,
   };
+
+  if (layerInfo.elementClass === "uint64") {
```
daniel-wer, line 192: See my other comment, shouldn't `int64` appear here as well?

frontend/javascripts/oxalis/model/sagas/volume/volume_interpolation_saga.ts
```diff
@@ -389,8 +428,8 @@ export default function* maybeInterpolateSegmentationLayer(): Saga<void> {
 
   for (let u = 0; u < size[firstDim]; u++) {
     for (let v = 0; v < size[secondDim]; v++) {
-      const firstVal = firstSliceDists.get(u, v);
-      const lastVal = lastSliceDists.get(u, v);
+      const firstVal = Number(firstSliceDists.get(u, v));
+      const lastVal = Number(lastSliceDists.get(u, v));
```
daniel-wer, line 432: Not that important, but isn't `firstSlice` and, therefore, `firstSliceDists` of type `NdArray<TypedArrayWithoutBigInt>`. I would have expected the `Number()` call to be superfluous.

frontend/javascripts/oxalis/shaders/main_data_fragment.glsl.ts
```diff
@@ -190,18 +191,21 @@ void main() {
   <% _.each(segmentationLayerNames, function(segmentationName, layerIndex) { %>
 
      // Color map (<= to fight rounding mistakes)
-     if ( length(<%= segmentationName%>_id) > 0.1 ) {
+     if ( length(<%= segmentationName%>_id_low) > 0.1 || length(<%= segmentationName%>_id_high) > 0.1 ) {
        // Increase cell opacity when cell is hovered
        float hoverAlphaIncrement =
          // Hover cell only if it's the active one, if the feature is enabled
```
daniel-wer, line 197: I think this has been removed before your PR, at least I don't see it

```suggestion
         // Hover cell only if it's the active one
```

frontend/javascripts/oxalis/shaders/texture_access.glsl.ts
```diff
@@ -41,50 +41,39 @@ export const getRgbaAtIndex: ShaderModule = {
...
+      vec4 getRgbaAtXYIndex_<%= name %>(float textureIdx, float x, float y) {
         // Since WebGL 1 doesnt allow dynamic texture indexing, we use an exhaustive if-else-construct
         // here which checks for each case individually. The else-if-branches are constructed via
         // lodash templates.
 
         <% if (dataTextureCountPerLayer === 1) { %>
             // Don't use if-else when there is only one data texture anyway
-            return texture2D(<%= name + "_textures" %>[0], accessPoint).rgba;
+
+            return texelFetch(<%= name + "_textures" %>[0], ivec2(x, y), 0).rgba;
```
daniel-wer, line 57: Nice, `texelFetch` is exactly what we need here - glad to see the workaround for this being removed.

frontend/javascripts/oxalis/view/merger_mode_modal_view.tsx
```diff
@@ -29,6 +34,16 @@ export default function MergerModeModalView({ isCloseable, onClose, progress }:
         </div>
       }
     >
+      {isUint64SegmentationVisible && (
+        <Alert
+          style={{ marginBottom: 12 }}
+          message="Warning"
+          description="The merger mode is limited to 32-bit. However, your current segmentation layer uses
+          64 bits. The rendering of segment ids larger than 2^32 will be incorrect."
```
daniel-wer, line 42:
```suggestion
          64 bits. The rendering of segment ids greater than or equal to 2^32 will be incorrect."
```

daniel-wer (commented):
Very nice that this didn't require large-scale changes and that quite a few workarounds could be removed! I have a couple of clarifying questions, but apart from that the code looks good to me :+1: 

Will report back after testing.

frontend/javascripts/libs/utils.ts
```diff
@@ -746,9 +747,33 @@ export function convertDecToBase256(num: number): Vector4 {
...
+export function castForArrayType(uncastNumber: number, data: TypedArray): number | bigint {
+  return data instanceof BigUint64Array ? BigInt(uncastNumber) : uncastNumber;
+}
+
+export function convertNumberTo64Bit(num: number): [Vector4, Vector4] {
+  // Cast to BigInt as bit-wise operations only work with 32 bits,
+  // even though Number uses 53 bits.
+  const hoveredSegmentId = BigInt(num);
+  let low;
+  let high;
```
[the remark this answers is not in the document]
philippotto: oh, this is an artifact from a point where the blocks used the same variable names (`const [r, g, b, a]`), but this is not necessary anymore. I'm cleaning it up now :)

frontend/javascripts/oxalis/model/sagas/volume/volume_interpolation_saga.ts
philippotto: you're right. this change also originates from an older version where big ints were allowed here, too. changing it now.

daniel-wer:
Issues I noticed during testing:
-  [ ] When opening the Segments tab with segments listed and hovering over one of them, there is an Error Toast, indicating an exception
  - ```
    console.js:13 TypeError: Cannot convert null to a BigInt
    at BigInt (<anonymous>)
    at Object.Me (utils.ts:761:28)
    at plane_material_factory.ts:529:33
    at s (listener_helpers.ts:32:7)
    at m (createStore.js:173:7)
    at redux-saga-core.esm.js:1410:22
    at overwrite_action_middleware.ts:62:16
    at action_logger_middleware.ts:41:14
    at setHoveredSegmentId (segments_view.tsx:133:5)
    at onMouseLeave (segment_list_item.tsx:375:9)
    ```
- [ ] The segments in the segments tab (and the isosurfaces) have a different color than the segment in the segmentation
- [ ] Flood fill doesn't work yet
  ![image](https://user-images.githubusercontent.com/1702075/183891032-cb1aa010-d102-4e3f-87d0-a84a99e75f32.png)
- [ ] When a mapping is activated, unmapped segments are no longer highlighted when hovered
  - [ ] Also, the "Hide unmapped segments" functionality seems to behave weirdly. Activating it doesn't make the unmapped segments disappear, but change color instead
- [ ] When adding a new volume annotation layer, it is 32-bit, although the original segmentation layer is 64-bit. Not sure whether  this needs to be changed in this PR. Could be a follow-up to allow to specify this upon annotation layer creation.

fm3:
> When adding a new volume annotation layer, it is 32-bit, although the original segmentation layer is 64-bit.

This sounds like a backend bug, I’ll have a look

daniel-wer:
> This sounds like a backend bug, I’ll have a look

Not sure. When basing the new layer on the original uint64 segmentation layer, the resulting layer is uint64 as well, so this works.
When adding a new layer not based on the fallback layer, it is uint32 automatically. I wouldn't know how to decide since there could be multiple other segmentation layers (uint32 and uint64), so I think one would need to let the user decide.

philippotto:
Thanks for the review and for testing, @daniel-wer :pray: 

> Issues I noticed during testing:

Oh, I'm sorry, that you ran into these several issues. I swear that flood fill and the segments tab worked at some point :see_no_evil: I should have tested again after the latest iteration. Will fix these now.

> When adding a new layer not based on the fallback layer, it is uint32 automatically. I wouldn't know how to decide since there could be multiple other segmentation layers (uint32 and uint64), so I think one would need to let the user decide.

I think, the use case for an empty segmentation layer which uses 64 bit is very rare. Similarly rare to explicitly requesting a 16 bit segmentation layer, I'd say (which is also not possible right now). Defaulting further to 32 bits seems like a good choice to me for now (especially, because the user doesn't have to think about the bit depth question).

fm3:
> Not sure. When basing the new layer on the original uint64 segmentation layer, the resulting layer is uint64 as well, so this works. When adding a new layer not based on the fallback layer, it is uint32 automatically. I wouldn't know how to decide since there could be multiple other segmentation layers (uint32 and uint64), so I think one would need to let the user decide.

Ah I see, I thought you were talking about with fallback layer. When creating a fresh volume layer, I’d say the existing layers are not relevant. IDs will start at 1 again for that layer, so uint32 should usually be enough. When later merging this layer with another volume layer, it may become necessary to switch to uint64 again. But this is a problem of the merging process. Since it is just one of several cases the merging process does not currently handle well, I’d say it is fair as a follow-up issue. I wrote #6389. 

Another case is download + reupload (if there is no fallback layer). I also wrote #6390 for proper handling of element class and largestSegmentId there. But again, problems should only arise if IDs that do not fit in uint32 are used, despite not having a fallback layer.

philippotto:
> When a mapping is activated, unmapped segments are no longer highlighted when hovered
> * [ ]  Also, the "Hide unmapped segments" functionality seems to behave weirdly. Activating it doesn't make the unmapped segments disappear, but change color instead

Did you test this with segment ids < 2**32? For these, both features work for me as intended. For greater IDs (regardless of whether they are part of the mapping or not), the mapping code doesn't work correctly (as it should be signaled by the toast message when activating the mapping).

philippotto:
@fm3 what do you think about the signed 64 bit support? I just wanted to test what happens with it (especially with negative ids) and noticed that the "create annotation" functionality breaks for signed 64 bit. See this stacktrace:

```
play.api.http.HttpErrorHandlerExceptions$$anon$1: Execution exception[[MatchError: int64 (of class scala.Enumeration$Val)]]
        at play.api.http.HttpErrorHandlerExceptions$.$anonfun$convertToPlayException$2(HttpErrorHandler.scala:381)
        at scala.Option.map(Option.scala:230)
        at play.api.http.HttpErrorHandlerExceptions$.convertToPlayException(HttpErrorHandler.scala:380)
        at play.api.http.HttpErrorHandlerExceptions$.throwableToUsefulException(HttpErrorHandler.scala:373)
        at play.api.http.DefaultHttpErrorHandler.onServerError(HttpErrorHandler.scala:264)
        at play.core.server.AkkaHttpServer$$anonfun$2.applyOrElse(AkkaHttpServer.scala:430)
        at play.core.server.AkkaHttpServer$$anonfun$2.applyOrElse(AkkaHttpServer.scala:422)
        at scala.concurrent.Future.$anonfun$recoverWith$1(Future.scala:417)
Caused by: scala.MatchError: int64 (of class scala.Enumeration$Val)
        at com.scalableminds.webknossos.datastore.models.datasource.ElementClass$.maxSegmentIdValue(DataLayer.scala:56)
        at com.scalableminds.webknossos.datastore.models.datasource.ElementClass$.largestSegmentIdIsInRange(DataLayer.scala:64)
        at models.annotation.AnnotationService.$anonfun$createTracingsForExplorational$3(AnnotationService.scala:194)
        at com.scalableminds.util.tools.Fox.$anonfun$flatMap$1(Fox.scala:252)
        at scala.concurrent.Future.$anonfun$flatMap$1(Future.scala:307)
        at akka.dispatch.BatchingExecutor$AbstractBatch.processBatch(BatchingExecutor.scala:63)
        at akka.dispatch.BatchingExecutor$BlockableBatch.$anonfun$run$1(BatchingExecutor.scala:100)
```

I'm wondering whether there is a real use case for signed ids. Maybe it would alright to just not support the signed variants?

philippotto:
> To be honest I’m not a fan of the idea of negative IDs. There is currently quite a bit of code in the backend that relies on them being positive. I also don’t really see a use case. So let’s create that issue when somebody really wants to use that ;)

Yes, I agree with you. So, as a consequence, there is also no need for signed 64 bit segmentations that only use positive ids, right? Then, I'd a front-end assertion so that `int64` is not allowed.

fm3:
Yes, sounds good!

frontend/javascripts/oxalis/model/accessors/dataset_accessor.ts
```diff
@@ -698,6 +698,12 @@ export function getVisibleOrLastSegmentationLayer(
 
   return null;
 }
+
+export function hasVisibleUint64Segmentation(state: OxalisState) {
```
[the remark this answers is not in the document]
philippotto: for the record: we decided to not support int64 for now.

philippotto:
The PR is ready from my point of view. I created https://github.com/scalableminds/webknossos/issues/6399 as a follow-up, but the priority needs to be discussed. Is something missing from your side, @fm3 and @daniel-wer?

fm3:
I think we created follow-up issues for everything I can think of at the moment :)

daniel-wer (commented):
Code LGTM :+1: 

However, in Chrome WebGL crashes every time I try to open any dataset of the deployed branch. Not sure whether this is due to the new Chrome version (104) or something else (I tried Chrome and computer restarts)? However, I remember successfully testing the older deployed version which crashed today as well (afterwards I deployed the newest version which also crashes for me).
Firefox works. Other branches or deployed instances work.
I can reproduce this locally as well.

Could you check your Chrome versions and try as well?

daniel-wer:
Git bisect reveals that the [very first frontend commit](https://github.com/scalableminds/webknossos/pull/6317/commits/3aa2bcfe15e610350b2c99c5444b2d86db844c7d) already triggers the WebGL crash for me. The one before works well. My test case was simply opening the l4_sample dataset in view mode.

philippotto:
> Code LGTM +1
> 
> However, in Chrome WebGL crashes every time I try to open any dataset of the deployed branch. Not sure whether this is due to the new Chrome version (104) or something else (I tried Chrome and computer restarts)? However, I remember successfully testing the older deployed version which crashed today as well (afterwards I deployed the newest version which also crashes for me). Firefox works. Other branches or deployed instances work. I can reproduce this locally as well.
> 
> Could you check your Chrome versions and try as well?

Weird. I just updated to Chrome v104 and the branch works without problems for me. I asked the rest of the team, but if this PR makes wk unusable with chrome for you, I'm a bit hesitant to merge it :see_no_evil: especially, because others from our user base might be running into the same problem.

Does the console show some info other than "webgl crashed"? RAM is probably alright, right? Maybe there is something in chrome://crashes/ ?

daniel-wer:
> Does the console show some info other than "webgl crashed"? RAM is probably alright, right? Maybe there is something in chrome://crashes/ ?

For the record, restarting chrome again and closing some other hibernated webKnossos tabs made it work for me as well. Console just printed the whole shader code. RAM was available and chrome://crashes showed the crash but no additional information.

daniel-wer (approved):
:+1:

PR #6324, opened 2022-07-07
3rd attempt: Merge "Shared Annotations" with "My annotations"

fm3:
This reverts commit 4ee95f30e770e98852e42b747c9986a574741850.

new attempt after #6287 still had a bug and had to be reverted

The list access query still showed too many annotations – this time only from the own organization, but from other users, without them using the share-with-team functionality.

Desired behavior/steps to test

| Scenario | See With Link | Listed |
| ------------- | ------------- | ------------- | 
| You are owner of annotation | ✓| ✓|
| Annotation is explicitly shared with a team you are in, visibility is internal or public | ✓| ✓|
| Annotation visibility is public, in any orga | ✓| |
| Annotation visibility is internal, in your orga | ✓| |
| Annotation visibility is private, you do not manage the author. (team sharing is overruled/disabled) | | |
| Annotation visibility is private, you manage the author | ✓| |
| Annotation visibility is private, you are the orga admin | ✓| |


app/models/annotation/Annotation.scala
```diff
@@ -178,16 +178,24 @@ class AnnotationDAO @Inject()(sqlClient: SQLClient, annotationLayerDAO: Annotati
...
+          and _id in
+          (select distinct a._annotation
+           from webknossos.annotation_sharedTeams a
+           join webknossos.user_team_roles t
+           on a._team = t._team
+           where t._user = '${requestingUserId.id}')))
+       """
+
+  override def readAccessQ(requestingUserId: ObjectId): String =
+    s"""    (visibility = '${AnnotationVisibility.Public}'
```
jstriebel, line 192: Please have the outer brackets on separate lines for readability, everything else is quite parsable :+1: 
```suggestion
    s"""(
             visibility = '${AnnotationVisibility.Public}'
```

jstriebel (approved):
Thanks for the great overview table! I double-checked the queries against this table, they seem to match. Not sure about the actual desired behavior, but it looks good at first glance :+1:

Maybe this is a good timepoint to introduce some tests for this?

PR #6325, opened 2022-07-07
Avoid proofreading actions while the last one is not finished

### LINKED ISSUE #6296: Creating multiple edges quickly in proofreading mode crashes
- Open an annotation with agglomerate mapping
- Activate agglomerate mapping
- Activate proofreading mode
- click on multiple segments
- quickly merge one tree with all other trees
- save

image

Either, the constraint should be softened or the front-end should be responsible for splitting the updates in distinct groups.

daniel-wer:
Set wk busy during proofreading and also modify the input catchers so that they not only show a loading spinner if wk is busy, but actually block all inputs.

- https://___.webknossos.xyz

### Steps to test:
- Open an annotation with an agglomerate view, activate it and switch to proofreading mode
- Click a couple of agglomerates to load their agglomerate skeletons
- Open the dev tools, switch to the "Network" tab and select "Slow 3G" throttling so that the proofreading action takes longer
- Do a proofreading action by adding or deleting an edge and try to perform another proofreading action in rapid succession
  - The second action should no longer be possible since all inputs to the viewports are blocked and a loading spinner is shown

### Issues:
- fixes #6296 

------
- [x] Ready for review


frontend/javascripts/oxalis/view/right-border-tabs/connectome_tab/synapse_tree.tsx
```diff
@@ -234,7 +234,7 @@ class SynapseTree extends React.Component<Props, State> {
           <Dropdown // Lazily create the dropdown menu and destroy it again, afterwards
             overlay={() => this.createSegmentDropdownMenu(data.id)}
             autoDestroy
-            placement="bottomCenter"
+            placement="bottom"
```
daniel-wer, line 237: Saw this deprecation warning: `Warning: [antd: Dropdown] You are using 'bottomCenter' placement in Dropdown, which is deprecated. Try to use 'bottom' instead. `

daniel-wer: I upgraded antd to a slightly newer version, because although this was deprecated, the type annotations didn't allow it yet :see_no_evil: 
As far as I can see, not much changed, but typing seems to have been improved and I was able to remove quite a few type ignores.

MichaelBuessemeyer (approved):
Looks good to me 🎉.

I only got one point, where I was unsure whether the deletion of a line was intentional. I'll approve anyway so you can directly merge when this was intentional. Otherwise might need to wait a bit again until I can approve :)

I only managed to sometimes load a proofreading annotation where the hdf5 mapping was already loaded directly after opening the annotation. But the proofreading tool was not visible in the toolbar. Only after I switched the mapping, the proofreading tool was visible. I think this is unrelated but still a bug, right?

frontend/javascripts/oxalis/model/sagas/proofread_saga.ts
```diff
@@ -279,13 +279,20 @@ function* splitOrMergeAgglomerate(action: MergeTreesAction | DeleteEdgeAction) {
...
+
+  yield* put(setBusyBlockingInfoAction(true, "Proofreading action"));
+
   if (!volumeTracing.mappingIsEditable) {
     try {
       yield* call(createEditableMapping);
     } catch (e) {
       console.error(e);
-      // Undo the last splitting/merging action since the proofreading action did not succeed
-      yield* put(undoAction());
```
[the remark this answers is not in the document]
daniel-wer: This was an oversight while merging the original proofreading PR. Undo/Redo is actually not allowed if the proofreading tool is active and a warning in that regard is shown to the user if it is triggered. This was added kind of last-minute, because the actual proofreading actions cannot be undone yet. Unfortunately, that's why this line had to be removed for now.

daniel-wer:
Thanks a lot for the review!

> I only managed to sometimes load a proofreading annotation where the hdf5 mapping was already loaded directly after opening the annotation. But the proofreading tool was not visible in the toolbar. Only after I switched the mapping, the proofreading tool was visible. I think this is unrelated but still a bug, right?

This sounds like a bug, I'll have a look :)

daniel-wer:
@MichaelBuessemeyer Thanks again for indicating the buggy behavior. I've pushed a fix :)

MichaelBuessemeyer (approved):
> Thanks again for indicating the buggy behavior. I've pushed a fix :)

I just tested the fix locally and it worked 🎉. Thanks for fixing this bug 🐛🧋.

The pr should be ready to merge. (I do not think a changelog entry for the fix is necessary.)

PR #6326, opened 2022-07-11
Make Annotation Layer Names Unique in DB

### LINKED ISSUE #6251: Make annotation layer names unique per annotation

fm3:
The Front-end already enforces unique url-safe layer names as of #6289 – this PR adds the same constraints in the back-end.
Note that the evolution is not fully automatic this time (would have become complex). Instead, instructions are added to the migration guide on how to manually migrate the db content.

Some backend changes followed from the name now being non-optional.

Additionally, The layer name displayed by the front-end for Skeleton Annotation Layers is changed from »Skeletons« to »Skeleton« as discussed [here](https://scm.slack.com/archives/C5AKLAV0B/p1655298529512849?thread_ts=1655298383.390979&cid=C5AKLAV0B) – Note that in the Skeleton case, the front-end does not use or set the annotation layer name from the DB, but instead uses this string. This is a special case enabled by the fact that there is currently always at most one skeleton layer.

### Steps to test:
- evolutions should be equal to schema (asserted by CI)
- users should be able to set volume layer names according to the constraints

### Issues:
- fixes #6251 

------
- [x] Ready for review


tools/postgres/schema.sql
```diff
@@ -56,7 +56,8 @@ CREATE TABLE webknossos.annotation_layers(
   _annotation CHAR(24) NOT NULL,
   tracingId CHAR(36) NOT NULL UNIQUE,
   typ webknossos.ANNOTATION_LAYER_TYPE NOT NULL,
-  name VARCHAR(256),
+  name VARCHAR(256) NOT NULL CHECK (name ~* '^[A-Za-z0-9\-_\.]+$'),
+  UNIQUE (name, _annotation),
```
jstriebel, line 60: Should those constraints also get names, similar to the evolution? Or are the names only needed to allow the reverse evolution?

jstriebel (approved):
LGTM :+1:

tools/postgres/schema.sql
fm3: postgres generates default names using a fixed schema, these are the ones I use in the reversion

PR #6330, opened 2022-07-13
Improve convert to hybrid workflow

### LINKED ISSUE #5695: [Follow up]: Multiple segmentation layers

### LINKED ISSUE #6209: Improve "convert to hybrid" flow with regard to segmentation layers
When having a skeleton-only annotation, one can click on "create to hybrid", but one doesn't have control over the creation of the volume layer (which one will be editable?).

In the case of a skeleton-only annotation, I'd argue that one should simply use the "Add Volume Annotation Layer" button (which should enable to select which layer should be used). Or even better: add a "make editable" button to each segmentation layer (which is not editable yet). In that case, the "convert to hybrid" button would be superfluous.

Dagobert42:
- https://___.webknossos.xyz

### Steps to test:
- create a volume-only annotation
- Annotation Type: X is removed from info tab
- the layers tab now displays an Add Skeleton Annotation Layer button with the old convert to hybrid functionality
- segmentation layers which were not previously editable now show an (un)lock icon button which shortcuts to the "add volume layer" modal with the layer being preselected
- should be disallowed for tasks

### Issues:
- fixes #6209

------
- [x] Ready for review

### DIFF
--- a/frontend/javascripts/components/hover_icon_button.tsx
+++ b/frontend/javascripts/components/hover_icon_button.tsx
@@ -1,0 +1,32 @@
+import { ButtonProps } from "antd";
+import * as React from "react";
+const { useState } = React;
+
+export type HoverButtonProps = Omit<ButtonProps, "icon"> & {
+  icon: React.ReactElement<any>;
+  hoveredIcon: React.ReactElement<any>;
+};
+
+export function HoverIconButton(props: HoverButtonProps) {
+  const [isMouseOver, setIsMouseOver] = useState<boolean>(false);
+
+  const onMouseEnter = (event: React.MouseEvent) => {
+    setIsMouseOver(true);
+    if (props.onMouseEnter != null) {
+      props.onMouseEnter(event);
+    }
+  };
+  const onMouseLeave = (event: React.MouseEvent) => {
+    setIsMouseOver(false);
+    if (props.onMouseLeave != null) {
+      props.onMouseLeave(event);
+    }
+  };
+
+  return React.cloneElement(isMouseOver ? props.hoveredIcon : props.icon, {
+    ...props,
+    onMouseEnter,
+    onMouseLeave,
+  });
+}
+export default {};
--- a/frontend/javascripts/oxalis/view/left-border-tabs/layer_settings_tab.tsx
+++ b/frontend/javascripts/oxalis/view/left-border-tabs/layer_settings_tab.tsx
@@ -8,6 +8,7 @@ import {
   WarningOutlined,
   PlusOutlined,
   VerticalAlignMiddleOutlined,
+  ToolOutlined,
 } from "@ant-design/icons";
 import { connect } from "react-redux";
 import React from "react";
@@ -49,6 +50,7 @@ import {
   getResolutionInfo,
   getResolutions,
   getVisibleSegmentationLayer,
+  getSegmentationLayers,
 } from "oxalis/model/accessors/dataset_accessor";
 import { getMaxZoomValueForResolution } from "oxalis/model/accessors/flycam_accessor";
 import {
@@ -124,6 +126,7 @@ type State = {
   // is shown for that VolumeTracing
   volumeTracingToDownsample: VolumeTracing | null | undefined;
   isAddVolumeLayerModalVisible: boolean;
+  preselectSegmentationLayerIndex: number | null | undefined;
   layerToMergeWithFallback: APIDataLayer | null | undefined;
 };
 
@@ -132,6 +135,7 @@ class DatasetSettings extends React.PureComponent<DatasetSettingsProps, State> {
   state: State = {
     volumeTracingToDownsample: null,
     isAddVolumeLayerModalVisible: false,
+    preselectSegmentationLayerIndex: 0,
     layerToMergeWithFallback: null,
   };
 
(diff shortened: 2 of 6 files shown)


frontend/javascripts/oxalis/view/left-border-tabs/modals/add_volume_layer_modal.tsx
```diff
@@ -84,14 +84,18 @@ export default function AddVolumeLayerModal({
   dataset,
   onCancel,
   tracing,
+  preselectedLayerIndex,
+  showLayerSelectionDisabled,
```
philippotto, line 88: I would name this `disableLayerSelection`.

frontend/javascripts/oxalis/view/left-border-tabs/modals/add_volume_layer_modal.tsx
```diff
@@ -84,14 +84,18 @@ export default function AddVolumeLayerModal({
...
 }: {
   dataset: APIDataset;
   onCancel: () => void;
   tracing: Tracing;
+  preselectedLayerIndex: number | null | undefined;
+  showLayerSelectionDisabled: boolean | null | undefined;
```
philippotto, line 94:
```suggestion
  preselectedLayerIndex: number | undefined;
  showLayerSelectionDisabled: boolean | undefined;
```

should be enough. the reason the `| null | undefined` is scattered through the entire code base is the automatic flow-to-typescript conversion and subtle differences between the semantics of their syntax. in the longterm, I'd rather get rid of this :)

frontend/javascripts/components/hover_icon_button.tsx
```diff
@@ -0,0 +1,32 @@
+import { ButtonProps } from "antd";
+import * as React from "react";
+const { useState } = React;
+
+export type HoverButtonProps = Omit<ButtonProps, "icon"> & {
+  icon: React.ReactElement<any>;
+  hoveredIcon: React.ReactElement<any>;
+};
+
+export function HoverIconButton(props: HoverButtonProps) {
```
philippotto, line 10: nice abstraction :+1:

frontend/javascripts/oxalis/view/left-border-tabs/layer_settings_tab.tsx
```diff
@@ -924,14 +979,34 @@ class DatasetSettings extends React.PureComponent<DatasetSettingsProps, State> {
...
                 Add Volume Annotation Layer
               </Button>
             </Row>
           </>
         ) : null}
 
+        {this.props.tracing.restrictions.allowUpdate && canBeMadeHybrid ? (
+          <Row justify="center" align="middle">
+            <Button
+              onClick={this.handleConvertToHybrid}
```
philippotto, line 998: Since this button now only has the "add skeleton layer" behavior, I'd rename `handleConvertToHybrid` to `addSkeletonAnnotationLayer`. Within that function, `convertToHybridTracing` can stay as is.

frontend/javascripts/oxalis/view/left-border-tabs/layer_settings_tab.tsx
```diff
@@ -894,9 +932,21 @@ class DatasetSettings extends React.PureComponent<DatasetSettingsProps, State> {
...
+      preselectedSegmentationLayerIndex: null,
     });
   };
 
+  handleConvertToHybrid = async () => {
+    await Model.ensureSavedState();
+    const maybeSegmentationLayer = getVisibleSegmentationLayer(Store.getState());
+    await convertToHybridTracing(
+      this.props.tracing.annotationId,
+      maybeSegmentationLayer != null ? maybeSegmentationLayer.name : null,
```
philippotto, line 945: if a skeleton layer is missing, this means that a volume layer already has to exist (otherwise, this wouldn't be an annotation). therefore, the converToHybridTracing function only needs to add a skeleton layer which is why the `fallbackLayerName` can simply be null. Then, there's no need to define `maybeSegmentationLayer`.

frontend/javascripts/oxalis/view/left-border-tabs/layer_settings_tab.tsx
```diff
@@ -457,6 +475,26 @@ class DatasetSettings extends React.PureComponent<DatasetSettingsProps, State> {
...
+              <Tooltip
+                title="Make this segmentation editable by adding a Volume Annotation Layer."
+                placement="left"
+              >
+                <HoverIconButton
+                  icon={<LockOutlined />}
+                  hoveredIcon={<UnlockOutlined />}
+                  onClick={() => {
+                    const segmentationLayers = getSegmentationLayers(dataset);
+                    const segmentationLayerIndex = segmentationLayers.indexOf(layer);
```
philippotto, line 488: I think, there's a bug here, because the `NewVolumeLayerSelection` component uses the index to access into  `const availableSegmentationLayers = _.differenceWith(segmentationLayers, volumeTracingLayers);` (as defined in `AddVolumeLayerModal`). Therefore, there are cases where the index will be incorrect.

To avoid that there are two places which depend on the semantics of the index, I'd rather pass a `preselectedSegmentationLayerName` instead of an `...Index`. Then, the `NewVolumeLayerSelection` component can find the correct index using the name itself.

philippotto (changes requested):
Cool stuff :+1: See my comments :)

philippotto (approved):
Great :+1: Works well for me. I don't think you need it to test it with another DS, since the new by-name logic should be way safer than the by-index logic.

philippotto:
But please update the changelog before merging ;)

PR #6331, opened 2022-07-13
Use coarser zoomStep for histogram clipping

### LINKED ISSUE #6115: Select coarser mag for automatic histogram clipping?
As discussed in https://github.com/scalableminds/webknossos/pull/6113#pullrequestreview-910653041 the automatic histogram clipping heavily depends on the currently rendered mag, especially if the datasets are very noisy.
We may want to use some sort of de-noising or just not ever use mag1 for the automatic histogram clipping.

Related: #5163

Dagobert42:
- https://usecoarsermagforhistogramclipping.webknossos.xyz

### Steps to test:
- clip a histogram

### Issues:
- fixes #6115 

------
- [x] Ready for review

### DIFF
--- a/frontend/javascripts/oxalis/api/api_latest.ts
+++ b/frontend/javascripts/oxalis/api/api_latest.ts
@@ -71,11 +71,12 @@ import {
   getResolutionInfo,
   getVisibleSegmentationLayer,
   getMappingInfo,
+  getMaxZoomStep,
 } from "oxalis/model/accessors/dataset_accessor";
 import {
   getPosition,
   getRotation,
-  getRequestLogZoomStep,
+  getMaximumZoomForAllResolutionsFromStore,
 } from "oxalis/model/accessors/flycam_accessor";
 import {
   loadAdHocMeshAction,
@@ -1349,7 +1350,7 @@ class DataApi {
     return this.cutOutCuboid(buckets, bbox, elementClass, resolutions, zoomStep);
   }
 
-  async getViewportData(viewport: OrthoView, layerName: string) {
+  async getViewportDataForHistogram(viewport: OrthoView, layerName: string) {
     const state = Store.getState();
     const [curX, curY, curZ] = dimensions.transDim(
       dimensions.roundCoordinate(getPosition(state.flycam)),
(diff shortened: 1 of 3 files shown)


daniel-wer (changes requested):
Nice work, but please see my suggestions :)

frontend/javascripts/oxalis/api/api_latest.ts
```diff
@@ -1367,7 +1363,12 @@ class DataApi {
       V3.add([curX, curY, curZ], [halfViewportExtentX, halfViewportExtentY, 1]),
       viewport,
     );
-    const resolutionIndex = getRequestLogZoomStep(state);
+
+    // Find a viable resolution to compute the histogram on
+    const layer = getLayerByName(state.dataset, layerName);
+    const resolutionInfo = getResolutionInfo(layer.resolutions);
+    const maybeResolutionIndex = 2;
```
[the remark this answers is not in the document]
Dagobert42: Yes thank you for the suggestion! This was exactly what was missing to make it work.

frontend/javascripts/oxalis/api/api_latest.ts
```diff
@@ -1367,7 +1367,14 @@ class DataApi {
...
       viewport,
     );
-    const resolutionIndex = getRequestLogZoomStep(state);
+
+    // Find a viable resolution to compute the histogram on
+    const layer = getLayerByName(state.dataset, layerName);
+    const resolutionInfo = getResolutionInfo(layer.resolutions);
+    // Ideally, we want to avoid resolutions 1 and 2 to keep
+    // the amount of data that has to be loaded small
+    const maybeResolutionIndex = Math.max(2, getRequestLogZoomStep(state));
```
philippotto, line 1376: How about changing this to:

```suggestion
    const maybeResolutionIndex = Math.max(2, getRequestLogZoomStep(state) + 1);
```

That way, the histogram clipping will be way faster (factor 8 in isotropic cases), since it would load the fallback data in most cases. Otherwise, all best-quality buckets for the current mag (if it's greater than mag 2) would need to be awaited which can take some seconds.

Dagobert42: I see :) Thanks!

frontend/javascripts/oxalis/model/sagas/clip_histogram_saga.ts
```diff
@@ -17,13 +18,20 @@ function onThresholdChange(layerName: string, [firstVal, secVal]: [number, numbe
...
 }
 
-async function getClippingValues(layerName: string, thresholdRatio: number = 0.005) {
+async function getClippingValues(layerName: string, thresholdRatio: number = 0.0001) {
   const { elementClass } = getLayerByName(Store.getState().dataset, layerName);
   const [TypedArrayClass] = getConstructorForElementClass(elementClass);
+
+  // Find a viable resolution to compute the histogram on
+  // Ideally, we want to avoid resolutions 1 and 2 to keep
+  // the amount of data that has to be loaded small
```
daniel-wer, line 27:
```suggestion
  // the amount of data that has to be loaded small and
  // to de-noise the data
```

daniel-wer (approved):
Awesome, thanks for incorporating the feedback! Works well :+1:

PR #6334, opened 2022-07-14
Disable navbar menu overflow due to a bug

daniel-wer:
Disabling the built-in overflow behavior is not ideal, but I'd argue it's the better solution for now, because:
- For the annotation view (where there are lots of elements in the navbar) we have custom-built overflow/collapsing behavior which continues to work
- Having such a narrow browser window that the built-in overflow behavior would be needed should be rather unusual

- https://navbaroverflow.webknossos.xyz

### Steps to test:
- Open the dashboard, the help button shouldn't be behind the overflow indicator

------
- [x] Ready for review


daniel-wer:
I just noticed in a last-minute test that the menus in the navbar are no longer clickable or at least they don't open on click :see_no_evil:

PR #6337, opened 2022-07-15
Allow dataset managers to set dataset team permissions

fm3:
Dataset managers are now allowed to change dataset team permissions for all teams they are a member of

- https://datasetmanagersetteams.webknossos.xyz

### Steps to test:
- create additional user
- make them dataset manager
- see that they can set dataset team permissions for teams they are a member of

### Issues:
- fixes https://discuss.webknossos.org/t/problem-about-public-dataset-result-in-webkonossos/1724/5

------
- [x] Ready for review

### DIFF
--- a/CHANGELOG.unreleased.md
+++ b/CHANGELOG.unreleased.md
@@ -19,6 +19,7 @@ For upgrade instructions, please check the [migration guide](MIGRATIONS.released
 ### Changed
 - Merged the "Shared Annotations" tab into the "Annotations" tab in the user's dashboard. If annotations are shared with you, you can see them in your dashboard. The table can be filtered by owner if you prefer to see only your own annotations. [#6230](https://github.com/scalableminds/webknossos/pull/6230)
 - Changed the name of the volume annotation layer in tasks to be "Volume" instead of "Volume Layer". [#6321](https://github.com/scalableminds/webknossos/pull/6321)
+- Dataset managers are now allowed to change dataset team permissions for all teams they are a member of. [#6336](https://github.com/scalableminds/webknossos/pull/6336)
 
 ### Fixed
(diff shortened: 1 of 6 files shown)


jstriebel (commented):
Nice! The backend code LGTM, but when testing it, I cannot see any teams in the frontend (being dataset-manager and being a member of teams). In the dropdown only teams appear when I'm a team manager there.

That's probably a frontend issue, I assume the following parts need `allowNonEditableTeams=true` or sth. similar (seems like false is an implicit default?):
* https://github.com/scalableminds/webknossos/blob/a05eac53c1535e57b0cf266088c4e6fad50dd595/frontend/javascripts/dashboard/dataset/dataset_settings_sharing_tab.tsx#L47
* https://github.com/scalableminds/webknossos/blob/a05eac53c1535e57b0cf266088c4e6fad50dd595/frontend/javascripts/admin/dataset/dataset_upload_view.tsx#L659

fm3:
Thanks for having a look! @philippotto Do you think this could easily be adapted in the frontend?

philippotto:
> @philippotto Do you think this could easily be adapted in the frontend?

Sure, I just added the property. However, I'm wondering what the remaining of the relevance `isEditable=false` behavior is. Now, that behavior is only used by `open_tasks_report_view.tsx` and `project_progress_report_view.tsx`. This probably means one can only see progress stats for tasks and projects of a team if one is a team member there. Probably this makes sense? Otherwise, one could simply remove the entire isEditable parameter..

philippotto:
For reference, see our discussion here: https://scm.slack.com/archives/C5AKLAV0B/p1658239442844309

jstriebel (approved):
Nice! Just checked it again, everything works now as described in the changes, I think that's fine for now :+1: Frontend changes also LGTM, as far as I can tell.

PR #6344, opened 2022-07-21
adds redirect to workflow frontend

normanrz:
Adds a redirect from `/workflows` to `https://workflows.voxelytics.com/workflows` (with sub paths). With this we can already use webKnossos-hosted URLs to point to workflow reports, before it is integrated into webKnossos itself.

------
- [x] Ready for review

### DIFF
--- a/frontend/javascripts/router.tsx
+++ b/frontend/javascripts/router.tsx
@@ -1,5 +1,5 @@
 // @ts-expect-error ts-migrate(2305) FIXME: Module '"react-router-dom"' has no exported member... Remove this comment to see the full error message
-import { ContextRouter } from "react-router-dom";
+import type { ContextRouter } from "react-router-dom";
 import { Redirect, Route, Router, Switch, useLocation } from "react-router-dom";
 import { Layout, Alert } from "antd";
 import { connect } from "react-redux";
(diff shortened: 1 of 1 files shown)


frontend/javascripts/router.tsx
```diff
@@ -100,6 +100,16 @@ function PageNotFoundView() {
   );
 }
 
+function RedirectToWorkflowViewer() {
+  const location = useLocation();
+
+  useEffect(() => {
+    windowLocation.assign(`https://workflows.voxelytics.com${location.pathname}${location.search}`);
```
philippotto, line 107: The final URL will be `https://workflows.voxelytics.com/workflows/...` then, right? It's a bit redundant, but probably acceptable. Alternatively, you could trim the leading `workflows` from the pathname.

philippotto (approved):
lgtm :+1:

frontend/javascripts/router.tsx
normanrz: Well the final URL will be `https://webknososs.org/workflows` ;-) I'd like to keep the URL schemes as close as possible, which introduces this (hopefully temporary) redundancy.

philippotto: ok, makes sense :+1:

PR #6345, opened 2022-07-21
improve histogram slider tooltip readability

### LINKED ISSUE #6343: Histogram slider tooltip values shown in e+ notation too soon
In an uint16 dataset, the values in the histogram slider tooltips are shown like 3,1868e+4 – I would instead expect 31868.
It may make sense to switch to e+ notation eventually, but only for larger values (or really small float values). Let’s discuss that. I’d argue that uint16 values should always be displayed in normal decimal notation.

fm3:
Histogrammschiebereglerwerkzeughinweislesbarkeitsverbesserung!

I noticed that uint64 brightness values in the histogram slider tooltip were shown with e notation, which was hard to read. I tweaked the condition for this a bit.

- https://slidertooltipreadable.webknossos.xyz

### Steps to test:
- play with the histogram sliders in the data types dataset, should feel natural

### Issues:
- fixes #6343 

------
- [x] Ready for review

### DIFF
--- a/frontend/javascripts/oxalis/view/left-border-tabs/histogram_view.tsx
+++ b/frontend/javascripts/oxalis/view/left-border-tabs/histogram_view.tsx
@@ -199,7 +199,9 @@ class Histogram extends React.PureComponent<HistogramProps, HistogramState> {
   };
 
   tipFormatter = (value: number) =>
-    value > 10000 ? value.toExponential() : roundTo(value, this.getPrecision()).toString();
+    value >= 100000 || (value < 0.001 && value > -0.001)
+      ? value.toExponential()
+      : roundTo(value, this.getPrecision()).toString();
 
   // eslint-disable-next-line react/sort-comp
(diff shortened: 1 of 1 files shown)


frontend/javascripts/oxalis/view/left-border-tabs/histogram_view.tsx
```diff
@@ -199,7 +199,9 @@ class Histogram extends React.PureComponent<HistogramProps, HistogramState> {
   };
 
   tipFormatter = (value: number) =>
-    value > 10000 ? value.toExponential() : roundTo(value, this.getPrecision()).toString();
+    value >= 100000 || (value < 0.001 && value > -0.001)
```
philippotto, line 202:
```suggestion
    value >= 100000 || (value < 0.001 && value > -0.001 && value !== 0)
```

otherwise, 0 renders as `0e+0` :)

philippotto (approved):
see my one comment :)

PR #6350, opened 2022-07-25
Upgrade ThreeJS and use WebGL 2

### LINKED ISSUE #5745: Consider using WebGL 2
## Detailed Description
Since iOS 15 supports webgl 2, we can consider switching to it. First, we should find out the impact of this (performance gains, code simplification?). Second, we should estimate the effort (including testing for iPad).

philippotto:
### Summary
- Enable WebGL2
- Update ThreeJS to r137 to fix issues with RG textures
- Replace GLSL validation package to support webgl 2
- Update node to 16 since new glsl package requires this
- Remove shader rendering spec due to incompatibilities with node 16 (see https://app.circleci.com/pipelines/github/scalableminds/webknossos/7286/workflows/e17826f1-461a-481c-ba60-246174b32df5/jobs/18622/parallel-runs/0/steps/0-109). there was only one spec, anyway, with very low impact. the e2e screenshot tests serve us better here.

The screenshots have slightly thicker borders (the red, green, blue ones) now apparently:
old: 
image

new: 
image

I'm not sure whether this is only in the screenshot tests or for the normal browser setup, too. However, I don't think that this is a blocker. Especially, because the pixel difference is so small that the screenshot tests pass.

- https://webgl.webknossos.xyz

### Todo:
- [x] clarify what cross sections are and check whether they work? (see cube.ts)

### Steps to test:
- [X] test on iPad (I already did this via browserstack and the rendering seems to work)
- [X] run nightly screenshot test against this branch
Also, try to test wk rather broad regarding rendering, but the following were changed during this PR which is why they should be tested with extra care:
- crosshair
	- in ortho
	- in flight mode
- mesh smoothing / computeVertexNormals
- Quaternion usage in 3D viewport / use the rotate buttons
- rendering of "cubes"

### Issues:
- fixes #5745 

------
- [x] Ready for review


frontend/javascripts/oxalis/controller/scene_controller.ts
```diff
@@ -235,7 +232,7 @@ class SceneController {
     }
 
     geometry.computeVertexNormals();
-    geometry.computeFaceNormals();
+    // geometry.computeFaceNormals();
```
philippotto, line 235: is there a noticeable setback due to this change?

frontend/javascripts/oxalis/geometries/cube.ts
```diff
@@ -99,7 +100,9 @@ class Cube {
       vec(min[0], min[1], max[2]),
       vec(min[0], max[1], max[2]),
       vec(min[0], max[1], min[2]),
-    ];
+    ]);
+
+    // todo: clarify what the cross sections are and fix them probably?
```
philippotto, line 105: todo

philippotto: @daniel-wer you probably have some insight here :)

daniel-wer: As far as I know, the cross sections simulate that the bounding box is a cuboid although it is actually only a wireframe of a cuboid. This is important so that the bounding box can be seen whenever one of its *faces* is visible in one of the viewports and not only whenever one of its *edges* is visible.
This is done by having one cross section per viewport which is a simple rectangle with 4 edges and moving it in the respective 3rd dimension whenever the flycam position changes.
Currently, this doesn't seem to work on the deployed branch.
Hope this makes sense, otherwise let's have a quick call :)

daniel-wer:
Quick note because I encountered this during a test:
- [ ] Downloading a computed ad-hoc mesh from the Segments list fails with ` TypeError: Da.Geometry is not a constructor`

frontend/javascripts/oxalis/geometries/cube.ts
philippotto: thank you, makes total sense :+1:

frontend/javascripts/libs/UpdatableTexture.ts
```diff
@@ -1,106 +1,98 @@
...
-  const imageData = ctx.createImageData(1, 1);
-  this.image = imageData;
-  this.magFilter = magFilter !== undefined ? magFilter : THREE.LinearFilter;
-  this.minFilter = minFilter !== undefined ? minFilter : THREE.LinearMipMapLinearFilter;
-  this.generateMipmaps = false;
-  this.flipY = false;
-  this.unpackAlignment = 1;
-  this.needsUpdate = true;
-}
+class UpdatableTexture extends THREE.Texture {
```
daniel-wer, line 4: Good job, cleaning this up and making it type-check :1st_place_medal:

daniel-wer (commented):
Awesome, great milestone, thank you for your persistence in getting this to work! Code mostly LGTM, I'll do some proper testing and will report back, afterwards :)

frontend/javascripts/oxalis/geometries/cube.ts
philippotto: should be fixed now :tada:

frontend/javascripts/oxalis/model/bucket_data_handling/texture_bucket_manager.ts
```diff
@@ -30,10 +31,12 @@ import type { ElementClass } from "types/api_flow_types";
...
-// At the moment, we only store one float f per bucket.
-// If f >= 0, f denotes the index in the data texture where the bucket is stored.
-// If f == -1, the bucket is not yet committed
-// If f == -2, the bucket is not supposed to be rendered. Out of bounds.
+// At the moment, we store two floats per bucket.
+// The first value f denotes the following:
+//   If f >= 0, f denotes the index in the data texture where the bucket is stored.
+//   If f == -1, the bucket is not yet committed
+//   If f == -2, the bucket is not supposed to be rendered. Out of bounds.
+// The second value v denotes the magnification value of the addressed bucket.
```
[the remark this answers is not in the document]
philippotto: it was wrong (outdated) before :)

frontend/javascripts/oxalis/controller/renderer.ts
```diff
@@ -13,13 +13,15 @@ function getRenderer() {
   const renderCanvasElement = document.getElementById("render-canvas");
   renderer =
     renderCanvasElement != null
-      ? new THREE.WebGLRenderer({
+      ? // Create a WebGL2 renderer
```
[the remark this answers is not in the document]
philippotto: no, it's meant to clarify that `WebGLRenderer` indeed refers to a webgl2 renderer :see_no_evil:

daniel-wer:
Notes from testing:
- [ ] The connectome_file_test_dataset prediction layer (uint24!) seems to have gaps in z, see screenshot of YZ viewport (shouldn't this have been catched by the screenshot tests? :thinking:)
  connectome_file_test_dataset_gaps
- [ ] Subjectively, I encounter missing buckets in the viewports much more often than usual, sometimes on page load or when moving. Could also be that the server is under heavy load :shrug: (Actually, I encountered this on master as well, so the culprit seems to be the [nine_float_layer_dataset](https://master.webknossos.xyz/datasets/sample_organization/nine_float_layer_dataset/view) or the dev server, not an issue of this PR)
  missing_buckets
  missing_buckets_2

frontend/javascripts/oxalis/controller/renderer.ts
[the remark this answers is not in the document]
daniel-wer: Ah makes sense, I think that's fine :+1:

philippotto:
> The brush preview sphere seems to be much brighter than usual. Ist this by design?

No, this was not intentional. I think some default behavior regarding the alpha value changed. I pushed a fix, which improves the color. I think, it's not 100% the same as on master, but it should be fine (and it's probably not worth the hassle to achieve the exact same output as before).

>  The connectome_file_test_dataset prediction layer (uint24!) seems to have gaps in z, see screenshot of YZ viewport (shouldn't this have been catched by the screenshot tests? :thinking: )

good catch :+1: The screenshot test for uint24 seems to test a quite unfortunate position (see image below). I'll look into this.

image

philippotto:
@daniel-wer I think, the PR is ready for a second look. I fixed all the issues that are relevant to this PR and the screenshot tests should be good now, too. I'll probably have our team do some additional testing, too, to get a good coverage of different systems (hardware, OS, browser etc).

.circleci/not-on-master.sh
```diff
@@ -1,8 +1,8 @@
 #!/usr/bin/env bash
 set -Eeuo pipefail
 
-if [ "${CIRCLE_BRANCH}" == "master" ]; then
+# if [ "${CIRCLE_BRANCH}" == "master" ]; then
```
daniel-wer, line 4: Reminder to re-enable this before merging

frontend/javascripts/oxalis/geometries/cube.ts
```diff
@@ -99,44 +103,44 @@ class Cube {
...
-      // @ts-expect-error ts-migrate(2339) FIXME: Property 'verticesNeedUpdate' does not exist on ty... Remove this comment to see the full error message
-      mesh.geometry.verticesNeedUpdate = true;
+      mesh.geometry.attributes.position.needsUpdate = true;
     }
 
     this.initialized = true;
-    this.updatePosition(getPosition(Store.getState().flycam));
+    this.updatePositionForCrossSections(getPosition(Store.getState().flycam));
+    this.crossSections[OrthoViews.PLANE_XZ].geometry.computeBoundingSphere();
+    this.crossSections[OrthoViews.PLANE_XZ].geometry.attributes.position.needsUpdate = true;
```
daniel-wer, line 138: Is this really needed, here and if so why only for the XZ plane?

daniel-wer (approved):
Nice fixes, everything:tm: working as intended now :+1:

frontend/javascripts/oxalis/geometries/cube.ts
philippotto: good catch. this wasn't intentional, as this is already done by `updatePositionForCrossSections` for all planes.

Issue #6297, opened 2022-06-23
Add layer name to export tiff job entry in jobs table


MichaelBuessemeyer:
Currently, the entry of an export tiff job does not contain information about which layer was exported. This should be added.

## Detailed Description

When triggering multiple export jobs of the same dataset, the user cannot derive which export job exported which layer. Thus the layer name should be added.

## Context

* [x] Specific to long-running jobs (set `jobsEnabled=true` in `application.conf`)

PR #6355, opened 2022-07-27
Add layer name to jobs table entry of tiff export job

MichaelBuessemeyer:
This pr adds the layer name to every version of the export tiff job. Previously the names of volume layers were the fallback layers' names if they existed. See
image

Now the name of the volume layer is used in every case. Even when the volume layer has no fallback.

image

- https://___.webknossos.xyz

### Steps to test:
- I suggest testing locally as a wk-worker is needed for testing.
- You need a running instance of the wk worker that is registered in the database (`yarn enable-jobs`)
- Open an annotation of any dataset with at least one color layer and a segmentation layer.
- create a small bounding box that has only 1 slice in z direction. The smaller the bounding box the faster the jobs are executed.
- Then use the download button in the bounding box tab and start a export job for one color layer and one volume layer without having drawn anything on that volume layer.
- Close the export modal, draw something in the just exported volume layer within the bounds of the bounding box.
- Create a new annotation layer without a fallback layer.
- Also draw something on the new volume layer without fallback within the bounding box. 
- Trigger another export job for volume layer with fallback layer and the volume layer without fallback layer.
- Go to the jobs page. Each export tiff job entry should have the exported's layer's name in it and if it is a volume annotation layer that was exported, there should be a link pointing to the annotation. Else the link should link to the dataset in view mode.

### Issues:
- fixes #[6297](https://github.com/scalableminds/webknossos/issues/6297)

------
- [x] Ready for review

### DIFF
--- a/frontend/javascripts/admin/admin_rest_api.ts
+++ b/frontend/javascripts/admin/admin_rest_api.ts
@@ -1003,6 +1003,7 @@ export async function getJobs(): Promise<APIJob[]> {
     datasetName: job.commandArgs.dataset_name,
     organizationName: job.commandArgs.organization_name,
     layerName: job.commandArgs.layer_name || job.commandArgs.volume_layer_name,
+    annotationLayerName: job.commandArgs.annotation_layer_name,
     boundingBox: job.commandArgs.bbox,
     exportFileName: job.commandArgs.export_file_name,
     tracingId: job.commandArgs.volume_tracing_id,
--- a/frontend/javascripts/admin/job/job_list_view.tsx
+++ b/frontend/javascripts/admin/job/job_list_view.tsx
@@ -135,16 +135,18 @@ class JobListView extends React.PureComponent<Props, State> {
     } else if (job.type === "export_tiff" && job.organizationName && job.datasetName) {
       const labelToAnnotationOrDataset =
         job.annotationId != null ? (
-          <Link to={`/annotations/${job.annotationId}`}>annotation of {job.datasetName}</Link>
+          <Link to={`/annotations/${job.annotationId}`}>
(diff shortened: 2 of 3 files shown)


MichaelBuessemeyer:
First thing:
@daniel-wer I do not know whether you have an installed version of the wk-worker locally. If you do not have one, please ask Philipp to review this pr. The review should be pretty fast and setting up the wk-worker locally only to review this pr is not worth it in my opinion.

I am unsure about the formulation of the export job entry in the table (see the 2nd screenshot in the pr description). The two "of"s in the sentence sound quirky. Do you have a better formulation @daniel-wer?

daniel-wer:
Unfortunately, I don't have a working wk-worker setup, locally, so I'll reassign the review.

daniel-wer:
> I am unsure about the formulation of the export job entry in the table (see the 2nd screenshot in the pr description). The two "of"s in the sentence sound quirky. Do you have a better formulation @daniel-wer?

Hmm, I would switch the of and from, move "layer" in front of the label and add "dataset" in front of the dataset name, so that it reads "Tiff export of layer {layerLabel} from dataset ...". With the annotation part it could then read "Tiff export of layer {layerLabel} from annotation of dataset ...".

philippotto (approved):
Looks good to me :+1: Since your testing steps are so thorough and backed with a screenshot, I think, another testing round is not necessary. Especially, since the changes are unlikely to break anything mission critical. Let me know in case you disagree. Otherwise, feel free to merge :)

MichaelBuessemeyer:
> I think, another testing round is not necessary.

Ok, thanks and let go :shipit:

Issue #6238, opened 2022-05-25
Keep name for editable segmentation


normanrz:
Assume I have a dataset with a segmentation layer called `segmentation`. Now, I create an annotation for this dataset. Magically, the layer is renamed to `Volume`. I find that quite confusing and would recommend to keep the name as is: `segmentation`.

PR #6357, opened 2022-07-29
Allow base layer name as new volume layer name

Dagobert42:
### Steps to test:
- create new volume layer based on segmentation layer and name it segmentation layer

------
- [x] Ready for review


Dagobert42:
It looks like we only call this through the add volume layer modal. Is that correct?

philippotto (approved):
great! changes look good and testing also showed that this works :+1: 

please wait with merging until monday, though :)

philippotto:
> It looks like we only call this through the add volume layer modal. Is that correct?

The code you changed is within `AddVolumeLayerModal` and only used for the modal. `validateReadableLayerName` itself is also used when renaming an existing layer, but there's no need to change anything there.

normanrz:
This does not fix #6238, does it?

philippotto:
> This does not fix #6238, does it?

You're right, this is independent of each other.

Issue #6356, opened 2022-07-28
Get rid of return statements in scala


fm3:
Got my first instances of `NonLocalReturn` and thus learned that return in scala is generally considered harmful. https://tpolecat.github.io/2014/05/09/return.html – I’d say it is worth taking the time to remove all `return` statements from the scala backend. Most are just if-guards that can easily be changed to if/else

Edit 2024: three instances of `return` remain, those are a little harder to rewrite

PR #6358, opened 2022-07-29
Remove most return statements from scala code

fm3:
return in scala is generally harmful, compare article at https://tpolecat.github.io/2014/05/09/return.html

This PR removes most return statements from our code
 - except those in the jzarr and datastore.storage packages (because PR #6335  does that
 - except those that I did not know how to convert without rewriting a lot of code (DataFinder, IsosurfaceService, CumsumParser)

It is recommended to hide whitespace in the diff view :)

- https://fewerreturnsinscala.webknossos.xyz

### Steps to test:
- create task, trace some, download + upload volume data, should still work as expected

### Issues:
- contributes to #6356 

------
- [x] Ready for review


leowe (approved):
LGTM 👍

PR #6362, opened 2022-08-01
Remember the forcefully disabled tool and reenable it if possible

### LINKED ISSUE #6293: Remember active tool during toggle segmentation
When brushing/erasing in a volume annotation, it is often helpful to temporarily hide the segmentation layer (shortcut 3) for a clear view of the raw data. When un-hiding it shortly thereafter, the user is stuck in the move tool, for the brush/erase tools were temporarily disabled. It would be great if the tool that was previously active could be remembered and if I could continue brushing after hitting “3” twice quickly to check the raw data.

MichaelBuessemeyer:
This PR tackles the problem described in #6293. When shortly disabling the segmentation layer via shortcut `3` to only see the color data the volume tool was deselected. The user then had to manually reselect the previous volume tool. To speed up annotating, the forcefully disabled tool is now remembered and reselected again if it is no longer disabled and no other tool has been selected after this tool was disabled.

- https://rememberactivetool.webknossos.xyz

### Steps to test:
- Open an annotation with a segmentation layer.
- Select the brush tool and hit `3` on the keyboard to disable the segmentation layer.
- The move tool should now be selected as the brush is disabled.
- Hit `3` again and the brush tool should now be automatically selected.
- Do the same for the other volume tools. This should also work.
- Now select the brush tool, hit `3`, and then select the skeleton tool. Hitting `3` again should not select the brush tool.
-  At last, select the brush tool, hit `3`, select the skeleton tool, and then select the move tool again. Hitting `3` again should not select the brush tool as the user intentionally changed the tools in between disabling and enabling the segmentation layer.

### Issues:
- fixes #6293

------
- [x] Ready for review

### DIFF
--- a/CHANGELOG.unreleased.md
+++ b/CHANGELOG.unreleased.md
@@ -12,6 +12,7 @@ For upgrade instructions, please check the [migration guide](MIGRATIONS.released
 
 ### Added
 - The owner of an annotation can allow other users, who may see the annotation, to also edit it. Note that parallel writes are not supported and would lead to conflicts. [#6236](https://github.com/scalableminds/webknossos/pull/6236)
+- WebKnossos now remembers the tool that was active in between enabling and disabling the segmentation layer. [#6362](https://github.com/scalableminds/webknossos/pull/6362)
 - Segmentation layers which were not previously editable now show an (un)lock icon button which shortcuts to the Add Volume Layer modal with the layer being preselected. [#6330](https://github.com/scalableminds/webknossos/pull/6330)
 - The NML file in volume annotation download now includes segment metadata like names and anchor positions. [#6347](https://github.com/scalableminds/webknossos/pull/6347)
 
--- a/CHANGELOG.unreleased.md
+++ b/CHANGELOG.unreleased.md
@@ -12,7 +12,7 @@ For upgrade instructions, please check the [migration guide](MIGRATIONS.released
 
 ### Added
 - The owner of an annotation can allow other users, who may see the annotation, to also edit it. Note that parallel writes are not supported and would lead to conflicts. [#6236](https://github.com/scalableminds/webknossos/pull/6236)
-- WebKnossos now remembers the tool that was active in between enabling and disabling the segmentation layer. [#6362](https://github.com/scalableminds/webknossos/pull/6362)
+- WebKnossos now remembers the tool that was active in between disabling and enabling the segmentation layer. [#6362](https://github.com/scalableminds/webknossos/pull/6362)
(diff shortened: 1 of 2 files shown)


MichaelBuessemeyer:
(Should probably be implemented as a saga)

_Originally posted by @philippotto in https://github.com/scalableminds/webknossos/issues/6293#issuecomment-1162975446_

Philipp suggested implementing this in a saga. But upon reading the code it was far simpler to implement the reenabling behaviour directly next to the code that disables the active tool. Thus this solution is likely to have less code and also be more understandable as the disabling and reenabling code is at one location.
Thats why this feature is not implemented as a saga. ^^

CHANGELOG.unreleased.md
```diff
@@ -12,6 +12,7 @@ For upgrade instructions, please check the [migration guide](MIGRATIONS.released
 
 ### Added
 - The owner of an annotation can allow other users, who may see the annotation, to also edit it. Note that parallel writes are not supported and would lead to conflicts. [#6236](https://github.com/scalableminds/webknossos/pull/6236)
+- WebKnossos now remembers the tool that was active in between enabling and disabling the segmentation layer. [#6362](https://github.com/scalableminds/webknossos/pull/6362)
```
daniel-wer, line 15:
```suggestion
- WebKnossos now remembers the tool that was active in between disabling and enabling the segmentation layer. [#6362](https://github.com/scalableminds/webknossos/pull/6362)
```

frontend/javascripts/oxalis/view/action-bar/toolbar_view.tsx
```diff
@@ -546,7 +549,19 @@ export default function ToolbarView() {
...
+      activeTool === AnnotationToolEnum.MOVE
+    ) {
+      // Reenable the tool that was disabled before.
+      setLastForcefulDisabledTool(null);
+      Store.dispatch(setToolAction(lastForcefulDisabledTool));
+    } else if (activeTool !== AnnotationToolEnum.MOVE) {
+      // Forget the last disabled tool as another tool besides the move tool was selected.
+      setLastForcefulDisabledTool(null);
     }
   }, [activeTool, disabledInfoForCurrentTool]);
```
daniel-wer, line 566: For completeness sake, `lastForcefulDisabledTool` should be included in the dependencies here as well, right?

daniel-wer (approved):
I agree that this solution is quite elegant :)

There is one use case it doesn't solve which is if tools are disabled because the user zoomed out too far. However, since this is not mentioned in the original issue, I would say it's fine to merge this as-is :+1:

PR #6363, opened 2022-08-01
Fix Zarr Segmentation Layer json parsing

### LINKED ISSUE #6188: Cannot create annotation with Zarr Segmentation as Fallback Layer
Trying to create an annotation with volume layer with fallback layer. The fallback layer is a zarr segmentation layer.
Looks like `getFallbackLayer` in the AnnotationService fails because the suppoed fallback layer is not case-matched against SegmentationLayer

fm3:
When parsing the datasource-properties.json, wk previously failed to recognize zarr segmentation layers. They were instead parsed as ZarrDataLayers (discarding segmentation-specific fields like largestSegmentId). The `category` field was still "segmentation" (creating an invalid state, and leading the frontend to in turn correctly show the layer as a segmentation). The symptom was that you could not create an annotation with this layer as fallback segmentation.

- https://fixzarrsegmentationparsing.webknossos.xyz

### Steps to test:
- talk to me for a dataset with a zarr segmentation
- load this dataset in your wk, refresh datasets with button in dashboard (as admin) or wait 1min for next scheduled refresh
- create annotation for this dataset (segmentation as fallback is default)
- should work, should be able to brush.

### Issues:
- fixes #6188

------
- [x] Ready for review


leowe (approved):
LGTM :+1:

PR #6310, opened 2022-06-29
Zarr streaming editable mapping support

leowe:
### URL of deployed dev instance (used for testing):
- https://___.webknossos.xyz

### Steps to test:
- Go to a volume annotation with a mapping file in webknossos
- edit the segmentation and create a new segment
- request a bounding box that contains that segment, e.g. `http://localhost:9000/tracings/volume/zarr/e1d6f00c-84ad-494c-9134-a9deefa4ecc7/1/0.2.4.1?token=secretSampleUserToken`
- save response and check whether it contains the new segment, e.g. in python
```
import numpy as np

f = open("file.bin", "r")
a = np.fromfile(f, dtype=np.uint32)
```

### Issues:
- fixes #6225 

------
- [x] Needs datastore update after deployment
- [x] Ready for review

PR #6366, opened 2022-08-02
Cleanup zarr directory representation

jstriebel:
This PR
* simplifies the zarr directory html pages
* makes them compatible with python fsspec
* and ensures that the pages are used in the browser always with a trailing slash 

### URL of deployed dev instance (used for testing):
- https://htmldirectoryfsspeccompat.webknossos.xyz

### Steps to test:
I tested those myself, review should be enough:
- Go to https://htmldirectoryfsspeccompat.webknossos.xyz/data/zarr/sample_organization/ROI2017_wkw?token=…,
  and explore some folders (token needs to be re-added manually)
- For the python client:
```python
from upath import UPath

p = UPath(
    "https://htmldirectoryfsspeccompat.webknossos.xyz/data/zarr/sample_organization/ROI2017_wkw",
    client_kwargs={"headers": {"X-Auth-Token": "…"}}
)

p.is_dir()
# True

list(p.iterdir())
# [
#   HTTPPath('https://htmldirectoryfsspeccompat.webknossos.xyz/data/zarr/sample_organization/ROI2017_wkw/color'),
#   HTTPPath('https://htmldirectoryfsspeccompat.webknossos.xyz/data/zarr/sample_organization/ROI2017_wkw/segmentation')
# ]

list(p.iterdir())[0].is_dir()
# True
```

### Issues:
- fixes #

------

## TODO
- [x] Needs datastore update after deployment
- [x] Ready for review

PR #6367, opened 2022-08-02
Zarr private links

### LINKED ISSUE #6193: mapping and other parameters for zarr streaming

leowe:
- https://zarrprivatelinks.webknossos.xyz/

### Steps to test (frontend):

- create an annotation
- use the dropdown next to the save button and select "zarr links"
- create a link
- edit the expiration date
- remove the link
- etc

### Steps to test (backend):

**First step: Get access token**
- create a new annotation in the UI
- take the annotation ID (e.g. from the URL)
- call `POST localhost:9000/api/annotations/zarr_private` with a body like:
```
{
    "_annotation": "<annotationId>",
    "expirationDateTime": null
}
```
- in the response you get a body with the corresponding information, like
```
{
    "_id": {
        "id": "62ebca5ee401004f04276595"
    },
    "_annotation": {
        "id": "62ebc6cfe40100a402276591"
    },
    "accessToken": "aw-Kd5TVGosY4sTOuAFnJbNWRQSpho9o",
    "isDeleted": false
}
```
The relevant part here is the `accessToken`.

- You can similarly test the update, delete, and get routes

**Second step: Using the private link endpoints**
- There are all the endpoints for `zarr_private` that also exists for the "normal" zarr routes
- a good first candidate for testing is the `datasource-properties.json`, since it directly shows whether the requests to the tracing-store and the internal request on the datastore works, it also checks all the access rights and does the lookup from the accessToken to the annotation on webknossos
- request looks like this: `GET localhost:9000/data/zarr_private/<accessToken>/datasource-properties.json`
- You can also check the data route (`GET /zarr_private/:accessToken/:dataLayerName/:mag/:cxyz`), see #6310 for ideas on how to check this

### Issues:
- fixes #6193 

------
- [X] Updated [(unreleased) migration guide](../blob/master/MIGRATIONS.unreleased.md#unreleased) if applicable
- [x] Ready for review


conf/evolutions/084-zarr-private-links.sql
```diff
@@ -0,0 +1,24 @@
+-- Note that this evolution introduces constraints which may not be met by existing data. See migration guide for manual steps
+
+START TRANSACTION;
+
+CREATE TABLE webknossos.annotation_private_links
```
normanrz, line 5: I guess it should be like this?
```suggestion
CREATE TABLE webknossos.annotation_privateLinks
```

leowe: Hm. I don't know. The `annotation_privateLinks` version leads to weird Table names in the Scala code (`AnnotationPrivatelinks` instead of `AnnotationPrivateLinks`)

normanrz: What about `CREATE TABLE webknossos."annotation_privateLinks"`?

fm3: I’m also in favor of `annotation_privateLinks` (like norman said), that is our convention. The generated slick classes do have weird names, but they are only accessed in very few places, so I’d say it’s fine

webknossos-datastore/app/com/scalableminds/webknossos/datastore/jzarr/ZarrHeader.scala
```diff
@@ -50,6 +53,30 @@ case class ZarrHeader(
 object ZarrHeader {
   val FILENAME_DOT_ZARRAY = ".zarray"
 
+  def fromLayer(dataLayer: DataLayer, mag: Vec3Int): ZarrHeader = {
```
fm3, line 56: please add explaining comment that this is used for exposing webknossos layers as zarr via the api (otherwise we may get confused in the future thinking it is about zarr layers on disk)

webknossos-datastore/app/com/scalableminds/webknossos/datastore/jzarr/ZarrHeader.scala
```diff
@@ -90,5 +117,23 @@ object ZarrHeader {
...
+        "order" -> zarrHeader.order,
+        "chunks" -> zarrHeader.chunks,
+        "compressor" -> zarrHeader.compressor,
+        "filters" -> None,
+        "shape" -> zarrHeader.shape,
+        "dimension_seperator" -> zarrHeader.dimension_separator
+      )
+  }
+
+//  implicit val jsonFormat: OFormat[ZarrHeader] = Json.using[Json.WithDefaultValues].format[ZarrHeader]
```
fm3, line 138: this can go I guess

webknossos-datastore/app/com/scalableminds/webknossos/datastore/rpc/RPCRequest.scala
```diff
@@ -220,6 +220,7 @@ class RPCRequest(val id: Int, val url: String, wsClient: WSClient)
             s"RequestBody: '$requestBodyPreview'. Status ${response.status}. " +
             s"ResponseBody: '${response.body.take(100)}'")
       }
+      logger.debug(f"parsed: ${Json.parse(response.body)}")
```
fm3, line 223:
```suggestion
```

webknossos-datastore/app/com/scalableminds/webknossos/datastore/services/DSRemoteTracingstoreClient.scala
```diff
@@ -0,0 +1,73 @@
...
+import play.api.libs.json.JsObject
+
+class DSRemoteTracingstoreClient @Inject()(
+    rpc: RPC,
+    config: DataStoreConfig,
+    val lifecycle: ApplicationLifecycle,
+) extends LazyLogging
+    with FoxImplicits {
+
+  // TODO key check in tracing store
```
fm3, line 20: I think it is fair to say that if the user has the accessToken, all the below requests can be considered authorized. We can therefore skip the dataStoreKey validation, which would be a bit more complicated here, since the tracingstore would have to forward that to webKnossos.

So please remove this comment as well as the dataStoreKey logic from this file.

webknossos-datastore/app/com/scalableminds/webknossos/datastore/services/DSRemoteTracingstoreClient.scala
```diff
@@ -0,0 +1,73 @@
...
+  private val dataStoreKey: String = config.Datastore.key
+
+  def getZArray(tracingId: String, mag: String, tracingStoreUri: String, accessId: String): Fox[ZarrHeader] =
```
fm3, line 23:
```suggestion
  def getZArray(tracingId: String, mag: String, tracingStoreUri: String, accessToken: String): Fox[ZarrHeader] =
```
(lets’s try to use the same name everywhere)

webknossos-datastore/conf/com.scalableminds.webknossos.datastore.routes
```diff
@@ -16,18 +16,26 @@ GET           /datasets/:organizationName/:dataSetName/layers/:dataLayerName/his
...
+GET           /zarr_private/:accessToken/:dataLayerName                                                                   @com.scalableminds.webknossos.datastore.controllers.ZarrStreamingController.dataLayerFolderContentsPrivateLink(accessToken: String,  dataLayerName: String)
+GET           /zarr_private/:accessToken/:dataLayerName/.zattrs                                                           @com.scalableminds.webknossos.datastore.controllers.ZarrStreamingController.zAttrsWithAnnotationPrivateLink(accessToken: String, dataLayerName: String)
+GET           /zarr_private/:accessToken/:dataLayerName/.zgroup                                                           @com.scalableminds.webknossos.datastore.controllers.ZarrStreamingController.zGroupPrivateLink(accessToken: String,  dataLayerName: String)
+GET           /zarr_private/:accessToken/:dataLayerName/:mag                                                              @com.scalableminds.webknossos.datastore.controllers.ZarrStreamingController.dataLayerMagFolderContentsPrivateLink(accessToken: String,  dataLayerName: String, mag: String)
```
fm3, line 36: please resolve the merge conflicts with the already-in-master pr https://github.com/scalableminds/webknossos/pull/6366 – the new thing there is that the directory listing routes should exist both with and without a trailing slash. This should be the case for your new routes as well

webknossos-tracingstore/app/com/scalableminds/webknossos/tracingstore/controllers/VolumeTracingController.scala
```diff
@@ -355,6 +365,24 @@ class VolumeTracingController @Inject()(
...
   }
 
+  def zarrSource(token: Option[String], tracingId: String): Action[AnyContent] = Action.async { implicit request =>
+    accessTokenService.validateAccess(UserAccessRequest.readTracing(tracingId), urlOrHeaderToken(token, request)) {
+      for {
+        tracing <- tracingService.find(tracingId) ?~> Messages("tracing.notFound") ~> 404
+
+        zarrLayer = ZarrSegmentationLayer(
+          name = tracingId,
+          largestSegmentId = if (tracing.segments.nonEmpty) tracing.segments.maxBy(_.segmentId).segmentId else 1,
```
fm3, line 375: volume tracings have a largestSegmentId field, not all have the segment list. please use that :)

app/models/annotation/AnnotationPrivateLink.scala
```diff
@@ -0,0 +1,84 @@
...
+
+object AnnotationPrivateLinkParams {
+  implicit val jsonFormat: OFormat[AnnotationPrivateLinkParams] = Json.format[AnnotationPrivateLinkParams]
+}
+
+class AnnotationPrivateLinkService @Inject()()(implicit ec: ExecutionContext, val bodyParsers: PlayBodyParsers) {
+
+  val DefaultAnnotationListLimit = 1000
+  def publicWrites(annotationPrivateLink: AnnotationPrivateLink): Fox[JsValue] =
+    Fox.successful(Json.toJson(annotationPrivateLink))
```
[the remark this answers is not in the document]
fm3: Then again, I just saw that the objectIds are sent as json objects, which is not very nice. So maybe let’s keep the writes here but instead make it explicit so that it can unpack the objectId and send them as strings only.
I created refactoring issue https://github.com/scalableminds/webknossos/issues/6375 to resolve this properly later

webknossos-datastore/app/com/scalableminds/webknossos/datastore/controllers/ZarrStreamingController.scala
```diff
@@ -215,75 +258,259 @@ class ZarrStreamingController @Inject()(
...
+            remoteTracingstoreClient
+              .getDataLayerMagFolderContents(annotationLayer.tracingId,
+                                             mag,
+                                             annotationSource.tracingStoreUrl,
+                                             accessToken)
+              .map(
+                layers =>
+                  Ok(
+                    views.html.datastoreZarrDatasourceDir(
+                      "Tracingstore",
```
fm3, line 364: That seems wrong here ;) Maybe even include the info that it is this combined annotation thingy in the html page.

conf/webknossos.latest.routes
```diff
@@ -141,6 +141,13 @@ GET           /annotations/:typ/:id/sharedTeams
 PATCH         /annotations/:typ/:id/sharedTeams                                         controllers.AnnotationController.updateSharedTeams(typ: String, id: String)
 PATCH         /annotations/:typ/:id/othersMayEdit                                       controllers.AnnotationController.updateOthersMayEdit(typ: String, id: String, othersMayEdit: Boolean)
 
+# Annotation Private Links
+GET           /annotations/zarr_private/byAccessToken/:accessId                         controllers.AnnotationPrivateLinkController.lookupPrivateLink(accessId: String)
+POST          /annotations/zarr_private                                                 controllers.AnnotationPrivateLinkController.create
+PUT           /annotations/zarr_private/:id                                             controllers.AnnotationPrivateLinkController.update(id: String)
+DELETE        /annotations/zarr_private/:id                                             controllers.AnnotationPrivateLinkController.delete(id: String)
+GET           /annotations/zarr_private/:id                                             controllers.AnnotationPrivateLinkController.get(id: String)
```
fm3, line 149: There should also be a list route. It should list all that the user is allowed see (note my other comment about the readAccessQuery)

fm3 (changes requested):
Wow, this is quite a piece of work :muscle: :slightly_smiling_face:  Thanks a lot! I added a couple of notes to the code. I’ll do testing in the next round, please ping me again after you adressed the feedback :slightly_smiling_face:

conf/webknossos.latest.routes
fm3: good point, we should go for camel case. Do you have an idea for a good name? just zarrPrivate? Or zarrPrivateLinks?

app/models/binary/DataSet.scala
```diff
@@ -111,8 +111,15 @@ class DataSetDAO @Inject()(sqlClient: SQLClient,
       )
     }
 
-  override def anonymousReadAccessQ(sharingToken: Option[String]): String =
-    "isPublic" + sharingToken.map(t => s" or sharingToken = '$t'").getOrElse("")
+  override def anonymousReadAccessQ(token: Option[String]): String =
+    "isPublic" + token
+      .map(
+        t =>
+          " or sharingToken = '$t' or _id in (select _dataset " +
```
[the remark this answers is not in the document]
leowe: Since in the row after the join there is only one column with the name `_dataset` no disambiguation is necessary, but I added it nonetheless.

conf/webknossos.latest.routes
normanrz: how about just „zarr“?

fm3: I think I’d rather go for `zarrPrivateLinks`, `privateLinks`, or `zarrAccessToken` or something along those lines here, since that is the object that the CRUD routes are dealing with. Note that this is not the user-facing URI that they paste into the external tool; this is only used by wk.

Those user-facing URIs should probably just say `zarr` or possibly `annotations/zarr/` or 'zarr/annotations/`, if it does not lead to conflicts with the existing dataset streaming routes

normanrz: I thought you were talking about user facing URIs.

philippotto:
> You can similarly test the update, delete, and get routes

I think the delete route doesn't really work yet. First, it doesn't return JSON (which might be ok, but it returns html which seems weird). Second, it returns a 200 status code, but does not change the isDeleted flag in the DB. maybe i'm doing something wrong, but I would expect an error if I that's the case.

conf/webknossos.latest.routes
leowe: There is also no other route that uses a second "name" after the initial "name" (usually it is only `annotations`). I think it should just be `/zarrPrivateLinks` instead of `annotations/zarrPrivateLinks`.

philippotto:
> > You can similarly test the update, delete, and get routes
> 
> I think the delete route doesn't really work yet. First, it doesn't return JSON (which might be ok, but it returns html which seems weird). Second, it returns a 200 status code, but does not change the isDeleted flag in the DB. maybe i'm doing something wrong, but I would expect an error if I that's the case.

I pushed a very basic / simplified front-end which can create and delete private links. Just open an annotation and a modal should open automatically which shows the new UI. deletion doesn't work as mentioned above. the error toast happens because the frontend tries to parse the response as JSON. however, the links don't seem to change at all in the db.

leowe:
> I think the delete route doesn't really work yet. First, it doesn't return JSON (which might be ok, but it returns html which seems weird). Second, it returns a 200 status code, but does not change the isDeleted flag in the DB.

Please try the new API routes, I think there were some conflicting DELETE routes.

philippotto:
> Please try the new API routes, I think there were some conflicting DELETE routes.

it still returns html and doesn't seem to do anything in the db :thinking:

frontend/javascripts/admin/admin_rest_api.ts
```diff
@@ -545,6 +545,29 @@ export async function getUsersWithActiveTasks(projectId: string): Promise<Array<
...
+}
+
+export function getPrivateLinks(): Promise<Array<unknown>> {
+  return Request.receiveJSON("/api/annotations/zarr_private");
+}
+
+export function deletePrivateLink(linkId: string): Promise<{
+  messages: Array<Message>;
+}> {
+  return Request.receiveJSON(`/annotations/zarr_private/${linkId}`, {
```
fm3, line 566: @philippotto shouldn’t this include `/api/`?

philippotto: ahhh, nice catch. @leowe this was my problem. now the delete route works :)

app/models/annotation/AnnotationPrivateLink.scala
```diff
@@ -0,0 +1,84 @@
...
+      _ <- run(
+        sqlu"""insert into webknossos.annotation_privateLinks(_id, _annotation, accessToken, expirationDateTime, isDeleted)
+                         values(${aPL._id.id}, ${aPL._annotation.id}, ${aPL.accessToken}, ${time}, ${aPL.isDeleted})""")
+    } yield ()
+  }
+
+  def updateOne(id: ObjectId, _annotation: ObjectId, expirationDateTime: Option[Long])(
+      implicit ctx: DBAccessContext): Fox[Unit] =
+    for {
+      _ <- assertUpdateAccess(id)
```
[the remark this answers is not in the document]
leowe: I am not quite sure whether it's possible to build the update access query with only the user id. Let's discuss this later in person.

normanrz:
> I pushed a very basic / simplified front-end which can create and delete private links. Just open an annotation and a modal should open automatically which shows the new UI. deletion doesn't work as mentioned above. the error toast happens because the frontend tries to parse the response as JSON. however, the links don't seem to change at all in the db.

can i haz screenshot?

philippotto:
> > I pushed a very basic / simplified front-end which can create and delete private links. Just open an annotation and a modal should open automatically which shows the new UI. deletion doesn't work as mentioned above. the error toast happens because the frontend tries to parse the response as JSON. however, the links don't seem to change at all in the db.
> 
> can i haz screenshot?

it's too ugly to share, I think ;) I'll draw a mockup now to explore how the UI could look like.

frontend/javascripts/oxalis/view/action-bar/private_links_view.tsx
```diff
@@ -0,0 +1,347 @@
...
+      Toast.error(`Could not delete link. ${err}`);
+      if (context) {
+        queryClient.setQueryData(mutationKey, context.previousLinks);
+      }
+    },
+  });
+}
+
+function UrlInput({ linkItem }: { linkItem: ZarrPrivateLink }) {
+  const url = `https://.../${linkItem.accessToken}`;
```
philippotto, line 132: @leowe can you tell me how this url should look like? :)

philippotto:
I think, this is ready for a first testing round. I deployed the branch here: https://zarrprivatelinks.webknossos.xyz/ The URLs are not missing the correct structure, but the rest should be good.

Naming/wording could probably be improved (UI wise and code wise). Not sure about the "Manage Zarr Endpoints".

@fm3 maybe you want to give some feedback for the UI? :)

frontend/javascripts/oxalis/view/action-bar/private_links_view.tsx
fm3: Good question! The base URI is `${DATASTORE_URI}/annotations/zarr/${accessToken}` However, this will yield a directory listing with only a .zgroup and then subdirectories for the layers. Not all tools will be able to handle the top-level link, I assume, but will need the layer-specific ones. I’m not sure what the best way to handle this would be. @normanrz maybe you have some input?

fm3:
The UI works very nicely for me! I like the table view and the expiry date handling.
I would have expected this as a tab in the sharing modal, not as a separate menu item, but that’s also fair.
One thing I noticed: The links are pretty long. I’d say tokens half this length would do just fine. @leowe I know this has to be changed in a couple of places, but I think it would be worth the time.

I also noticed that “segmentation” is listed twice in the directory listing 🤔 I created a default hybrid annotation on the l4_sample dataset. I’d say if a segmentation layer is used as a fallback layer for one of the volume layers, it should not be listed separately. (Also their name may be identical which breaks the URIs anyway)

When a link is deleted, and then requested (I tried the directory listing), I get a 400 there. Should probably be 404.

Also note that @jstriebel recently merged https://github.com/scalableminds/webknossos/pull/6398 and https://github.com/scalableminds/webknossos/pull/6397 – @leowe  could you merge the master branch into this one adapt the new zarr endpoints in the same way?

I did not yet test the actual data streaming, I hope I’ll get to that in the coming days.

frontend/javascripts/oxalis/view/action-bar/private_links_view.tsx
```diff
@@ -0,0 +1,347 @@
...
+        </Tooltip>
+      ),
+    },
+  ];
+
+  return (
+    <div>
+      <div style={{ marginBottom: 8 }}>
+        You can create one or multiple Zarr link to this annotation below. The link can be used by
+        other tools to access the data in a streaming manner.
```
fm3, line 295:
```suggestion
      Create <a href="https://zarr.readthedocs.io/en/stable/">Zarr</a> streaming links to allow other tools to load the data of this annotation.
      Note that anyone with these links can access the data, regardless of other sharing settings.
```

frontend/javascripts/oxalis/view/action-bar/private_links_view.tsx
philippotto: > Not all tools will be able to handle the top-level link, I assume, but will need the layer-specific ones.

Do you know which tools are capable of using the "short" url? One could list these as an example. Maybe providing only the short version also applies some pressure on other software to incorporate support for that...
Alternatively, we could have a dropdown for the copy-button which allows to select the layer one wants to reference. However, it's probably tedious to use (especially if it's unclear which software demands which format). which is why I'd rather avoid that..

normanrz: I don't think there is any tool that can work with "short" urls, sadly. Copy buttons for individual layers would be user-friendly. Perhaps in a nested modal or popover?

normanrz:
I'd call the feature `Zarr Links` instead of `Zarr endpoints`.

I'm not sure whether these links should have an expiration date by default or not. I lean towards a default expiration.

frontend/javascripts/oxalis/view/action-bar/private_links_view.tsx
philippotto: I pushed a change which opens a dropdown list when clicking the copy button. I also thought about the modal or popover approach, but I think the dropdown is the cleanest option.

philippotto:
> I'd call the feature `Zarr Links` instead of `Zarr endpoints`.

Done :+1: 

> I'm not sure whether these links should have an expiration date by default or not. I lean towards a default expiration.

I tend towards the opposite :see_no_evil: Three reasons:

- if the expiration date is overlooked, it could catch the user off-guard ((our) other sharing links usually don't include an expiration date which is why this could be surprising)
- if there is a default expiration date, the user might think that there **has to be** an expiration date. however, it's not mandatory. deleting it can be done via the UI, but the user might not even bother with figuring out how to do it.
- a default expiration often will be less than ideal. if the user has to change it, they got to use the date picker. on the other hand, adding an expiration date, is a bit ux-friendlier since we provide the dropdown with "1d, 1 week etc".

just my three cents, though. I don't have a strong opinion on this matter.

philippotto:
@daniel-wer feel free to have a look at the front-end code :) I assume the basic structure won't change much, anymore. happy to jump on a call, too.

normanrz:
> > I'm not sure whether these links should have an expiration date by default or not. I lean towards a default expiration.
> 
> I tend towards the opposite 🙈 Three reasons:
> 
> * if the expiration date is overlooked, it could catch the user off-guard ((our) other sharing links usually don't include an expiration date which is why this could be surprising)
> * if there is a default expiration date, the user might think that there **has to be** an expiration date. however, it's not mandatory. deleting it can be done via the UI, but the user might not even bother with figuring out how to do it.
> * a default expiration often will be less than ideal. if the user has to change it, they got to use the date picker. on the other hand, adding an expiration date, is a bit ux-friendlier since we provide the dropdown with "1d, 1 week etc".
> 
> just my three cents, though. I don't have a strong opinion on this matter.

My argument is that these private links effectively circumvent the webKnossos security model and therefore should only do so for a limited amount of time. Only, if the users know what they are doing, this should be indefinite.

> * if there is a default expiration date, the user might think that there **has to be** an expiration date. however, it's not mandatory. deleting it can be done via the UI, but the user might not even bother with figuring out how to do it.

I wouldn't object to mandatory expiration dates.

> * if the expiration date is overlooked, it could catch the user off-guard ((our) other sharing links usually don't include an expiration date which is why this could be surprising)

Also, expiration dates can be easily extended.

frontend/javascripts/oxalis/view/action-bar/tracing_actions_view.tsx
```diff
@@ -585,6 +588,16 @@ class TracingActionsView extends React.PureComponent<Props, State> {
         Share
       </Menu.Item>,
     );
+    elements.push(
+      <Menu.Item
+        key="zarr-links-button"
+        onClick={() => this.setState({ isZarrPrivateLinksModalOpen: true })}
+      >
+        <LinkOutlined />
+        Manage Zarr Links
```
normanrz, line 597:
```suggestion
       Zarr Links
```

philippotto:
> > > I'm not sure whether these links should have an expiration date by default or not. I lean towards a default expiration.
> > 
> > 
> > I tend towards the opposite see_no_evil Three reasons:
> > 
> > * if the expiration date is overlooked, it could catch the user off-guard ((our) other sharing links usually don't include an expiration date which is why this could be surprising)
> > * if there is a default expiration date, the user might think that there **has to be** an expiration date. however, it's not mandatory. deleting it can be done via the UI, but the user might not even bother with figuring out how to do it.
> > * a default expiration often will be less than ideal. if the user has to change it, they got to use the date picker. on the other hand, adding an expiration date, is a bit ux-friendlier since we provide the dropdown with "1d, 1 week etc".
> > 
> > just my three cents, though. I don't have a strong opinion on this matter.
> 
> My argument is that these private links effectively circumvent the webKnossos security model and therefore should only do so for a limited amount of time. Only, if the users know what they are doing, this should be indefinite.

Ok, I understand. Which default would you suggest? 1 week?

normanrz:
> Ok, I understand. Which default would you suggest? 1 week?

I'd say 1 month.

hotzenklotz:
@philippotto BTW, when I click on the copy URL button, the browser only copies the short hand URL, e.g. `https://.../P4ZeUDcKHQWCsBzv-2rMnB_Kw-HzZeKm`. 🐛 

I would also be great if the general/non-private Zarr URL for (public) datasets would also be listed in the modal. Perhaps as a footer. (see also #6408 )

philippotto:
> @philippotto BTW, when I click on the copy URL button, the browser only copies the short hand URL, e.g. `https://.../P4ZeUDcKHQWCsBzv-2rMnB_Kw-HzZeKm`.

Sorry, the deployed dev instance was outdated. I updated it now. Please check again.

> I would also be great if the general/non-private Zarr URL for (public) datasets would also be listed in the modal. Perhaps as a footer. (see also #6408 )

Makes sense. Could somebody tell me how these URLs are built so that I can integrate them in the UI?

hotzenklotz:
> Makes sense. Could somebody tell me how these URLs are built so that I can integrate them in the UI?

@philippotto Very similar to the private URLs:
`https://<webKnossos data store url>/data/zarr/<organization_id>/<dataset name>/<layer name>`

frontend/javascripts/oxalis/view/action-bar/private_links_view.tsx
fm3: maybe this needs a target=_blank, just accidentally lost my wk tab :sweat_smile:

frontend/javascripts/oxalis/view/action-bar/private_links_view.tsx
fm3: I forgot to include `/data` in the above URI. It should be `{datastore}/data/annotations/zarr/{tokenValue}`

fm3:
Two more things I noticed during testing:
 - The datastore listing used `2-2-2` format for mags, while tracingstore wrote `2`
image
@leowe maybe you could unify this :)
 - In the datasource-properties.json, the volume layer name is not used, instead the tracingId is inserted. should be the former :) 
 
Other than that, things look good!

daniel-wer (commented):
Not much to add, frontend code looks very good. The react-query library seems to be helpful! During testing everything worked beautifully. I didn't test the links in any other tools, though.

- [ ] I think there should be a loading indicator when opening the "Zarr Links" modal. At first I thought the links weren't saved and then suddenly they popped up. (A little later I saw that the OK button shows a loading indicator, but I think that is too subtle. I would expect a larger loading spinner above the "Create Zarr Link" button. Not that high priority of course)

frontend/javascripts/oxalis/view/layouting/tracing_layout_view.tsx
```diff
@@ -123,16 +123,27 @@ class TracingLayoutView extends React.PureComponent<PropsWithRouter, State> {
...
+      const mainContainer = document.getElementById("main-container");
+      if (mainContainer) {
+        document.body.removeChild(mainContainer);
+      }
+    }
     window.removeEventListener("resize", this.debouncedOnLayoutChange);
-    const refreshMessage = document.createElement("p");
+
+    const refreshMessageContainer = document.createElement("div");
+    refreshMessageContainer.style.display = "grid";
```
philippotto, line 135: @daniel-wer I sneaked this change in this PR, since it caught my eye during testing. just as a heads-up ^^

frontend/javascripts/admin/admin_rest_api.ts
```diff
@@ -548,11 +548,14 @@ export async function getUsersWithActiveTasks(projectId: string): Promise<Array<
...
-export function createPrivateLink(annotationId: string): Promise<ZarrPrivateLink> {
+export function createPrivateLink(
+  annotationId: string,
+  initialExpirationPeriodInDays: number = 30,
+): Promise<ZarrPrivateLink> {
   return Request.sendJSONReceiveJSON("/api/zarrPrivateLinks", {
     data: {
       annotation: annotationId,
-      expirationDateTime: null,
+      expirationDateTime: new Date().getTime() + initialExpirationPeriodInDays * 24 * 3600 * 1000,
```
daniel-wer, line 558: More readable, less error prone (`moment` might need to be imported):

```suggestion
      expirationDateTime: moment().add(initialExpirationPeriodInDays, "days").valueOf(),
```

frontend/javascripts/oxalis/view/layouting/tracing_layout_view.tsx
daniel-wer: Awesome :tada:

daniel-wer (commented):
Nice work!

- [ ] @philippotto I think you mentioned that you wanted to add the possibility to share layers via zarr links in `View` mode. I would also vote to allow that if it's easily possible.

- [ ] @leowe One question regarding permissions. Who is allowed to create these sharing links? Anyone who can see a layer/an annotation? I'm thinking about the use case where someone shared an annotation with me privately via link and then later revokes the link. If I created a private zarr link in between, would I still be able to access the annotation?

- [ ] Also, the skeleton is probably never shared, right? Maybe the text in the sharing modal could be adapted to make it more clear what exactly is being shared :thinking:

philippotto:
@leowe I think this PR only includes the forward sql migration, but the reversion is missing. Is this intended?

philippotto:
The missing front-end tasks should be done now. I added the dataset zarr link to the existing sharing modal for datasets, as I think, it fits there best.

fm3:
> I think this PR only includes the forward sql migration, but the reversion is missing. Is this intended?

Hmm that may also have been caused by a merge. At least leo did mark this conversation as resolved some time ago https://github.com/scalableminds/webknossos/pull/6367#discussion_r938503105

leowe:
> * [ ]  @leowe One question regarding permissions. Who is allowed to create these sharing links? Anyone who can see a layer/an annotation? I'm thinking about the use case where someone shared an annotation with me privately via link and then later revokes the link. If I created a private zarr link in between, would I still be able to access the annotation?

The permission check right now is against the annotation owner, i.e. you can only create / update / delete private links when you own the annotation.

leowe:
> > I think this PR only includes the forward sql migration, but the reversion is missing. Is this intended?
> 
> Hmm that may also have been caused by a merge. At least leo did mark this conversation as resolved some time ago [#6367 (comment)](https://github.com/scalableminds/webknossos/pull/6367#discussion_r938503105)

Yes that seemed to have gone missing somewhere...

app/controllers/AnnotationPrivateLinkController.scala
```diff
@@ -0,0 +1,102 @@
...
+
+      annotationPrivateLink <- annotationPrivateLinkDAO.findOne(idValidated)
+      _ <- bool2Fox(annotationPrivateLink.expirationDateTime.forall(_ > System.currentTimeMillis())) ?~> "Token expired" ~> 404
+      _ <- annotationDAO.findOne(annotationPrivateLink._annotation) ?~> "annotation.notFound" ~> NOT_FOUND
+
+      annotationPrivateLinkJs <- annotationPrivateLinkService.publicWrites(annotationPrivateLink)
+    } yield Ok(annotationPrivateLinkJs)
+  }
+
+  def create: Action[AnnotationPrivateLinkParams] = sil.SecuredAction.async(validateJson[AnnotationPrivateLinkParams]) {
```
jstriebel, line 64: @leowe Please add some more information to the swagger annotations. All of those endpoints should have a proper `operationId` in https://zarrprivatelinks.webknossos.xyz/swagger.json, by default it is the function name. You can specify a different one via `nickname`, as seen on some other endpoints. Please also add some descriptions, especially for endpoints that expect a specific json format.

jstriebel: Only for the wk endpoints should be enough :+1: Let's see what we come up with tomorrow.

leowe:
I added all the changes that were requested for the backend.

daniel-wer (approved):
Frontend LGTM :+1:

app/controllers/AnnotationPrivateLinkController.scala
jstriebel: Thanks!

fm3 (approved):
LGTM! I also did some testing again, seemed to work as intended :) Thanks for your persistence with this!

PR #6369, opened 2022-08-03
Isosurfaces for volume annotation layers with fallback segmentation

### LINKED ISSUE #4930: Add Isosurface Support for Volume tracings with Fallback layer
Currently, the on the fly Isosurface generation of a tracing done by the backend is only supported for tracings without a fallback layer.
It would be nice if this feature would also support tracings with fallback layers.
The following questions need to be discussed:
- Is this actually a use case? Would our users use the isosurface generation in tracings with a fallback layer?
- What has to needs to be implemented for this feature?
  - @fm3 mentioned that to generate the isosurface data from the tracing and the datastore are required to be mixed. Currently, both stores do not have a way of communicating with each other.
- Does the value of this feature outweigh the implementation afford?

fm3:
- https://isosurfacefallback.webknossos.xyz

### Steps to test:
- Create volume annotation with fallback layer (default for datasets with segmentation layers)
- Pick an existing segment’s color, brush some (brush in multiple z-layers, so that isosurface subsampling won’t skip over it)
- right click segment, select ad-hoc mesh, wait, watch 3d viewport
- full segment’s isosurface should appear, featuring the brushed changes.
- also validate that isosurface loading in dataset view mode still works
- also validate that isosurface loading for volume annotations _without_ fallback segmentation still works

### Issues:
- fixes #4930 

------
- [x] Ready for review


fm3:
@philippotto the front-end change is tiny but maybe you could still have a look to make sure I didn’t miss a case there :)

frontend/javascripts/oxalis/model/sagas/isosurface_saga.ts
```diff
@@ -359,7 +359,7 @@ function* maybeLoadIsosurface(
   const tracingStoreUrl = `${tracingStoreHost}/tracings/volume/${layer.name}`;
   const volumeTracing = yield* select((state) => getActiveSegmentationTracing(state));
   // Fetch from datastore if no volumetracing exists or if the tracing has a fallback layer ...
```
philippotto, line 361:
```suggestion
  // Fetch from datastore if no volumetracing ...
```

philippotto:
awesome :star_struck: :star_struck: :star_struck: really cool to see this feature coming now :+1: worked very well for me too. front-end change makes sense, as well :+1:

frontend/javascripts/oxalis/model/sagas/isosurface_saga.ts
```diff
@@ -358,8 +358,8 @@ function* maybeLoadIsosurface(
   }`;
   const tracingStoreUrl = `${tracingStoreHost}/tracings/volume/${layer.name}`;
   const volumeTracing = yield* select((state) => getActiveSegmentationTracing(state));
-  // Fetch from datastore if no volumetracing exists or if the tracing has a fallback layer ...
-  let useDataStore = volumeTracing == null || volumeTracing.fallbackLayer != null;
+  // Fetch from datastore if no volumetracing ...
+  let useDataStore = volumeTracing == null;
```
jstriebel, line 362: This means, that the requests go to the tracingstore, when there's a volume annotation + fallback data, right? But shouldn't this rather go to the datastore, since it probably has the larger amount of data? I'd assume this is just the same arguments as for zarr annotation streaming.

fm3: While this could be a potential future pervormance improvement, I would argue against this at this time.
- In fact the zarr annotation streaming draft in https://github.com/scalableminds/webknossos/pull/6367 _also_ does not currently load volume layer fallback data from the datastore directly. Instead, there the datastore checks requests to see if they are for datastore-only layers, then it answers them itself, otherwise it forwards them to the tracingstore (which may in turn fetch fallback buckets from the datastore)
- Even when supplying the fallback data directly, the datastore would still have to send a request to the tracingstore to find out which buckets the tracingstore has. Only on failure it would then load the data directly. So the benefit is not that big
- The datastore currently has no models for annotations and their layers. The PR mentioned above goes around this by requesting a lot of stuff from webKnossos. To ensure that no permission leaks come up, this is done via those private links only. We would need a concept for that here too
- So yes, to reduce the data volume, we may change this in the future, but I’d say that is not part of this issue’s scope

jstriebel: > While this could be a potential future pervormance improvement, I would argue against it at this time.

Sounds fair to me. Could you add an issue for it? And maybe we should document somewhere, which APIs on which server contain which content, as this is now getting slightly complicated to remember.

> * In fact the zarr annotation streaming draft in [Zarr private links #6367](https://github.com/scalableminds/webknossos/pull/6367) _also_ does not currently load volume layer fallback data from the datastore directly. Instead, there the datastore checks requests to see if they are for datastore-only layers, then it answers them itself, otherwise it forwards them to the tracingstore (which may in turn fetch fallback buckets from the datastore)

Ok :+1: 

> * Even when supplying the fallback data directly, the datastore would still have to send a request to the tracingstore to find out which buckets the tracingstore has. Only on failure it would then load the data directly. So the benefit is not that big

Yes, the tracingstore would still need to answer, but without doing the additional request from the tracingstore to the datastore for the fallback chunks. Since tracingstore and datastore are most of the time not on the same machine in our deployments, I think this might still save quite some time, so it's not that small either.

> * The datastore currently has no models for annotations and their layers. The PR mentioned above goes around this by requesting a lot of stuff from webKnossos. To ensure that no permission leaks come up, this is done via those private links only. We would need a concept for that here too
> * So yes, to reduce the data volume, we may change this in the future, but I’d say that is not part of this issue’s scope

:+1:

fm3: https://github.com/scalableminds/webknossos/issues/6374

jstriebel (approved):
LGTM, didn't do any testing myself though :+1:

normanrz:
This will make some of our users very happy!

PR #6370, opened 2022-08-03
Allow to extrude segment instead of interpolate (similar to the old "copy from previous slice" feature)

### LINKED ISSUE #6338: Bring back copy from previous slice (overload volume interpolation feature)
## Detailed Description
The new interpolation feature replaced the old copy-from-previous slice feature. However, the new interpolation code could be adapted so that no interpolation happens if you don't want it to. That way, users who still would like to use the old feature could do so without needing to re-add the old implementation.

However, then the UX question remains. Another button? Which shortcut? Or alternatively: perhaps a switch between "use interpolation" and "only copy", which changes what the shortcut (and the button) does.

That way, we could also introduce the feature of copying a slice across n sections.

## Context

philippotto:
This PR generalizes the volume interpolation feature to also provide an "extrusion" feature (i.e., copy a segment from one slice to N slices). This is similar to the old copy-from-previous slice feature, but a bit more powerful (since one can label N slices at the same time).
wK remembers whether interpolation or extrusion was used last and switches the main button accordingly.

Screenshots:
image

image

image

- https://restorecopyfromslice.webknossos.xyz/

### Steps to test:
- use volume interpolation as usual (via shortcut V or via button in the toolbar)
- label a slice, move a few slices via `f` and then use the extrusion feature via the dropdown in the toolbar
- the segment should be copied to all slices between the start and the current slice (including the current slice)
- also, the button should change to the extrusion icon
- using the feature again with the extrusion button (without the dropdown) should also work as expected

### Issues:
- fixes #6338 

------
- [X] Ready for review


frontend/javascripts/oxalis/view/action-bar/toolbar_view.tsx
```diff
@@ -233,26 +235,62 @@ function OverwriteModeSwitch({
   );
 }
 
+const INTERPOLATION_ICON = {
```
daniel-wer, line 238: Very creative icon usage :100:

frontend/javascripts/oxalis/view/action-bar/toolbar_view.tsx
```diff
@@ -233,26 +235,62 @@ function OverwriteModeSwitch({
...
+    <Menu
+      onClick={(e) => {
+        dispatch(updateUserSettingAction("interpolationMode", e.key as InterpolationMode));
+        onInterpolateClick(null);
+      }}
+      items={[
+        {
+          label: "Interpolate current segment",
+          key: InterpolationModeEnum.INTERPOLATE,
+          icon: <i className="fas fa-align-center fa-rotate-90" />,
```
daniel-wer, line 267: You could also use the `INTERPOLATION_ICON` object here and in line 272 to avoid code duplication

daniel-wer (approved):
Works beautifully! The only thing I'm not sure about is whether users will be familiar with the term "segment extrusion" :thinking: It's the first time I've heard of it, but I'm also not a power-user of image manipulation software - did you encounter this term before and that's why you've used it?

frontend/javascripts/oxalis/view/action-bar/toolbar_view.tsx
```diff
@@ -233,26 +235,62 @@ function OverwriteModeSwitch({
...
+        onInterpolateClick(null);
+      }}
+      items={[
+        {
+          label: "Interpolate current segment",
+          key: InterpolationModeEnum.INTERPOLATE,
+          icon: INTERPOLATION_ICON[InterpolationModeEnum.INTERPOLATE],
+        },
+        {
+          label: "Extrude current segment",
```
philippotto, line 270:
```suggestion
          label: "Extrude (copy) current segment",
```

frontend/javascripts/oxalis/model/sagas/volume/volume_interpolation_saga.ts
```diff
@@ -100,9 +121,12 @@ export function getInterpolationInfo(state: OxalisState, explanationPrefix: stri
   const isPossible = disabledExplanation == null;
   tooltipAddendum = disabledExplanation || tooltipAddendum;
 
+  const tooltipMain = onlyExtrude
+    ? "Extrude current segment from last labeled until current slice (V)"
```
philippotto, line 125:
```suggestion
    ? "Extrude (copy) current segment from last labeled until current slice (V)"
```

PR #6376, opened 2022-08-05
List unlisted users to superUsers

### LINKED ISSUE #6353: List unlisted users for super users
Users can be unlisted (e.g. super users that switched into orgas for debugging purposes).
Other superusers should still be able to list them (including themselves).
This will require a separate list access query, like we already have for annotations.

fm3:
To enable better debugging in other organizations, unlisted users are now listed IF the requesting user is a superUser

### Steps to test:
 - In application.conf, set isDemoInstance = true.
 - In private window, sign up as new user, creating organization, in which you are admin
 - As sample@scm.io, switch to that new orga. In user list, you should see yourself there
 - In the private window, as the new orga’s admin, you should not see sample@scm.io (they are still unlisted to you)

### Issues:
- fixes #6353

------
- [x] Ready for review


leowe (approved):
LGTM :+1:

PR #6377, opened 2022-08-05
Filter datasets by organization, use that in task creation view

### LINKED ISSUE #3568: Allow task creation for public datasets of other organizations
Needed steps:
 - discuss if we want this at all
 - introduce organization (name) parameter in task creation for dataset disambiguation (currently datasets are looked up using the user’s organization)
 - other organization’s datasets in task creation selection form, possibly grouped or even optional to avoid dataset spam

### LINKED ISSUE #5598: Task Creation View lists too many datasets
## Context
When creating a new task for my organization I am asked to choose a dataset from the dropdown. The dropdown list all public datasets available in the dataset gallery in addition to my organization's datasets.

image

## Expected Behavior
I only want to see my orga's dataset.

## Current Behavior
I see my orga's datasets + all public datasets form the publication gallery.

## Steps to Reproduce the bug
1. Create a new task
2. Click on dataset dropdown
3. Tada!

## Your Environment for bug
* Version of WebKnossos (Release or Commit): wk.org

### LINKED ISSUE #5923: task creation allows selection of unavailable datasets
## Context
I was creating a task as admin of an organisation

## Expected Behavior
All datasets that can be selected in the dropdown menu should create task without error message

## Current Behavior
For some datasets one receives an error message

## Steps to Reproduce the bug
Create new organisation, login as admin, create new task type, create new project, create new experience, create new task and select Neuromast_T1 as dataset

## Your Environment for bug
Windows 10, Chrome Version 96.0.4664.110, webknossos.org

image

### LINKED ISSUE #6329: Allow to filter /api/datasets by organization, uploader
## Detailed Description
Atm `/api/datasets` shows all datasets (even public ones of other orgs, which we might want to reconsider). It would be nice to be able to filter those for the python client by the owning organization, the uploader, and tags (using `any` semantics for the tags).

fm3:
- Public datasets of other organizations are no longer listed during task creation. (Previously they were listed but still ultimately failed, compare #3568)
- Added new filter options for the dataset list api at `api/datasets`: `organizationName: String`, `onlyMyOrganization: Boolean`, `uploaderId: String`

### Steps to test:
- Set up two organizations, add datasets for both, set some datasets to public
- In task creation view you should not see other organization’s datasets in the dropdown menu, even if they are public
- try new filter options for dataset list: `organizationName: String`, `onlyMyOrganization: Boolean`, `uploaderId: String`

### Issues:
- fixes #5923
- fixes #5598
- fixes #6329
- related: #3568

------
- [x] Ready for review


philippotto (commented):
front-end lgtm :+1:

util/src/main/scala/com/scalableminds/util/tools/Converter.scala
```diff
@@ -10,6 +10,10 @@ object DefaultConverters {
       }
   }
 
+  implicit object StringToString extends Converter[String, String] {
```
jstriebel, line 13: :grey_question:
(Why is this needed? Doesn't do much harm, just wondering…)

jstriebel (changes requested):
The backend code looks mostly fine, I'd just like to have a Swagger annotation for the parameters of the `datasetList` endpoint :writing_hand:

util/src/main/scala/com/scalableminds/util/tools/Converter.scala
fm3: The current implementation of Filters extracts the get parameters, and uses these Converters to type them. There wasn’t a string one before, and this implicit method was needed.

However, I guess we should stop using this method, and go for explicit get parameters, also allowing for standard swagger anotations. It may be that I can remove these converters then. I’ll have a look

fm3:
@jstriebel I added swagger annotations and listed the parameters explicitly, including types. Please have another look!

jstriebel (approved):
Perfect, LGTM :+1:

PR #6379, opened 2022-08-05
Strict permission checks during all steps of dataset upload

### LINKED ISSUE #5787: Dataset Upload: Separate initialTeams timeout from name reservation
## Detailed Description
follow-up for #5786 – The initialTeams timeout mechanism is supposed to guard against users setting the initial teams of datasets even though they are not the uploader (or admin). It was introduced in #4663

However, since #4999 the name reserving for datasets during the upload also happens by inserting a dataset row into the database. Since this can happen significantly earlier than the finishUpload request, the timeout mechanism was broken and gives false positives.

We should separate those two mechanisms again.

#5786 increases the timeout to 24h as a hotfix.

We should make sure that the timeout only counts from reporting the finished upload until setting the teams again, and then reduce it to 1h again.

fm3:
Tightened and cleaned up security for dataset uploads
 - for every chunk, and finishUpload, the token is validated against the upload id (don’t worry, there is a cache)
 - the initial teams timeout check was removed (since it has already previously been moved to the reserveUpload request which can be sent only once per dataset)
 - administrateDataSources access checks are now organization-specific for better isolation
 - removed unused (and broken) datasource list route

### Steps to test:
- upload a dataset as admin or dataset manager or team manager, should go smoothly, including setting teams
- try to forge requests screwing with anyone else’s uploads

### Issues:
- fixes #5787

------
- [x] Ready for review


leowe (approved):
LGTM :+1: although I didn't really know how to test the "forge requests screwing with anyone else’s uploads"

PR #6380, opened 2022-08-05
Show reload layer button only to those with permission

### LINKED ISSUE #5734: Reload layer in incognito fails
## Expected Behavior
Probably, the button shouldn't be available for non-authorized users.

## Current Behavior
It fails with unhelpful error messages.
Screenshot 2021-09-17 at 16 19 46

## Steps to Reproduce the bug
I was clicking the "reload" button on the color layer in this dataset: https://webknossos.org/datasets/scalable_minds/l4dense_motta_et_al_demo/view

fm3:
The “reload data” button for dataset and annotation layers is now only shown to users who have administrative permissions for the dataset (because the button clears server caches and file handles).

- https://hidereloadlayerbutton.webknossos.xyz

### Steps to test:
- Register as second user, then as admin activate that user
- share a dataset with their team
- as that user, open the dataset. Layer tab should not feature reload buttons
- as admin/dataset manager/team manager of datast teams, you should still see the reload buttons. 

### Issues:
- fixes #5734

------
- [x] Ready for review


philippotto (approved):
thank you!

PR #6381, opened 2022-08-06
Return 404 for not-found zarr chunks

### LINKED ISSUE #6179: Zarr Streaming - 404 for out-of-bounding box chunks
## Detailed Description
Follow-up for [#6144](https://github.com/scalableminds/webknossos/pull/6144).

For the Zarr streaming API and for datasets that have an offset != (0,0,0), the size of the bounding box in the `.zarray` is extended to include the original bounding box. In the `.zarray` file there is also a field `fill_value` for filling empty chunks.

Currently, when a bucket outside of the original bounding box is requested, an empty `200` Response is sent. In the analogue code in the `parr-python` library (https://github.com/zarr-developers/zarr-python/blob/master/zarr/core.py#L1938-L1945 and
https://github.com/zarr-developers/zarr-python/blob/master/zarr/storage.py#L1360-L1361), a `404` is interpreted as empty and correspondingly filled with the `fill_value`. 

We could do the same thing for every empty bucket (i.e. bucket that only contains `0`s)

normanrz:
Return 404 for not-found zarr chunks. Not sure if this plays well with the other currently in-review PRs for Zarr. Happy to close this one again, if you think that makes sense. I was just curious how a fix could be implemented.

### Steps to test:
- Assuming you have the `l4_sample` dataset and set it to "public":
- `curl -I http://localhost:9000/data/zarr/sample_organization/l4_sample/color/1-1-1/0.112.112.32` -> 200 Ok
- `curl -I http://localhost:9000/data/zarr/sample_organization/l4_sample/color/1-1-1/0.0.0.0` -> 404 Not found

### Issues:
- fixes #6179

------
- [x] Ready for review

### DIFF
--- a/webknossos-datastore/app/com/scalableminds/webknossos/datastore/controllers/ZarrStreamingController.scala
+++ b/webknossos-datastore/app/com/scalableminds/webknossos/datastore/controllers/ZarrStreamingController.scala
@@ -212,7 +212,8 @@ class ZarrStreamingController @Inject()(
           ),
           DataServiceRequestSettings(halfByte = false)
         )
-        (data, _) <- binaryDataService.handleDataRequests(List(request))
+        (data, notFoundIndices) <- binaryDataService.handleDataRequests(List(request))
+        _ <- bool2Fox(notFoundIndices.size == 0) ~> "zarr.chunkNotFound" ~> 404
       } yield Ok(data)
     }
   }
--- a/webknossos-datastore/app/com/scalableminds/webknossos/datastore/controllers/ZarrStreamingController.scala
+++ b/webknossos-datastore/app/com/scalableminds/webknossos/datastore/controllers/ZarrStreamingController.scala
@@ -43,7 +43,7 @@ class ZarrStreamingController @Inject()(
                                         urlOrHeaderToken(token, request)) {
         for {
           dataSource <- dataSourceRepository.findUsable(DataSourceId(dataSetName, organizationName)).toFox ?~> Messages(
-            "dataSource.notFound") ~> 404
+            "dataSource.notFound") ~> NOT_FOUND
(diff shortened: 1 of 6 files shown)


webknossos-datastore/app/com/scalableminds/webknossos/datastore/controllers/ZarrStreamingController.scala
```diff
@@ -212,7 +212,8 @@ class ZarrStreamingController @Inject()(
           ),
           DataServiceRequestSettings(halfByte = false)
         )
-        (data, _) <- binaryDataService.handleDataRequests(List(request))
+        (data, notFoundIndices) <- binaryDataService.handleDataRequests(List(request))
+        _ <- bool2Fox(notFoundIndices.size == 0) ~> "zarr.chunkNotFound" ~> 404
```
fm3, line 216:
```suggestion
        _ <- bool2Fox(notFoundIndices.isEmpty) ~> "zarr.chunkNotFound" ~> 404
```

fm3 (approved):
I think there is a second method that needs your change, somewhere in VolumeTracingController. Otherwise, LGTM

normanrz:
Seems to me `VolumeTracingController` already returns a 404 on missing chunks.

webknossos-datastore/app/com/scalableminds/webknossos/datastore/controllers/ZarrStreamingController.scala
normanrz: The other lines above also use `~> 404` should I replace them all?

fm3: if it’s no trouble, yes please :)

fm3 (approved):

PR #6382, opened 2022-08-08
Dataset Upload: Improve heuristic to detect mag/layer-only dir

### LINKED ISSUE #5884: wkw datasets with numerical layer name strings are not loaded correctly
## Context
When creating a WKW dataset, setting a layer name to a numerical string will cause webknossos to not load correctly the entire dataset

## Expected Behavior
The dataset is loaded correctly

## Current Behavior
When trying to read the dataset in webknossos two errors are raised:
 

- `Error processing layer 'cdbbcad7d293338e/test_dataset_1638452336/color' - Error reading resolutions`

- `Error processing resolution 'Left(2513)' - binaryData/cdbbcad7d293338e/test_dataset_1638452336/color/2513/header.wkw (No such file or directory)`

## Steps to Reproduce the bug
This python script creates a bug-reproducing dataset and uploads it

``` [python]
from pathlib import Path
import random

import numpy as np
from datetime import datetime
import webknossos as wk

dataset_name = "test_dataset_" + str(int(round(datetime.timestamp(datetime.now()))))
output_path = Path(dataset_name)

shape = (100, 100, 2)
data = np.random.randint(0, 255, shape, dtype=np.uint8)
with wk.webknossos_context(token="my_token"):
    dataset = wk.Dataset.create(dataset_path=str(output_path), name=dataset_name, scale=(1, 1, 1))
    color_layer = dataset.add_layer(layer_name=str(random.randint(0, 9999)), category="color", dtype_per_layer=data.dtype)
    mag = color_layer.add_mag(1)
    mag.write(data)
    mag.compress()

    url = dataset.upload()
    print(f"error reproducing dataset uploaded to {url}")
```

## Workaround
Use alphanumeric strings for layer names

## Environment

- python==3.8
- webknossos==0.8.24

fm3:
When users upload a wkw dataset with only a mag directory or only a layer directory (and no datasource-properties.json), there is a heuristic in place to fix the directory structure (i.e. create the mag parent or layer parent dir). Based on this, the explore route can later infer a datasource-properties.json.
However, the heuristic had false positives on numerical-only layer names. This PR replaces the heuristic with a new one that searches for "header.wkw" files and then looks at how deep in the directory tree they occur.

### Steps to test:
- Upload a normal wkw dataset, should still work
- Upload just the content of a layer directory, should be wrapped in "color" directory (see binaryData dir)
- Upload just the content of a mag directory, should be wrapped in "color/1" directory (see binaryData dir)
- Note that those repaired datasets will not be imported immediately. Rather, you need to go to the settings page first and import, to make them usable (i.e. create a datasource-properties.json file. This did not change in this PR)

### Issues:
- fixes #5884 

------
- [x] Ready for review


jstriebel (approved):
LGTM, didn't test it though

PR #6384, opened 2022-08-08
Fix job export link in slack notification

### LINKED ISSUE #6080: Result link for exported tiff is broken in slack notification
Result link for export tiff job contains webknossos.org prefix, should be absolute and pointing to datastore. (this only happens for remote datastore)

Example: https://scm.slack.com/archives/CMBMU5684/p1645699434336199

- [x] Specific to long-running jobs (set `jobsEnabled=true` in `application.conf`)
- [x] Specific to webKnossos.org (set `isDemoInstance=true` in `application.conf`)

fm3:
Note, though, that the datastore token is not automatically added to the link, but adding the user token is still one step less work than before, when you also had to fix the uri :grin: 

fixes #6080


jstriebel (approved):
:+1:

PR #6386, opened 2022-08-08
adds button for duplicating annotations

### LINKED ISSUE #5570: Create Duplicate of Annotation
There is “copy to my account” if you view someone else’s annotation. Sometimes you want to do the same for one of your own annotations. The backend should already support this :tm:

normanrz:
### Steps to test:
- Open an annotation (yours or from somebody else)
- Open the menu next to "save", click duplicate
- The current annotation will be saved, a duplicate will be created and you will be redirected to the duplicate

### Issues:
- fixes #5570

------
- [x] Ready for review


philippotto (approved):
lgtm :+1:

normanrz:
@philippotto Can you think of situations when we don't want to allow duplicating?

philippotto:
> @philippotto Can you think of situations when we don't want to allow duplicating?

no, even if we would forbid it in certain circumstances, downloading+reuploading would be a workaround, anyway (which we don't forbid, either).

normanrz:
> > @philippotto Can you think of situations when we don't want to allow duplicating?
> 
> no, even if we would forbid it in certain circumstances, downloading+reuploading would be a workaround, anyway (which we don't forbid, either).

Well, there is an `allowDownload` restriction.

fm3:
allowDownload is currently always identical to allowAccess

Issue #6387, opened 2022-08-09
Use brightness min/max from configuration for thumbnails, support float data


fm3:
Some datasets use only a small range of their data type’s possible intensity range. The wk frontend histogram sliders allow looking at that range in detail, but thumbnails will remain low-contrast.
It would be desirable to use this information and scale the data brightness when generating thumbnails.
With that, it would also be possible to support thumbnails for float data

PR #6388, opened 2022-08-09
Fix uint16 grayscale to rgb conversion for thumbnails

fm3:
The function returns an rgb int for every position. For uint16 grayscale data, I now consider the upper byte the color value for r,g and b. This discards the precision of the lower bytes, but for uint16 data that uses its full intensity range, produces reasonable-looking images.

I am not certain what the semantics of the leftmost byte in this rgb int is, but it looks like it is discarded. I am now writing 0xFF there, as is done for uint8 grayscale data.

Note that follow-up issue #6387 will make converting the input byte data to the correct number type necessary, so this function will be completely rewritten then, hopefully also fixing the 4 byte case.

I also changed the cuboid dimension limit (which was previously a too-strict limit on thumbnail size)

### Steps to test:
- I tested that uint16 data now produces reasonable-looking thumbnails

------
- [x] Ready for review


jstriebel (commented):
LGTM :+1: 

> I am not certain what the semantics of the leftmost byte in this rgb int is, but it looks like it is discarded. I am now writing 0xFF there, as is done for uint8 grayscale data.

:+1: (I guess it's a placeholder for an unused alpha value?)

jstriebel (approved):

PR #6407, opened 2022-08-16
Fixed "clip histogram" button not resetting in some cases

### LINKED ISSUE #6341: After histogram clipping fails (e.g. no data in viewport), button stays disabled
I clicked the histogram-auto-clipping button while being in a black-only area in the dataset. It correctly showed a warning that this does not work. However, I then moved to a spot with real data and wanted to click again, but the button stays in the spinner mode and is disabled. After the error it should go back to a clickable state.

image

hotzenklotz:
This PR fix a small issue with an AsyncButton for clipping the histogram not resetting to it's "base"/cklickable state in some case but instead "loading"/"spinning" forever. 

- https://___.webknossos.xyz

### Steps to test:
- Move to an area with no data ("black")
- Click "clip histogram" button
- Oberserve that the button turns into a spinner and returns to its base state ready for clicking again

### Issues:
- fixes #6341

------
- [x] Ready for review


philippotto (approved):
awesome, thank you for debugging this :+1:

PR #6410, opened 2022-08-17
Fix expanding the sidebar preventing hotkeys from working immediately

### LINKED ISSUE #5827: Keyboard shortcuts don't work after using button to collapse/expand sidebars
a `blur()` is probably needed when clicking these buttons.

## Steps to Reproduce the bug

1. expand/collapse sidebar via button
2. press f to move across a dataset (this does not work currently; a click on the data viewport is necessary first)
3.
4.

## Your Environment for bug

hotzenklotz:
This PR fixes a bug whereby expanding/collpasing the sidebar using the little icon buttons would make the main XYZ viewports lose focs/not respond to keyboard inputs anymore. Instead of having to click in the viewport to restore focs and continue using the keyboard hotkeys this now should work without any addtional click/refocusing.

Techical details:
- It seems to me that the first `blur()` call only every blurred the <div> element of the icon itself. For some reasons, the surround button element still retained focus/control. By also calling `blur()`onthe parent button element the issue resolved itself.

- https://___.webknossos.xyz

### Steps to test:
- Expand/Collapse sidebar via button
- Press any hotkey, e.g. f or SPACE, and observe that the intended behavior, e.g. movement, occurs

### Issues:
- fixes #5827

------
- [x] Ready for review

### DIFF
--- a/frontend/javascripts/oxalis/view/components/border_toggle_button.tsx
+++ b/frontend/javascripts/oxalis/view/components/border_toggle_button.tsx
@@ -46,7 +46,9 @@ function BorderToggleButton(props: Props) {
         onClick={(event) => {
           if (event != null) {
             // @ts-expect-error ts-migrate(2339) FIXME: Property 'blur' does not exist on type 'EventTarge... Remove this comment to see the full error message
-            event.target.blur();
+            event.target.blur(); // this will only blur the the wrapped icon <div> element
+            // @ts-expect-error ts-migrate(2339) FIXME: Property 'blur' does not exist on type 'EventTarge... Remove this comment to see the full error message
+            event.target.parentElement.blur(); // this will only blur the <button> element
           }
 
           onClick();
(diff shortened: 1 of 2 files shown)


frontend/javascripts/oxalis/view/components/border_toggle_button.tsx
```diff
@@ -46,7 +46,9 @@ function BorderToggleButton(props: Props) {
         onClick={(event) => {
           if (event != null) {
             // @ts-expect-error ts-migrate(2339) FIXME: Property 'blur' does not exist on type 'EventTarge... Remove this comment to see the full error message
-            event.target.blur();
```
philippotto, line 49: I think this blur() can be removed, as it probably doesn't do anything. As far as I know, only one blur will have an effect, since only one element can have the page focus. I also just tested it and the second blur seems to be enough to solve the issue :tada:

philippotto (commented):
thanks for digging into this!

frontend/javascripts/oxalis/view/components/border_toggle_button.tsx
philippotto: ah, it seems like it depends on where the user clicks exactly. sometimes the inner element gets the focus and sometimes the outer element. this is why both blurs make sense (however, with a slightly differing explanation than in your PR :)).

philippotto (approved):

PR #6411, opened 2022-08-18
If Layer BoundingBoxes intersect, Use Intersection Center for all Layer Thumbnails

### LINKED ISSUE #6405: Thumbnails for layers don’t seem to use the same position/zoom
We noticed that for some dataset thumbnails the layers don’t seem to show the same position/zoom. https://scm.slack.com/archives/C02A9H8CU/p1660572352968639

fm3:
- https://thumbnailstablecenter.webknossos.xyz

### Steps to test:
- Set up a dataset with properties so that the layer bounding boxes are not equal
- Request thumbnails for these layers, i.e. `http://localhost:9000/api/datasets/sample_organization/dsA/layers/segmentation/thumbnail`, `http://localhost:9000/api/datasets/sample_organization/dsA/layers/color/thumbnail`
- The thumbnails should show the same region of the dataset

### Issues:
- fixes #6405 

------
- [x] Ready for review

### DIFF
--- a/util/src/main/scala/com/scalableminds/util/geometry/BoundingBox.scala
+++ b/util/src/main/scala/com/scalableminds/util/geometry/BoundingBox.scala
@@ -11,7 +11,27 @@ case class BoundingBox(topLeft: Vec3Int, width: Int, height: Int, depth: Int) {
       math.max(topLeft.y, other.topLeft.y) < math.min(bottomRight.y, other.bottomRight.y) &&
       math.max(topLeft.z, other.topLeft.z) < math.min(bottomRight.z, other.bottomRight.z)
 
-  def combineWith(other: BoundingBox): BoundingBox = {
+  def intersection(other: BoundingBox): Option[BoundingBox] = {
+    val newTopLeft = Vec3Int(
+      math.max(topLeft.x, other.topLeft.x),
+      math.max(topLeft.y, other.topLeft.y),
+      math.max(topLeft.z, other.topLeft.z)
+    )
+    val newBottomRight = Vec3Int(
+      math.min(bottomRight.x, other.bottomRight.x),
+      math.min(bottomRight.y, other.bottomRight.y),
+      math.min(bottomRight.z, other.bottomRight.z)
+    )
+    if (newTopLeft.x < newBottomRight.x && newTopLeft.y < newBottomRight.y && newTopLeft.z < newBottomRight.z) {
+      Some(
+        BoundingBox(newTopLeft,
+                    newBottomRight.x - newTopLeft.x,
+                    newBottomRight.y - newTopLeft.y,
+                    newBottomRight.z - newTopLeft.z))
+    } else None
+  }
+
+  def union(other: BoundingBox): BoundingBox = {
     val x = math.min(other.topLeft.x, topLeft.x)
     val y = math.min(other.topLeft.y, topLeft.y)
     val z = math.min(other.topLeft.z, topLeft.z)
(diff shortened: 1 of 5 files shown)


util/src/main/scala/com/scalableminds/util/geometry/BoundingBox.scala
```diff
@@ -77,13 +97,23 @@ object BoundingBox {
...
     }
 
+  def intersection(bbs: List[BoundingBox]): Option[BoundingBox] =
+    bbs match {
+      case head :: tail =>
+        tail.foldLeft[Option[BoundingBox]](Some(head)) { (aOpt, b) =>
+          aOpt.flatMap(_ intersection b)
+        }
+      case _ =>
+        None
```
jstriebel, line 115: I guess any empty bbox would also be fine, but doesn't matter probably.

util/src/main/scala/com/scalableminds/util/geometry/BoundingBox.scala
```diff
@@ -11,7 +11,27 @@ case class BoundingBox(topLeft: Vec3Int, width: Int, height: Int, depth: Int) {
...
+      math.min(bottomRight.y, other.bottomRight.y),
+      math.min(bottomRight.z, other.bottomRight.z)
+    )
+    if (newTopLeft.x < newBottomRight.x && newTopLeft.y < newBottomRight.y && newTopLeft.z < newBottomRight.z) {
+      Some(
+        BoundingBox(newTopLeft,
+                    newBottomRight.x - newTopLeft.x,
+                    newBottomRight.y - newTopLeft.y,
+                    newBottomRight.z - newTopLeft.z))
+    } else None
```
jstriebel, line 31: This might also be an empty bbox IMO, but doesn't really matter.

jstriebel (approved):
LGTM :+1:

philippotto:
:+1: 

What happens if the intersection is empty? Rather unlikely, but a reasonable default would be cool.

fm3:
I wanted to return None if there is no intersection, because otherwise the returned empty bbox carries an arbitrary topleft, which seemed weird. Also, the center of a zero-volume box may still be defined, depending on how you understand it. And my code needs to find out if there is no intersection, because then it passes None to the datastore as center, which will then again use each layer’s individual bounding box, to answer @philippotto’s question as well

util/src/main/scala/com/scalableminds/util/geometry/BoundingBox.scala
```diff
@@ -77,13 +97,23 @@ object BoundingBox {
...
       None
 
-  def combine(bbs: List[BoundingBox]): BoundingBox =
+  def union(bbs: List[BoundingBox]): BoundingBox =
     bbs match {
       case head :: tail =>
-        tail.foldLeft(head)(_ combineWith _)
+        tail.foldLeft(head)(_ union _)
       case _ =>
         BoundingBox(Vec3Int(0, 0, 0), 0, 0, 0)
```
jstriebel, line 105: Maybe this should be None as well then? But it probably just makes other checks necessary that can be avoided this way.

fm3: You’re right, that would be cleaner. However, an empty union bbox only happens when an empty list of boxes is passed, which is way more rare than for intersection, where this happens on different bboxes being passed that just happen not to intersect. I think I won’t touch the union code in the scope of this PR. Maybe it will come up again, then I’m open to changing this to None as well :)

PR #6412, opened 2022-08-18
Prevent mag warning from showing in 3D viewport

### LINKED ISSUE #6354: Maximizing 3D viewport triggers mag warning
I hit “.” to maximize the 3D viewport. This toast was shown:
image
I don’t think it applies here, as no actual data is shown in this case, only wireframes + meshes

hotzenklotz:
The PR fix a small isse whereby a toast warning would show up every time one opened the 3D viewport fullscreen. The message is about not having apprirate data for the loaded mag and does not really make sense when the 3D viewport is in fullscreen mode.

This PR prevents this message from showing when maximizing the 3D viewport.

- https://___.webknossos.xyz

### Steps to test:
- Maximize 3D viewport
- Oberserve no more message popping up

### Issues:
- fixes #6354

------
- [x] Ready for review


frontend/javascripts/oxalis/model/sagas/dataset_saga.ts
```diff
@@ -50,9 +51,13 @@ export function* watchZ1Downsampling(): Saga<void> {
     const minVoxelPerPixel = 0.1;
     if (!userClosedWarning) {
       // checking only the downsampled dimensions x and y
+      const is3DViewportFullscreen = yield* select(
```
philippotto, line 54: the `activeViewport` is the one that the mouse hovers currently (or most recently). it's not necessarily in fullscreen then. this means that the toast would likely disappear and re-appear when moving the mouse to the 3d viewport and out of it again.
to get the fullscreen info, I'd do something like the following:

```
const extent_xy = getViewportExtents(state)["PLANE_XY"];
const extent_yz = getViewportExtents(state)["PLANE_YZ"];
...
const areas = [extent_xy, extent_yz, ...].map(([width, height]) => width * height);
const areDataviewportsInvisible = _.sum(areas) == 0;
```

could be a helper function next to `getViewportExtents` in view_mode_accessors.

hotzenklotz: @philippotto I applied your suggestion, please have a look again.

philippotto (approved):
excellent, works well :+1:

PR #6413, opened 2022-08-18
Create unified json format for ObjectId

### LINKED ISSUE #6375: Unify ObjectId json format
We should check if the ObjectId can get a unified json format that reads and writes to just a string, not wrapped in a json object.
This should be relatively easy to create, but we should double check that there are no places where the object is currently expected.

fm3:
When writing an objectId to json (as to string), the result is now just a string (or rather JsString), no longer an `{"id":"myId"}` object.
ObjectId can now also be read from json strings directly..
When reading from Json, it is also validated against the mongo id schema.

I believe we could now replace most occurences of `id.id` by just `id` (not only in json, but also in both sql and strings), but I did not want to create too many line changes in this pr, since there are many other open PRs. Let’s just replace those gradually when editing different areas of the code.

- https://objectidjson.webknossos.xyz

### Steps to test:
- Create a project, save, should work.
- in dataset settings, change allowed teams, should work

### Issues:
- fixes #6375 

---

- [x] Ready for review


leowe (approved):
LGTM :+1: Thanks for doing this.

PR #6414, opened 2022-08-19
Make Largest Segment ID Optional

### LINKED ISSUE #6039: Remove requirement of largestSegmentId
webKnossos currently requires a `largestSegmentId > 0` to be set for each segmentation layer. This was meant to force the user to set the value in order to be able to do volume annotations without re-using existing segment ids. However, often this largest id is not known to the user and the requirement becomes a nuisance and guesses are used. With multiple volume annotation layers, it is now possible to add annotations without overwriting any data.
Also, the `largestSegmentId` of a segmentation layer can be `0` when completely empty. Although, it would be practical to leave out the layer entirely, we shouldn't consider it invalid state.

I propose to allow `0` for `largestSegmentId` and to make the field optional.
Opinions? @fm3 @philippotto

### LINKED ISSUE #6390: preserve volume annotation largestSegmentId and elementClass across re-upload
When downloading and re-uploading a volume annotation, the largestSegmentId and elementClass properties are lost.
In case of an existing fallback layer, the latter is harmless, as the elementClass of the fallback layer is always used. But if there is no fallback layer, the elementClass is set to uint32 in upload, which may not be large enough.

fm3:
Makes the largestSegmentId for segmentation layers optional.
Persists the largestSegmentId for volume annotation layers in the NML

### TODO Backend
 - [x] make largestSegmentId optional in case classes
 - [x] adapt datasource validation
 - [x] create unset value for volumeTracings (cannot use None here because of proto backwards compatibility)
 - [x] volumeTracing largestSegmentId: persist to and read from NML
 - [x] volumeTracing largestSegmentId: make optional

### TODO Frontend
 - [X] when creating a new segment id, check if the volumeTracing’s largestSegmentId is `null`. If so, tell the user about it and provide a way to input a largest segment id
 - [x] Relax the json validation in the dataset edit view
 - [x] Adapt warning message and validation in add-remote-zarr-dataset view

### TODO python client
 - [x] make sure that existing requests don’t break if the segmentation layers do not have a largestSegmentId field https://github.com/scalableminds/webknossos-libs/pull/786

- https://largestsegmentidoptional.webknossos.xyz/auth/login

### Steps to test:
- dataset settings page
  - edit a dataset with a segmentation and set the "largest segment id" field to empty
  - switch to the "advanced" menu and doublecheck that the largest segment id property is either set to null or does not exist
  - add the property there and set it to 10
  - switch back to simple
  - switch back to advanced and remove the property again
  - switch back to simple
  - the field should be empty
  - save
- create an annotation for that dataset (fallback layer should be the segmentation layer without the largest segment id)
  - switch to brush
  - try to brush -> the active segment id should be zero which shouldn't do anything and a toast should warn you about this
  - try to create a new segment id via "c" or the toolbar button --> a toast should appear
  - click the button in the toast and a modal should open which allows to input a largest segment id
  - do so and click ok
  - now, the "c" shortcut should work
- create an annotation without fallback layer --> brushing should just work

### Issues:
- fixes #6039 
- fixes #6390

------
- [x] Ready for review


fm3:
@philippotto Would you be interested in implementing the front-end todos here?

@jstriebel Would you be interested in adapting the python client? I think the important part is that the segmentation layers in the info request may no longer contain the largestSegmentId, which might break existing requests? If the typing is that strict in the front-end.

jstriebel:
> @jstriebel Would you be interested in adapting the python client? I think the important part is that the segmentation layers in the info request may no longer contain the largestSegmentId, which might break existing requests? If the typing is that strict in the front-end.

Yep: https://github.com/scalableminds/webknossos-libs/pull/786

philippotto:
> @philippotto Would you be interested in implementing the front-end todos here?

Yes, I think, I can do that. Not sure when exactly I'll get to it, though :see_no_evil:

webknossos-tracingstore/app/com/scalableminds/webknossos/tracingstore/tracings/volume/VolumeTracingDefaults.scala
```diff
@@ -11,6 +11,8 @@ object VolumeTracingDefaults {
 
   val largestSegmentId = 0L
 
+  val unsetLargestSegmentId = -1L
```
philippotto, line 14: @fm3  would it be possible to deliver `null` instead of `-1` to the front-end? then, we could make the typescript type nullable which would give us type safety.

philippotto: Great, thank you!

fm3:
@philippotto I changed the proto file to allow None for largestSegmentIds. Note that in the update actions they are still not nullable.
When I tried to create an empty volume annotation, I got the please-enter-id prompt. I believe the backend would have set the Id to Some(0). Do you know if the frontend interprets that as null? Otherwise there might still be something wrong in the backend.

philippotto:
> @philippotto I changed the proto file to allow None for largestSegmentIds. Note that in the update actions they are still not nullable.

Can we make the property in the update action nullable? Or does something speak against it?

> When I tried to create an empty volume annotation (no fallback layer), I got the please-enter-id prompt. I believe the backend would have set the Id to Some(0). Do you know if the frontend interprets that as null? Otherwise there might still be something wrong in the backend.

Yes, sorry, there was an error in the front-end when I introduced a temporary logic to handle `-1` as a value.

philippotto:
I think the front-end is ready for review. @fm3 Do you think the back-end is ready, too? Feel free to test the frontend behavior now :) If you agree with the approach, I'd ask for a review.

fm3:
Yes, I’d say the backend is also ready for review.
I tested the behavior and it makes sense to me. 

However, I’d suggest that we explain to the user a bit more what is happening, or guide them towards a safe path. I guess the standard use case is that the user will create an annotation and start brushing. The popup currently only says The segment id is 0, please change it. »Cannot brush with segment id 0. Please change the segment id by [etc] or create a new unused id in the toolbar.« (or something) – where the user will then also be asked to set the largest segment id. It might even be nice to add a link to the dataset edit view, if the user has write access to the dataset, so that they can then set the largest segment id for the on-disk layer. I hope this makes sense, what do you think?

philippotto:
> However, I’d suggest that we explain to the user a bit more what is happening, or guide them towards a safe path. I guess the standard use case is that the user will create an annotation and start brushing. The popup currently only says The segment id is 0, please change it. »Cannot brush with segment id 0. Please change the segment id by [etc] or create a new unused id in the toolbar.« (or something) – where the user will then also be asked to set the largest segment id. It might even be nice to add a link to the dataset edit view, if the user has write access to the dataset, so that they can then set the largest segment id for the on-disk layer. I hope this makes sense, what do you think?

I extended the docs for the largest segment ID and also linked the "dataset edit view" in the new modal. Other than that, I think, that the problem is explained sufficiently in the modal, no? The two new toasts are not too detailed, but I don't think that they should be much longer.

Do you think this is alright?

app/controllers/AnnotationIOController.scala
```diff
@@ -281,10 +281,19 @@ Expects:
           .map(layer => elementClassToProto(layer.elementClass))
           .getOrElse(elementClassToProto(VolumeTracingDefaults.elementClass)),
         fallbackLayer = fallbackLayer.map(_.name),
-        largestSegmentId = fallbackLayer.map(_.largestSegmentId).getOrElse(VolumeTracingDefaults.largestSegmentId)
+        largestSegmentId =
+          combineLargestSegmentIds(volumeTracing.largestSegmentId, fallbackLayer.map(_.largestSegmentId))
```
jstriebel, line 285: `volumeTracing.largestSegmentId` is from the loaded nml and has precedence now, is this correct? It's not clear to me how `adaptPropertiesToFallbackLayer` is used, but probably fine.

app/controllers/AnnotationIOController.scala
```diff
@@ -281,10 +281,19 @@ Expects:
...
       )
     }
 
+  private def combineLargestSegmentIds(fromNml: Option[Long], fromFallbackLayer: Option[Option[Long]]): Option[Long] =
```
jstriebel, line 289: From this name I would have expected some max operation. Why exactly this precedence order, what is the use-case here?

app/models/annotation/AnnotationService.scala
```diff
@@ -155,14 +155,19 @@ class AnnotationService @Inject()(
...
         VolumeTracingDefaults.zoomLevel,
         organizationName = Some(organizationName),
         resolutions = resolutionsRestricted.map(vec3IntToProto)
       )
   }
 
+  private def initialLargestSegmentId(fromFallbackLayer: Option[Option[Long]]): Option[Long] =
+    if (fromFallbackLayer.isDefined) {
+      fromFallbackLayer.flatten
+    } else VolumeTracingDefaults.largestSegmentId
```
jstriebel, line 169: This means the outer option is checked, right? So if there is a fallback layer, its largest segment id is used, even if its None, correct?

webknossos-datastore/app/com/scalableminds/webknossos/datastore/models/datasource/DataLayer.scala
```diff
@@ -63,8 +63,11 @@ object ElementClass extends ExtendedEnumeration {
   }
 
   def largestSegmentIdIsInRange(largestSegmentId: Long, elementClass: ElementClass.Value): Boolean =
-    largestSegmentId >= 0L && segmentationElementClasses.contains(elementClass) && largestSegmentId <= maxSegmentIdValue(
-      elementClass)
+    largestSegmentIdIsInRange(Some(largestSegmentId), elementClass)
+
+  def largestSegmentIdIsInRange(largestSegmentIdOpt: Option[Long], elementClass: ElementClass.Value): Boolean =
+    segmentationElementClasses.contains(elementClass) && largestSegmentIdOpt.forall(largestSegmentId =>
```
jstriebel, line 69: That `=>` had me thinking it's a `>=` for a second^^ :D

app/controllers/AnnotationIOController.scala
jstriebel: PS: The method `combineLargestSegmentIds` in `VolumeTracingService.scala` does what I assumed what this function would do.

fm3:
@jstriebel Thanks for having a look! This is why reviews are important, this code was really quite cryptic :D I added some comments now, hopefully answering your questions. I also merged the precedence methods from AnnotationService and AnnotationIOController into one, and while I was there changed the artificially Fox-y method to synchronous.
`adaptPropertiesToFallbackLayer` changes the uploaded VolumeTracing’s properties, depending on the fallback layer. If there is one, its properties are used, if there is none, the default properties for no-fallback-layer tracings are inserted. I agree that the name does not fully reflect that, but I don’t really have a better idea that doesn’t get ridiculously long. Please have another look, I hope the changes are explained now :)

jstriebel (commented):
Awesome, backend LGTM :+1: Didn't do any testing myself so far.

frontend/javascripts/libs/render_independently.tsx
```diff
@@ -0,0 +1,42 @@
...
+import { document } from "libs/window";
+import { Provider } from "react-redux";
+
+type DestroyFunction = () => void; // The returned promise gets resolved once the element is destroyed.
+
+export default function renderIndependently(
+  getComponent: (arg0: DestroyFunction) => React.ReactElement<React.ComponentProps<any>, any>,
+): Promise<void> {
+  return new Promise((resolve) => {
+    import("oxalis/throttled_store").then((_Store) => {
```
daniel-wer, line 12: Would a normal import cause circular dependencies or why did you choose to perform the dynamic import?

frontend/javascripts/libs/render_independently.tsx
```diff
@@ -0,0 +1,42 @@
...
+        const unmountResult = ReactDOM.unmountComponentAtNode(div);
+
+        if (unmountResult && div.parentNode) {
+          div.parentNode.removeChild(div);
+        }
+
+        resolve();
+      }
+
+      // const ContextProvider = createContextProvider(context);
```
daniel-wer, line 33:
```suggestion
```

frontend/javascripts/oxalis/model/actions/volumetracing_actions.ts
```diff
@@ -87,9 +87,15 @@ export const initializeEditableMappingAction = (mapping: ServerEditableMapping)
     mapping,
   } as const);
 
-export const createCellAction = () =>
+/*
+ * The largestSegmentId parameter is required to enforce that the dispatcher of the action
+ * has dealt with the case where the maximum segment id id is not set. In that case,
```
daniel-wer, line 92:
```suggestion
 * has dealt with the case where the maximum segment id is not set. In that case,
```

frontend/javascripts/oxalis/model/reducers/volumetracing_reducer_helpers.ts
```diff
@@ -47,22 +47,19 @@ export function setActiveCellReducer(state: OxalisState, volumeTracing: VolumeTr
     activeCellId: id,
   });
 }
-export function createCellReducer(state: OxalisState, volumeTracing: VolumeTracing, id?: number) {
-  if (id === 0) {
-    // cellId 0 means there is no annotation, so there must not be a cell with id 0
-    return state;
-  }
```
daniel-wer, line 54: Was this check no longer needed?

frontend/javascripts/oxalis/model/sagas/mapping_saga.ts
```diff
@@ -234,6 +234,12 @@ function* getLargestSegmentId(layerName: string): Saga<number> {
     throw new Error("Mappings only exist for segmentation layers.");
   }
 
+  if (segmentationLayer.largestSegmentId == null) {
+    // todo: fix this once the custom segment colors are merged into master
```
daniel-wer, line 238: Reminder to fix this ToDo.

frontend/javascripts/oxalis/model/sagas/update_actions.ts
```diff
@@ -109,7 +109,8 @@ type UpdateVolumeTracingUpdateAction = {
     activeSegmentId: number;
     editPosition: Vector3;
     editRotation: Vector3;
-    largestSegmentId: number;
+    // todo: clarify that the backend accepts this
```
daniel-wer, line 112: As far as I can see this was clarified.

```suggestion
```

frontend/javascripts/test/reducers/volumetracing_reducer.spec.ts
```diff
@@ -17,6 +17,7 @@ mockRequire("app", {
   },
 });
 test("VolumeTracing should set a new active cell", (t) => {
+  // todo
```
daniel-wer, line 20: Is this still a ToDo?

frontend/javascripts/oxalis/view/largest_segment_id_modal.tsx
```diff
@@ -0,0 +1,97 @@
...
+
+      <p>
+        If you know the largest segment ID in this segmentation layer or if you know which value
+        would be safe to use, please input it below:
+      </p>
+      <div style={{ display: "grid", placeItems: "center" }}>
+        <InputNumber
+          size="large"
+          min={1}
+          max={100000}
```
daniel-wer, line 88: This seems to small, maybe a copy & paste remnant. There should either be no max or, better, it should depend on the data type of the segmentation layer.

daniel-wer (commented):
Nice changes :+1: Let me know if you think this should be tested again :)

frontend/javascripts/oxalis/model/reducers/volumetracing_reducer_helpers.ts
philippotto: The previous signature allowed to pass an explicit `id` which should be set for the active segment id. This was only used by the `INITIALIZE_VOLUMETRACING` reducer. I think, this made the code a bit hard to understand and also goes against the name of `createCellReducer`.

Therefore, I adapted the function so that its only purpose is to really create a new cell id. Therefore, the old `id` parameter doesn't exist, anymore, and doesn't need to be checked against `0` anymore. Hope that makes sense :)

conf/messages
```diff
@@ -228,7 +228,7 @@ tracing.wrongMag=Tracing “{0}” does not have mag “{1}”
 
 annotation.create.failed=Failed to create annotation.
 annotation.create.forbidden=You do not have permission to create annotations for this dataset.
-annotation.volume.negativeLargestSegmentId=The resulting volume annotation would have an invalid largest segment id (must be nonnegative, within element class range, and < 2^53)
+annotation.volume.invalidLargestSegmentId=Cannot create tasks with fallback segmentation layers that do not have a valid largest segment id.
```
philippotto, line 231: @fm3  This is only a theoretical warning, right? e.g, if we would allow to create volume tasks by specifying a layer, this would come up (not possible currently).
Just asking, because I just checked volume tasks by specifying an annotation ID or a zip as a base and it worked.

fm3: Yes, I think this is not currently possible. But some of the code paths already exist and I wanted to make sure they’re guarded

philippotto: ok, great :+1:

frontend/javascripts/oxalis/model/sagas/mapping_saga.ts
philippotto: This is resolved now. Since the new custom segment color mechanism doesn't need reassignment of segment ids (the old code had to reassign the IDs so that the texture access was aligned with its 256**2 dimension), I don't even need to access the largest segment anymore. This is quite lucky, as otherwise I would have had to disable mappings with custom colors for segmentations with unknown largest segment id.
By the way, I also added an analytics event so that we know whether custom mapping colors are used at all.

philippotto (approved):
This PR is ready from my point of view. @fm3 Do you agree? I just tested it a bit and it seems good to go for me. Maybe you want to have a look, too. Then, we could merge&deploy on Monday, I'd say.

PR #6415, opened 2022-08-19
New worker job: Find Largest Segment ID

fm3:
wk side for https://github.com/scalableminds/webknossos-worker/pull/114

- https://___.webknossos.xyz

### Steps to test:
- Edit a dataset. Should look like this (if the largest segment id is empty):
image
- Click the Detect button
image
image

- Should show a toast:
image
- After reloading:
image
(the message would typically be something like `Updated largest segment id from 1 to 1470623.`)

It would be even better if the largest segment id was automatically updated in the form, but for long running jobs this wouldn't really matter (as one can close the tab) and the interface of the API doesn't really lend itself to that currently.

### Issues:
- follow-up for #6039 

------
- [X] Ready for review


fm3:
@philippotto I adapted the job list view so that it displays the returned result message in the actions column. I’m aware that it does not really fit the column title but I don’t think creating a new column is worth it here. What do you think? Also, where do you think would be a good place for a button to trigger this job? Maybe you could have a look at the front-end part here?

philippotto:
I added some prototypical code to trigger the job (mainly so that I could test the worker). however, for a proper integration, I'd like to wait until #6485 is merged.

philippotto:
cc @fm3 the frontend is ready for review. you might want to request a review for the backend, too :)

frontend/javascripts/dashboard/dataset/dataset_settings_data_tab.tsx
```diff
@@ -312,6 +323,21 @@ function SimpleLayerForm({
...
+        ? () =>
+            startFindLargestSegmentIdJob(datasetId.name, datasetId.owningOrganization, layer.name)
+        : null,
+    findJobPred: (job) =>
+      job.type === "find_largest_segment_id" && job.datasetName === datasetId?.name,
+    jobStartedMessage:
+      "A job was scheduled to compute the largest segment ID. It will be automatically updated for the dataset. You may close this tab now.",
+    successMessage:
+      "The computation of the largest segment id for this dataset has finished. Reload the page to see it.",
+    failureMessage: "The computation of a mesh file for this dataset didn't finish properly.",
```
daniel-wer, line 338:
```suggestion
    failureMessage: "The computation of the largest segment id for this dataset didn't finish properly.",
```

frontend/javascripts/admin/job/job_hooks.ts
```diff
@@ -0,0 +1,107 @@
...
+  activeJob: APIJob | null;
+  mostRecentSuccessfulJob: APIJob | undefined;
+} {
+  const activeUser = useSelector((state: OxalisState) => state.activeUser);
+  const [jobs, setJobs] = useState<APIJob[] | null>(null);
+  const [activeJob, setActiveJob] = useState<APIJob | null>(null);
+  const areJobsEnabled = features().jobsEnabled;
+  const potentialJobs = (jobs || []).filter(findJobPred);
+  const mostRecentSuccessfulJob = potentialJobs.find(
+    (job) => findJobPred(job) && job.state === "SUCCESS",
```
daniel-wer, line 35: Shouldn't `(job) => findJobPred(job)` be true for all jobs in `potentialJobs`?

daniel-wer (commented):
Great abstraction and UX as always! :)

I did not test this PR, mainly because I don't have a running worker setup locally, atm.

frontend/javascripts/admin/job/job_hooks.ts
```diff
@@ -0,0 +1,107 @@
+import features from "features";
+import { getJobs } from "admin/admin_rest_api";
+import { useInterval } from "beautiful-react-hooks";
+import Toast from "libs/toast";
+import { OxalisState } from "oxalis/store";
+import { useEffect, useState } from "react";
+import { useSelector } from "react-redux";
+import { APIJob } from "types/api_flow_types";
+
+export function useStartAndPollJob({
```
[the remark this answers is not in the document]
philippotto: Good point! I created a new issue: https://github.com/scalableminds/webknossos/issues/6599
I think, refactoring existing code might not be worth it (especially, since the hook cannot be used in class-components), but especially for new job-integrations, we should keep this in mind :)

fm3:
I can’t approve my own PR but I’d say if philipp tested that the backend passes the parameters through so that the job runs as expected, the backend does not need further review :) :shipit:

philippotto (approved):

fm3:
I did not look at the front-end code in deatil, but one question just to make sure: is this disabled if jobs are not enabled for the instance and the dataset? (I’m asking with https://github.com/scalableminds/webknossos/issues/6582 in mind)

PR #6416, opened 2022-08-19
Voxelytics routes

normanrz:
Adds the backend code for the Voxelytics workflow viewer to webKnossos.

### Steps to test:
- Checkout the https://github.com/scalableminds/webknossos/pull/6460 PR and test there

------
- [x] Ready for review


app/controllers/VoxelyticsController.scala
```diff
@@ -0,0 +1,283 @@
...
+
+      for {
+        runId <- voxelyticsDAO.getRunIdByName(runName, request.identity._organization) ~> NOT_FOUND
+        _ <- voxelyticsService.checkAuth(runId, request.identity) ~> UNAUTHORIZED
+        _ <- Fox.sequence(request.body.map(event => createWorkflowEvent(runId, event)))
+      } yield Ok
+    }
+
+  @ApiOperation(hidden = true, value = "")
+  def getChunkStatistics(workflowHash: String, runIdS: String, taskName: String): Action[AnyContent] =
```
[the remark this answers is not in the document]
normanrz: It's meant for String. Ideally the type in the signature would be `ObjectId` directly

app/models/voxelytics/ElasticsearchClient.scala
```diff
@@ -0,0 +1,95 @@
...
+class ElasticsearchClient @Inject()(wkConf: WkConf, rpc: RPC)(implicit ec: ExecutionContext)
+    extends LazyLogging
+    with MimeTypes {
+
+  private lazy val conf = wkConf.Voxelytics.Elasticsearch
+  val SCROLL_SIZE = 10000
+
+  private def pollUntilServerIsAvailable: Fox[Unit] =
+    // TODO: Implement polling with conf.startupTimeout
+    Fox.successful()
```
normanrz, line 26: @fm3 I need some help here. In JS I implemented a function that waits until the elasticsearch server is available and initializes the index (if necessary). The function result is memoized, so all callers get the same "promise". A promise is probably the same as a future in Scala. Now, I don't know how to implement `setTimeout` in Scala. I think you mentioned, I need an actor system for that? https://github.com/scalableminds/voxelytics/blob/master/voxelytics/workflows/workflow-viewer/server/logging.js#L10-L62

normanrz (commented):
Thanks for the feedback! I went all your comments.

app/models/voxelytics/ElasticsearchClient.scala
normanrz: I think I figured it out now. Not sure what the `lifecycle` does, though. https://github.com/scalableminds/webknossos/pull/6416/files#diff-078a862b7f24a708ab9cb5323ad2768985ed7a68dd2e32540273717a05ecf723R23-R123

app/utils/SQLHelpers.scala
```diff
@@ -119,6 +119,34 @@ class SimpleSQLDAO @Inject()(sqlClient: SQLClient)(implicit ec: ExecutionContext
     }
   }
 
+  def escapeLiteral(aString: String): String = {
```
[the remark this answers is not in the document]
fm3: I wrote follow-up issue #6455

conf/application.conf
```diff
@@ -82,6 +82,7 @@ features {
   allowDeleteDatasets = true
   # to enable jobs for local development, use "yarn enable-jobs" to also activate it in the database
   jobsEnabled = false
+  voxelyticsEnabled = true
```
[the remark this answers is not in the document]
normanrz: No, that's just for development.

app/RequestHandler.scala
```diff
@@ -27,10 +39,15 @@ class RequestHandler @Inject()(webCommands: WebCommands,
...
     with InjectedController
-    with LazyLogging {
+    with LazyLogging
+    with AdditionalHeaders {
 
   override def routeRequest(request: RequestHeader): Option[Handler] =
-    if (request.uri.matches("^(/api/|/data/|/tracings/|/swagger).*$")) {
+    if (request.method == "OPTIONS") {
+      Some(Action {
+        options(request)
```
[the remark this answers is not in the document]
fm3: I’d refer that to our security engineers @daniel-wer and @frcroth – my guess is that it’s fine if every route gets cors headers (that would actually simplify the code as well), but I am unfamiliar with xss attack scenarios.

However, many of the actual routes don’t have the cors headers at this time, but the options request lists it for all. I’m not sure if that might cuase problems.

app/models/voxelytics/VoxelyticsService.scala
```diff
@@ -0,0 +1,205 @@
...
+import scala.util.Try
+
+case class RunEntry(id: ObjectId,
+                    name: String,
+                    username: String,
+                    hostname: String,
+                    voxelyticsVersion: String,
+                    workflow_hash: String,
+                    workflow_yamlContent: String,
+                    workflow_config: JsObject,
```
fm3, line 21: I guess the mix of snake case and camel case stems from what we get from voxelytics? Then that’s fine

app/models/voxelytics/VoxelyticsService.scala
```diff
@@ -0,0 +1,205 @@
...
+      val artifacts = artifactKV._2
+      (taskName, JsObject(artifacts.map(artifactEntryPublicWrites)))
+    }))
+  }
+
+  def workflowConfigPublicWrites(workflowConfig: JsObject, tasks: List[TaskEntry]): JsObject =
+    workflowConfig ++ Json.obj("tasks" -> JsObject(tasks.map(t => (t.name, t.config ++ Json.obj("task" -> t.task)))))
+
+  def aggregateBeginEndTime(runs: List[RunEntry]): (VoxelyticsRunState, Instant, Option[Instant]) = {
+    val state = runs.maxBy(_.beginTime).state
```
fm3, line 152: maxBy throws an exception when called on an empty list. What is the expected behavior in this case?

app/models/voxelytics/VoxelyticsService.scala
```diff
@@ -0,0 +1,205 @@
...
+    (state, beginTime, endTime)
+  }
+
+  def combineTaskRuns(allTaskRuns: List[TaskRunEntry], mostRecentRunId: ObjectId): List[TaskRunEntry] =
+    allTaskRuns
+      .filter(task => task.runId == mostRecentRunId)
+      .map(task => {
+        val thisTaskRuns = allTaskRuns.filter(t => t.taskName == task.taskName).sortBy(_.beginTime)
+        val nonWaitingTaskRuns = thisTaskRuns.filter(t => {
+          t.state == VoxelyticsRunState.RUNNING || t.state == VoxelyticsRunState.COMPLETE || t.state == VoxelyticsRunState.FAILED || t.state == VoxelyticsRunState.CANCELLED
```
fm3, line 165: could there be a semantic name for this list of states? if so, this could become something like `VoxelyticsRunState.verySpecialStates.contains(t.state)`

app/models/voxelytics/VoxelyticsService.scala
```diff
@@ -0,0 +1,205 @@
...
+        })
+        if (nonWaitingTaskRuns.nonEmpty) {
+          nonWaitingTaskRuns.head
+        } else {
+          thisTaskRuns.head
+        }
```
fm3, line 171:
```suggestion
        nonWaitingTaskRuns.headOption.getOrElse(thisTaskRuns.head)
```
(though It would of course be great if the second `.head` could also be replaced by headOption, but that may change the function signature)

app/models/voxelytics/VoxelyticsService.scala
```diff
@@ -0,0 +1,205 @@
...
+      })
+
+  def upsertTask(runId: ObjectId,
```
fm3, line 174:
```suggestion
  def upsertTaskWithArtifacts(runId: ObjectId,
```

package.json
```diff
@@ -122,6 +122,7 @@
     "lint-no-cache": "node_modules/.bin/eslint --quiet --ext .ts,.tsx frontend/javascripts/",
     "pretty": "node_modules/.bin/prettier --write --config .prettierrc \"frontend/javascripts/**/*.ts\" \"frontend/javascripts/**/*.tsx\" \"tools/**/*.js\"",
     "pretty-backend": "sbt \";scalafmt; util/scalafmt; webknossosTracingstore/scalafmt; webknossosDatastore/scalafmt\"",
+    "lint-backend": "sbt \";scapegoat; scalafmtCheck; util/scalafmtCheck; webknossosTracingstore/scalafmtCheck; webknossosDatastore/scalafmtCheck\"",
```
fm3, line 125: Did you use this locally? I tend to just run pretty-backend, which mutates the file.
Maybe this could also be used in the CI directly?

app/models/voxelytics/ElasticsearchClient.scala
```diff
@@ -0,0 +1,192 @@
...
+      scrollId = (scroll \ "_scroll_id").as[String]
+      scrollHits = (scroll \ "hits" \ "hits").as[List[JsValue]]
+      _ = buffer ++= scrollHits
+      lastScrollId <- fetchBatch(buffer, scrollId)
+      _ <- rpc(s"${conf.uri}/_search/scroll/$lastScrollId").delete()
+    } yield JsArray(buffer)
+
+  }
+
+  private def fetchBatch(buffer: ListBuffer[JsValue], scrollId: String): Fox[String] =
```
fm3, line 159:
```suggestion
  private def fetchBatchToBuffer(bufferMutable: ListBuffer[JsValue], scrollId: String): Fox[String] =
```
Since mutating passed variables is a very special case in the scala backend, I’d like to make it immediately recognizable (also goes for the other functions)

app/models/voxelytics/ElasticsearchClient.scala
```diff
@@ -0,0 +1,192 @@
...
+
+    for {
+      _ <- serverStartupFuture
+      scroll <- rpc(s"${conf.uri}/${conf.index}/_search?scroll=1m")
+        .postJsonWithJsonResponse[JsValue, JsValue](scrollBody) ~> "Could not fetch logs"
+      scrollId = (scroll \ "_scroll_id").as[String]
+      scrollHits = (scroll \ "hits" \ "hits").as[List[JsValue]]
+      _ = buffer ++= scrollHits
+      lastScrollId <- fetchBatch(buffer, scrollId)
+      _ <- rpc(s"${conf.uri}/_search/scroll/$lastScrollId").delete()
```
fm3, line 154: will there be a lot of these? if so, it might make sense to add `.silent` to the rpc to avoid spam in the backend logging

app/models/voxelytics/ElasticsearchClient.scala
```diff
@@ -0,0 +1,192 @@
...
+      } else {
+        logger.info("Elasticsearch is available.")
+        Fox.successful(())
+      }
+    } yield ()
+  }
+
+  private def bootstrapIndexOnServer: Fox[Unit] =
+    for {
+      indexExists <- rpc(s"${conf.uri}/${conf.index}").head.map(_ => true).getOrElse(false)
```
fm3, line 116: Took me a while to read that, I didn’t even know Fox had a getOrElse! But well, fair enough :D

app/models/voxelytics/ElasticsearchClient.scala
```diff
@@ -0,0 +1,192 @@
...
+      _ <- if (!indexExists) {
+        logger.info("Bootstrapping Elasticsearch index for Voxelytics.")
+        rpc(s"${conf.uri}/${conf.index}").put(elasticsearchSchema)
+      } else {
+        Fox.successful(())
+      }
```
fm3, line 122:
```suggestion
      _ <- Fox.runIf(!indexExists) {
        logger.info("Bootstrapping Elasticsearch index for Voxelytics.")
        rpc(s"${conf.uri}/${conf.index}").put(elasticsearchSchema)
      }
```

fm3 (commented):
This is moving in a very good direction :) You have been busy! I added a few more comments to the backend code. Looking forward to testing with the frontend next.

package.json
normanrz: It also runs `scapegoat`, which I guess is the linter? Perhaps the `scalafmt` doesn't belong in this call

app/models/voxelytics/VoxelyticsService.scala
fm3: alright :) maybe add a comment in this line that the callers take care of this

app/RequestHandler.scala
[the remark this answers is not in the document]
normanrz: After educating myself again about XSRF, I would rather allow remote origin access everywhere. It should be limited to selected routes, preferably public routes. I'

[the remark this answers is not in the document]
fm3: > I would rather allow remote origin access everywhere

I think there may be a “not” missing somewhere in there ;) Ok cool. I just double-checked and on the wk side (where browsers will send cookies) there is only one route that allows remote origin: /api/buildinfo. If I understand correctly, the datastore/tracingstore side, which is authenticated with GET parameter token or X-Auth-Token header should be safe from XSRF.

app/models/voxelytics/ElasticsearchClient.scala
fm3: - The lazy val should work as a memoized future, that looks good to me and seems to work well :+1: 
- I don’t think you’ll need the ApplicationLifecycle for this.
- Play will sometimes instantiate multiple instances of injected classes, and garbage-collect them again later. To ensure that this function is never run superfluously, you could add ElasticSearchClient to the `WebKnossosModule` setup, and bind it as singleton there. This is already done for a couple of other stateful services to enable shared caching. Restart after incremental compilation will still re-run everything in them.
- The `akka.pattern.after` seems like a nice way to sleep without blocking the thread from the thread pool. Didn’t know about this :)

fm3 (approved):
Backend LGTM and seems to work :) I will admit I did not test every code path, and we will likely find more things as this lives on, but I’d say for a first version this is ready for some dogfooding.

Please add a changelog entry and migration guide, remove the for-development-only config and some debug logging (I think), then when the frontend is approved too this should be good to go :)

app/models/voxelytics/ElasticsearchClient.scala
normanrz: Thanks for looking into this!

PR #6422, opened 2022-08-24
Small Zarr Fixes

fm3:
While working on S3 without credentials, I noticed two other zarr bugs that this PR addresses
 - the `.array()` call in the bytes converter is superfluous (copy paste relic), and for float datasets throws exceptions.
 - .zarray files with missing values could not be correctly parsed, due to missing `WithDefaultValues`

### Steps to test:
- I tested that I could now loat a float zarr dataset
- I tested that reading remote zarr datasets with missing `dimension_separator` key in .zarray files works again

------
- [x] Ready for review


leowe (approved):
Good catch :+1:

PR #6423, opened 2022-08-25
Get shortest path between two nodes

### LINKED ISSUE #6349: Create tree for shortest path between two points
## Detailed Description
If one node is active and the user rightclicks another node that is in the same tree, there could be a context menu entry à la "Extract shortest path to this node" (better wording welcome). We already have a JS api method for `measurePathLengthBetweenNodes`. The code could be (copied and) adapted to not return a scalar, but to create a new tree.

This feature was requested from external users.

## Context

Dagobert42:
- https://___.webknossos.xyz

### Steps to test:
- create a tree
- select node
- open context menu on another node in the same tree
- extract shortest Tree to this node

### Issues:
- fixes #6349 

------
- [x] Ready for review

### DIFF
--- a/frontend/javascripts/oxalis/api/api_latest.ts
+++ b/frontend/javascripts/oxalis/api/api_latest.ts
@@ -747,9 +747,18 @@ class TracingApi {
   }
 
   /**
-   * Returns the length of the shortest path between two nodes in nanometer and in voxels.
+   * Returns the length of the shortest path between two nodes in nanometer and in voxels
+   * as well as the shortest path between two nodes as an array of node IDs.
+   * Note that the array is ordered from target to source node.
    */
-  measurePathLengthBetweenNodes(sourceNodeId: number, targetNodeId: number): [number, number] {
+  measurePathLengthBetweenNodes(
+    sourceNodeId: number,
+    targetNodeId: number,
+  ): {
+    lengthNm: number;
+    lengthVx: number;
+    shortestPath: number[];
+  } {
     const skeletonTracing = assertSkeleton(Store.getState().tracing);
     const { node: sourceNode, tree: sourceTree } = getNodeAndTreeOrNull(
       skeletonTracing,
@@ -775,6 +784,10 @@ class TracingApi {
     // used when returning the final distance. The actual path finding is only done in
     // the physical space (nm-based).
     const distanceMapVx: Record<number, number> = {};
+    // We keep track of each nodes parent node ID in order to reconstruct an array of
+    // node IDs for the shortest path.
+    const parentMap: Record<number, number> = {};
+    parentMap[sourceNode.id] = -1;
 
     const getDistance = (nodeId: number) =>
       distanceMap[nodeId] != null ? distanceMap[nodeId] : Number.POSITIVE_INFINITY;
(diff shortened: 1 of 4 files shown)


frontend/javascripts/oxalis/api/api_latest.ts
```diff
@@ -747,9 +747,18 @@ class TracingApi {
...
 
   /**
-   * Returns the length of the shortest path between two nodes in nanometer and in voxels.
-   */
-  measurePathLengthBetweenNodes(sourceNodeId: number, targetNodeId: number): [number, number] {
+   * Returns the length of the shortest path between two nodes in nanometer and in voxels
+   * as well as the shortest path between two nodes as an array of node IDs.
+   * Note that the array is ordered from target to source node.
+   */
+  measurePathLengthBetweenNodes(
```
philippotto, line 754: this is a backwards-incompatible change right now. I'd rename this function and add another `measurePathLengthBetweenNodes` function which delegates to this implementation and simply extracts the lengthVx and lengthNm parameter to return it.

frontend/javascripts/oxalis/model/reducers/skeletontracing_reducer_helpers.ts
```diff
@@ -859,3 +859,25 @@ export function removeMissingGroupsFromTrees(
   });
   return changedTrees;
 }
+export function extractPathAsNewTree(
+  state: OxalisState,
+  sourceTree: Tree,
+  pathOfNodeIds: number[],
+): Maybe<Tree> {
+  return createTree(state, Date.now()).map((newTree) => {
```
philippotto, line 867:
```suggestion
  return createTree(state, Date.now(), true, `Shortest path from node ${pathOfNodeIds[0]} to ${...}`).map((newTree) => {
```

the `true` is so that the tree is added to the active group. the fourth parameter doesn't exist yet, but it would be cool if you could add it as an optional name parameter. Then, the tree would get a nice descriptive name :)

frontend/javascripts/oxalis/view/context_menu.tsx
```diff
@@ -420,6 +439,19 @@ function NodeContextMenuOptions({
...
+            <Menu.Item
+              key="extract-shortest-path"
+              disabled={activeNodeId == null || !areInSameTree || isTheSameNode}
+              onClick={() =>
+                activeNodeId != null
+                  ? extractShortestPathAsNewTree(clickedTree, activeNodeId, clickedNodeId)
+                  : null
+              }
+            >
+              Extract shortest Tree to this Node
```
philippotto, line 452:
```suggestion
              Extract shortest Path to this Node
```

frontend/javascripts/oxalis/api/api_latest.ts
```diff
@@ -747,9 +747,18 @@ class TracingApi {
   }
 
   /**
-   * Returns the length of the shortest path between two nodes in nanometer and in voxels.
-   */
-  measurePathLengthBetweenNodes(sourceNodeId: number, targetNodeId: number): [number, number] {
+   * Returns the length of the shortest path between two nodes in nanometer and in voxels
+   * as well as the shortest path between two nodes as an array of node IDs.
+   * Note that the array is ordered from target to source node.
```
philippotto, line 752: the order from target to source is a bit weird. I'd swap it. you can do so by changing `push` to `unshift` at the end of the function when iterating across the parents.

philippotto (changes requested):
cool stuff :) just a few nitpicks..

philippotto (commented):
Great! Only the changelog entry is missing now :)

philippotto (approved):
:tada:

PR #6431, opened 2022-08-29
Make Annotation Zarr Endponits Compatible With Standard WK Auth

### LINKED ISSUE #6418: Adapt annotation zarr streaming endpoint to also accept annotation id and normal user token
The zarr private links are path-only URIs to enable streaming to other tools. However, this circumvents the normal wk auth.
When we have full control of the client, it should also be possible to stream annotations as zarr with the normal auth mechanism (user token as get parameter). The lookup should then be by annotation id rather than access id.

fm3:
The zarr annotation streaming backend routes on the datastore now also take annotation id instead of just private link access tokens. In this case, a normal user token can be supplied to allow access to more than just public annotations.

- https://annotationzarr.webknossos.xyz

### Issues:
- fixes #6418

------
- [x] Ready for review


leowe (approved):
Nice refactor! I like the now more consistent access validation in the ZarrStreamingController.

leowe (approved):
Good catch :+1:

PR #6433, opened 2022-08-30
Fix clip histogram button if ideal mag does not exist

### LINKED ISSUE #6432: Auto-Clip histogram while layer is not rendered due to missing mag breaks core UI
The layer in question was only available in mag1, and I was zoomed out further (which was possible because other layers were available in more mags). I clicked the auto-clip button of said layer, and the UI turned red and an alert said Internal error. »Please reload the page to avoid losing data.«

I would assume the button to just do nothing if there is no data available for its logic (or possibly even show a toast explaining that I need to zoom)

philippotto:
- https://___.webknossos.xyz

### Steps to test:
- test the clip histogram button with a dataset where one layer only has mag 1 and another layer has higher mags
- zoom out very far (e.g., mag 8) and use the clip-histogram for the layer which only has mag 1
- --> a toast should explain that the action didn't succeed and to zoom in
- zoom in further (e.g., mag 4) --> the button should work (for me, it loaded 80 MB of data which is quite a lot but alright for this edge case in my opinion)

### Issues:
- fixes #6432

------
- [X] Ready for review


philippotto:
@fm3 I couldn't really reproduce the exact error scenario you experienced in your issue (do you happen to have an error message?). Initially, I thought that a non existent mag was used, but the code did actually deal with that correctly. However, I noticed that there was no guard against loading very large volumes of data (e.g., only mag 1 exists, but you click on the layer in mag 32). Therefore, I adapted the code to gracefully handle this situation. I hope this helps!

fm3:
> I couldn't really reproduce the exact error scenario you experienced

For the record, as discussed in slack, my issue was not tested on the latest master and would with the newer code have instead triggered exactly what this issue solves 👍

philippotto:
@fm3 maybe you can have a look at the diff :)

frontend/javascripts/oxalis/api/api_latest.ts
```diff
@@ -1355,31 +1355,40 @@ class DataApi {
...
-    if (maybeResolutionIndex == null) {
-      zoomStep = getRequestLogZoomStep(state);
-    } else {
-      const layer = getLayerByName(state.dataset, layerName);
-      const resolutionInfo = getResolutionInfo(layer.resolutions);
-      zoomStep = resolutionInfo.getClosestExistingIndex(maybeResolutionIndex);
+    const resolution = resolutionInfo.getResolutionByIndexOrThrow(zoomStep);
+    const resolutionUVX = dimensions.transDim(resolution, viewport);
+    const widthInVoxel = Math.ceil(halfViewportExtentU * resolutionUVX[0]);
+    const heightInVoxel = Math.ceil(halfViewportExtentV * resolutionUVX[1]);
```
philippotto, line 1385:
```suggestion
    const widthInVoxel = Math.ceil(halfViewportExtentU / resolutionUVX[0]);
    const heightInVoxel = Math.ceil(halfViewportExtentV / resolutionUVX[1]);
```

fm3 (approved):
Works for me :)

Issue #5639, opened 2021-07-27
Make datasource editor schema errors human readable


hotzenklotz:
## Detailed Description
Ich no comprende :-(
<img width="1311" alt="image" src="https://user-images.githubusercontent.com/1105056/127154173-39bb9334-0b21-4a30-b874-958c27407750.png">

These schema errors are not helpful at all. 

## Context
I sometimes need to edit the data-source.json thingy manually from the advanced web UI. It is nice that the there is some validation... however the error message as abysmal. Rather then telling me it is not valid, it should me tell either what property is missing or which property value is invalid.

* [ ] Specific to long-running jobs (set `jobsEnabled=true` in `application.conf`)
* [ ] Specific to webKnossos.org (set `isDemoInstance=true` in `application.conf`)

PR #6434, opened 2022-08-30
Add JSON schema for zarr datasources

philippotto:
- https://___.webknossos.xyz

### Steps to test:
- edit the settings for a zarr dataset (e.g., add `https://data-humerus.webknossos.org/data/zarr/scalable_minds/l4dense_motta_et_al_demo/segmentation` via the new ZARR import front-end)
- editing the JSON should work without "invalid JSON" errors

### Issues:
- follow-up #6383 

------
- [X] Ready for review


frontend/javascripts/types/schemas/datasource.schema.ts
```diff
@@ -64,36 +65,44 @@ export default {
       },
       required: ["dataFormat", "boundingBox", "wkwResolutions"],
     },
-    "types::DataLayerKnossosPartial": {
```
philippotto, line 67: @fm3 there's no use in keeping the KNOSSOS spec here, right?

frontend/javascripts/types/schemas/datasource.schema.ts
```diff
@@ -64,36 +65,44 @@ export default {
...
               },
-              boundingBox: {
-                $ref: "#/definitions/types::BoundingBox",
+              axisOrder: {
+                type: "object",
+                additionalProperties: { type: "number" },
               },
             },
-            required: ["name", "resolutions", "boundingBox"],
+            required: ["mag", "path"],
```
fm3, line 101: Note that `path` in the mag object is not required. Another optional field within the mag object is `credentials`.      example:
```    "credentials": {
            "user": "asdf",
            "password": "asdf"
          }
```

philippotto: thanks, will update that! out of curiosity: where would wk fetch the data from if the path is not provided?

fm3: it will look on the local datastore disc, relative to the dataset dir in `layer_name/2-2-1` (short form for scalar mags, i.e. `layer_name/1`)

philippotto: ah, makes sense :+1: thanks!

fm3 (approved):
Works for me :) I have to say that the error messages are still not particularly readable, but that’s a separate issue I guess.

philippotto:
> I have to say that the error messages are still not particularly readable, but that’s a separate issue I guess.

Yes, that's https://github.com/scalableminds/webknossos/issues/5639. Unfortunately, the jsonschema package doesn't really help with readable error messages..
I'd rather put more work into having a form UI so that the user doesn't have to edit the JSON in the first place.

PR #6435, opened 2022-08-30
Migrations: fix sql links

jstriebel:
Minor change, fixing links to sql evolutions in the migrations document, see also https://docs.webknossos.org/webknossos/MIGRATIONS.released.html


fm3 (approved):
Thanks!

PR #6438, opened 2022-08-31
Use sharing token from URL for sharing links if it exists

philippotto:
- https://___.webknossos.xyz

### Steps to test:
- open a private dataset
- create a sharing link (A)
- open that link in an incognito window
- click the share button -> the copied url should contain the same token as in (A)

### Issues:
- see https://scm.slack.com/archives/C5AKLAV0B/p1660204300742279?thread_ts=1660202133.556449&cid=C5AKLAV0B

------
- [X] Ready for review

### DIFF
--- a/frontend/javascripts/oxalis/view/action-bar/share_modal_view.tsx
+++ b/frontend/javascripts/oxalis/view/action-bar/share_modal_view.tsx
@@ -16,6 +16,7 @@ import {
   editAnnotation,
   sendAnalyticsEvent,
   setOthersMayEditForAnnotation,
+  getSharingTokenFromUrlParameters,
 } from "admin/admin_rest_api";
 import TeamSelectionComponent from "dashboard/dataset/team_selection_component";
 import Toast from "libs/toast";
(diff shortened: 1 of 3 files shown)


daniel-wer (commented):
Hmm, did this work for you during testing? For me, the copied link in the incognito tab doesn't contain a token, neither for a shared dataset nor for an annotation.

philippotto:
> Hmm, did this work for you during testing? For me, the copied link in the incognito tab doesn't contain a token, neither for a shared dataset nor for an annotation.

Sorry for the trouble! It worked for me when I opened the PR, but apparently I must have been logged in when I tried to acquire the link. Maybe I didn't use incognito or I was already logged-in in my incognito session (since chrome only supports one incognito session at the same time).

Now it should work :)

frontend/javascripts/oxalis/view/action-bar/share_modal_view.tsx
```diff
@@ -84,11 +87,8 @@ export function useDatasetSharingToken(dataset: APIDataset) {
   };
 
   useEffect(() => {
-    if (!activeUser) {
-      return;
-    }
     getAndSetToken();
-  }, [dataset, activeUser]);
+  }, [dataset]);
```
daniel-wer, line 91: I think, technically, `activeUser` should still be part of the dependencies since `getAndSetToken` should be re-executed if it changes. At least that's my understanding of it, not sure if the linter complains about that :thinking:

daniel-wer (approved):
Works like a charm now :+1:

frontend/javascripts/oxalis/view/action-bar/share_modal_view.tsx
philippotto: Good point! I added it now :)

PR #6440, opened 2022-09-01
Remove wrongly-typed field from Mailchimp tag response

fm3:
The id field in mailchimp tag listing responses is not a string, but int. This lead to the response parsing failing :facepalm: We don’t even use the id.

This should fix the tagging for inactive/active users after the third week. (see MailchimpTicker.scala)

### Steps to test:
- I tested that the response can now be parsed and that the contains checks for existing tags works.

- [x] Ready for review


jstriebel (approved):
:shipit:

PR #6441, opened 2022-09-01
Remove some more return statements from scala code

fm3:
### Issues:
- contributes to #6356 

- [x] Ready for review


leowe (approved):
LGTM 👍

PR #6214, opened 2022-05-17
Create mailchimp tags for inactive/active users after three weeks

fm3:
Merge this after June 1 – Blocked by #6213 – that one needs to be deployed for at least two weeks before this PR makes sense

Tag users in MailChimp, three weeks after their registration. If, at that point, they have activity in the previous two weeks, tag them as active (WasActiveInWeeksTwoOrThree), otherwise inactive (WasInactiveInWeeksTwoAndThree).

### Issues:
- fixes #6112

------
- [x] Ready for review

PR #6442, opened 2022-09-01
extensive error logging for mailchimp tagging

fm3:
it still did not tag any users :scream: 

follow up for #6214


leowe (approved):
Hopefully it works now :)

PR #6444, opened 2022-09-01
Skip fetching lastActivity for multiusers with zero active users

fm3:
Hopefully the last PR for this feature. #6442 showed that the problem was that fetching lastActivity for inactive users failed.

This was only a problem because the surrounding Fox.serialCombined stopped when the first user tagging couldn’t happen. Replacing it by `foreach` solved that (already in the previous pr). Now this PR also adds skipping this fetch entirely so that the if-else logic for isActivated can run normally. This will reduce the error logging.


leowe (approved):
👍

PR #6445, opened 2022-09-02
In e2e tests, disable datastore filesystem watcher

### LINKED ISSUE #6419: Snapshot tests occasionally fail: Dataset confocal-multi_knossos is not imported
CI occasionally fails at createExplorational on the dataset confocal-multi_knossos
example: https://app.circleci.com/pipelines/github/scalableminds/webknossos/7719/workflows/fcb48411-59fa-4be8-a5a6-e2dec132c6bf/jobs/19068
re-run fixed it: https://app.circleci.com/pipelines/github/scalableminds/webknossos/7719/workflows/93cf7c64-e249-4931-9deb-931fce9f3383/jobs/19069

One hypothesis is that in the CI no datasets exist on disk, so there may be a race condition of the disk-scanning disabling the dataset that was previously just set as existing by the set-up database (see datasets.csv)

fm3:
The snapshot tests use a custom mocked db content including datasets. Those were overwritten sometimes by the filesystem watching of the datastore. Additionally there was a race condition, as this watching happens in 1min intervals, asynchronous to the test’s reset-db task.

This PR passes an additional argument to the test runner, disabling this file system watcher entirely for the tests.

### Steps to test:
- I tested locally that even with `datastore.watchFileSystem.interval = 5 seconds` in the application.conf, the logging did not show file system scanning
- CI should pass

### Issues:
- fixes #6419

------
- [x] Ready for review

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -185,7 +185,8 @@ services:
         "testOnly e2e.* --
           -Dtracingstore.fossildb.address=fossildb
           -Dtracingstore.redis.address=redis
-          -Ddatastore.redis.address=redis"
+          -Ddatastore.redis.address=redis
+          -Ddatastore.watchFileSystem.enabled=false"
     volumes:
       - ./binaryData/Connectomics department:/home/${USER_NAME:-sbt-user}/webknossos/binaryData/Organization_X
 


philippotto (approved):
awesome!

PR #6451, opened 2022-09-08
Poll wk version to notify during upgrade

fm3:
- https://___.webknossos.xyz

### Steps to test:
- open help menu in navbar
- check in dev tools that every 2s a request is made to build info
- close the menu
- --> requests should stop
- update the server (or monkeypatch it somehow) --> a message should appear in the help menu (I tested this and it worked)

------
- [X] Ready for review


philippotto:
@fm3 I added some small changes and think the code is fine now :) do you want to test this, too?

fm3:
If it works for you, I think this is good to go. This can’t really break any existing functionality and I assume you tested it

philippotto (approved):

PR #6458, opened 2022-09-12
Fix isosurface loading for volume annotations with mappings

fm3:
- Set mapping parameters to None in volume tracing service before passing request to Isosurface Service. This is the bugfix
- Refactoring: use `None` instead of `null` for mappingService
- remove some unused imports 

### Steps to test:
- Create volume annotation based on segmentation layer, activate a mapping. Request isosurface (e.g. through proofreading tool before first split/merge action). Should work and look right

---

- [x] Ready for review

### DIFF
--- a/webknossos-tracingstore/app/com/scalableminds/webknossos/tracingstore/tracings/volume/VolumeTracingService.scala
+++ b/webknossos-tracingstore/app/com/scalableminds/webknossos/tracingstore/tracings/volume/VolumeTracingService.scala
@@ -74,7 +74,7 @@ class VolumeTracingService @Inject()(
 
   /* We want to reuse the bucket loading methods from binaryDataService for the volume tracings, however, it does not
      actually load anything from disk, unlike its “normal” instance in the datastore (only from the volume tracing store) */
-  val binaryDataService = new BinaryDataService(Paths.get(""), 100, null)
+  val binaryDataService = new BinaryDataService(Paths.get(""), 100, None)
 
   isosurfaceServiceHolder.tracingStoreIsosurfaceConfig = (binaryDataService, 30 seconds, 1)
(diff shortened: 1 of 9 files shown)


webknossos-tracingstore/app/com/scalableminds/webknossos/tracingstore/tracings/volume/VolumeTracingService.scala
```diff
@@ -391,8 +391,8 @@ class VolumeTracingService @Inject()(
         request.segmentId,
         request.subsamplingStrides,
         request.scale,
-        request.mapping,
-        request.mappingType
```
philippotto, line 395: It makes sense, that these are not passed because we are in the VolumeTracingService and not in the EditableMappingService, but the latter doesn't pass these mapping parameters, either. See https://github.com/scalableminds/webknossos/pull/6458/files?diff=split&w=1#diff-5a1cef6ee4682e274fee1f59abbba1bbe211a63e0142a7cea4f61b2a580767bdR596
Where are these parameters passed then? I can't really wrap my head around it.

fm3: Understandable :D Only in the datastore, where the backend does not know which mapping is active, there is a need to pass this parameter with the request. In the tracingstore, however, the backend knows the selected mapping, and the data is already mapped before it is passed to the isosurface service. This goes for both editable mapping annotations and normal volume annotations.

philippotto: ah, makes sense :)

philippotto (approved):
works for me and lgtm, too :)

PR #6460, opened 2022-09-12
Voxelytics frontend

### LINKED ISSUE #6468: Add Voxelytics Runs Overview
voxelytics runs reporting to webKnossos

normanrz:
### Steps to test:
- Get voxelytics and elasticsearch
- Set the `WORKFLOW_URL` env var to `http://localhost:9000`
- Set the `DEFAULT_WEBKNOSSOS_URL` env var to `http://localhost:9000`
- Set the `WORKFLOW_TOKEN` env var to a valid auth token in the above configured webKnossos instance
- Run `vx voxelytics/connect/test/configs/segem_test_2d.yaml --cpu`
- Open http://localhost:9000/workflows

fixes #6468 

------
- [ ] Merge together with #6416 
- [x] Ready for review


frontend/javascripts/libs/polling.ts
```diff
@@ -0,0 +1,37 @@
+import { useEffect, useRef } from "react";
+
+export default function usePolling(callback: () => Promise<void>, delay: number | null) {
```
philippotto, line 3: could be moved into the react_hooks module, I suppose.

frontend/javascripts/types/api_flow_types.ts
```diff
@@ -702,3 +703,166 @@ export type ZarrPrivateLink = {
   accessToken: string;
   expirationDateTime: number | null;
 };
+
+export enum VoxelyticsRunState {
```
philippotto, line 707: If you want to lazy-import a vx bundle, it might be useful to add these types to a new module. In general, we should reorganize some code a bit (e.g., this module currently imports the store which imports the sagas which imports a big part of the code base). However, this is rather out of scope for this PR.

frontend/javascripts/admin/voxelytics/task_list_view.tsx
```diff
@@ -0,0 +1,598 @@
...
+}
+
+function removeUrlParam(location: ReturnType<typeof useLocation>, key: string) {
+  const search = new URLSearchParams(location.search);
+  search.delete(key);
+  return `${location.pathname}?${search.toString()}`;
+}
+
+function TaskStateTag({ taskInfo }: { taskInfo: VoxelyticsTaskInfo }) {
+  useUpdateEvery(10000);
```
philippotto, line 89:
```suggestion
  // Re-render every 10s so that the durations that are relative to the current time
  // are updated regularly.
  useUpdateEvery(10000);
```

took me a moment to figure out why this is necessary.

philippotto (commented):
didn't test yet and only skimmed the new modules, but looks good overall :+1: will try to test now.

frontend/javascripts/types/api_flow_types.ts
normanrz: I think it makes more sense to refactor that in a coordinated effort. If moving these types in a separate file, we at least also need to move the vx-related functions from `admin_rest_api.ts` to a separate file.

normanrz:
> a comparison of the old and new bundle sizes would be interesting before we merge this.

6.6M vs 7.2M uncompressed (calculated with `yarn build && du -csh public/bundle/*.{js,css}`)

Bundle splitting turns out to be quite difficult :-/

philippotto:
> > a comparison of the old and new bundle sizes would be interesting before we merge this.
> 
> 6.6M vs 7.2M uncompressed (calculated with `yarn build && du -csh public/bundle/*.{js,css}`)
> Bundle splitting turns out to be quite difficult :-/

I had a look at the devtools on https://master.webknossos.xyz vs https://vxfrontend.webknossos.xyz/ and there the difference is 381 kB vs. 380 kB for `main.js` (so, basically nothing) and 1.1 MB vs. 966 kB for vendors~main.js. So, this means bundle splitting would probably only be beneficial to separate the new vendors code into a new module. Is this also difficult? Either way, it's probably not really necessary, right now. Still, it would be good to know whether further vendor splitting might be doable without too much effort.

philippotto (approved):
:lollipop:

normanrz:
> Still, it would be good to know whether further vendor splitting might be doable without too much effort.

It could be done, but we'd need to rework the code splitting config in webpack. We need to figure out how to assign stable names to the individual bundles (currently, hard-coded to `vendors~main`) or figure out dynamic imports in our system (currently, the vendors bundle is statically imported via HTML script tags).

PR #6464, opened 2022-09-13
Allow merge and min-cut when rightclicking on voxel (instead of node)

### LINKED ISSUE #6426: [Super Voxel Editing] Merge segments along shortest path
## Detailed Description
When editing the super voxel graph, it should be easy to combine two segments (=merge operation). Instead of having to manually create edge between the underlaying graphs, wK should automatically create the best/shortest/easiest? connection between the two graphs when executing this operation.

Ideally, I could imaging this workflow like this:
1. Select first segment / select a node of the first segment2. 
2. Click on the second segment/Right-click on the second segment and select merge from context menu.
3. Baaaam, they merge.

Ideally, this workflow does not require to load any of the underlaying super voxel graphs at all. It should be as few clicks as possible and comparable to the existing "Merger Mode".

## Context

philippotto:
- [https://___.webknossos.xyzhttps://proofreadingqolsplitandmerge.webknossos.xyz/](https://proofreadingqolsplitandmerge.webknossos.xyz/)

### Steps to test:
- create annotation for test-agglomerate-file DS
- select hdf5 mapping
- select proofreading tool
- click on segment to load agglomerate skeleton
- merging
  - select a node
  - rightclick on another segment (not a node) and select "merge with active segment"
    - --> segments should get merged (agglomerate skeleton will be reloaded)

~- splitting~
  ~- select a node~
    ~- rightclick on the same segment somewhere where it makes sense* and select "split from active segment"~
    ~- segments should get split~
- min-cutting
  - select a node
    - rightclick on the same segment somewhere where it makes sense** and select "min-cut from active segment"
    - segments should get split

Footnotes:
~\* for the user it's virtually impossible to know which voxel position would work for splitting, since one cannot see the oversegmentation (a toast will appear which tells you about this, though). if one clicks in the vicinity of another node of the same agglomerate skeleton, the chances are good that the split will succeed. however, it still might be the case that one clicks on the same over-segment. even if the split works, the user might get confused because nothing visibly happens. this is the case, when there are multiple edges which need to be split (however, the user action would only remove one edge).~
** for min-cutting there is the same risk of rightclicking the very same oversegment. however, if the target position refers to another over-segment, the split should succeed regardless of how many (indirect) edges exist, because they will be deleted by the min-cut.

the above problems might be only of theoretical nature. when working with actual segmentations and mergers/splits (and not dummy ones), the problem might be negligible. if these still appear, wk will recommend to use the agglomerate structure to perform the split or min-cut.

### Issues:
- fixes #6426

------
- [X] Ready for review

### DIFF
--- a/frontend/javascripts/oxalis/model/sagas/proofread_saga.ts
+++ b/frontend/javascripts/oxalis/model/sagas/proofread_saga.ts
@@ -461,8 +461,10 @@ function* splitOrMergeOrMinCutAgglomerate(
 
   yield* call([api.data, api.data.reloadBuckets], layerName);
 
-  const newSourceNodeAgglomerateId = yield* call(getDataValue, sourceNodePosition);
-  const newTargetNodeAgglomerateId = yield* call(getDataValue, targetNodePosition);
+  const [newSourceNodeAgglomerateId, newTargetNodeAgglomerateId] = yield* all([
+    call(getDataValue, sourceNodePosition),
+    call(getDataValue, targetNodePosition),
+  ]);
 
   /* Rename agglomerate skeleton(s) according to their new id and mapping name */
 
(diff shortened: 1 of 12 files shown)


frontend/javascripts/oxalis/model/sagas/proofread_saga.ts
```diff
@@ -399,31 +374,112 @@ function* splitOrMergeOrMinCutAgglomerate(
...
+  sourceNodePosition: Vector3,
+  targetNodePosition: Vector3,
+  agglomerateFileMag: Vector3,
+  editableMappingId: string,
+  volumeTracingId: string,
+  sourceTree: Tree | null,
+  items: UpdateAction[],
+): Saga<boolean> {
+  if (sourceNodeAgglomerateId !== targetNodeAgglomerateId) {
+    Toast.error("Segments need to be in the same agglomerate to perform a min-cut.");
```
hotzenklotz, line 461:
```suggestion
    Toast.error("Segments need to be in the same agglomerate to perform a min-cut splitting operation.");
```

hotzenklotz (approved):

PR #6472, opened 2022-09-15
Avoid int overflows for agglomerate tree id

### LINKED ISSUE #6467: Agglomerate Skeleton request returned negative tree ids
see https://scm.slack.com/archives/C5AKLAV0B/p1663147447844559?thread_ts=1663072356.819119&cid=C5AKLAV0B

fm3:
Protobuf skeletons only support int tree ids, not long. We math.abs the id now as a hash function so that a unique deterministic tree color can be selected but no overflows occur, as the frontend won’t handle negative tree ids.

### Steps to test:
- load agglomerate skeletons for agglomerates with very large ids
- they should show

### Issues:
- fixes #6467 

------
- [x] added changelog entry
- [x] Ready for review


philippotto (approved):
Great :+1:

PR #6477, opened 2022-09-16
Miscellaneous improvements to the proofreading tool

### LINKED ISSUE #6288: Followup for Editable Mappings aka Supervoxel Proofreading
#### Backend
 - [ ] duplicate route should work on this sort of volumetracing
 - [ ] revert to version update action for editable mapping
 - [ ] export
   - [ ] include it in downloaded annotation?
   - [ ] define exchange format
   - [ ] merge back into agglomerate file
   - [ ] merge back into dataset
   - [ ] apply in annotation zarr streaming
 - [ ] perf
    - [ ] Optimize Roundtrip?
    - [ ] Use low-level typed data processing (no UnsignedInteger)
    - [ ] Is the size limit for agglomerate skeletons too tight?

#### Frontend
  - [ ] Implement undo/redo support
  
Also see #6267 and #6265.

philippotto:
- fix race condition in conflict-checking code by re-reading current version from store
- use base mapping name of editable mapping in mapping UI if possible
- don't try to load agglomerate skeleton if user selects a node in proofreading tool (this could load another skeleton if a node was only visible due to a high clipping distance which was confusing)

- https://___.webknossos.xyz

### Steps to test:
- open DS with agglomerate mapping and select it
- select proofreading tool and use click on a segment
- the left sidebar should show a readable name for the mapping
- select a node in the tree --> nothing else should happen

### Issues:
- contributes to #6288

------
- [X] Ready for review


hotzenklotz (approved):
LGTM

PR #6479, opened 2022-09-20
Remove resolution parameter from requestRawCuboid

### LINKED ISSUE #6189: Remove zoomStep entirely, after client has been updated
After https://github.com/scalableminds/webknossos/pull/6159/ and https://github.com/scalableminds/webknossos-libs/issues/719 have been in production for some time (say, a month or two), we should remove the resolution/zoomStep paramter entirely from the requestRawCuboid route (datastore.routes; BinaryDataController), and make the new mag format non-optional. This should also make DataLayerLike.magFromExponent obsolete.

frcroth:
### Issues:
- fixes #6189 

Since this is an API change, I guess it needs some documentation?

------
- [x] Ready for review


fm3 (approved):
Looking good :) 
On API Documentation: the old, textual documentation at https://docs.webknossos.org/webknossos/rest_api.html includes only a subset of api endpoints that were at the time considered relevant to users.
Now we encourage using the python library, which uses @ApiParam annotations, which are composed into a swagger.json file as annotation.
So I’d say no further changes are required. However, I’d ask to add a changelog entry, which we can point towards if anything breaks.

fm3 (approved):
👍

PR #6481, opened 2022-09-20
Mag information is not lost on empty volumes

### LINKED ISSUE #6128: Make mag list detection for volume uploads more resilient
The mag list detection in the volume annotation upload cannot handle some edge cases:
 - buckets that are all-zero are skipped and if (in a mag) all buckets are zero, that mag is not added to the Mag list. Example zip: https://scm.slack.com/archives/C5AKLAV0B/p1648576115769449
 - volume annotations with no buckets at all (could still have header.wkw files per mag) also result in an empty Mag list.

https://github.com/scalableminds/webknossos/pull/6116 is a workaround that will transform this empty list to a complete list. But this should only happen if there is really no way to obtain the correct list. I’d say we should also consider all-zero buckets and header.wkw files when detecting the Mag set.

frcroth:
- https://emptyvolumesdownloadupload.webknossos.xyz

### Steps to test:
- Select a dataset with multiple mags
- Create a volume with fewer mags
- Download that volume
- Reload the dataset
- Upload the volume
- Ensure that only the selected mags are available in the volume

### Issues:
- fixes #6128

The original handling is still largely there to handle previously exported files.

------
- [x] Ready for review


CHANGELOG.unreleased.md
```diff
@@ -22,6 +22,7 @@ For upgrade instructions, please check the [migration guide](MIGRATIONS.released
 - Selecting a node with the proofreading tool won't have any side effects anymore. Previous versions could load additional agglomerate skeletons in certain scenarios which could be confusing. [#6477](https://github.com/scalableminds/webknossos/pull/6477)
 - Sharing links are shortened by default. Within the sharing modal, this shortening behavior can be disabled. [#6461](https://github.com/scalableminds/webknossos/pull/6461)
 - Removed optional "resolution" parameter from /datasets/:organizationName/:dataSetName/layers/:dataLayerName/data route. Use mag instead. [#6479](https://github.com/scalableminds/webknossos/pull/6479)
+- Changed how volumes containing no data are stored. Now the selection of magnitudes is correctly exported and imported. [#6481](https://github.com/scalableminds/webknossos/pull/6481)
```
fm3, line 25:
```suggestion
- Changed how volumes containing no data are stored. Now the selection of magnifications is correctly exported and imported. [#6481](https://github.com/scalableminds/webknossos/pull/6481)
```

webknossos-datastore/app/com/scalableminds/webknossos/datastore/dataformats/wkw/WKWBucketStreamSink.scala
```diff
@@ -1,32 +1,30 @@
...
+import com.scalableminds.webknossos.datastore.geometry.Vec3IntProto
 import com.scalableminds.webknossos.wrap.{BlockType, WKWFile, WKWHeader}
 
 import scala.collection.mutable
 import scala.concurrent.Future
 
 class WKWBucketStreamSink(val layer: DataLayer) extends WKWDataFormatHelper {
 
-  def apply(bucketStream: Iterator[(BucketPosition, Array[Byte])]): Iterator[NamedStream] = {
+  def apply(bucketStream: Iterator[(BucketPosition, Array[Byte])], mags: Seq[Vec3IntProto]): Iterator[NamedStream] = {
```
fm3, line 16: I’d suggest passing a Vec3Int in here, (not the proto version). The proto version does not have all the nice convenience methods we use all around the scala code. You can have the caller extend the ProtoGeometryImplicits trait, which should make `vec3IntFromProto`  available (and since it is implicit, you might not even have to call it if the function signatures make the types unambiguous)

fm3 (approved):
Looking good :) Please see my small comment on Vec3Int

fm3 (approved):

PR #6483, opened 2022-09-21
Make version history view paginated

### LINKED ISSUE #5421: Pagination for update action log
Some annotations have very large update action logs. We may want to not fetch the whole thing at once, but rather introduce some pagination.

It is used for the version restore view of annotations.

frcroth:
- [paginationforupdateactionlog.webknossos.xyz](https://padingationforupdateactionlog.webknossos.xyz/)

### Steps to test:
- Create some annotations
- Check the update history

### Issues:
- fixes #5421 

------
- [x] Ready for review


fm3:
@philippotto this issue became relevant again recently (a user reported data loss and we could not explore older annotation version as the update action log request did not complete in reasonable time). Do you think you have some capacity to adapt the front-end to use pagination? And if so, do you think the API with newestVersion and oldestVersion makes sense? Or would you prefer to send a limit and page? The front-end should already know the newest possible version of the tracing, right? and the oldest should always be zero. This is not urgent, but if you have some time, maybe you could have a look :)

philippotto:
> @philippotto this issue became relevant again recently (a user reported data loss and we could not explore older annotation version as the update action log request did not complete in reasonable time). Do you think you have some capacity to adapt the front-end to use pagination? And if so, do you think the API with newestVersion and oldestVersion makes sense? Or would you prefer to send a limit and page number? The front-end should already know the newest possible version of the tracing, right? and the oldest should always be zero. This is not urgent, but if you have some time, maybe you could have a look :)

I'm having a look :)

@frcroth could you do the same for the SkeletonTracingController? I think it makes sense that both version routes use the pagination mechanism.

philippotto:
The front-end is ready from my side. For easier testing I set the items per page parameter to `ENTRIES_PER_PAGE`. However, it could probably be something like 500? I also made the groups auto-expand for the testing round, but I'd revert that, too.

The only thing missing is that the back-end also accepts the new parameters for skeleton tracings.

fm3:
Yes 500 seems reasonable. We can still tweak it again later if we find that with real-world annotations this is still slow

frcroth:
@philippotto The route for skeleton tracing is implemented

philippotto:
> @philippotto The route for skeleton tracing is implemented

Thanks!

@fm3 Would you like to test this PR? Maybe you could also review the frontend? In that case, I think it would be easiest to have  a quick call where I guide you through it :)

frontend/javascripts/oxalis/view/version_list.tsx
```diff
@@ -147,73 +147,173 @@ const getGroupedAndChunkedVersions = _.memoize(
...
+    // function would not work correctly.
+    throw new Error("ENTRIES_PER_PAGE should be divisible by 10.");
+  }
+
+  // For example, the following parameters would be a valid variable set:
+  // baseVersion = 23
+  // relativePageNumber = 1
+  // absolutePageNumber = 10
+  // newestVersion = 22
+  // oldestVersion = 20
```
fm3, line 170:
```suggestion
  // oldestVersion = 21
```

frontend/javascripts/oxalis/view/version_list.tsx
```diff
@@ -147,73 +147,173 @@ const getGroupedAndChunkedVersions = _.memoize(
...
+    tracingId,
+    versionedObjectType,
+    oldestVersion,
+    newestVersion,
+  );
 
-  async function fetchData() {
+  // Insert version 0
+  if (oldestVersion === 1) {
+    updateActionLog.unshift({
```
fm3, line 190:
```suggestion
    updateActionLog.push({
```

frontend/javascripts/oxalis/view/version_entry_group.tsx
```diff
@@ -53,7 +53,7 @@ function GroupHeader({
 }
 export default class VersionEntryGroup extends React.Component<Props, State> {
   state: State = {
-    expanded: false,
+    expanded: false, // todo: revert after testing
```
fm3, line 56:
```suggestion
    expanded: false,
```

frontend/javascripts/oxalis/view/version_list.tsx
```diff
@@ -25,6 +25,9 @@ import Store from "oxalis/store";
 import VersionEntryGroup from "oxalis/view/version_entry_group";
 import api from "oxalis/api/internal_api";
 import Toast from "libs/toast";
+import { useInfiniteQuery, useQueryClient } from "@tanstack/react-query";
+
+const ENTRIES_PER_PAGE = 5;
```
fm3, line 30:
```suggestion
const ENTRIES_PER_PAGE = 500;
```

fm3 (approved):
Works for me :) Thanks for guiding me through the code in person. Maybe you can investigate why the first two pages are fetched twice each

philippotto:
> Maybe you can investigate why the first two pages are fetched twice each

Found the bug! I had to change a `||` to `&&`. Will merge now unless you veto :)

PR #6485, opened 2022-09-21
Improve Zarr Import UI (and general editing of datasets)

### LINKED ISSUE #6409: Improve datasource editing for remote zarr import
When adding a zarr dataset in the upload view we are displaying the datasource json as clean text.
Nobody should see the datasource json. We should show a slimmed-down nice representation of it:

- Dataset name (editable)
- Dataset scale (editable)
- List of Layers
  - Name (editable)
  - Category (editable)
  - Bounding box (read-only)
  - List of mags (read-only)
  - Optional largest segment id (editable)

philippotto:
- https://___.webknossos.xyz

### Steps to test:
- In general do some monkeytesting with the following:
  - use new "Add Zarr UI" (e.g., with layer url == `https://data-humerus.webknossos.org/data/zarr/scalable_minds/l4dense_motta_et_al_demo/segmentation`)
  - edit an existing dataset (zarr or wkw)

### Issues:
- fixes #6409

------
- [X] Ready for review


frontend/javascripts/dashboard/dataset/dataset_settings_data_tab.tsx
```diff
@@ -138,37 +167,70 @@ function SimpleDatasetForm({
...
+                name={["dataSource", "id", "name"]}
+                label="Name"
+                info="The name of the dataset"
+                rules={[
+                  {
+                    required: true,
+                    message: "Please provide a name for the dataset.",
+                  },
+                  {
+                    validator: syncValidator((value) => value.length > 3, "..."),
```
philippotto, line 187: @fm3 Do you know what the constraints are for the name of a dataset?

fm3: same as for dataset upload https://github.com/scalableminds/webknossos/blob/master/app/models/binary/DataSetService.scala#L53
```
      _ <- bool2Fox(name.matches("[A-Za-z0-9_\\-\\.]*")) ?~> "dataSet.name.invalid.characters"
      _ <- bool2Fox(!name.startsWith(".")) ?~> "dataSet.name.invalid.startsWithDot"
      _ <- bool2Fox(!name.endsWith(".")) ?~> "dataSet.name.invalid.endsWithDot"
      _ <- bool2Fox(name.length >= 3) ?~> "dataSet.name.invalid.lessThanThreeCharacters"
```

philippotto: Great, thank you. I found the frontend code which already validated this for the dataset name in the other import view. However, it didn't include the `.` rules. I assume that this is a bit newer and the frontend and backend diverged on this? Maybe you can add a comment to the back-end that this is also checked by front-end (it's a bit redundant, but I find it good for better readable messages).

fm3: Yes, the dot rules are newer. I’m not sure anymore if the endsWith dot should be removed again. What do you think? The startsWith dot is to avoid the dataset becoming a hidden directory

philippotto: > I’m not sure anymore if the endsWith dot should be removed again. What do you think?

I don't know whether there could be problems with zarr routes because the `.` could be a delimiter there? However, a dot in the middle of the name also seems to be supported at the moment, so a trailing `.` might also be ok. Other than the zarr topic, I don't see a reason to forbid trailing dots.

fm3: Then let’s just undo that (do you want to do this here and remove the line in the backend?)

philippotto: done in backend and frontend :)

hotzenklotz:
The form does not quite work for me, e.g.: When in "advanced" mode, I want to change the `numChannels` from `1` to `3`. As soon as I delete the value `1` the form crashes, because it has become invalid for a second... I guess the same applies to other properties as well.

hotzenklotz:
The default name for dataset is the same as the layer name of the first layer. I am not sure if this makes sense. In case of your test DS above, the dataset name should have been `l4_dense_motta`.  I am not sure if the name can also be derived from the Zarr file/spec or with should either leave the field empty or with a generic name.

image

hotzenklotz:
I almost overlooked the little "delete button". Perhaps, we could make it into very lightweight icon button (just a border, no background) to slightly emphasize it.

image
image

https://ant.design/components/button/

hotzenklotz:
Maybe not technically caused by this PR but wK offers a config suggestion, i.e. to remove all datalayer (`dataLayer = []`) when clearly the advanced view still has content/values for this property... 

image

hotzenklotz:
Nitpick: I have a dataset with exactly 1 layer. I can click the "delete" button and the UI will remove the layer while at the same complaining about an invalid config which it cannot save. The good thing first: I indeed cannot save and destroy my dataset. 
Ideas:
- Perhaps the `Save` button should be disabled when encountering an invalid config. 
- Maybe the `Remove Layer` button should either be disabled or not rendered at all for a dataset with just one layer.

hotzenklotz:
Another small nitpick: The error message contains a reference to the server time...? 😕 
image

frontend/javascripts/admin/dataset/dataset_add_view.tsx
```diff
@@ -116,7 +116,7 @@ function DatasetAddView({ history }: RouteComponentProps) {
     <React.Fragment>
       <Layout>
         <Content>
-          <Tabs defaultActiveKey="1" className="container">
+          <Tabs defaultActiveKey="2" className="container">
```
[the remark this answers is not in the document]
philippotto: it's only meant for dev.

- [ ] undo

philippotto:
> [ ] The default name for dataset is the same as the layer name of the first layer. I am not sure if this makes sense. In case of your test DS above, the dataset name should have been `l4_dense_motta`. I am not sure if the name can also be derived from the Zarr file/spec or with should either leave the field empty or with a generic name.

@fm3 I think the initial name is coming from the back-end, right? defaulting to an empty string sounds like a good solution for now (definitely, better than a poor guess, I'd say).

fm3:
> backend only infers what's on disk
Yes, makes sense to me :+1: 

> initial name is coming from the back-end, right?
correct

> defaulting to an empty string sounds like a good solution for now (definitely, better than a poor guess, I'd say).
ok I’ll include that change in #6520, should be merged soon

hotzenklotz:
I re-tested and LGTM. Ready to merge :-)

philippotto:
> I re-tested and LGTM. Ready to merge :-)

I need your official acceptance to merge this PR :)

fm3 (approved):

Issue #6465, opened 2022-09-13
remove default '' from id fields in sql schema


fm3:
This default is not a valid value, it should be removed.

Issue #5944, opened 2022-01-06
jobs._dataStore should be VARCHAR instead of CHAR


fm3:
In the postgres table, the reference to the datastore is currently stored as CHAR(256) in the webknossos.jobs table. This leads to padding with spaces, which needs to be trimmed when loading from the db. Should be VARCHAR.

PR #6486, opened 2022-09-22
Allow task type summary to identify task type when creating tasks in bulk

### LINKED ISSUE #4990: Allow selecting task type by name in bulk task creation
Looking up the task type ID is tedious, depending on your workflow.
Backwards-compatibility should be maintained, but if the entered value is not an ObjectId, we can add a lookup by name

frcroth:
- https://tasktypebysummary.webknossos.xyz

### Steps to test (end-to-end):
 - Create tasks in bulk, try by task type id and by summary. example:
```
ROI2017_wkw, sampleTaskType, sampleExp, 1, 512,512,512, 0, 0, 0, 1, 500, 500, 500, 100, 100, 100, sampleProject
ROI2017_wkw, sampleTaskType, sampleExp, 1, 512,512,512, 0, 0, 0, 1, 500, 500, 500, 100, 100, 100, sampleProject
ROI2017_wkw, sampleTaskType, sampleExp, 1, 512,512,512, 0, 0, 0, 1, 500, 500, 500, 100, 100, 100, sampleProject
```

Note that by summary does not work if the task type summary contains commas, since that breaks the frontend csv parsing (quote escaping is not implemented there).  However, we checked two large wk instances and nobody had task types with commas anyway

### Steps to test (backend only):
- Create a task via the POST /api/tasks route using a taskTypeSummary in the field taskTypeIdOrSummary rather than a taskTypeId
- Create a task via the POST /api/tasks route using a taskTypeId to see that it still works

### Issues:
- fixes #4990

------
- [x] Ready for review


fm3 (approved):
Backend LGTM. @philippotto could you have a quick look at the frontend changes?

app/models/task/TaskCreationParameters.scala
```diff
@@ -5,7 +5,7 @@ import models.user.Experience
 import play.api.libs.json.{Format, Json}
 
 case class TaskParameters(
-    taskTypeId: String,
+    taskTypeIdOrSummary: String,
```
fm3, line 8: could you add a comment here explaining that after initial processing this will only contain ids?

frontend/javascripts/admin/task/task_create_bulk_view.tsx
```diff
@@ -257,7 +257,7 @@ function TaskCreateBulkView() {
             Specify each new task on a separate line as comma seperated values (CSV) in the
             following format:
             <br />
-            <a href="/dashboard">dataSet</a>, <a href="/taskTypes">taskTypeId</a>, experienceDomain,
+            <a href="/dashboard">dataSet</a>, <a href="/taskTypes">taskType</a>, experienceDomain,
```
philippotto, line 260:
```suggestion
            <a href="/dashboard">dataSet</a>, <a href="/taskTypes">taskType (ID or summary)</a>, experienceDomain,
```

frontend/javascripts/admin/task/task_create_bulk_view.tsx
```diff
@@ -299,7 +299,7 @@ function TaskCreateBulkView() {
             >
               <TextArea
                 className="input-monospace"
-                placeholder="dataSet, taskTypeId, experienceDomain, minExperience, x, y, z, rotX, rotY, rotZ, instances, minX, minY, minZ, width, height, depth, project[, scriptId, baseAnnotationId]"
+                placeholder="dataset, taskType, experienceDomain, minExperience, x, y, z, rotX, rotY, rotZ, instances, minX, minY, minZ, width, height, depth, project[, scriptId, baseAnnotationId]"
```
philippotto, line 302:
```suggestion
                placeholder="dataset, taskType (ID or summary), experienceDomain, minExperience, x, y, z, rotX, rotY, rotZ, instances, minX, minY, minZ, width, height, depth, project[, scriptId, baseAnnotationId]"
```

philippotto (approved):
Overall, the changes lgtm :+1: I only added two suggestions for communicating to the user that id and summary are both valid.

However, while reading the code, I had another thought: Would it make sense to rename "Summary" to "Name"? A summary as an identifying feature seems a bit weird to me. A name would make more sense in my opinion. What do you think?

fm3:
> Would it make sense to rename "Summary" to "Name"?

Yes I think that would make sense. Not high priority, though

philippotto:
> > Would it make sense to rename "Summary" to "Name"?
> 
> Yes I think that would make sense. Not high priority, though

True, but it could fit well with this PR, I think :)

> Note that by summary does not work if the task type summary contains commas, since that breaks the frontend csv parsing (quote escaping is not implemented there).  However, we checked two large wk instances and nobody had task types with commas anyway

I added a form validation to the summary field so that commas may not be included for future creations/edits.

fm3:
I plan on doing a combined sql cleanup pr soon for #6465, #5944 and renaming privateLinks.accessToken to key. Maybe I can include this renaming in that evolution and propagate the new name throughout the code.

philippotto:
Ok, cool :+1:

PR #6491, opened 2022-09-22
New Mesh File Support

### LINKED ISSUE #5830: Support new Mesh File Format
Once the new mesh file format is finalized, reading support can be added to the wk backend.

leowe:
- https://newmeshfilesupport.webknossos.xyz

### Steps to test:
- Use a mesh file with the new format and use "Show precomputed mesh" for a segment in webknossos

### Issues:
- fixes #5830 

------
- [x] Remove Draco dummy route & functions

------
- [x] Ready for review


webknossos-datastore/app/com/scalableminds/webknossos/datastore/storage/Hdf5FileCache.scala
```diff
@@ -41,3 +48,26 @@ class Hdf5FileCache(val maxEntries: Int) extends LRUConcurrentCache[String, Cach
...
+        block
+      }.toFox
+    } yield result
+
+  def safeExecuteBox[T](filePath: Path, meshFileCache: Hdf5FileCache)(block: CachedHdf5File => T): Box[T] =
+    for {
+      _ <- if (filePath.toFile.exists()) {
+        Full(true)
+      } else {
+        Empty ~> "mesh.file.open.failed"
```
fm3, line 67:
```suggestion
        Failure("mesh.file.open.failed")
```
(I believe the ~> operator turns the Empty box into a Failure anyway)

webknossos-datastore/app/com/scalableminds/webknossos/datastore/storage/Hdf5FileCache.scala
```diff
@@ -41,3 +48,26 @@ class Hdf5FileCache(val maxEntries: Int) extends LRUConcurrentCache[String, Cach
     }
   }
 }
+
+object CachedHdf5Utils {
+  def safeExecute[T](filePath: Path, meshFileCache: Hdf5FileCache)(block: CachedHdf5File => T)(
```
fm3, line 53: Do you think the Box implementation could be used everywhere, and then be converted to Fox when needed? Or is there Fox-specific logic that is needed in some usages?

webknossos-datastore/app/com/scalableminds/webknossos/datastore/services/MeshFileService.scala
```diff
@@ -105,43 +170,199 @@ class MeshFileService @Inject()(config: DataStoreConfig)(implicit ec: ExecutionC
...
+      .map(
+        lod =>
+          MeshLodInfo(scale = segmentInfo.lodScales(lod).toInt,
+                      vertexOffset = segmentInfo.vertexOffsets(lod),
+                      chunkShape = segmentInfo.chunkShape,
+                      chunks = chunks(lod)))
+      .toList
+    MeshSegmentInfo(chunkShape = segmentInfo.chunkShape, gridOrigin = segmentInfo.gridOrigin, lods = meshfileLods)
+  }
+  private def parseNeuroglancerManifest(manifest: Array[Byte]): NeuroglancerSegmentInfo = {
```
fm3, line 248: maybe this could be moved to NeuroglancerSegmentInfo.fromBytes()?

webknossos-datastore/app/com/scalableminds/webknossos/datastore/services/MeshFileService.scala
```diff
@@ -105,43 +170,199 @@ class MeshFileService @Inject()(config: DataStoreConfig)(implicit ec: ExecutionC
...
+      val lodScaleMultiplier = cachedMeshFile.reader.float64().getAttr("/", "lod_scale_multiplier")
+      val transform = cachedMeshFile.reader.float64().getMatrixAttr("/", "transform")
+
+      val (neuroglancerStart, neuroglancerEnd) = getNeuroglancerOffsets(segmentId, cachedMeshFile)
+
+      val manifest = cachedMeshFile.reader
+        .uint8()
+        .readArrayBlockWithOffset("/neuroglancer", (neuroglancerEnd - neuroglancerStart).toInt, neuroglancerStart)
+      val segmentInfo = parseNeuroglancerManifest(manifest)
+      val meshfile = getChunksFromSegmentInfo(segmentInfo, lodScaleMultiplier, neuroglancerStart)
```
fm3, line 207: What exactly is returned here? I feel like meshfile isn’t the perfect name. The type seemst to be MeshSegmentInfo? Also we now have mixed usages of meshfile vs meshFile

fm3 (commented):
Cool stuff, the Backend already looks very good :) Maybe you can have a look at my comments today

util/src/main/scala/com/scalableminds/util/tools/Fox.scala
```diff
@@ -33,7 +33,7 @@ trait FoxImplicits {
 
   implicit def try2Fox[T](t: Try[T])(implicit ec: ExecutionContext): Fox[T] = t match {
     case Success(result)       => Fox.successful(result)
-    case scala.util.Failure(e) => Fox.failure(e.getMessage)
+    case scala.util.Failure(e) => Fox.failure(s"${e.toString}")
```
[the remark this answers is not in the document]
leowe: Yes this helped and prints the message and the exception class

frontend/javascripts/oxalis/model/sagas/isosurface_saga.ts
```diff
@@ -546,17 +545,42 @@ function* loadPrecomputedMeshForSegmentId(
...
   try {
-    availableChunks = yield* call(
-      getMeshfileChunksForSegment,
-      dataset.dataStore.url,
-      dataset,
-      getBaseSegmentationName(segmentationLayer),
-      meshFileName,
-      id,
-    );
+    // todo: should actually check against 3, but some new mesh files
```
philippotto, line 562: todo

frontend/javascripts/oxalis/controller/scene_controller.ts
```diff
@@ -302,12 +309,66 @@ class SceneController {
...
-    this.isosurfacesRootGroup.add(ambientLightForIsosurfaces);
-    const ambientLightForMeshes = new THREE.AmbientLight(0x404040, 15); // soft white light
+    // const ambientLightForIsosurfaces = new THREE.AmbientLight(0x404040, 15); // soft white light
+    // this.isosurfacesRootGroup.add(ambientLightForIsosurfaces);
+
+    // const ambientLightForMeshes = new THREE.AmbientLight(0x404040, 5); // soft white light
+    // this.meshesRootGroup.add(ambientLightForMeshes);
+
+    let unsubscribe = () => {};
+    const testLights = (a: number, b: number, c: number, d: number) => {
```
philippotto, line 319: todo: clean up when there's feedback on the lighting

frontend/javascripts/oxalis/view/components/setting_input_views.tsx
```diff
@@ -9,6 +9,8 @@ import messages from "messages";
 
 const ROW_GUTTER = 1;
 
+const ANTD_TOTAL_SPAN = 24;
```
philippotto, line 12: I added these changes because I added a setting for the "smooth mesh" behavior at first. However, I removed the setting again. The changes here are quite generic, though, and might be useful later which is why I left them here.

frontend/javascripts/libs/draco.ts
```diff
@@ -0,0 +1,35 @@
...
+
+export function getDracoLoader(): CustomDRACOLoader {
+  if (_dracoLoader) {
+    return _dracoLoader;
+  }
+  // @ts-ignore
+  _dracoLoader = new CustomDRACOLoader();
+
+  _dracoLoader.setDecoderPath(
+    "https://raw.githubusercontent.com/mrdoob/three.js/r145/examples/js/libs/draco/",
```
philippotto, line 26: This will load the WASM file from githubusercontent. It's probably better to copy the module to our public folder, right?

philippotto (commented):
@daniel-wer I think, I'll request your review later, but already a note for you:

- the code in token.ts and mesh_v0.ts was simply moved there
- the files BufferGeometryUtils.ts and DRACOLoader.ts were copied from the node package. these modules exist in two formats within the three package (commonjs and esm). the commonjs modules cannot be found by webpack (so, compilation fails) and the esm modules cannot be imported by babel (and also not by esbuild) (so, the tests failed). this is why I resorted to simply copying them for now.

fm3 (commented):
Backend LGTM :sunglasses:

philippotto:
@daniel-wer I think you can already review the PR. For the lighting part, I'm still waiting for other feedback, but that code isn't really critical anyway. For the version comparison with `2`, I'm a bit hesitant, since I know that there are already mesh files in circulation with v=2. So, maybe it's alright to leave the comparison at `>= 2`. What do you think?

frontend/javascripts/oxalis/model/sagas/isosurface_saga.ts
daniel-wer: > For the version comparison with 2, I'm a bit hesitant, since I know that there are already mesh files in circulation with v=2. So, maybe it's alright to leave the comparison at >= 2. What do you think?

Are we talking about many (or important) mesh files with version 2? Otherwise, I would lean towards bumping the comparison to 3 and assert that the mesh file version is neither 1 nor 2. This would make the code easier to understand and forfeit the need to explain that everything is named v3 but some mesh files have v2. Also we would need to make sure that there exists a test dataset that has a mesh file with the correct version 3.

frontend/javascripts/oxalis/model/sagas/isosurface_saga.ts
```diff
@@ -546,17 +615,48 @@ function* loadPrecomputedMeshForSegmentId(
...
+    if (version >= 2) {
+      const segmentInfo = yield* call(
+        meshV3.getMeshfileChunksForSegment,
+        dataset.dataStore.url,
+        dataset,
+        getBaseSegmentationName(segmentationLayer),
+        meshFileName,
+        id,
+      );
+      availableChunks = _.first(segmentInfo.chunks.lods)?.chunks || [];
```
daniel-wer, line 649: Is there a documentation of the new mesh file format somewhere? I would like to know how the `lods` work and whether they are sorted, for example :)

frontend/javascripts/types/api_flow_types.ts
```diff
@@ -693,6 +693,10 @@ export type ServerEditableMapping = {
 export type APIMeshFile = {
   meshFileName: string;
   mappingName?: string | null | undefined;
+  // 0   - is the first mesh file version
+  // 1-2 - skipped for consistency with VX artifact versioning
+  // 3   - (some artifacts might have used 2, too) is the newer version with draco encoding.
```
daniel-wer, line 698: See my other comment. If you agree, the part in the parentheses can be removed.

public/wasm/draco_wasm_wrapper.js
```diff
@@ -0,0 +1,104 @@
+var $jscomp=$jscomp||{};$jscomp.scope={};$jscomp.arrayIteratorImpl=function(f){var m=0;return function(){return m<f.length?{done:!1,value:f[m++]}:{done:!0}}};$jscomp.arrayIterator=function(f){return{next:$jscomp.arrayIteratorImpl(f)}};$jscomp.makeIterator=function(f){var m="undefined"!=typeof Symbol&&Symbol.iterator&&f[Symbol.iterator];return m?m.call(f):$jscomp.arrayIterator(f)};
```
daniel-wer, line 1: Would it make sense to include a comment where this was copied from?

daniel-wer (commented):
Nice work, I will report back after testing :+1:

frontend/javascripts/oxalis/model/sagas/isosurface_saga.ts
```diff
@@ -506,6 +515,66 @@ function* _refreshIsosurfaceWithMap(
...
+    }
+  }
+
+  // If a deferred already exists, the one can be awaited (regardless of
+  // whether it's finished or not) and its content used to call the callback.
+  // If mustRequest was set to true, a new deferred will be created which
+  // replaces the old one (old references to the first Deferred will still
+  // work and will be resolved by the corresponding saga execution).
+  if (fetchDeferredsPerLayer[layerName] && !mustRequest) {
+    const meshes = yield* call(() => fetchDeferredsPerLayer[layerName].promise());
```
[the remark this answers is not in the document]
philippotto: TS doesn't complain, but the `this` context is lost in that case. I could change the methods of the Deferred class to use something like `resolve = () => {}` instead of `resolve() {}`, but if we go down this route, we would have to do this everywhere in the long run, which feels weird :thinking: Not too sure about this. Feel free to disagree :)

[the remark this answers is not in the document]
daniel-wer: Ah, makes sense, let's leave it as-is then :+1:

frontend/javascripts/oxalis/model/sagas/isosurface_saga.ts
```diff
@@ -506,6 +515,66 @@ function* _refreshIsosurfaceWithMap(
...
+  // else return the stored files (might be empty array). Or if we force a reload.
+  if (!files || mustRequest) {
+    const availableMeshFiles = yield* call(
+      getMeshfilesForDatasetLayer,
+      dataset.dataStore.url,
+      dataset,
+      getBaseSegmentationName(segmentationLayer),
+    );
+    yield* put(updateMeshFileListAction(layerName, availableMeshFiles));
+    deferred.resolve(availableMeshFiles);
```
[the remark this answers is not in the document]
philippotto: I think, the scenario you described cannot occur, since there will always be a deferred if the mesh files were already fetched. However, the code didn't really make use of this fact (and thus was hard to read). I simplified the code now accordingly :)

frontend/javascripts/oxalis/model/sagas/isosurface_saga.ts
philippotto: See https://www.notion.so/scalableminds/Design-Doc-Mesh-file-069d8fe05fff4017bc88584c1f6f92c6. I think, the lods are not sorted, but since we haven't really done anything multi-lod, yet, I'd defer this until we can properly test it with a multi-lod-file.

daniel-wer (commented):
The mesh loading with new and old mesh file formats worked very well :+1: 

I only noticed one issue during testing. When opening a sharing link that included an active mesh file, for example, this one: https://newmeshfilesupport.webknossos.xyz/annotations/633d73740100009b005f634a#%7B%22position%22:%5B3517,3800,1025%5D,%22mode%22:%22orthogonal%22,%22zoomStep%22:1,%22stateByLayer%22:%7B%22071103ab-b959-4d83-9538-d7512ad225c9%22:%7B%22meshInfo%22:%7B%22meshFileName%22:%22meshfile_v3%22,%22meshes%22:%5B%7B%22segmentId%22:58624,%22seedPosition%22:%5B3436,3687,1025%5D,%22isPrecomputed%22:true,%22meshFileName%22:%22meshfile_v3%22%7D%5D%7D%7D%7D%7D

the other mesh file is activated, not the "meshfile_v3". (The mesh that is encoded in the URL as well is loaded using the correct mesh file!)

philippotto:
> the other mesh file is activated, not the "meshfile_v3". (The mesh that is encoded in the URL as well is loaded using the correct mesh file!)

good catch! I fixed it now :) if you give your final "go", I'd merge this :ship:

frontend/javascripts/oxalis/controller/scene_controller.ts
```diff
@@ -306,69 +306,36 @@ class SceneController {
...
-
-    // const light1 = new THREE.DirectionalLight(0xefefff, 12);
-    // light1.position.set(1, 1, 1).normalize();
-    // this.isosurfacesRootGroup.add(light1);
-
-    // const light2 = new THREE.DirectionalLight(0xffefef, 8);
-    // light2.position.set(-1, -1, -1).normalize();
-    // this.isosurfacesRootGroup.add(light2);
+    const AMBIENT_INTENSITY = 30;
+    const DIRECTIONAL_INTENSITRY = 5;
```
daniel-wer, line 313:
```suggestion
    const DIRECTIONAL_INTENSITY = 5;
```

frontend/javascripts/oxalis/controller/scene_controller.ts
```diff
@@ -306,69 +306,36 @@ class SceneController {
...
+    const POINT_INTENSITY = 5;
+
+    const ambientLight = new THREE.AmbientLight(2105376, AMBIENT_INTENSITY);
+
+    const directionalLight = new THREE.DirectionalLight(16777215, DIRECTIONAL_INTENSITRY);
```
daniel-wer, line 318:
```suggestion
    const directionalLight = new THREE.DirectionalLight(16777215, DIRECTIONAL_INTENSITY);
```

frontend/javascripts/oxalis/controller/scene_controller.ts
```diff
@@ -306,69 +306,36 @@ class SceneController {
...
+    directionalLight.position.x = 1;
+    directionalLight.position.y = 1;
+    directionalLight.position.z = 1;
+    directionalLight.position.normalize();
+
+    const directionalLight2 = new THREE.DirectionalLight(16777215, DIRECTIONAL_INTENSITRY);
```
daniel-wer, line 324:
```suggestion
    const directionalLight2 = new THREE.DirectionalLight(16777215, DIRECTIONAL_INTENSITY);
```

frontend/javascripts/oxalis/model/sagas/isosurface_saga.ts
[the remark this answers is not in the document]
daniel-wer: Thank you, much more clear now :+1:

daniel-wer (approved):
LGTM, please see the code suggestions to fix a remaining typo. :+1:

PR #6492, opened 2022-09-26
SQL schema cleanup

### LINKED ISSUE #5944: jobs._dataStore should be VARCHAR instead of CHAR

### LINKED ISSUE #6465: remove default '' from id fields in sql schema

fm3:
Normalizes the sql schema a bit, adds missing foreign key constraint.

- https://sqlcleanup.webknossos.xyz

### Steps to test:
- should still be able to manage/view datasets

### Issues:
- fixes #6465 
- fixes #5944 

------
- [x] Ready for review


leowe (approved):
LGTM 👍

PR #6493, opened 2022-09-26
Add "shift + w" shortcut to cycle backwards through tools

### LINKED ISSUE #6299: Shift+W could cycle current tool backwards
Shift+W could cycle current tool backwards.

Dagobert42:
- https://___.webknossos.xyz

### Steps to test:
- open an annotation and use shift + w to cycle through tools backwards

### Issues:
- fixes #6299 

------
- [x] Ready for review


philippotto (approved):
:+1:

Since "Shift" is also a modifier to temporarily switch from one tool to another, the preview is  a bit weird (keeping shift pressed down will often show the color picker icon). However, I don't see a good way to circumvent this and also the switching itself works properly. So, let's go with it for now :)

philippotto:
Please also update the documentation here: https://github.com/scalableminds/webknossos/blob/master/docs/keyboard_shortcuts.md?plain=1#L103

Would be cool if you could also rename "toggle mode" to "cycle through tools" :)

PR #6494, opened 2022-09-26
Load 5000 versions instead of 500 for the paginated version restore list

philippotto:
Loading 500 versions only takes ~100ms and there are tracings with more than 500k versions. So, the page size should be higher.

- https://___.webknossos.xyz

### Steps to test:
- shouldn't be necessary

### Issues:
- follow up for #5421


fm3 (approved):
50💯

PR #6497, opened 2022-09-27
Fix validation of layer selection when trying to start globalization of floodfills

philippotto:
Additionally, I also changed the globalize button to be always visible, but explain when it's disabled via a tooltip.

- https://___.webknossos.xyz

### Steps to test:
- setup worker
- use fill tool for a large cell so that it's only partially computed
- click globalize floodfill button
- select layer
- confirm modal -> shouldn't an invalid form error

------
- [X] Ready for review


fm3 (approved):
LGTM

PR #6504, opened 2022-09-28
Script for creating GH release

philippotto:
The script can be called with `python3 tools/create-gh-release.py` which will then:
- extract the newest release notes from `CHANGELOG.released.md`
- format the notes with the header like we did it before manually for GH releases
- create a GH release with the CLI tool `gh`

We could also plug this into a GH action, but since there are manual steps involved in the releases anyway (e.g., curating the "Highlights"), I don't have any problems with manually starting this script. It's still a big improvement to manually fiddling the release notes together (especially, the anchor links were error prone in the past which motivated me to automate this).

-------
- [X] Ready for review


philippotto:
by the way, I tested this with a private repo. it looks like this:

image

normanrz (approved):
nice stuff!

PR #6517, opened 2022-09-29
Proofreading - Using Precomputed mesh for Oversegmentation

### LINKED ISSUE #6506: [Proofreading] Make use of precomputed mesh of oversegmentation
If a precomputed mesh of the oversegmentation exists, this should be used when meshing a segments. One approach could be that the backend applies the agglomerate mapping on the mesh data and serves that to the frontend. The chunk borders might need special handling.

Algorithmically, I imagine something like this roughly:
- frontend sends segment ID (mapped)
- backend looks up which (unmapped) ids in the oversegmentation map to the given ID using the agglomerate file (is this operation possible?)
- backend looks up the meshes of each id and generates a list of mesh chunks
- the frontend can load the mesh chunks

leowe:
- https://___.webknossos.xyz

### Steps to test:
- [agglomerate-test-file](https://www.notion.so/scalableminds/Test-Datasets-c0563be9c4a4499dae4e16d9b2497cfb#4f27c86bc8fc485e9d04b0dc1b1488c9) and
[meshfile_new_format.hdf5.zip](https://github.com/scalableminds/webknossos/files/9684395/meshfile_new_format.hdf5.zip)

### Issues:
- fixes #6506 
- depends on #6491 

------
- [ ] Switch defaults in route

------
- [ ] Ready for review


leowe:
I am not sure why, but some of the unmapped segment ids don't occur in the mesh file. I therefore allowed it that some of them fail.

fm3:
> I am not sure why, but some of the unmapped segment ids don't occur in the mesh file. I therefore allowed it that some of them fail.

Possibly because the mesh file was not computed on mag1 and some small segments get lost in downsampling?

fm3:
@philippotto looks like this branch diverged from #6491 before that one was merged and squashed. Since I don’t really know what this PR is for I can’t piece together which commits are relevant here. Maybe we could have a call about this PR and how to move forward here

fm3:
superseded by #6569

PR #6518, opened 2022-09-29
makes workflow_yamlContent optional

normanrz:
Actually, the field `workflow_yamlContent` is optional. For example, when constructing the workflow in code (like in the tests).

------
- [x] Ready for review


fm3 (approved):
Looks like it was already optional in the schema – how did this work before, was there a default value when parsing from sql?

normanrz:
> how did this work before

It didn't. I didn't test it with the vx tests yet.

PR #6520, opened 2022-09-30
[WIP] Explore remote N5 datasets

fm3:
- https://___.webknossos.xyz

### Steps to test:
- abc

### Issues:
- fixes #

------
- [ ] Ready for review


fm3:
@philippotto if you have time, maybe you could adapt the front-end to reflect the change that remote datasets can now also be n5 instead of zarr (not sure how prominently we want to promote this feature). 
I’d say that the backend part could also be merged as is, so the frontend changes do not necessarily block this PR

app/models/binary/explore/N5MultiscalesExplorer.scala
```diff
@@ -0,0 +1,129 @@
...
+      case "zm"        => Fox.successful(1e-12)
+      case "am"        => Fox.successful(1e-9)
+      case "fm"        => Fox.successful(1e-6)
+      case "pm"        => Fox.successful(1e-3)
+      case "nm"        => Fox.successful(1.0)
+      case "µm"        => Fox.successful(1e3)
+      case "mm"        => Fox.successful(1e6)
+      case "cm"        => Fox.successful(1e7)
+      case "dm"        => Fox.successful(1e8)
+      case "m"         => Fox.successful(1e9)
```
jstriebel, line 81: Maybe this can be shared with the other unit definitions? Or are those too different?

jstriebel (approved):
LGTM, love the new `Explorer` hierarchy :heart:

app/models/binary/explore/N5MultiscalesExplorer.scala
fm3: From what I’ve seen, ngff uses `nanometer` while n5 uses `nm` so they are pretty distinct. Of course, we could still merge the matching into one big function, what would you prefer?

jstriebel: Let's keep it separate as-is for now :+1:
The only reference I found is https://github.com/saalfeldlab/n5-viewer/blob/23f86b32e77ff1856893fa955523fb016cef3d82/README.md#container-structure,
where it says that not setting anything assumes `µm`, I think that is currently not the case. Also, it specifically shows the string `um`, which is missing in this list.

PR #6527, opened 2022-10-04
Replace babel with esbuild for tests and for webpack

philippotto:
- https://esbuildtests.webknossos.xyz

### Steps to test:
- CI
- do some monkey testing on the deployed branch

### Issues:
- improves build time and especially test compilation time (30s to 1s)
- build time should go down further if we remove webpack (to which the current PR can be seen as a precursor)


webpack.config.js
```diff
@@ -77,7 +78,17 @@ module.exports = function (env = {}) {
         {
           test: /\.tsx?$/,
           exclude: /(node_modules|bower_components)/,
-          use: "babel-loader",
+          loader: "esbuild-loader",
+          options: {
+            loader: "tsx", // Or 'ts' if you don't need tsx
```
philippotto, line 83:
```suggestion
            loader: "tsx", // also supports 'ts'
```

tools/test.sh
```diff
@@ -19,9 +19,28 @@ function prepare {
...
   pbjs -t json "webknossos-datastore/proto/VolumeTracing.proto" > "$testBundlePath/VolumeTracing.proto.json"
-  # --copy-files will copy files that are present in the source dir but are not transpiled (e.g.: json files)
-  # "$@" inside this function refers to all arguments of the prepare function, not all arguments of this bash script
-  BABEL_ENV=test babel --extensions .ts,.tsx "$jsPath" --out-dir "$testBundlePath" --copy-files "$@"
+
+  node_modules/.bin/esbuild \
+    --platform=node \
+    --format=cjs \
+    --define:process.env.BABEL_ENV=\"test\" \
+    --target=node10.4 \
```
daniel-wer, line 27: In https://github.com/scalableminds/webknossos/blob/master/DEV_INSTALL.md we recommend node 16 or 18. Would it make sense to use a similar target, here? Maybe this will speed things up, because newer features can be used.

tools/test.sh
```diff
@@ -19,9 +19,28 @@ function prepare {
...
   pbjs -t json "webknossos-datastore/proto/SkeletonTracing.proto" > "$testBundlePath/SkeletonTracing.proto.json"
   pbjs -t json "webknossos-datastore/proto/VolumeTracing.proto" > "$testBundlePath/VolumeTracing.proto.json"
-  # --copy-files will copy files that are present in the source dir but are not transpiled (e.g.: json files)
-  # "$@" inside this function refers to all arguments of the prepare function, not all arguments of this bash script
-  BABEL_ENV=test babel --extensions .ts,.tsx "$jsPath" --out-dir "$testBundlePath" --copy-files "$@"
+
+  node_modules/.bin/esbuild \
+    --platform=node \
+    --format=cjs \
+    --define:process.env.BABEL_ENV=\"test\" \
```
daniel-wer, line 26: Not that important, but since we don't use BABEL any longer, this might be confusing. We could rename it everywhere to something more fitting. From what I can see, it is basically used as a boolean indicating whether the tests run or not, so maybe `IS_TESTING`.

tools/test.sh
```diff
@@ -19,9 +19,28 @@ function prepare {
...
+    --target=node10.4 \
+    --outdir="$testBundlePath" $($FIND frontend/javascripts \( -name "*.ts" -o -name "*.tsx" \))
+
+  # Copy files which were not compiled by esbuild (e.g., snapshots).
```
daniel-wer, line 30: This is an impressive command, but also slightly intimidating. Could you add an example file path before and after the transformation. This should make it easier to adapt the command in the future, in case the folder structure changes.

daniel-wer (approved):
Awesome! The test compilation speedup is quite unreal, also 1000 lines less in yarn.lock! :exploding_head: 

I didn't encounter any issues during testing.

tools/test.sh
philippotto: Unfortunately, node >= 13 fails to execute some tests because of a dynamic import (which gets compiled to a `require` with lower node targets). I couldn't really find out why the import fails, but I added a comment which explains the target config.

tools/test.sh
philippotto: Good point! I just extended the comment :)

daniel-wer (approved):
Thank you, LGTM :+1:

PR #6532, opened 2022-10-05
Avoid allocating spire uint objects during apply agglomerate

fm3:
When applying an agglomerate mapping, each value of the segmentation data needs to be converted to Long to then look up the agglomerate id in the hdf5 array. Since the jvm does not support unsigned types, this was previously done by wrapping each signed value in a spire unsigned integer object, and then calling toLong. 

This PR changes the code to skip using the spire classes and directly calling the to-long method (which performs bitwise and to do the unsigned logic) on each value.

It also moves the filter-zero logic out of the data conversion since this had a performance cost even if the filterZero boolean was false.

It also adds a pre-allocated LongBuffer instead of using `array.map(_.toLong)`, which is slightly faster, I guess the old version did not know the final length of the output array before.

My measurements show that the conversion to Long is about six times faster with this PR, almost halving the time to apply an agglomerate mapping on my test data.

cc @youri-k 

- https://applymappingfunctionaltyping.webknossos.xyz

### Steps to test:
- View dataset with agglomerate mapping; activate it, should show mapped data
- Compute ad-hoc mesh, should show meshes for mapped data
- Histogram and find data should still work for different dtypes (e.g. uint8, uint16, float, uint24-rgb)

------
- [x] Ready for review


webknossos-datastore/app/com/scalableminds/webknossos/datastore/services/AgglomerateService.scala
```diff
@@ -64,16 +60,26 @@ class AgglomerateService @Inject()(config: DataStoreConfig) extends DataConverte
...
+        convertToAgglomerate(longBuffer.array, bytesPerElement, putByte)
+      case data: Array[Short] =>
+        val longBuffer = LongBuffer.allocate(data.length)
+        data.foreach(e => longBuffer.put(uShortToLong(e)))
+        convertToAgglomerate(longBuffer.array, bytesPerElement, putShort)
+      case data: Array[Int] =>
+        val longBuffer = LongBuffer.allocate(data.length)
+        data.foreach(e => longBuffer.put(uIntToLong(e)))
+        convertToAgglomerate(longBuffer.array, bytesPerElement, putInt)
+      case data: Array[Long] => convertToAgglomerate(data, bytesPerElement, putLong)
```
jstriebel, line 81: Please add a comment with a short explanation about the reasoning here (probably similar to PR description)

jstriebel (approved):
Nice, mostly LGTM, just a comment about the optimizations would be nice. I tested the behavior on the deployed dev-instance, and skimmed through the code, not checking everything in detail. Do you think that's fine?

PR #6534, opened 2022-10-05
Add OIDC authentication

### LINKED ISSUE #6437: Single Sign On (SSO) Authentication
OIDC and SAML authentication. Maybe upgrade silhouette?

frcroth:
- https://___.webknossos.xyz

### Steps to test:
- Create a local OIDC provider. See the comment below. (note: if keycloak is started via docker (not compose), change application.conf to `providerURL = "http://localhost:8080/realms/master/"`)
- also set `oidcEnabled` to true in the application.conf `features` block
- open http://localhost:9000/auth/login and use the "login with sso" button
- you should be redirected to the keycloak UI
- log in with admin:admin
- you should be redirected to webknossos (a new account was created and you are logged in)
- log out and log back in (e.g. with incognito tab), the same account should be logged in
- disable oidcEnabled again and verify that the log in/out mechanism works as usual

### Open questions:
- Is client authentication required (currently not implemented)
- [X] Should there be an option to disable OIDC?

### Issues:
- fixes #6437 

------
- [x] Ready for review


frcroth:
You can use keycloak, just start this with `docker compose up`.
```
version: '3'

volumes:
  mysql_data:
      driver: local

services:
  mysql:
      image: mysql:5.7
      volumes:
        - mysql_data:/var/lib/mysql
      environment:
        MYSQL_ROOT_PASSWORD: root
        MYSQL_DATABASE: keycloak
        MYSQL_USER: keycloak
        MYSQL_PASSWORD: password
  keycloak:
      image: quay.io/keycloak/keycloak:legacy
      environment:
        DB_VENDOR: MYSQL
        DB_ADDR: mysql
        DB_DATABASE: keycloak
        DB_USER: keycloak
        DB_PASSWORD: password
        KEYCLOAK_USER: admin
        KEYCLOAK_PASSWORD: admin
        # Uncomment the line below if you want to specify JDBC parameters. The parameter below is just an example, and it shouldn't be used in production without knowledge. It is highly recommended that you read the MySQL JDBC driver documentation in order to use it.
        #JDBC_PARAMS: "connectTimeout=30000"
      ports:
        - 8080:8080
      depends_on:
        - mysql

```
You can import a client with this JSON:
```
{
  "clientId": "myclient",
  "name": "My client",
  "description": "",
  "rootUrl": "",
  "adminUrl": "",
  "baseUrl": "http://localhost:8080/realms/master/myclient/",
  "surrogateAuthRequired": false,
  "enabled": true,
  "alwaysDisplayInConsole": false,
  "clientAuthenticatorType": "client-secret",
  "redirectUris": [
    "",
    "*",
    "http://localhost:9000/api/auth/oidc/callback",
    "https://www.keycloak.org/app/*"
  ],
  "webOrigins": [],
  "notBefore": 0,
  "bearerOnly": false,
  "consentRequired": false,
  "standardFlowEnabled": true,
  "implicitFlowEnabled": false,
  "directAccessGrantsEnabled": true,
  "serviceAccountsEnabled": false,
  "publicClient": true,
  "frontchannelLogout": true,
  "protocol": "openid-connect",
  "attributes": {
    "oidc.ciba.grant.enabled": "false",
    "display.on.consent.screen": "false",
    "oauth2.device.authorization.grant.enabled": "false",
    "backchannel.logout.session.required": "true",
    "backchannel.logout.revoke.offline.tokens": "false"
  },
  "authenticationFlowBindingOverrides": {},
  "fullScopeAllowed": true,
  "nodeReRegistrationTimeout": -1,
  "defaultClientScopes": [
    "web-origins",
    "acr",
    "roles",
    "profile",
    "email"
  ],
  "optionalClientScopes": [
    "address",
    "phone",
    "offline_access",
    "microprofile-jwt"
  ],
  "access": {
    "view": true,
    "configure": true,
    "manage": true
  }
}
```

app/models/user/UserService.scala
```diff
@@ -200,6 +200,9 @@ class UserService @Inject()(conf: WkConf,
       _ <- multiUserDAO.updatePasswordInfo(user._multiUser, passwordInfo)(GlobalAccessContext)
     } yield passwordInfo
 
+  def getOIDCPasswordInfo: PasswordInfo =
```
fm3, line 203:
```suggestion
  def getOpenIdConnectPasswordInfo: PasswordInfo =
```

app/oxalis/security/OpenIdConnectClient.scala
```diff
@@ -0,0 +1,93 @@
...
+import pdi.jwt.{JwtJson, JwtOptions}
+import play.api.libs.ws._
+import utils.WkConf
+
+import java.net.URLEncoder
+import java.nio.charset.StandardCharsets
+import javax.inject.Inject
+import scala.concurrent.ExecutionContext
+
+class OpenIDConnectClient @Inject()(rpc: RPC, conf: WkConf)(implicit executionContext: ExecutionContext) {
```
fm3, line 16:
```suggestion
class OpenIdConnectClient @Inject()(rpc: RPC, conf: WkConf)(implicit executionContext: ExecutionContext) {
```

app/oxalis/security/OpenIdConnectClient.scala
```diff
@@ -0,0 +1,93 @@
...
+        "redirect_uri" -> redirectUrl,
+        "scope" -> openIdClient.scope,
+        "response_type" -> "code",
+      )
+      serverInfos.authorization_endpoint + "?" +
+        queryParams.map(v => v._1 + "=" + URLEncoder.encode(v._2, StandardCharsets.UTF_8.toString)).mkString("&")
+    }
+
+  /*
+  Create token request (https://openid.net/specs/openid-connect-core-1_0.html#TokenRequest), fields described by
```
fm3, line 34:
```suggestion
  Fetches token from the oidc provider (https://openid.net/specs/openid-connect-core-1_0.html#TokenRequest), fields described by
```

app/oxalis/security/OpenIdConnectClient.scala
```diff
@@ -0,0 +1,93 @@
...
+    for {
+      serverInfos <- discover(openIdClient)
+      tokenResponse <- rpc(serverInfos.token_endpoint).postFormParseJson[OpenIdConnectTokenResponse](
+        Map(
+          "grant_type" -> "authorization_code",
+          "client_id" -> openIdClient.clientId,
+          "redirect_uri" -> redirectUrl,
+          "code" -> code
+        ))
+      new_code <- JwtJson
```
fm3, line 47:
```suggestion
      newToken <- JwtJson
```

app/utils/WkConf.scala
```diff
@@ -60,6 +60,13 @@ class WkConf @Inject()(configuration: Configuration) extends ConfigReader with L
       val children = List(User)
     }
 
+    object OIDC {
```
fm3, line 63:
```suggestion
    object SingleSignOn {
      object OpenIdConnect {
```

It’s a bit arbitrary but I’d move this out of the webKnossos config block and into the top level, since it is not a feature that only makes sense in the webKnossos context. Instead, I’d wrap it in SingleSignOn, since we may add more providers in the future.

app/utils/WkConf.scala
```diff
@@ -60,6 +60,13 @@ class WkConf @Inject()(configuration: Configuration) extends ConfigReader with L
...
+      val providerURL: String = get[String]("webKnossos.OIDC.providerURL")
```
fm3, line 64:
```suggestion
      val providerUrl: String = get[String]("singleSignOn.openIdConnect.providerUrl")
```

conf/application.conf
```diff
@@ -70,6 +70,13 @@ webKnossos {
 
     Please add the information of the operator to comply with GDPR.
   """
+
+  OIDC {
+    providerURL = "http://localhost:8080/auth/realms/master/"
```
fm3, line 75: I’d say this should all be empty strings in the repository, as this feature should be disabled by default. I’d say sample values for testing can be found in the PR description/discussion and this is enough, we should be able to find the PR again later.

conf/evolutions/091-oidc.sql
```diff
@@ -0,0 +1,35 @@
+BEGIN transaction;
+
+DROP VIEW webknossos.userInfos;
+DROP VIEW webknossos.multiUsers_;
+
+-- Cannot alter enum in transaction block, workaround required
```
fm3, line 6: Cool that this works :) When this happened previously I added the enum type outside of the transaction, which I think is also fine (since even if the rest of the transaction fails, this should not cause actual problems). But it is good to have this version, also as reference for the future.

util/src/main/scala/com/scalableminds/util/mvc/ExtendedController.scala
```diff
@@ -177,6 +177,7 @@ trait MimeTypes {
   val xmlMimeType: String = "application/xml"
   val zipMimeType: String = "application/zip"
   val jsonMimeType: String = "application/json"
+  val formUrlEncoded: String = "application/x-www-form-urlencoded"
```
fm3, line 180:
```suggestion
  val formUrlEncodedMimeType: String = "application/x-www-form-urlencoded"
```
(for consistency with the others)

app/controllers/AuthenticationController.scala
```diff
@@ -429,6 +431,91 @@ class AuthenticationController @Inject()(
     }
   }
 
+  private lazy val oidcClientId = conf.WebKnossos.OIDC.clientId
+  private lazy val oidcProviderUrl = conf.WebKnossos.OIDC.providerURL
```
fm3, line 435: since these have only one usage and the readability should still be ok, I’d inline them below

app/controllers/AuthenticationController.scala
```diff
@@ -429,6 +431,91 @@ class AuthenticationController @Inject()(
...
+
+  lazy val oidcConfig: OpenIdConnectConfig = OpenIdConnectConfig(oidcProviderUrl, oidcClientId)
+  lazy val absoluteOidcCallbackURL = "http://localhost:9000/api/auth/oidc/callback"
```
fm3, line 438: Should be WkConf.Http.uri instead of hard-coded localhost:9000
```suggestion
  lazy val absoluteCallbackUrl = s"${conf.Http.uri}/api/auth/oidc/callback"
```

And I think it would be fair to move this to the OpenIdConnectClient class, as it is only used there

app/controllers/AuthenticationController.scala
```diff
@@ -429,6 +431,91 @@ class AuthenticationController @Inject()(
...
+
+  // Claims from https://openid.net/specs/openid-connect-core-1_0.html#StandardClaims
+  case class OpenConnectId(iss: String,
```
fm3, line 441: I’d move this to the OpenIdConnectClient file as well, and import it here if needed

app/controllers/AuthenticationController.scala
```diff
@@ -54,6 +55,7 @@ class AuthenticationController @Inject()(
     conf: WkConf,
     annotationDAO: AnnotationDAO,
     wkSilhouetteEnvironment: WkSilhouetteEnvironment,
+    openIdClient: OpenIDConnectClient,
```
fm3, line 58:
```suggestion
    openIdConnectClient: OpenIDConnectClient,
```

app/oxalis/security/OpenIdConnectClient.scala
```diff
@@ -0,0 +1,93 @@
...
+import java.nio.charset.StandardCharsets
+import javax.inject.Inject
+import scala.concurrent.ExecutionContext
+
+class OpenIDConnectClient @Inject()(rpc: RPC, conf: WkConf)(implicit executionContext: ExecutionContext) {
+
+  /*
+   Build redirect URL to redirect to OIDC provider for auth request (https://openid.net/specs/openid-connect-core-1_0.html#AuthRequest)
+   */
+  def redirectUrl(openIdClient: OpenIdConnectConfig, redirectUrl: String): Fox[String] =
```
fm3, line 21:
```suggestion
  def getRedirectUrl: Fox[String] =
```

I’d add the `get` because `redirect` also reads a bit like a verb.

I’d say it is not necessary to pass the parameters in from the caller, but rather have lazy vals for the client itself, I thinkt it already has everything it needs to construct them (mostly the injected WkConf).
Maybe the OpenIdConnectConfig case class is not needed at all, but the fields can be read from the passed WkConf where needed?

app/controllers/AuthenticationController.scala
```diff
@@ -429,6 +431,91 @@ class AuthenticationController @Inject()(
...
+              GlobalAccessContext)
+            user: User <- userService.insert(organization._id,
+                                             oidc.email,
+                                             oidc.given_name,
+                                             oidc.family_name,
+                                             isActive = true,
+                                             userService.getOIDCPasswordInfo)
+            multiUser: MultiUser <- multiUserDAO.findOne(user._multiUser)(GlobalAccessContext)
+            _ = analyticsService.track(SignupEvent(user, hadInvite = false))
+            //  brainDBResult <- brainTracing.registerIfNeeded(user, signUpData.password).toFox // Is this necessary?
```
fm3, line 492: Not needed, as discussed in person.

app/controllers/AuthenticationController.scala
```diff
@@ -429,6 +431,91 @@ class AuthenticationController @Inject()(
...
+            _ = if (conf.Features.isDemoInstance) {
+              mailchimpClient.registerUser(user, multiUser, tag = MailchimpTag.RegisteredAsUser)
+            } else {
+              Mailer ! Send(defaultMails.newUserMail(user.name, oidc.email, None, enableAutoVerify = true))
+            }
```
fm3, line 497: I wonder if some of this function’s logic can be extracted into a method. A lot of this exists twice now, once for oidc and once for the normal registration/login

app/controllers/AuthenticationController.scala
```diff
@@ -429,6 +431,91 @@ class AuthenticationController @Inject()(
...
+      }
+  }
+
+  def openIdCallback(): Action[AnyContent] = Action.async { implicit request =>
+    for {
+      code <- openIdClient.getToken(oidcConfig,
+                                    absoluteOidcCallbackURL,
+                                    request.queryString.get("code").flatMap(_.headOption).getOrElse("missing code"),
+      )
+      oidc: OpenConnectId <- validateJsValue[OpenConnectId](code).toFox
```
fm3, line 514:
```suggestion
      openConnectId <- validateJsValue[OpenConnectId](code).toFox
```
(if I understand correctly that OpenIdConnect uses OpenConnectIds? Seems confusing to me, but may be correct if that’s what the protocol states)

fm3 (commented):
Thanks for fighting your way through this! I think this is already going in a great direction. I added a few comments, mostly about naming (let’s avoid the abbreviation oidc), and about accessing the config.

We might want to discuss later if we want to redirect the user directly or keep the API somewhat rest-ful, and return json uris and do the redirection in the front-end but that is a change we can build later on, when a frontend is added.

app/utils/WkConf.scala
frcroth: sure

conf/application.conf
frcroth: Agreed, since these are values for testing on my side. However, I think this could be done last. Should there also be a configuration option that disables OIDC entirely? Or should that be reflected by empty values? Deactivated OIDC should result in some different error messages on the routes.

fm3: > I think this could be done last

fair :)

> Should there also be a configuration option that disables OIDC entirely? Or should that be reflected by empty values? Deactivated OIDC should result in some different error messages on the routes.

I’d say an empty providerUrl means oidc auth is disabled. Some of the config options have this semantic already. I agree that the routes should have an assertion on this and emit a readable error message then.

app/controllers/AuthenticationController.scala
frcroth: I'm not sure about moving this to the oidcc class, because the client should exist independently of the controller and the callback url is very specific controller knowledge

app/controllers/AuthenticationController.scala
frcroth: Agreed

app/controllers/AuthenticationController.scala
frcroth: This is not something that is directly part of the protocol. I renamed it to OpenIdConnectClaimSet since it describes it more accurately

app/controllers/AuthenticationController.scala
frcroth: done now

app/utils/WkConf.scala
```diff
@@ -64,6 +64,13 @@ class WkConf @Inject()(configuration: Configuration) extends ConfigReader with L
     val children = List(User, Tasks, Cache, SampleOrganization)
   }
 
+  object SingleSignOn {
+    object OIDC {
+      val providerUrl: String = get[String]("singleSignOn.OIDC.providerURL")
+      val clientId: String = get[String]("singleSignOn.OIDC.clientID")
+    }
```
fm3, line 71:
```suggestion
    object OpenIdConnect {
      val providerUrl: String = get[String]("singleSignOn.openIdConnect.providerUrl")
      val clientId: String = get[String]("singleSignOn.openIdConnect.clientId")
    }
```
I’d prefer to avoid abbreviations and all-uppercase URL/ID also in the config :)

app/controllers/AuthenticationController.scala
```diff
@@ -429,6 +453,58 @@ class AuthenticationController @Inject()(
     }
   }
 
+  lazy val absoluteCallbackURL = s"${conf.Http.uri}/api/auth/oidc/callback"
```
fm3, line 456:
```suggestion
  lazy val absoluteOpenIdConnectCallbackUrl = s"${conf.Http.uri}/api/auth/oidc/callback"
```
(if this stays in the controller, I’d prefer it be named explicitly to more quickly know what it is about)

app/controllers/AuthenticationController.scala
```diff
@@ -429,6 +453,58 @@ class AuthenticationController @Inject()(
...
+
+  def loginViaOpenIdConnect(): Action[AnyContent] = sil.UserAwareAction.async { implicit request =>
+    openIdConnectClient.getRedirectUrl(absoluteCallbackURL).map(url => Redirect(url))
+  }
+
+  def loginUser(user: User, loginInfo: LoginInfo)(implicit request: Request[AnyContent]): Future[Result] =
```
fm3, line 462: looks like the passed user is not used (a new variable by the same name is immediately fetched using loginInfo). In this case, I’d remove the user argument

fm3 (commented):
Nice work refactoring! I added some more comments (far fewer this time).

Also, please coordinate with @philippotto which parts should be moved to the front-end and what the API should look like then (redirect in frontend for example)

philippotto:
I finally got around to implementing the front-end part. The log in form now has a SSO button which redirects as necessary (once) which already works for me. #6596  mentions that the backend does a redirect which should be done by the frontend. I wasn't sure how exactly this should look like. @frcroth Could you remove the backend redirection and tell me what the frontend should do exactly?

Also I noticed the following things during testing:
- I needed to adapt `application.conf` so that *both* `OpenIdConnect` and `openIdConnect` exist. Otherwise, wk would complain that one config section was missing.
- If the keycloak user has no email (which was the case for the default user), wk shows this error `{"messages":[{"error":"Validating Json Failed: {\"obj.given_name\":[{\"msg\":[\"error.path.missing\"],\"args\":[]}],\"obj.family_name\":[{\"msg\":[\"error.path.missing\"],\"args\":[]}],\"obj.email\":[{\"msg\":[\"error.path.missing\"],\"args\":[]}]}"}]}` after sign in. since this is a rare edge case, this might be alright.
- If the keycloak user has no name (which was the case for the default user), wk will show an empty name. Maybe change it to anonymous?
- If the name was updated in keycloak, wk will still show the old name. Not sure whether this is important, but could lead to confusion if the data sources diverge.

Apart from the above, some questions:

- If the organization is empty, will the first user that registered be granted admin rights? That's probably a good idea.
- Can the backend provide information, whether SSO is supported? maybe via the `features` block in application.conf. Then, the frontend will know whether to display the corresponding UI.
- SSO won't be relevant for webknossos.org for now, right? otherwise, we would need to have multiple singleSignOn providers.

@frcroth also, could you pull the master into this branch and resolve the merge conflicts? thank you :)

frcroth:
You already implemented the necessary redirect. A redirect for the actual login only makes sense in the backend IMO since the front end is not involved. The application.conf problem should be fixed. I added the feature flag. Should this disable the backend routes?

>     If the organization is empty, will the first user that registered be granted admin rights? That's probably a good idea.
>     Can the backend provide information, whether SSO is supported? maybe via the features block in application.conf. Then, the frontend will know whether to display the corresponding UI.
>     SSO won't be relevant for webknossos.org for now, right? otherwise, we would need to have multiple singleSignOn providers.

1. Currently not as I understand it
2. Do you want another route to check that?
3. I don't know

fm3:
>    Currently not as I understand it

Making the first user of the orga admin sounds like a good idea.


> Do you want another route to check that?

The whole features config block is exposed via the features route, so the frontend can check it that way now.

> Should this disable the backend routes?

I’d say so, yes. As we checked the provider.isEmpty before, now this should be ANDed with the feature flag.

>    I don't know

I think multiple providers is something that we may add in the future, but not in this iteration. The feature is not currently built for webknossos.org

frcroth:
Making the first user of the orga admin could also be a separate issue, right? Or should this be adressed here?

fm3:
If I understand this correctly, there is currently no way of creating a new orga via oidc, right? I’d even think there is no way to create an empty organization via webKnossos. So the very first user would presumably have to either sign up in the traditional way, or, more likely, we create the organization manually, in which case I’d say it is also fair to set the user to admin manually. So yes, I’d say we will see what other user flows can be implemented when we see that they are needed for real-world usage.

philippotto:
> So yes, I’d say we will see what other user flows can be implemented when we see that they are needed for real-world usage. Do you agree @frcroth @philippotto ?

Yes, I agree :+1:

I incorporated the new feature flag to hide the SSO button if oidc is disabled. From my point of view, this PR could be ready to be merged? Do you agree @fm3 @frcroth ?


> If the name was updated in keycloak, wk will still show the old name. Not sure whether this is important, but could lead to confusion if the data sources diverge.

Just want to re-raise this: Do you have thoughts on that? I don't think it's important for the first iteration, but I'd be interested in your thoughts.

fm3:
> If the name was updated in keycloak, wk will still show the old name

I’d say this is pretty harmless. If a user reports a divergence, it is trivial to change the name in the db or add this feature. I’d not include this now.

> could be ready to be merged?

Could you edit the PR description to include full steps to test for the front-end too? I’d like to test the differing scenarios again. (Also I guess someone should have a quick look at the frontend code changes?). Other than that, I guess it’s good to go (with the new feature flag being set to false, the other new values in application.conf shouldn’t do harm, and may even help with understanding what needs to be inserted there for production later on)

philippotto:
> > could be ready to be merged?
> 
> Could you edit the PR description to include full steps to test for the front-end too?

Done!

> Also I guess someone should have a quick look at the frontend code changes?

@daniel-wer Could you have a short look?

fm3:
Any chance we could rename features.oidcEnabled to features.openIdConnectEnabled? That would be more consistent with the other application.conf keys. (As mentioned above we tried to avoid the acronym for consistency)

fm3:
@frcroth what exactly is OIDC Implicit flow? I noticed that in the steps to test the client has `implicitFlowEnabled: false` but you named the PR this way – does that mean we are testing something else?

fm3 (commented):
I just re-tested, worked as expected. Maybe you could add to the steps to test that a user with name, email and password has to exist in keycloak.

I’d say from my perspective this is ready to go :)

daniel-wer (commented):
Frontend code LGTM :+1:

frcroth:
> @frcroth what exactly is OIDC Implicit flow? I noticed that in the steps to test the client has `implicitFlowEnabled: false` but you named the PR this way – does that mean we are testing something else?

Yes, I was a bit confused. We are using code flow authorization with openid scope (https://darutk.medium.com/diagrams-of-all-the-openid-connect-flows-6968e3990660). I was a bit confused because we do not use a client secret atm, which is often associated with implicit flow, but it works because keycloak is configured to not use authorization.

frcroth:
> Any chance we could rename features.oidcEnabled to features.openIdConnectEnabled? That would be more consistent with the other application.conf keys. (As mentioned above we tried to avoid the acronym for consistency)

Did that in the backend, frontend needs to be updated, also e2e tests will fail again, but I couldn't build them new after  `yarn remove-snapshots` with `docker compose up e2e-tests` failing

fm3:
Thanks!

> but I couldn't build them new after yarn remove-snapshots with docker compose up e2e-tests failing

did that work before? `yarn refresh-snapshots` is the standard way now, but I think that’s pretty much the same thing you described. What’s the error output? 🤔

frcroth:
> Thanks!
> 
> > but I couldn't build them new after yarn remove-snapshots with docker compose up e2e-tests failing
> 
> did that work before? `yarn refresh-snapshots` is the standard way now, but I think that’s pretty much the same thing you described. What’s the error output? thinking

I get 
```
webknossos-e2e-tests-1  | 
webknossos-e2e-tests-1  |   Uncaught exception in public-test/test-bundle/test/backend-snapshot-tests/annotations.e2e.js
webknossos-e2e-tests-1  | 
webknossos-e2e-tests-1  |   Error: Cannot find module 'lz4-wasm-nodejs'
webknossos-e2e-tests-1  |   Require stack:
webknossos-e2e-tests-1  |   - /home/sbt-user/webknossos/public-test/test-bundle/test/mocks/lz4.js
webknossos-e2e-tests-1  |   - /home/sbt-user/webknossos/public-test/test-bundle/test/enzyme/e2e-setup.js
webknossos-e2e-tests-1  |   - /home/sbt-user/webknossos/public-test/test-bundle/test/backend-snapshot-tests/annotations.e2e.js
webknossos-e2e-tests-1  |   - /home/sbt-user/webknossos/node_modules/ava/lib/worker/subprocess.js
webknossos-e2e-tests-1  | 
webknossos-e2e-tests-1  |   › - node_modules/ava/lib/worker/subprocess.js
webknossos-e2e-tests-1  |   › Function.Module._load (node_modules/mock-require/index.js:14:22)
webknossos-e2e-tests-1  |   › Object.<anonymous> (public-test/test-bundle/test/mocks/lz4.js:5:35)
webknossos-e2e-tests-1  | 
webknossos-e2e-tests-1  |   ✖ public-test/test-bundle/test/backend-snapshot-tests/annotations.e2e.js exited with a non-zero exit code: 1
webknossos-e2e-tests-1  | 
webknossos-e2e-tests-1  |   Uncaught exception in public-test/test-bundle/test/backend-snapshot-tests/datasets.e2e.js
webknossos-e2e-tests-1  | 
webknossos-e2e-tests-1  |   Error: Cannot find module 'lz4-wasm-nodejs'
webknossos-e2e-tests-1  |   Require stack:
webknossos-e2e-tests-1  |   - /home/sbt-user/webknossos/public-test/test-bundle/test/mocks/lz4.js
webknossos-e2e-tests-1  |   - /home/sbt-user/webknossos/public-test/test-bundle/test/enzyme/e2e-setup.js
webknossos-e2e-tests-1  |   - /home/sbt-user/webknossos/public-test/test-bundle/test/backend-snapshot-tests/datasets.e2e.js
webknossos-e2e-tests-1  |   - /home/sbt-user/webknossos/node_modules/ava/lib/worker/subprocess.js
webknossos-e2e-tests-1  | 
webknossos-e2e-tests-1  |   › - node_modules/ava/lib/worker/subprocess.js
webknossos-e2e-tests-1  |   › Function.Module._load (node_modules/mock-require/index.js:14:22)
webknossos-e2e-tests-1  |   › Object.<anonymous> (public-test/test-bundle/test/mocks/lz4.js:5:35)
webknossos-e2e-tests-1  | 
webknossos-e2e-tests-1  |   ✖ public-test/test-bundle/test/backend-snapshot-tests/datasets.e2e.js exited with a non-zero exit code: 1
webknossos-e2e-tests-1  | 
webknossos-e2e-tests-1  |   Uncaught exception in public-test/test-bundle/test/backend-snapshot-tests/misc.e2e.js
webknossos-e2e-tests-1  | 
webknossos-e2e-tests-1  |   Error: Cannot find module 'lz4-wasm-nodejs'
webknossos-e2e-tests-1  |   Require stack:
webknossos-e2e-tests-1  |   - /home/sbt-user/webknossos/public-test/test-bundle/test/mocks/lz4.js
webknossos-e2e-tests-1  |   - /home/sbt-user/webknossos/public-test/test-bundle/test/enzyme/e2e-setup.js
webknossos-e2e-tests-1  |   - /home/sbt-user/webknossos/public-test/test-bundle/test/backend-snapshot-tests/misc.e2e.js
webknossos-e2e-tests-1  |   - /home/sbt-user/webknossos/node_modules/ava/lib/worker/subprocess.js
webknossos-e2e-tests-1  | 
webknossos-e2e-tests-1  |   › - node_modules/ava/lib/worker/subprocess.js
webknossos-e2e-tests-1  |   › Function.Module._load (node_modules/mock-require/index.js:14:22)
webknossos-e2e-tests-1  |   › Object.<anonymous> (public-test/test-bundle/test/mocks/lz4.js:5:35)
webknossos-e2e-tests-1  | 
webknossos-e2e-tests-1  |   ✖ public-test/test-bundle/test/backend-snapshot-tests/misc.e2e.js exited with a non-zero exit code: 1
webknossos-e2e-tests-1  | 
webknossos-e2e-tests-1  |   Uncaught exception in public-test/test-bundle/test/backend-snapshot-tests/projects.e2e.js
webknossos-e2e-tests-1  | 
webknossos-e2e-tests-1  |   Error: Cannot find module 'lz4-wasm-nodejs'
webknossos-e2e-tests-1  |   Require stack:
webknossos-e2e-tests-1  |   - /home/sbt-user/webknossos/public-test/test-bundle/test/mocks/lz4.js
webknossos-e2e-tests-1  |   - /home/sbt-user/webknossos/public-test/test-bundle/test/enzyme/e2e-setup.js
webknossos-e2e-tests-1  |   - /home/sbt-user/webknossos/public-test/test-bundle/test/backend-snapshot-tests/projects.e2e.js
webknossos-e2e-tests-1  |   - /home/sbt-user/webknossos/node_modules/ava/lib/worker/subprocess.js
webknossos-e2e-tests-1  | 
webknossos-e2e-tests-1  |   › - node_modules/ava/lib/worker/subprocess.js
webknossos-e2e-tests-1  |   › Function.Module._load (node_modules/mock-require/index.js:14:22)
webknossos-e2e-tests-1  |   › Object.<anonymous> (public-test/test-bundle/test/mocks/lz4.js:5:35)
webknossos-e2e-tests-1  | 
webknossos-e2e-tests-1  |   ✖ public-test/test-bundle/test/backend-snapshot-tests/projects.e2e.js exited with a non-zero exit code: 1
webknossos-e2e-tests-1  | 
webknossos-e2e-tests-1  |   Uncaught exception in public-test/test-bundle/test/backend-snapshot-tests/scripts.e2e.js
webknossos-e2e-tests-1  | 
webknossos-e2e-tests-1  |   Error: Cannot find module 'lz4-wasm-nodejs'
webknossos-e2e-tests-1  |   Require stack:
webknossos-e2e-tests-1  |   - /home/sbt-user/webknossos/public-test/test-bundle/test/mocks/lz4.js
webknossos-e2e-tests-1  |   - /home/sbt-user/webknossos/public-test/test-bundle/test/enzyme/e2e-setup.js
webknossos-e2e-tests-1  |   - /home/sbt-user/webknossos/public-test/test-bundle/test/backend-snapshot-tests/scripts.e2e.js
webknossos-e2e-tests-1  |   - /home/sbt-user/webknossos/node_modules/ava/lib/worker/subprocess.js
webknossos-e2e-tests-1  | 
webknossos-e2e-tests-1  |   › - node_modules/ava/lib/worker/subprocess.js
webknossos-e2e-tests-1  |   › Function.Module._load (node_modules/mock-require/index.js:14:22)
webknossos-e2e-tests-1  |   › Object.<anonymous> (public-test/test-bundle/test/mocks/lz4.js:5:35)
webknossos-e2e-tests-1  | 
webknossos-e2e-tests-1  |   ✖ public-test/test-bundle/test/backend-snapshot-tests/scripts.e2e.js exited with a non-zero exit code: 1
webknossos-e2e-tests-1  | 
webknossos-e2e-tests-1  |   Uncaught exception in public-test/test-bundle/test/backend-snapshot-tests/tasks.e2e.js
webknossos-e2e-tests-1  | 
webknossos-e2e-tests-1  |   Error: Cannot find module 'lz4-wasm-nodejs'
webknossos-e2e-tests-1  |   Require stack:
webknossos-e2e-tests-1  |   - /home/sbt-user/webknossos/public-test/test-bundle/test/mocks/lz4.js
webknossos-e2e-tests-1  |   - /home/sbt-user/webknossos/public-test/test-bundle/test/enzyme/e2e-setup.js
webknossos-e2e-tests-1  |   - /home/sbt-user/webknossos/public-test/test-bundle/test/backend-snapshot-tests/tasks.e2e.js
webknossos-e2e-tests-1  |   - /home/sbt-user/webknossos/node_modules/ava/lib/worker/subprocess.js
webknossos-e2e-tests-1  | 
webknossos-e2e-tests-1  |   › - node_modules/ava/lib/worker/subprocess.js
webknossos-e2e-tests-1  |   › Function.Module._load (node_modules/mock-require/index.js:14:22)
webknossos-e2e-tests-1  |   › Object.<anonymous> (public-test/test-bundle/test/mocks/lz4.js:5:35)
webknossos-e2e-tests-1  | 
webknossos-e2e-tests-1  |   ✖ public-test/test-bundle/test/backend-snapshot-tests/tasks.e2e.js exited with a non-zero exit code: 1
webknossos-e2e-tests-1  | 
webknossos-e2e-tests-1  |   Uncaught exception in public-test/test-bundle/test/backend-snapshot-tests/tasktypes.e2e.js
webknossos-e2e-tests-1  | 
webknossos-e2e-tests-1  |   Error: Cannot find module 'lz4-wasm-nodejs'
webknossos-e2e-tests-1  |   Require stack:
webknossos-e2e-tests-1  |   - /home/sbt-user/webknossos/public-test/test-bundle/test/mocks/lz4.js
webknossos-e2e-tests-1  |   - /home/sbt-user/webknossos/public-test/test-bundle/test/enzyme/e2e-setup.js
webknossos-e2e-tests-1  |   - /home/sbt-user/webknossos/public-test/test-bundle/test/backend-snapshot-tests/tasktypes.e2e.js
webknossos-e2e-tests-1  |   - /home/sbt-user/webknossos/node_modules/ava/lib/worker/subprocess.js
webknossos-e2e-tests-1  | 
webknossos-e2e-tests-1  |   › - node_modules/ava/lib/worker/subprocess.js
webknossos-e2e-tests-1  |   › Function.Module._load (node_modules/mock-require/index.js:14:22)
webknossos-e2e-tests-1  |   › Object.<anonymous> (public-test/test-bundle/test/mocks/lz4.js:5:35)
webknossos-e2e-tests-1  | 
webknossos-e2e-tests-1  |   ✖ public-test/test-bundle/test/backend-snapshot-tests/tasktypes.e2e.js exited with a non-zero exit code: 1
webknossos-e2e-tests-1  | 
webknossos-e2e-tests-1  |   Uncaught exception in public-test/test-bundle/test/backend-snapshot-tests/teams.e2e.js
webknossos-e2e-tests-1  | 
webknossos-e2e-tests-1  |   Error: Cannot find module 'lz4-wasm-nodejs'
webknossos-e2e-tests-1  |   Require stack:
webknossos-e2e-tests-1  |   - /home/sbt-user/webknossos/public-test/test-bundle/test/mocks/lz4.js
webknossos-e2e-tests-1  |   - /home/sbt-user/webknossos/public-test/test-bundle/test/enzyme/e2e-setup.js
webknossos-e2e-tests-1  |   - /home/sbt-user/webknossos/public-test/test-bundle/test/backend-snapshot-tests/teams.e2e.js
webknossos-e2e-tests-1  |   - /home/sbt-user/webknossos/node_modules/ava/lib/worker/subprocess.js
webknossos-e2e-tests-1  | 
webknossos-e2e-tests-1  |   › - node_modules/ava/lib/worker/subprocess.js
webknossos-e2e-tests-1  |   › Function.Module._load (node_modules/mock-require/index.js:14:22)
webknossos-e2e-tests-1  |   › Object.<anonymous> (public-test/test-bundle/test/mocks/lz4.js:5:35)
webknossos-e2e-tests-1  | 
webknossos-e2e-tests-1  |   ✖ public-test/test-bundle/test/backend-snapshot-tests/teams.e2e.js exited with a non-zero exit code: 1
webknossos-e2e-tests-1  | 
webknossos-e2e-tests-1  |   Uncaught exception in public-test/test-bundle/test/backend-snapshot-tests/teamstructure.e2e.js
webknossos-e2e-tests-1  | 
webknossos-e2e-tests-1  |   Error: Cannot find module 'lz4-wasm-nodejs'
webknossos-e2e-tests-1  |   Require stack:
webknossos-e2e-tests-1  |   - /home/sbt-user/webknossos/public-test/test-bundle/test/mocks/lz4.js
webknossos-e2e-tests-1  |   - /home/sbt-user/webknossos/public-test/test-bundle/test/enzyme/e2e-setup.js
webknossos-e2e-tests-1  |   - /home/sbt-user/webknossos/public-test/test-bundle/test/backend-snapshot-tests/teamstructure.e2e.js
webknossos-e2e-tests-1  |   - /home/sbt-user/webknossos/node_modules/ava/lib/worker/subprocess.js
webknossos-e2e-tests-1  | 
webknossos-e2e-tests-1  |   › - node_modules/ava/lib/worker/subprocess.js
webknossos-e2e-tests-1  |   › Function.Module._load (node_modules/mock-require/index.js:14:22)
webknossos-e2e-tests-1  |   › Object.<anonymous> (public-test/test-bundle/test/mocks/lz4.js:5:35)
webknossos-e2e-tests-1  | 
webknossos-e2e-tests-1  |   ✖ public-test/test-bundle/test/backend-snapshot-tests/teamstructure.e2e.js exited with a non-zero exit code: 1
webknossos-e2e-tests-1  | 
webknossos-e2e-tests-1  |   Uncaught exception in public-test/test-bundle/test/backend-snapshot-tests/timetracking.e2e.js
webknossos-e2e-tests-1  | 
webknossos-e2e-tests-1  |   Error: Cannot find module 'lz4-wasm-nodejs'
webknossos-e2e-tests-1  |   Require stack:
webknossos-e2e-tests-1  |   - /home/sbt-user/webknossos/public-test/test-bundle/test/mocks/lz4.js
webknossos-e2e-tests-1  |   - /home/sbt-user/webknossos/public-test/test-bundle/test/enzyme/e2e-setup.js
webknossos-e2e-tests-1  |   - /home/sbt-user/webknossos/public-test/test-bundle/test/backend-snapshot-tests/timetracking.e2e.js
webknossos-e2e-tests-1  |   - /home/sbt-user/webknossos/node_modules/ava/lib/worker/subprocess.js
webknossos-e2e-tests-1  | 
webknossos-e2e-tests-1  |   › - node_modules/ava/lib/worker/subprocess.js
webknossos-e2e-tests-1  |   › Function.Module._load (node_modules/mock-require/index.js:14:22)
webknossos-e2e-tests-1  |   › Object.<anonymous> (public-test/test-bundle/test/mocks/lz4.js:5:35)
webknossos-e2e-tests-1  | 
webknossos-e2e-tests-1  |   ✖ public-test/test-bundle/test/backend-snapshot-tests/timetracking.e2e.js exited with a non-zero exit code: 1
webknossos-e2e-tests-1  | 
webknossos-e2e-tests-1  |   Uncaught exception in public-test/test-bundle/test/backend-snapshot-tests/users.e2e.js
webknossos-e2e-tests-1  | 
webknossos-e2e-tests-1  |   Error: Cannot find module 'lz4-wasm-nodejs'
webknossos-e2e-tests-1  |   Require stack:
webknossos-e2e-tests-1  |   - /home/sbt-user/webknossos/public-test/test-bundle/test/mocks/lz4.js
webknossos-e2e-tests-1  |   - /home/sbt-user/webknossos/public-test/test-bundle/test/enzyme/e2e-setup.js
webknossos-e2e-tests-1  |   - /home/sbt-user/webknossos/public-test/test-bundle/test/backend-snapshot-tests/users.e2e.js
webknossos-e2e-tests-1  |   - /home/sbt-user/webknossos/node_modules/ava/lib/worker/subprocess.js
webknossos-e2e-tests-1  | 
webknossos-e2e-tests-1  |   › - node_modules/ava/lib/worker/subprocess.js
webknossos-e2e-tests-1  |   › Function.Module._load (node_modules/mock-require/index.js:14:22)
webknossos-e2e-tests-1  |   › Object.<anonymous> (public-test/test-bundle/test/mocks/lz4.js:5:35)
webknossos-e2e-tests-1  | 
webknossos-e2e-tests-1  |   ✖ public-test/test-bundle/test/backend-snapshot-tests/users.e2e.js exited with a non-zero exit code: 1
webknossos-e2e-tests-1  |   ─
webknossos-e2e-tests-1  | 
webknossos-e2e-tests-1  |   11 uncaught exceptions
webknossos-e2e-tests-1  | error Command failed with exit code 1.
webknossos-e2e-tests-1  | info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
webknossos-e2e-tests-1  | 2022-11-24 16:24:30,916 [INFO] akka.actor.CoordinatedShutdown - Running CoordinatedShutdown with reason [ServerStoppedReason]
webknossos-e2e-tests-1  | 2022-11-24 16:24:30,952 [INFO] Startup - Deleting temporary files
webknossos-e2e-tests-1  | 2022-11-24 16:24:31,007 [INFO] Startup - Closing SQL Database handle
webknossos-e2e-tests-1  | 2022-11-24 16:24:31,013 [INFO] com.zaxxer.hikari.HikariDataSource - slick.db - Shutdown initiated...
webknossos-e2e-tests-1  | 2022-11-24 16:24:31,056 [INFO] com.zaxxer.hikari.HikariDataSource - slick.db - Shutdown completed.
webknossos-e2e-tests-1  | 2022-11-24 16:24:31,065 [INFO] com.scalableminds.webknossos.tracingstore.tracings.TracingDataStore - Closing TracingStore grpc channels...
webknossos-e2e-tests-1  | 2022-11-24 16:24:31,117 [INFO] swagger - Swagger - stopped.
webknossos-e2e-tests-1  | 2022-11-24 16:24:31,119 [DEBUG] controllers.AssetsMetadataProvider - Cleaning AssetsMetadata instance
webknossos-e2e-tests-1  | [error]   x pass the e2e tests
webknossos-e2e-tests-1  | [error]    1 != 0 (End2EndSpec.scala:35)
webknossos-e2e-tests-1  | [info] Total for specification End2EndSpec
webknossos-e2e-tests-1  | [info] Finished in 1 minute 18 seconds, 62 ms
webknossos-e2e-tests-1  | [info] 1 example, 1 failure, 0 error
webknossos-e2e-tests-1  | [error] Failed: Total 1, Failed 1, Errors 0, Passed 0
webknossos-e2e-tests-1  | [error] Failed tests:
webknossos-e2e-tests-1  | [error]       e2e.End2EndSpec
webknossos-e2e-tests-1  | [error] (Test / testOnly) sbt.TestsFailedException: Tests unsuccessful
webknossos-e2e-tests-1  | [error] Total time: 130 s (02:10), completed Nov 24, 2022, 4:24:34 PM
webknossos-e2e-tests-1 exited with code 1

```
,also for yarn install I get:

```
[4/6] ⢀ core-js
error /home/felix/scm/webknossos/node_modules/rome: Command failed.
Exit code: 1
Command: node scripts/postinstall.js
Arguments: 
Directory: /home/felix/scm/webknossos/node_modules/rome
Output:
/home/felix/scm/webknossos/node_modules/rome/scripts/postinstall.js:18
const binName = PLATFORMS?.[platform]?.[arch];
                          ^

SyntaxError: Unexpected token '.'
    at Object.compileFunction (vm.js:344:18)
    at wrapSafe (internal/modules/cjs/loader.js:1048:15)
    at Module._compile (internal/modules/cjs/loader.js:1082:27)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1138:10)
    at Module.load (internal/modules/cjs/loader.js:982:32)

```

fm3:
`yarn install` was the right call for the failing snapshot tests. The new lib `rome` was added to the package.json in https://github.com/scalableminds/webknossos/pull/6618 – so this is likely an effect of pulling the master into this branch.

Now why exactly yarn install fails, I’m uncertain. Which node version do you have? @philippotto do you have any insights here? Could it be specific to the linux distribution?

In any case, I can refresh the snapshots here to avoid blocking this PR by the install issue. But we should figure that out.

frcroth:
Nvm, with the new command the e2e tests run :shrug:

philippotto:
@frcroth Could you have a look at the CI? See https://app.circleci.com/pipelines/github/scalableminds/webknossos/9133/workflows/463f4087-80d1-43fc-9a94-a4fb2b9cb8cc/jobs/20501

```
[info] compiling 6 Scala sources and 1 Java source to /home/circleci/webknossos/target/scala-2.12/classes ...
[error] /home/circleci/webknossos/app/controllers/AuthenticationController.scala:125:26: private default argument in class AuthenticationController is never used
[error]                          autoActivate: Boolean = true,
[error]                          ^
[error] one error found
[error] (Compile / compileIncremental) Compilation failed
```

frcroth:
> @frcroth Could you have a look at the CI? See [app.circleci.com/pipelines/github/scalableminds/webknossos/9133/workflows/463f4087-80d1-43fc-9a94-a4fb2b9cb8cc/jobs/20501](https://app.circleci.com/pipelines/github/scalableminds/webknossos/9133/workflows/463f4087-80d1-43fc-9a94-a4fb2b9cb8cc/jobs/20501)
> 
> ```
> [info] compiling 6 Scala sources and 1 Java source to /home/circleci/webknossos/target/scala-2.12/classes ...
> [error] /home/circleci/webknossos/app/controllers/AuthenticationController.scala:125:26: private default argument in class AuthenticationController is never used
> [error]                          autoActivate: Boolean = true,
> [error]                          ^
> [error] one error found
> [error] (Compile / compileIncremental) Compilation failed
> ```

@philippotto Fixed now

frcroth:
@philippotto I think front end needs to be adapted again because a name changed.

> Any chance we could rename features.oidcEnabled to features.openIdConnectEnabled? That would be more consistent with the other application.conf keys. (As mentioned above we tried to avoid the acronym for consistency)

philippotto:
> @philippotto Fixed now

Great :+1: 

> @philippotto I think front end needs to be adapted again because a name changed.

Done! For such small things, such as renaming a var, feel free to do a search&replace in the frontend folder next time :)

fm3 (approved):

frcroth:
Will be merged next week.

PR #6536, opened 2022-10-06
Fix MeshChunk byteOffset (Long, not Int)

fm3:
ByteOffset can be larger than Int range, should be treated as Long all the way through.

- [x] Ready for review


daniel-wer (approved):
Thanks for fixing :+1:

PR #6537, opened 2022-10-06
Reload histogram when reloading layer

### LINKED ISSUE #4479: Also reload histogram when layer is reloaded from backend
Currently, only after the initialization of the tracing view, the histogram is being fetched from the backend. If the data in the backend changes and a layer is being reloaded the histogram should also update accordingly.

daniel-wer:
I had to move around the code in the histogram saga a little bit, reviewing ignoring white space changes should help.

- https://___.webknossos.xyz

### Steps to test:
- Open a dataset locally, remove some mag 1 wkw files on disk and reload the layer - the histogram should change

### Issues:
- fixes #4479 

------
- [x] Ready for review

### DIFF
--- a/frontend/javascripts/oxalis/view/statusbar.tsx
+++ b/frontend/javascripts/oxalis/view/statusbar.tsx
@@ -194,7 +194,7 @@ function ShortcutsInfo() {
         </span>
         <span key="zoom" className="shortcut-info-element">
           <span
-            key="zoom-i"
+            key="space-forward-i"
             className="keyboard-key-icon-small"
             style={{
               borderColor: lineColor,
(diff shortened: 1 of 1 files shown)


frontend/javascripts/oxalis/model/sagas/load_histogram_data_saga.ts
```diff
@@ -1,88 +1,99 @@
...
 
-async function fetchAllHistogramsForLayers(
-  dataLayers: Array<DataLayer>,
-  dataset: APIDataset,
-): Promise<HistogramDataForAllLayers> {
-  const histograms: HistogramDataForAllLayers = {};
+export default function* loadHistogramDataSaga(): Saga<void> {
+  yield* take("WK_READY");
+  yield* takeEvery("RELOAD_HISTOGRAM", reloadHistogramForLayer);
+  const dataLayers: Array<DataLayer> = yield* call([Model, Model.getColorLayers]) as any;
```
philippotto, line 19:
```suggestion
  const dataLayers: Array<DataLayer> = yield* call([Model, Model.getColorLayers]);
```

I think, it is not necessary anymore because of TS (at least I just tested this on another branch).

frontend/javascripts/oxalis/model/sagas/load_histogram_data_saga.ts
```diff
@@ -1,88 +1,99 @@
...
-    // Adjust the intensityRange of the layer to be within the range of the actual (sampled) data.
-    const histogram = histograms[layerName];
-    const allMinValues = histogram.map((currentHistogramData) => currentHistogramData.min);
-    const allMaxValues = histogram.map((currentHistogramData) => currentHistogramData.max);
-    const minimumInHistogramData = Math.min(...allMinValues);
-    const maximumInHistogramData = Math.max(...allMaxValues);
-    let newIntensityRange = [];
-    const currentLayerConfig = layerConfigurations[layerName];
+  const histogram = yield* call(loadHistogramForLayer, layerName, layerConfigurations[layerName]);
+  // Also update histogram data if it is undefined since apparently it can no longer be generated
```
philippotto, line 38: I don't understand the "since apparently it can no longer be generated" bit :thinking:

frontend/javascripts/oxalis/model/sagas/load_histogram_data_saga.ts
```diff
@@ -1,88 +1,99 @@
...
-
-      if (Array.isArray(data) && data.length > 0) {
-        histograms[dataLayer.name] = data;
-      }
-    } catch (e) {
-      console.warn(`Error: Could not fetch the histogram data for layer ${dataLayer.name}.`, e);
-    }
+    const layerName = dataLayer.name;
+    // We send the histogram data requests sequentially so there is less blockage of bucket data requests.
+    const histogram = yield* call(loadHistogramForLayer, layerName, layerConfigurations[layerName]);
```
philippotto, line 26: I think, it would simplify things if `loadHistogramForLayer` doesn't take the `layerConfiguration` parameter (especially, because there are two call sites where this needs to be prepared).

frontend/javascripts/oxalis/model/sagas/load_histogram_data_saga.ts
```diff
@@ -1,88 +1,99 @@
...
+    if (histogram != null) histograms[layerName] = histogram;
   }
 
-  return histograms;
+  yield* put(setHistogramDataAction(histograms));
```
philippotto, line 30: what do you think about calling `yield* put(setHistogramDataForLayerAction(layerName, histogram));` within the loop above? then, the action type `setHistogramDataAction` wouldn't be necessary anymore and the `HistogramDataForAllLayers` stuff probably wouldn't be, either.

philippotto (changes requested):
nice refactoring :+1: I left some suggestions, though :) also, while you are at it, the following would be cool (not hi-pri, but maybe quick to do):

If the histogram loading fails (this especially happens when I just restarted the server locally), there will be a toast indicating that. I would rather have no toast and then have the histogram view show something like "histogram not available. click here to try again".

frontend/javascripts/oxalis/model/sagas/load_histogram_data_saga.ts
daniel-wer: Yes that didn't make too much sense, I removed it.

frontend/javascripts/oxalis/model/sagas/load_histogram_data_saga.ts
daniel-wer: Yes, much better :)

daniel-wer:
Nice suggestions! I added a loading spinner while I was at it. See the gif for what it currently looks like, open for design suggestions :)

histogram_reload

philippotto (approved):
awesome stuff :) see only my minor style suggestion (might need prettier afterwards).

PR #6545, opened 2022-10-11
Switch default volume layer source for advanced annotation creation modal

hotzenklotz:
When creating a new annotation using the more advanced/detailed modal from the overflow menu, the default option for new annotations is to create an empty volume layer. Our internal discussion showed, that more often than not, you would likely want to base your volume annotation on an existing segmentation layer instead.

This PR adds the following changes:
- switch the default volume layer to existing segmentation. layer instead of empty one
- a small cosmetic CSS tweak to prevent longer mag labels from wrapping, e.g. `1024-1024-512`

switch default volume layer to existing segmentation layers instead of empty volume layer when using detailed modal to create a new annotation

image

- https://___.webknossos.xyz

### Steps to test:
- Go to dashboard
- Create new annotation using the overflow menu (three dots ...)
- Any existing segmentation layer should be pre-selected by default

### Issues:
- fixes None

------
- [x] Ready for review


frontend/javascripts/dashboard/advanced_dataset/create_explorative_modal.tsx
```diff
@@ -65,18 +65,19 @@ export function NewVolumeLayerSelection({
...
       >
-        <Radio key={-1} value={-1}>
-          Create empty layer
-        </Radio>
         {segmentationLayers.map((segmentationLayer, index) => (
           <Radio
             key={segmentationLayer.name}
             value={index}
             disabled={!doesSupportVolumeWithFallback(dataset, segmentationLayer)}
+            style={{ fontStyle: "italic" }}
```
philippotto, line 73: I don't really understand why this is italic now? Making `Create empty layer` italic would make some sense to me, but turning all items italic except for the "empty layer" seems counter-intuitive to me (highlighting everything but one). maybe remove this line again?

philippotto (approved):
looks good, but I'd remove the `italic` bit again. see my other comment.

frontend/javascripts/dashboard/advanced_dataset/create_explorative_modal.tsx
hotzenklotz: I wanted to highlight that the selection features a layer name and not a "technical term". In other words, the selectable options are named by a user and have no technical relevance.

In this screenshot I found it really odd that according to the wording my volume annotation is based an "a segmentation". `Segmentation` the name of a layer and not a technical property. This would have been clearer if the layer was called `synapse` or `cell borders` or something.
image

Maybe we can make this more clear in a different way than italics.

philippotto: ah, I see. what about putting the layer names in quotation marks to make it clear that these are layer names? or put " layer" behind each name?

PR #6546, opened 2022-10-12
Fix docs for the annotation download file format

fm3:
- [x] Ready for review

### DIFF
--- a/docs/rest_api.md
+++ b/docs/rest_api.md
@@ -344,12 +344,16 @@ Download an annotation as NML/ZIP
 
 #### Returns
  - As chunked file stream:
-   - In case of a volume annotation:
-     - A ZIP file containing both an NML file and a data.zip file with the volume data
-   - In case of a single explorative or task annotation with no volume annotation:
+   - In case of a single skeleton-only annotation:
      - A single NML file
+   - In case of a single annotation with volume layers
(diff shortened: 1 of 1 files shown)


docs/rest_api.md
```diff
@@ -344,13 +344,17 @@ Download an annotation as NML/ZIP
...
+   - In case of a single skeleton-only annotation:
      - A single NML file
+   - In case of a single annotation with volume layers
+     - A ZIP file containing both an NML file and data zip files with the volume data for each volume layer. The data zip file names are referenced in the `<volume>` tags in the NML file.
    - In case of compound downloads (CompoundTask/CompoundProject/CompoundTaskType):
      - A ZIP file containing individual NML files for all associated annotations
 
+
+#### Note
+ - The volume data zip file name used to always be `data.zip`, when wk supported only a single volume layer per annotation. Please use the file names referenced in the `<volume>` tags in the NML file instead.
```
jstriebel, line 356: You could also use an admonition from mkdocs here, see
https://squidfunk.github.io/mkdocs-material/reference/admonitions/#usage

(They are already configured in the docs, so just adding the syntax would be enough.)

Feel free to leave as-is as well.

jstriebel (approved):
LGTM :+1:

PR #6547, opened 2022-10-12
adds user_auth_token to every job and removes webknossos_token from jobs

normanrz:
This removes the job-specific `user_auth_token` and `webknossos_token` in favor of a new `user_auth_token` field that is always available to the worker when requesting jobs. The user token will be generated when the worker requests the job, so changed tokens in between scheduling and requesting will be used.

### Steps to test:
- Create a job and request it via the remote-wk-worker routes
- `user_auth_token` should always be present.

Code is ready for review. To test, I need to finish the changes in the worker.

------
- [ ] Needs worker update after deployment
- [x] Ready for review

### DIFF
--- a/app/models/job/Job.scala
+++ b/app/models/job/Job.scala
@@ -1,6 +1,5 @@
 package models.job
 
-import java.sql.Timestamp
 import akka.actor.ActorSystem
 import com.scalableminds.util.accesscontext.{DBAccessContext, GlobalAccessContext}
 import com.scalableminds.util.geometry.BoundingBox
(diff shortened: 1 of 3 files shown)


app/models/job/Job.scala
```diff
@@ -341,7 +341,7 @@ class JobService @Inject()(wkConf: WkConf,
       Json.obj(
         "id" -> job._id.id,
         "command" -> job.command,
-        "commandArgs" -> (job.commandArgs - "webknossos_token" - "user_auth_token"),
+        "commandArgs" -> (job.commandArgs - "webknossos_token"),
```
fm3, line 344:
```suggestion
        "commandArgs" -> (job.commandArgs - "webknossos_token" - "user_auth_token"),
```
This should remain here to ensure we are not leaking tokens used in existing jobs (otherwise a DB migration would be required removing them, that would be fine by me too)

fm3 (commented):
Mostly LGTM, see my small suggestion below.

Did you test that the user token has sufficient permissions to do what the webknossos_token was used for before? (I think dataset import triggering?) I’m not sure if the set of users who are allowed to trigger jobs is the same as users who may `administrateDataSources` (admins + team managers)

normanrz:
> Did you test that the user token has sufficient permissions to do what the webknossos_token was used for before? (I think dataset import triggering?) I’m not sure if the set of users who are allowed to trigger jobs is the same as users who may `administrateDataSources` (admins + team managers)

I have not tested that yet. How would I even test that properly?

fm3:
> How would I even test that properly?

I suppose by setting up a local wk environment with multiple users and starting jobs as a non-privileged user.

I’m not certain what behavior we want here. Should non-dataset managers/admins be allowed start jobs at all?

normanrz:
We finally want to tackle this in this sprint. Not sure if based on this PR or a new one, though. cc @fm3

normanrz:
I think we should make the changes to the UI to prevent members to run jobs that create datasets. That is not ideal, but unless we do something about the dataset vs. annotation situation this is the only consistent option.

fm3 (approved):

PR #6548, opened 2022-10-12
Look up annotation dataset by tracing id instead of (broken) organizationName field

### LINKED ISSUE #6521: volume tracing zattrs route broken for annotations on foreign organization datasets
got `"Unsuccessful WS request to https://webknossos.org/tracings/volume/zarr/379ffeaf-e8a9-4a75-b302-0615f92a4b1f/.zattrs (ID: 0).Status: 404. Response: {\"messages\":[{\"error\":\"Unsuccessful WS request to http://localhost:9000/api/tracingstores/localhost/dataSource/l4dense_motta_et_al_dev (ID: 2646).Status: 403. Response: {\\\"messages\\\":[{\\\"error\\\":\\\"Could not access DataSet l4dense_motta_et_al_dev. Does y"` – the private link token should have enabled this access

### LINKED ISSUE #6522: Tracing proto organizationName field – dataset vs user orga
Looks like createExplorational sets the organizationName field using the user orga, not the dataset orga, as it is described in the comment in the proto files. This creates problems like #6521 – we should consider how to deal with this disparity

fm3:
The organizationName field in the Tracing proto objects is used to look up the dataset.
Unfortunately, in createExplorational, not the dataset organization but the user organization was entered into that field, breaking the features which rely on this lookup for all annotations where the dataset orga is not the same as the user orga.
Those features include:
 - Ad-Hoc meshing with fallback layer
 - Zarr streming
 - Editable Mappings
 - Redownsampling

Fixing this in createExplorational will not repair existing tracing objects, mutating them in the fossildb is tricky. We should discuss how to proceed here:
- (A) Do nothing, these features will stay unavailable for foreign-datset annotations created before the fix
- (B) In the dataset lookup, on failure, look for datasets of the same name of other organizations that the user may read, and use those. This should repair *most* annotations, except where datasets of the same name exist in multiple organizations
- (C) Create a python script that fixes this in the fossildb retroactively for all tracing versions (hard, risky)
- (D) Create adapter code in wk that repairs all annotations when opening them (creates update actions)
- Edit: (E) from comment below: ditch the organizationName and datasetName fields entirely and perform the (cached) lookup each time by the tracing id (the wk-side postgres table should be able to connect tracing id and dataset via foreign keys)

This PR implements E

### TODO
 - [x] caching
 - [x] cleanup
 - [x] double-check that front-end does not use the relevant fields in proto
 - [x] nml writer: pass in dataset name

- https://___.webknossos.xyz

### Steps to test:
- Test the four features described above with a multi-organization setup on  a public dataset of orga A, with a user of orga B.

### Issues:
- fixes #6522
- fixes #6521

------
- [x] Ready for review


webknossos-datastore/app/com/scalableminds/webknossos/datastore/services/DSRemoteTracingstoreClient.scala
```diff
@@ -40,7 +40,7 @@ class DSRemoteTracingstoreClient @Inject()(
                      cxyz: String,
                      tracingStoreUri: String,
                      token: Option[String]): Fox[Array[Byte]] =
-    rpc(s"$tracingStoreUri/tracings/volume/zarr/$tracingId/$mag/$cxyz")
+    rpc(s"$tracingStoreUri/tracings/volume/zarr/$tracingId/$mag/$cxyz").silent
```
fm3, line 43: This slipped in here: avoid logging for every chunk request

fm3:
Discussion result:
(E) ditch the organizationName and datasetName fields entirely and perform the (cached) lookup each time by the tracing id (the wk-side postgres table should be able to connect tracing id and dataset via foreign keys)

jstriebel (approved):
LGTM (besides the two open TODOs) :+1:

PR #6553, opened 2022-10-14
[Skeletons] Allow deletion of root group

### LINKED ISSUE #5773: Allow to set color for all trees in a group
## Detailed Description
Right-clicking a group should show a "Set color for all trees" menu entry.

## Context
When having lots of single-node-trees which semantically belong to the same cluster, it would be helpful to give all of them the same color.

hotzenklotz:
There is no quick way to delete all skeleton groups and trees. Deleting the `Root` group would do that but has been disabled up until point. This PR changes that and allows the Root group and all it's children (aka everything) to be deleted.

- https://___.webknossos.xyz

### Steps to test:
- Open annotation with sub-groups and trees
- Right-click on `Root` group
- Select `Delete Group`
- Everything should be gone. The whole Skeleton tab should be empty. A Root node will be automatically re-created when placing a new node.

### Issues:
- fixes #5773

------
- [x] Ready for review


philippotto (approved):
works well :+1:

PR #6558, opened 2022-10-17
Highlight Menu in Navbar Menu

hotzenklotz:
During a discussion, we were left unsatisfied that the main navbar menu with many of the important dataset-wide actions items (Download, Sharing, Zarr Links) etc is very subtle / underrepresented:
- In "View" mode, the menu is easily overlooked since it is just a down-facing arrow (next to a super prominent, primary-color toggle switch for the left-hand side panel).
- In "Tracing Mode", the menu is also still tucked away next to the Save button and easily overlooked.

This PR proposed the following changes:
- Make the menu a separate dropdown button with an explicit title to give it the space it deserves in both *View* and *Tracing* mode
- Remove the cog icon, which on toggles the left-hand side bar. This is a legacy UI element and there are 2 alternative toggle buttons + a keyboard shortcut ("K") for that.

View Mode:
image

Tracing Mode:
image

- https://___.webknossos.xyz

### Steps to test:
- Open any dataset/annotation in both *view* and *tracing* modes

### Issues:
- No issue

------
- [x] Ready for review


fm3:
> @fm3 maybe you also have an idea/opinion?

I think “Menu” is acceptable as well, “Actions” or “More“ also works for me. I’m afraid I don’t really have another idea

hotzenklotz:
@normanrz Do you have a opinion on that? 

I am still in favor of "Menu" even so it is slightly generic. Yet, it is very much to the point, does not require any explanation and give users a clear idea of what might be behind the button.

normanrz:
I think "Menu" is good. "Actions" would also be fine. "More" would be missing a context imho.

normanrz (approved):

PR #6560, opened 2022-10-17
WIP: Added Help button

### LINKED ISSUE #6502: Help button
We want to make it as easy as possible for them to write us and ask questions. 

For this, we would like to have a button in the webKnossos interface for users to ask questions. When clicked, a form-like window should open that users can fill out. When sent, we should receive their answer by email. It should be possible to click it away. 

We want the button to focus on new users. It should only appear for the first three weeks after a user sign-up.

hotzenklotz:
The PR adds a small help button to the UI. Users can send us a question or give feedback. 

- https://___.webknossos.xyz

### Steps to test:
- Click on help button
- Write a something nice
- Click "Send"
- Check you emails if you were CCed on the help email 

### Issues:
- fixes #6502 

------
- [ ] Ready for review

### DIFF
--- a/frontend/javascripts/oxalis/view/help_modal.tsx
+++ b/frontend/javascripts/oxalis/view/help_modal.tsx
@@ -1,3 +1,4 @@
+import { sendHelpEmail } from "admin/admin_rest_api";
 import { Button, Modal, Input } from "antd";
 import React, { useState } from "react";
 
@@ -9,7 +10,7 @@ function HelpModal() {
     setModalOpen(false);
 
     if (helpText.length > 0) {
-      // API Call or Email
+      sendHelpEmail(helpText);
     }
   };
 
@@ -37,7 +38,7 @@ function HelpModal() {
           rows={6}
           value={helpText}
           onChange={(e: React.ChangeEvent<HTMLTextAreaElement>) => setHelpText(e.target.value)}
-        ></Input.TextArea>
+        />
       </Modal>
     </>
   );
--- a/app/controllers/Application.scala
+++ b/app/controllers/Application.scala
@@ -1,28 +1,39 @@
 package controllers
 
+import com.scalableminds.util.accesscontext.{GlobalAccessContext}
+import akka.actor.ActorSystem
 import com.mohiva.play.silhouette.api.Silhouette
 import com.scalableminds.util.tools.{Fox, FoxImplicits}
 import com.typesafe.config.ConfigRenderOptions
 import io.swagger.annotations.{Api, ApiOperation, ApiResponse, ApiResponses}
 import models.analytics.{AnalyticsService, FrontendAnalyticsEvent}
-import models.user.MultiUserDAO
+import models.user.{MultiUserDAO, UserService}
+import models.organization.OrganizationDAO
 import oxalis.security.WkEnv
 import play.api.libs.json.{JsObject, Json}
 import play.api.mvc.{Action, AnyContent, PlayBodyParsers}
 import slick.jdbc.PostgresProfile.api._
 import utils.{SQLClient, SimpleSQLDAO, StoreModules, WkConf}
+import oxalis.mail.{DefaultMails, Send}
 
 import javax.inject.Inject
 import scala.concurrent.ExecutionContext
 
 @Api
 class Application @Inject()(multiUserDAO: MultiUserDAO,
+                            actorSystem: ActorSystem,
                             analyticsService: AnalyticsService,
+                            userService: UserService,
                             releaseInformationDAO: ReleaseInformationDAO,
+                            organizationDAO: OrganizationDAO,
                             conf: WkConf,
+                            defaultMails: DefaultMails,
                             storeModules: StoreModules,
                             sil: Silhouette[WkEnv])(implicit ec: ExecutionContext, bodyParsers: PlayBodyParsers)
     extends Controller {
+
+  private lazy val Mailer =
+    actorSystem.actorSelection("/user/mailActor")
 
   @ApiOperation(value = "Information about the version of webKnossos")
   @ApiResponses(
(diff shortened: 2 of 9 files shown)


hotzenklotz:
I could not really decide on a color scheme. Here are the two options I tried:


1. default color scheme of antd's `<Alert type="info">`:
image

2. More prominent color scheme, similar to antd's primary `<Button>`:
image

hotzenklotz:
> Should the user maybe get a slightly different email along the lines of “we received the following feedback message and will get back to you soon”?

I don't think that it makes much of a difference. The email copy for the user serves primarily to give him some feedback that his request/question was entered into the system. It also nicely ties together the conversation, when we answer directly to that email.

> Should this always be active? Or just for the webknossos.org instance?

Yes, I think it should be available for all instances. It does not hurt if we get questions from the open-source instances or dedicated hosted ones. If we get to too many questions (doubtful) then we can always be more selective with this feature.

frontend/javascripts/oxalis/view/help_modal.tsx
```diff
@@ -0,0 +1,98 @@
...
+import { setActiveUserAction } from "oxalis/model/actions/user_actions";
+import { OxalisState } from "oxalis/store";
+import React, { CSSProperties, useState } from "react";
+import { useDispatch, useSelector } from "react-redux";
+
+function HelpButton() {
+  const [isModalOpen, setModalOpen] = useState(false);
+
+  const dispatch = useDispatch();
+  const activeUser = useSelector((state: OxalisState) => enforceActiveUser(state.activeUser));
```
philippotto, line 13: this crashes wk when the user is not logged in. why use `enforceActiveUser` when you are guarding against `!activeUser` later, anyway?

frontend/javascripts/navbar.tsx
```diff
@@ -345,10 +350,22 @@ function HelpSubMenu({
           Keyboard Shortcuts
         </a>
       </Menu.Item>
+      {isAuthenticated ? (
+        <>
+          <Menu.Item key="get_help" onClick={() => setIsHelpModalOpen(true)}>
+            Get help / Ask a question
```
philippotto, line 356:
```suggestion
            Get help / Ask a question
```

I'm not really fond of the `/`. "Ask a question" alone should cover it pretty well, no?

philippotto (changes requested):
this crashes when being logged out, so i couldn't test the actual UX yet :see_no_evil:

frontend/javascripts/oxalis/view/help_modal.tsx
hotzenklotz: Copy pasta. That's why 😊

frontend/javascripts/navbar.tsx
hotzenklotz: Fair enough. I shortened it to just "Ask a question"

frontend/javascripts/oxalis/view/help_modal.tsx
hotzenklotz: Fixed

hotzenklotz:
@philippotto The logged-out crash should be fixed. Thanks for spotting that.

frontend/javascripts/navbar.tsx
```diff
@@ -353,7 +353,7 @@ function HelpSubMenu({
       {isAuthenticated ? (
         <>
           <Menu.Item key="get_help" onClick={() => setIsHelpModalOpen(true)}>
-            Get help / Ask a question
+            Ask a question
```
philippotto, line 356: the rest of the menu items are in title case, so I'd adapt this here for consistency.

frontend/javascripts/oxalis/view/help_modal.tsx
```diff
@@ -0,0 +1,96 @@
...
+
+  const dispatch = useDispatch();
+  const activeUser = useSelector((state: OxalisState) => state.activeUser);
+
+  const discardButton = (e: React.MouseEvent<HTMLButtonElement, MouseEvent>) => {
+    // prevent the modal from also being shown
+    e.stopPropagation();
+    
+    if (!activeUser) return null;
+    if (activeUser && activeUser.novelUserExperienceInfos.hasDiscardedHelpButton) return null;
```
philippotto, line 19: should logged-out users see the help button and be able to send emails? we won't know their email to reply though, right? if we support this somehow, the close-button should be hidden, though, since it doesn't do anything.

hotzenklotz: Logged out users should not see the help button:
1. We don't have an email to respond to
2. there is not much you can do in logged out state anyway. you likely be browsing the landing page
  2.1 alternatively, you can be logged out for viewing a public dataset/annotation. That is a read-only interaction anyway and would not expect someone to ask questions

frontend/javascripts/navbar.tsx
hotzenklotz: Done

frontend/javascripts/oxalis/view/help_modal.tsx
```diff
@@ -0,0 +1,98 @@
...
+    if (!activeUser) return;
+
+    const [newUserSync] = updateNovelUserExperienceInfos(activeUser, {
+      hasDiscardedHelpButton: true,
+    });
+    dispatch(setActiveUserAction(newUserSync));
+  };
+
+  if (!activeUser) return null;
+  if (activeUser && activeUser.novelUserExperienceInfos.hasDiscardedHelpButton) return null;
```
philippotto, line 27:
```suggestion
  if (activeUser.novelUserExperienceInfos.hasDiscardedHelpButton) return null;
```

since you are already checking for `!activeUser`

philippotto (approved):
works for me :+1: let's see how often this is used. on a big screen, the help button is rather low-key, but that's better than being too offensive, I guess.

PR #6561, opened 2022-10-17
Fix crash related to WEBGL_debug_renderer_info

philippotto:
[Context](https://scm.slack.com/archives/C5AKLAV0B/p1665156066659499)

- https://___.webknossos.xyz

### Steps to test:
- hard to test well; inspect code thoroughly instead. the outer try-catch should make the heuristic pretty robust :)

------
- [X] Ready for review


daniel-wer (approved):
LGTM with one small suggestion :+1:

thopp-tudelft:
With this change webknossos no longer crashes when webgl.enable-debug-renderer-info is set to false in firefox. :+1: 
The try catch is never triggered, checking if the returned parameters are null was all that was needed.

PR #6565, opened 2022-10-18
Update immutability-helper package to v3.1.1 with TS support

### LINKED ISSUE #3055: Update immutability-helper to make use of TS typing

hotzenklotz:
As per @philippotto suggestion, I updated the immutability-helper package to v3.1.1. The updated package has been re-written in Typescript and hopefully should provide better type coverage.

@philippotto and I did a quick spot check in some reducers and were pleased that we were given proper type errors :-)

I ran the TS type coverage reporter (`yarn ts-coverage`) but that did not report any differences between `master` and this PR. 🤷 

### Steps to test:
- Introduce some errors in any reducer and be greeted with (helpful?) type errors

### Issues:
- fixes #3055

------
- [x] Ready for review

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -71,5 +71,6 @@ frontend/javascripts/test/snapshots/type-check/*
 **/screenshots/*.new.png
 .nyc_output
 coverage
+coverage-ts
 *.vscode
 .bsp
--- a/package.json
+++ b/package.json
@@ -138,7 +138,8 @@
     "postinstall": "cd tools/proxy && yarn install",
     "typecheck": "yarn tsc",
     "ts": "yarn tsc",
-    "ts-watch": "yarn tsc --watch"
+    "ts-watch": "yarn tsc --watch",
+    "ts-coverage": "typescript-coverage-report"
   },
   "lint-staged": {
     "*.ts": [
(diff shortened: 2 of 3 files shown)


philippotto (approved):
sweet :lollipop:

PR #6566, opened 2022-10-19
When downloading with skipVolumeData, keep volume tag

fm3:
- https://volumeskipdataincludetag.webknossos.xyz

### Steps to test:
- Download a volume annotation with and without the include-volume-data checkbox active
- The volume tag and its children (segments) should always be included
- in the skip-volume-data case, the `location` attribute should be omitted, and a comment added instead.

------
- [x] Adapted [wk-libs python client](https://github.com/scalableminds/webknossos-libs/tree/master/webknossos/webknossos/client) if relevant API parts change (@jstriebel do you know if something needs to be adapted to handle the missing location attribute?)
- [x] Ready for review


fm3:
Looks like this does break the download with skipVolumeData for libs versions 0.10.18 and 0.10.19. Do you think we should accept that and encourage users to upgrade? Or build a backwards compatible version of this? I’m hesitant to just make the location emptystring, maybe you have other ideas?

jstriebel (commented):
The code itself LGTM :+1: 

> Looks like this does break the download with skipVolumeData for libs versions 0.10.18 and 0.10.19. Do you think we should accept that and encourage users to upgrade? Or build a backwards compatible version of this? I’m hesitant to just make the location emptystring, maybe you have other ideas?

As a first measure we should release a wk-libs version that is compatible with the change, I prepared a PR for that (https://github.com/scalableminds/webknossos-libs/pull/814). Then I think it's ok to merge this, since it only affects new downloads, and `skip_volume_data` must be set, which was just introduced three weeks ago with version `0.10.18` [here](https://github.com/scalableminds/webknossos-libs/commit/7c152fb4b8616c18845ac35c2c9d196835d02910).

fm3:
Thanks for having a look and thanks for pushing the change in the python client!

Do you think we should wait for some days before deploying this, so that more people would switch from 0.10.18 to 0.10.20? Or maybe it’s a rare enough case that it doesn’t make sense to wait?

jstriebel:
> Do you think we should wait for some days before deploying this, so that more people would switch from 0.10.18 to 0.10.20? Or maybe it’s a rare enough case that it doesn’t make sense to wait?

Let's just wait a couple of days, as this change is not pressing IMO. But in general I think that it should be quite rare, still.

jstriebel (approved):
Enough waiting IMO :rocket:

PR #6569, opened 2022-10-19
Load Oversegmentation Meshes during Proofreading

fm3:
Second try with cherry-picked commits from #6517

- https://___.webknossos.xyz

### Steps to test:
- abc

### Issues:
- fixes #

------
- [ ] Ready for review


fm3:
@philippotto Do you know if the front-end _knows_ when it is requesting these precomputed meshes with an _editable_ mapping? Could it send this info as a boolean paramter? Or does the datastore need to figure that out on its own?

philippotto:
> Do you know if the front-end knows when it is requesting these precomputed meshes with an editable mapping? Could it send this info as a boolean paramter?

Yes, this should be possible for the front-end.

fm3:
>> Do you know if the front-end knows when it is requesting these precomputed meshes with an editable mapping? Could it send this info as a boolean paramter?

> Yes, this should be possible for the front-end.

That’s good news :) I added the optional boolean param targetMappingIsEditable. Fetching the actual mapping from the tracingstore is not implemented yet, but if you have time, you could already adapt the frontend to use the parameter accordingly.

fm3:
@philippotto ok so here’s what I’ve come up with:
The request should include
 - `editableMappingTracingId: Option[String]` (the tracing id, not the editable mapping id). If this is set, it is assumed that the request is about an editable mapping
 - `targetMappingName: Option[String]` (as before) the on-disk mapping name. In case of an editable mapping, this should _still_ be the on-disk base mapping name (so that agglomerates that are untouched by the editable mapping can be looked up there without another round-trip between tracingstore and datastore)

Could you adapt the frontend like that?

philippotto:
> @philippotto ok so here’s what I’ve come up with: The request should include
> 
> * `editableMappingTracingId: Option[String]` (the tracing id, not the editable mapping id). If this is set, it is assumed that the request is about an editable mapping
> * `targetMappingName: Option[String]` (as before) the on-disk mapping name. In case of an editable mapping, this should _still_ be the on-disk base mapping name (so that agglomerates that are untouched by the editable mapping can be looked up there without another round-trip between tracingstore and datastore)
> 
> Could you adapt the frontend like that?

I just adapted the call for the mesh v3 route. Should the v0 route be changed, too?

I didn

fm3:
> I just adapted the call for the mesh v3 route. Should the v0 route be changed, too?

For the record, we decided that adapting the v0 route is not needed.


I’d say the backend is ready for review now :)

frontend/javascripts/oxalis/model/sagas/isosurface_saga.ts
```diff
@@ -617,6 +623,16 @@ function* loadPrecomputedMeshForSegmentId(
 
   const version = meshFile.formatVersion;
   try {
+    const isosurfaceExtraInfo = yield* call(getIsosurfaceExtraInfo, segmentationLayer.name, null);
+    const mappingName = meshFile.mappingName == null ? isosurfaceExtraInfo.mappingName : null;
```
daniel-wer, line 627: I would have expected this to always be `isosurfaceExtraInfo.mappingName`. If the active mesh file was computed for a mapping, but the mapping is not active, the segments in the viewports won't correspond to the segments in the mesh file. That's why before loading a precomputed mesh, the user is prompted to activate the corresponding mapping (if there is one). Does that make sense?

frontend/javascripts/oxalis/model/sagas/proofread_saga.ts
```diff
@@ -86,20 +89,35 @@ function proofreadUsingMeshes(): boolean {
...
-      mappingType,
-      preferredQuality,
-    }),
-  );
+  if (
+    currentMeshFile != null &&
+    currentMeshFile.formatVersion >= 3 &&
+    currentMeshFile.mappingName == null
+  ) {
+    // If a mesh file exists which was computed without a mapping, use that instead of computing
```
daniel-wer, line 101:
```suggestion
    // If a mesh file is active which was computed without a mapping, use that instead of computing
```

daniel-wer (commented):
This is pretty cool! The performance impact is already noticeable even on the small toy dataset. Will be interesting to test this on a large dataset :)

If we ever want to use the oversegmentation meshes to support the proofreading flow itself (by highlighting them separately in the 3D viewport, for example), we'll need to rethink this approach, because for the frontend, the separate meshes appear as one, right?

webknossos-datastore/app/com/scalableminds/webknossos/datastore/controllers/MeshController.scala
```diff
@@ -66,12 +77,81 @@ class MeshController @Inject()(
                 "mesh.file.listChunks.failed",
                 request.body.segmentId.toString,
                 request.body.meshFile) ?~> Messages("mesh.file.load.failed", request.body.segmentId.toString) ~> BAD_REQUEST
```
jstriebel, line 79: Is this still necessary?

webknossos-datastore/app/com/scalableminds/webknossos/datastore/controllers/MeshController.scala
```diff
@@ -55,7 +59,14 @@ class MeshController @Inject()(
                                          organizationName: String,
                                          dataSetName: String,
                                          dataLayerName: String,
-                                         formatVersion: Int): Action[ListMeshChunksRequest] =
+                                         formatVersion: Int,
+                                         /* If targetMappingName is set, assume that meshfile contains meshes for
+                                           the oversegmentation. Collect mesh chunks of all *unmapped* segment ids
+                                           belonging to the supplied agglomerate id.
+                                           If it is not set, use meshfile as is, assume passed id is present in meshfile
+                                          */
```
jstriebel, line 67: How does one get the meshes of mapped segment ids, without using the oversegmentation? Or is this handled somewhere else?

webknossos-datastore/app/com/scalableminds/webknossos/datastore/services/MeshFileService.scala
```diff
@@ -147,7 +145,20 @@ case class MeshSegmentInfo(chunkShape: Vec3Float, gridOrigin: Vec3Float, lods: L
 object MeshSegmentInfo {
   implicit val jsonFormat: OFormat[MeshSegmentInfo] = Json.format[MeshSegmentInfo]
 }
-case class WebknossosSegmentInfo(transform: Array[Array[Double]], meshFormat: String, chunks: MeshSegmentInfo)
+case class WebknossosSegmentInfo(transform: Array[Array[Double]], meshFormat: String, chunks: MeshSegmentInfo) {
+  def merge(that: WebknossosSegmentInfo): WebknossosSegmentInfo =
+    // assume that transform, meshFormat, chunkShape, gridOrigin, scale and vertexOffset are the same
```
jstriebel, line 150: It looks like two things are named chunks here, the `MeshSegmentInfo` object and also the `lod.chunks` items. Can we rename the outer one?

Only the (inner)chunks are merged here, right? So maybe calling the method `merge_chunks` makes the assumption more clear when being called as well.

jstriebel (commented):
The backend mostly LGTM, just added some questions in the comments:

philippotto:
> If we ever want to use the oversegmentation meshes to support the proofreading flow itself (by highlighting them separately in the 3D viewport, for example), we'll need to rethink this approach, because for the frontend, the separate meshes appear as one, right?

I think, the isosurface saga could simply use the mesh file and request the meshes for the supervoxels like we did it in an earlier version (but without using ad-hoc meshing). It should be fairly simple, but since we don't use that behavior currently, I left it as is.

> P.S. I added the oversegmentation mesh file to the test-agglomerate-file dataset on humerus, so that it is always available on the dev instances.

Awesome, thank you! I didn't think about that.

frontend/javascripts/oxalis/model/sagas/isosurface_saga.ts
philippotto: Good point, this was a bit fuzzy here. I assumed that the mappings would match, but now I added a proper assertion to make sure that the caller didn't request something weird. I also tried to simplify the logic a bit. Please have a look at my latest commit :)

webknossos-datastore/app/com/scalableminds/webknossos/datastore/services/MeshFileService.scala
fm3: These data structures are a bit hard to name as they kind of live between the wk terminology and neuroglancer meshfile manifest terminology. I renamed this to segmentInfo here now, but to be honest, I’m not 100% certain leo’s reasoning for naming choices here :sweat_smile:

webknossos-datastore/app/com/scalableminds/webknossos/datastore/controllers/MeshController.scala
fm3: Well spotted, this was dead code, I removed it now

webknossos-datastore/app/com/scalableminds/webknossos/datastore/controllers/MeshController.scala
fm3: If the selected meshfile is computed with a mapping, the passed id is assumed to be an agglomerate id. There is no difference here for this case. (or rather, it is expressed solely by which meshfile the frontend selects)

jstriebel: Makes sense, thanks :+1:

webknossos-datastore/app/com/scalableminds/webknossos/datastore/services/MeshFileService.scala
jstriebel: Fair enough. What do you think about renaming the funtion to `merge_chunks`?

jstriebel: Yep, fine for me :+1:

jstriebel (commented):
Backend LGTM, leaving the final approval for @daniel-wer :rocket:

daniel-wer (approved):
Frontend LGTM as well :+1:

PR #6572, opened 2022-10-20
adds buttons for switching organizations for workflows

normanrz:
### Steps to test:
- Make a second organization
- Run a vx workflow
- Open workflow report in wk
- Switch to other organization
- See the "switch to org" screen and click the button
- Voilà

------
- [x] Ready for review


frontend/javascripts/admin/voxelytics/workflow_view.tsx
```diff
@@ -339,11 +350,21 @@ export default function WorkflowView() {
...
+          user != null ? await isWorkflowAccessibleBySwitching(workflowName) : null;
+        setLoadingState({
+          status: "FAILED",
+          organizationToSwitchTo: organization ?? undefined,
+          error: err as Error,
+        });
+      } catch (err1) {
+        console.log(err1);
+        Toast.error("Could not load workflow report.");
+        setLoadingState({ status: "FAILED", error: err as Error });
```
philippotto, line 366:
```suggestion
      } catch (accessibleBySwitchingError) {
        console.log(accessibleBySwitchingError);
        Toast.error("Could not load workflow report.");
        setLoadingState({ status: "FAILED", error: accessibleBySwitchingError as Error });
```

frontend/javascripts/admin/voxelytics/workflow_view.tsx
```diff
@@ -373,8 +394,17 @@ export default function WorkflowView() {
     report == null || report.run.state === VoxelyticsRunState.RUNNING ? VX_POLLING_INTERVAL : null,
   );
 
+  if (loadingState.status === "FAILED" && user != null) {
```
philippotto, line 397: what happens if status is failed and user == null? this should probably show a log in form, right?

normanrz: the router shows a login form, if the user is not logged in. so effectively, `user == null` is never true for the scope of this component

frontend/javascripts/admin/voxelytics/workflow_view.tsx
philippotto: note that this is not only renaming vars but also changes which err object is reported.

philippotto (approved):
:+1:

app/models/voxelytics/VoxelyticsDAO.scala
```diff
@@ -79,7 +79,16 @@ class VoxelyticsDAO @Inject()(sqlClient: SQLClient)(implicit ec: ExecutionContex
           WHERE hash = $workflowHash AND _organization = $organizationId
           """.as[(String, String)])
       (name, hash) <- r.headOption
-    } yield WorkflowEntry(name, hash)
+    } yield WorkflowEntry(name, hash, organizationId)
+
+  def findWorkflowOrganizationsByHash(workflowHash: String): Fox[List[WorkflowEntry]] =
```
fm3, line 84: As discussed in person, I’d rename this function to findWorkflowByHash (and return just the head option), and rename the function above to findWorkflowByHashAndOrganization

fm3 (approved):

PR #6574, opened 2022-10-21
Fix suggested wklibs snippet

philippotto:
The snippet was missing to commas which caused syntax errors when executing it verbatim. Also, the parameter names weren't correct anymore.

------
- [X] Ready for review


jstriebel (approved):
Thanks, LGTM :+1:

PR #6588, opened 2022-10-25
Fix precomputed mesh positions

daniel-wer:
by scaling the whole group instead of each chunk on its own. Fixing a regression introduced in https://github.com/scalableminds/webknossos/pull/6552.  See https://scm.slack.com/archives/C5AKLAV0B/p1666683368029709 for context.

- https://___.webknossos.xyz

### Steps to test:
- Use a v3 meshfile and load a precomputed mesh consisting of multiple chunks. The meshes should appear at the correct position.

------
- [x] Ready for review


philippotto (approved):
LGTM, but didn't test. Can't do much (more) harm, though, I think :)

PR #6589, opened 2022-10-25
Ensure that node positions are maintained as integers in front-end

### LINKED ISSUE #5800: Rethink integer-casting of node positions
## Context
Skeleton tracings. Also see (and post update there): https://discuss.webknossos.org/t/internal-node-positions-with-sub-voxel-resolution/1741

## Expected Behavior
Floating values for node positions should be correctly stored /edit: or the positions should be immediately rounded after creation/movement of a node.

## Current Behavior
Floating values for node positions are cast to integers.

## Steps to Reproduce the bug

1. create nodes which are very close to each other in the YZ viewport (zoom in quite a lot)
2. ensure the nodes are in the same z slices, but not on top of each other
3. save & reload --> now all nodes are on the same z
4.

## Your Environment for bug

philippotto:
Node positions are already stored as ints by the backend. This change ensures that the nodes are also ints in the front-end. The only exception is when currently moving a node as the this would lead to a weird UX, otherwise.

- https://intnodeposition.webknossos.xyz

### Steps to test:
- create and drag nodes
- disable interpolation
- zoom in and check that the nodes are at pixel corners

### Issues:
- fixes #5800 

------
- [X] Ready for review

### DIFF
--- a/frontend/javascripts/oxalis/controller/combinations/skeleton_handlers.ts
+++ b/frontend/javascripts/oxalis/controller/combinations/skeleton_handlers.ts
@@ -177,6 +177,17 @@ export function moveNode(dx: number, dy: number, nodeId?: number | null | undefi
     }),
   );
 }
+
+export function finishNodeMovement(nodeId: number) {
+  getSkeletonTracing(Store.getState().tracing).map((skeletonTracing) =>
+    getNodeAndTree(skeletonTracing, nodeId).map(([activeTree, node]) => {
+      Store.dispatch(
+        setNodePositionAction(V3.round(node.position, [0, 0, 0]), node.id, activeTree.treeId),
+      );
+    }),
+  );
+}
+
 export function setWaypoint(
   position: Vector3,
   activeViewport: OrthoView,
--- a/frontend/javascripts/oxalis/controller/combinations/skeleton_handlers.ts
+++ b/frontend/javascripts/oxalis/controller/combinations/skeleton_handlers.ts
@@ -152,7 +152,16 @@ export function handleOpenContextMenu(
     activeViewport,
   );
 }
-export function moveNode(dx: number, dy: number, nodeId?: number | null | undefined) {
+export function moveNode(
+  dx: number,
+  dy: number,
+  nodeId?: number | null | undefined,
+  // If useFloat is false, the node will be moved by integer deltas.
+  // If useFloat is true, the caller should ensure that the node position
+  // is changed to integers afterwards (e.g., when dragging a node where
+  // the floats are used temporary for a smooth UX).
+  useFloat: boolean = false,
+) {
   // dx and dy are measured in pixel.
   getSkeletonTracing(Store.getState().tracing).map((skeletonTracing) =>
     getNodeAndTree(skeletonTracing, nodeId).map(([activeTree, activeNode]) => {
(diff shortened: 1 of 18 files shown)


frontend/javascripts/oxalis/controller/combinations/skeleton_handlers.ts
```diff
@@ -177,6 +177,17 @@ export function moveNode(dx: number, dy: number, nodeId?: number | null | undefi
     }),
   );
 }
+
+export function finishNodeMovement(nodeId: number) {
+  getSkeletonTracing(Store.getState().tracing).map((skeletonTracing) =>
+    getNodeAndTree(skeletonTracing, nodeId).map(([activeTree, node]) => {
+      Store.dispatch(
+        setNodePositionAction(V3.round(node.position, [0, 0, 0]), node.id, activeTree.treeId),
```
philippotto, line 185: I think, `round` makes most sense here. Note that the second parameter is the output variable and is necessary, because the mjs module would create a typed array otherwise (which immediately crashed during the freeze operation of the reducer). Probably, the TypedArray default of the mjs could use some re-thinking, too..

frontend/javascripts/oxalis/model/reducers/skeletontracing_reducer_helpers.ts
```diff
@@ -143,6 +144,7 @@ export function createNode(
     .getOrElse(Constants.DEFAULT_NODE_RADIUS);
   // Find new node id by increasing the max node id.
   const nextNewId = skeletonTracing.cachedMaxNodeId + 1;
+  const position = V3.floor(positionFloat);
```
philippotto, line 147: This could be a `round` operation, too, but when creating a node at z=1.6, this would change the behavior. Until now, that node would be stored at z=1. Rounding would store it at z=2. Depending on the clipping distance, this could be extra weird. For moving a node, this issue isn't relevant, I think, because the movement is 2D.

frontend/javascripts/oxalis/model/reducers/skeletontracing_reducer.ts
```diff
@@ -676,6 +676,8 @@ function SkeletonTracingReducer(state: OxalisState, action: Action): OxalisState
                 node.id,
                 update(node, {
                   position: {
+                    // Don't round here, since this would make the continous
```
daniel-wer, line 679:
```suggestion
                    // Don't round here, since this would make the continuous
```

daniel-wer (commented):
Nice :+1: 

Three more actions that come to mind that could potentially provoke sub-voxel node positions (the first confirmed, the other two untested):
- [x] moving nodes via `ctrl + arrow-keys`
- [ ] nml import
- [ ] agglomerate skeleton loading

Also, I noticed during testing that clicking multiple times in a small area easily creates nodes on top of each other. However, I can't think of a nice method to indicate that to the user. Probably ok for now.

philippotto:
> Three more actions that come to mind that could potentially provoke sub-voxel node positions (the first confirmed, the other two untested):
> 
> * [x]  moving nodes via `ctrl + arrow-keys`
> * [x]  nml import
> * [x]  agglomerate skeleton loading

Good points :+1: I did test the NML import initially and found it to produce only integers, but I think I used the back-end export for creating the NML. Now, the code guards against floats during import and export.
Re agglomerate skeleton loading, I tested it and the back-end returns integers :tada:

> Also, I noticed during testing that clicking multiple times in a small area easily creates nodes on top of each other. However, I can't think of a nice method to indicate that to the user. Probably ok for now.

While testing this, I noticed that the casting felt a bit weird (since the code only relied on Math.trunc). Now, I added a Math.round in the getMousePosition code (so that it only affects the UV plane so that we don't have trouble with the third dimension). Now, it feels nicer. The problem you observed still exists, but I think we can't do much (and I also think that it's acceptable).

Feel free to have another look :)

daniel-wer (approved):
Nice, LGTM! :+1:

PR #6591, opened 2022-10-25
Virtual Folder Structure for Datasets

fm3:
- https://___.webknossos.xyz

### Steps to test:
- abc

### Issues:
- fixes #

------
- [ ] Ready for review

### DIFF
--- a/app/controllers/InitialDataController.scala
+++ b/app/controllers/InitialDataController.scala
@@ -6,6 +6,7 @@ import com.scalableminds.util.tools.{Fox, FoxImplicits}
 import com.typesafe.scalalogging.LazyLogging
 import models.annotation.{TracingStore, TracingStoreDAO}
 import models.binary._
+import models.folder.FolderDAO
 import models.project.{Project, ProjectDAO}
 import models.task.{TaskType, TaskTypeDAO}
 import models.team._
@@ -15,6 +16,7 @@ import org.joda.time.DateTime
 import oxalis.security._
 import play.api.libs.json.Json
 import utils.{ObjectId, StoreModules, WkConf}
+
 import javax.inject.Inject
 import models.organization.{Organization, OrganizationDAO}
 import play.api.mvc.{Action, AnyContent}
@@ -40,6 +42,7 @@ class InitialDataService @Inject()(userService: UserService,
                                    userExperiencesDAO: UserExperiencesDAO,
                                    taskTypeDAO: TaskTypeDAO,
                                    dataStoreDAO: DataStoreDAO,
+                                   folderDAO: FolderDAO,
                                    tracingStoreDAO: TracingStoreDAO,
                                    teamDAO: TeamDAO,
                                    tokenDAO: TokenDAO,
@@ -107,6 +110,7 @@ Samplecountry
     for {
       _ <- updateLocalDataStorePublicUri()
       _ <- updateLocalTracingStorePublicUri()
+      _ <- insertRootFolder()
       _ <- insertLocalDataStoreIfEnabled()
       _ <- insertLocalTracingStoreIfEnabled()
       _ <- assertInitialDataEnabled
@@ -130,6 +134,12 @@ Samplecountry
       organizations <- organizationDAO.findAll
       _ <- bool2Fox(organizations.isEmpty) ?~> "initialData.organizationsNotEmpty"
     } yield ()
+
+  private def insertRootFolder(): Fox[Unit] =
+    folderDAO.getRoot.futureBox.flatMap {
+      case Full(_) => Fox.successful(())
+      case _ => folderDAO.insertRoot()
+    }
 
   private def insertDefaultUser(): Fox[Unit] =
     userService
--- a/app/controllers/InitialDataController.scala
+++ b/app/controllers/InitialDataController.scala
@@ -138,7 +138,7 @@ Samplecountry
   private def insertRootFolder(): Fox[Unit] =
     folderDAO.getRoot.futureBox.flatMap {
       case Full(_) => Fox.successful(())
-      case _ => folderDAO.insertRoot()
+      case _       => folderDAO.insertRoot()
     }
 
   private def insertDefaultUser(): Fox[Unit] =
--- a/app/controllers/InitialDataController.scala
+++ b/app/controllers/InitialDataController.scala
@@ -6,7 +6,7 @@ import com.scalableminds.util.tools.{Fox, FoxImplicits}
 import com.typesafe.scalalogging.LazyLogging
 import models.annotation.{TracingStore, TracingStoreDAO}
 import models.binary._
-import models.folder.FolderDAO
+import models.folder.{Folder, FolderDAO}
 import models.project.{Project, ProjectDAO}
 import models.task.{TaskType, TaskTypeDAO}
 import models.team._
@@ -71,7 +71,8 @@ Samplecountry
                  additionalInformation,
                  "/assets/images/oxalis.svg",
                  "Sample Organization",
-                 PricingPlan.Custom)
+                 PricingPlan.Custom,
+                 ObjectId.generate)
   private val organizationTeam =
     Team(organizationTeamId, defaultOrganization._id, defaultOrganization.name, isOrganizationTeam = true)
   private val userId = ObjectId.generate
@@ -136,9 +137,9 @@ Samplecountry
     } yield ()
 
   private def insertRootFolder(): Fox[Unit] =
-    folderDAO.getRoot.futureBox.flatMap {
+    folderDAO.findOne(defaultOrganization._rootFolder).futureBox.flatMap {
       case Full(_) => Fox.successful(())
-      case _       => folderDAO.insertRoot()
+      case _       => folderDAO.insertAsRoot(Folder(defaultOrganization._rootFolder, defaultOrganization.name))
     }
 
   private def insertDefaultUser(): Fox[Unit] =
--- a/app/models/folder/Folder.scala
+++ b/app/models/folder/Folder.scala
@@ -22,7 +22,7 @@ class FolderDAO @Inject()(sqlClient: SQLClient)(implicit ec: ExecutionContext)
   def parse(r: FoldersRow): Fox[Folder] =
     Fox.successful(Folder(ObjectId(r._Id), r.name))
 
-  def insertRoot(f: Folder): Fox[Unit] = {
+  def insertAsRoot(f: Folder): Fox[Unit] = {
     val insertPathQuery =
       sqlu"INSERT INTO webknossos.folder_paths(_ancestor, _descendant, depth) VALUES(${f._id}, ${f._id}, 0)"
     for {
@@ -42,7 +42,7 @@ class FolderDAO @Inject()(sqlClient: SQLClient)(implicit ec: ExecutionContext)
       parsed <- parseAll(rows)
     } yield parsed
 
-  def insertChild(parentId: ObjectId, f: Folder): Fox[Unit] = {
+  def insertAsChild(parentId: ObjectId, f: Folder): Fox[Unit] = {
     val insertPathQuery =
       sqlu"""INSERT INTO webknossos.folder_paths(_ancestor, descendant, depth)
              SELECT ancestor, ${f._id}, depth + 1 from webknossos.folder_paths WHERE descendant = $parentId -- links to ancestors
--- a/app/models/folder/Folder.scala
+++ b/app/models/folder/Folder.scala
@@ -22,8 +22,7 @@ class FolderDAO @Inject()(sqlClient: SQLClient)(implicit ec: ExecutionContext)
   def parse(r: FoldersRow): Fox[Folder] =
     Fox.successful(Folder(ObjectId(r._Id), r.name))
 
-  def insertRoot(): Fox[Unit] = {
-    val f = Folder(ObjectId.generate, "root")
+  def insertRoot(f: Folder): Fox[Unit] = {
     val insertPathQuery =
       sqlu"INSERT INTO webknossos.folder_paths(_ancestor, _descendant, depth) VALUES(${f._id}, ${f._id}, 0)"
     for {
@@ -31,12 +30,10 @@ class FolderDAO @Inject()(sqlClient: SQLClient)(implicit ec: ExecutionContext)
     } yield ()
   }
 
-  def getRoot: Fox[Folder] =
+  def findOne(folderId: ObjectId): Fox[Folder] =
     for {
-      row <- run(sql"""
-             SELECT #$columns FROM webknossos.folders
-        WHERE NOT EXISTS ( SELECT _descendant FROM webknossos.folder_paths WHERE depth > 0)""".as[FoldersRow])
-      parsed <- parseFirst(row, "root")
+      rows <- run(sql"SELECT #$columns FROM webknossos.folders WHERE _id = $folderId".as[FoldersRow])
+      parsed <- parseFirst(rows, "id")
     } yield parsed
 
   def findChildren(folderId: ObjectId): Fox[List[Folder]] =
@@ -64,4 +61,10 @@ class FolderDAO @Inject()(sqlClient: SQLClient)(implicit ec: ExecutionContext)
       _ <- run(sqlu"update webknossos.folders set name = ${f.name} where _id = ${f._id}")
     } yield ()
 
+// TODO  mkdir -p route
+
+  // set permissions for folder
+  // load full tree (sans datasets)
+  // enforce unique names per layer
+  // Root per organization!
 }
--- a/app/models/folder/Folder.scala
+++ b/app/models/folder/Folder.scala
@@ -2,6 +2,7 @@ package models.folder
 
 import com.scalableminds.util.tools.Fox
 import com.scalableminds.webknossos.schema.Tables.{Folders, _}
+import play.api.libs.json.{JsObject, Json}
 import slick.jdbc.PostgresProfile.api._
 import slick.lifted.Rep
 import slick.sql.SqlAction
@@ -11,6 +12,11 @@ import javax.inject.Inject
 import scala.concurrent.ExecutionContext
 
 case class Folder(_id: ObjectId, name: String)
+
+class FolderService @Inject()()(implicit ec: ExecutionContext) {
+  def publicWrites(folder: Folder): Fox[JsObject] =
+    Fox.successful(Json.obj("id" -> folder._id, "name" -> folder.name))
+}
 
 class FolderDAO @Inject()(sqlClient: SQLClient)(implicit ec: ExecutionContext)
     extends SQLDAO[Folder, FoldersRow, Folders](sqlClient) {
--- a/app/models/folder/Folder.scala
+++ b/app/models/folder/Folder.scala
@@ -1,0 +1,44 @@
+package models.folder
+
+import com.scalableminds.util.tools.Fox
+import com.scalableminds.webknossos.schema.Tables.{Folders, _}
+import slick.jdbc.PostgresProfile.api._
+import slick.lifted.Rep
+import utils.{ObjectId, SQLClient, SQLDAO}
+
+import javax.inject.Inject
+import scala.concurrent.ExecutionContext
+
+case class Folder(_id: ObjectId, name: String)
+
+class FolderDAO @Inject()(sqlClient: SQLClient)(implicit ec: ExecutionContext)
+  extends SQLDAO[Folder, FoldersRow, Folders](sqlClient) {
+
+  val collection = Folders
+  def idColumn(x: Folders): Rep[String] = x._Id
+  def isDeletedColumn(x: Folders): Rep[Boolean] = x.isdeleted
+
+  def parse(r: FoldersRow): Fox[Folder] =
+    Fox.successful(Folder(ObjectId(r._Id), r.name))
+
+  def insertRoot(): Fox[Unit] = {
+    val id = ObjectId.generate
+    val name = "root"
+    val insertFolderQuery = sqlu"insert into webknossos.folders(_id, name) values($id, $name)"
+    val insertPathQuery = sqlu"insert into webknossos.folder_paths(_ancestor, _descendant, depth) values($id, $id, 0)"
+    for {
+      _ <- run(DBIO.sequence(List(insertFolderQuery, insertPathQuery)).transactionally)
+    } yield ()
+  }
+
+  def getRoot: Fox[Folder] = {
+    for {
+      row <-
+        run(sql"""
+             SELECT _id, name FROM webknossos.folders
+             WHERE NOT EXISTS ( SELECT TOP 1 1 descendant FROM webknossos.folder_paths WHERE depth > 0)""".as[FoldersRow])
+      parsed <- parseFirst(row, "root")
+    } yield parsed
+  }
+
+}
--- a/app/models/folder/Folder.scala
+++ b/app/models/folder/Folder.scala
@@ -4,6 +4,7 @@ import com.scalableminds.util.tools.Fox
 import com.scalableminds.webknossos.schema.Tables.{Folders, _}
 import slick.jdbc.PostgresProfile.api._
 import slick.lifted.Rep
+import slick.sql.SqlAction
 import utils.{ObjectId, SQLClient, SQLDAO}
 
 import javax.inject.Inject
@@ -12,7 +13,7 @@ import scala.concurrent.ExecutionContext
 case class Folder(_id: ObjectId, name: String)
 
 class FolderDAO @Inject()(sqlClient: SQLClient)(implicit ec: ExecutionContext)
-  extends SQLDAO[Folder, FoldersRow, Folders](sqlClient) {
+    extends SQLDAO[Folder, FoldersRow, Folders](sqlClient) {
 
   val collection = Folders
   def idColumn(x: Folders): Rep[String] = x._Id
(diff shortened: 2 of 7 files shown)


tools/postgres/schema.sql
```diff
@@ -99,6 +99,7 @@ CREATE TABLE webknossos.dataSets(
   _organization CHAR(24) NOT NULL,
   _publication CHAR(24),
   _uploader CHAR(24),
+  _folder CHAR(24) NOT NULL,
```
fm3, line 102: note to self: add index

app/models/organization/OrganizationService.scala
```diff
@@ -77,15 +80,18 @@ class OrganizationService @Inject()(organizationDAO: OrganizationDAO,
...
                                   "",
                                   "",
                                   organizationDisplayName,
-                                  initialPricingPlan)
+                                  initialPricingPlan,
+                                  organizationRootFolder._id)
       organizationTeam = Team(ObjectId.generate, organization._id, "Default", isOrganizationTeam = true)
       _ <- organizationDAO.insertOne(organization)
       _ <- teamDAO.insertOne(organizationTeam)
+      _ <- folderDAO.insertAsRoot(organizationRootFolder)
```
fm3, line 94: double-check that this works, what about the foreign key constraint? changing the order might help

app/controllers/InitialDataController.scala
```diff
@@ -107,6 +111,7 @@ Samplecountry
     for {
       _ <- updateLocalDataStorePublicUri()
       _ <- updateLocalTracingStorePublicUri()
+      _ <- insertRootFolder()
```
fm3, line 114: move below assertNoOrganizationsPresent?

app/controllers/DataSetController.scala
```diff
@@ -56,7 +56,8 @@ class DataSetController @Inject()(userService: UserService,
       (__ \ 'displayName).readNullable[String] and
       (__ \ 'sortingKey).readNullable[Long] and
       (__ \ 'isPublic).read[Boolean] and
-      (__ \ 'tags).read[List[String]]).tupled
+      (__ \ 'tags).read[List[String]] and
+      (__ \ 'folderId).read[ObjectId]).tupled
```
fm3, line 60: make nullable

fm3: double check that the frontend does not send very old values, loaded long before the move

app/models/binary/DataSet.scala
```diff
@@ -135,13 +163,36 @@ class DataSetDAO @Inject()(sqlClient: SQLClient,
...
       parsed <- parseAll(r)
     } yield parsed
 
+  private def buildSearchPredicate(searchQueryOpt: Option[String]): String =
+    searchQueryOpt match {
+      case None => "true"
+      case Some(searchQuery) =>
+        val searchQueryFiltered = searchQuery.replaceAll("[^A-Za-z0-9_\\-. ]", "").toLowerCase()
+        val segments = searchQueryFiltered.trim.split(" +")
+        segments.map(queryToken => s" LOWER(name) LIKE '%$queryToken%' ").mkString(" AND ")
```
fm3, line 187: underscore needs to be escaped in LIKE syntax? let’s dive deeper here

app/models/binary/DataSet.scala
```diff
@@ -112,20 +114,46 @@ class DataSetDAO @Inject()(sqlClient: SQLClient,
...
+          )
+        )
+        OR ( -- user is in a team that is allowed for the folder or its ancestors
+          _folder IN (
+            SELECT fp._descendant
+            FROM webknossos.folder_paths fp
+            WHERE fp._ancestor IN (
+              SELECT at._folder
+              FROM webknossos.folder_allowedTeams at
+              JOIN webknossos.user_team_roles tr ON at._team = tr._team
```
fm3, line 151: name user_team_roles ut here, as above

app/models/folder/Folder.scala
```diff
@@ -0,0 +1,307 @@
...
+          -- is descendant of a folder with allowed teams the user is in
+          ${prefix}_id IN (
+            SELECT fp._descendant
+            FROM webknossos.folder_paths fp
+            WHERE fp._ancestor IN (
+              SELECT at._folder
+              FROM webknossos.folder_allowedTeams at
+              JOIN webknossos.user_team_roles tr ON at._team = tr._team
+              WHERE tr._user = '$requestingUserId'
+              $writeAccessPredicate
```
fm3, line 155:
```suggestion
              AND $writeAccessPredicate
```
(change above to or("true"))

app/models/folder/Folder.scala
```diff
@@ -0,0 +1,307 @@
...
+      rows <- run(sql"""SELECT f._id, f.name, fp._ancestor
+              FROM webknossos.folders_ f
+              JOIN webknossos.folder_paths fp -- join to find immediate parent, this will also kick out self
+              ON f._id = fp._descendant
+              WHERE fp.depth = 1 AND f._id IN
+              (SELECT _descendant  -- find all folder ids that are descendants
+              FROM webknossos.folder_paths
+              WHERE _ancestor = $folderId)
+              AND #$accessQueryWithPrefix
+              UNION ALL SELECT _id, name, NULL from webknossos.folders_  -- find self again, with no parent
```
fm3, line 246:
```suggestion
              UNION ALL SELECT _id, name, NULL -- find self again, with no parent
              FROM webknossos.folders_  
```

app/models/folder/Folder.scala
```diff
@@ -0,0 +1,307 @@
...
+              WHERE _id = $folderId
+              AND #$accessQuery
+              """.as[(String, String, Option[String])])
+      parsed <- Fox.combined(rows.toList.map(parseWithParent))
+    } yield parsed
+
+  def insertAsChild(parentId: ObjectId, f: Folder)(implicit ctx: DBAccessContext): Fox[Unit] = {
+    val insertPathQuery =
+      sqlu"""INSERT INTO webknossos.folder_paths(_ancestor, _descendant, depth)
+             SELECT _ancestor, ${f._id}, depth + 1 from webknossos.folder_paths WHERE _descendant = $parentId -- links to ancestors
```
fm3, line 256:
```suggestion
             SELECT _ancestor, ${f._id}, depth + 1 -- links to ancestors
             FROM webknossos.folder_paths
             WHERE _descendant = $parentId 
```

app/models/folder/Folder.scala
```diff
@@ -0,0 +1,307 @@
...
+         AND _ancestor NOT IN (SELECT _descendant FROM webknossos.folder_paths WHERE _ancestor = $idValidated)
+        """
+    val insertNewPathsQuery =
+      sqlu"""
+        INSERT INTO webknossos.folder_paths (_ancestor, _descendant, depth)
+        SELECT supertree._ancestor, subtree._descendant, supertree.depth + subtree.depth + 1
+        FROM webknossos.folder_paths supertree
+        CROSS JOIN webknossos.folder_paths subtree
+        WHERE subtree._ancestor = $idValidated
+        AND supertree._descendant = $newParentIdValidated
```
fm3, line 279:
```suggestion
        WHERE subtree._ancestor = $idValidated -- subtree is the OLD tree including self
        AND supertree._descendant = $newParentIdValidated  -- supertree is the ancestors of the NEW parent, including it
```

docs/datasets.md
```diff
@@ -207,6 +207,19 @@ Offers an option to delete a dataset and completely remove it from webKnossos. B
...
+## Dataset Organization
+
+In the dashboard, you can see all datasets of your organization (if you have the necessary permissions).
+Datasets can be organized within folders by simply dragging a dataset entry from the table to a folder in the left sidebar.
+New folders can be created by right-clicking an existing folder in the sidebar and selecting "New Folder".
+
+A folder can be edited (also via the context menu) to change its name or its access permissions.
+In the access permissions field, a list of teams can be provided that controls which teams should have access to that folder.
+Note that even a folder with an empty access permissions field can be accessed by users which have access to its parent folder.
+This is because the access permissions are handled cumulatively.
```
philippotto, line 219: @fm3 does this explanation make sense to you? I found it hard to explain the concept in a simple way and I'm also not sure whether it's correct :see_no_evil:

fm3: Sure, works for me :)

philippotto:
@daniel-wer I think, this PR is ready for review and testing. Let me know if you want to have a call about this :)

daniel-wer (commented):
First part of the frontend review. I did not dig into the core additions of the PR yet.

frontend/javascripts/dashboard/advanced_dataset/dataset_action_view.tsx
```diff
@@ -240,5 +237,76 @@ class DatasetActionView extends React.PureComponent<Props, State> {
     );
   }
 }
+const onClearCache = async (
```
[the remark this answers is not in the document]
philippotto: Yes! I opened https://github.com/scalableminds/webknossos/issues/6654 as a follow-up.

frontend/javascripts/admin/project/project_list_view.tsx
```diff
@@ -97,9 +94,8 @@ class ProjectListView extends React.PureComponent<PropsWithRouter, State> {
     this.fetchData();
   }
 
-  // @ts-expect-error ts-migrate(7006) FIXME: Parameter 'prevProps' implicitly has an 'any' type... Remove this comment to see the full error message
-  componentDidUpdate(prevProps) {
-    persistence.persist(this.props.history, this.state);
+  componentDidUpdate(prevProps: Props) {
+    persistence.persist(this.state);
```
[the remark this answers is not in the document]
philippotto: good catch!

frontend/javascripts/admin/api/folders.ts
```diff
@@ -0,0 +1,46 @@
...
+
+export async function createFolder(parentId: string, name: string): Promise<Folder> {
+  const params = new URLSearchParams();
+  params.append("parentId", parentId);
+  params.append("name", name);
+
+  const folder = await Request.receiveJSON(`/api/folders/create?${params}`, {
+    method: "POST",
+  });
+  folder.parent = parentId;
```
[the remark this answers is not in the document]
philippotto: Oh, this was a leftover and is not needed anymore. I removed it.

app/controllers/DataSetController.scala
jstriebel: > make nullable

@fm3 I did not find a change for that, but I might have missed it.

> double check that the frontend does not send very old values, loaded long before the move

Was this checked? @fm3 @philippotto

fm3: Good catch, I missed this. I now pushed a change where the backend dataset update route expects the folderId as nullable, and does not change the folderId if no new one is passed.

@philippotto Jonathan pointed out that when multiple users edit a dataset (or the same user in multiple tabs), this could lead to old values being re-used. Assume in tab A the dashboard is open, in tab B the dataset settings. In tab B the dataset settings are changed (say, isPublic), then, hours later, in tab A, the dataset is dragged to a new folder, triggering a new update request. This would set isPublic to false again, right? We thinkt he frontend should probably guard against this by fetching the dataset info again before updating it. Or alternatively, the fields should all be nullable and changed only if sent. For that we would of course need a new way of encoding setting stuff no None, e.g. the description.

app/controllers/DataSetController.scala
```diff
@@ -57,7 +57,7 @@ class DataSetController @Inject()(userService: UserService,
       (__ \ 'sortingKey).readNullable[Long] and
       (__ \ 'isPublic).read[Boolean] and
       (__ \ 'tags).read[List[String]] and
-      (__ \ 'folderId).read[ObjectId]).tupled
+      (__ \ 'folderId).readNullable[ObjectId]).tupled
```
jstriebel, line 60: PS: I just saw that the `folderId` is missing in the Swagger docs below (line 332)

app/controllers/DataSetController.scala
philippotto: Good point! I think, we have this problem potentially at other places, too, and it existed before for datasets in the dashboard as well (e.g., when editing tags). However, the issue is likely exacerbated now due to the folder property of datasets.

> Or alternatively, the fields should all be nullable and changed only if sent. For that we would of course need a new way of encoding setting stuff no None, e.g. the description. What do you think?

I think, this would be the best way. It doesn't solve all problems automatically (e.g., the tags are a single property of datasets where conflicting writes could also lose earlier updates), but it's a very good measure to avoid that unrelated change sets disturb each other. This is also how idiomatic, immutable updates in javascript are often done.

Differentiating between "should not be updated" and "should be set to null" is usually done by checking whether the key exists at all. So, this update payload `{description: null}` would remove the description, while this update payload `{tags: ["a tag"]}` wouldn't touch the description.

Is this possible in scala? If so, this would be great. Then, I can update the update-code in the front-end.

Additionally, regular polling would probably be a good idea, too, but I'd need to take care to not introduce janky updates to the lists.

> We thinkt he frontend should probably guard against this by fetching the dataset info again before updating it.

I'm not really fond of the added round trip and would consider this rather as a last resort.

fm3: > Differentiating between "should not be updated" and "should be set to null" is usually done by checking whether the key exists at all.

Hmm, I’m afraid there is no direct way to represent this in play/scala  – compare https://discuss.lightbend.com/t/distinguishing-between-omitted-and-null-values/7105

I think there are a few options
 - add something like `updateDescription: bool` to the parameters for all nullable fields. If this is true, then the value in the db should be set to the value passed in `description`, be it null or something else
 - create different routes for each field, call only those that need changing
 - use fully custom deserialization in the backend, and use double wrapped `Option[Option[String]]` to represent the different states, compare https://stackoverflow.com/questions/48510822/deserialize-json-distinguising-missing-and-null-values
 - do the roundtrip as indicated before

My personal preference at this time would be the first variant using booleans. What do you think? cc @jstriebel

philippotto: (1) and (2) sound a bit tedious as an interface to be honest. (3) would be fine with me, but it's probably tedious to implement, too.

Could the backend list the available keys of the passed JsObject to know whether a `None` value was actually listed or not? That would be akin to passing booleans (because the is-key-included-in-key-list check would return a boolean, too), but without explicitly passing these booleans?

jstriebel: I agree that all of the options are not super convincing. Would it be possible to add a custom type, e.g. Unset vs Set, which would be similar to an Option, and Unset is the default? We use something similar in the python libs:
https://github.com/scalableminds/webknossos-libs/blob/master/webknossos/webknossos/client/_generated/models/current_user_info_response_200.py#L36
Or would that also involve custom deserialization? I'd hope that there's a way to register custom deserialization hooks?

frontend/javascripts/dashboard/dataset/queries.tsx
```diff
@@ -0,0 +1,435 @@
...
+
+        if (context) {
+          queryClient.setQueryData(mutationKey, context.previousFolders);
+        }
+      },
+    },
+  );
+}
+
+export function useUpdateDatasetMutation(folderId: string | null) {
```
daniel-wer, line 292: The `update` in the name tripped me up a bit, because I was searching for where the updated values are sent to the server. I think this more of a refresh isn't it?

daniel-wer (commented):
Very impressive work :clap: I'm looking forward to when the old dataset list can be removed :)

I did some basic testing and noticed the following things:
- [x] In the search results the dataset selection no longer works
- [ ] ~The per-folder dataset count is wrong when "Show missing datasets" is activated~ see comment below by @philippotto 
- [x] Having more than two levels of folders will likely result in a horizontal scroll bar which doesn't look great and is rather clunky to work with. I had a look at what drive does and it seems they simply cut the folder name while giving the user the option to resize the left bar using the mouse but also resizing the left bar automatically once a folder with a too long name appears (within a limit it seems). Not high-priority, but I wanted to comment on that.
  expanding

Also, I think users will want to upload or add datasets to specific folders automatically in the future. It could be part of a followup PR to integrate this into the dataset upload and to specify how 3rd-party-tools which add datasets can assign them to a specific folder.

frontend/javascripts/dashboard/folders/folder_tree.tsx
```diff
@@ -0,0 +1,306 @@
+import React, { useEffect, useState } from "react";
+import { DropTargetMonitor, useDrop } from "react-dnd";
+import SortableTree, {
+  ExtendedNodeData,
+  FullTree,
+  NodeData,
+  OnDragPreviousAndNextLocation,
+  OnMovePreviousAndNextLocation,
+} from "react-sortable-tree";
```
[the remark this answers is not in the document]
philippotto: I looked into this today and the initial experience with the antd module was quite pleasant. However, I ran into a roadblock: I couldn't get dragging a dataset from the table into the sidebar to work. Apparently, antds own drag-and-drop behavior (for moving folders) prevents the custom "drop target" from working. If I disable the `draggable` property, dragging datasets to folders works again, but obviously moving folders doesn't work then, anymore.

I found the following related issues:

https://github.com/react-component/tree/issues/376
https://github.com/react-component/tree/issues/419
https://github.com/react-component/tree/issues/443
https://github.com/react-component/tree/issues/470

So, it seems to be known, but not being really actively worked on. Maybe we could find a workaround without touching the library code, but I don't really know how this could be done, since I don't know how the drag/drop events are intercepted in the first place.

I pushed my changes to `folders-antd`, but unless you have a good idea, I would think that it's not really worth to investigate further right now :( It's a bummer, since the styling really looked far better.. Maybe I'll adapt the styling a bit manually, so that we can draw some benefits from antd in the end.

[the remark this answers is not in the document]
daniel-wer: Argh, that's too bad :( Thank you for trying! Indeed, the issues and the one PR look effectively abandoned. Also, the code in the PR looks like a larger change so I am not too optimistic that there's an easy solution. I'll have a quick look at the `folders-antd` branch, though :)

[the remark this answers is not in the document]
philippotto: Ok, after reading my comment from above I had an idea how to workaround this issue! I don't want to toot my own horn, but I think it's pretty great (and simple) :smile: See https://github.com/scalableminds/webknossos/commit/7cdae93cd934b4a217ab791ab5be124085255efd. The idea is that folder DnD and dataset DnD are effectively incompatible, but they don't need to be compatible **at the same time**. So, as soon as a dataset is being dragged, I disable the tree's DnD behavior (and luckily, this information is already available :)).

I don't know whether there will be other roadblocks, but I think the critical stuff is working then (folder + dataset DnD, context menu).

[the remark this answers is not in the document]
daniel-wer: Uuh nice, that would be a great workaround, fingers crossed :grin:

app/controllers/DataSetController.scala
fm3: A custom wrapper type is an interesting Idea, could be used just for such update requests. I’ll have a look.
The other question is whether this should block this PR. The problem certainly existed before folders, as philipp has already stated. I’d vote to make this a follow-up issue, what do you think?

frontend/javascripts/dashboard/dataset/dataset_collection_context.tsx
```diff
@@ -0,0 +1,276 @@
...
+
+const ACTIVE_FOLDER_ID_STORAGE_KEY = "activeFolderId";
+
+export default function DatasetCollectionContextProvider({
+  children,
+}: {
+  children: React.ReactNode;
+}) {
+  const [activeFolderId, setActiveFolderId] = useState<string | null>(
+    UserLocalStorage.getItem(ACTIVE_FOLDER_ID_STORAGE_KEY) || null,
```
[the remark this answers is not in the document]
fm3: Hmm, good question. No strong opinion, but I would probably expect it to be at the root folder when I open it again. That seems to be how google tools behave as well. On the other hand, if a user works with the same three datasets in a particular folder for an extended period of time, they might find the additional clicks bothersome. I’d say it depends on how quickly they can get to their folder again … (assuming they don’t want to bookmark the url with folder id). Is the left tree view expanded by default? If not, maybe its last expansion state could be persisted? Or alternatively, we could add a “bookmarks” section or even “recently used“ for folders and/or datasets? Those would be follow-up issues of course.

app/controllers/DataSetController.scala
jstriebel: > I’d vote to make this a follow-up issue, what do you think?

Sound's fair for me. I'd say this issue is more pressing now, since different views (that easily are opened in separate tabs and for different users) now provoke this problem, and not just editing a single UI view in parallel. But it does not need to block this PR :+1:

fm3: I’m afraid even this was already the case before (I believe tags can be added in the dashboard) 😅 

In either case, I created the issue https://github.com/scalableminds/webknossos/issues/6657

jstriebel (commented):
Backend LGTM :+1:

frontend/javascripts/dashboard/dataset/queries.tsx
philippotto: Admittedly, this is not super elegant, but the this mutation can in fact do both (and both are used). I added this comment now:

```js
  /*
   * This mutation can either trigger a simple refresh of the dataset
   * (only pass the APIDatasetId) or it can update the actual dataset
   * when the tuple [APIMaybeUnimportedDataset, string] is passed.
   */
```

It's also used to edit tags, for example.

philippotto:
> * [ ]  The per-folder dataset count is wrong when "Show missing datasets" is activated

I see that this can be very confusing, but in the end the number reflects the amount of datasets in the folder and not how many datasets are shown. I adapted the UI to make this more clear with an asterisk and a tooltip. What do you think?

image

Adapting the number to the current filters would be a bit more work and I'm not sure if I'm fully on board with that :see_no_evil: (I'm obviously biased, though).

Is this alright with you? If not, I'd put this into a follow-up, as I don't think that this should block the PR.

frontend/javascripts/dashboard/dataset/queries.tsx
```diff
@@ -0,0 +1,435 @@
...
+                  parent: newParentId,
+                }
+              : oldFolder,
+          ),
+        );
+        return {
+          previousFolders,
+        };
+      },
+      onSuccess: (updatedFolder, [folderId, newParentId]) => {
```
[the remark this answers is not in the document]
philippotto: It's not strictly necessary, but I thought, if there is a fresh folder instance from the server, why not use it, as it could have been changed by someone else (e.g., renamed). However, I DRYed the code a bit to avoid the redundancy :)

frontend/javascripts/dashboard/dataset/queries.tsx
```diff
@@ -0,0 +1,435 @@
...
+    {
+      mutationKey,
+      onSuccess: (updatedDataset) => {
+        console.log("setQueryData for", mutationKey);
+        queryClient.setQueryData(mutationKey, (oldItems: APIMaybeUnimportedDataset[] | undefined) =>
+          updateDatasetInQueryData(updatedDataset, folderId, oldItems),
+        );
+        const targetFolderId = updatedDataset.folderId;
+        if (targetFolderId !== folderId) {
+          // The dataset was moved to another folder. Add the dataset to that target folder
```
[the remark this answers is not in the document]
philippotto: Yes! This happens above line 314. The code called `updateDatasetInQueryData` to do that, however, I just inlined the code since the additional function didn't really make sense (originally, I refactored it, because I thought I could reuse the code).

frontend/javascripts/dashboard/dataset/dataset_collection_context.tsx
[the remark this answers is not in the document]
philippotto: > On the other hand, if a user works with the same three datasets in a particular folder for an extended period of time, they might find the additional clicks bothersome.

That's what I had in mind.

> Or alternatively, we could add a “bookmarks” section or even “recently used“ for folders and/or datasets?

Yes, I think this could be a cool upgrade in the future. However, let's wait for some real user feedback.

I'd vote to keep it now as is (i.e., remember the last active folder).

philippotto:
@daniel-wer Thank you for your thorough review! I just finished the feedback and would be happy if you could have a look at my newer commits. The antd tree works pretty well. The only downside is that the library doesn't support the `canDrop` property. However, I think it's okay. The user will see a toast warning, if they try to drop a folder somewhere where it's not allowed.

app/controllers/DataSetController.scala
philippotto: Just for the record: I implemented a polling mechanism so that the datasets are updated regularly and automatically. Therefore, the chance of conflicting updates is greatly reduced, since the updates would need to be in the same time window (I think it's 60 seconds).

frontend/javascripts/dashboard/dataset/queries.tsx
```diff
@@ -0,0 +1,435 @@
...
+
+  return useMutation((folder: FolderUpdater) => updateFolder(folder), {
+    mutationKey,
+    onSuccess: (updatedFolder) => {
+      queryClient.setQueryData(mutationKey, (oldItems: FlatFolderTreeItem[] | undefined) =>
+        (oldItems || []).map((oldFolder: FlatFolderTreeItem) =>
+          oldFolder.id === updatedFolder.id
+            ? {
+                ...updatedFolder,
+                parent: oldFolder.parent,
```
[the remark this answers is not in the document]
daniel-wer: Got it, thanks for clarifying.

daniel-wer (commented):
This is looking better and better! Thank you for investing the time to replace the react-sortable-tree with the antd one :1st_place_medal: 

> Adapting the number to the current filters would be a bit more work and I'm not sure if I'm fully on board with that see_no_evil (I'm obviously biased, though).
> 
> Is this alright with you? If not, I'd put this into a follow-up, as I don't think that this should block the PR.

Yes, agreed. I think the solution with the tooltip works well :)

I did some more testing:
- [x] Testing with a non-Super-Admin, there were a couple of inconsistencies (your call whether they should be part of a followup PR):
  - [x] Folder creation failed with the message "folder.create.failed" which should get a proper description
  - [x] Trying to update a folder's allowed teams gave "Failed to update the folder’s name" which is misleading, because the folder name was not changed.
  - [x] Trying to add a tag gave two error messages: "Could not update dataset. [object Object]" and "You are not authorized to view or edit this resource" which should be consolidated and maybe made more descriptive.

- [x] Also, I noticed one (minor) issue directly after logging in. For some time (depending on the network connection I would assume), a view is shown, indicating that there are no datasets and that one should be imported. After half a second, the screen rerenders and the dataset list is shown. Maybe there could be a loading spinner, instead?

philippotto:
@daniel-wer Thanks for your testing feedback! The points should be done now. Please have another look at my last commits :)

daniel-wer (approved):
Frontend and tests LGTM :tada: 

What I didn't properly test are issues that result from multiple people modifying the folder structure at once. I know that the dataset polling exists to mitigate this problem a bit, but I'm not sure about the actual folder hierarchy. Did you discuss this?

philippotto:
> What I didn't properly test are issues that result from multiple people modifying the folder structure at once. I know that the dataset polling exists to mitigate this problem a bit, but I'm not sure about the actual folder hierarchy. Did you discuss this?

When a folder is edited, its data is fetched before displaying the modal. So, the risk of conflicts is rather low then. For the folder hierarchy, I just enabled a polling mechanism so that the risk should be quite low there, too.

If you don't veto, I'd merge and deploy this tomorrow morning :)

philippotto:
@fm3 I added the sql migration to the migration guide. Nothing else needs to happen for the migration, right? And `Needs datastore update after deployment` is not needed, right?

fm3:
Sounds about right to me :) Go for it! If you don’t deploy too early in the morning I should also be able to be there and support should anything go wrong

frontend/javascripts/oxalis/view/context_menu.tsx
```diff
@@ -953,12 +959,20 @@ function ContextMenuContainer(props: Props) {
...
 }
 
+function ContextMenuContainer(props: Props) {
+  return (
+    <GenericContextMenuContainer {...props}>
+      <ContextMenuInner {...props} />
+    </GenericContextMenuContainer>
+  );
+}
+
```
MichaelBuessemeyer, line 975: It looks to me like these changes introduced some kind of bug I still cannot understand. Due to some reason, the hooks of ContextMenuInner are not called consistently in the same order.
image
I'll open an issue for this.

PR #6593, opened 2022-10-27
WIP: Add route to remove annotation layer

### LINKED ISSUE #5802: Add option to delete layers from an annotation
#5771 introduced the concept of annotation layers. An annotation can for example have multiple volume annotation layers (which can later be toggled visible vs invisible) – There should be routes and a UI to add a new or delete an existing layer from an annotation.

frcroth:
- https://___.webknossos.xyz

### Steps to test:
- abc

### Issues:
- fixes #5802 

------
- [ ] Ready for review


app/controllers/AnnotationController.scala
```diff
@@ -225,6 +225,26 @@ class AnnotationController @Inject()(
       } yield result
     }
 
+  @ApiOperation(hidden = true, value = "")
+  def removeAnnotationLayer(typ: String, id: String, layerName: String): Action[AnyContent] =
+    sil.SecuredAction.async { implicit request =>
+      for {
+        _ <- bool2Fox(AnnotationType.Explorational.toString == typ) ?~> "annotation.makeHybrid.explorationalsOnly"
+        annotation <- provider.provideAnnotation(typ, id, request.identity)
+        _ <- annotationService.findLayer(annotation, layerName) ?~> "No such layer found"
```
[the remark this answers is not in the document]
fm3: Also, I think the annotation returned by the above provideAnnotation already has an annotationLayers list which can be used to check if the layer exists, so the findLayer function might not be needed after all.

fm3 (commented):
Works for me, very cool!

I added a few minor suggestions. Also two things I’d like to ask for: Could you add
 - logging: something along the lines of `"Deleting annotation layer with tracing id ${tracingId} for annotation ${annotationId}"` so that we can restore the layer in case of emergency.
 - prevent zero-layer annotations. In the delete route, there should be an assertion that there are at least two layers so that deleting leaves one in place. (one could also think about adding an sql constraint there but I think it’s not worth the effort. if a user manages to create a zero-layer annotation despite the assertion, it will have kind of a broken state but it’s not the end of the world)

fm3:
@daniel-wer Do you maybe have capacity to add a front-end button and confirm modal here?

app/controllers/AnnotationController.scala
```diff
@@ -225,6 +225,26 @@ class AnnotationController @Inject()(
       } yield result
     }
 
+  @ApiOperation(hidden = true, value = "")
+  def removeAnnotationLayer(typ: String, id: String, layerName: String): Action[AnyContent] =
+    sil.SecuredAction.async { implicit request =>
+      for {
+        _ <- bool2Fox(AnnotationType.Explorational.toString == typ) ?~> "annotation.makeHybrid.explorationalsOnly"
+        annotation <- provider.provideAnnotation(typ, id, request.identity)
```
fm3, line 233:
```suggestion
        annotation <- provider.provideAnnotation(typ, id, request.identity)
        _ <- bool2Fox(annotation._user == request.identity._id) ?~> "notAllowed" ~> FORBIDDEN
```

app/controllers/AnnotationController.scala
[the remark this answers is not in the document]
frcroth: ah yes

app/controllers/AnnotationController.scala
frcroth: It is clear that there needs to be some authorization, but is it really intended that only the creator of the annotation can edit it?

fm3: Fair question! The answer is not entirely clear, but yes, I’d build it like this for now.

There are two other groups to consider
 - people who “administrate” this user (organization admins and team managers): here the usual workflow is that they are *not* allowed to edit the annotation at all. Instead they should use “copy to my account” if they want to create a copy to modify themselves. So not allowing them to delete a layer is in line with the existing behavior.
 - people who the annotation was shared with if `othersMayEdit` was set to true by the annotation owner: here we could also allow them to delete layers but I think since it is a potentially drastic destructive action it would be safer not to allow this at this time. If people who are using the othersMayEdit feature request this in the future, we may still change this.

fm3 (commented):
Backend LGTM! Let’s wait to see if the front-end will be implemented in this PR

app/controllers/AnnotationController.scala
frcroth: :+1:

daniel-wer:
@fm3 Thanks for taking care of the frontend!

The code looks good to me, but during testing I noticed that removing a skeleton layer led to a broken annotation. I investigated and traced this back to the so called "viewMode". For legacy reasons there seem to be four different view modes (`["orthogonal", "flight", "oblique", "volume"]`) instead of the expected three ("volume" isnt't really different from "orthogonal" afaics). The backend sends the `allowedModes` as part of the "/info" request's `settings` property. The `viewMode` is also part of the url. Now, in a "hybrid" annotation, the view mode seems to be "orthogonal", but after removing the skeleton layer, this mode is no longer allowed, the backend only sends "volume" - which leads to the broken annotation.
As you can see this is a real mess. I'm wondering whether we should use this opportunity to clean this up. I think we should be able to replace "volume" with "orthogonal" while making sure that we keep the url parsing backwards-compatible. Any thoughts on that? Also /cc @philippotto 

P.S. I also found one comment in the code from @fm3: "// todo: refactor MODE_VOLUME away or make this logic compatible"

philippotto:
> I think we should be able to replace "volume" with "orthogonal" while making sure that we keep the url parsing backwards-compatible. Any thoughts on that? Also /cc @philippotto

Yes, I think so, too. I don't know why the differentiation for "volume" should still be necessary.

> P.S. I also found one comment in the code from @fm3: "// todo: refactor MODE_VOLUME away or make this logic compatible"

I think, I vaguely remember writing that comment. Probably within a PR merged by @fm3.


Should I have a look regarding the removal of the "volume" mode in the frontend? As I understand it, the backend also would need to be adapted, right?

daniel-wer:
> Should I have a look regarding the removal of the "volume" mode in the frontend? As I understand it, the backend also would need to be adapted, right?

If you have the time, that would be great! Yes, I think the backend would need to be adapted as well. I'm not sure whether that's complicated - depends on whether the `allowedModes` are generated on-the-fly or are read from the DB, I would assume.

fm3:
In the backend the logic seems to be
```
if (isTask):
  send allowed modes for its tasktype as stored in the db
else:
  generate on the fly depending on existing annotation layers
```

The backend itself does not use the allowed modes, it only sends them to the frontend. 

So we should still define a migration for the task types. Should volume then always be replaced by orthogonal (while keeping the allowed modes distinct)?

fm3:
Apparently the front-end never allowed “volume” as an allowed mode for task types, so there aren’t actually any task types with “volume”, which simplifies things

philippotto:
> Apparently the front-end never allowed “volume” as an allowed mode for task types, so there aren’t actually any task types with “volume”, which simplifies things

Great :+1: I didn't find any handling of the "volume" case, anyway. So, this adds up at least. I cleaned up the typing in the front-end and hadn't to do any behavior changes. So, this PR should be good to go. @daniel-wer Could you have a look at my last commit?

fm3:
My guess is that we should remove the on-the-fly-sent “volume” in the backend as well?

philippotto:
> My guess is that we should remove the on-the-fly-sent “volume” in the backend as well?

Yes, this value for allowedModes should simply die, I think.

fm3:
I removed the volume task type from the backend, now sending orthogonal in that case.
I was wondering if there may be any trouble if the user is in flight mode first, and then that one disappears from the list of allowed modes (leaving only orthogonal)? It seemed to work fine when I tested it, but maybe you have more insight?

Otherwise, I’d say this PR is ready to go if there are no objections?

philippotto:
> I removed the volume task type from the backend, now sending orthogonal in that case.

:+1: 

> I was wondering if there may be any trouble if the user is in flight mode first, and then that one disappears from the list of allowed modes (leaving only orthogonal)? It seemed to work fine when I tested it, but maybe you have more insight?

In general, when loading an annotation, the allowed modes are verified against the active mode and switched when necessary. So, there shouldn't be any problem, as far as I know.

fm3 (approved):

daniel-wer:
@philippotto Sorry, I wasn't able to comment in time since I was busy with meetings. Did you test whether older links which have `viewMode` 3 in the url (volume), still work - so the url manager? Also, in your last commit I saw `ViewModeValuesIndices` in the constants.ts - is this still used somewhere?

philippotto:
> Did you test whether older links which have `viewMode` 3 in the url (volume), still work - so the url manager?

Yes, that works :)

> Also, in your last commit I saw `ViewModeValuesIndices` in the constants.ts - is this still used somewhere?

Yes, it's still used :)

PR #6595, opened 2022-10-27
Assert that there are workers for a datastore on job creation

### LINKED ISSUE #6582: Some jobs can be submitted for datastores without workers
I saw two instances of forever-pending export-tiff jobs that were somehow submitted for datasets that are on datastores with no associated workers. This should not be offered in the front-end nor should it be accepted by the backend.

frcroth:
- https://___.webknossos.xyz

### Steps to test:
- Create a job on a datastore without a worker
- Create a job on a datastore with a worker

### Issues:
- fixes #6582 

------
- [x] Ready for review


app/models/job/Worker.scala
```diff
@@ -81,6 +83,12 @@ class WorkerService @Inject()(conf: WkConf) {
       "lastHeartBeatIsRecent" -> lastHeartBeatIsRecent(worker)
     )
 
+  def assertDataStoreHasWorkers(dataStoreName: String)(implicit ctx: DBAccessContext): Fox[Boolean] =
```
fm3, line 86:
```suggestion
  def assertDataStoreHasWorkers(dataStoreName: String)(implicit ctx: DBAccessContext): Fox[Unit] =
```

app/models/job/Worker.scala
```diff
@@ -81,6 +83,12 @@ class WorkerService @Inject()(conf: WkConf) {
...
+    for {
+      _ <- dataStoreDAO.findOneByName(dataStoreName)
+      _ <- workerDAO.findOneByDataStore(dataStoreName)
+    } yield true
```
fm3, line 90:
```suggestion
    } yield ()
```
(since the boolean return value is never read, I’d instead return unit)

app/models/job/Worker.scala
```diff
@@ -81,6 +83,12 @@ class WorkerService @Inject()(conf: WkConf) {
       "lastHeartBeatIsRecent" -> lastHeartBeatIsRecent(worker)
     )
 
+  def assertDataStoreHasWorkers(dataStoreName: String)(implicit ctx: DBAccessContext): Fox[Boolean] =
+    for {
+      _ <- dataStoreDAO.findOneByName(dataStoreName)
+      _ <- workerDAO.findOneByDataStore(dataStoreName)
```
fm3, line 89:
```suggestion
      _ <- workerDAO.findOneByDataStore(dataStoreName) ?~> "jobs.noWorkerForDatastore"
```
If adding the messages happens here, it can be removed from the controller, saving quite a few redundant strings :)
In the mesesages file, I’d suggest `Processing jobs are not available for the datastore this dataset belongs to`

fm3 (commented):
LGTM, see my small comments

fm3 (approved):
:tada:

PR #6601, opened 2022-11-02
Hotfix Dataset View Mode

philippotto:
always create annotationToolsGeometries to avoid accidental null access; especially, this fixes the broken view-mode

- https://___.webknossos.xyz

### Steps to test:
- open DS in view mode

### Issues:
- follow-up to #6542 

------
- [X] Ready for review

### DIFF
--- a/frontend/javascripts/oxalis/controller/scene_controller.ts
+++ b/frontend/javascripts/oxalis/controller/scene_controller.ts
@@ -392,13 +392,11 @@ class SceneController {
     const taskBoundingBox = getSomeTracing(state.tracing).boundingBox;
     this.buildTaskingBoundingBox(taskBoundingBox);
 
-    if (state.tracing.volumes.length > 0) {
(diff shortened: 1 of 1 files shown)


frontend/javascripts/oxalis/controller/scene_controller.ts
```diff
@@ -392,13 +392,11 @@ class SceneController {
     const taskBoundingBox = getSomeTracing(state.tracing).boundingBox;
     this.buildTaskingBoundingBox(taskBoundingBox);
 
-    if (state.tracing.volumes.length > 0) {
```
philippotto, line 395: the if-check was an optimization which isn't really worth it in my opinion. the instance var types of the classes where not-nullable (which wasn't correct). making them nullable would require a lot of other null-checks in various places. this is why I went with making them actually not-nullable.

daniel-wer (approved):

PR #6603, opened 2022-11-03
Fix getNullBucket crash and remove dead code

philippotto:
- https://hotfixgetNullBucketcrash.webknossos.xyz

### Steps to test:
- create a new annotation and restrict the mags during creation (e.g., to one mag)
- open the annotation --> it should not crash

### Issues:
- follow-up for #6542
- see https://scm.airbrake.io/projects/95438/groups/3398209786679717674/notices/3398729977235811716?tab=notice-detail

------
- [X] Ready for review


frontend/javascripts/oxalis/model/bucket_data_handling/data_cube.ts
```diff
@@ -165,12 +156,8 @@ class DataCube {
     this.temporalBucketManager = new TemporalBucketManager(this.pullQueue, this.pushQueue);
   }
 
-  getNullBucket(address: Vector4): Bucket {
-    if (this.boundingBox.containsBucket(address, this.resolutionInfo)) {
-      return NULL_BUCKET;
-    } else {
-      return NULL_BUCKET_OUT_OF_BB;
-    }
```
philippotto, line 173: As far as I see, this distinction was only used in the `ArbitraryCubeAdapter`. However, that wasn't used at all, anymore. So, I unified the two `NullBuckets`.

yarn.lock
```diff
@@ -11613,7 +11613,7 @@ rc-table@~7.26.0:
     rc-util "^5.22.5"
     shallowequal "^1.1.0"
 
-rc-tabs@^12.2.2, rc-tabs@~12.2.0:
+rc-tabs@~12.2.0:
```
philippotto, line 11616: `yarn install` produced this change (not for the first time). I think, it needs to be there.

frontend/javascripts/oxalis/model/bucket_data_handling/bounding_box.ts
```diff
@@ -47,9 +47,15 @@ class BoundingBox {
   });
 
   containsBucket([x, y, z, zoomStep]: Vector4, resolutionInfo: ResolutionInfo): boolean {
-    const { min, max } = this.getBoxForZoomStep(
-      resolutionInfo.getResolutionByIndexOrThrow(zoomStep),
-    );
+    /* Checks whether a bucket is contained in the active bounding box.
+     * If the passed resolutionInfo does not contain the passed zoomStep, this method
+     * returns false.
+     */
```
philippotto, line 53: This method is only used by `DataCube.isWithinBounds` which itself also checks for `this.cubes[zoomStep] == null`. So, the new behavior here is in line with the existing behavior on the call site.
The other usage of `containsBucket` was removed due to the `NullBucket` consolidation.

daniel-wer (approved):
Nice fix and cleanup :+1: I'll do a quick test once the CI is ready.

Do you follow why this suddenly was an issue after the quick select PR?

philippotto:
> I'll do a quick test once the CI is ready.

The CI currently has some problems. I hope it resolves itself:

```
[error]   download error: Caught java.net.ConnectException (Connection timed out (Connection timed out)) while downloading https://repo1.maven.org/maven2/com/google/protobuf/protobuf-java/3.15.8/protobuf-java-3.15.8.pom
````

> Do you follow why this suddenly was an issue after the quick select PR?

Yes, the quick-select PR changed the bounding box method to accept the resolution info. The old code simply accessed the `Store` to get the resolutions (for the entire dataset). I had to change this due to cyclic references. The resolution info that is passed only refers to the layer (instead of the datasets resolutions). This made sense to me because the bounding box was used in the context of a layer, but I didn't think of the case where non-existing zoom steps are requested. Changing it back to use the dataset resolutions would have probably been ok, too, but I wanted to avoid the usage of `Store` (especially, because I was burnt with the cyclic dependencies before).

daniel-wer:
Thank you for clarifying, makes sense :)

PR #6609, opened 2022-11-04
Create a layer for each channel of NGFF-Zarr datasets

### LINKED ISSUE #6578: Add ChannelIndex to MagLocator, adapt Zarr Import for Multichannel datasets
Remote Zarr datasets can have n channels. wk can only handle single channels per layer (or uint24 rgb layers, which is a special case). To import multi-channel datasets in wk, the MagLocator could store the channelIndex, and data loading could then use the selected channel. Multiple layers could then each load one of the source channels.

frcroth:
- ~~https://channelindex.webknossos.xyz/~~

### Steps to test:
- Import https://uk1s3.embassy.ebi.ac.uk/idr/zarr/v0.4/idr0062A/6001240.zarr via remote zarr import
- Should result in two layers, one for each channel
- give data loading a minute or two, then data should load (use wk layer coloring in left sidebar to distinguish the channels)

Not tested: Importing zarr datasets with only one channel. This should result in only one layer.

### Issues:
- fixes #6578 

------
- [x] Ready for review


app/models/binary/explore/NgffExplorer.scala
```diff
@@ -25,36 +25,66 @@ class NgffExplorer extends RemoteLayerExplorer {
...
+          layers <- layersFromNgffMultiscale(multiScale, remotePath, credentials, channelCount)
+        } yield layers
+      })
+      layers: List[(ZarrLayer, Vec3Double)] = layerLists.flatten
     } yield layers
 
-  private def layerFromNgffMultiscale(multiscale: NgffMultiscalesItem,
-                                      remotePath: Path,
-                                      credentials: Option[FileSystemCredentials]): Fox[(ZarrLayer, Vec3Double)] =
+  private def getNgffMultiScaleChannelCount(multiscale: NgffMultiscalesItem, remotePath: Path): Fox[Int] =
```
fm3, line 38:
```suggestion
  private def getNgffMultiscaleChannelCount(multiscale: NgffMultiscalesItem, remotePath: Path): Fox[Int] =
```
Please check that the spelling is consistent throught the file :) I’d go for multiscale as one word

app/models/binary/explore/NgffExplorer.scala
```diff
@@ -25,36 +25,66 @@ class NgffExplorer extends RemoteLayerExplorer {
...
       nameFromPath <- guessNameFromPath(remotePath)
       name = multiscale.name.getOrElse(nameFromPath)
-      voxelSizeNanometers = voxelSizeInAxisUnits * axisUnitFactors
-      layer: ZarrLayer = if (looksLikeSegmentationLayer(name, elementClass)) {
-        ZarrSegmentationLayer(name, boundingBox, elementClass, magsWithAttributes.map(_.mag), largestSegmentId = None)
-      } else ZarrDataLayer(name, Category.color, boundingBox, elementClass, magsWithAttributes.map(_.mag))
-    } yield (layer, voxelSizeNanometers)
+      layerTuples <- Fox.combined(
+        (0 until channelCount)
+          .map(channelIndex => {
```
fm3, line 63:
```suggestion
      layerTuples <- Fox.serialCombined(0 until channelCount) { channelIndex =>
```
serialCombined is a shortcut for this exact case (we don’t have multiple foxes before, but want to iterate over a sequence, and with each item we will get a new fox by the block)

fm3 (commented):
Looking good, and works for me! Please see my two small refactoring suggestions.

Also: could you add the passing of the channelIndex also for the N5Array (at N5BucketProvider)? The N5 explorer does not currently support inferring channel metadata (which is fine for now), but if added manually to the datasource-properties.json, this should also work.

frcroth:
> Also: could you add the passing of the channelIndex also for the N5Array (at N5BucketProvider)? The N5 explorer does not currently support inferring channel metadata (which is fine for now), but if added manually to the datasource-properties.json, this should also work.

I don't exactly know what you intend for the N5Array. Should the N5Array have another property channelIndex?

fm3:
> I don't exactly know what you intend for the N5Array. Should the N5Array have another property channelIndex?

Yes, and pass it to its superclass (that should be enough)

fm3 (approved):
:tada:

PR #6615, opened 2022-11-07
Fix importing a dataset from disk

philippotto:
- https://___.webknossos.xyz

### Steps to test:
- remove the datasource-properties.json for an existing dataset and reload the list in the dashboard
- try to import the dataset by using the inferred values from wk

### Issues:
- fixes a regression introduced with #6485 

------
- [X] Ready for review


philippotto:
> I think you branched off of folders – could you cherry pick the relevant commit here? innocent

Argh, good catch. I didn't branch wrong, though. Instead, I pulled the folders branch into this branch :facepalm:

philippotto:
> Maybe unrelated, but I noticed that there is no field to enter the `largestSegmentId` for the dataset I tested this with (ROI2017_wkw), although it has a segmentation layer. Is this expected or a regression (from another PR)?

Good catch! I think, it's a regression from the same PR. I fixed it now :)

daniel-wer (approved):
LGTM :+1:

PR #6616, opened 2022-11-07
Rephrased Error Messages

hotzenklotz:
Based on a recent issue and some ambiguity around the error messages, I had a quick look at the `messages.conf`. Changes:
- general rephrased selected messages to (hopefully) improve clarity
- replaced all shorthand `don't`, `can't` etc with proper long form `do not`, `cannot` etc
- removed 80 unused messages
  - message use count was done using a simple string search and might be prone to error
  - general command, e.g. `grep -H -R -o  "error.noUser"  ./app ./webknossos-datastore ./webknossos-tracingstore | wc -l | xargs -I{} echo "error.noUser"  {} >> message_counts.txt`
  - I removed the matched keys with inverted grep matching: `grep -v -f matches.txt conf/messages > conf/messages_filtered`
  - [count_messages_usage.sh.zip](https://github.com/scalableminds/webknossos/files/9953930/count_messages_usage.sh.zip)
  - [matches.txt](https://github.com/scalableminds/webknossos/files/9953920/matches.txt)
  - [message_counts.txt](https://github.com/scalableminds/webknossos/files/9953921/message_counts.txt)

### Steps to test:
- Read changes
- ??? 

### Issues:
- None

------
- [x] Ready for review


conf/messages
```diff
@@ -33,102 +19,71 @@ image.create.failed=Failed to create image
...
-organization.name.alreadyInUse=The chosen organization name is already in use.
+organization.name.invalid=This organization name contains illegal characters. Please only use letters and numbers.
+organization.name.alreadyInUse=This name is already claimed by a different organization and not available anymore. Please choose a different name.
 organization.alreadyJoined=Your account is already associated with the selected organization.
 
-user.activated={0} has been activated
 user.notFound=User not found
 user.noAdmin=Access denied. Only admin users can execute this operation.
-user.deactivated=Your account hasn’t been activated yet
+user.deactivated=Your account has not been activated by an admin yet. Please contact your organisation's admin for help.
```
fm3, line 43:
```suggestion
user.deactivated=Your account has not been activated by an admin yet. Please contact your organization’s admin for help.
```

conf/messages
```diff
@@ -33,102 +19,71 @@ image.create.failed=Failed to create image
...
-dataSet.file=Dataset ZIP File
 dataSet.notFound=Dataset {0} does not exist or could not be accessed
 dataSet.notFoundConsiderLogin=Dataset {0} does not exist or could not be accessed. You may need to log in.
 dataSet.notFoundForAnnotation=The Dataset for this annotation does not exist or could not be accessed.
 dataSet.noAccess=Could not access DataSet {0}. Does your team have access?
 dataSet.noAccessById=Could not access the corresponding DataSet. This is likely because you are not a member of a team that has access to it.
 dataSet.notImported=Dataset {0} is not imported
-dataSet.name.invalid=The dataset name is invalid
-dataSet.name.invalid.characters=Dataset names may only contain letters, digits, dots, underscorees, dashes.
+dataSet.name.invalid=The dataset name contains illegal characters. Please only use letters, numbers, and hypens.
```
fm3, line 72: This is not always the case, one of the three messages below should be in the chain. We may want to remove this message entirely so the ones from below can bubble up to the main message

conf/messages
```diff
@@ -345,27 +257,24 @@ serialization.failed=Failed to serialize data.
...
 job.couldNotRunGlobalizeFloodfills = Failed to start job for globalizing floodfills.
-job.couldNotRunApplyMergerMode = Failed to start job to apply merger mode tracing.
+job.couldNotRunApplyMergerMode = Failed to start job to apply merger mode annotation.
 job.disabled = Long-running jobs are not enabled for this webKnossos instance.
 jobs.worker.notFound = Could not find this worker in the database.
 job.export.fileNotFound = Exported file not found. The link may be expired.
 job.export.tiff.invalidBoundingBox = The selected bounding box could not be parsed, must be x,y,z,width,height,depth
 job.export.tiff.volumeExceeded = The volume of the selected bounding box is too large.
 job.export.tiff.edgeLengthExceeded = An edge length of the selected bounding box is too large.
-job.export.notAllowed.organization = Currently tiff export is only allowed for datasets of your own organization.
```
fm3, line 362: I feel like this should not be unused 🤔 could you have a quick look in the jobs controller? Should be analogous to the other jobs (see lines directly below here)

conf/messages
```diff
@@ -378,6 +287,4 @@ voxelytics.zeroRunWorkflow = No run for this workflow found
 
 publication.notFound = Publication could not be found.
 
-agglomerateSkeleton.failed=Could not generate agglomerate skeleton.
-
-unexpectedReturn=Unexpected return of parameters
+agglomerateSkeleton.failed=Could not generate agglomerate skeleton.
```
fm3, line 290: final line break please :)

fm3 (commented):
Very nice refactoring! Please see my small comments. Also could you add a changelog entry?
I can add a refresh-snapshots run when all else is ready

conf/messages
fm3: We use organization with a z (american english) in wk everywhere, as far as I can tell

conf/messages
hotzenklotz: Looks like this was replaced with very specific responses instead. This general form is not used anywhere as far as I can tell with string search.

image

conf/messages
hotzenklotz: I am not sure what you want me to do here? None of the messages were removed. I just rephrased them.

conf/messages
fm3: Ok, apparently this was intentional after all, see https://github.com/scalableminds/webknossos/pull/6219 – that means the tiff-export-specific message was indeed intentionally unused :+1:

fm3 (approved):
With a changelog entry, this should be good to go :)

PR #6618, opened 2022-11-09
Integrate Rome Linting

philippotto:
Integrates rome linter and disables eslint except for the rules-of-hooks which are not yet supported by rome unfortunately. Nevertheless, the CI time is reduced by ~30s. Once we integrate rome formatting, we can shave of another 10 seconds.

- https://___.webknossos.xyz

### Steps to test:
- CI should be enough

### Issues:
- improves speed for linting

------
- [X] Ready for review


package.json
```diff
@@ -180,7 +172,6 @@
     "deep-freeze": "0.0.1",
     "dice-coefficient": "^2.1.0",
     "distance-transform": "^1.0.2",
-    "eslint-plugin-ava": "^13.2.0",
```
daniel-wer, line 183: There's probably no replacement for the ava https://github.com/avajs/eslint-plugin-ava/blob/main/docs/rules/no-only-test.md, right? I remember that we added this one just recently, because we got bitten by that.

daniel-wer (approved):
Awesome, thanks for taking care of this! :+1:

package.json
philippotto: Good point! I replaced it with a very simplistic bash file: https://github.com/scalableminds/webknossos/pull/6618/commits/11fbb9292c90cd5c60b96643325f297dd7dcd0e6
It's not super accurate (since the only modifier could be formulated in another way), but I think it should catch 95% of the cases :)

daniel-wer: Sweet, thank you!

PR #6622, opened 2022-11-09
Allow viewing vx workflows via link if organization matches

### LINKED ISSUE #6619: Voxelytics reports should be viewable for anyone in the right orga who has the link
Currently, voxelytics reports have no concept of unlisted-but-accessible. We want to change this, so that anyone in the right organization can view a workflow report if they have the link

fm3:
- https://vxunlisted.webknossos.xyz

### Steps to test:
- Run vx workflow, reporting to this instance
- You should be able to see and list the workflow
- Another user in the same orga should see via link but not list the workflow

### Issues:
- fixes #6619 

------
- [x] Ready for review


jstriebel (approved):
LGTM :+1:

PR #6625, opened 2022-11-09
Proofreading without skeletons

### LINKED ISSUE #6507: [Proofreading] Make agglomerate skeletons unnecessary
Agglomerate skeletons are rather hard to understand as a user. We should try to make them non-mandatory for most use cases.

- [ ] for merging it should not be necessary to load the skeleton. instead, there should be the notion of an "active segment". thus, rightclicking another segment will make the source-target pair clear for the merge operation
- [ ] for splitting we might want to use a different technique to show the super-voxels. one way could be to use the meshes and render them in the data viewports with cleverly applied clipping. that way, we don't need another segmentation layer for that. I imagine this to show the contours of the super voxels effectively.

daniel-wer:
Allow to proofread by clicking on segments instead of agglomerate skeleton nodes.
The position of the active segment is now updated when clicking onto a segment and also highlighted in proofreading mode using a small white cross hair. Afterwards, users can right-click somewhere else and select the entry to split or merge (a) segment(s).
Additionally, the active segment is highlighted a bit more than the currently hovered segment to make it more prominent and more easily recognizable.

- https://proofreadingwithoutskeletons.webknossos.xyz

### Steps to test:
- Create an annotation, activate an agglomerate view mapping and select the proofreading tool
  - Click on some segments - the click position should be highlighted with a zoom-independent cross hair + the active segment should be highlighted with higher opacity
  - Right-click somewhere else and select "Merge with active Segment" or "Split from active Segment"
- Open datasets with different scales - the cross hair should always have the same size.
- Use "Shift + Leftclick" and "Ctrl + Leftclick" to merge/split segments without the context menu. The description in the bottom bar should adapt automatically.
- Loading agglomerate skeletons and deleting/adding edges should still work for proofreading, although it's less discoverable (by design, might be removed in the near future)

### Issues:
- fixes #6507

------
- [x] Ready for review


frontend/javascripts/oxalis/controller/combinations/bounding_box_handlers.ts
```diff
@@ -227,7 +227,8 @@ export function getClosestHoveredBoundingBox(
 export const highlightAndSetCursorOnHoveredBoundingBox = _.throttle(
   (delta: Point2, position: Point2, planeId: OrthoView) => {
     const hoveredEdgesInfo = getClosestHoveredBoundingBox(position, planeId);
-    const inputCatcher = document.getElementById(`inputcatcher_${planeId}`);
+    // Access the parent element as that is where the cursor style property is set
+    const inputCatcher = document.getElementById(`inputcatcher_${planeId}`)?.parentElement;
```
daniel-wer, line 231: During testing I noticed that the cursor never changed again after activating the bounding box tool, because the style property was set for the wrong element (and never was removed again).

frontend/javascripts/oxalis/geometries/materials/plane_material_factory.ts
```diff
@@ -148,6 +155,9 @@ class PlaneMaterialFactory {
       globalMousePosition: {
         value: new THREE.Vector3(0, 0, 0),
       },
+      activeSegmentPosition: {
+        value: new THREE.Vector3(0, 0, 0),
```
philippotto, line 159: will this show a crosshair at 0, 0, 0 when switching to the proofreading tool initially?

philippotto:
Great work :+1: Feels already pretty good! See my comments about the overloading of the left-click. During testing, I clicked on a node which was weird, because the proofreading position was set to the click position, but afterwards the z-coordinate was changed to the node (due to clipping distance > 1). Therefore, I couldn't see the crosshair, anymore.

frontend/javascripts/oxalis/controller/combinations/tool_controls.ts
```diff
@@ -754,14 +758,10 @@ export class ProofreadTool {
...
     plane: OrthoView,
-    _event: MouseEvent,
+    event: MouseEvent,
     isTouch: boolean,
   ) {
-    const didSelectNode = SkeletonHandlers.handleSelectNode(planeView, pos, plane, isTouch);
-    if (didSelectNode) {
-      // Don't do anything else
-      return;
-    }
```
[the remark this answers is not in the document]
daniel-wer: You're right, I'm not sure why I removed the condition.
I didn't know that it was possible to select a node using right-click. In that case, yes, let's remove the left-click node selection from proofreading mode as we want to discourage working with agglomerate skeletons, anyways.

frontend/javascripts/oxalis/geometries/materials/plane_material_factory.ts
daniel-wer: Yep it is if the bounding box starts at 0, 0, 0. There's a small party going on there, another guest is the center dot of the quick select tool (took me some time to figure out where the dot comes from) :grin: 

I'll take care of both issues :)

philippotto (approved):
Excellent :100: Also thank you for fixing the quick select geometry at 0,0,0 :pray:

daniel-wer:
I'll merge this on Monday in case anything goes wrong :)

PR #6628, opened 2022-11-10
Fix rerender after each layer name keypress in dataset import view

### LINKED ISSUE #6624: Dataset Edit View: Typing in Layer Name field triggers rerender, input loses focus
In the dataset edit view, when typing in the layer name, every keypress triggers a rerender of the layer list item, because of ```<List.Item key={`layer-${layer.name}`}>``` (with the name changing).

This makes the layer name input very cumbersome, as i have to click it again after every typed letter.

What can be used as a react key instead to avoid this?

fm3:
- https://editviewfixrerender.webknossos.xyz (I’ll deploy after CI success)

### Steps to test:
- Import a remote zarr dataset
- In the edit view, change a layer name
- typing there should not trigger re-render after each keypress, causing the text field to lose focus

### Issues:
- fixes #6624

------
- [x] Ready for review


hotzenklotz (approved):
LGTM

PR #6632, opened 2022-11-10
Add Terms of Service Acceptance Concept

### LINKED ISSUE #6611: Add owner for organizations and ToS acceptance
## Detailed Description
We want to have a single user that is the "owner" of an organization. That user is responsible for accepting the terms of service and our contact person for inquiries about the organization. 

We also need owners to accept our updated terms of service. For that, we need to show a modal in case there are new terms of service to accept them and persist the acceptance. I would suggest to store a version number of the ToS and show the modal whenever there is a newer ToS version (e.g. in the server config). We should also store the date of the last ToS acceptance.
We should probably block access for regular users to the organization, in case the owner has not accepted the new ToS.

## Context
* [x] Specific to webKnossos.org (set `isDemoInstance=true` in `application.conf`)

fm3:
New config keys `webKnossos.termsOfService.{enabled, version, content}`, new routes
`GET /termsOfService`, `POST termsOfService/accept`, `GET /termsOfService/acceptanceNeeded`

The idea is that after login, the frontend checks acceptanceNeeded and shows a popup asking for new TOS acceptance. If the user is orga owner, they cann call `accept` and that’s it, if they are not, they are told `please contact your orga owner`.

Note that if termsOfService is not explicitly enabled in the config, `acceptanceNeeded` should always return false.

Note that when migrating an existing instance, every orga should get one organization owner.

### TODO Frontend
 -  [X] adapt types (user json now contains isOrganizationOwner)
 -  [X] use routes, show modals, as described above

- https://tos.webknossos.xyz

### Steps to test:
- set `termsOfService.enabled` to true in application.conf
- ensure that you are logged in
- open some page on wk
- a modal should appear and prompt you to accept the ToS
- log in as a second user (which is not an admin)
- a modal should appear and ask you to contact the orga admin
- snoozing
  - the default deadline is currently in 2023, so the modals should be closable
  - reload the page --> the modals shouldn't re-appear
  - wait more than 20 seconds --> the modals should appear again
- with exceeded deadline:
  - change the deadline in application conf to a past date (e.g., 2021)
  - check that the modals cannot be closed without hitting accept
  - after acceptance, the modals shouldn't appear again
- bump the ToS version in the application.conf
  - the modals should appear again

### Issues:
- fixes #6611 

------
- [X] Ready for review


philippotto:
I finished the basic functionality. Looks like this:

image


@fm3 When I tested this, I stumbled a bit, since I had to set the tos version to 1 (since the lastTOSAcceptanceVersion defaults to 0). maybe it could default to -1? or one checks the last TOS acceptance time against zero..

conf/evolutions/reversions/091-terms-of-service.sql
```diff
@@ -0,0 +1,28 @@
...
+SELECT
+u._id AS _user, m.email, u.firstName, u.lastname, o.displayName AS organization_displayName,
+u.isDeactivated, u.isDatasetManager, u.isAdmin, m.isSuperUser,
+u._organization, o.name AS organization_name, u.created AS user_created,
+m.created AS multiuser_created, u._multiUser, m._lastLoggedInIdentity, u.lastActivity
+FROM webknossos.users_ u
+JOIN webknossos.organizations_ o ON u._organization = o._id
+JOIN webknossos.multiUsers_ m on u._multiUser = m._id;
+
+UPDATE webknossos.releaseInformation SET schemaVersion = 91;
```
daniel-wer, line 25:
```suggestion
UPDATE webknossos.releaseInformation SET schemaVersion = 90;
```

This also seems to be wrong for `evolutions/reversions/090-cleanup.sql`!

conf/messages
```diff
@@ -38,6 +38,8 @@ organization.name.invalid=This organization name contains illegal characters. Pl
 organization.name.alreadyInUse=This name is already claimed by a different organization and not available anymore. Please choose a different name.
 organization.alreadyJoined=Your account is already associated with the selected organization.
 
+termsOfService.versionMismatch=Terms of service version mismatch. Current version is {0}, recieved acceptance for {1}
```
daniel-wer, line 41:
```suggestion
termsOfService.versionMismatch=Terms of service version mismatch. Current version is {0}, received acceptance for {1}
```

daniel-wer (commented):
Frontend LGTM :+1: 

I think the backend still needs a review?

I'm a little bit worried that this might be fairly disruptive. In case the organization owner is not immediately available, all members need to wait for the ToS acceptance. I think a more user friendly pattern would be to prompt the organization owner for acceptance a couple of weeks before (repeatedly), noting the deadline and that work of the members will be disrupted if the ToS are not accepted by then. I know that this would be more work to implement, but I wanted to note it. Have you thought about or discussed this?

conf/evolutions/reversions/091-terms-of-service.sql
fm3: Whoops, luckily we never had to run that reversion. (Not that it would have done much harm). Fixed it now in both!

fm3:
> I'm a little bit worried that this might be fairly disruptive. In case the organization owner is not immediately available, all members need to wait for the ToS acceptance. I think a more user friendly pattern would be to prompt the organization owner for acceptance a couple of weeks before (repeatedly), noting the deadline and that work of the members will be disrupted if the ToS are not accepted by then. I know that this would be more work to implement, but I wanted to note it. Have you thought about or discussed this?

Would forward this to @normanrz – I think apart from the implementation effort, there may be legal things to consider?

philippotto:
> > I'm a little bit worried that this might be fairly disruptive. In case the organization owner is not immediately available, all members need to wait for the ToS acceptance. I think a more user friendly pattern would be to prompt the organization owner for acceptance a couple of weeks before (repeatedly), noting the deadline and that work of the members will be disrupted if the ToS are not accepted by then. I know that this would be more work to implement, but I wanted to note it. Have you thought about or discussed this?
> 
> Would forward this to @normanrz – I think apart from the implementation effort, there may be legal things to consider?

@normanrz and I briefly discussed this and agreed that we should try to find a simple solution to mitigate the described problem. One idea would be the following:

The application.conf gets another ToS parameter which would be a `deadlineTimestamp` that contains the (ideally, a human-readable?) date on which the new ToS have to be agreed upon. When requesting the ToS state, the result could be `{needsAcceptance: boolean, deadlineTimestamp: timestamp}`. If the deadline is not reached yet, the modals would be closable.

What do you think about this @fm3 ?

normanrz:
I was thinking, we might want to put the ToS text on weblium and just include an iframe in webKnossos.

normanrz:
> I was thinking, we might want to put the ToS text on weblium and just include an iframe in webKnossos.

But markdown is also fine. No need to derail this PR.

normanrz:
We just got the first draft of the ToS. I now think, we should definitely use an iframe instead of inline markdown. The main reason is that we need to show both German and English versions and I would like to use a table for side-by-side texts. That is probably annoying to do in markdown.

philippotto:
Ok, do you already have a tentative URL ready to test it with?

normanrz:
https://weblium.webknossos.org/terms-of-service

philippotto:
Ok, weblium needs to allow iframes, though. Currently, I'm getting `Refused to frame 'https://weblium.webknossos.org/' because an ancestor violates the following Content Security Policy directive: "frame-ancestors 'none'".`

normanrz:
Hmm. Maybe that will work on same-origin.

fm3:
@philippotto I added the two new fields `acceptanceDeadline` (number) and `acceptanceDeadlinePassed` Boolean, using current server time, to the `/termsOfService/acceptanceNeeded` route.

Note that we should set enabled to false again in the config before merging this PR

app/oxalis/security/CombinedAuthenticatorService.scala
```diff
@@ -26,11 +25,6 @@ case class CombinedAuthenticator(actualAuthenticator: StorableAuthenticator) ext
   override def isValid: Boolean = actualAuthenticator.isValid
 }
 
-class IdentityCookieSigner extends Signer {
-  override def sign(data: String): String = data
-  override def extract(message: String): Try[String] = Success(message)
-}
```
jstriebel, line 32: :+1:

jstriebel (commented):
Nice stuff, backend mostly LGTM :+1: Only left some minor comments:

tools/postgres/schema.sql
```diff
@@ -301,6 +303,7 @@ CREATE TABLE webknossos.users(
   userConfiguration JSONB NOT NULL,
   isDeactivated BOOLEAN NOT NULL DEFAULT false,
   isAdmin BOOLEAN NOT NULL DEFAULT false,
+  isOrganizationOwner BOOLEAN NOT NULL DEFAULT false,
```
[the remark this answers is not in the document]
fm3: I had a look at this as well, but there are apparently a few downsides to the unique index method, see the comments here https://stackoverflow.com/a/16236566/7564727 – This might not be a problem for us, but since we don’t have a hard requirement for just a single owner, I’d leave it as is for now.

util/src/main/scala/com/scalableminds/util/tools/ConfigReader.scala
```diff
@@ -1,10 +1,20 @@
...
 
+import com.typesafe.config.Config
 import play.api.{ConfigLoader, Configuration}
 
+import java.time.Instant
+
 trait ConfigReader {
   def raw: Configuration
 
+  implicit val instantConfigLoader: ConfigLoader[Instant] = new ConfigLoader[Instant] {
```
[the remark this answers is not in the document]
fm3: Yes, this is used when reading the new config value for the acceptanceDeadline. It is of type `Instant`, which means that an implicit `ConfigLoader[Instant]` needs to be in scope (but play doesn’t come with one).

jstriebel (commented):
Backend is fine from my side, didn't do any testing though. Leaving the approval for the frontend reviewer.

frontend/javascripts/components/terms_of_services_check.tsx
```diff
@@ -0,0 +1,149 @@
...
+import React, { useState } from "react";
+import { useSelector } from "react-redux";
+import { formatDateInLocalTimeZone } from "./formatted_date";
+
+const LAST_TERM_OF_SERVICE_WARNING_KEY = "lastTermOfServiceWarning";
+
+export function CheckTermsOfServices() {
+  const lastWarningString = UserLocalStorage.getItem(LAST_TERM_OF_SERVICE_WARNING_KEY);
+  const lastWarning = moment(lastWarningString ? parseInt(lastWarningString) : 0);
+  const isLastWarningOld = moment().diff(lastWarning, "seconds") > 20;
```
philippotto, line 22: todo: extract and set to reasonable constant (e.g., 3 days?).

conf/application.conf
```diff
@@ -74,6 +74,13 @@ webKnossos {
 
     Please add the information of the operator to comply with GDPR.
   """
+  termsOfService {
+    enabled = true
```
philippotto, line 78: set to false before merging

philippotto:
@daniel-wer The front-end part is ready for a second review. I integrate a snooze functionality so that the ToS acceptance can be snoozed if the deadline hasn't passed yet. The modals will re-appear on the next page load if a specific duration has passed in the meantime (currently, 20 seconds for testing; probably will change to 3 days).

daniel-wer (approved):
LGTM if the remaining ToDos are taken care of. Thank you for making it easy to test on the dev instance :pray: 

- [ ] The ToS shouldn't include the hamburger menu or website footer. Not sure whether that's something for this PR or needs to be changed in the ToS weblium page.

philippotto:
> [ ] The ToS shouldn't include the hamburger menu or website footer. Not sure whether that's something for this PR or needs to be changed in the ToS weblium page.

I'd assume, it would be the easiest to change this on weblium. @normanrz Is this possible? Currently, it looks like this:

image

conf/application.conf
philippotto, line 78:
```suggestion
    enabled = false
```

PR #6633, opened 2022-11-10
Fix creating a task using a base NML

philippotto:
- https://___.webknossos.xyz

### Steps to test:
- open the "create a task" dialog
- select "Upload NML File" and upload an NML
- fill the remaining values
- create a task --> should work

### Issues:
- follow-up for https://github.com/scalableminds/webknossos/pull/6486/files

------
- [X] Ready for review


fm3 (commented):
I’m afraid this breaks task creation *without* nml.

Note that the form and bulk task creation use the same request unless NMLs are added, so there will probably be a need to distinguish between with or without NML for this field

philippotto:
Hm, ok, then, I'd suggest that the backend should always accept/expect `taskTypeIdOrSummary`. Would this be alright?

fm3:
I created a new PR for the backend version https://github.com/scalableminds/webknossos/pull/6635

hotzenklotz (approved):
🚀

philippotto:
Closed in favor of #6635.

Issue #5080, opened 2021-01-22
Annotation sharing: should you see the layer config of the annotation owner?


fm3:
When sharing your annotation it may be desirable that the other user sees it just as you do, i.e. with your layer settings (same visibility + intensity settings). Let’s discuss if this is a good idea and when to copy settings from where to where

philippotto:
I think, the most plausible way would be to provide a checkbox in the sharing modal à la: "Use my current view settings (e.g., layer visibility, intensity)". Then, the settings are persisted somehow so that they can be restored when the URL is opened. However, it's probably hard to pinpoint the exact "when" of the snapshot creation. Especially, when we want to remove the explicit save button from the modal.

That's why an easier variant could be to just copy the settings from the owner when the link is opened (as you probably meant in the issue description). This would probably be "good enough" for most usecases, anyway. However, we would need to pay attention that the wording is precise in the UI so that the user understands that their actions *after* creating the sharing link can influence the behavior of the link.

The layer settings are not version-managed by any chance? Then, the solution could be to allow both variants..

fm3:
> The layer settings are not version-managed by any chance?

Not at this time, never had a reason to. I’d think the overhead is also not really justified for this use case

> copy the settings from the owner when the link is opened

Yes, that could be a solution.

Another problem concerns the state of the viewing user’s settings. Assuming they are a logged-in user, and not just a one-time viewer. The annotation author’s settings for both the dataset and annotation tab are copied. Where are they copied to? Are they just used while viewing this particular annotation? What happens if the viewing user changes a setting afterwards? Is that persisted to their own view configuration? I’m afraid both doing so and not doing so could create confusion. I don’t know how to clearly communicate this. (Also I’m not 100% clear about the settings behavior myself, the infos from both tabs are saved in the same place, right? UserDatasetViewConfiguration?)

philippotto:
We should probably discuss this in person :speech_balloon: Additional relevant topics could be:

- sharing the activated mapping
- sharing loaded isosurfaces
- more?

normanrz:
Could we encode the layer settings in the url? Maybe with an integrated backend-implemented URL shortening.

philippotto:
Most of the discussion above is outdated by now. The current sharing code can encode layer-specific things in the URL and already does that for mappings, meshes, connectomes etc.

It's probably a good idea to at least add the following:
- layer visibility

[...]

PR #6634, opened 2022-11-10
Encode layer visibility in sharing link

daniel-wer:
- https://sharelayervisibility.webknossos.xyz

### Steps to test:
- Open an annotation, enable and disable layers, set the annotation to public and copy the sharing link
  - Open the annotation in an incognito tab, the layer visibility should be equal to the original annotation (also for the Skeleton layer)
  - Reload the original annotation, the layer visibility should remain the same

### Issues:
- contributes to #5080

------
- [x] Ready for review


philippotto (commented):
The code looks good to me :+1: However, during testing I got the following toast when opening the sharing link:

`URL configuration values for the layer "color" are ignored, because: The requested layer with name color is not a segmentation layer.`

This should probably not happen, right?

The visibility was correct for my layers, though. It was a public annotation for a private dataset with with a visible color layer, invisible volume layer and visible skeleton layer. I'll share the link with you privately.

daniel-wer:
> However, during testing I got the following toast when opening the sharing link:
> 
> URL configuration values for the layer "color" are ignored, because: The requested layer with name color is not a segmentation layer.
> 
> This should probably not happen, right?

Uh, good job catching that :see_no_evil: Fixed in the newest version.

philippotto (approved):
Awesome, works for me now :+1:

PR #6635, opened 2022-11-10
Fix task creation with base NML

fm3:
Turns out the frontend now also sends the id with the json key taskTypeIdOrSummary in the createFromFiles case. This PR renames the expected backend parameter to match that.

It does not, however, allow for summaries, this would have made the code quite a bit more complicated.

I’m starting to think that this whole IdOrSummary thing may have been a mistake (as in, the harder-to-maintain code is not justified by the small ux improvement). Let’s discuss that in person next week.

- https://taskcreationnml.webknossos.xyz

### Steps to test:
- test that task creation works with and without passed nml file, and also using the bulk creation tab.

------
- [x] Ready for review


philippotto (approved):
the changes look good to me :+1: 

> Let’s discuss that in person next week.

we could discuss this today after the standup if you want.

PR #6637, opened 2022-11-11
Fix NGFF import for datasets with no channel axis

fm3:
follow-up for #6609 – axisOrder.c being None wasn’t handled.

### Steps to test (I tested this locally):
- import ngff dataset with no channel dimension (e.g. https://s3.embl.de/i2k-2020/platy-raw.ome.zarr), should run through + load data
- import ngff dataset with channel dimension (e.g. https://uk1s3.embassy.ebi.ac.uk/idr/zarr/v0.4/idr0062A/6001240.zarr), should also still run through + load data

------
- [x] Ready for review


frcroth (approved):
Thanks!

Issue #6630, opened 2022-11-10
Share chunkContents cache across multiple arrays


fm3:
We imported https://uk1s3.embassy.ebi.ac.uk/idr/zarr/v0.4/idr0047A/4496763.zarr and data loading took forever (like >5 minutes)
It is expected that populating the chunk cache is relatively slow (chunk shape is 1*1*2048*2048) but after that, data loading should be faster.

We should double check that now superfluous requests are made, that all caches are shared as expected, and investigate whatever else may be slowing this down.

normanrz:
Are the histogram requests connected to the cache as well?

fm3:
Should be, yes

fm3:
Looks like the chunk cache is limited to 100 MB per array, that’s not enough for 32 slices of this dataset.

We should look into sharing a single cache for all arrays, with a fixed size and shared LFU strategy.

fm3:
related: #6969 (item weight)

PR #6639, opened 2022-11-14
Hotfix: increase zarr chunk cache size

fm3:
Hotfix for #6630 – ultimately, a shared cache would be better, but that is more complex to build. I’d say increasing the per-array cache size is unlikely to give us memory problems at this time, as the entire array including its cache should be cleared by the outer caching (DataCubeCache) regularly.

I will look into this in more detail when tackling the issue #6630 properly with a shared cache.


philippotto (approved):
:+1:
we should keep the higher RAM consumption in mind after merging, in case trouble arises.

PR #6644, opened 2022-11-16
Include scale bar in screenshots

### LINKED ISSUE #3978: Scale bar not saved with the screenshot
See https://discuss.webknossos.org/t/scale-bar-not-saved-with-the-screenshot/1446

daniel-wer:
Was easier than thought :) The lines added only stem from me adding new lines between functions in the utils module.

These are automatic screenshots made using the webKnossos functionality:

l4_sample__3584_3584_1024__TDView
l4_sample__3584_3584_1024__PLANE_XY

- https://includescalebarinscreenshots.webknossos.xyz

### Steps to test:
- Press `Q` or select "Screenshot" from the menu. The scale bar should be included in the screenshots if it was activated. The 3d view buttons should not be included.

### Issues:
- fixes #3978 

------
- [x] Ready for review

### DIFF
--- a/CHANGELOG.unreleased.md
+++ b/CHANGELOG.unreleased.md
@@ -17,6 +17,7 @@ For upgrade instructions, please check the [migration guide](MIGRATIONS.released
 - Added ability to shuffle / set colors for a whole tree group. [#6586](https://github.com/scalableminds/webknossos/pull/6586)
 - Annotation layers can now be removed. [#6593](https://github.com/scalableminds/webknossos/pull/6593)
 - When adding remote Zarr datasets with multiple channels, channels are converted into layers. [#6609](https://github.com/scalableminds/webknossos/pull/6609)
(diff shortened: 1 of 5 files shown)


daniel-wer:
I did not find a workaround for the "blurry" lines of the scale bar. It seems to depend on the screen resolution and size of the viewports. Also happens without the transparent background or with z-index set :shrug: I don't think it's a blocker, but feel free to test yourself if you have any additional ideas :)

philippotto (approved):
Awesome :man_dancing: By the way, on my setup, there are no blurry lines in the screenshots :)

daniel-wer:
> By the way, on my setup, there are no blurry lines in the screenshots :)

Interesting. I suspected the 200% scaling of my Hi-DPI screen to cause the blurry lines, but that's not the case. Instead, I noticed that the overlay did not work on Hi-DPI screens with a `devicePixelRatio > 1`. Was easy to fix and I added a note to the data canvas Hi-DPI issue to remove this workaround again if we ever complete said issue.

PR #6647, opened 2022-11-17
Create Annotation From View Mode: Keep Activated Mapping

### LINKED ISSUE #6645: When creating annotation form view mode, keep active mapping
The Create Annotation button in dataset view mode should pass the previously selected mapping to the createExplorational call, which should propagate it to the created VolumeTracing object.

fm3:
- https://createexplorationalmapping.webknossos.xyz

### Steps to test:
- Create annotation from dashboard, should work as before
- Create annotation from view mode (color layer only), should work as before
- Create annotation from view mode (with segmentation with no mapping active), should work as before
- Create annotation from view mode (with segmentation and hdf5 mapping selected), annotation should have that mapping active in the fallback segmentation

### Issues:
- fixes #6645

------
- [x] Ready for review


philippotto:
> @philippotto I wasn’t sure how to access the currently active mapping from the action bar view, could you maybe have a quick look?

Yes, I'm looking into it. Should this work with JSON mappings, too, or only hdf5 mappings?

jstriebel (commented):
Backend LGTM

daniel-wer (approved):
Works well for me :+1: 

Do you think the same should be the case when making a segmentation layer editable (and a mapping is activated)? To me it looks like everything is there to support that, so it might be a quick thing to add. Otherwise, feel free to ignore and merge this PR as-is :)

philippotto:
> Do you think the same should be the case when making a segmentation layer editable (and a mapping is activated)? To me it looks like everything is there to support that, so it might be a quick thing to add. Otherwise, feel free to ignore and merge this PR as-is :)

I would wait until #6506 is done, because this will probably affect how we want to present the default precomputation options (e.g., make the unmapped precomputation the default, because it's the most powerful one) :)

philippotto:
> > Do you think the same should be the case when making a segmentation layer editable (and a mapping is activated)? To me it looks like everything is there to support that, so it might be a quick thing to add. Otherwise, feel free to ignore and merge this PR as-is :)
> 
> I would wait until #6506 is done, because this will probably affect how we want to present the default precomputation options (e.g., make the unmapped precomputation the default, because it's the most powerful one) :)

Oh, I got confused with #6651. I'll have a look!

philippotto:
The `/api/annotations/${annotationType}/${annotationId}/addAnnotationLayer` would need to be adapted to consume the mapping parameter, too. @fm3 is this an easy addition? if not, feel free to say no and we will tackle this once it pops up again :)

fm3:
> The /api/annotations/${annotationType}/${annotationId}/addAnnotationLayer would need to be adapted to consume the mapping parameter, too

Did you try it? I’d say it should already be adapted as well automagically :)

philippotto:
Cool :) Yes, it works. I pushed the changes now. @daniel-wer Do you want to have another look?

daniel-wer (approved):
Thank you! LGTM :+1:

PR #6648, opened 2022-11-17
Deduplicate bboxes when importing NML

### LINKED ISSUE #6562: When merging annotations, deduplicate bounding boxes
When we introduced user bounding boxes in https://github.com/scalableminds/webknossos/pull/4536 we already talked about deduplicating the bounding boxes during annotation merge (both backend and frontend). We discarded the attempts at the time, because the first implementations didn’t perform as expeted and we didn’t want to block the PR. Now users have reported that the bbox buildup is annoying. We should look into that again.

Should be done in:
- [x] back-end
- [ ] front-end

philippotto:
- https://deduplicatebboxes.webknossos.xyz

### Steps to test:
- create an annotation
- create a bbox
- download the NML (use the dropdown menu in the toolbar)
- import the NML with DnD
- the bbox should not be duplicated

### Issues:
- fixes #6562 

------
- [X] Ready for review


daniel-wer (approved):
Nice, LGTM! :+1:

PR #6651, opened 2022-11-18
Pass mapping name to precompute meshes

### LINKED ISSUE #6585: agglomerateView mapping name is not passed to compute mesh file job call
The compute mesh file job can take an agglomerateView parameter. However, it looks like this is never set. See `startComputeMeshFileJob` in segments_view.tsx. I would have expected this to be set to the currently selected mapping when launching the job

philippotto:
- https://___.webknossos.xyz

### Steps to test:
- I opened a dataset with an agglomerate mapping
- activated the mapping
- hit precompute meshes
- tested that the meshes work with that mapping

### Issues:
- fixes #6585

------
- [X] Ready for review


daniel-wer (approved):
Code LGTM :+1:, did not test again.

PR #6652, opened 2022-11-21
Use LZ4 with WASM for volume saving/undo/redo

### LINKED ISSUE #5970: Speed up undo/redo for volume data (with WASM based lz4)
## Detailed Description
For undo/redo (and saving) of volume data we currently use the [lz4js](https://www.npmjs.com/package/lz4js) package which roughly takes 20ms on my machine per bucket. I'd hope to get faster speeds with a wasm-based lz4 approach. At the moment, these two WASM based lz4 packages might be useful:
https://www.npmjs.com/package/lz4-wasm
https://www.npmjs.com/package/wasm-lz4

For saving to back-end, we need a frame-less version of lz4. In the worst case, we could stick to the old lz4js library for that part, though.

## Context

This would speed up working with volume tracings.

philippotto:
See #5970 and https://github.com/scalableminds/webknossos/pull/6652#issuecomment-1324805564 for context and numbers.

- https://lzwasm.webknossos.xyz/

### Steps to test:
- brush and use undo/redo
- brush and save and reload page
- everything should work as usual

### Issues:
- fixes #5970

------
- [X] Ready for review


frontend/javascripts/libs/utils.ts
```diff
@@ -886,7 +886,7 @@ export function castForArrayType(uncastNumber: number, data: TypedArray): number
 }
 
 export function convertNumberTo64Bit(num: number | null): [Vector4, Vector4] {
-  if (num == null) {
+  if (num == null || isNaN(num)) {
```
philippotto, line 889: not sure how I triggered this, but an error was thrown here at one point which is why I slipped this fix in here.

frontend/javascripts/test/model/volumetracing/volume_annotation_sampling.spec.ts
```diff
@@ -8,7 +9,7 @@ import anyTest from "ava";
 import datasetServerObject from "test/fixtures/dataset_server_object";
 import mockRequire from "mock-require";
 import sinon from "sinon";
-mockRequire.stopAll();
```
philippotto, line 11: the `stopAll`s reverted the lz4 mocking (which is necessary since the lz4-wasm module is different for browser and node, unfortunately). it took me a while to figure out that stopAll was the culprit, since it's so inconspicuously in the code. luckily, these statements weren't even necessary.

it's a bit unfortunate that the lz4 mock needs to be at that many places, but I couldn't figure out a better way..

daniel-wer (commented):
Wow, extremely impressive improvements! :exploding_head:  Thank you for researching and testing :pray:

Unfortunately, I don't get the mocking yet. Why do files that don't have any direct lz4 usages need to import the lz4 mock? - and wouldn't that also import the mock in production? Why not mock only in the test files where it's needed using `mockRequire` like we usually do? :thinking:

philippotto:
> Why do files that don't have any direct lz4 usages need to import the lz4 mock? - and wouldn't that also import the mock in production? Why not mock only in the test files where it's needed using `mockRequire` like we usually do?

Oh, two non-spec files slipped in there by accident. I corrected these now.

philippotto:
Hmm, the wasm doesn't load on the dev instance -.- I'm looking into it..

philippotto:
> Hmm, the wasm doesn't load on the dev instance -.- I'm looking into it..

@fm3 I think the problem is that the backend serves the `*.wasm` file with `content-type: application/octet-stream`. However, it needs to be `application/wasm`. Could you have a look for how to adapt this?

philippotto:
@daniel-wer please have another look :) now, the deployment at https://lzwasm.webknossos.xyz should work, too.

frontend/javascripts/test/backend-snapshot-tests/annotations.e2e.ts
```diff
@@ -1,3 +1,4 @@
+import "test/mocks/lz4";
```
daniel-wer, line 1: Instead of importing the mock in every test file, I think it should work and would be more foolproof to import it once in `frontend/javascripts/test/enzyme/e2e-setup.ts` or do you think the import order could be problematic, then? :thinking: 

Sorry to comment on the mocking again. It's no big deal if you leave it as-is.

daniel-wer (approved):
LGTM with or without my suggestion :+1:

During testing of the dev instance, the performance improvement compared to master didn't appear that large. Is that the case for you too? Still, there is an improvement :)

philippotto:
I also bumped the pako library (use for gzip compression when sending save requests) because the version we used dates back to 2015 :see_no_evil: The new version is two weeks old. I couldn't see measurable speed improvements, but it shouldn't hurt nonetheless.

Issue #6604, opened 2022-11-03
Cannot save volume annotation for dataset consisting of zarr links


philippotto:
## Steps to Reproduce the bug

- [ ] Cannot reproduce the bug anymore / needs deeper investigation.

1. Open an annotation on webknossos.org (see here for a link: https://scm.slack.com/archives/C5AKLAV0B/p1667468655289759)
2. Create a zarr private link
3. Use Import Zarr on localhost to import all layers from the private link
4. Change the category to segmentation for the segmentation layers
5. Save
6. Create a new annotation for that dataset
7. Brush & Save --> crashes

PR #6660, opened 2022-11-24
Guard against invalid-mag bucket volume save actions

fm3:
This PR ensures that the front-end doesn't try to save buckets for magnifications which don't exist. This caused save errors before (without any helpful message).
The scenario only happened when layers with differing mags existed (e.g., mag 2-2-2 vs mag 2-2-1) which is not supported, anyway. However, the save logic used layer-independent resolutions which is not the correct way of handling this, which is why I took the opportunity to clean this up a bit.

- https://volumemagerror.webknossos.xyz

### Steps to test:
- The actual error scenario is hard to reproduce (the same dataset worked for @fm3 but not for me). I'd suggest to test the happy paths instead, since these are far more important.
- Create an annotation and brush (also use floodfill, since a bit of code changed there) --> save and reload
- Create another annotation (e.g., with layer restrictions) and test again

### Issues:
- https://scm.slack.com/archives/C5AKLAV0B/p1667468655289759
- related to https://github.com/scalableminds/webknossos/issues/6604

------
- [X] Ready for review


webknossos-tracingstore/app/com/scalableminds/webknossos/tracingstore/tracings/volume/VolumeTracingService.scala
```diff
@@ -95,7 +95,7 @@ class VolumeTracingService @Inject()(
               case a: UpdateBucketVolumeAction =>
                 if (tracing.getMappingIsEditable) {
                   Fox.failure("Cannot mutate volume data in annotation with editable mapping.")
-                } else updateBucket(tracingId, tracing, a, updateGroup.version)
+                } else updateBucket(tracingId, tracing, a, updateGroup.version) ?~> "Failed to save volume data."
```
philippotto, line 98: When I provoked the error, I only saw this message here and not the "received a mag...". is this intentional? should there be an error chain?

frontend/javascripts/test/model/binary/layers/wkstore_adapter.spec.ts
```diff
@@ -15,12 +15,18 @@ const RequestMock = {
 const { dataSource } = datasetServerObject;
 let _fourBit = false;
 
+// test
+
```
philippotto, line 19:
```suggestion
```

daniel-wer (approved):
Code LGTM :+1: 

I encountered one issue during testing:
- [x] When restricting the volume annotation (for example, to start at 4-4-4) and then using the quick select tool in 1-1-1, the following error toast is shown: "Error: Magnification with index 0 does not exist.". This might be alright, but I wanted to mention it since all other annotation tools work in this mag.

philippotto:
> [x] When restricting the volume annotation (for example, to start at 4-4-4) and then using the quick select tool in 1-1-1, the following error toast is shown: "Error: Magnification with index 0 does not exist.". This might be alright, but I wanted to mention it since all other annotation tools work in this mag.

Thanks for catching this! The issue was unrelated to the PR, but I fixed it, anyway :) See my last commit.

webknossos-tracingstore/app/com/scalableminds/webknossos/tracingstore/tracings/volume/VolumeTracingService.scala
fm3: I think there should be a chain. I’ll have another look

fm3: You were right, there was a conversion that ate the error chain. Should be fixed now :)

fm3:
I’d say the backend changes are simple enough that they don’t need another review. @philippotto assuming you are happy with the front-end, feel free to merge :)

PR #6663, opened 2022-11-28
Ability to recover from webGL context loss

### LINKED ISSUE #6653: Recover from webGL context loss
Handle the `webglcontextrestored` event to recover from a context loss. Needs investigation to find out how much threejs can do for us here and how much of our own rendering logic has to be touched.

philippotto:
- https://restorefromcontextloss.webknossos.xyz

### Steps to test:
- open a dataset or an annotation
- open the console
- execute `window.testContextLoss()`
- wait for a few seconds (the recovery event is dispatched after 2.5 seconds which then triggers the actual recovery mechanism)

### Issues:
- fixes #6653 

------
- [x] Ready for review

### DIFF
--- a/.circleci/not-on-master.sh
+++ b/.circleci/not-on-master.sh
@@ -1,8 +1,8 @@
 #!/usr/bin/env bash
 set -Eeuo pipefail
 
-# if [ "${CIRCLE_BRANCH}" == "master" ]; then
+if [ "${CIRCLE_BRANCH}" == "master" ]; then
   echo "Skipping this step on master..."
-# else
-#  exec "$@"
-# fi
+else
+  exec "$@"
+fi
--- a/.circleci/not-on-master.sh
+++ b/.circleci/not-on-master.sh
@@ -1,8 +1,8 @@
 #!/usr/bin/env bash
 set -Eeuo pipefail
 
-if [ "${CIRCLE_BRANCH}" == "master" ]; then
+# if [ "${CIRCLE_BRANCH}" == "master" ]; then
   echo "Skipping this step on master..."
-else
-  exec "$@"
-fi
+# else
+#  exec "$@"
+# fi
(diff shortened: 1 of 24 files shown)


frontend/javascripts/libs/UpdatableTexture.ts
```diff
@@ -152,7 +152,6 @@ class UpdatableTexture extends THREE.Texture {
       src,
     );
     this.gl.bindTexture(this.gl.TEXTURE_2D, activeTexture);
-    this.image = null;
```
philippotto, line 155: This resource is necessary when re-uploading textures to the GPU after a context recovery. I compared the heap memory consumption between this branch and the master and could not find a noticeable difference.

frontend/javascripts/oxalis/controller/renderer.ts
```diff
@@ -20,10 +17,24 @@ function getRenderer() {
...
+  const ext = renderer.getContext().getExtension("WEBGL_lose_context");
+  if (ext == null) {
+    return;
+  }
+  ext.loseContext();
+  setTimeout(() => ext.restoreContext(), 2500);
+}
+
+// @ts-ignore
+window.testContextLoss = testContextLoss;
```
philippotto, line 37: I'll remove this after testing.

.circleci/not-on-master.sh
```diff
@@ -1,8 +1,8 @@
 #!/usr/bin/env bash
 set -Eeuo pipefail
 
-if [ "${CIRCLE_BRANCH}" == "master" ]; then
+# if [ "${CIRCLE_BRANCH}" == "master" ]; then
```
daniel-wer, line 4: Reminder to comment this back in

daniel-wer (approved):
Nice solution, works very well :+1: 

After adding a changelog entry and reverting the temporary changes, this should be good to go.

PR #6666, opened 2022-11-29
Add second (non-admin) user to "initial data" trigger

philippotto:
Testing permission-related functionality often requires a user which doesn't have admin rights. Setting that user up is relatively tedious (especially on dev instances) and I don't see a reason why there shouldn't be a second default user by default (of course, only when `sampleOrganization == true`). Since the same password is used for both default users, one only needs to adapt one character in the email to log into the other account.

@fm3 What do you think about this? The config doesn't read as elegant as before, but I think the gains are worth it. Configuring another password for the second user didn't seem worth it to me, on the other hand.

- https://___.webknossos.xyz

### Steps to test:
- log in with `sample@scm.io`:`secret`
- log in with `sample2@scm.io`:`secret`

------
- [X] Ready for review

### DIFF
--- a/app/controllers/InitialDataController.scala
+++ b/app/controllers/InitialDataController.scala
@@ -53,6 +53,7 @@ class InitialDataService @Inject()(userService: UserService,
   implicit val ctx: GlobalAccessContext.type = GlobalAccessContext
 
   private val defaultUserEmail = conf.WebKnossos.SampleOrganization.User.email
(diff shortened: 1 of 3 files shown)


fm3 (commented):
Sure, makes sense.
Two suggestions
 - Maybe use a different full name for this user so that the initials wk displays at the top-right are discernible from those of Sample User.
 - Maybe `insertDefaultUser` can be parameterized to avoid duplicating basically the whole method?

PR #6668, opened 2022-11-30
Fix import of N5 datasets by adding n5 schema to frontend validation

philippotto:
- https://fixn5schema.webknossos.xyz

### Steps to test:
- try to import `s3://janelia-cosem-datasets/jrc_hela-3/jrc_hela-3.n5/em/fibsem-uint16`
- you should be able to see data after import (might be a bit slow)

------
- [X] Ready for review


frontend/javascripts/types/schemas/datasource.schema.ts
```diff
@@ -115,6 +115,56 @@ export default {
...
+                properties: {
+                  user: { type: "string" },
+                  password: { type: "string" },
+                },
+                required: ["user", "password"],
+              },
+              axisOrder: {
+                type: "object",
+                additionalProperties: { type: "number" },
+              },
```
daniel-wer, line 160: What does `axisOrder` look like if it is supplied?

daniel-wer (approved):
LGTM with one question :+1:

frontend/javascripts/types/schemas/datasource.schema.ts
philippotto: I think, it's something like `{x: 0, y: 1, z: 2}` (should be the same as for zarr).

PR #6670, opened 2022-11-30
Improves mag and voxelSize inferral for remote datasets

### LINKED ISSUE #6667: Allow compatible OME-NGFF multiscales to be imported
## Context
OME-NGFF datasets can have raw data and segmentations (labels) attached to them. Support for labels was added in #6638. However, when a labels layer does not contain all entries of the resolution pyramid, import doesn't work.
Example:
- https://s3.embl.de/i2k-2020/platy-raw.ome.zarr, voxel sizes: `10,10,25   20,20,25   40,40,50   80,80,100   160,160,200   …`
- https://s3.embl.de/i2k-2020/platy-raw.ome.zarr/labels/cells, voxel sizes: `0,80,100   160,160,200   …`

## Steps to Reproduce the bug
Try and import dataset: https://s3.embl.de/i2k-2020/platy-raw.ome.zarr

normanrz:
### Steps to test:
- Import https://s3.embl.de/i2k-2020/platy-raw.ome.zarr
- You should be able to see both the raw and segmentation data

### Issues:
- fixes #6667 

------
- [x] Ready for review

### DIFF
--- a/app/models/binary/explore/ExploreRemoteLayerService.scala
+++ b/app/models/binary/explore/ExploreRemoteLayerService.scala
@@ -88,7 +88,7 @@ class ExploreRemoteLayerService @Inject()() extends FoxImplicits with LazyLoggin
 
   private def rescaleLayersByCommonVoxelSize(layersWithVoxelSizes: List[(DataLayer, Vec3Double)])(
       implicit ec: ExecutionContext): Fox[(List[DataLayer], Vec3Double)] = {
-    val allVoxelSizes: Set[Vec3Double] = layersWithVoxelSizes
+    val allVoxelSizes = layersWithVoxelSizes
       .flatMap(layerWithVoxelSize => {
         val layer = layerWithVoxelSize._1
         val voxelSize = layerWithVoxelSize._2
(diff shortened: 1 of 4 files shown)


app/models/binary/explore/ExploreRemoteLayerService.scala
```diff
@@ -66,11 +70,62 @@ class ExploreRemoteLayerService @Inject()() extends FoxImplicits with LazyLoggin
...
+                   boundingBox = l.boundingBox * magFactors)
+          case l: ZarrSegmentationLayer =>
+            l.copy(mags = l.mags.map(mag => mag.copy(mag = mag.mag * magFactors)),
+                   boundingBox = l.boundingBox * magFactors)
+          case l: N5DataLayer =>
+            l.copy(mags = l.mags.map(mag => mag.copy(mag = mag.mag * magFactors)),
+                   boundingBox = l.boundingBox * magFactors)
+          case l: N5SegmentationLayer =>
+            l.copy(mags = l.mags.map(mag => mag.copy(mag = mag.mag * magFactors)),
+                   boundingBox = l.boundingBox * magFactors)
```
frcroth, line 123: As I understand it this could be simplified by using | to match multiple cases. Although I think this would require a new trait common to datasets (since DataSet/DataSetLike does not have e.g. mags)

frcroth (approved):

app/models/binary/explore/ExploreRemoteLayerService.scala
normanrz: @fm3 Do you have a good idea how this could be improved? Using a common trait is difficult, because the code relies on the `copy` method, which is only defined for case classes.

Issue #6669, opened 2022-11-30
[Follow-up] Folder Enhancements


philippotto:
Bugs:
- [x] switching to another orga causes error-toast to appear because folder id from url is loaded
- [x] dragged row has full opacity on some browsers / OS ? makes it hard to see where it's being dropped over. use `setDragImage` to generate a small drag image so that the wide row doesn't block other elements anymore.
- [x] right sidebar doesn't mention limiting of search results when stating total amount (should do the same as the headline)
- [x] cancelling new-folder input still creates folder
- ~[ ] refresh-datasets-button doesn't work properly sometimes? still showed datasets which were moved in the meantime~
  - cannot reproduce :thinking: 
- [X] using context menu on not-selected folder should not select it
- [X] deleting active folder should select another folder

Improvements:
- [x] clearer display of "default" permissions (tag can be confused with actual team)

Features
- [x] more prominent "add folder" button
- [x] target folder when adding dataset should be configurable
  - [x] clicking "add dataset" in folder X should import the dataset into folder X
- [x] multi-select for datasets (e.g., for dragging datasets into a folder)
- [x] show path / breadcrumb for datasets when searching
  - e.g., instead of datastore as a tag
- [x] ability to limit search to a folder
- [ ] #6883

Cleanup:
- [x] remove "experimental" label
- [x] https://github.com/scalableminds/webknossos/issues/6654 (grace period?)

philippotto:
> target folder when adding dataset should be configurable

@fm3 The back-end doesn't support this yet, right? For the upload route, the folder id might be passed to `WKRemoteDataStoreController.reserveDataSetUpload` and for zarr/n5 to `DataSourceController.add`?

fm3:
Correct, we can add that to those routes.

philippotto:
I extracted the remaining part into https://github.com/scalableminds/webknossos/issues/6883.

PR #6674, opened 2022-12-01
Fix and improve miscellaneous things in Folders tab

philippotto:
Quoting from #6669

> Bugs:
> - [x] switching to another orga causes error-toast to appear because folder id from url is loaded
> - [x] dragged row has full opacity on some browsers / OS ? makes it hard to see where it's being dropped over. use `setDragImage` to generate a small drag image so that the wide row doesn't block other elements anymore.
> - [x] right sidebar doesn't mention limiting of search results when stating total amount (should do the same as the headline)
> - [x] cancelling new-folder input still creates folder
> - [X] using context menu on not-selected folder should not select it
> - [X] deleting active folder should select another folder
> 
> Improvements:
> - [x] clearer display of "default" permissions (tag can be confused with actual team)

- https://foldersfollowup.webknossos.xyz

### Steps to test:
- I tested all of the described things. I think, there's no need to test the time-consuming things again (e.g., orga switching or search result limit). However, the other things, such as dragging a dataset should be easy to test :)

### Issues:
- contributes to #6669

------
- [x] Ready for review


daniel-wer (approved):
Very nice, everything works as described, LGTM :+1: 

Great improvement for the drag preview, in particular!

PR #6677, opened 2022-12-02
Implement folder specific search for datasets

philippotto:
- folder-specific search with optional recursion
- folder path displayed as tag

image
image

- https://folderspecificsearch.webknossos.xyz

### Steps to test:
- create some folders to  have a folder hierarchy
- move some datasets
- use the global search to find a dataset
- customize the search to only  search a specific folder and check that the search-subfolders option works
- check that the paths next to the dataset names look good
- switching to another folder while the search is active should change the folder that is searched (earlier versions cleared the search instead)
- if the global search is enabled, there shouldn't be an active folder. activating a folder in that case, should switch automatically to the folder-specific search.
- when going from global search to folder-specific search, the most recently active folder should be activated again (the same goes for clearing the search box)

### Issues:
- contributes to #6669

------
- [x] Ready for review


frontend/javascripts/dashboard/dataset/queries.tsx
```diff
@@ -498,3 +518,58 @@ function getUnobtrusivelyUpdatedDatasets(
...
   });
 }
+
+type FolderHierarchy = {
+  tree: FolderItem[];
+  itemById: Record<string, FolderItem>;
+  flatItems: FlatFolderTreeItem[];
+};
+
+export function getFolderHierarchy(folderTree: FlatFolderTreeItem[]): FolderHierarchy {
```
philippotto, line 528: this function was basically moved from `frontend/javascripts/dashboard/folders/folder_tree.tsx`.

frontend/javascripts/dashboard/dataset/queries.tsx
```diff
@@ -60,8 +65,13 @@ export function useDatasetSearchQuery(query: string | null) {
   );
 }
 
-export function useFolderTreeQuery() {
-  return useQuery(["folders"], getFolderTree, {
+async function fetchTreeHierarchy() {
+  const flatTreeItems = await getFolderTree();
+  return getFolderHierarchy(flatTreeItems);
```
philippotto, line 70: The query stores the hierarchical data instead of the flat backend data now. This is also why the (optimistic) updates in the queries had to be changed (they were only adapted to use the new helper `transformHierarchy` so that the old update code can be reused 1:1).

app/controllers/DataSetController.scala
```diff
@@ -178,6 +178,10 @@ class DataSetController @Inject()(userService: UserService,
       uploaderId: Option[String],
       @ApiParam(value = "Optional filtering: List only datasets in the folder with this id")
       folderId: Option[String],
+      @ApiParam(
+        value =
+          "Optional filtering: If a folderId was specified, this parameter controls whether subfolders should be considered, too (default: false)")
+      recursive: Option[Boolean],
```
fm3, line 184: The name `recursive` would be fine if the entire rout would be about searching in folders. But since it is broader, and the arguments can be combined in different combinations, I’d prefer something more explanatory. Hmm, not sure, how about “alsoSearchSubfolders”?

app/models/binary/DataSet.scala
```diff
@@ -163,11 +163,18 @@ class DataSetDAO @Inject()(sqlClient: SQLClient,
...
+  def findAllWithSearch(folderIdOpt: Option[ObjectId], searchQuery: Option[String], recursive: Boolean = false)(
       implicit ctx: DBAccessContext): Fox[List[DataSet]] =
     for {
       accessQuery <- readAccessQuery
-      folderPredicate = folderIdOpt.map(folderId => s"_folder = '$folderId'").getOrElse("true")
+      folderPredicate = folderIdOpt
+        .map(
+          folderId =>
+            if (recursive)
+              s"_folder in (select _descendant FROM webknossos.folder_paths fp WHERE fp._ancestor = '$folderId')"
```
fm3, line 174:
```suggestion
              s"_folder IN (SELECT _descendant FROM webknossos.folder_paths fp WHERE fp._ancestor = '$folderId')"
```

app/models/binary/DataSet.scala
```diff
@@ -163,11 +163,18 @@ class DataSetDAO @Inject()(sqlClient: SQLClient,
...
-  def findAllWithSearch(folderIdOpt: Option[ObjectId], searchQuery: Option[String])(
+  def findAllWithSearch(folderIdOpt: Option[ObjectId], searchQuery: Option[String], recursive: Boolean = false)(
       implicit ctx: DBAccessContext): Fox[List[DataSet]] =
     for {
       accessQuery <- readAccessQuery
-      folderPredicate = folderIdOpt.map(folderId => s"_folder = '$folderId'").getOrElse("true")
+      folderPredicate = folderIdOpt
+        .map(
+          folderId =>
+            if (recursive)
```
fm3, line 173: Since there are three possible code paths here now, I’d replace the map/if/else/getOrElse with a match expression

```suggestion
      folderPredicate = folderIdOpt match {
        case Some(folderId) if (recursive) =>
```

and then `case Some(folderId) => … `  and then `case None => "true"`

fm3 (commented):
Cool stuff :) I added some suggestions, only concerning code style in the backend.

daniel-wer (commented):
Works beautifully! Not much to add from my side :)

app/controllers/DataSetController.scala
philippotto: Hmm, I think the route param can also be used without searching. For example, one could list all datasets of a folder recursively by providing folderId, recursive=true and no search query. I think that would work (even though it's not used that way). Would you still prefer another name?

fm3: Yes, but my suggested name might not be the best option. How about “alsoIncludeSubfolders”? or just “includeSubfolders”? My concern is that the bare “recursive” is not going to be associated with the whole folder concept at all

frontend/javascripts/dashboard/folders/folder_tree.tsx
```diff
@@ -58,12 +58,10 @@ export function FolderTreeSidebar({
...
-      // No search is active and no folder is selected. For example, this happens
-      // after clearing the search box.
-      // Activate the root folder.
-      if (treeData.length > 0) {
-        context.setActiveFolderId(treeData[0].key);
-      }
+      // No search is active and no folder is selected. For example, this can happen
+      // after clearing the search box (when the search was global).
+      // Activate the most recently used folder or the root folder.
+      context.setActiveFolderId(context.mostRecentlyUsedActiveFolderId || treeData[0].key);
```
daniel-wer, line 64: Great :+1:

frontend/javascripts/libs/react_hooks.ts
```diff
@@ -2,13 +2,15 @@ import { useState, useEffect, useRef } from "react";
 import { useLocation } from "react-router-dom";
 
 // Adapted from: https://usehooks.com/usePrevious/
-export function usePrevious<T>(value: T): T | null | undefined {
+export function usePrevious<T>(value: T, ignoreFalsyValues: boolean = false): T | null {
```
daniel-wer, line 5: Slightly pedantic, it's not ignoring falsy values, but only null and undefined, isn't it? Not sure what to call those :see_no_evil: Maybe `ignoreNullValues`, although it's also not entirely correct :shrug:

daniel-wer (commented):
Great improvements! :1st_place_medal:  I'll test once the CI is ready.

frontend/javascripts/libs/react_hooks.ts
philippotto: Of course, you are right. I think the right term is "nullable" :) I'll rename it!

app/controllers/DataSetController.scala
philippotto: Ah, I see. I like `includeSubfolders` :+1:

frontend/javascripts/libs/react_hooks.ts
fm3: Is it really? :grimacing: I always read “nullable” as “can be null _or_ have a value” (similar to an option type), while in this context it reads more like “is actually null (or undefined)“. I may be wrong, though

daniel-wer: :see_no_evil: One could name it explicitly,`ignoreNullAndUndefined`.

daniel-wer:
Currently, the deployed dev branch fails to render with `TypeError: Cannot read properties of undefined (reading 'key') at folder_tree.tsx:64:87`. I think there needs to be a check whether `treeData` is empty.

frontend/javascripts/libs/react_hooks.ts
philippotto: You're right. I got a bit confused reading https://www.typescriptlang.org/docs/handbook/advanced-types.html#nullable-types which doesn't really define what a nullable type is, but @fm3's take is how I actually see it, too. I renamed it to `ignoreNullAndUndefined` :)

philippotto:
> Currently, the deployed dev branch fails to render with `TypeError: Cannot read properties of undefined (reading 'key') at folder_tree.tsx:64:87`. I think there needs to be a check whether `treeData` is empty.

Fixed now :)

> Edit: Ironically, this is the line I commented with "Great +1" sweat_smile

:rofl: I relied too much on TS and didn't test the last code change obviously.

daniel-wer (approved):
It's a pleasure to use this feature, especially with the newest improvements! :+1:

PR #6679, opened 2022-12-05
Fix s3fs region access, s3 url styles

fm3:
Two fixes for s3fs
 - Set `withForceGlobalBucketAccessEnabled` to true, since we do not know in advance in which s3 region the relevant buckets will be stored.
 - Fix improper handling of non-standard URI formats. Now all three variants can be accepted:
```
s3://webknossos-zarr/demodata/l4_sample_zarr/color.zarr/1/
s3://webknossos-zarr.s3.amazonaws.com/demodata/l4_sample_zarr/color.zarr/1/
s3://s3.amazonaws.com/webknossos-zarr/demodata/l4_sample_zarr/color.zarr/1/
```
(Note that these examples need credentials, talk to us for testing)

Note on Implementation: The FileSystem itself now holds bucket information IF the bucket is in the base uri (as is the case in the first two examples above). When getPath is called on that fileSystem, the bucket is prepended to the path arguments.

Note that non-standard endpoints are no longer supported due to this change. The endpoint is now always s3.amazon.com (with default region us-east-1 but globalBucketAccess). We never encountered non-standard endpoints so far anyway and this case was not properly tested before either. When we encounter them, we can still try to re-add it without breaking this existing logic.

I did not clean up the s3fs java code. We could consider converting it to scala, rewriting it with our style, and remove the parts we do not use. However, I’d suggest we wait with this since there have been discussions about a new s3fs fork with active support, which may replace our package in the future.

### Steps to test:
- Test importing, and loading data from, the above sample zarr dataset URIs

------
- [x] Ready for review


frcroth (approved):
Very nice! I haven't look to much at the Java code but it seems to work

PR #6680, opened 2022-12-05
WIP: Add telemetry and health status for InternalError sigbus

### LINKED ISSUE #6650: Catch SIGBUS/java.lang.InternalError and post to telemetry, trigger restart?
We got a bunch of `java.lang.InternalError: a fault occurred in a recent unsafe memory access operation in compiled Java code` in a short period of time in a datastore, combined with corrupted data loading.

This should be detected and sent to the telemetry channel, and possibly set the health check to false for some time to trigger a restart in autoheal environments.

I think it would make the most sense to add a catch block at some level during data loading. `WKWBucketProvider.loadFromUnderlying` was often in the stack.

frcroth:
- https://___.webknossos.xyz

### Steps to test:
- abc

### Issues:
- fixes #6650 

------
- [ ] Ready for review

### DIFF
--- a/webknossos-datastore/app/com/scalableminds/webknossos/datastore/DataStoreModule.scala
+++ b/webknossos-datastore/app/com/scalableminds/webknossos/datastore/DataStoreModule.scala
@@ -21,5 +21,6 @@ class DataStoreModule extends AbstractModule {
     bind(classOf[MappingService]).asEagerSingleton()
     bind(classOf[AgglomerateService]).asEagerSingleton()
     bind(classOf[IsosurfaceServiceHolder]).asEagerSingleton()
+    bind(classOf[ApplicationHealthService]).asEagerSingleton()
   }
 }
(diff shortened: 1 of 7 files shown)


frcroth:
Is this roughly what you had in mind @fm3 ? I don't yet know how to pass the DSSlackNotificationService to the WKWBucketProvider (because it is created in WKWDataLayers).

fm3:
Yes, this is gonig in the right direction :)
 - Could you test throwing the exception inside wkw data loading to simulate the error? Does the exception bubble all the way up to requestData in the BinaryDataController? In this case, we wouldn’t have to worry about connecting the BucktProviders themselves, which is a bit harder, since they exist per layer and are not injected. Otherwise, maybe another point in the stack is a good entry point for this.
 - I would suggest adding a new healthService that is then injected into the place where the exception is caught, and also into the Application controller, where the health request is answered. It can have a variable storing last occurences of the error with timestamp. Then, the health api can report false if the error occurred within the last, say, 30 seconds. This new service could be marked asEagerSingleton in the DatastoreModule.

I hope this makes sense, feel free to ping me again if you run into more questions :)

frcroth:
Closed in favor of #6705

PR #6683, opened 2022-12-06
Implement multi selection of datasets in folder tab

philippotto:
- https://multiselect.webknossos.xyz

### Steps to test:
- play around with selecting single datasets and multiple ones (by pressing CTRL)
- test dragging datasets into another folder
- the ctrl behavior is inspired by the way the gnome file explorer works
- I tested the following cases:
```
Input state: none selected
	user action: click on A:
		output state: A is selected
			
	user action: ctrl click on A:
		output state: A is selected
			
	user action: drag on A:
		output state: A is selected
			
Input state: A selected
	user action: click on A
		output state: none is selected

	user action: ctrl click on A
		output state: none is selected

	user action: drag A
		output state: A is selected
		
	user action: click on B
		output state: B is selected
		
	user action: ctrl click on B
		output state: A + B is selected

	user action: drag B
		output state: B is selected
		
Input state: A + B selected
	user action: click on A
		output state: A is selected
		
	user action: ctrl click on A
		output state: B is selected

	user action: drag A
		output state: A + B is selected

	user action: click on C
		output state: should select C		

	user action: ctrl click on C
		output state: A, B, C are selected

	user action: drag C
		output state: C is selected
```
- range select via shift is a bit simpler. I'm simply extending the current selection from the closest selected record. This differs from nautilus a bit, but only for special cases (e.g., there's already a range or scattered) and I think, this behaves quite intuitive for the most common cases.

### Issues:
- contributes to #6669

------
- [X] Ready for review


philippotto:
@daniel-wer If you can ignore 925a481e0df8f4ed51abb88d4acb3f257d98783a during review (e.g., by reviewing the commits before and after that commit separately), the diff should be easier to grasp :see_no_evil:

philippotto:
@daniel-wer The branch is now merge-conflict free and also includes the range selection via shift+click. Feel free to review and test now :)

daniel-wer (commented):
Code LGTM, this will be very helpful to a lot of users!

Notes from my testing:
- [x] The shift + left click selection behaves weirdly inside a folder (even after a reload). See this gif:
strange-shift-selection-in-folder
- [ ] Selecting datasets with ctrl + left click rather often does not work. The same sometimes happens even without the ctrl modifier. I click into the large white space in the tags column. Ah, I think the culprit might be this overly large div in the tags column (that I seem to hit pretty accurately :sweat_smile:):
image
- [ ] When selecting multiple datasets, the context menu is a little misleading as it only executes the action for the dataset that was right-clicked on, although one could expect it to be executed for all selected datasets (e.g. "Reload"). This is also what happens in my file explorer when having selected multiple files. Maybe the context menu should be disabled for the multi-selection case.

philippotto:
@daniel-wer Thank you for your testing feedback :pray: I incorporated everything now. Please have another look :)

daniel-wer (approved):
Works very well! :+1: 

I found one more minor issue which could be fixed:
- [x] When moving datasets into the same folder they already are in, the progress modal and success toast still appear. This should be ignored or one could show a warning toast. I'm leaning towards the warning toast to notify the user that the last action didn't make any sense and was probably unintended - but no hard opinion.

daniel-wer:
:ship: it!

PR #6686, opened 2022-12-07
Unify timestamp representation with Instant

### LINKED ISSUE #6456: In backend, unify timestamp data types
I think I’ve seen java.time.Instant, java.time.LocalDateTime, Long, java.sql.Timestamp and org.joda.time.DateTime for timestamps. We should decide which one works best and stick to it (possibly adding json writes/reads and postgres de/serialization).
Looks like duration already pretty consistently uses scala.concurrent.duration. Edit: ok now I also saw java.time.Duration…

fm3:
Introducing Instant, a case class representing a point in time in UTC. It replaces our mixed usages of Long, java.time.Instant, java.time.LocalDateTime, org.joda.time.DateTime and java.sql.Timestamp. I probably did not catch all usages of Long, but those can be replaced step by step. I searched for the others and they remain only where really needed.

I first wanted to go with java.time.Instant everywhere, but unfortunately backwards compatibility forbids it, since it serializes as ISO time string, while our API expects epochMillis numbers in jsons. Adding our own case class has a couple of other benefits, though, like nice integration with FiniteDuration.

Note that voxelytics reports expected an ISO time string, so to unify this with the rest of our API, i changed the types in the frontend there, to also expect number.

- https://instant.webknossos.xyz

### Steps to test:
- Test task annotating, time tracking report (staticstics view + time tracking view)
- Test Voxelytics Reporting (both running a workflow and then viewing its report)
- Test around different functionalities, annotating, worker job, edit a task type, etc

### Issues:
- fixes #6456 

------
- [x] Ready for review

### DIFF
--- a/app/controllers/AnnotationController.scala
+++ b/app/controllers/AnnotationController.scala
@@ -68,7 +68,7 @@ class AnnotationController @Inject()(
     with FoxImplicits {
 
   implicit val timeout: Timeout = Timeout(5 seconds)
-  private val taskReopenAllowed = (conf.Features.taskReopenAllowed + (10 seconds)).toMillis
+  private val taskReopenAllowed = conf.Features.taskReopenAllowed + (10 seconds)
 
   @ApiOperation(value = "Information about an annotation, supplying the type explicitly",
                 nickname = "annotationInfoByType")
@@ -187,7 +187,7 @@ class AnnotationController @Inject()(
         isAdminOrTeamManager <- userService.isTeamManagerOrAdminOf(user, annotation._team)
         _ <- bool2Fox(annotation.state == AnnotationState.Finished) ?~> "annotation.reopen.notFinished"
         _ <- bool2Fox(isAdminOrTeamManager || annotation._user == user._id) ?~> "annotation.reopen.notAllowed"
-        _ <- bool2Fox(isAdminOrTeamManager || System.currentTimeMillis - annotation.modified < taskReopenAllowed) ?~> "annotation.reopen.tooLate"
+        _ <- bool2Fox(isAdminOrTeamManager || (annotation.modified + taskReopenAllowed).isPast) ?~> "annotation.reopen.tooLate"
       } yield ()
 
     for {
--- a/app/controllers/AnnotationController.scala
+++ b/app/controllers/AnnotationController.scala
@@ -4,6 +4,7 @@ import akka.util.Timeout
 import com.mohiva.play.silhouette.api.Silhouette
 import com.scalableminds.util.accesscontext.{DBAccessContext, GlobalAccessContext}
 import com.scalableminds.util.geometry.BoundingBox
+import com.scalableminds.util.time.Instant
 import com.scalableminds.util.tools.{Fox, FoxImplicits}
 import com.scalableminds.webknossos.datastore.models.annotation.AnnotationLayerType.AnnotationLayerType
 import com.scalableminds.webknossos.datastore.models.annotation.{AnnotationLayer, AnnotationLayerType}
@@ -98,7 +99,8 @@ class AnnotationController @Inject()(
           .publicWrites(annotation, request.identity, Some(restrictions)) ?~> "annotation.write.failed"
         _ <- Fox.runOptional(request.identity) { user =>
           if (typedTyp == AnnotationType.Task || typedTyp == AnnotationType.Explorational) {
-            timeSpanService.logUserInteraction(timestamp, user, annotation) // log time when a user starts working
+            timeSpanService
+              .logUserInteraction(Instant(timestamp), user, annotation) // log time when a user starts working
           } else Fox.successful(())
         }
         _ = Fox.runOptional(request.identity)(user => userDAO.updateLastActivity(user._id))
@@ -345,7 +347,7 @@ class AnnotationController @Inject()(
       } yield result
   }
 
-  private def finishAnnotation(typ: String, id: String, issuingUser: User, timestamp: Long)(
+  private def finishAnnotation(typ: String, id: String, issuingUser: User, timestamp: Instant)(
       implicit ctx: DBAccessContext): Fox[(Annotation, String)] =
     for {
       annotation <- provider.provideAnnotation(typ, id, issuingUser) ~> NOT_FOUND
@@ -360,7 +362,7 @@ class AnnotationController @Inject()(
     implicit request =>
       log() {
         for {
-          (updated, message) <- finishAnnotation(typ, id, request.identity, timestamp) ?~> "annotation.finish.failed"
+          (updated, message) <- finishAnnotation(typ, id, request.identity, Instant(timestamp)) ?~> "annotation.finish.failed"
           restrictions <- provider.restrictionsFor(typ, id)
           json <- annotationService.publicWrites(updated, Some(request.identity), Some(restrictions))
         } yield JsonOk(json, Messages(message))
@@ -373,7 +375,7 @@ class AnnotationController @Inject()(
       log() {
         withJsonAs[JsArray](request.body \ "annotations") { annotationIds =>
           val results = Fox.serialSequence(annotationIds.value.toList) { jsValue =>
-            jsValue.asOpt[String].toFox.flatMap(id => finishAnnotation(typ, id, request.identity, timestamp))
+            jsValue.asOpt[String].toFox.flatMap(id => finishAnnotation(typ, id, request.identity, Instant(timestamp)))
           }
 
           results.map { _ =>
--- a/app/controllers/MaintenanceController.scala
+++ b/app/controllers/MaintenanceController.scala
@@ -42,7 +42,7 @@ class MaintenanceController @Inject()(sil: Silhouette[WkEnv],
     for {
       multiUser <- multiUserDAO.findOne(request.identity._multiUser)
       _ <- bool2Fox(multiUser.isSuperUser)
-      _ <- maintenanceDAO.updateExpirationTime(Instant(0))
+      _ <- maintenanceDAO.updateExpirationTime(Instant.zero)
     } yield Ok("maintenance.close.success")
   }
 
--- a/app/controllers/MaintenanceController.scala
+++ b/app/controllers/MaintenanceController.scala
@@ -1,7 +1,9 @@
 package controllers
 
 import com.mohiva.play.silhouette.api.Silhouette
+import com.scalableminds.util.time.Instant
 import com.scalableminds.util.tools.{Fox, FoxImplicits}
+
 import javax.inject.Inject
 import models.user.MultiUserDAO
 import oxalis.security.WkEnv
@@ -9,6 +11,7 @@ import play.api.libs.json.Json
 import play.api.mvc.{Action, AnyContent}
 import slick.jdbc.PostgresProfile.api._
 import utils.{SQLClient, SimpleSQLDAO}
+import scala.concurrent.duration._
 
 import scala.concurrent.ExecutionContext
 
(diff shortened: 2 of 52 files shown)


app/controllers/AnnotationController.scala
```diff
@@ -187,7 +189,7 @@ class AnnotationController @Inject()(
         isAdminOrTeamManager <- userService.isTeamManagerOrAdminOf(user, annotation._team)
         _ <- bool2Fox(annotation.state == AnnotationState.Finished) ?~> "annotation.reopen.notFinished"
         _ <- bool2Fox(isAdminOrTeamManager || annotation._user == user._id) ?~> "annotation.reopen.notAllowed"
-        _ <- bool2Fox(isAdminOrTeamManager || System.currentTimeMillis - annotation.modified < taskReopenAllowed) ?~> "annotation.reopen.tooLate"
+        _ <- bool2Fox(isAdminOrTeamManager || (annotation.modified + taskReopenAllowed).isPast) ?~> "annotation.reopen.tooLate"
```
jstriebel, line 192: :heart:

app/controllers/MaintenanceController.scala
```diff
@@ -18,41 +21,43 @@ class MaintenanceController @Inject()(sil: Silhouette[WkEnv],
     extends Controller
     with FoxImplicits {
 
+  private val maintenanceDuration: FiniteDuration = 10 minutes
+
   def info: Action[AnyContent] = sil.UserAwareAction.async { implicit request =>
     for {
       expirationTime <- maintenanceDAO.getExpirationTime
-      isMaintenance = expirationTime.getTime >= System.currentTimeMillis
+      isMaintenance = !expirationTime.isPast
```
jstriebel, line 29: You could also add `isFuture`, but also fine as-is.

app/controllers/MaintenanceController.scala
```diff
@@ -18,41 +21,43 @@ class MaintenanceController @Inject()(sil: Silhouette[WkEnv],
...
     } yield Ok(Json.obj("isMaintenance" -> Json.toJson(isMaintenance)))
   }
 
   def initMaintenance: Action[AnyContent] = sil.SecuredAction.async { implicit request =>
     for {
       multiUser <- multiUserDAO.findOne(request.identity._multiUser)
       _ <- bool2Fox(multiUser.isSuperUser)
-      _ <- maintenanceDAO.updateExpirationTime(new java.sql.Timestamp(System.currentTimeMillis + 1000 * 60 * 10))
+      _ <- maintenanceDAO.updateExpirationTime(Instant.in(maintenanceDuration))
```
jstriebel, line 37: :heart:

app/controllers/MaintenanceController.scala
```diff
@@ -18,41 +21,43 @@ class MaintenanceController @Inject()(sil: Silhouette[WkEnv],
...
     } yield Ok("maintenance.init.success")
   }
 
   def closeMaintenance: Action[AnyContent] = sil.SecuredAction.async { implicit request =>
     for {
       multiUser <- multiUserDAO.findOne(request.identity._multiUser)
       _ <- bool2Fox(multiUser.isSuperUser)
-      _ <- maintenanceDAO.updateExpirationTime(new java.sql.Timestamp(0))
+      _ <- maintenanceDAO.updateExpirationTime(Instant(0))
```
jstriebel, line 45: Maybe `Instant.min`, similar to `max`?

app/controllers/OrganizationController.scala
```diff
@@ -70,7 +70,7 @@ class OrganizationController @Inject()(organizationDAO: OrganizationDAO,
     implicit val ctx: DBAccessContext = GlobalAccessContext
     for {
       invite <- inviteDAO.findOneByTokenValue(inviteToken)
-      _ <- bool2Fox(invite.expirationDateTime.isAfterNow)
+      _ <- bool2Fox(!invite.expirationDateTime.isPast)
```
jstriebel, line 73: Same as above, maybe use `isFuture`?

app/controllers/TimeController.scala
```diff
@@ -104,8 +105,8 @@ class TimeController @Inject()(userService: UserService,
     for {
       userJs <- userService.compactWrites(user)
       timeJs <- timeSpanDAO.findAllByUserWithTask(user._id,
-                                                  Some(startDate.getTimeInMillis),
-                                                  Some(endDate.getTimeInMillis))
+                                                  Some(Instant(startDate.getTimeInMillis)),
+                                                  Some(Instant(endDate.getTimeInMillis)))
```
jstriebel, line 109: Maybe add another conversion to `Instant` itself, similar to the other ones that are already in place?

app/models/annotation/nml/NmlWriter.scala
```diff
@@ -329,7 +329,7 @@ class NmlWriter @Inject()(implicit ec: ExecutionContext) extends FoxImplicits {
     }
     Xml.withinElementSync("meta") {
       writer.writeAttribute("name", "timestamp")
-      writer.writeAttribute("content", DateTime.now().getMillis.toString)
+      writer.writeAttribute("content", Instant.now.epochMillis.toString)
```
jstriebel, line 332: I've seen this twice now, maybe add another method on Instant for this? (Not sure if it comes another time, if not, also fine as-is.)

app/models/binary/DataSetService.scala
```diff
@@ -257,13 +258,13 @@ class DataSetService @Inject()(organizationDAO: OrganizationDAO,
...
-  def lastUsedTimeFor(_dataSet: ObjectId, userOpt: Option[User]): Fox[Long] =
+  def lastUsedTimeFor(_dataSet: ObjectId, userOpt: Option[User]): Fox[Instant] =
     userOpt match {
       case Some(user) =>
         (for {
           lastUsedTime <- dataSetLastUsedTimesDAO.findForDataSetAndUser(_dataSet, user._id).futureBox
-        } yield lastUsedTime.toOption.getOrElse(0L)).toFox
-      case _ => Fox.successful(0L)
+        } yield lastUsedTime.toOption.getOrElse(Instant(0L))).toFox
+      case _ => Fox.successful(Instant(0L))
```
jstriebel, line 267: Again, maybe `Instant.min`?

app/models/binary/Publication.scala
```diff
@@ -87,8 +89,6 @@ class PublicationDAO @Inject()(sqlClient: SQLClient)(implicit ec: ExecutionConte
     for {
       _ <- run(
         sqlu"""insert into webknossos.publications(_id, publicationDate, imageUrl, title, description, created, isDeleted)
-                         values(${p._id.id}, ${p.publicationDate
-          .map(new java.sql.Timestamp(_))}, ${p.imageUrl}, ${p.title}, ${p.description}, ${new java.sql.Timestamp(
-          p.created)}, ${p.isDeleted})""")
+                         values(${p._id.id}, ${p.publicationDate}, ${p.imageUrl}, ${p.title}, ${p.description}, ${p.created}, ${p.isDeleted})""")
```
jstriebel, line 92: I assume the previous `map` is not needed anymore since publicationDate is an Option and can be deserialized recursively (e.g. first the Option deserialization, if there's a value the Instant is deserialized), is this correct?

app/models/user/time/TimeSpan.scala
```diff
@@ -63,43 +59,43 @@ class TimeSpanDAO @Inject()(sqlClient: SQLClient)(implicit ec: ExecutionContext)
...
+  def findAllByUser(userId: ObjectId, start: Option[Instant], end: Option[Instant]): Fox[List[TimeSpan]] =
     for {
       r <- run(
         Timespans
-          .filter(r =>
-            notdel(r) && r._User === userId.id && (r.created >= new java.sql.Timestamp(start.getOrElse(0))) && r.created <= new java.sql.Timestamp(
-              end.getOrElse(MAX_TIMESTAMP)))
+          .filter(
+            r =>
+              notdel(r) && r._User === userId.id && (r.created >= start.getOrElse(Instant(0)).toSql) && r.created <= end
```
jstriebel, line 74: Instant.min?

app/models/user/time/TimeSpan.scala
```diff
@@ -124,24 +120,24 @@ class TimeSpanDAO @Inject()(sqlClient: SQLClient)(implicit ec: ExecutionContext)
...
+  def findAllByAnnotation(annotationId: ObjectId, start: Option[Instant], end: Option[Instant]): Fox[List[TimeSpan]] =
     for {
       r <- run(
         Timespans
           .filter(
             r =>
-              notdel(r) && r._Annotation === annotationId.id && (r.created >= new java.sql.Timestamp(
-                start.getOrElse(0))) && r.created <= new java.sql.Timestamp(end.getOrElse(MAX_TIMESTAMP)))
+              notdel(r) && r._Annotation === annotationId.id && (r.created >= start
+                .getOrElse(Instant(0))
```
jstriebel, line 130: The formatting seems a bit weird to me here, since `.getOrElse` is only about `start`. But no strong opinion.

app/models/voxelytics/ElasticsearchClient.scala
```diff
@@ -68,16 +68,16 @@ class ElasticsearchClient @Inject()(wkConf: WkConf, rpc: RPC, val system: ActorS
     for {
       _ <- bool2Fox(enabled) ?~> "Elasticsearch is not enabled."
       _ = logger.info("Waiting for Elasticsearch to become available.")
-      _ <- pollUntilServerStartedUp(LocalDateTime.now.plus(Duration.ofMillis(conf.startupTimeout.toMillis))) ~> 500
+      _ <- pollUntilServerStartedUp(Instant.now + conf.startupTimeout) ~> 500
```
jstriebel, line 71:
```suggestion
      _ <- pollUntilServerStartedUp(Instant.in(conf.startupTimeout)) ~> 500
```

jstriebel (commented):
Awesome stuff! :tada: This is so much better :heart: 
Left a couple of comment, but nothing major.

app/models/binary/Publication.scala
fm3: Almost correct! It’s not exactly recursively, but SQLTypeImplicits now has `SetInstantOpt` as well as `SetInstant`

app/models/user/time/TimeSpan.scala
fm3: Yeah, that’s what our formatter does here. I think mid-term, we’d want to replace this kind of query with the sql syntax ones, then we don’t have this problem anymore

app/controllers/MaintenanceController.scala
fm3: I decided against isFuture for the moment. Might add it later.

fm3:
I also now tested that the voxelytics reporting still works :) I think if you could have a quick look at the frontend code changes this should be good to go  :) @philippotto maybe you have a minute to do that?

frontend/javascripts/admin/voxelytics/workflow_list_view.tsx
```diff
@@ -17,28 +17,26 @@ import { VX_POLLING_INTERVAL } from "./workflow_view";
 function parseTaskInfo(taskInfo: VoxelyticsTaskInfo): VoxelyticsTaskInfo {
   return {
     ...taskInfo,
-    beginTime: taskInfo.beginTime != null ? new Date(taskInfo.beginTime) : null,
-    endTime: taskInfo.endTime != null ? new Date(taskInfo.endTime) : null,
+    beginTime: taskInfo.beginTime != null ? taskInfo.beginTime : null,
+    endTime: taskInfo.endTime != null ? taskInfo.endTime : null,
```
philippotto, line 21:
```suggestion
```

See my other comment for an explanation. Also here the short form if deleting doesnt work: `taskInfo.beginTime ?? null`

frontend/javascripts/admin/voxelytics/workflow_list_view.tsx
```diff
@@ -17,28 +17,26 @@ import { VX_POLLING_INTERVAL } from "./workflow_view";
...
 }
 
 function parseRunInfo(runInfo: VoxelyticsRunInfo): VoxelyticsRunInfo {
   return {
     ...runInfo,
     tasks: runInfo.tasks.map(parseTaskInfo),
-    beginTime: new Date(runInfo.beginTime),
-    endTime: runInfo.endTime != null ? new Date(runInfo.endTime) : null,
+    beginTime: runInfo.beginTime,
+    endTime: runInfo.endTime != null ? runInfo.endTime : null,
```
philippotto, line 30: same here

jstriebel (approved):
LGTM :tada:

PR #6689, opened 2022-12-09
fixes infinite loop in getLineCount

normanrz:
Fixes a bug in `getLineCount` that would result in an infinite loop.

### Steps to test:
- Call `getLineCount('DEBUG    voxelytics.commons.workflow.resumable_executor.resumable_executor Looking for checkpoint: artifacts/xxxxxxxx-xxxx-xx-xx-xxxxx/compute_segment_stats/segment_stats__xxxxxxxxxx/resumable_executions/main/step/xxxxxx,xxxxxx,xxxx,xxxx,xx,xxx/checkpoint. Exists: True', 110)` (should result in `4`)
------
- [x] Ready for review


daniel-wer (approved):
Thank you, LGTM! :+1:

Issue #6688, opened 2022-12-09
webknossos can't send emails


thopp-tudelft:
## Context

Whenever the instance of the webknossos docker image version 22.12.0 tries to send an email the following message is placed in the log:
```
2022-12-09 11:12:25,503 [ERROR] akka.actor.OneForOneStrategy - Cannot find valid hostname for mail session
org.apache.commons.mail.EmailException: Cannot find valid hostname for mail session
	at org.apache.commons.mail.Email.getMailSession(Email.java:647)
	at oxalis.mail.Mailer.oxalis$mail$Mailer$$send(Mailer.scala:59)
	at oxalis.mail.Mailer$$anonfun$receive$1.applyOrElse(Mailer.scala:31)
	at akka.actor.Actor.aroundReceive(Actor.scala:537)
	at akka.actor.Actor.aroundReceive$(Actor.scala:535)
	at oxalis.mail.Mailer.aroundReceive(Mailer.scala:27)
	at akka.actor.ActorCell.receiveMessage(ActorCell.scala:580)
	at akka.actor.ActorCell.invoke(ActorCell.scala:548)
	at akka.dispatch.Mailbox.processMailbox(Mailbox.scala:270)
	at akka.dispatch.Mailbox.run(Mailbox.scala:231)
```
The documentation has no information about specific settings that need to be configured to have this work, nor is the error message very insightful.

## Expected Behavior

An email should be sent. Or a message that says there is a configuration error should appear in the logs instead of an exception.

## Current Behavior

Nothing, the interface gives a message to the user that an email will be sent but it never is.

## Steps to Reproduce the bug
1. Start the docker image with the default parameters from the documentation [here](https://docs.webknossos.org/webknossos/installation.html) with VIRTUAL_HOST set and the datastores etc...
2. Set it up etc...
3. Try to register an account
4. An email should be sent here to the admin but it isn't

## Your Environment for bug

- Browser name and version: all browsers
- Operating System and version: Ubuntu 20.04
- Version of WebKnossos (Release or Commit): 22.12.0

thopp-tudelft:
Based on reading the configuration [here](https://github.com/scalableminds/webknossos/blob/7b095666c889fd7506081f9c7787c79629c43a22/conf/application.conf#L173) I have added the following parameters to the docker container command: `-Dmail.smtp.host=host.docker.internal -Dmail.smtp.port=25 -Dmail.smtp.tls=false -Dmail.smtp.auth=false`

[...]

PR #6692, opened 2022-12-09
Respect mail.smtp.auth setting

fm3:
As reported in #6688 the wk setting mail.smtp.auth was not respected. I added an if-condition to only set the auth fields if this is enabled.
The PR also adds some logging if email adresses could not be set.

------
- [x] Ready for review


jstriebel (approved):
LGTM :+1:

PR #6695, opened 2022-12-12
Check that user does not upload NML in dataset upload view

### LINKED ISSUE #6694: In dataset upload view, auto-detect NMLs and warn/redirect the user to annotation upload
We saw a few more uploads of NML files in the dataset upload. I believe the frontend already runs some analysis on the files before uploading? Maybe it would be simple to auto-detect NMLs and notify the user that this is the wrong page for those?

Alternatively, we could unify the upload altogether, but that would certainly be more complex (and I’m not sure it would work better for users)

philippotto:
- https://___.webknossos.xyz

### Steps to test:
- enable jobs (otherwise, there will be a warning anyway saying that the NML file needs conversion)
- drag nml file into upload dataset view
- will look like this:

image

The submit button is apparently always enabled by antd even though the validation fails. I didn't find an easy way to improve this, but 1) couldn't find a quick solution and 2) it shouldn't be big problem and 3) it's been like this before.

### Issues:
- fixes #6694

------
- [X] Ready for review

### DIFF
--- a/frontend/javascripts/admin/dataset/dataset_upload_view.tsx
+++ b/frontend/javascripts/admin/dataset/dataset_upload_view.tsx
@@ -777,6 +777,14 @@ class DatasetUploadView extends React.Component<PropsWithFormAndRouter, State> {
                 {
                   validator: syncValidator(
                     (files: FileWithPath[]) =>
+                      files.filter((file) => Utils.isFileExtensionEqualTo(file.path, ["nml"]))
+                        .length === 0,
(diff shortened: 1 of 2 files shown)


fm3 (approved):
Thanks!
(btw: does this also catch volume annotation zips? I guess zip files containing at least one NML file could also trigger this, what do you think?)

philippotto:
> (btw: does this also catch volume annotation zips? I guess zip files containing at least one NML file could also trigger this, what do you think?)

Good point. I added a warning modal for that case (this was easier than using the default form validation and should be better than nothing).

PR #6698, opened 2022-12-14
add protected and private modifiers to DAO hierarchy

fm3:
Re-generating Tables.scala from a changed SQL schema led to a ton of unneeded recompilation.

This PR adds protected/private access modifiers in the DAO hierarchy to make everything that has stuff from Tables.scala in the signature protected/private in the DAO scope. This was implicitly already the case for most methods. The exception was parse, where two non-private usages existed in associated DAOs (SharedAnnotationsDAO, UserTeamRolesDAO). I merged these into the main DAOs AnnotationsDAO and UserDAO respectively to avoid having to make parse public.

- https://daoprivate.webknossos.xyz

### Steps to test:
- CI should be enough to ensure we didn’t break anything
- To see the speedup, add a new column to a table in schema.sql, execute yarn refresh-schema and recompile → Should not recompile over a hundred source files

------
- [x] Ready for review

### DIFF
--- a/app/WebKnossosModule.scala
+++ b/app/WebKnossosModule.scala
@@ -21,7 +21,6 @@ class WebKnossosModule extends AbstractModule {
     bind(classOf[UserService]).asEagerSingleton()
     bind(classOf[TaskService]).asEagerSingleton()
     bind(classOf[UserDAO]).asEagerSingleton()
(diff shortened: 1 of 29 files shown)


jstriebel (approved):
LGTM :+1:

fm3:
cc @frcroth (please do the same thing in the new DAO in #6646 when pulling in the master)

PR #6700, opened 2022-12-15
Fix crash in publication page and add error boundaries

philippotto:
- adds error boundaries to the individual routes (except for the tracing views since these have their own boundaries)
  - a "nice" error page will be shown to the user and the error will be logged to airbrake, too
- adds top-level error boundary for route-independent react errors
- fixes rendering error when a publication doesn't have any active datasets or annotations

- https://___.webknossos.xyz

### Steps to test:
- the publication fix is a bit tedious to set up. I don't think, it needs additional testing. It will look like this (a bit naked, but this should be only an edge case, anyway):
image
- the error boundaries can be tested by:
  - navigating to admin > users
  - image
  - clicking "help" in the navbar and "ask a question" (this error boundary is route independent)
image

### Issues:
- see https://scm.slack.com/archives/C5AKLAV0B/p1671099320485249 for context

------
- [X] Ready for review

### DIFF
--- a/frontend/javascripts/admin/user/user_list_view.tsx
+++ b/frontend/javascripts/admin/user/user_list_view.tsx
@@ -288,9 +288,6 @@ class UserListView extends React.PureComponent<Props, State> {
   };
 
   render() {
-    if (Math.random() > -1) {
-      throw new Error("Dummy error to test error boundary");
-    }
     const hasRowsSelected = this.state.selectedUserIds.length > 0;
     const rowSelection = {
       preserveSelectedRowKeys: true,
--- a/frontend/javascripts/admin/user/user_list_view.tsx
+++ b/frontend/javascripts/admin/user/user_list_view.tsx
@@ -288,6 +288,7 @@ class UserListView extends React.PureComponent<Props, State> {
   };
 
   render() {
+    throw new Error("Dummy error to test error boundary");
     const hasRowsSelected = this.state.selectedUserIds.length > 0;
     const rowSelection = {
       preserveSelectedRowKeys: true,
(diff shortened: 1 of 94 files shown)


frontend/javascripts/admin/user/user_list_view.tsx
```diff
@@ -288,6 +288,7 @@ class UserListView extends React.PureComponent<Props, State> {
   };
 
   render() {
+    throw new Error("Dummy error to test error boundary");
```
philippotto, line 291:
```suggestion
```

remove after testing/approval.

frontend/javascripts/navbar.tsx
```diff
@@ -316,6 +316,10 @@ function HelpSubMenu({
 } & SubMenuProps) {
   const [isHelpModalOpen, setIsHelpModalOpen] = useState(false);
 
+  if (isHelpModalOpen) {
+    throw new Error("Throwing dummy error because you clicked on Help > Ask a question.");
+  }
+
```
philippotto, line 322:
```suggestion
```

remove after testing/approval.

philippotto (commented):
review will be easier by ignoring whitespace changes

normanrz (approved):
code looks good. i'll do some testing now

normanrz:
Tested -> works. I was wondering if the toast is still required? Seems a bit redundant.

philippotto:
> Tested -> works. I was wondering if the toast is still required? Seems a bit redundant.

I assume you tested it locally, right? Orthogonal to the error boundaries, there is a global `window.onerror` handler (mainly for errors unrelated to react). Due to a peculiarity of react, that event handler is triggered again after an error boundary was activated, but only in DEV mode. See this [react issue](https://github.com/facebook/react/issues/10474). So, in the production build, there won't be the redundant error toast. I just tested it at https://fixpublicationpage.webknossos.xyz/users.

PR #6704, opened 2022-12-19
Easier folder creation and configurable target folder for dataset upload

philippotto:
- add "Add Folder" button next to "Add Dataset" button
image
- dummy context menu when rightclicking into the blank space of the sidebar:
image
- clicking "add dataset" will encode the active folder id in the url so that the import view can use that to initial the new target-folder selection
image
image

~The target folder functionality is currently front-end only (it simply uses the API to move the DS after the upload). This won't work if the dataset needs to be converted first. Not sure whether this should block the PR. The backend part probably won't happen in this year. Merging this now could be confusing to users when the DS doesn't pop up in the desired folder. However, this can be confusing right now, too~

- https://folderfollowup.webknossos.xyz

### Steps to test:
- use context menu in sidebar on folders and on blank space
- use new add-folder button
  - use the target folder UI to change the folder (should have initialized to the folder that was active in the dashboard)
  - upload something (zarr or normal)
  - dataset should appear there

### Issues:
- contributes to #6669 

------
- [X] Ready for review


frontend/javascripts/dashboard/dataset_view.tsx
```diff
@@ -268,11 +268,30 @@ function DatasetView(props: Props) {
...
+                ? `/datasets/upload?to=${context.activeFolderId}`
+                : "/datasets/upload"
+            }
+            style={margin}
+          >
             <Button type="primary" icon={<PlusOutlined />}>
               Add Dataset
             </Button>
           </Link>
+          {"queries" in context && context.activeFolderId != null && (
```
daniel-wer, line 283:
```suggestion
          {"activeFolderId" in context && context.activeFolderId != null && (
```

daniel-wer (approved):
Nicely done! Worked well during my testing, both for dataset upload and remote zarr import :+1: 

> The backend part probably won't happen in this year. Merging this now could be confusing to users when the DS doesn't pop up in the desired folder. However, this can be confusing right now, too :thinking:

Since the frontend gets to know whether the dataset needs to be converted, a simple toast/message could help to avoid confusion. It could state that this is not supported yet and that the dataset needs to be moved after the conversion.

philippotto:
> Since the frontend gets to know whether the dataset needs to be converted, a simple toast/message could help to avoid confusion. It could state that this is not supported yet and that the dataset needs to be moved after the conversion.

Good idea! Instead of showing a prompt, I used the antd form validation which also works nicely, I think.

image

and the normal case (as before)
image

philippotto:
@daniel-wer I updated the front-end to incorporate the newest changes by @fm3 so that the target folder also works with datasets which are converted after upload. The target folder *must* be specified now and the front-end ensures that the user may upload the DS to the folder by using the `disabled` property of the antd tree component. The "Allowed Teams", on the other hand, are not required anymore during upload. Also, the "Allowed Teams" warning/hint in the dataset settings were removed because there are not really necessary, anymore (due to the cumulative permission concept, users should be far more likely to have the necessary permissions).

frontend/javascripts/admin/dataset/dataset_upload_view.tsx
```diff
@@ -650,17 +668,8 @@ class DatasetUploadView extends React.Component<PropsWithFormAndRouter, State> {
...
-                      required: !isDatasetManagerOrAdmin,
-                      // @ts-expect-error ts-migrate(2322) FIXME: Type 'string | null' is not assignable to type 'st... Remove this comment to see the full error message
-                      message: !isDatasetManagerOrAdmin
-                        ? messages["dataset.import.required.initialTeam"]
-                        : null,
-                    },
-                  ]}
                 >
-                  <Tooltip title="Except for administrators and dataset managers, only members of the teams defined here will be able to view this dataset.">
+                  <Tooltip title="The dataset can be seen by administrators, dataset managers and by teams that have access to the folder to which the dataset is uploaded. If you want to grant additional teams access, define these teams here.">
```
daniel-wer, line 672: I cannot trigger this tooltip any longer. Maybe it should be triggered by an info icon as the tooltip next to the "Target Folder" form item.

daniel-wer (commented):
Very nice, that datasets that need a conversion can be uploaded to a specific folder now, too.

During testing I noticed
- [x] When selecting "Add Dataset" from a folder, the target folder is preselected which is nice. However, when trying to upload, webknossos complains that no target folder is selected (manually selecting the same folder fixes the issue). See screenshot
image
- [ ] The target folder preselection is also performed for folders the user doesn't have edit permissions for and which, therefore, couldn't be selected manually, because their entries are disabled
- [ ] [Optional] When selecting "Go to Dashboard" after a successful dataset upload, it would be cool to auto-activate the target folder so the user immediately sees the uploaded dataset

philippotto:
@daniel-wer Thank you for your testing feedback! I incorporated the first two points. I agree, the third point would be nice, but I decided against it for now due to the priority/cost ratio.

daniel-wer (approved):
LGTM, works as advertised :+1:

PR #6705, opened 2022-12-19
Catch java internal error in data loading

### LINKED ISSUE #6650: Catch SIGBUS/java.lang.InternalError and post to telemetry, trigger restart?

frcroth:
- https://___.webknossos.xyz

### Steps to test:
- The datastore is healthy /data/health -> 200
- Currently an InternalError is automatically thrown when loading data, so just open some local dataset.
- The data will not be fetched. Datastore is now unhealthy (for 30 seconds) /data/health != 200
- Slack notifactions are currently not tested, as I am not sure what to do here

### Issues:
- fixes #6650 

------
- [x] Ready for review


webknossos-datastore/app/com/scalableminds/webknossos/datastore/services/ApplicationHealthService.scala
```diff
@@ -0,0 +1,28 @@
...
+    slackNotificationService.notifyForInternalError(e)
+  }
+
+  def getRecentProblem(duration: Integer = 30): Option[HealthProblem] = {
+    if (encounteredProblems.isEmpty) return None
+    if (ChronoUnit.SECONDS.between(encounteredProblems.last.timestamp, Instant.now()) <= duration) {
+      Some(encounteredProblems.last)
+    } else {
+      None
+    }
```
fm3, line 25: Two suggestions
 - use prepend rather than append in the list (way more performant, makes headOption use possible, which is safer than last)
 - use FiniteDuration, which plays nicely with our Instant class, no need for ChronoUnit. `import scala.concurrent.duration.DurationInt` should make the `30 seconds` syntax possible.

```suggestion
  def getRecentProblem(duration: FiniteDuration = 30 seconds): Option[HealthProblem] = {
    encounteredProblems.headOption.filter { problem => 
!(problem.timestamp + duration).isPast
}
```
(did not test this code, may contain errors, but you get the idea)

fm3 (commented):
I added a few refactoring comments concerning the recent error check, I hope this makes sense to you :) I will do a final testing round later.

(Also, we need to make sure we ultimately remove the artificial errors before merging)

webknossos-datastore/app/com/scalableminds/webknossos/datastore/services/ApplicationHealthService.scala
frcroth: This use of filter on an option to get an option is really nice!

fm3 (approved):

PR #6706, opened 2022-12-19
Automatically open (and close) quick select settings when labeling in…

philippotto:
… preview mode. That way, bulk labelings with preview mode don't require constantly opening the settings manually.

- https://___.webknossos.xyz

### Steps to test:
- use quick select tool with and without preview mode and check usability of settings

------
- [X] Ready for review


hotzenklotz (approved):
Thanks for the quick action in response to my feedback. PR LGTM

PR #6711, opened 2022-12-20
Use less ram when initializing data textures

philippotto:
See code comments.

- https://lessram.webknossos.xyz

### Steps to test:
- open a dataset with a seg layer
- check that console does not show webgl errors
- change the color of a segment
- run `window.testContextLoss()` in the console and check that wk is usable again after a few seconds of flickering

### Issues:
- perf perf perf

------
- [X] Ready for review


frontend/javascripts/libs/UpdatableTexture.ts
```diff
@@ -134,21 +83,27 @@ class UpdatableTexture extends THREE.Texture {
     width: number,
     height: number,
   ) {
+    if (originalTexSubImage2D == null) {
+      // See explanation at declaration of originalTexSubImage2D.
+      originalTexSubImage2D = this.gl.texSubImage2D.bind(this.gl);
+      this.gl.texSubImage2D = _.noop;
+    }
```
daniel-wer, line 90: This workaround is a bit different from when we last talked, I think. If I see it correctly, `gl.texSubImage2D` is permanently noop'ed, instead of only temporarily during the texture initialization. Couldn't this lead to issues if the function is called somewhere else?

daniel-wer (commented):
As far as I can see, everything still works.

When I switch to flight mode for the first time, I get a WebGL warning `[.WebGL-0xcdc0032ea00] GL_INVALID_VALUE: Texture dimensions must all be greater than zero.` which is one less than on master (where there is also `[.WebGL-0xcdc050ba900] GL_INVALID_OPERATION: Level of detail outside of range.`), so that's good I suppose :grin:

Your call whether it makes sense to execute the screenshot tests for this branch to make sure everything still renders as usual (since you can better assess what changed in this PR).

Edit: Also a final test on an iphone whether this still fixes the RAM initialization issue would be good :)

frontend/javascripts/libs/UpdatableTexture.ts
philippotto: It's only slightly different, because the old behavior basically also permanently noop'ed the webgl function, but temporarily swapped it back in before calling the function in `update`. After the update, the function was noop'ed again. So, if this function would be called elsewhere nothing would happen with both approaches. Therefore, I removed the temporary pointer juggling from the update function, since it didn't really have a benefit.

> instead of only temporarily during the texture initialization

This would be ideal, but I don't know how we would do this elegantly, since threejs initializes the textures at its own will (most likely on the first render call). For normal initialization, it might be feasible to override it only temporarily, but if a webgl context is lost and recovered, we would also need to plug into that (if we care about that scenario).

> Couldn't this lead to issues if the function is called somewhere else?

Yes, I think, if we would use other textures somewhere for rendering that wouldn't work anymore. However, we have never needed that which is why I think we can ignore that problem for now.

daniel-wer: Ah, ok makes sense, thanks for explaining. I agree that we should leave it as-is, then :)

philippotto:
> Your call whether it makes sense to execute the screenshot tests for this branch to make sure everything still renders as usual (since you can better assess what changed in this PR).

This was a good call :+1: The nightly showed that the skeletons were broken, because a normal THREE.DataTexture was used for that :see_no_evil: I adapted the workaround so, that only gl calls are noop'ed if an empty TypedArray is passed. I think, this should be way more robust.

The last nightly still failed, but this is because the connectome tab is not enabled for the particular dev instance. I tested that case manually and it worked. Please have a final look at my last changes and then I'd merge this PR :crossed_fingers:

daniel-wer (approved):
Nice work! Glad you made the call to execute the screenshot tests and caught that bug :1st_place_medal:

Issue #3401, opened 2018-10-25
Try to get rid of circular dependencies in JS


philippotto:
From time to time, we are running into problems due to such cycles. We should try to enforce cycle-free import hierarchies. Before that, we need to figure out how to refactor the code to avoid the cycles in the first place.

philippotto:
Cycles found with https://github.com/aackerman/circular-dependency-plugin:

```
WARNING in Circular dependency detected:
frontend/javascripts/admin/admin_rest_api.js -> frontend/javascripts/libs/request.js -> frontend/javascripts/admin/datastore_health_check.js -> frontend/javascripts/admin/admin_rest_api.js

WARNING in Circular dependency detected:
frontend/javascripts/admin/datastore_health_check.js -> frontend/javascripts/libs/request.js -> frontend/javascripts/admin/datastore_health_check.js

WARNING in Circular dependency detected:
frontend/javascripts/admin/task/task_create_bulk_view.js -> frontend/javascripts/admin/task/task_create_form_view.js -> frontend/javascripts/admin/task/task_create_bulk_view.js

WARNING in Circular dependency detected:
frontend/javascripts/admin/task/task_create_form_view.js -> frontend/javascripts/admin/task/task_create_bulk_view.js -> frontend/javascripts/admin/task/task_create_form_view.js

WARNING in Circular dependency detected:
frontend/javascripts/libs/user_local_storage.js -> frontend/javascripts/oxalis/store.js -> frontend/javascripts/oxalis/model/sagas/root_saga.js -> frontend/javascripts/oxalis/model/sagas/volumetracing_saga.js -> frontend/javascripts/oxalis/model/sagas/automatic_brush_saga.js -> frontend/javascripts/oxalis/api/internal_api.js -> frontend/javascripts/oxalis/api/api_latest.js -> frontend/javascripts/libs/user_local_storage.js

WARNING in Circular dependency detected:
[...]

PR #6713, opened 2022-12-21
Remove most cyclic dependencies in front-end

philippotto:
This PR removes most, but not all cyclic dependencies. I'd suggest to tackle the remaining one's once we've got some air for that. I added a check to the CI, though, that ensures that a new PR won't increase the amount of cycles. I hope this works alright (if not we can disable that check again).

Changed code is 99% just moved around. The singletons module is new and due to it's explicit set-functions breaks up some critical cyclic dependencies (e.g., Model, Store etc).

- https://cyclicdeps.webknossos.xyz

### Steps to test:
- CI

### Issues:
- contributes to #3401

------
- [X] Ready for review


frontend/javascripts/oxalis/model/accessors/flycam_accessor.ts
```diff
@@ -307,7 +312,7 @@ export function getValidTaskZoomRange(
   state: OxalisState,
   respectRestriction: boolean = false,
 ): [number, number] {
-  const defaultRange = [userSettings.zoom.minimum, Infinity];
+  const defaultRange = [baseDatasetViewConfiguration.zoom.minimum, Infinity];
```
daniel-wer, line 315: Do you know why typescript did not complain about this, before?

daniel-wer (commented):
Splendid work, thank you for taking the time! I'll do some quick monkey testing and will report back, afterwards.

frontend/javascripts/oxalis/model/accessors/flycam_accessor.ts
philippotto: If I remember correctly, both code versions are fine value-wise (i.e., they are set to the same value), but the old one caused a cyclic dependencies, about which TS doesn't care :)

daniel-wer: Ah, I searched for `zoom` in the `user_settings.schema` and couldn't find it, but there's a `...baseDatasetViewConfiguration,` in there that I missed :+1:

daniel-wer (approved):
LGTM :tada:

PR #6715, opened 2023-01-02
add license checker

normanrz:
Adds a license checker for the backend. It can be run with `sbt dumpLicenseReport` (not sure where to document this). The frontend already has that in the form of `yarn licenses ls`. 

------
- [x] Ready for review

### DIFF
--- a/project/license.sbt
+++ b/project/license.sbt
@@ -1,0 +1,1 @@
+addSbtPlugin("com.typesafe.sbt" % "sbt-license-report" % "1.2.0")


fm3 (approved):
Most of the development commands are defined in package.json (even those that in turn call sbt) – maybe that would be a good place to “document” this? Otherwise, we could start a more generalized readme markdown file for development, or start a list in notion with such commands. There is also the public webknossos wiki, but I think it is becoming disused 😬

PR #6718, opened 2023-01-03
Nested sql interpolation

normanrz:
- https://___.webknossos.xyz

### Steps to test:
- abc

### Issues:
- fixes #

------
- [ ] Ready for review

### DIFF
--- a/app/models/voxelytics/VoxelyticsDAO.scala
+++ b/app/models/voxelytics/VoxelyticsDAO.scala
@@ -5,6 +5,7 @@ import com.scalableminds.util.tools.Fox
 import models.user.User
 import play.api.libs.json._
 import slick.jdbc.PostgresProfile.api._
+import utils.CustomSQLInterpolation2.customSQLInterpolation
 import utils.{ObjectId, SQLClient, SimpleSQLDAO}
 
 import javax.inject.Inject
@@ -74,7 +75,7 @@ class VoxelyticsDAO @Inject()(sqlClient: SQLClient)(implicit ec: ExecutionContex
 
   def findWorkflowByHashAndOrganization(organizationId: ObjectId, workflowHash: String): Fox[WorkflowEntry] =
     for {
-      r <- run(sql"""
+      r <- run(nsql"""
           SELECT name, hash
           FROM webknossos.voxelytics_workflows
           WHERE hash = $workflowHash AND _organization = $organizationId
@@ -84,7 +85,7 @@ class VoxelyticsDAO @Inject()(sqlClient: SQLClient)(implicit ec: ExecutionContex
 
   def findWorkflowByHash(workflowHash: String): Fox[WorkflowEntry] =
     for {
-      r <- run(sql"""
+      r <- run(nsql"""
           SELECT name, hash, _organization
           FROM webknossos.voxelytics_workflows
           WHERE hash = $workflowHash
@@ -277,7 +278,7 @@ class VoxelyticsDAO @Inject()(sqlClient: SQLClient)(implicit ec: ExecutionContex
   def upsertArtifactChecksumEvent(artifactId: ObjectId, ev: ArtifactFileChecksumEvent): Fox[Unit] =
     for {
       _ <- run(
-        sqlu"""INSERT INTO webknossos.voxelytics_artifactFileChecksumEvents (_artifact, path, resolvedPath, checksumMethod, checksum, fileSize, lastModified, timestamp)
+        nsql"""INSERT INTO webknossos.voxelytics_artifactFileChecksumEvents (_artifact, path, resolvedPath, checksumMethod, checksum, fileSize, lastModified, timestamp)
                VALUES ($artifactId, ${ev.path}, ${ev.resolvedPath}, ${ev.checksumMethod}, ${ev.checksum}, ${ev.fileSize}, ${ev.lastModified}, ${ev.timestamp})
                ON CONFLICT (_artifact, path, timestamp)
                  DO UPDATE SET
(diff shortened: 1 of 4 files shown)


app/utils/SQL.scala
```diff
@@ -0,0 +1,251 @@
...
+    SqlToken(sql = outputSql.toString, values = outputValues.toList)
+  }
+}
+
+object NestedSQLInterpolation {
+  implicit def nestedSQLInterpolation(s: StringContext): NestedSQLInterpolator = new NestedSQLInterpolator(s)
+}
+
+case class SqlToken(sql: String, values: List[SqlValue] = List()) {
+  def debug: String = {
```
fm3, line 53:
```suggestion
  def debugInfo: String = {
```

app/utils/SQL.scala
```diff
@@ -0,0 +1,251 @@
...
+
+      override def getDumpInfo = DumpInfo(DumpInfo.simpleNameFor(getClass), mainInfo = s"[${debug}]")
+
+      protected[this] def createBuilder = Vector.newBuilder[R]
+    }
+
+  def asUpdate = as[Int](GetUpdateValue).head
+
+  def stripMargin(marginChar: Char): SqlToken = copy(sql = sql.stripMargin(marginChar))
+  def stripMargin: SqlToken = copy(sql = sql.stripMargin)
```
fm3, line 82: those seem unused? are they supposed to be part of the API?

app/utils/SQL.scala
```diff
@@ -0,0 +1,251 @@
...
+
+        protected def extractValue(rs: PositionedResult): R = rconv(rs)
+      }
+
+      override def getDumpInfo = DumpInfo(DumpInfo.simpleNameFor(getClass), mainInfo = s"[${debug}]")
+
+      protected[this] def createBuilder = Vector.newBuilder[R]
+    }
+
+  def asUpdate = as[Int](GetUpdateValue).head
```
fm3, line 79: Please annotate all public functions with a return type (and make private what can be private)

app/utils/SQL.scala
```diff
@@ -0,0 +1,251 @@
...
+
+  def tupleList(values: List[List[Any]]): SqlToken = {
+    val sqlValueLists = values.map(list => list.map(SqlValue.makeSqlValue))
+    SqlToken(sql = sqlValueLists.map(list => s"(${list.map(_.getPlaceholder()).mkString(", ")})").mkString(", "),
+             values = sqlValueLists.flatten)
+  }
+
+  def raw(s: String) = SqlToken(s)
+
+  def empty(): SqlToken = raw("")
```
fm3, line 121:
```suggestion
  def empty: SqlToken = raw("")
```
(parens are usually added only where side effects happen)

app/utils/SQL.scala
```diff
@@ -0,0 +1,251 @@
...
+
+  def identifier(id: String): SqlToken = raw('"' + id + '"')
+}
+
+trait SqlValue {
+  def setParameter(pp: PositionedParameters): Unit
+
+  def getPlaceholder(): String = "?"
```
fm3, line 129:
```suggestion
  def placeholder: String = "?"
```

app/utils/SQL.scala
```diff
@@ -0,0 +1,251 @@
...
+
+case class InstantValue(v: Instant) extends SqlValue {
+  override def setParameter(pp: PositionedParameters): Unit = pp.setTimestamp(v.toSql)
+
+  override def getPlaceholder(): String = "?::TIMESTAMPTZ"
+
+  override def debug(): String = s"'${v.toString}'::TIMESTAMPTZ"
+}
+
+case class DurationValue(v: Duration) extends SqlValue {
```
fm3, line 208: I think we should use FiniteDuration – the generic Duration can also hold infinity, which will lead to unexpected behavior when calling length. I don’t think we currently have any semantics where infinity is needed in webKnossos.

fm3 (commented):
This is good stuff, I’m looking forward to using it everywhere!
I added a few minor comments.

I don’t think we can make this all that much cleaner, you already got pretty far there.

On naming the string modifier – how about just q for query? I think we would want to use this for all queries, whether we will use the nesting feature or not – then the n is no longer so relevant for the usages. What do you think? sql would also be nice, but that might cause confusion with the existing one provided by slick.

normanrz:
> On naming the string modifier – how about just q for query?

I like `q`.

How would you structure the files/packages? Should this live under `utils` or `utils.sql` or something?

fm3:
`utils.sql` sounds good. This could also hold the existing SQLHelpers file (which should probably be split into parts, but that can wait for a later PR)

app/utils/SQL.scala
normanrz: I had them in, because they are common with `s` and `sql[u]`-based string interpolation. I took them out again, because they don't seem to be used anywhere in our codebase.

app/models/voxelytics/VoxelyticsDAO.scala
```diff
@@ -4,8 +4,9 @@ import com.scalableminds.util.time.Instant
 import com.scalableminds.util.tools.Fox
 import models.user.User
 import play.api.libs.json._
-import slick.jdbc.PostgresProfile.api._
-import utils.{ObjectId, SQLClient, SimpleSQLDAO}
+import utils.sql.SqlInterpolation.sqlInterpolation
```
normanrz, line 7: Any idea how to make this more concise? I'm afraid my "implicit"-foo is not sufficient to optimize this further.

fm3: I added a commit, forwarding the implicit definition to the SimpleSQLDAO (from which the VoxelyticsDAO inherits, so no need for the import there). Is this what you had in mind?

normanrz: Yes! Perfect!

app/utils/sql/SqlInterpolation.scala
```diff
@@ -0,0 +1,248 @@
...
+      case x: FiniteDuration => DurationValue(x)
+      case x: ObjectId       => ObjectIdValue(x)
+      case x: JsValue        => JsonValue(x)
+    }
+}
+
+case class StringValue(v: String) extends SqlValue {
+  override def setParameter(pp: PositionedParameters): Unit = pp.setString(v)
+
+  override def debugInfo: String = "'" + v + "'"
```
fm3, line 158: the comment says that debugInfo should be pastable in an sql editor – I guess this might mean that v should be escaped here?

test/backend/SQLTestSuite.scala
```diff
@@ -0,0 +1,140 @@
...
+import com.scalableminds.util.time.Instant
+import org.scalatestplus.play.PlaySpec
+import play.api.libs.json.Json
+import utils.ObjectId
+import utils.sql.SqlInterpolation.sqlInterpolation
+import utils.sql._
+
+import scala.concurrent.duration.DurationInt
+
+class SQLTestSuite extends PlaySpec {
```
fm3, line 12:
```suggestion
class SqlInterpolationTestSuite extends PlaySpec {
```

fm3 (commented):
I added another commit following some code style recommendations from intellij, nothing to worry about.
I also added two more minor comments:

fm3 (approved):
Awesome 🎉 

I think we should take some time soon to propagate this to all the DAOs. In the process, we will likely notice more features missing here, but they can then be added as needed.

PR #6724, opened 2023-01-05
Fix occasional broken node selection

philippotto:
In addition to the incorrect base, I adapted the context menu so that it won't crash in such a case. Also, I removed some `Maybe` wrappers as I don't see a benefit in it anymore, since we have a good typechecker.

- https://___.webknossos.xyz

### Steps to test:
[broken-id-repro.zip](https://github.com/scalableminds/webknossos/files/10353812/broken-id-repro.zip)
- import the NML from the zip
- activate the node id 4878140 via the status bar
- right click the centered node (on master, this fails)

### Issues:
- https://scm.slack.com/archives/C02H5T8Q08P/p1672909057564449

------
- [X] Ready for review


daniel-wer (approved):
Awesome find! :tada: LGTM :+1:

PR #6728, opened 2023-01-09
Add explicit name for foreign-key constraint to fix false positives in diff schema

fm3:
After a postgres upgrade the automatic naming for tuple foreign keys was changed, resulting in false positives in the schema difference detection.

### Steps to test:
- the foreign key constraint referencing `webknossos.voxelytics_workflows(_organization, hash)` in the `voxelytics_runs` table should be named `voxelytics_runs__organization_workflow_hash_fkey` regardless of postgres version

------
- [x] Ready for review

### DIFF
--- a/MIGRATIONS.unreleased.md
+++ b/MIGRATIONS.unreleased.md
@@ -11,3 +11,4 @@ User-facing changes are documented in the [changelog](CHANGELOG.released.md).
 ### Postgres Evolutions:
 
 - [094-pricing-plans.sql](conf/evolutions/reversions/094-pricing-plans.sql)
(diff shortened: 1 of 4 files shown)


normanrz (commented):
What about the other fkeys? Should we also provide names explicitly?

fm3:
Fair question. Up until this point this naming always been deterministic. Apparently this is not guaranteed by postgres, but as long as this does not create problems, it should be fine. Normal db operation is not affected by the names, only evolutions are. Thanks to the diff_schema script, we should notice immediately once this diverges again somewhere. So I’d vote to keep the code compact by not adding explicit names everywhere. Do you agree?

normanrz (approved):

PR #6732, opened 2023-01-10
Vx followup

normanrz:
- Optimizes inserts, when many events are sent in a single request, by grouping them into single database queries
- Optimizes workflow listing by batching database queries
- Provides consolidated view of tasks that have been resumed, i.e. combining the information about the chunks of multiple runs
- Removes some cyclic dependencies in the frontend
- Adds support for Enumerations in SqlInterpolation (stolen from #6723)

### Steps to test:
- Run a voxelytics workflow with multiple chunks, interrupt the run and resume
- Once complete, the workflow report should provide accurate information about all completed chunks

------
- [x] Ready for review


app/utils/sql/SqlInterpolation.scala
```diff
@@ -53,6 +53,9 @@ case class SqlToken(sql: String, values: List[SqlValue] = List()) {
   def debugInfo: String = {
     // The debugInfo should be pastable in an SQL client
     val parts = sql.split("\\?", -1)
+    if (parts.tail.length != values.length) {
+      println(s"[$sql] {${values}}")
```
fm3, line 57: println should not stay in here. maybe return this instead?

app/controllers/VoxelyticsController.scala
```diff
@@ -159,87 +158,78 @@ class VoxelyticsController @Inject()(
 
   def storeWorkflowEvents(workflowHash: String, runName: String): Action[List[WorkflowEvent]] =
     sil.SecuredAction.async(validateJson[List[WorkflowEvent]]) { implicit request =>
-      def createWorkflowEvent(runId: ObjectId, event: WorkflowEvent): Fox[Unit] =
-        event match {
-          case ev: RunStateChangeEvent => voxelyticsDAO.upsertRunStateChangeEvent(runId, ev)
+      def createWorkflowEvent(runId: ObjectId, events: List[WorkflowEvent]): Fox[Unit] =
+        if (events.nonEmpty) {
+          events.head match {
```
fm3, line 163: I’d prefer functional style here
```suggestion
        events.headOption.map { event =>
          event match { … }
        }.getOrElse(Fox.successful())
```

app/models/voxelytics/VoxelyticsService.scala
```diff
@@ -118,10 +150,41 @@ class VoxelyticsService @Inject()(voxelyticsDAO: VoxelyticsDAO)(implicit ec: Exe
...
+          val sortedTaskRuns = group._2.sortBy(_.beginTime).reverse
+          val combinedTaskRun = combinedTaskRuns.find(_.taskName == group._1)
+          Json.obj(
+            "taskName" -> group._1,
+            "state" -> sortedTaskRuns
+              .map(_.state)
+              .find(s => s != VoxelyticsRunState.SKIPPED && s != VoxelyticsRunState.PENDING)
+              .orElse(sortedTaskRuns.headOption.map(_.state))
+              .getOrElse(VoxelyticsRunState.SKIPPED)
+              .asInstanceOf[VoxelyticsRunState],
```
fm3, line 167: Could you add a comment explaining why a cast is needed? (Or better yet, avoid it?)

app/models/voxelytics/VoxelyticsService.scala
```diff
@@ -118,10 +150,41 @@ class VoxelyticsService @Inject()(voxelyticsDAO: VoxelyticsDAO)(implicit ec: Exe
...
+            "beginTime" -> Try(sortedTaskRuns.flatMap(_.beginTime).min).toOption,
+            "endTime" -> Try(sortedTaskRuns.flatMap(_.endTime).max).toOption,
+            "currentExecutionId" -> combinedTaskRun.flatMap(_.currentExecutionId),
+            "chunks" -> combinedTaskRun
+              .map(_.chunks)
+              .getOrElse(ChunkStatistics(0, 0, 0, 0, 0))
+              .asInstanceOf[ChunkStatistics],
```
fm3, line 174: Again it is not obvious to me why a cast is needed. Don’t the items have this type already?

app/models/voxelytics/VoxelyticsDAO.scala
```diff
@@ -257,358 +415,287 @@ class VoxelyticsDAO @Inject()(sqlClient: SQLClient)(implicit ec: ExecutionContex
...
+          profiling.stddev_cpuSystem AS stddev_cpuSystem,
+          durations.max_duration AS max_duration,
+          durations.median_duration AS median_duration,
+          durations.stddev_duration AS stddev_duration,
+          durations.sum_duration AS sum_duration
+        FROM ( -- Chunks grouped by executionId
+          SELECT
+            c.executionId,
+            COUNT(c._chunk) AS countTotal,
+            SUM(CASE WHEN c.state = 'FAILED' THEN 1 ELSE 0 END) AS countFailed,
```
fm3, line 634: Should possibly use the enums as SqlValue? That way, usage search works here.

fm3 (commented):
Scala Backend mostly LGTM. I added a few comments.

I did not dive deeply into the very complex SQL queries. If you think they should be reviewed in detail, maybe we could schedule a call.

app/models/voxelytics/VoxelyticsService.scala
normanrz: Maybe you can help me here. If I leave this out, the compiler says that it can't figure out the type. If I put the same expression in a `val` and then in the `Json.obj(...)` it works, aswell.

frontend/javascripts/admin/voxelytics/statistics_tab.tsx
```diff
@@ -85,12 +90,27 @@ export default function StatisticsTab({
...
-                {row.executionId}
-                <br />
-                <span className="stats-label">
-                  {row.countFinished !== row.countTotal && <>{row.countFinished} of </>}
-                  {row.countTotal} chunk{row.countTotal !== 1 && "s"} completed
-                </span>
+                <Tooltip
+                  overlay={
+                    <>
+                      {row.counts.complete}/{row.counts.total - row.counts.skipped} complete,{" "}
```
philippotto, line 96: not sure whether it is intuitive to subtract the skipped ones from `total`. how about: `r remaining, x complete, y cancelled, z failed, s skipped, t total` where `r = total - complete - skipped`.

frontend/javascripts/admin/voxelytics/statistics_tab.tsx
```diff
@@ -85,12 +90,27 @@ export default function StatisticsTab({
...
+                >
+                  <span>
+                    {row.executionId}
+                    <br />
+                    <span className="stats-label">
+                      {row.counts.complete !== row.counts.total - row.counts.skipped && (
+                        <>{row.counts.complete} of </>
+                      )}
+                      {row.counts.total - row.counts.skipped} chunk
+                      {row.counts.total - row.counts.skipped !== 1 && "s"} completed
```
philippotto, line 110:
```suggestion
                      {row.counts.total - row.counts.skipped} {pluralize("chunk", row.counts.total - row.counts.skipped} completed
```

needs to be imported

frontend/javascripts/admin/voxelytics/task_view.tsx
```diff
@@ -105,13 +77,13 @@ function TaskView({
               }}
             >
               <Progress
-                percent={(taskInfo.chunksFinished / taskInfo.chunksTotal) * 100}
+                percent={(taskInfo.chunks.complete / taskInfo.chunks.total) * 100}
```
philippotto, line 80: here, I'd think that skipped ones need to be subtracted from total?

frontend/javascripts/admin/voxelytics/workflow_list_view.tsx
```diff
@@ -149,9 +106,11 @@ export default function WorkflowListView() {
     return (
       <Tooltip title={label}>
         <Progress
-          percent={Math.round(((completeCount + failedCount) / runnableCount) * 100)}
+          percent={Math.round(
+            ((completeCount + cancelledCount + failedCount) / run.taskStatistics.total) * 100,
+          )}
           status={runStateToStatus(run.state)}
-          success={{ percent: Math.round((completeCount / runnableCount) * 100) }}
+          success={{ percent: Math.round((completeCount / run.taskStatistics.total) * 100) }}
```
philippotto, line 113: divide by `runnableCount` instead of `total`?

philippotto (approved):
lgtm mostly :+1: only left some comments regarding how the progress is calculated and represented. ignore my comments if you think, the current way makes more sense. I didn't use the new reports heavily yet, so I can only comment coming from an outside perspective.

app/models/voxelytics/VoxelyticsService.scala
fm3: Huh, I can reproduce the behavior. This actually smells a bit like a bug in the scala compiler to me. I would very much prefer the `val` variant over `asInstanceOf` though, if that’s ok with you.

app/utils/sql/SqlInterpolation.scala
normanrz: gone

app/models/voxelytics/VoxelyticsDAO.scala
normanrz: good idea

normanrz:
@daniel-wer what do you think about @philippotto's comments? I think you have spent the most time with the reports so far.

daniel-wer:
In my opinion, the main thing that is interesting to me when looking at a report with a currently running task is:
- how many chunks are already done, have failed, still need to be computed (whether the done tasks were computed in this specific run or before is not really important, but the distinction could be helpful)
- what is the ETA

In the workflow list, the progress bar is chunked with different colors. Couldn't the progress bar for a running task also show this distinction? Always show the overall number of chunks as the total, then color the skipped ones gray, the finished ones blue or green and the failed ones red.

When looking at the statistics tab, I'd like to see the aggregated stats for all completed chunks.

normanrz:
> When looking at the statistics tab, I'd like to see the aggregated stats for all completed chunks.

Should that include running and failed chunks?

fm3 (approved):

normanrz:
I now solved the performance issue by restructuring the database: `voxelytics_{run,task,chunk}StateChangeEvents` are gone in favor or `state`, `beginTime`, `endTime` fields are added to `voxelytics_{runs,tasks,chunks}`. The API with vx is unchanged.
The forward evolution is straightforward, but lossy. Therefore, I don't know how to do the reverse evolution. I'd argue to not implement a full reverse evolution and just keep the old tables until we know everything works. @fm3 wdyt?

fm3:
@normanrz that sounds fair to me. If the app image is rolled back after we notice that something is wrong, the old tables are filled again by the old code. Will that be okay / create an acceptable state? If so, fine by me. Also, I’d ask you to add a follow-up issue to drop the unused tables in another evolution after a grace period

normanrz:
Ingesting 120k chunks now only took a few seconds (on my Mac) 🎉 💯 
Screenshot 2023-01-16 at 17 50 26

Issue #6721, opened 2023-01-04
Use composable SQL interpolation syntax for all queries


fm3:
#6718 introduces a way to write sql syntax with composable string interpolation. This should be used for all sql queries throughout the webknossos backend.

fm3:
closed via #6733 and #6734

Issue #6730, opened 2023-01-10
SQL helper functions for updates, inserts, upserts


fm3:
With the new SqlInterpolation it could make sense to create helper functions making inserts, upserts, updates less verbose.

normanrz:
```scala
  def insert(tableName: String)(fields: (String, Any)*): SqlToken = {
    val keys = fields.map(_._1)
    val values = fields.map(f => SqlValue.makeSqlValue(f._2)).toList
    val sql =
      s"""
        INSERT INTO ${identifier(tableName)} (${keys.map(identifier).mkString(", ")})
        VALUES (${values.map(_.placeholder).mkString(", ")})"""
    SqlToken(sql = sql, values = values)
  }

  def update(tableName: String)(whereFields: (String, Any)*)(updateFields: (String, Any)*): SqlToken = {
    val whereKeys = whereFields.map(_._1)
    val whereValues = whereFields.map(f => SqlValue.makeSqlValue(f._2)).toList
    val updateKeys = updateFields.map(_._1)
    val updateValues = updateFields.map(f => SqlValue.makeSqlValue(f._2)).toList
    val sql =
      s"""
        UPDATE ${identifier(tableName)}
        WHERE ${whereKeys.zip(whereValues).map(kv => s"${identifier(kv._1)} = ${kv._2.placeholder}").mkString(" AND ")}
        SET ${updateKeys.zip(updateValues).map(kv => s"${identifier(kv._1)} = ${kv._2.placeholder}").mkString(", ")}"""
    SqlToken(sql = sql, values = whereValues ++ updateValues)
  }
  ```

PR #6734, opened 2023-01-12
Sql refactoring part two

fm3:
Adapts the remaining queries, second part to #6723

- https://sqlrefactoringparttwo.webknossos.xyz

### Steps to test:
- same as in #6723 , this time focus on task management, users, multi-orga setup, worker jobs

### Issues:
- fixes #https://github.com/scalableminds/webknossos/issues/6721

------
- [x] Ready for review

### DIFF
--- a/app/models/job/Job.scala
+++ b/app/models/job/Job.scala
@@ -1,6 +1,5 @@
 package models.job
 
-import akka.actor.ActorSystem
 import com.scalableminds.util.accesscontext.{DBAccessContext, GlobalAccessContext}
 import com.scalableminds.util.geometry.BoundingBox
 import com.scalableminds.util.mvc.Formatter
@@ -8,25 +7,22 @@ import com.scalableminds.util.time.Instant
 import com.scalableminds.util.tools.{Fox, FoxImplicits}
 import com.scalableminds.webknossos.schema.Tables._
 import com.typesafe.scalalogging.LazyLogging
-
-import javax.inject.Inject
 import models.analytics.{AnalyticsService, FailedJobEvent, RunJobEvent}
 import models.binary.{DataSetDAO, DataStoreDAO}
 import models.job.JobState.JobState
 import models.organization.OrganizationDAO
 import models.user.{MultiUserDAO, User, UserDAO}
 import oxalis.telemetry.SlackNotificationService
-import play.api.inject.ApplicationLifecycle
 import play.api.libs.json.{JsObject, Json}
 import slick.jdbc.PostgresProfile.api._
 import slick.jdbc.TransactionIsolation.Serializable
 import slick.lifted.Rep
-import utils.sql.{SqlClient, SQLDAO}
+import utils.sql.{SQLDAO, SqlClient, SqlToken}
 import utils.{ObjectId, WkConf}
 
+import javax.inject.Inject
 import scala.concurrent.ExecutionContext
-import scala.concurrent.duration.FiniteDuration
-import scala.concurrent.duration._
+import scala.concurrent.duration.{FiniteDuration, _}
 
 case class Job(
     _id: ObjectId,
@@ -127,44 +123,43 @@ class JobDAO @Inject()(sqlClient: SqlClient)(implicit ec: ExecutionContext)
   override def findAll(implicit ctx: DBAccessContext): Fox[List[Job]] =
     for {
       accessQuery <- readAccessQuery
-      r <- run(
-        sql"select #${columns.debugInfo} from #${existingCollectionName.debugInfo} where #${accessQuery.debugInfo} order by created"
-          .as[JobsRow])
+      r <- run(q"select $columns from $existingCollectionName where $accessQuery order by created".as[JobsRow])
       parsed <- parseAll(r)
     } yield parsed
 
   override def findOne(jobId: ObjectId)(implicit ctx: DBAccessContext): Fox[Job] =
     for {
       accessQuery <- readAccessQuery
-      r <- run(
-        sql"select #${columns.debugInfo} from #${existingCollectionName.debugInfo} where #${accessQuery.debugInfo} and _id = $jobId"
-          .as[JobsRow])
+      r <- run(q"select $columns from $existingCollectionName where $accessQuery and _id = $jobId".as[JobsRow])
       parsed <- parseFirst(r, jobId)
     } yield parsed
 
-  def countUnassignedPendingForDataStore(_dataStore: String): Fox[Int] =
+  def countUnassignedPendingForDataStore(dataStoreName: String): Fox[Int] =
     for {
-      r <- run(sql"""select count(_id) from #${existingCollectionName.debugInfo}
-              where state = '#${JobState.PENDING}'
-              and manualState is null
-              and _dataStore = ${_dataStore}
-              and _worker is null""".as[Int])
+      r <- run(q"""select count(_id) from $existingCollectionName
+                   where state = ${JobState.PENDING}
+                   and manualState is null
+                   and _dataStore = $dataStoreName
+                   and _worker is null""".as[Int])
       head <- r.headOption
     } yield head
 
   def countUnfinishedByWorker(workerId: ObjectId): Fox[Int] =
     for {
-      r <- run(
-        sql"select count(_id) from #${existingCollectionName.debugInfo} where _worker = $workerId and state in ('#${JobState.PENDING}', '#${JobState.STARTED}') and manualState is null"
-          .as[Int])
+      r <- run(q"""SELECT COUNT(_id)
+                   FROM $existingCollectionName
+                   WHERE _worker = $workerId
+                   AND state in ${SqlToken.tuple(List(JobState.PENDING, JobState.STARTED))}
+                   AND manualState IS NULL""".as[Int])
       head <- r.headOption
     } yield head
 
   def findAllUnfinishedByWorker(workerId: ObjectId): Fox[List[Job]] =
     for {
-      r <- run(
-        sql"select #${columns.debugInfo} from #${existingCollectionName.debugInfo} where _worker = $workerId and state in ('#${JobState.PENDING}', '#${JobState.STARTED}') and manualState is null order by created"
-          .as[JobsRow])
+      r <- run(q"""SELECT $columns from $existingCollectionName
+                   WHERE _worker = $workerId and state in ${SqlToken.tuple(List(JobState.PENDING, JobState.STARTED))}
+                   AND manualState IS NULL
+                   ORDER BY created""".as[JobsRow])
       parsed <- parseAll(r)
     } yield parsed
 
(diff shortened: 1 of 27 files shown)


app/utils/sql/SqlTypeImplicits.scala
```diff
@@ -1,35 +1,110 @@
...
+
+  implicit def boundingBoxToSqlValue(v: BoundingBox): SqlValue = BoundingBoxValue(v)
+
+  implicit def vec3IntToSqlValue(v: Vec3Int): SqlValue = Vector3Value(v.toVec3Double)
+
+  implicit def vec3DoubleToSqlValue(v: Vec3Double): SqlValue = Vector3Value(v)
+
+  implicit def finiteDurationToSqlValue(v: FiniteDuration): SqlValue = DurationValue(v)
+
+  // Conversions to SqlToken. This forwarded conversion is needed as implicits are not automatically chained.
```
normanrz, line 65: I was wondering, if we actually need the `*toSqlValue` functions? Maybe we could just pass around tokens.

app/utils/sql/SqlInterpolation.scala
```diff
@@ -102,16 +94,12 @@ object SqlToken {
     SqlToken(sql = outputSql.toString, values = outputValues.toList)
   }
 
-  def tuple(values: Seq[Any]): SqlToken = {
-    val sqlValues = values.map(SqlValue.makeSqlValue)
-    SqlToken(sql = s"(${sqlValues.map(_.placeholder).mkString(", ")})", values = sqlValues.toList)
-  }
+  def tuple(values: List[SqlValue]): SqlToken =
+    SqlToken(sql = s"(${values.map(_.placeholder).mkString(", ")})", values = values)
```
normanrz, line 98:
```suggestion
  def tupleFromList(values: List[SqlValue]): SqlToken =
    SqlToken(sql = s"(${values.map(_.placeholder).mkString(", ")})", values = values)
  def tuple(values: SqlValue*): SqlToken =
    SqlToken(sql = s"(${values.map(_.placeholder).mkString(", ")})", values = values.toList)  
```

app/models/job/Job.scala
```diff
@@ -176,59 +171,52 @@ class JobDAO @Inject()(sqlClient: SqlClient)(implicit ec: ExecutionContext)
...
-                         values(${j._id}, ${j._owner}, ${j._dataStore}, ${j.command}, '#${sanitize(
-          j.commandArgs.toString)}',
-                          '#${j.state.toString}', #${optionLiteralSanitized(j.manualState.map(_.toString))},
-                          #${optionLiteral(j._worker.map(_.toString))},
-                          #${optionLiteralSanitized(j.latestRunId)},
-                          #${optionLiteralSanitized(j.returnValue)},
-                          #${optionLiteral(j.started.map(_.toString))},
-                          #${optionLiteral(j.ended.map(_.toString))},
-                          ${j.created}, ${j.isDeleted})""")
+      _ <- run(q"""INSERT INTO webknossos.jobs(
```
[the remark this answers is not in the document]
fm3: existingCollectionName points to webknossos.jobs_, which  is the view with isDeleted = false. It cannot be used for inserts. We could add the same for collectionName, which might make sense. But tbh I’m not certain if either of them really make all that much sense. existingCollectionName stems from a time when I tried to do as much as possible in the SQLDAO superclass. Do you have an opinion on this?

app/models/task/TaskType.scala
```diff
@@ -120,80 +113,73 @@ class TaskTypeDAO @Inject()(sqlClient: SqlClient)(implicit ec: ExecutionContext)
...
-                              ${t.settings.volumeInterpolationAllowed},
-                              ${t.settings.mergerMode},
-                              #${optionLiteral(t.settings.resolutionRestrictions.min.map(_.toString))},
-                              #${optionLiteral(t.settings.resolutionRestrictions.max.map(_.toString))},
-                              #${optionLiteral(t.recommendedConfiguration.map(c => sanitize(Json.toJson(c).toString)))},
-                              '#${t.tracingType.toString}',
-                              ${t.created}, ${t.isDeleted})
-                       """)
+                    values(${t._id}, $organizationId, ${t._team}, ${t.summary}, ${t.description},
+                           ${SqlToken.raw(s"${enumArrayLiteral(t.settings.allowedModes)}")},
```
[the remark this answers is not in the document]
fm3: I could not figure out a way to get `setParameter` for `Array[Enumeration#Value]` to work. I created follow-up issue https://github.com/scalableminds/webknossos/issues/6736 for this

app/models/user/MultiUser.scala
```diff
@@ -54,105 +53,100 @@ class MultiUserDAO @Inject()(sqlClient: SqlClient)(implicit ec: ExecutionContext
...
-                       isSuperUser, novelUserExperienceInfos, selectedTheme, created, isDeleted)
-                     values(${u._id}, ${u.email}, '#${sanitize(u.passwordInfo.hasher)}', ${u.passwordInfo.password},
-                      ${u.isSuperUser}, '#$novelUserExperienceInfosString', '#${u.selectedTheme}',
-                     ${u.created}, ${u.isDeleted})
-          """)
+      _ <- run(q"""insert into webknossos.multiusers(_id, email, passwordInfo_hasher,
+                                                     passwordInfo_password,
+                                                     isSuperUser, novelUserExperienceInfos, selectedTheme,
+                                                     created, isDeleted)
+                   values(${u._id}, ${u.email}, ${SqlToken.raw(escapeLiteral(u.passwordInfo.hasher))},
```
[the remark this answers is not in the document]
fm3: On the scala side, this is a string, while on the sql side it is an enum, which cannot be passed. Again, I’d suggest to schedule a follow-up issue https://github.com/scalableminds/webknossos/issues/6737 (If you think it is important, I can also do this here already)

app/utils/sql/SqlTypeImplicits.scala
```diff
@@ -1,35 +1,110 @@
...
+
+  implicit def jsValueToSqlValue(v: JsValue): SqlValue = JsonValue(v)
+
+  implicit def optionToSqlValue[T](v: Option[T])(implicit innerConversion: T => SqlValue): SqlValue =
+    v match {
+      case Some(inner) => innerConversion(inner)
+      case None        => NoneValue()
+    }
+
+  implicit def stringIterableToSqlValue(v: Iterable[String]): SqlValue = ArrayValue(v.toList)
```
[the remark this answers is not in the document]
fm3: I had this originally, and it took the enum array mentioned above. However, setParamter then failed with a type mismatch. So I’d rather catch the type mismatch already in the scala compiler, which is why I restricted this to String for the moment.

app/utils/sql/SqlTypeImplicits.scala
fm3: I am unsure about this. It would work with tokens, but in the tuple method, tokens that are more complex than single values don’t really make sense. With this two-step conversion, the signature of tuple can take List[SqlValue] and ensure nothing more complex is passed. But I agree that the additional conversions are also not great.

normanrz: ok, fine.

app/utils/sql/SqlTypeImplicits.scala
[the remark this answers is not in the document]
normanrz: fair enough. i think we should have versions of this for all used types of arrays, then.

app/models/job/Job.scala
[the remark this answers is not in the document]
normanrz: no strong opinion. `existingCollectionName` is probably harder to miss than a single `_`. consistency would be nice, tho. most inserts should probably be replaced by helper functions #6730.

app/utils/sql/SqlTypeImplicits.scala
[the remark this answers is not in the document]
fm3: I agree! VARCHAR[] and webknossos.TASKTYPE_MODES[] are the only ones right now. Please have a look at the lastest commit 63c06a2ab to see if you think my way of passing the sql enum name in there is ok, even though it does not work as implicit conversion.

app/utils/sql/SqlInterpolation.scala
fm3: How important do you think this is? My concern is that one might accidentally use tuple instead of tupleFromList. It will compile fine (as List[SqlValue] has an implicit conversion to SqlValue), but will not perform as expected. I currently see no usages where tuple is called in this way.

The nested one seems more relevant, but there this trick doesn’t work :grimacing:

normanrz: I think that would improve the readability. I used that in #6732 for INSERTs, e.g. https://github.com/scalableminds/webknossos/pull/6732/files#diff-a2ec134373baf5603901ac361985d256c48f4f0e17025a96c4e5d6f0fec98c5fR945-R955, and for inline tuples, e.g. https://github.com/scalableminds/webknossos/pull/6732/files#diff-a2ec134373baf5603901ac361985d256c48f4f0e17025a96c4e5d6f0fec98c5fR464.

We can debate naming, of course.

`tupleLists` could be replaced by `SqlToken.join(list.map(i => SqlToken.tuple(i._1, i._2)), ",")`. Could `Iterable[SqlToken]` be extended to have a `toTupleList` function (with implicit magic)?

app/models/task/TaskType.scala
```diff
@@ -120,80 +113,73 @@ class TaskTypeDAO @Inject()(sqlClient: SqlClient)(implicit ec: ExecutionContext)
...
-                           settings_resolutionRestrictions_max = #$resolutionMaxLiteral,
-                           recommendedConfiguration = #$configurationLiteral,
-                           isDeleted = ${t.isDeleted}
-                          where _id = ${t._id.id}""")
+      _ <- run(q"""UPDATE webknossos.taskTypes
+                   SET
+                   _team = ${t._team},
+                   summary = ${t.summary},
+                   description = ${t.description},
+                   settings_allowedModes = ${EnumerationArrayValue(t.settings.allowedModes, "webknossos.TASKTYPE_MODES")},
```
normanrz, line 167: Would it be possible to add an sqlTypeName attribute to the enumeration classes? Or make dedicated *toSqlValue functions? I would prefer to not have this boilerplate in the queries.

fm3: I tried that but the only solution I could figure out involved [this](https://stackoverflow.com/questions/44390775/scala-reflection-get-class-of-enumaration-value) private method to access the Enum class from the value object. I’d say that is not worth it at this time. I’d defer finding a better solution here, and since it task type allowed modes is the only occurence of enum-arrays in postgres, let’s keep it like this for the moment.

app/utils/sql/SqlInterpolation.scala
fm3: Fair enough. I also renamed tuple to tupleFromValues to explicitly distinguish it from tupleFromLists.

normanrz (approved):

PR #6739, opened 2023-01-13
Rebranding

normanrz:
- Adds the new WEBKNOSSOS logo
- Changes the primary color to #5660FF
- Changes brand name from webKnossos to WEBKNOSSOS

- https://rebranding.webknossos.xyz

### Steps to test:
- Click around on https://rebranding.webknossos.xyz

------
- [x] Ready for review


frontend/stylesheets/dark.less
```diff
@@ -1,6 +1,10 @@
 @import "../../node_modules/antd/dist/antd.dark.less";
 @import "./main.less";
 
+body {
+  background-color: #000;
```
philippotto, line 5: To quote from the Refactoring UI pdf:

> True black tends to look pretty unnatural, so start with a *really* dark grey and work your way up to white in steady increments.

I think the same goes for a white background. Maybe you feel like toning these down a tiny notch.

normanrz: This just fixes an inconsistency (see below). Tuning the background is a bigger job. 
Screenshot 2023-01-13 at 12 00 37

fm3:
Favicon seems to be #6456ff, not #5660FF, is that intended?

normanrz:
> Favicon seems to be #6456ff, not #5660FF, is that intended?

That shouldn't be the case. Where are you seeing that?

jstriebel:
Just found this gray text which I found hard to read in dark-mode, it's okayish in light-mode (might not even be affected by the new colors):
Screenshot from 2023-01-13 13-34-39
Screenshot from 2023-01-13 13-37-09

jstriebel:
I'm not quite sure if Nunito is used correctly, I couldn't find it in the sources. Maybe move it to fonts, next to the Titilium woff2 files?

fm3:
> Where are you seeing that?

image

(favicon purple-ish, big svg bluer)

philippotto:
Regarding the new font: it still doesn't seem to be used on my machine. Or is this expected?

image

normanrz:
For me it loads and uses the network font, now.
Screenshot 2023-01-13 at 15 08 00

philippotto:
I found some spots where old colors are still used:

- in the dataset folder view the left sidebar's background is still the old blue when dragging a dataset
- the active entries in the tree list and segment list view in an annotation

The rest is looking good :+1:

normanrz:
> The only thing I noticed is that the margins now seem different for multi-line tags. Not sure which change caused this but on master, this doesn't seem to be the case:
> 
> image
> 
> new vs vs
> 
> image
> 
> But this doesn't need to block the PR in my opinion.

How about this?
Screenshot 2023-01-17 at 09 45 03

philippotto:
> How about this?

Yes, this looks good :+1:

PR #6745, opened 2023-01-16
Purge last usages of SqlToken.raw with user input

fm3:
- Introduces two new enum classes to pass to sql (note that they are not used in the scala models as their usages are in library code, which expects strings)
- Introducs `SqlToken.joinBySeparator` which is here used with the separator AND to construct the dataset search query
- The remaining usages of SqlToken.raw contain no user-supplied data (construction of column lists, sql view name existingCollectionName)

- https://avoidsqlraw.webknossos.xyz

### Steps to test:
- Search for a dataset
- Register a new user
- Should work

------
- [x] Ready for review

### DIFF
--- a/app/utils/sql/SqlInterpolation.scala
+++ b/app/utils/sql/SqlInterpolation.scala
@@ -86,8 +86,8 @@ object SqlToken {
 
   def raw(s: String): SqlToken = SqlToken(s)
 
-  def joinBySeparator(tokens: Iterable[SqlToken], sepator: String): SqlToken =
-    SqlToken(sql = tokens.map(_.sql).mkString(sepator), values = tokens.flatMap(_.values).toList)
+  def joinBySeparator(tokens: Iterable[SqlToken], separator: String): SqlToken =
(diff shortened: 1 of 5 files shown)


normanrz (approved):

app/utils/sql/SqlInterpolation.scala
```diff
@@ -106,6 +86,9 @@ object SqlToken {
 
   def raw(s: String): SqlToken = SqlToken(s)
 
+  def joinBySeparator(tokens: Iterable[SqlToken], sepator: String): SqlToken =
+    SqlToken(sql = tokens.map(_.sql).mkString(sepator), values = tokens.flatMap(_.values).toList)
```
normanrz, line 90:
```suggestion
  def joinBySeparator(tokens: Iterable[SqlToken], separator: String): SqlToken =
    SqlToken(sql = tokens.map(_.sql).mkString(separator), values = tokens.flatMap(_.values).toList)
```

Issue #4018, opened 2019-04-15
Texture count is not used per layer in fragment shader


philippotto:
Instead, the maximum is used which works per se, but can cause unnecessary branching --> performance.

PR #6748, opened 2023-01-17
Prepare multi modality support

### LINKED ISSUE #6085: Support multi-modality datasets
## Detailed Description
In webKnossos all layers of a dataset must have the same voxel resolution (aka "scale). When imaging a dataset with different modalities (e.g., EM and LM), this constraint is usually not met. Supporting these different modalities in one dataset would be a good feature to have.

One idea for the implementation would be to have one scene transform per modality (or per layer). Currently, there is only one global scene scale which would need to be split up.

Other thoughts:
- a first version might be restricted so that only one modality is visible at a given time (and switching between the modalities switches the global scale)
- some sort of semantic zoom would be probably be a nice feature: when zooming out, one sees LM data and when zooming in, modalities with higher resolutions appear (e.g., CT and then EM)

Not sure whether the backend would need to do something here, too?

Quoting from https://github.com/scalableminds/webknossos/issues/4026#issuecomment-485691651:

> For this, we would need to figure out how to unify the addressing for the different scales. Right now, addressing data positions is done via voxel coordinates (see user-exposed position UI). In case, of differing scales, this would not work, anymore. Due to backwards compatibility (also for NMLs), I assume that we would need to come up with some (inelegant) workaround for this problem.

## Todo

- [x] gather example data (i.e., at least two DS layers with different scales and potentially affine transforms)
- [x] do an implementation spike get a better picture

## Context

philippotto:
### Summary
- Supports rendering a dataset where layers can be transformed each (with an affine matrix).
- Transformation can be specified via datasource-properties.json or via JS API (will be ephemeral, then)
- Rendering was re-architectured to use cuckoo hashing for the dynamic look up of bucket addresses.
- Progressive rendering is done on the GPU (instead of on the CPU).
- For performance reasons, the plane geometries are now tessellated and the vertices are mutated by the vertex shader so that they are bucket-aligned. This enables us to precompute the cuckoo lookup in the vertex shader.
- The pre-computation is disabled when interpolation is enabled (would cause problems at chunk borders, otherwise). Also as a workaround for issues at chunk borders, pre-computation is ignored there, too (independent of interpolation).
- Transformed layers won't be rendered using the above optimization (since their buckets won't align with the vertices).
- Since enabling interpolation can incur a performance cost, it's disabled by default (won't affect old preferences) and the user will get a small hint in the left sidebar if it's enabled.
- As a consequence of the individually transformed layers, it can happen that one layer is rendered in a different magnification than another layer. This used to be uniform, but isn't anymore now (even though, it should still be uniform when no transformations are used).

- https://multimodality.webknossos.xyz

### To do:
- [X] allow to define transforms per layer which are used in the shader
- [X] adapt bucket picking to respect layer transformations
- [X] allow to make the active mag dependent on each layer (one layer can be rendered in mag 1 and the other in mag 4)
- [X] remove anchor point mechanism in favor of hashing to address buckets on the GPU
- [X] Backend: Store matrix from datasource-properties.json in postgres, include in dataset json sent to frontend
- Restore old behavior/functionality:
  - [X] progressive rendering
  - [x] data that is still being loaded should be gray instead of black
  - [x] "The currently rendered quality is not optimal due to the available magnifications and the viewport arrangement." shows false positively
  - [x] find my data needs to be adapted.
- Fix bugs:
  - [x] having more layers in a DS than can be rendered at the same time results in rendering issues
  - [x] rotation from arbitrary mode is not removed when switching back to orthogonal mode
  - [x] deep freeze breaks for transform matrices apparently
  - [X] thin lines at bucket borders (solved by disabling anti aliasing which prevents flat varyings to be interpolated errorenously)
  - [X] data offset by 32vx
  - [X] missing data at border
  - [X] weird triangles at border
  - [x] sometimes bucket strip shows at viewport boundary shows wrong data
  - [x] some transformations cause incorrect data to be rendered (needs investigation)
  - [x] XZ and YZ rendering is corrupt (probably the computation in the vertex shader is not adapted yet)
  - [x] sometimes cross hairs are invisible (probably due to disabled antialiasing)
  - [X] support segmentation layers again
  - [x] fix line artifact when rendering segmentation layer (this is caused by the segmentation patterns which rely on partial derivatives; unfortunately, these seem to be buggy now that there's a special tessellation)
- [X] Performance (solved by performing bucket look up in vertex shader)
- [ ] clean up UI
  - [x] display correct mag etc
  - [x] fix CI tests
  - [x] fix screenshot tests
  - [x] add warning for performance for interpolation?
  - [x] disable performance optimization when interpolation is enabled
  - [x] change default for interpolation
  - [ ] doublecheck bucket loading order for transformed layers
- [x] go through to do comments in code

### Steps to test:
- test view mode and annotation mode for multiple datasets
- pay attention to performance and rendering issues

### Issues:
- fixes #6085

------
- [x] Ready for review


frontend/javascripts/oxalis/model_initialization.ts
```diff
@@ -222,6 +224,122 @@ export async function initialize(
     setInitialTool();
   }
 
+  const makeTranslation = (x: number, y: number, z: number): Matrix4x4 =>
```
philippotto, line 227: move into own module?

frontend/javascripts/oxalis/shaders/segmentation.glsl.ts
```diff
@@ -273,12 +262,15 @@ export const getSegmentationId: ShaderModule = {
...
         getSegmentIdOrFallback(
-          <%= segmentationName %>_lookup_texture,
+          lookup_texture,
           <%= formatNumberAsGLSLFloat(colorLayerNames.length + layerIndex) %>,
           <%= segmentationName %>_data_texture_width,
           <%= formatNumberAsGLSLFloat(packingDegreeLookup[segmentationName]) %>,
           worldPositionUVW,
-          vec4(0.0, 0.0, 0.0, 0.0)
+          vec4(0.0, 0.0, 0.0, 0.0),
+          // todo
```
philippotto, line 271: todo

frontend/javascripts/oxalis/shaders/texture_access.glsl.ts
```diff
@@ -37,6 +39,15 @@ export const getRgbaAtIndex: ShaderModule = {
           )
         ).rgba;
     }
+
+    uvec4 getUnsignedRgbaAtIndex(highp usampler2D dtexture, float textureWidth, float idx) {
```
philippotto, line 43: use overloading instead of separate name?

frontend/javascripts/oxalis/shaders/texture_access.glsl.ts
```diff
@@ -104,56 +200,66 @@ export const getColorForCoords: ShaderModule = {
...
-          relativeBucketPosition.x < 0.0 ||
-          relativeBucketPosition.y < 0.0 ||
-          relativeBucketPosition.z < 0.0) {
-        // In theory, the current magnification should always be selected
-        // so that we won't have to address data outside of the addresSpaceDimensions.
-        // Nevertheless, we explicitly guard against this situation here to avoid
-        // rendering wrong data.
-        returnValue[1] = vec4(1.0, 1.0, 0.0, 1.0);
+      if (worldPositionUVW.x < 0. || worldPositionUVW.y < 0. || worldPositionUVW.z < 0.) {
+        // should alpha be -1.0?
```
philippotto, line 204: todo

frontend/javascripts/oxalis/geometries/plane.ts
```diff
@@ -4,20 +4,23 @@ import { getBaseVoxelFactors } from "oxalis/model/scaleinfo";
...
-import type { OrthoView, Vector3, Vector4 } from "oxalis/constants";
+import type { OrthoView, Vector3 } from "oxalis/constants";
 import constants, {
   OrthoViewColors,
   OrthoViewCrosshairColors,
   OrthoViewGrayCrosshairColor,
   OrthoViewValues,
 } from "oxalis/constants";
 
+export const PLANE_SUBDIVISION = 200;
```
philippotto, line 15: use in shader

frontend/javascripts/oxalis/shaders/main_data_fragment.glsl.ts
```diff
@@ -251,5 +282,254 @@ void main() {
...
+  }
+
+  flatVertexPos = worldCoordUVW;
+  float NOT_YET_COMMITTED_VALUE = pow(2., 21.) - 1.;
+  const float bucketWidth = 32.;
+  const float bucketSize = 32.*32.*32.;
+
+  const uint FALLBACK_COUNT = 3u;
+
+  // todo: seg layer is missing!
```
philippotto, line 493: todo

frontend/javascripts/oxalis/shaders/main_data_fragment.glsl.ts
```diff
@@ -251,5 +282,254 @@ void main() {
...
+  vUv = uv;
+  modelCoord = vec4(position, 1.0);
+  savedModelMatrix = modelMatrix;
+mat4 modelInv = inverseMatrix(modelMatrix);
+  worldCoord = modelMatrix * vec4(position, 1.0);
+
+  gl_Position = projectionMatrix * modelViewMatrix * vec4(position, 1.0);
+
+  float planeWidth = 376.;
+  float subdivisionCount = 200.;
```
philippotto, line 419: dont use hardcoded values

frontend/javascripts/oxalis/model/accessors/flycam_accessor.ts
```diff
@@ -312,32 +364,35 @@ export function getValidTaskZoomRange(
...
+  // }
+
+  // const min = getMinMax(resolutionRestrictions.min, true);
+  // const max = getMinMax(resolutionRestrictions.max, false);
+  // return [min, max];
 }
 export function isMagRestrictionViolated(state: OxalisState): boolean {
   const { resolutionRestrictions } = state.tracing.restrictions;
-  const zoomStep = getRequestLogZoomStep(state);
+  // todo: the resolution restrictions need to be stored per layer (backend + frontend).
```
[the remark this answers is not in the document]
philippotto: @fm3 How would you gauge the necessary effort to store the resolution restrictions per layer? I think, this PR should be merged soon and a follow-up one could tackle this, but I'd still be interested in your thoughts.

frontend/javascripts/oxalis/geometries/materials/plane_material_factory.ts
```diff
@@ -455,6 +489,35 @@ class PlaneMaterialFactory {
...
+            const hasTransform = !_.isEqual(layer.transformMatrix || Identity4x4, Identity4x4);
+            this.uniforms[`${name}_has_transform`] = {
+              value: hasTransform,
+            };
+            if (representativeLayerIdxForMag == null && !hasTransform) {
+              representativeLayerIdxForMag = layerIdx;
+            }
+          }
+
+          // If all layers have a transform, the representativeLayerIdxForMag
```
daniel-wer, line 525: Could you briefly explain what the `representativeLayerIdxForMag` is needed/used for?

frontend/javascripts/oxalis/geometries/materials/plane_material_factory.ts
```diff
@@ -697,17 +763,18 @@ class PlaneMaterialFactory {
     // layers were least-recently activated (but are now disabled).
     // The first array contains the color layer names and the second the segmentation layer names.
```
daniel-wer, line 777: What does the third return value represent? From the code it looks like it is the overall number of layers, but then, shouldn't that also be returned in line 766?

frontend/javascripts/oxalis/geometries/materials/plane_material_factory.ts
```diff
@@ -759,17 +826,25 @@ class PlaneMaterialFactory {
     );
   };
 
-  getFragmentShader(): string {
+  getFragmentShaderWithUniforms(): [string, Uniforms] {
     const { initializedGpuFactor, maximumLayerCountToRender } =
       Store.getState().temporaryConfiguration.gpuSetup;
     // Don't compile code for segmentation in arbitrary mode
```
daniel-wer, line 845: This comment no longer seems relevant (even before your change) :thinking:

frontend/javascripts/oxalis/geometries/materials/plane_material_factory.ts
```diff
@@ -780,24 +855,36 @@ class PlaneMaterialFactory {
...
-void main() {
-  vUv = uv;
-  modelCoord = vec4(position, 1.0);
-  savedModelMatrix = modelMatrix;
-  worldCoord = modelMatrix * vec4(position, 1.0);
-  gl_Position = projectionMatrix * modelViewMatrix * vec4(position, 1.0);
-}`;
+    const { initializedGpuFactor, maximumLayerCountToRender } =
+      Store.getState().temporaryConfiguration.gpuSetup;
+    // Don't compile code for segmentation in arbitrary mode
```
daniel-wer, line 880: See my comment above.

frontend/javascripts/oxalis/geometries/materials/plane_material_factory.ts
```diff
@@ -780,24 +855,36 @@ class PlaneMaterialFactory {
...
+    const datasetScale = dataset.dataSource.scale;
+    const lookupTextureWidth = getLookupBufferSize(initializedGpuFactor);
+
+    return getMainVertexShader({
+      globalLayerCount,
+      colorLayerNames,
+      segmentationLayerNames,
+      packingDegreeLookup,
+      // Todo: this is not computed per layer. See #4018
+      dataTextureCountPerLayer: Model.maximumTextureCountForLayer,
```
daniel-wer, line 895: In https://github.com/scalableminds/webknossos/issues/4018 you mention that this could improve performance if it was "fixed". Just wanted to mention it in case you had forgotten about it (like me), no need to do it in this PR.

frontend/javascripts/oxalis/geometries/plane.ts
daniel-wer: Maybe you could also add a comment on what the rationale behind this number is. It's probably not important that it's exactly 200, but explaining why it isn't 20 or 2000 might help future readers :)

frontend/javascripts/oxalis/model/bucket_data_handling/abstract_cuckoo_table.ts
```diff
@@ -0,0 +1,303 @@
...
+      CUCKOO_TWIDTH: this.textureWidth,
+    };
+  }
+
+  flushTableToTexture() {
+    this._texture.update(this.table, 0, 0, this.textureWidth, this.textureWidth);
+  }
+
+  checkValidKey(key: K) {
+    if (key === EMPTY_KEY) {
```
daniel-wer, line 123: There are a couple of usages of `EMPTY_KEY` in this class. Shouldn't they use `getEmptyKey()`?

frontend/javascripts/oxalis/model/bucket_data_handling/abstract_cuckoo_table.ts
```diff
@@ -0,0 +1,303 @@
...
+      THREE.RGBAIntegerFormat,
+    );
+    cachedNullTexture.internalFormat = "RGBA32UI";
+
+    return cachedNullTexture;
+  }
+
+  private initializeTableArray() {
+    this.table = new Uint32Array(AbstractCuckooTable.ELEMENTS_PER_ENTRY * this.entryCapacity).fill(
+      EMPTY_KEY_VALUE,
```
daniel-wer, line 77: I'm stumbling about this line, since a subclass could decide not to use `EMPTY_KEY_VALUE` and instead return something else in `getEmptyValue()`. If you use that instead, there should be a check whether `ELEMENTS_PER_ENTRY > 1` and if so that all entries of the empty key are the same, before using `getEmptyValue()[0]` to fill the array.

daniel-wer (commented):
First part of my review :) I'll continue tomorrow.

frontend/javascripts/oxalis/geometries/plane.ts
philippotto: good point :) I just thought the value through (the initial value was more a gut feeling) and decreased it to 80 now (and added the explanation of course).

frontend/javascripts/oxalis/geometries/materials/plane_material_factory.ts
```diff
@@ -232,6 +226,14 @@ class PlaneMaterialFactory {
       this.uniforms[`${layerName}_unrenderable`] = {
         value: 0,
       };
+      const layer = getLayerByName(Store.getState().dataset, dataLayer.name);
+
+      this.uniforms[`${layerName}_transform`] = {
+        value: invertAndTranspose(layer.transformMatrix || Identity4x4),
```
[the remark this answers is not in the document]
philippotto: I'm not 100% sure what you mean, but it might be correct :smile: I added a comment to the `invertAndTranspose` function which explains its existence. Copied here so that you don't need to look it up:

> // Transposition is often needed so that the matrix has the right format
// for matrix operations (e.g., on the GPU; but not for ThreeJS).
// Inversion is needed when the position of an "output voxel" (e.g., during
// rendering in the fragment shader) needs to be mapped to its original
// data position (i.e., how it's stored without the transformation).
// Without the inversion, the matrix maps from stored position to the position
// where it should be rendered.

frontend/javascripts/oxalis/shaders/texture_access.glsl.ts
```diff
@@ -86,16 +88,85 @@ export const getColorForCoords: ShaderModule = {
...
+      return float(address);
+    }
+
+    float lookUpBucket(uint globalLayerIndex, uvec4 bucketAddress, bool supportsPrecomputedBucketAddress) {
+      // The fragment shader can read the entry that was
+      // calculated by the vertex shader. However, this won't always
+      // be precise because the triangles of the plane won't necessarily
+      // align with the bucket borders.
+      <% if (isFragment) { %>
+        if (supportsPrecomputedBucketAddress) {
```
daniel-wer, line 148: I did not find an explanation for this parameter. I expect it to be related to the triangle border artifacts you experienced in an intermediate version. Is that correct?

frontend/javascripts/oxalis/shaders/texture_access.glsl.ts
```diff
@@ -104,56 +175,66 @@ export const getColorForCoords: ShaderModule = {
...
-        return returnValue;
+      uint globalLayerIndex = availableLayerIndexToGlobalLayerIndex[uint(localLayerIndex)];
+      uint activeMagIdx = uint(activeMagIndices[int(globalLayerIndex)]);
+
+      float bucketAddress;
+      vec3 offsetInBucket;
+      uint renderedMagIdx;
+
+      if (!supportsPrecomputedBucketAddress) {
+        for (uint i = 0u; i < 4u; i++) {
```
daniel-wer, line 192: Is the `< 4` coming from the maximum allowed zoom step diff that is still rendered? If so, I'd reuse the respective constant via templating.

frontend/javascripts/oxalis/controller/scene_controller.ts
```diff
@@ -450,13 +465,21 @@ class SceneController {
...
     // Only set the visibility of the dataset bounding box for the TDView.
     // This has to happen before updateForCam is called as otherwise cross section visibility
     // might be changed unintentionally.
     this.datasetBoundingBox.setVisibility(id !== OrthoViews.TDView || tdViewDisplayDatasetBorders);
     this.datasetBoundingBox.updateForCam(id);
     this.userBoundingBoxes.forEach((bbCube) => bbCube.updateForCam(id));
+    this.layerBoundingBoxes.forEach((bbCube) => {
+      const visible = id !== OrthoViews.TDView || tdViewDisplayLayerBorders;
+      bbCube.setVisibility(visible);
+      if (visible) {
```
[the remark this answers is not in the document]
philippotto: good point! I just adapted the visibility of the bboxes to the layers and this was indeed a bug :+1: (however, the old code at least respected the tdViewDisplayLayerBorders setting successfully for some reason).

frontend/javascripts/oxalis/shaders/main_data_fragment.glsl.ts
```diff
@@ -134,15 +154,28 @@ ${compileShader(
...
-    return;
+
+  if (floor(flatVertexPos.x) == floor(worldCoordUVW.x) && floor(flatVertexPos.y) == floor(worldCoordUVW.y)) {
+    gl_FragColor = vec4(1., 0., 1., 1.);
+    // return;
+  }
+
+  if (index.x == 200.) {
+      gl_FragColor = vec4(0., 1., 1., 1.);
+//      return;
```
daniel-wer, line 165: Looks like remains from testing that can be removed

frontend/javascripts/oxalis/shaders/main_data_fragment.glsl.ts
```diff
@@ -251,5 +284,249 @@ void main() {
...
+  uniform float <%= name %>_alpha;
+  uniform float <%= name %>_gammaCorrectionValue;
+  uniform float <%= name %>_unrenderable;
+  uniform mat4 <%= name %>_transform;
+  uniform bool <%= name %>_has_transform;
+<% }) %>
+
+uniform float activeMagIndices[<%= globalLayerCount %>];
+uniform uint availableLayerIndexToGlobalLayerIndex[<%= globalLayerCount %>];
+uniform vec3 resolutions[20];
```
daniel-wer, line 339: The 20 probably shouldn't be hard-coded

frontend/javascripts/oxalis/shaders/main_data_fragment.glsl.ts
```diff
@@ -251,5 +284,249 @@ void main() {
       `vec3(${vector3.map(formatNumberAsGLSLFloat).join(", ")})`,
     OrthoViewIndices: _.mapValues(OrthoViewIndices, formatNumberAsGLSLFloat),
     hasSegmentation,
+    isFragment: true,
+  });
+}
+
+export function getMainVertexShader(params: Params) {
```
daniel-wer, line 291: The name of the file suggests that it only contains a fragment shader. It should be adapted.

frontend/javascripts/oxalis/shaders/main_data_fragment.glsl.ts
```diff
@@ -251,5 +284,249 @@ void main() {
...
+uniform float brushSizeInPixel;
+uniform float planeID;
+uniform vec3 addressSpaceDimensions;
+uniform vec4 hoveredSegmentIdLow;
+uniform vec4 hoveredSegmentIdHigh;
+
+const float bucketWidth = <%= bucketWidth %>;
+const float bucketSize = <%= bucketSize %>;
+const float l_texture_width = <%= l_texture_width %>;
+
```
daniel-wer, line 386: Does the above section mirror the whole variable declarations of the fragment shader?

frontend/javascripts/oxalis/shaders/main_data_fragment.glsl.ts
```diff
@@ -251,5 +284,249 @@ void main() {
...
+  // - the second last vertex is moved so that it's positioned at the beginning of the last
+  //   rendered bucket. If this isn't handled additionally, the second last vertex might be
+  //   still correct in most cases, but at certain positions, the second last vertex is
+  //   already in the second last bucket and would be moved to far to the left.
+  //   Example:
+  //   *------|------|------|----a-|---*
+  //   ^ first vertex
+  //          ^ second vertex
+  //                   ...             ^ last vertex
+  //                             ^ if a is the second last vertex, it needs to be moved to the right
```
daniel-wer, line 441: Doesn't the same issue exist for the third last vertex then, sometimes. That it maybe would need to be moved to the right, as otherwise the second last bucket would not be covered? (and so on...)

frontend/javascripts/oxalis/shaders/main_data_fragment.glsl.ts
```diff
@@ -251,5 +284,249 @@ void main() {
...
+  //                               instead of the left.
+
+  // Calculate the index of the vertex (e.g., index.x=0 is the first horizontal vertex).
+  index = (position.xy / (PLANE_WIDTH / 2.) + 1.) / 2. * PLANE_SUBDIVISION;
```
daniel-wer, line 445: Does it work to use the `VIEWPORT_WIDTH` constant for that? Don't the orthogonal viewport possibly have wildly varying sizes?
Also, I'm failing to follow the calculation I have to admit :see_no_evil:

frontend/javascripts/oxalis/model/accessors/flycam_accessor.ts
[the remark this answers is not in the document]
philippotto: for the record: this only affects task annotations. for now, I added a heuristic which uses the first color layer to read the active mag. for layers without transforms this doesn't matter, anyway.

frontend/javascripts/oxalis/controller/combinations/skeleton_handlers.ts
```diff
@@ -268,7 +268,8 @@ function addNode(
       position,
       rotation,
       OrthoViewToNumber[Store.getState().viewModeData.plane.activeViewport],
-      getRequestLogZoomStep(Store.getState()),
+      // todo: this would be a breaking change. clarify.
```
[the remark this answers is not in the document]
philippotto: see https://scm.slack.com/archives/C5AKLAV0B/p1677079138825069 for discussion.

frontend/javascripts/oxalis/shaders/main_data_fragment.glsl.ts
```diff
@@ -251,5 +284,249 @@ void main() {
...
+
+  uint activeMagIdx = uint(activeMagIndices[representativeLayerIdxForMag]);
+  vec2 d = transDim(vec3(32.) * getResolution(activeMagIdx)).xy;
```
daniel-wer, line 448: Could you clarify what `d` is? Maybe we should have a look at this code together if it's easier :)

frontend/javascripts/oxalis/shaders/main_data_fragment.glsl.ts
```diff
@@ -251,5 +284,249 @@ void main() {
...
+  // Offset the bucket calculation for the current vertex by a voxel to ensure
+  // that the provoking vertex (the one that is used by the flat varyings in
+  // the corresponding triangle) looks up the correct bucket. Otherwise,
+  // a rendering offset of 32 vx occurs.
+  worldCoordUVW.x -= 1.;
+  worldCoordUVW.y += 1.;
+
+  flatVertexPos = worldCoordUVW;
+  float NOT_YET_COMMITTED_VALUE = pow(2., 21.) - 1.;
+  const float bucketWidth = 32.;
```
daniel-wer, line 486: I'd suggest to use `BUCKET_WIDTH` via templating here as well. Same for the FALLBACK_COUNT in line 489.

daniel-wer (commented):
I've gone through all the files once, now :) Very impressive work! Most of the changes were really easy to follow :+1: 

With testing, I'd wait until the first round of feedback is addressed and most ToDos have been taken care of.

frontend/javascripts/oxalis/model_initialization.ts
philippotto: I moved it into the front-end api.

frontend/javascripts/oxalis/controller/combinations/move_handlers.ts
```diff
@@ -45,9 +47,10 @@ export const moveW = (deltaW: number, oneSlide: boolean): void => {
   }
 
   if (oneSlide) {
-    const logZoomStep = getRequestLogZoomStep(Store.getState());
+    const activeMagIndices = getActiveMagIndicesForLayers(Store.getState());
+    const finestMagIdx = _.min(values(activeMagIndices)) || 0;
     const wDim = Dimensions.getIndices(activeViewport)[2];
-    const wStep = getResolutions(Store.getState().dataset)[logZoomStep][wDim];
+    const wStep = getResolutions(Store.getState().dataset)[finestMagIdx][wDim];
```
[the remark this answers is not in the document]
philippotto: > On a different note, maybe it's more intuitive if this only takes into account layers that are currently visible? Otherwise, if the layer with the finest resolution is not visible, the step might be too small and there will be no change in the viewports.

This is a good point :+1: 

> I'm not sure I understand how this continues to work 
My current assumption is that there are layers with varying scales and, therefore, the one with the finest resolution not necessarily is the one with the smallest voxel size.
Instead, it's likely the case that the resolutions refer to a base scale - in which case it does make sense again.

Let's discuss this in-person :)

frontend/javascripts/oxalis/geometries/materials/plane_material_factory.ts
philippotto: I think the case where `0` is returned doesn't really matter as nothing will be rendered, anyway, but it's probably less confusing to return the global layer count. I'll adapt the code (and add a comment) :)

frontend/javascripts/oxalis/model/bucket_data_handling/abstract_cuckoo_table.ts
philippotto: You are right! I guess, my generalization of these classes got a bit too flexibel (which wouldn't have worked if the subclasses were to push the limits). I cleaned this up (mainly by making the restrictions clearer in the comments + interface). See 7e4db9587.

frontend/javascripts/oxalis/shaders/main_data_fragment.glsl.ts
philippotto: I added a better comment to the code. I hope this makes it clearer :) See de948413f

frontend/javascripts/oxalis/model_initialization.ts
```diff
@@ -222,6 +224,122 @@ export async function initialize(
...
+
+  const setRotation = (layerName: string, _theta: number, pos: Vector3 | null = null) => {
+    const theta = (_theta / 360) * 2 * Math.PI;
+    if (pos == null) {
+      pos = [3496, 3379, 0];
+    }
+    setTransforms(layerName, makeRotation(theta, pos));
+  };
+
+  type Spec =
```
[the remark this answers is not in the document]
philippotto: I put them into the frontend api for now and marked them as experimental (I also slimmed it down a bit). Hope that's okay!

frontend/javascripts/test/puppeteer/dataset_rendering.screenshot.ts
```diff
@@ -1,3 +1,4 @@
+import "test/mocks/lz4";
```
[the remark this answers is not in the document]
philippotto: Good question :see_no_evil: I assume that the import dependencies changed somehow so that the lz4 dependency is imported now (which fails on node). I didn't bother to investigate further. Hope this is alright!

frontend/javascripts/oxalis/model/reducers/settings_reducer.ts
```diff
@@ -178,9 +179,14 @@ function SettingsReducer(state: OxalisState, action: Action): OxalisState {
...
       if (allowedModes.includes(action.viewMode)) {
-        return updateTemporaryConfig(state, {
+        const newState = updateTemporaryConfig(state, {
           viewMode: action.viewMode,
         });
+        if (action.viewMode !== "orthogonal") {
+          return newState;
+        }
+        // Restore rotation
+        return setRotationReducer(newState, [0, 0, 0]);
```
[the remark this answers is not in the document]
philippotto: The rotation is only changed if the user does so in flight/oblique mode. Since this affects the matrix (which is also used in orthogonal mode), the rotation needs to be reset. Does this make sense? I also extended the comment a bit.

frontend/javascripts/oxalis/model/bucket_data_handling/layer_rendering_manager.ts
```diff
@@ -105,21 +92,42 @@ function consumeBucketsFromArrayBuffer(
   return bucketsWithPriorities;
 }
 
+export function getGlobalLayerIndexForLayerName(
+  layerName: string,
+  optSanitizer?: (arg: string) => string,
+): number {
+  const sanitizer = optSanitizer || _.identity;
+  const allColorLayers = getColorLayers(Store.getState().dataset);
```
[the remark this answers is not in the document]
philippotto: good point! I don't know why I implemented it so convoluted. thanks :)

frontend/javascripts/oxalis/model/bucket_data_handling/layer_rendering_manager.ts
```diff
@@ -175,58 +186,56 @@ export default class LayerRenderingManager {
...
+      M4x4.mul(layerMatrix, state.flycam.currentMatrix),
     );
-    const areas = getAreasFromState(state);
-    const matrix = getZoomedMatrix(state.flycam);
+
     const { viewMode } = state.temporaryConfiguration;
-    const isArbitrary = constants.MODES_ARBITRARY.includes(viewMode);
     const { sphericalCapRadius } = state.userConfiguration;
     const isVisible = isLayerVisible(dataset, this.name, datasetConfiguration, viewMode);
+    const rects = getViewportRects(Store.getState());
```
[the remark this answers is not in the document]
philippotto: I think, the "areas" are a concept which are a bit outdated now. They basically contain top, left, right, bottom in bucket coordinates. This means that the areas didn't necessarily change when one moved only by one voxel (since the same buckets were at the border of the viewport). With the transformation of layers, this doesn't really make much sense anymore, though. The oblique bucket picker needs the exact extent of the rendered viewport to calculate its things (one could probably use a similar optimization technique, though, but I don't feel like this would get us much).

The areas are only used by the prefetching saga now. This could use some overhauling, too, but obviously it's out of scope for this PR.

frontend/javascripts/oxalis/model/bucket_data_handling/texture_bucket_manager.ts
```diff
@@ -3,40 +3,41 @@ import _ from "lodash";
...
 // Buckets which are not needed anymore will be replaced by other buckets.
 // A bucket is considered "active" if it is supposed to be in the data texture.
 // A bucket is considered "committed" if it is indeed in the data texture.
 // Active buckets will be pushed into a writerQueue which is processed by
 // writing buckets to the data texture (i.e., "committing the buckets").
-// At the moment, we store two floats per bucket.
-// The first value f denotes the following:
-//   If f >= 0, f denotes the index in the data texture where the bucket is stored.
-//   If f == -1, the bucket is not yet committed
-//   If f == -2, the bucket is not supposed to be rendered. Out of bounds.
```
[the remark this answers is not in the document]
philippotto: The cuckoo table now accomplishes these cases as follows:
f >= 0 ==> [key, address] is in table
f == -1 ==> [key, NOT_YET_COMMITTED_VALUE] is in table
f == -2 ==> key doesn't exist in table

frontend/javascripts/oxalis/shaders/texture_access.glsl.ts
philippotto: `supportsPrecomputedBucketAddress` is true if a layer doesn't have special transforms and if interpolation is not used. Only in these cases, the bucket address that was precomputed in the vertex shader can be used.

frontend/javascripts/oxalis/geometries/materials/plane_material_factory.ts
philippotto: Good point! I added the following comment:

> // The vertex shader looks up the buckets for rendering so that the
// fragment shader doesn't need to do so. Currently, this only works
// for layers that don't have a transformation (otherwise, the differing
// grids wouldn't align with each other).
// To align the vertices with the buckets, the current magnification is
// needed. Since the current mag can differ from layer to layer, the shader
// needs to know which mag is safe to use.
// For this purpose, we define the representativeLayerIdxForMag which is
// essentially an index to *some* layer that doesn't have any transforms.
// If all layers have a transform, the representativeLayerIdxForMag
// isn't relevant which is why it can default to 0.

frontend/javascripts/oxalis/shaders/main_data_fragment.glsl.ts
philippotto: good point! I could dry most of it :)

frontend/javascripts/oxalis/model/bucket_data_handling/bucket_picker_strategies/oblique_bucket_picker.ts
```diff
@@ -3,103 +3,162 @@ import type { EnqueueFunction } from "oxalis/model/bucket_data_handling/layer_re
...
+      resolutions,
+      logZoomStep,
+    );
+    const steps = stepRateBuckets.length + 1;
+    const stepSize = [extent[0] / steps, extent[1] / steps];
+    // This array holds the start and end points
+    // of horizontal lines which cover the entire rendered plane.
+    // These "scan lines" are traversed to find out which buckets need to be
+    // sent to the GPU.
+    const zDiff = 10;
```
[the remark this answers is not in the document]
philippotto: The coordinates are essentially in "screen space" and multiplied with the `queryMatrix`. So, a static value makes somewhat sense in my opinion. However, I'm sure this could be improved somehow. However, since I'm not sure how, I'd leave it as is for the moment :smile:

frontend/javascripts/oxalis/model/bucket_data_handling/layer_rendering_manager.ts
```diff
@@ -45,19 +44,7 @@ const asyncBucketPick: typeof asyncBucketPickRaw = memoizeOne(
...
-
-  const roundToNearestBucketBoundary = (pos: Vector3, dimension: 0 | 1 | 2) => {
-    const bucketExtentInVoxel = constants.BUCKET_WIDTH * resolution[dimension];
-    // Math.round returns 0 or 1 which will be mapped to -1 or 1
-    return Math.round(modAndDivide(pos[dimension], bucketExtentInVoxel)) * 2 - 1;
-  };
-
-  return map3((_pos, idx) => roundToNearestBucketBoundary(position, idx), position);
-}
+const getLookUpCuckooTable = memoizeOne(() => new CuckooTableVec5(256));
```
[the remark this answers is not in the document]
philippotto: Yes, it's a lazy-initialized singleton. I renamed it to `getSharedLookUpCuckooTable` and added a comment about it being a lazy singleton.

frontend/javascripts/oxalis/shaders/main_data_shaders.glsl.ts
```diff
@@ -470,8 +415,11 @@ void main() {
   }
 
   worldCoord = vec4(transDim(transWorldCoord), 1.);
-  vec3 posRec = (modelInv * worldCoord).xyz;
-  gl_Position = projectionMatrix * modelViewMatrix * vec4(posRec, 1.0);
+  new_worldCoordUVW = getWorldCoordUVW();
+  debugVal = getWorldCoordUVW();
+
+  gl_Position = projectionMatrix * viewMatrix * worldCoord;
+  gl_Position.z = originalZ;
```
daniel-wer, line 389: Maybe you could add a small comment here, explaining that due to floating point inaccuracies z might change unintendedly, otherwise

frontend/javascripts/oxalis/shaders/segmentation.glsl.ts
```diff
@@ -271,12 +271,18 @@ export const getSegmentationId: ShaderModule = {
 
   <% _.each(segmentationLayerNames, function(segmentationName, layerIndex) { %>
     vec4[2] getSegmentationId_<%= segmentationName %>(vec3 worldPositionUVW) {
-      vec4[2] volume_color =
+      vec4[2] volume_color;
+      vec3 transformedCoordUVW = transDim((<%= segmentationName %>_transform * vec4(transDim(worldPositionUVW), 1.0)).xyz);
+      if (isOutsideOfBoundingBox(transformedCoordUVW)) {
+        return volume_color;
```
daniel-wer, line 277: `volume_color` is uninitialized at this point, is the returned value deterministic in that case or should we initialize it to be sure?

frontend/javascripts/test/model/binary/data_rendering_logic.spec.ts
```diff
@@ -1,10 +1,11 @@
-// @ts-nocheck
```
daniel-wer, line 1: :tada:

frontend/javascripts/test/puppeteer/dataset_rendering.screenshot.ts
```diff
@@ -137,13 +152,18 @@ const datasetConfigOverrides: Record<string, PartialDatasetConfiguration> = {
...
+    interpolation: true,
+  },
 };
 
 async function withRetry(
   retryCount: number,
   testFn: () => Promise<boolean>,
   resolveFn: (arg0: boolean) => void,
 ) {
+  // todo: revert
```
daniel-wer, line 165: Todo comment

frontend/javascripts/oxalis/controller/combinations/move_handlers.ts
```diff
@@ -48,6 +48,9 @@ export const moveW = (deltaW: number, oneSlide: boolean): void => {
 
   if (oneSlide) {
     const activeMagIndices = getActiveMagIndicesForLayers(Store.getState());
+    // The following logic might not always make sense when having layers
```
daniel-wer, line 51: Another candidate for the followup (limitation documentation) issue

daniel-wer (commented):
Looking very good, please see my additional comments :)

- [x] Quite a few of the screenshots are missing the cross hair in one or more of the data viewports. I think this was fixed in the mean time, so it's probably enough to re-generate the screenshots
- [x] What do you think about adding a transformed dataset to the screenshot tests?

If I remember correctly, the team already did some monkey testing. What do you think remains to be tested or should we do a final monkey testing round in the end?

frontend/javascripts/oxalis/shaders/segmentation.glsl.ts
philippotto: good question! just tested it and it should be safe :)

frontend/javascripts/oxalis/controller/scene_controller.ts
```diff
@@ -474,12 +475,13 @@ class SceneController {
     this.datasetBoundingBox.setVisibility(id !== OrthoViews.TDView || tdViewDisplayDatasetBorders);
     this.datasetBoundingBox.updateForCam(id);
     this.userBoundingBoxes.forEach((bbCube) => bbCube.updateForCam(id));
-    this.layerBoundingBoxes.forEach((bbCube) => {
-      const visible = id !== OrthoViews.TDView || tdViewDisplayLayerBorders;
+    const layerNameToIsDisabled = getLayerNameToIsDisabled(Store.getState().datasetConfiguration);
+    Object.keys(this.layerBoundingBoxes).forEach((layerName) => {
+      const bbCube = this.layerBoundingBoxes[layerName];
+      const visible =
+        id === OrthoViews.TDView && tdViewDisplayLayerBorders && !layerNameToIsDisabled[layerName];
```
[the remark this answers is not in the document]
philippotto: Yeah, I think, I went that route because the new layer bboxes are yellow now (to have at least a bit of a distinguishing characteristic) and if the layer bbox == dataset bbox, the borders would be yellow in the data viewports which felt a bit dominant.

frontend/javascripts/oxalis/model/accessors/flycam_accessor.ts
```diff
@@ -552,3 +566,34 @@ function _getUnrenderableLayerInfosForCurrentZoom(
 export const getUnrenderableLayerInfosForCurrentZoom = reuseInstanceOnEquality(
   _getUnrenderableLayerInfosForCurrentZoom,
 );
+
+function _getActiveResolutionInfo(state: OxalisState) {
```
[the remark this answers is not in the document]
philippotto: Both callers of this function (status bar and info tab view) make proper use of the return values :)

frontend/javascripts/oxalis/model/accessors/flycam_accessor.ts
```diff
@@ -372,33 +374,46 @@ export function getValidTaskZoomRange(
 ): [number, number] {
   const defaultRange = [baseDatasetViewConfiguration.zoom.minimum, Infinity] as Vector2;
   const { resolutionRestrictions } = state.tracing.restrictions;
+  // We use the first color layer as a heuristic to check the validity of the zoom range,
+  // as we don't know to which layer a restriction is meant to be applied.
+  // If the layers don't have any transforms, the layer choice doesn't matter, anyway.
```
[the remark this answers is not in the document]
philippotto: Good point! I opened https://github.com/scalableminds/webknossos/issues/6926.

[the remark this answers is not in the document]
philippotto: Good point! I opened https://github.com/scalableminds/webknossos/issues/6926.

philippotto:
@daniel-wer I incorporated your remaining feedback and also updated the screenshots. I'll add an additional screenshot test for the transformations in a follow-up PR. If you accept the PR, I'll merge it today :)

frontend/javascripts/oxalis/api/api_latest.ts
```diff
@@ -1970,9 +1970,14 @@ class DataApi {
   }
 
   /*
-   * _Experimental_ API for creating transformation matrices based on an array of TransformerSpecs.
+   * _Experimental_ API for creating transformation matrices based on an array of TransformerSpec's.
```
daniel-wer, line 1973:
```suggestion
   * _Experimental_ API for creating transformation matrices based on an array of TransformerSpecs.
```

daniel-wer (approved):
LGTM :tada:

PR #6749, opened 2023-01-17
Adapt color of multiuser switch indicator

fm3:
image

did not see a way to use the less variable for this :grimacing:

### DIFF
--- a/frontend/javascripts/navbar.tsx
+++ b/frontend/javascripts/navbar.tsx
@@ -162,20 +162,7 @@ function UserInitials({
       </Avatar>
 
       {isMultiMember ? (
-        <SwapOutlined
-          style={{
-            position: "absolute",
-            top: 2,
-            right: -5,
-            marginRight: 0,
-            minWidth: 12,
-            height: 12,
-            lineHeight: "12px",
-            fontSize: 12,
-            color: "#59F8E8",
-          }}
-          title="You are member of multiple organizations. Click the avatar to switch between them."
-        />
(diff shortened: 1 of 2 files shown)


frontend/javascripts/navbar.tsx
```diff
@@ -172,7 +172,7 @@ function UserInitials({
             height: 12,
             lineHeight: "12px",
             fontSize: 12,
-            color: "#75df4a",
+            color: "#59F8E8",
```
philippotto, line 175: could you put the style into a css class into `main.less`? Then, you can use `@blue-zircon` (compare with the `.ant-avatar` class).

frontend/javascripts/navbar.tsx
```diff
@@ -162,20 +162,7 @@ function UserInitials({
...
-            marginRight: 0,
-            minWidth: 12,
-            height: 12,
-            lineHeight: "12px",
-            fontSize: 12,
-            color: "#75df4a",
-          }}
-          title="You are member of multiple organizations. Click the avatar to switch between them."
-        />
+        <SwapOutlined title="You are member of multiple organizations. Click the avatar to switch between them." />
```
philippotto, line 165: thanks for the extraction! one last thing: could you add a `className="swap-account-icon"` (or another name) and use that class name in the less file? the current way would also change other swap-icons if they are used (now or in the future).

philippotto (approved):
Excellent, thanks for fixing this!

PR #6754, opened 2023-01-17
Fix font import

normanrz:
- [x] Ready for review


philippotto (approved):
I don't know the `tech(variations)` stuff, but you probably tested it. For the record: where was the font import broken? For me, it already worked..

normanrz:
`tech(variations)` doesn't seem to be necessary, but broke the font import on Safari.

PR #6756, opened 2023-01-17
Fix superuser organization switching

fm3:
At this point the user access ctx does not have update access for its own new identity. This code needs to run with global access context.

Also in this PR: incrase threshold for slow request slack notifications to 30s (we mostly ignored anything faster anyway, it just became spam)

### Steps to test:
- setup multiorga setup, create second orga, as superuser of orga 1, try to join orga 2, should work (I tested this locally)

------
- [x] Ready for review


philippotto (approved):
:+1:

PR #6759, opened 2023-01-18
Speed Up Dataset List Request

### LINKED ISSUE #6089: speed up dataset list request
We should do some profiling and optimize the performance for the dataset list request.

### LINKED ISSUE #6773: Dataset Folder Tab: Public datasets are not shown if user has otherwise no access
Datasets (even of their own organization) that a registered user can only see because the datasets are public are shown in the legacy datasets tab but not in the new dataset tab.

fm3:
- Slightly speeds up the dataset list request by moving the filter logic to composed SQL.
- An optional boolean parameter `compact` can be used to get a more compact response with less information. This speeds up the route significantly more, as the complete response is constructed with just one SQL query.
- the optional `isEditable` GET parameter in the dataset list request has been removed (it was, to our best knowledge, unused)
- The front-end does not yet use this new `compact` version, but this is planned, cc @philippotto 
- Note that the work by @normanrz in the now superseded #6766 paved the way for this PR :)
- Also slipped in here: bugfix for folder access query (make visible folders that include public datasets)
- Also slipped in here: bugfix for annotation list total count header

- https://fastdatasetlist.webknossos.xyz

### Steps to test:
- query http://localhost:9000/api/datasets?compact=true
- should be fast (compared to no compact=true)
- should still work with filters like isActive, searchQuery, folderId, … (except isEditable, that one has been removed)
- isEditable boolean field should be correct (that one needs most scrutiny I’d say)

### Issues:
- fixes #6089
- fixes #6773

------
- [x] Ready for review


app/models/binary/DataSet.scala
```diff
@@ -164,8 +185,117 @@ class DataSetDAO @Inject()(sqlClient: SqlClient,
...
+                  (u.isAdmin AND u._organization = d._organization) OR
+                  u.isDatasetManager OR
+                  d._uploader = u._id OR
+                  d._id IN (              -- team manager of team that has access to the dataset
+                    SELECT _dataSet
+                    FROM webknossos.dataSet_allowedTeams dt
+                    JOIN webknossos.user_team_roles utr ON dt._team = utr._team
+                    WHERE utr._user = u._id AND utr.isTeamManager
+                  ) OR
+                  d. _folder IN (        -- team manager of team that has (cumulative) access to dataset folder
```
normanrz, line 247:
```suggestion
                  d._folder IN (        -- team manager of team that has (cumulative) access to dataset folder
```

app/models/binary/DataSet.scala
```diff
@@ -164,8 +185,117 @@ class DataSetDAO @Inject()(sqlClient: SqlClient,
...
+              o.name,
+              d._folder,
+              d.isUsable,
+              d.displayName,
+              d.created,
+              COALESCE(
+                (
+                  (u.isAdmin AND u._organization = d._organization) OR
+                  u.isDatasetManager OR
+                  d._uploader = u._id OR
```
normanrz, line 240: We may need to revise that in the future. There needs to be a way to revoke access to a dataset even for the person that happens to have uploaded the data.

app/models/binary/DataSet.scala
```diff
@@ -164,8 +185,117 @@ class DataSetDAO @Inject()(sqlClient: SqlClient,
...
+      query = q"""
+            SELECT
+              d._id,
+              d.name,
+              o.name,
+              d._folder,
+              d.isUsable,
+              d.displayName,
+              d.created,
+              COALESCE(
```
normanrz, line 236: Btw this could also be implemented as a view: a view of all dataset+user combinations with an attribute whether the user can read/write the dataset. The benefit of a view would be that we can easily examine it with pgadmin et al and it could potentially be used by multiple queries.

normanrz (approved):
Very good stuff. Looking forward to seeing the perf improvements!

I am wondering if a complex query like this (or parts of it) might be better placed in a view for reusability and debuggability.

app/models/binary/DataSet.scala
fm3: If you agree we can just remove this from the relevant places now. In fact, to be allowed to upload (to a folder), you need to be admin/dataset manager or folder team manager already. So this change will not affect many (and there it may actually be intended)

app/models/binary/DataSet.scala
fm3: I agree about the views. However, we need to figure out a way to handle them during migrations. The existing userInfos view is already frequently a source of trouble (needs to be dropped + recreated in evolutions). I would not like to have to do this for two more views all the time :grimacing:  Maybe there is a way to get around this.

normanrz: What's the problem with dropping/recreating?

fm3: Not a conceptual problem, but time and again we wrote evolutions with only their semantics in mind, then saw them fail because the view handling was not yet included. It of course makes the evolutions a lot bigger as well. None of this totally blocks this, but I would only go for it if we see concrete benefit from creating more views

normanrz:
> What are your feelings on this? Do you know if there are still users querying the Rest API directly, without our libs?

I would mention it in migration guide and break it. I'd say that the REST API is only officially supported through our official libraries.

app/models/binary/DataSet.scala
fm3: I created https://github.com/scalableminds/webknossos/issues/6780 – I decided not to do it right away because I need to double-check that this does not break the uploading process itself

PR #6761, opened 2023-01-18
When re-activating user, check orga user limit

### LINKED ISSUE #6757: Enforce user limits when activating users
## Context
In the admin>users page, you can reactivate users. This can lead to you going over your organization's limit of included users. The backend doesn't to yet enforce/prevent this.

## Expected Behavior
You should not be able activate users if you are already at the limit of your orga's users.

## Current Behavior
You can activate all users without any checks and run over your limit.

## Your Environment for bug

- Browser name and version: all browsers
- Operating System and version: all OS
- Version of WebKnossos (Release or Commit): Version: 21170

fm3:
### Steps to test:
- for an orga, register several users
- in db, set user limit
- deactivate some users so that the orga user count is within limits, then try to re-activate one, should be blocked
- after relaxing the limit in the db, it should be allowed again

### Issues:
- fixes #6757

------
- [x] Ready for review


app/models/organization/OrganizationService.scala
```diff
@@ -115,7 +115,7 @@ class OrganizationService @Inject()(organizationDAO: OrganizationDAO,
     def sendRPCToDataStore(dataStore: DataStore) =
       rpc(s"${dataStore.url}/data/triggers/newOrganizationFolder")
         .addQueryString("token" -> dataStoreToken, "organizationName" -> organizationName)
-        .post
+        .post()
```
frcroth, line 118: ?

frcroth (approved):

app/models/organization/OrganizationService.scala
fm3: scala style guide recommends adding () to parameter-less function calls iff the function contain side effects. Intellij reminded me that this is the case here

PR #6770, opened 2023-01-20
Loki for voxelytics logs

normanrz:
- Replaces elasticsearch with loki for vx log storage
- Changes the log frontend to only show the latest 1000 log entries

Screenshot 2023-01-20 at 20 09 03

- https://lokilogs.webknossos.xyz

### Steps to test:
- Run a vx workflow against https://lokilogs.webknossos.xyz and check the logs

------
- [x] Ready for review


normanrz:
Not sure whether to show the latest log messages first or last 🤔

app/models/voxelytics/LokiClient.scala
```diff
@@ -0,0 +1,173 @@
...
+      _ <- serverStartupFuture
+      res <- rpc(s"${conf.uri}/loki/api/v1/query_range?$queryString").silent.getWithJsonResponse[JsValue]
+    } yield
+      JsArray(
+        (res \ "data" \ "result")
+          .as[List[JsValue]]
+          .flatMap(
+            stream =>
+              (stream \ "values")
+                .as[List[(String, String)]]
```
fm3, line 120: Did you check what happens if the received json does not match the expected schema? Since this happens down here in the yield statement, it may become a bare exception that bubbles to the user. It might be worth it to move this up to the for part and wrap it with a tryo or one of the json parsing convenience functions that we already have (those may not suit this nested processing, though)

app/models/voxelytics/LokiClient.scala
```diff
@@ -0,0 +1,173 @@
...
+                    "timestamp" -> Instant(value._1.substring(0, value._1.length - 6).toLong)))))
+  }
+
+  def bulkInsert(logEntries: List[JsValue], organizationId: ObjectId): Fox[Unit] =
+    if (logEntries.nonEmpty) {
+      val streams = logEntries
+        .groupBy(
+          entry => ((entry \ "vx" \ "workflow_hash").as[String], (entry \ "vx" \ "run_name").as[String])
+        )
+        .map(kv =>
```
fm3, line 132:
```suggestion
        .map(keyValueTuple =>
```

app/models/voxelytics/LokiClient.scala
```diff
@@ -0,0 +1,173 @@
...
+          Json.obj(
+            "stream" -> Json.obj("vx_workflow_hash" -> kv._1._1,
+                                 "vx_run_name" -> kv._1._2,
+                                 "wk_url" -> wkConf.Http.uri,
+                                 "wk_org" -> organizationId.id),
+            "values" -> JsArray(kv._2.map(entry => {
+              val timestamp = java.time.Instant.parse((entry \ "@timestamp").as[String]).toEpochMilli
```
fm3, line 139: Could this use our Instant class as well? please add the missing functions like toNanoseconds

fm3 (commented):
Backend mostly LGTM, I added some small comments :)

util/src/main/scala/com/scalableminds/util/time/Instant.scala
```diff
@@ -11,10 +12,15 @@ case class Instant(epochMillis: Long) extends Ordered[Instant] {
   override def toString: String = DateTimeFormatter.ISO_INSTANT.format(toJavaInstant)
 
   def toJavaInstant: java.time.Instant = java.time.Instant.ofEpochMilli(epochMillis)
+
   def toJodaDateTime: org.joda.time.DateTime = new org.joda.time.DateTime(epochMillis)
+
   def toSql: java.sql.Timestamp = new java.sql.Timestamp(epochMillis)
 
+  def toNanosecondsString: String = s"${epochMillis}000000"
```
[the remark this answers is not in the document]
normanrz: Long would overflow in 2262

fm3 (commented):
Backend LGTM :) Leaving final approval for frontend review

conf/application.conf
```diff
@@ -281,9 +281,8 @@ swagger.api.info {
 
 voxelytics {
   staleTimeout = 20 minutes
-  elasticsearch {
+  loki {
```
daniel-wer, line 284: Should this appear in the migration guide or in the breaking changes section of the changelog?

frontend/javascripts/admin/voxelytics/log_tab.tsx
```diff
@@ -143,23 +144,34 @@ export default function LogTab({
...
 }: {
   workflowHash: string;
   runId: string;
   taskName: string;
   isRunning: boolean;
+  beginTime: Date | null;
+  endTime: Date | null;
 }) {
-  const [showTimestamps, setShowTimestamps] = useState(false);
+  const [showTimestamps, setShowTimestamps] = useState(true);
```
daniel-wer, line 157: Thank you :pray:

frontend/javascripts/admin/voxelytics/log_tab.tsx
```diff
@@ -143,23 +144,34 @@ export default function LogTab({
...
   const [isFullscreen, setIsFullscreen] = useState(false);
   const [isLoading, setIsLoading] = useState(true);
-  const [level, setLevel] = useState("DEBUG");
+  const [level, setLevel] = useState(LOG_LEVELS.INFO);
```
daniel-wer, line 160: Did this work? On the deployed dev instance, the default log level still seems to be `DEBUG`.

frontend/javascripts/admin/voxelytics/log_tab.tsx
```diff
@@ -192,6 +204,24 @@ export default function LogTab({
     }
   }, [logResult, showTimestamps]);
 
+  async function downloadFullLog() {
```
daniel-wer, line 207: I think this is fine for now, but we'll have to see whether this is feasible for very long logs. The alternative would be for the server to serve this log file in a streaming fashion, for direct download. Probably, voxelytics should log less if this becomes problematic, though :sweat_smile:

frontend/javascripts/admin/voxelytics/task_list_view.tsx
```diff
@@ -370,10 +371,8 @@ export default function TaskListView({
         message.error("Please select a specific run for log download.");
```
daniel-wer, line 371: It would be helpful to show this as a tooltip when the button is disabled, instead (or too).

frontend/javascripts/admin/voxelytics/utils.ts
```diff
@@ -63,3 +65,29 @@ export async function copyToClipboad(text: string) {
...
+      runId,
+      taskName,
+      minLevel,
+      startTime,
+      currentEndTime,
+      batchSize,
+    );
+    if (batch.length === 0) break;
+    buffer = batch.concat(buffer);
+    currentEndTime = new Date(batch[batch.length - 1].timestamp - 1);
```
daniel-wer, line 90: Shouldn't `startTime` be updated here as well?

daniel-wer (commented):
This is a nice improvement! :+1: Maybe we can test this with a run that logs a little bit more?

frontend/javascripts/admin/voxelytics/log_tab.tsx
normanrz: I just redeployed the dev instance.

frontend/javascripts/admin/voxelytics/log_tab.tsx
normanrz: 💣

frontend/javascripts/admin/voxelytics/utils.ts
normanrz: I don't think that is necessary. The `startTime` stays at the minimum `beginTime ?? new Date(0)`, because the logs are fetch in backward order from the `currentEndTime`.

conf/application.conf
normanrz: Yes. I fully forgot to add changelog and migration guide. Added now.

frontend/javascripts/admin/voxelytics/utils.ts
daniel-wer: Ah I see, this works because of the limit when requesting the logs. Maybe the `getVoxelyticsLogs` function should contain a comment indicating that the limit is counted from `endTime`, so that the latest logs are fetched first, to avoid such confusion.

normanrz: I added a comment :+1:

normanrz:
> Maybe we can test this with a run that logs a little bit more?

I'll let the test suite run tonight.

daniel-wer (approved):
LGTM if the tests run through successfully :+1:

normanrz:
@fm3 Would you mind having another look at my latest commits? I moved the batched-fetching from the frontend to the backend.

fm3 (commented):
LGTM :)
Do you think it is likely that we will want to change the constants POLLING_INTERVAL, LOG_TIME_BATCH_INTERVAL, LOG_ENTRY_BATCH_SIZE again soon? If so, maybe they should be extracted to the config? But if you think this unlikely, feel free to leave as is

normanrz:
@daniel-wer here is the dev instance with a longer run: https://lokilogs.webknossos.xyz/workflows/e8185485dc?runId=63cfcd920100008b00668fc9

daniel-wer:
> @daniel-wer here is the dev instance with a longer run: https://lokilogs.webknossos.xyz/workflows/e8185485dc?runId=63cfcd920100008b00668fc9

Thanks, looking good, but I noticed two things:
- [ ] Sometimes text is overlaid on top of each other (seems to only happen for lines that are a tiny bit too long for a single line)
  overlaid_text_bug
- [ ] The last couple of lines of the train model log (and other tasks), actually belong to the next task
  - ```
    2023-01-24T12:26:07.587Z NOTICE   voxelytics 

    🚀🚀🚀 Executing evaluate_segmentation@predict_batch
    2023-01-24T12:26:07.588Z NOTICE   voxelytics Metadata for the current workflow execution can be found at 
    runs/segem_test_until_predict.2023-01-24_13-22-19
    ```

normanrz:
> * Sometimes text is overlaid on top of each other (seems to only happen for lines that are a tiny bit too long for a single line)
>   overlaid_text_bug

That is probably a problem because of the scrollbar, which changes the width of the container. Hard for me to debug on Mac.

> * The last couple of lines of the train model log (and other tasks), actually belong to the next task
> ```
>   2023-01-24T12:26:07.587Z NOTICE   voxelytics 
>   
>   🚀🚀🚀 Executing evaluate_segmentation@predict_batch
>   2023-01-24T12:26:07.588Z NOTICE   voxelytics Metadata for the current workflow execution can be found at 
>   runs/segem_test_until_predict.2023-01-24_13-22-19
>   ```

That is actually a vx problem. Not sure how to fix it.

normanrz:
Since we only show 1000 lines max, I decided to get rid of virtualized log list. This should solve the overlapping line problem. @daniel-wer would you mind having another look? https://lokilogs.webknossos.xyz/workflows/e8185485dc?runId=63cfcd920100008b00668fc9

daniel-wer (approved):
Nice, LGTM :tada:

PR #6775, opened 2023-01-23
Access Google Cloud Storage via NIO

### LINKED ISSUE #6626: Support Zarr/N5 streaming from Google Cloud Storage
## Detailed Description
Zarr/N5 streaming should support Google Cloud Storage as an another data source in addition to streaming from HTTPs / S3. 

- Public datasets on GCP should be supported
- Private dataset with proper authentication should be supported
- URIs starting with `gs:// ...` should be supported.

fm3:
- Adds support for remote datasets hosted on Google Cloud Storage `gs://` with optional `GoogleServiceAccountCredential`s
- Refactor NIO usage: do not get file system using NIO lookup by schema, but handle this explicitly. This results in a lot less magic and more direct and typesafe passing of different kinds of credentials to the FileSystems. Also no need for the META-INF files anymore :tada: 
- Note: the paths passed to the FileSystem.getPath are no longer full URIs but only paths inside of the bucket scope. The GCS one requires this, and the others support it as well. This has to be converted back to URIs when passing it to MagLocator
- Slightly refactor handling of credentials, rename some fields.
- Use asynchronous caching for file system creation to avoid duplication due to parallel requests
- try to gunzip data returned by NIO file system (I was surprised to randomly receive some gzipped data for a gcs bucket)

- https://googlecloud.webknossos.xyz

### TODO
 - [x] access GCS data anonymously
 - [x] access GCS data with service account credential json
 - [x] integrate GCS file system into managed file systems creation
 - [x] unify file system handling
 - [x] clean up credential naming, enums
 - [x] re-check that s3 and https credentials and uri styles still work (rework so that paths never contain url schema and host/bucket?)
 - [x] add route to create GCS credential
 - [x] adapt front-end for uploading service account credential json
 - [x] remove fast-start changes to application.conf

### Steps to test:
- Front-end is not ready yet, but this feature can already be tested by pasting a `gs://` zarr uri and ignoring the front-ends warnings
- Adding credential is also possible by using whichever secret field (and entering a non-empty arbitrary string into the left username/keyid field)
- also test that s3 and https datasets can still be explored and viewed, with and without credentials.

### Issues:
- fixes #6626

------
- [x] Ready for review


frontend/javascripts/admin/dataset/dataset_add_zarr_view.tsx
```diff
@@ -311,7 +311,6 @@ function AddZarrLayer({
       (userInput.indexOf("https://") !== 0 && userInput.indexOf("s3://") === 0)
     ) {
       setSelectedProtocol(userInput.indexOf("https://") === 0 ? "https" : "s3");
-      setShowCredentialsFields(userInput.indexOf("s3://") === 0);
```
fm3, line 314: note that all supported schemas now support their form of credentials so this is no longer needed

fm3:
@frcroth I’d appreciate it if you could already have a look at the backend changes even though the frontend part is not yet complete. @philippotto agreed to adapt the front-end in the coming days.

webknossos-datastore/app/com/scalableminds/webknossos/datastore/storage/FileSystemCredentials.scala
```diff
@@ -1,23 +1,50 @@
...
+                                 secretAccessKey: String,
+                                 user: String,
+                                 organization: String)
+    extends FileSystemCredential
 
 object S3AccessKeyCredential {
   implicit val jsonFormat: OFormat[S3AccessKeyCredential] = Json.format[S3AccessKeyCredential]
 }
+
+case class GoogleServiceAccountCredential(name: String, secretJson: JsValue, user: String, organization: String)
```
frcroth, line 29: I am not sure about the naming here. In the enum CredentialType there are HTTP_Basic_Auth and S3_Access_Key, which have the same name as the matching case classes (but with Credential) added here, while the Credential is called GoogleServiceAccountCredential and the type is called `CredentialType.GCS`. Maybe rename the enum type if this is the correct name?

app/models/binary/explore/ExploreRemoteLayerService.scala
```diff
@@ -138,25 +140,27 @@ class ExploreRemoteLayerService @Inject()(credentialService: CredentialService)
...
+      uri <- tryo(new URI(normalizeUri(layerUri))) ?~> s"Received invalid URI: $layerUri"
+      credentialOpt = credentialService.createCredentialOpt(uri,
+                                                            credentialIdentifier,
+                                                            credentialSecret,
+                                                            requestingUser._id,
+                                                            requestingUser._organization)
+      remoteSource = RemoteSourceDescriptor(uri, credentialOpt)
+      credentialId <- Fox.runOptional(credentialOpt)(c => credentialService.insertOne(c)) ?~> "Failed to store file system credential"
+      fileSystem <- FileSystemsHolder.getOrCreate(remoteSource) ?~> "Failed to set up remote file system"
+      remotePath <- tryo(fileSystem.getPath(FileSystemsHolder.pathFromUri(remoteSource.uri))) ?~> "Failed to get remote path"
```
frcroth, line 157: Move to messages?

util/src/main/scala/com/scalableminds/util/io/ZipIO.scala
```diff
@@ -145,6 +145,27 @@ object ZipIO extends LazyLogging {
...
   }
 
+  def tryGunzip(possiblyCompressed: Array[Byte]): Array[Byte] =
+    tryo(gunzip(possiblyCompressed)).toOption.getOrElse(possiblyCompressed)
+
+  def gunzip(compressed: Array[Byte]): Array[Byte] = {
+    val is = new GZIPInputStream(new ByteArrayInputStream(compressed))
+    val os = new ByteArrayOutputStream()
+    try {
+      val buffer = new Array[Byte](1024)
```
frcroth, line 155: Any reason for this specific buffer length?

webknossos-datastore/app/com/scalableminds/webknossos/datastore/s3fs/S3FileSystemProvider.java
```diff
@@ -37,8 +37,8 @@
 /**
  * Spec:
  * <p>
- * URI: s3://[endpoint]/{bucket}/{key} If endpoint is missing, it's assumed to
- * be the default S3 endpoint (s3.amazonaws.com)
+ * URI: s3://[user@]{bucket}/{key}.
```
frcroth, line 40: Does the dot at the end mean end of statement or include a dot in the URL?

fm3: good point, I removed it to avoid confusion. No dot in the URLs please

util/src/main/scala/com/scalableminds/util/io/ZipIO.scala
fm3: I just stole it from the gzip method. Since all bytes should already be in memory in this case, the buffer size matters little. I believe powers of two are recommended if the data source or destination is on disk, to match whole multiples of the disk block size, but here it does not really matter that much. If you have any more insight about the ideal value here, please let me know :)

philippotto:
@fm3 I just pushed the front-end changes. I hope it makes sense the way I did it. I didn't have a gs URL + credentials on my hand, so didn't really test it e2e, but the content of a dummy JSON were sent properly, I think :) Let me know if you need anything else!

fm3:
@philippotto Great, thanks! Yes, the json is sent correctly :) Compare https://scm.slack.com/archives/C5AKLAV0B/p1675339672375929?thread_ts=1674548652.080749&cid=C5AKLAV0B for access to example data with credentials.

I noticed two small things:
 -  The label of the dropzone says „(Optional)“, which seems a bit redundant since above there is already the radio selection between anonymous and with credential. Maybe this could be removed (or remove the radio selection and make the auth fields optional for all cases?)
 - If I press reset after a dataset has been recovered, the status which protocol was selected seems to be lost. The gs uri is still in the field, but the auth fields are showing the basicauth case. Not super important, since the reset button will probably be used rarely :shrug:

philippotto:
> @philippotto Great, thanks! Yes, the json is sent correctly :) Compare https://scm.slack.com/archives/C5AKLAV0B/p1675339672375929?thread_ts=1674548652.080749&cid=C5AKLAV0B for access to example data with credentials.

Perfect :+1: Exploration works, but the data fetch requests don't really work for me. The front-end waits forever since the requests don't finish (until they'll probably time out). The console says:

```
java.lang.NoClassDefFoundError: Could not initialize class org.blosc.IBloscDll
        at com.scalableminds.webknossos.datastore.datareaders.BloscCompressor.cbufferSizes(Compressor.scala:254)
        at com.scalableminds.webknossos.datastore.datareaders.BloscCompressor.uncompress(Compressor.scala:240)
        at com.scalableminds.webknossos.datastore.datareaders.ChunkReader.$anonfun$readBytes$2(ChunkReader.scala:38)
        at scala.Option.map(Option.scala:230)
        at com.scalableminds.webknossos.datastore.datareaders.ChunkReader.$anonfun$readBytes$1(ChunkReader.scala:35)
        at scala.util.Using$Manager.scala$util$Using$Manager$$manage(Using.scala:171)
        at scala.util.Using$Manager$.$anonfun$apply$2(Using.scala:225)
        at scala.util.Try$.apply(Try.scala:213)
        at scala.util.Using$Manager$.apply(Using.scala:225)
        at com.scalableminds.webknossos.datastore.datareaders.ChunkReader.readBytes(ChunkReader.scala:34)
        at com.scalableminds.webknossos.datastore.datareaders.ChunkReader.read(ChunkReader.scala:31)
        at com.scalableminds.webknossos.datastore.datareaders.DatasetArray.$anonfun$getSourceChunkDataWithCache$1(DatasetArray.scala:96)
        at akka.http.caching.LfuCache$.$anonfun$toJavaMappingFunction$2(LfuCache.scala:97)
        at scala.compat.java8.functionConverterImpls.AsJavaBiFunction.apply(FunctionConverters.scala:41)
        at com.github.benmanes.caffeine.cache.LocalAsyncCache.lambda$get$2(LocalAsyncCache.java:92)
        at com.github.benmanes.caffeine.cache.BoundedLocalCache.lambda$doComputeIfAbsent$14(BoundedLocalCache.java:2413)
        at java.util.concurrent.ConcurrentHashMap.compute(ConcurrentHashMap.java:1853)
        at com.github.benmanes.caffeine.cache.BoundedLocalCache.doComputeIfAbsent(BoundedLocalCache.java:2411)
        at com.github.benmanes.caffeine.cache.BoundedLocalCache.computeIfAbsent(BoundedLocalCache.java:2394)
        at com.github.benmanes.caffeine.cache.LocalAsyncCache.get(LocalAsyncCache.java:91)
        at com.github.benmanes.caffeine.cache.LocalAsyncCache.get(LocalAsyncCache.java:82)
        at akka.http.caching.LfuCache.getOrLoad(LfuCache.scala:126)
        at com.scalableminds.webknossos.datastore.datareaders.DatasetArray.getSourceChunkDataWithCache(DatasetArray.scala:96)
        at com.scalableminds.webknossos.datastore.datareaders.DatasetArray.$anonfun$readAsFortranOrder$4(DatasetArray.scala:77)
        at com.scalableminds.util.tools.Fox$.runNext$3(Fox.scala:131)
        at com.scalableminds.util.tools.Fox$.serialCombined(Fox.scala:137)
        at com.scalableminds.webknossos.datastore.datareaders.DatasetArray.readAsFortranOrder(DatasetArray.scala:75)
        at com.scalableminds.webknossos.datastore.datareaders.DatasetArray.readBytes(DatasetArray.scala:54)
        at com.scalableminds.webknossos.datastore.datareaders.DatasetArray.readBytesXYZ(DatasetArray.scala:46)
        at com.scalableminds.webknossos.datastore.dataformats.zarr.ZarrCubeHandle.cutOutBucket(ZarrBucketProvider.scala:23)
        at com.scalableminds.webknossos.datastore.dataformats.BucketProvider.$anonfun$load$2(BucketProvider.scala:23)
        at com.scalableminds.webknossos.datastore.storage.DataCubeCache.$anonfun$withCache$5(DataCubeCache.scala:76)
        at com.scalableminds.util.tools.Fox.$anonfun$flatMap$1(Fox.scala:259)
        at scala.concurrent.Future.$anonfun$flatMap$1(Future.scala:307)
        at java.util.concurrent.ForkJoinTask$RunnableExecuteAction.exec(ForkJoinTask.java:1402)
        at java.util.concurrent.ForkJoinTask.doExec(ForkJoinTask.java:289)
        at java.util.concurrent.ForkJoinPool$WorkQueue.runTask(ForkJoinPool.java:1056)
        at java.util.concurrent.ForkJoinPool.runWorker(ForkJoinPool.java:1692)
        at java.util.concurrent.ForkJoinWorkerThread.run(ForkJoinWorkerThread.java:175)
```

Is there something wrong with my setup? I did a `./clean` after checking out this branch..
 
> * The label of the dropzone says „(Optional)“, which seems a bit redundant since above there is already the radio selection between anonymous and with credential. Maybe this could be removed (or remove the radio selection and make the auth fields optional for all cases?)

Done :+1: 

> * If I press reset after a dataset has been recovered, the status which protocol was selected seems to be lost. The gs uri is still in the field, but the auth fields are showing the basicauth case. Not super important, since the reset button will probably be used rarely shrug

Good point. I adapted the reset button to also reset the original input field.

fm3:
Thanks!

The error reads to me as if blosc is not installed. That is a data (de)compression library. Does loading other blosc-compressed zarr datasets work for you? Could you try `apt install libblosc1`? Compare https://github.com/scalableminds/webknossos/blob/master/DEV_INSTALL.md

frontend/javascripts/admin/dataset/dataset_add_neuroglancer_view.tsx
```diff
@@ -26,9 +27,68 @@ type StateProps = {
...
+  }
+  const jsonString = await readFileAsText(file);
+  return JSON.parse(jsonString);
+};
+
+export function GoogleAuthFormItem({
+  fileList,
+  handleChange,
+  showOptionalHint,
+}: {
```
philippotto, line 45: This part (and following) was merely moved.

CHANGELOG.unreleased.md
```diff
@@ -11,6 +11,7 @@ For upgrade instructions, please check the [migration guide](MIGRATIONS.released
 [Commits](https://github.com/scalableminds/webknossos/compare/23.02.0...HEAD)
 
 ### Added
+- Remote datasets can now also be streamed from Google Cloud Storage uris (`gs://`). [#6775](https://github.com/scalableminds/webknossos/pull/6775)
```
daniel-wer, line 14:
```suggestion
- Remote datasets can now also be streamed from Google Cloud Storage URIs (`gs://`). [#6775](https://github.com/scalableminds/webknossos/pull/6775)
```

At least to me this looks less strange :)

frontend/javascripts/admin/dataset/dataset_add_neuroglancer_view.tsx
```diff
@@ -26,9 +27,68 @@ type StateProps = {
...
-}>;
+
+export type FileList = UploadFile<any>[];
+
+export const parseCredentials = async (file: RcFile | undefined): Promise<Object | null> => {
+  if (!file) {
+    return null;
+  }
+  const jsonString = await readFileAsText(file);
+  return JSON.parse(jsonString);
```
daniel-wer, line 38: Should this be guarded somehow? The page shouldn't crash if a file with the wrong format was uploaded.

frontend/javascripts/admin/dataset/dataset_add_zarr_view.tsx
```diff
@@ -327,13 +343,28 @@ function AddZarrLayer({
...
-        : await exploreRemoteDataset([datasourceUrl], {
+    const { dataSource: newDataSource, report } = await (async () => {
+      if (showCredentialsFields) {
+        if (selectedProtocol === "gs") {
+          const credentials =
+            fileList.length > 0 ? await parseCredentials(fileList[0]?.originFileObj) : null;
+          if (credentials) {
+            return exploreRemoteDataset([datasourceUrl], {
+              username: "",
+              pass: JSON.stringify(credentials),
```
daniel-wer, line 354: I tried to find out which type `credentials` has, but it's not strictly specified in wk. The only thing I found was `Record<string, any>` in the `NeuroglancerDatasetConfig`. Is there a more specific definition of the credential type? And is stringifying and passing all of it as the password here, correct?

daniel-wer: Ah I found the link to the documentation which shows what the credentials file looks like (https://cloud.google.com/iam/docs/creating-managing-service-account-keys?hl=de#creating). In that case, there doesn't need to be a more specific type and it also makes sense to pass all of it as `pass` here, so nevermind :)

daniel-wer (commented):
Frontend code almost LGTM. I'll try to test and will report back.

daniel-wer:
Works very nicely :+1: 

Notes from testing:

- [ ] The "Add Remote Zarr / N5 Dataset" page should list which storage types are supported: https, s3, gcs afaik. There is a validation error if a non-supported url is pasted, but it would be nice to find out, before.
  - [ ] There is a typo on the page: "segmentattion" with double t

Not from this PR, but I really like the error log, if the import doesn't work! :1st_place_medal:

frontend/javascripts/admin/dataset/dataset_add_neuroglancer_view.tsx
daniel-wer: I tested this case. Currently, the page doesn't crash, but also nothing happens and no error is shown.

philippotto:
> [x] The "Add Remote Zarr / N5 Dataset" page should list which storage types are supported: https, s3, gcs afaik. There is a validation error if a non-supported url is pasted, but it would be nice to find out, before.
> * [x]  There is a typo on the page: "segmentattion" with double t

Done :)

frcroth (approved):

Issue #6509, opened 2022-09-28
Implement Pricing Concept


philippotto:
- [x] Organization Page should be given some love. The page should also show what the possible pricing plans are and how to upgrade. @hotzenklotz 
- [x] Enforce user limit: Disable accepting invites when full
- [x] New organization fields
   - [x] paidUntil
   - [x] included users, storage
   - [x] Plans: Free, Team, Power, Team-Trial, Power-Trial, Custom
----------------------
- [x] If a feature is not supported by a pricing plan, the UI should forbid the corresponding action and tell the user how to upgrade.
   - [x] Admin pages: tasks, projects, teams, tasks tab in dashboard #6767
   - [x] Sharing modal: team sharing  #6776
   - [x] Dataset settings: allowed teams  #6776
   - [x] Pre-computed meshes: context menu, sidebar #6776
   - [x] Segmentation proof-reading button #6776
   - [x] New dataset dashboard: folder view on the left, metadata, tags, permissions #6767
- [ ] Enforce storage limit
   - [x] Maintain current storage usage
   - [ ] Disable uploading when full (also in libs) #6808 
- [ ] #6855

Also see:
- https://www.notion.so/scalableminds/webKnossos-Monetization-236159a5b7c8487a8d09945aba9f00dc
- https://docs.google.com/spreadsheets/d/1eD3UeAZd6LhlhazUlUcLLGQMptrIgOivxtq4N6lRQIg/edit#gid=0

normanrz:
Mockup for organization page:

![organization mockup](https://user-images.githubusercontent.com/335438/195882581-e56bc8eb-5ef1-47b2-8ab8-ea027dfa26d0.png)

Albane-dlv:
Brainstorm wording and design for unsupported features in notion page: https://www.notion.so/scalableminds/Advertise-pricing-3386806f8cc84e689bf98e2c60526f10

PR #6776, opened 2023-01-23
Enforce Pricing for Team Sharing

hotzenklotz:
PR adds a new wrapper component blur components/features that are not part of an organization's selected plan.

Changes:
- new blurring wrapper component
- affects Team sharing modal and dataset setting's sharing tab
- moved `PricingPlanEnum` to `pricing_plan_utils` to unroll a cyclic dependency 

- https://moreplans.webknossos.xyz/

### Steps to test:
- Set orga plan accordingly in your DB
- Check annotation team sharing modal
- Check dataset sharing tab

### Issues:
- contributes to #6509 

------
- [x] Ready for review


frontend/javascripts/components/pricing_enforcers.tsx
```diff
@@ -75,14 +64,42 @@ export function PricingEnforcedButton({
...
+
+export const PricingEnforcedBlur: React.FunctionComponent<RequiredPricingProps> = ({
+  children,
+  requiredPricingPlan,
+}) => {
+  const currentPricingPlan = useSelector((state: OxalisState) =>
+    state.activeOrganization ? state.activeOrganization.pricingPlan : PricingPlanEnum.Basic,
+  );
+  const isFeatureAllowed = isPricingPlanGreaterEqualThan(currentPricingPlan, requiredPricingPlan);
+  if (isFeatureAllowed) return <>children</>;
```
philippotto, line 84: I think, this would render the string "children" :thinking: 

```suggestion
  if (isFeatureAllowed) return <>{children}</>;
```
or:
```
  if (isFeatureAllowed) return children;
```

philippotto (commented):
Code looks good except for maybe one thing :+1: Didn't test yet, though.

frontend/javascripts/components/pricing_enforcers.tsx
hotzenklotz: Ah, good catch. Unfortunately, `return children` does not work... well at least Typescript complains... (React actually was able to work with this nonetheless). Hence, I had to wrap it in a Fragment to satisfy the type checker.

philippotto (approved):
Also works well :+1: 

However, it took me a bit until I could fully "parse" the blurred component. Especially when the blurred area is relatively large, the lock icon is very easy to miss. For example:

image

image

I'm not sure how this could be improved. Initially, I had the idea to simply always show the tooltip content next to the lock icon in the center of the blurred area (including the black tooltip background). However, I don't know whether it will look good and will work with the wide teams input that is not very tall. Maybe another color for the lock icon?

Maybe you want to experiment a bit. If it stays as is, it's probably okay, too, since the tooltip will show why things are blurred, too.

By the way, during testing, my Spotify "Discover Weekly" playlist decided to play a song called "Blurry Mess". I'm feeling watched :sweat_smile:

frontend/javascripts/oxalis/view/action-bar/tracing_actions_view.tsx
```diff
@@ -123,6 +123,8 @@ export function LayoutMenu(props: LayoutMenuProps) {
     autoSaveLayouts,
     setAutoSaveLayouts,
     saveCurrentLayout,
+    // rome-ignore lint/correctness/noUnusedVariables: underscore prefix does not work with object destructuring
+    setCurrentLayout,
```
hotzenklotz, line 127: fix for accidentally introduced bug in my recent Linter PR

hotzenklotz:
@philippotto Please re-review with latest changes:
- I added helper function like we discussed, see `isFeatureAllowedByPricingPlan`
- Added checks for `Pre-Computed Mesh` creation button
- Added checks for Proof-Reading tool

When testing, please try anonymous access / publicly shared annotation as well. Please also consider how else some of these features can be activated, e.g. through keyboard shortcuts.

frontend/javascripts/oxalis/view/action-bar/tracing_actions_view.tsx
hotzenklotz: React/Antd complaining about the `setCurrentLayout` prop being present in/leaking into a `<Menu>` component, where it was not expected/supported.

philippotto:
Testing went well :+1:

One thing I noticed is that if a feature is disabled due to the incorrect orga plan, the user won't know which plan is necessary. For example, for the proofreading tool, one will have to look up the plan comparison to see that the power plan is required. One could mention this in the tooltip. However, it would just be nice to have, I think.

PR #6782, opened 2023-01-25
Catch zero-sized buckets in backend, fix http nio timeout handling

### LINKED ISSUE #6709: Sometimes zarr data requests return too few buckets
## Steps to Reproduce the bug

1. Set up this dataset here https://scm.slack.com/archives/C3RUHRXBM/p1671526934198699?thread_ts=1671526852.728559&cid=C3RUHRXBM
2. View dataset and move in mag 2
3. Sometimes a warning is shown: `bucket.ts?f9e9:578 bucket.data has unexpected length {arrayBuffer: Uint8Array(0), actual: 0, expected: 32768, channelCount: 1}` which can result in corrupt rendering

Some details for the occurrence I inspected:
- one request for the uint8 color layer requested 6 buckets, but got only 4.
  - requested:
```
[4253, 5940, 123, 1]
[4254, 5945, 123, 1]
[4261, 5944, 123, 1]
[4252, 5940, 123, 1]
[4253, 5945, 123, 1]
[4252, 5944, 123, 1]
```
- the missingbuckets header was set to `[]` though
- the received buffer was 131072 entries long (131072 / 32**3 == 4)
- for some other volume request, the missingbuckets header contained buckets. so, the header itself seems to work at least sometimes

## Your Environment for bug
Localhost

fm3:
Turns out the our https filesystem provider did not correctly handle read timeouts. This produced seemingly successful bucket loading, but the arrays had zero length. This PR
 - fixes the timeout handling of those http requests
 - adds a guard against this kind of error in BinaryDataService – all zero-length arrays are considered Failures. This matches the frontend expectations.

### Steps to test:
- In HttpsSeekableByteChannel.scala, change the timeouts to be extremely short, like 1ms
- Requesting remote zarr data should not result in rendering errors, but rather correctly behave like that the buckets could not be loaded

### Issues:
- fixes #6709 

------
- [x] Ready for review

### DIFF
--- a/CHANGELOG.unreleased.md
+++ b/CHANGELOG.unreleased.md
@@ -37,6 +37,7 @@ For upgrade instructions, please check the [migration guide](MIGRATIONS.released
 - Fixed deprecation warnings for antd <Modal> props. [#6765](https://github.com/scalableminds/webknossos/pull/6765)
 - Fixed a bug where direct task assignment to a single user would fail. [#6777](https://github.com/scalableminds/webknossos/pull/6777)
 - Fixed a bug where the dataset folders view would not list public datasets if the requesting user could not also access the dataset for other reasons, like being admin. [#6759](https://github.com/scalableminds/webknossos/pull/6759)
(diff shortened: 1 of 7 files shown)


jstriebel (approved):
LGTM :+1:

PR #6792, opened 2023-01-31
adds "Ok" to health check

normanrz:
This makes it easier to see whether the health check worked in the browser or via curl.

------
- [x] Ready for review


fm3 (approved):
Ok

PR #6794, opened 2023-01-31
Fix datastore file system watcher

fm3:
Fixes regression introduced in #6788 – apparently the execution order caused this override value to be unset, leading the ticker to be disabled. Changing this from val to def fixes it. I like this better than the `lazy val` this was before, as the superclass also has this as def.

I also decided to lower the default delay from 10s to 5s for this to shorten the time the datasources are unavailable.

### Steps to test:
- run wk with default config, should log `Scanning inbox (binaryData)...` shortly after startup

------
- [x] Ready for review


philippotto (approved):
LGTM! Great, that you found this :+1:

PR #6795, opened 2023-01-31
Fix proofreading when mag 1 doesn't exist for segmentation layer

philippotto:
- https://___.webknossos.xyz

### Steps to test:
- I duplicated the test-agglomerate-ds, added another color layer (which has mag 1), downsampled the seg layer and removed mag 1 so that only mag 2 exists (therefore, color exists in mag 1 but the segmentation layer only in mag 2)
- then I used the proofreading tool to merge segments (didn't work before in this constellation)
- I also doublechecked that one cannot merge with segment 0

### Issues:
- see https://scm.slack.com/archives/C02H5T8Q08P/p1675179308527579?thread_ts=1675151659.552189&cid=C02H5T8Q08P

------
- [x] Ready for review


daniel-wer (approved):
The fix makes perfect sense to me :+1:

PR #6799, opened 2023-02-01
Fix editable text style

hotzenklotz:
This PR mainly fixes some small line breaks in the some places in the UI that I found annoying.

Changes:
- Annotation Info Tab
  - Ensure annotation name and pencil icon are on the same line
  - Ensure annotation description and pencil icon are on the same line
- Segments Tab
  - Ensure each list entry is on one line
- Refactored some typescript typing and removed suppressions/increased coverage

Before: 
image

After:
image

Before:
image

After:
image

Layout/CSS tested in Safari 16.03, Chrome 109.0.5414.119, Firefox 109.0.1 on MacOS.

- https://___.webknossos.xyz

### Steps to test:
- Check that the line breaks work in your browser
- Double-check where else the changed components might be use, .e.g. dataset list on dashboard

### Issues:
- None

------
- [x] Ready for review


frontend/javascripts/oxalis/view/components/input_component.tsx
```diff
@@ -96,19 +119,31 @@ class InputComponent extends React.PureComponent<InputComponentProps, InputCompo
 
   render() {
     const { isTextArea, onPressEnter, title, style, ...inputProps } = this.props;
-    const InputComponentType: typeof TextArea | typeof Input = isTextArea ? TextArea : Input;
-    const input = (
-      <InputComponentType
+    // const InputComponentType: typeof TextArea | typeof Input = isTextArea ? TextArea : Input;
```
philippotto, line 122: remove?

frontend/javascripts/oxalis/view/components/input_component.tsx
```diff
@@ -96,19 +119,31 @@ class InputComponent extends React.PureComponent<InputComponentProps, InputCompo
...
+    const input = isTextArea ? (
+      <TextArea
+        {...inputProps}
+        style={title == null ? style : undefined}
+        onChange={this.handleChange}
+        onFocus={this.handleFocus}
+        onBlur={this.handleBlur}
+        value={this.state.currentValue}
+        onPressEnter={onPressEnter}
```
philippotto, line 131: why no `blurYourself` anymore as a fallback?

frontend/javascripts/oxalis/view/components/input_component.tsx
```diff
@@ -96,19 +119,31 @@ class InputComponent extends React.PureComponent<InputComponentProps, InputCompo
...
+    ) : (
+      <Input
         {...inputProps}
         style={title == null ? style : undefined}
         onChange={this.handleChange}
         onFocus={this.handleFocus}
         onBlur={this.handleBlur}
         value={this.state.currentValue}
-        onPressEnter={onPressEnter != null ? onPressEnter : this.blurYourself}
+        onPressEnter={onPressEnter}
```
philippotto, line 142: This could probably be put into an object and then spread into here and into `TextArea` (`{ style: title == .., onChange: ... }`) to avoid the redundancy.

philippotto (changes requested):
Thank you for doing this! Only got some smaller things. Will test after these are resolved.

frontend/javascripts/oxalis/view/components/input_component.tsx
hotzenklotz: Hehehe, good catch. I removed it so the review can find something ;-) Will restore it.

frontend/javascripts/oxalis/view/components/input_component.tsx
hotzenklotz: Nice idea, but it seems like this is not possible. When specifying all the props individually like in this PR then TS can do its discriminative typing thing and resolve the correct type information for all the callbacks, either `EventHandler<Input>` or `EventHandler<TextArea>`. 
When shortening the props into a single object, each callback will have a mixed `EventHandler<Input|TextArea>` type and TS will complain.

philippotto: Ok, fair enough, then let's keep it as is :+1:

philippotto:
Works great :+1: However, I found one problem (not sure whether it's new but definitely related): When editing the description of an annotation, the "enter" key does not produce line breaks. If this is fixed, the PR should be good to go :)

hotzenklotz:
@philippotto I think I fixed the behavior of the ENTER key for the multiline text area. Please see my commit: https://github.com/scalableminds/webknossos/pull/6799/commits/973339a1c0ef8e45618440ad5999e7fce76d1646

philippotto (approved):

PR #6803, opened 2023-02-01
Rewrite database tooling and support for postgres passwords

### LINKED ISSUE #6727: Use postgres with password
webKnossos currently requires the postgres database to be configured with default username and password `postgres:postgres`. For security reasons, it would be desirable to allow arbitrary username/password combinations.
I believe JDBC and Slick already support that with the connection URL syntax `jdbs:postgres://username:password@host/database`. However, all adjacent scripts (e.g. `ensure_db.sh`) don't.

normanrz:
This PR rewrites the previous database tools from `tools/postgres` into a single JS file with multiple commands. The `dbtool` (I'll happily take other name suggestions) has some handling of parsing the Scala-supplied postgres connection config into regular postgres URLs (eg. `postgres://user:password@host:port/database`) for use with `psql`.

The Scala code now calls this tool by invoking node as a command instead of JVM-executed JS. Therefore, I added node to the Docker image. I could remove a few Scala and node dependencies.

`diff-schema` was a JS script already, but I also rewrote to have everything in one place. I tested that it produces exactly the same dump output.

To test everything, I configured the CI to use non-standard user/password combination and set up a different postgres user on the dev instance.

- https://postgrespassword.webknossos.xyz

### Steps to test:
- Create a new postgres superuser: `CREATE ROLE webknossos WITH LOGIN SUPERUSER PASSWORD '...';`
- Add user and password to `slick.conf` (or as env var)
- Try to run webKnossos
- Try the commands of the `tools/postgres/dbtool.js`; although, most are checked during runtime or CI

### Issues:
- fixes #6727 

------
- [x] Ready for review


normanrz:
ftr https://stackoverflow.com/posts/70103077/revisions to change the owner of postgres objects.

docker-compose.yml
```diff
@@ -35,7 +35,9 @@ services:
     volumes:
       - ./binaryData:/webknossos/binaryData
     environment:
-      - POSTGRES_URL=jdbc:postgresql://postgres/webknossos
+      POSTGRES_URL: jdbc:postgresql://postgres/webknossos
+      PGUSER: webknossos
```
philippotto, line 39: what do you think about defaulting to `webknossos_user`? I find this easier to understand when seeing it as a string somewhere. Feel free to ignore if you think it would be too unusual.

tools/postgres/dbtool.js
```diff
@@ -0,0 +1,447 @@
...
+  urlWithoutDatabase.pathname = "";
+
+  return {
+    url: url.toString(),
+    urlWithoutDatabase: urlWithoutDatabase.toString(),
+    username: url.username,
+    password: url.password,
+    hostname: url.hostname,
+    port: url.port,
+    database: url.pathname.length > 1 ? url.pathname.substring(1) : "webknossos",
```
fm3, line 40: could you add a comment explaining the substring(1) here? What character should be dropped?

package.json
```diff
@@ -112,13 +108,13 @@
...
     "docs": "node_modules/.bin/documentation build --shallow frontend/javascripts/oxalis/api/api_loader.ts frontend/javascripts/oxalis/api/api_latest.ts --github --project-name \"WEBKNOSSOS Frontend API\" --format html --output public/docs/frontend-api",
-    "refresh-schema": "./tools/postgres/refresh_schema.sh && rm -f target/scala-2.12/src_managed/schema/com/scalableminds/webknossos/schema/Tables.scala",
-    "enable-jobs": "sed -i -e 's/jobsEnabled = false/jobsEnabled = true/g' ./conf/application.conf; ./tools/postgres/set_jobs.sh true",
-    "disable-jobs": "sed -i -e 's/jobsEnabled = true/jobsEnabled = false/g' ./conf/application.conf; ./tools/postgres/set_jobs.sh false",
+    "refresh-schema": "./tools/postgres/dbtool.js refresh-schema && rm -f target/scala-2.12/src_managed/schema/com/scalableminds/webknossos/schema/Tables.scala",
+    "enable-jobs": "sed -i -e 's/jobsEnabled = false/jobsEnabled = true/g' ./conf/application.conf; ./tools/postgres/dbtool.js enable-jobs",
+    "disable-jobs": "sed -i -e 's/jobsEnabled = true/jobsEnabled = false/g' ./conf/application.conf; ./tools/postgres/dbtool.js enable-jobs",
```
philippotto, line 113:
```suggestion
    "disable-jobs": "sed -i -e 's/jobsEnabled = true/jobsEnabled = false/g' ./conf/application.conf; ./tools/postgres/dbtool.js disable-jobs",
```

tools/postgres/dbtool.js
```diff
@@ -0,0 +1,447 @@
...
+    console.log(
+      safePsqlSpawn([
+        PG_CONFIG.urlWithoutDatabase,
+        "-tAc",
+        `CREATE DATABASE ${PG_CONFIG.database};`,
+      ]),
+    );
+  }
+
+  const doesSchemaExist = callPsql(
```
fm3, line 94:
```suggestion
  const existingSchemaName = callPsql(
```

tools/postgres/dbtool.js
```diff
@@ -0,0 +1,447 @@
...
+  console.log(safePsqlSpawn([PG_CONFIG.url, "-v", "ON_ERROR_STOP=ON", "-f", schemaPath]));
+}
+
+function dumpSchema(databaseUrl, schemaDir, silent = false) {
+  if (!fs.existsSync(schemaDir) || !fs.statSync(schemaDir).isDirectory) {
+    console.error("Schema directory $schemadir does not exist, aborting!");
+    process.exit(1);
+  }
+
+  for (const fn of fs.readdirSync(schemaDir)) {
```
fm3, line 140:
```suggestion
  for (const filename of fs.readdirSync(schemaDir)) {
```
(i read it as function first and was confused. I think it’s filename?)

tools/postgres/dbtool.js
```diff
@@ -0,0 +1,447 @@
...
+      fs.rmSync(evolutionsDumpDir, { recursive: true, force: true });
+    }
+    if (fs.existsSync(schemaDumpDir)) {
+      fs.rmSync(schemaDumpDir, { recursive: true, force: true });
+    }
+  }
+  console.log("✨✨ Evolutions match with schema file");
+}
+
+function applyEvolutions() {
```
fm3, line 268: I don’t think this is used anywhere, do you think we should keep it anyway?

fm3 (commented):
Cool stuff!

tools/postgres/dbtool.js
```diff
@@ -0,0 +1,447 @@
...
+      ? url.username
+      : process.env.PGUSER != null
+      ? process.env.PGUSER
+      : "postgres";
+  url.password =
+    url.password !== "" && url.password != null
+      ? url.password
+      : process.env.PGPASSWORD != null
+      ? process.env.PGPASSWORD
+      : "postgres";
```
philippotto: maybe it makes sense to also check against an empty string for user and password? also, just checking for falsy (e.g., `url.password ? ... : ...`) would work for that, too, in case the loss of explicitness is fine for you.

tools/postgres/dbtool.js
```diff
@@ -0,0 +1,447 @@
...
+        .replace(/,$/gm, "")
+        .replace(/\\r/gm, "  ")
+        .split("\n")
+        .sort()
+        .join("\n"),
+    );
+  }
+}
+
+function dumpSchemaViaSqlFiles(filePaths) {
```
philippotto, line 179: `dumpExpectedSchema` would be clearer to me. The parameter could be named `sqlFilePaths` to convey the sql bit. Similarily, `dumpSchema` could be renamed to `dumpCurrentSchema`.

tools/postgres/dbtool.js
```diff
@@ -0,0 +1,447 @@
...
+    console.log("✨✨ Successfully applied the evolutions");
+  } else {
+    console.log("There are no evolutions that can be applied.");
+  }
+}
+
+function assertUniqueEvolutionNumbers() {
+  const groupedEvolutions = new Map();
+  for (const filename of fs.readdirSync(evolutionsPath)) {
+    const num = parseInt(filename.split("-")[0], 10);
```
philippotto, line 308: maybe check for NaN to be extra careful.

tools/postgres/dbtool.js
```diff
@@ -0,0 +1,447 @@
...
+program
+  .command("drop-db")
+  .description("Drop the current database")
+  .action(() => {
+    dropDb();
+    console.log("✨✨ Done");
+  });
+
+program
+  .command("refresh-db")
```
philippotto, line 340: Maybe call it `drop-and-refresh-db`? I always found the existing naming a bit dangerous.

philippotto (approved):
Looks great :+1: Only left a couple of nitpicks and suggestions.

tools/postgres/dbtool.js
normanrz: let's keep it

fm3 (approved):
✨

PR #6810, opened 2023-02-03
Fix some login problems

philippotto:
This PR fixes two problems:
- after logging in, an error boundary UI briefly showed up, because the active organization wasn't set yet (but was enforced by the dashboard)
- due to the above bug I noticed that there was apparently a hard refresh after logging in which I fixed now since it was unnecessary.

- https://___.webknossos.xyz

### Steps to test:
- log in from `/auth/login` and from the navbar (e.g., open a public dataset for that)
- logging in should not show up an error message

### Issues:
- partly related to #6767

------
- [X] Ready for review

### DIFF
--- a/CHANGELOG.unreleased.md
+++ b/CHANGELOG.unreleased.md
@@ -13,9 +13,10 @@ For upgrade instructions, please check the [migration guide](MIGRATIONS.released
 ### Added
 
 ### Changed
-- Limit paid team sharing features to respective organization plans. [6767](https://github.com/scalableminds/webknossos/pull/6776)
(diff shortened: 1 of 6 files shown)


hotzenklotz (approved):
Tested and works well. Thanks.

What component caused the full page refresh?

philippotto:
> What component caused the full page refresh?

See here: https://github.com/scalableminds/webknossos/blob/master/frontend/javascripts/router.tsx#L273
I think it's necessary so that the user can be redirected to `webknossos.org` after logging out (which doesn't live inside this repo).

PR #6811, opened 2023-02-03
Add owner information to organization

hotzenklotz:
- Added owner information to organization REST API route and organization page. 
  - Flo and I decided that we did not want share the whole orga owner user object through the API but rather just a string since some API calls expose the default orga without any authentication
- Also added the owner name to some of the pricing tooltips to help guide people to right person. 
- Added a button to the tooltip/popover in case you are the owner to upgrade seamlessly.

### Steps to test:
- Hover over any feature that is restricted to certain plans
- Check out the owner name in the organization page

### Issues:
- contributes to #6509 

------
- [x] Ready for review


hotzenklotz:
@fm3 Please review backend parts
@philippotto Please review frontend parts.

frontend/javascripts/admin/organization/upgrade_plan_modal.tsx
```diff
@@ -299,6 +299,7 @@ function UpgradePricingPlanModal({
           ) : null}
         </>
       }
+      zIndex={10000} // overlay everything
```
hotzenklotz, line 302: @philippotto I tried to click somewhere in the background (`document.body.click()`) to make the menu disappear but that did not work. With the high zIndex you can at least be sure that it overlays anything.

frontend/javascripts/messages.tsx
```diff
@@ -436,6 +436,9 @@ instead. Only enable this option if you understand its effect. All layers will n
   "ui.no_form_active": "Could not set the initial form values as the form could not be loaded.",
   "organization.plan.upgrage_request_sent":
     "An email with your upgrade request has been sent to the WEBKNOSSOS sales team.",
-  "organization.plan.feature_not_available": (requiredPlan: string) =>
-    `This feature is not available in your organization's plan. Ask your organization owner to upgrade at least a ${requiredPlan} plan.`,
+  "organization.plan.feature_not_available": (
+    requiredPlan: string,
+    organizationOwnerName: string,
+  ) =>
+    `This feature is not available in your organization's plan. Ask your organization owner ${organizationOwnerName} to upgrade to at least a ${requiredPlan} plan.`,
```
philippotto, line 443: "Ask the owner of your organization (John Doe)" would be more idiomatic english than "organization owner", I think.
Also, I'd change the type to `organizationOwnerName: string | null` and then do a conditional on that so that the double space is avoided if the owner is not known.

frontend/javascripts/admin/organization/pricing_plan_utils.ts
```diff
@@ -83,3 +84,23 @@ export function isFeatureAllowedByPricingPlan(
...
   return isPricingPlanGreaterEqualThan(organization.pricingPlan, requiredPricingPlan);
 }
+
+export function getFeatureNotAvailabeInPlanMessage(
+  requiredPricingPlan: PricingPlanEnum,
+  organization: APIOrganization | null,
+) {
+  let organizationOwnerName = "";
+
+  // expected naming schema for owner: "(M. Mustermann)" | ""
```
philippotto, line 94: what is the reasoning for abbreviating the first name? privacy? conciseness?

frontend/javascripts/components/pricing_enforcers.tsx
```diff
@@ -24,18 +32,41 @@ const handleMenuClick: MenuClickEventHandler = (info) => {
...
   if (isFeatureAllowed) return <Menu.Item {...menuItemProps}>{children}</Menu.Item>;
 
   return (
-    <Tooltip
-      title={messages["organization.plan.feature_not_available"](requiredPricingPlan)}
+    <Popover
+      color={PRIMARY_COLOR_HEX}
+      content={
+        <div style={popOverStyle}>
+          {getFeatureNotAvailabeInPlanMessage(requiredPricingPlan, activeOrganization)}
```
philippotto, line 64: typo in `Availabe`

philippotto (changes requested):
:+1: Only left some nitpicks/suggestions. Will test later.

app/models/user/User.scala
```diff
@@ -159,6 +159,19 @@ class UserDAO @Inject()(sqlClient: SqlClient)(implicit ec: ExecutionContext)
...
+      accessQuery <- accessQueryFromAccessQ(listAccessQ)
+      r <- run(q"""select $columns
+                   from $existingCollectionName
+                   where $accessQuery
+                   and isOrganizationOwner
+                   and not isDeactivated
+                   and _organization = $organizationId
+                   order by _id""".as[UsersRow])
+      parsed <- parseFirst(r, organizationId)
+    } yield parsed
```
fm3, line 173:
```suggestion
  def findOwnerByOrg(organizationId: ObjectId): Fox[User] =
    for {
      r <- run(q"""select $columns
                   from $existingCollectionName
                   where isOrganizationOwner
                   and not isDeactivated
                   and _organization = $organizationId
                   order by _id
                   limit 1""".as[UsersRow])
      parsed <- parseFirst(r, organizationId)
    } yield parsed
```
The access query is not needed because everyone who may see the organization json may also fetch the owner. This also makes the DBAccessContext parameter unnecessary.

app/models/organization/OrganizationService.scala
```diff
@@ -38,6 +40,8 @@ class OrganizationService @Inject()(organizationDAO: OrganizationDAO,
     } else Json.obj()
     for {
       usedStorageBytes <- organizationDAO.getUsedStorage(organization._id)
+      owner <- userDAO.findOwnerByOrg(organization._id)(GlobalAccessContext)
+      ownerName = s"${owner.firstName} ${owner.lastName}"
```
fm3, line 44:
```suggestion
      ownerBox <- userDAO.findOwnerByOrg(organization._id).futureBox
      ownerNameOpt = ownerBox.toOption.map(o => s"${o.firstName} ${o.lastName}")
```

There is not currently any code guaranteeing that there exists an owner. While this is kind of an invalid state, I’d add this safeguard here to avoid breaking the entire organization request in this case. I think philipp also already talked about making the ownerName property optional in the frontend.

app/models/organization/OrganizationService.scala
```diff
@@ -49,7 +53,8 @@ class OrganizationService @Inject()(organizationDAO: OrganizationDAO,
         "paidUntil" -> organization.paidUntil,
         "includedUsers" -> organization.includedUsers,
         "includedStorageBytes" -> organization.includedStorageBytes,
-        "usedStorageBytes" -> usedStorageBytes
+        "usedStorageBytes" -> usedStorageBytes,
+        "ownerName" -> ownerName
```
fm3, line 57:
```suggestion
        "ownerName" -> ownerNameOpt
```

fm3 (changes requested):
Thanks for looking into this! I added some suggestions for the scala code :)

frontend/javascripts/admin/organization/pricing_plan_utils.ts
hotzenklotz: Conciseness. I wanted to keep the tooltip as short as possible.

frontend/javascripts/messages.tsx
hotzenklotz: > Also, I'd change the type to organizationOwnerName: string | null and then do a conditional on that so that the double space is avoided if the owner is not known.

Mhm, I was hoping that `getFeatureNotAvailabeInPlanMessage()` already checks against null to avoid doing this logic in the `messages` module. it, does not, however, prevent a double space.

frontend/javascripts/components/pricing_enforcers.tsx
hotzenklotz: Done

frontend/javascripts/messages.tsx
philippotto: ok, i see. i don't even know whether the double space is rendered at all? I think this could get collapsed into a single space by the browser? if it's rendered, fixing the doublespace without moving the null-check could work something like this: `... owner ${organizationOwnerName}to` (no space before "to") and then `organizationOwnerName` should have a trailing space if it's not empty...

hotzenklotz: TBH, I would not optimize for extremely unlikely case that an orga has no owner. This can only ever be the case if a DB migration was not executed correctly.

philippotto: good point :+1:

fm3 (approved):
Backend LGTM. Did not test, though

philippotto:
I noticed two things during testing:

- [ ] the tooltip says "Ask the owner (...)" even when the current user is the owner actually. I think this is weird and we should try to avoid this wording in that case.
- [ ] in case of the proofreading tool, there is no popover but still the old tooltip. is this intended?

Other than that, it works great :+1:

hotzenklotz:
>  in case of the proofreading tool, there is no popover but still the old tooltip. is this intended?

Yes, this is intended. The toolbar has uses a tooltip for the description of all the individual tools that was "hijacked" for this message. This keeps things simple.

philippotto:
> > in case of the proofreading tool, there is no popover but still the old tooltip. is this intended?
> 
> Yes, this is intended. The toolbar has uses a tooltip for the description of all the individual tools that was "hijacked" for this message. This keeps things simple.

Ok, fair enough :+1:

frontend/javascripts/oxalis/model/accessors/tool_accessor.ts
```diff
@@ -236,6 +242,7 @@ export function getDisabledInfoForTools(state: OxalisState): Record<
     hasAgglomerateMappings,
     genericDisabledExplanation,
     state.activeOrganization,
+    enforceActiveUser(state.activeUser),
```
philippotto, line 245: why enforce? `getFeatureNotAvailableInPlanMessage` should be able to handle null for the active user. i think this could crash currently when viewing a public annotation.

philippotto (approved):
Great :+1:

Issue #5792, opened 2021-10-21
Add `Show in list` item to context menu for node, segments and bounding boxes.


MichaelBuessemeyer:
## Detailed Description

Currently, the user does not directly know which node/skeleton, segment, or bounding box is hovered at the moment. Thus, a menu item in the context menu that triggers a jump to the segment list to the hovered segment would be very nice. The same functionality could be used for node/skeletons and bounding boxes.

A good starting point for this issue would be the Docs / Code of the layouting lib `FlexLayout`. This lib should make it possible to open a specific tab upon a certain user action (e.g. a context menu item click).

## Context

* [ ] Specific to long-running jobs (set `jobsEnabled=true` in `application.conf`)
* [ ] Specific to webKnossos.org (set `isDemoInstance=true` in `application.conf`)

MichaelBuessemeyer:
Update:
- [ ] Only skeletons / nodes do not support this

Segments and bounding boxes already have a "Focus in <view>" menu item

PR #6813, opened 2023-02-03
Add mesh actions to 3d viewport context menu

### LINKED ISSUE #5803: Implement context-menu for 3D viewport
## Detailed Description
The other viewports already have a context menu. For the 3D viewport the following actions could be interesting:

- all node-related context actions we already have in the other viewports
- mesh-related actions, such as:
  - jump to position
  - remove mesh
  - reload mesh
  - reveal mesh in sidebar (see #5792)
- settings (e.g., "hide wireframe")? we already have that via a settings icon, but a context menu might also be a good place.

Would make the 3D viewport easier to use.

## Context

MichaelBuessemeyer:
This pr adds some mesh-related items to the context menu in the 3d viewport.
Note that finding/opening the mesh/segment of the mesh in the segments tab is not included, as this requires #5792 to be implemented. IMO this pr shouldn't wait for these changes as the current items are already valuable enough to be merged :shipit:.

Note, that this pr is affected by bug #6814. (The context menu needs to be opened twice under certain circumstances)

- https://addmeshactionstodcontextmenu.webknossos.xyz/

### Steps to test:
- Open a tracing with meshes. E.g. open a new hybrid tracing with an existing segmentation layer. Compute an ad-hoc mesh using the context menu in the e.g. xy viewport.
- Hover the mesh in the 3d viewport and open the context menu (likely needed twice). 
- Play around with the new options (hide, remove, jump, and listing the name / segment id)
- In case the name of the segment of the mesh is very long, the name will be abbreviated and fully available via a tooltip.

### Issues:
- fixes #5803

------
~~- [ ] Updated [(unreleased) migration guide](../blob/master/MIGRATIONS.unreleased.md#unreleased) if applicable~~
~~- [ ] Adapted [wk-connect](https://github.com/scalableminds/webknossos-connect) if datastore API changes~~
- [x] Ready for review


frontend/javascripts/oxalis/view/context_menu.tsx
```diff
@@ -1090,15 +1189,26 @@ function ContextMenuInner(propsWithInputRef: Props) {
...
         </InfoMenuItem>,
       );
     }
+    if (segments != null && maybeClickedMeshId != null) {
+      const segmentName = getNameOfSegment(segments.get(maybeClickedMeshId));
+      if (segmentName != null) {
+        infoRows.push(
+          <InfoMenuItem key="copy-cell">
+            <div className="cell-context-icon" />
+            Segment Name: {segmentName}
```
MichaelBuessemeyer, line 1207: I am unsure how to indicate that this menu field displays the segment's name. First, I wrote `Segment Name:` as a prefix. But this seems to be too long IMO. Therefore I replayed it with a name tag icon. Does this make sense? Do you have a better suggestion?

frontend/javascripts/oxalis/view/right-border-tabs/segments_tab/segment_list_item.tsx
```diff
@@ -176,6 +180,7 @@ type Props = {
 
 function getSegmentTooltip(segment: Segment) {
   if (segment.creationTime == null) {
+    // TODO: Ask whether the segment name would be better here.
     return `Segment ${segment.id}`;
   }
```
MichaelBuessemeyer, line 185: I think this function was added before segments had their own names. As I extracted the `getNameOfSegment` method, I would suggest also using it here. But as this hasn't been done previously, there might be a reason for this that I don't know.

So: Should I keep it at the current state or use the new extracted method to show the name of the segment, if it has one?

MichaelBuessemeyer:
I'll do the docs update and the changelog entry on Thursday (I think)

MichaelBuessemeyer:
@philippotto If you think daniel has more time / is better suited for the review, please just assign him 🙏

frontend/javascripts/oxalis/view/context_menu.tsx
```diff
@@ -342,9 +353,63 @@ function getMaybeMinCutItem(
...
+      key="reload-mesh"
+      onClick={() => refreshMesh(visibleSegmentationLayer.name, maybeClickedMeshId)}
+    >
+      Reload Mesh
+    </Menu.Item>,
+    <Menu.Item
+      key="jump-to-mesh"
+      onClick={() => {
+        const storeState = Store.getState();
+        const datasetScale = storeState.dataset.dataSource.scale;
```
philippotto, line 390: would it be easy to already receive the unscaled position here so that this doesn't need to happen here? I think, this division is done somewhere else, too. So, it could by DRYed a bit. However, I'm not sure whether the scale is easily known "further up".

frontend/javascripts/oxalis/view/context_menu.tsx
```diff
@@ -1090,15 +1188,29 @@ function ContextMenuInner(propsWithInputRef: Props) {
       );
     }
 
-    if (segmentIdAtPosition > 0) {
+    if (segmentIdAtPosition > 0 || maybeClickedMeshId != null) {
```
philippotto, line 1191: Could it be that both are defined? Or is one case only for the data viewports and the second case only for the 3D viewport? If both were possible, the user wouldn't know which ID is shown, right?

frontend/javascripts/oxalis/view/right-border-tabs/segments_tab/segment_list_item.tsx
philippotto: The main purpose of this function is to convey the "creation time" of the segment when hovering it in the segments list (only for very old segments, this information doesn't exist). The id is only mentioned for the sake of completeness (and because it's short). The name doesn't need to be mentioned here in my opinion, because the name is already what the user is currently hovering.

frontend/javascripts/libs/utils.ts
```diff
@@ -1103,3 +1103,7 @@ export function conjugate(
   }
   return `${verbStr}s`;
 }
+
+export function truncateStringToLength(str: string, length: number): string {
+  return str.length > length ? `${str.substring(0, length)}...` : str;
```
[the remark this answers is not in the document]
MichaelBuessemeyer: But isn't the condition necessary for the `...`?

Sure, I could move this to the context menu file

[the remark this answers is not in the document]
philippotto: > But isn't the condition necessary for the `...`?

Ah, sorry, yeah, you are right of course. Ignore my comment :)

frontend/javascripts/oxalis/view/context_menu.tsx
MichaelBuessemeyer: Yeah, the scale is easily available in the parent component. I just overlooked it 🙈

frontend/javascripts/oxalis/view/context_menu.tsx
MichaelBuessemeyer: It is not possible that both are defined. The segmentId is calculated based upon the calculated global position of where the context menu is opened. The function used for that is [_calculateMaybeGlobalPos](https://github.com/scalableminds/webknossos/blob/b3673cdfa979ac882baf5a0b5b35c65f4b8f79f3/frontend/javascripts/oxalis/model/accessors/view_mode_accessor.ts#L98). Thus in case of the 3d viewport, the segmentId will never be defined.
And the `maybeClickedMeshI` can only be defined when the context menu is opened in the 3d viewport. Thus if on variable is defined / `> 0` the other cannot. They exclude each other 🎉 .


I am personally not perfectly happy with my solution here, but other alternatives would me more code duplication. Thus I added a comment shortly explaining that only one of them can be defined.

frontend/javascripts/oxalis/view/context_menu.tsx
```diff
@@ -1188,6 +1190,10 @@ function ContextMenuInner(propsWithInputRef: Props) {
       );
     }
 
+    // Currently either segmentIdAtPosition or maybeClickedMeshId is set, but not both.
+    // segmentIdAtPosition is only set if a segment is hovered in one of the xy,x z, or yz viewports.
```
philippotto, line 1194:
```suggestion
    // segmentIdAtPosition is only set if a segment is hovered in one of the xy, xz, or yz viewports.
```

frontend/javascripts/oxalis/view/context_menu.tsx
philippotto: > It is not possible that both are defined. The segmentId is calculated based upon the calculated global position of where the context menu is opened. The function used for that is [_calculateMaybeGlobalPos](https://github.com/scalableminds/webknossos/blob/b3673cdfa979ac882baf5a0b5b35c65f4b8f79f3/frontend/javascripts/oxalis/model/accessors/view_mode_accessor.ts#L98). Thus in case of the 3d viewport, the segmentId will never be defined.
And the maybeClickedMeshI can only be defined when the context menu is opened in the 3d viewport. Thus if on variable is defined / > 0 the other cannot. They exclude each other tada .

Great :+1: Thanks for the explanation!

> I am personally not perfectly happy with my solution here, but other alternatives would me more code duplication. Thus I added a comment shortly explaining that only one of them can be defined.

Yeah, I think, it's a good trade-off.

philippotto:
Just gave it a whirl and it worked quite well :+1: Great work!

> Not, that this pr is affected by bug https://github.com/scalableminds/webknossos/issues/6814. (The context menu needs to be opened twice under certain circumstances)

Do you know these circumstances? For me, it always worked on the first try :thinking: 

> I am unsure how to indicate that this menu field displays the segment's name. First, I wrote Segment Name: as a prefix. But this seems to be too long IMO. Therefore I replayed it with a name tag icon. Does this make sense? Do you have a better suggestion?

Two ideas:
- [ ] I like the name tag icon, but I'd give it a tooltip (only the icon) which says "Segment Name"
- [ ] I'd only display the name if it really is defined. Otherwise, it's redundant (see screenshot).
image

One thing which caught my eye is this:
- [ ] If I rightclick somewhere in the 3d viewport where no mesh or skeleton is, I get the following context menu:
image
At first, I thought: cool, now I can even load a mesh for a segment which is rendered onto one of the planes in the 3d viewports. However, a) this doesn't work (would be complicated to support probably) and b) these context options also show up when rightclicking a white area. So, I assume this case is just not handled properly somehow? I think, it should simply show "No actions available" or something like this (I think, this string already exists in the context menu module somewhere).

MichaelBuessemeyer:
>Do you know these circumstances? For me, it always worked on the first try 🤔

Yeah, it is actually hard to reproduce. For me, it always worked when I created an ad-hoc mesh in the zy viewport and then immediately hovered over a mesh in the 3d viewport without clicking anywhere and then opening the context menu. Then it says "no data" 🤷‍♂️. 
The same bug already exists on the master:
- create a tracing
- create some nodes in the yz viewport
- open the context menu in the yz viewport and close it by also clicking somewhere in the yz viewport
- hover a skeleton node in the 3d viewport and right-click. The menu is empty although it shouldn't be. Then reponing the context menu shows options.


> One thing which caught my eye is this:

This definitely shouldn't be the case. I'll investigate this :)

MichaelBuessemeyer:
And also: Please note, that I added the change of words in the upload view where "uploaded" changed to "imported".

MichaelBuessemeyer:
I fixed the bug and mentioned the new mesh-related context menu items in the docs

philippotto (approved):
Awesome, works great :+1:

PR #6817, opened 2023-02-06
Fix saving allowed teams in dataset settings

philippotto:
- https://___.webknossos.xyz

### Steps to test:
- open the settings for a dataset
- add an allowed team in the sharing tab
- save
- ensure that the saving worked

### Issues:
- follow-up to #6776 

------
- [X] Ready for review


hotzenklotz (approved):
Thanks for fixing this. Tested and works!

PR #6819, opened 2023-02-06
Annotation Locking Mechanism

### LINKED ISSUE #6800: Automatic Annotation Locking Mechanism (for collaboration)
When allowing other users to edit an annotation, we currently expect the users to coordinate so that they don't work concurrently on the same annotation (this would result in conflicts). Until #6495 and #6791 are tackled, this can lead to problematic behavior for volume data.

This issue suggests an automatic locking mechanism which enforces that users cannot cause conflicts. I imagine it somewhat like this:

- the back-end holds a mutex per annotation ~(or per tracing?)~ (simpler than per tracing)
- when a user A opens an annotation, the mutex is acquired automatically and kept alive via a heartbeat (only necessary when other users may edit the annotation; so, don't do it when not shared?)
- when another user B opens an annotation and the mutex acquisition fails, a message à la `User X is currently editing the annotation. Only showing read-only mode.`. The front-end switches to read-only mode then.
- the front-end retries the mutex acquisition (analoguous to the heartbeat) and if it succeeds a message à la `You may now edit the annotation. Reload the page to do so.` (necessary to get the newest changes).
- a mutex is automatically released once the heartbeat is not received anymore (thus, when closing an annotation, this will happen)

A downside of this approach is that a mutex can be lost when being offline temporarily. However, the chances for this are low and a potential data loss is somewhat understandable when being offline without saving first (will only happen when somebody else edits inbetween obviously).

--------
Note that the `watchForSaveConflicts` needs to be adapted probably so that the wording and semantics make sense with the locking mechanism.

fm3:
new route `POST /api/annotations/<annotationId>/acquireMutex` (no body) will, for othersMayEdit annotations, return json body with two fields: canEdit: bool, blockedByUser: user (optional, only set if blocked. Uses compact version of user json, just like in user/owner in annotation json)

- https://___.webknossos.xyz

### Steps to test:
- create an additional user account / promote sample user2 to also be an admin.
- create an annotation with sample user 1, make it editable by others keep it open.
- Open the annotation in a private tab with sample user2. There it should switch to read only and show a toast that sample user is currently editing the annotation
- When you close the annotation with sample user and wait for max 3 minutes, the annotation should be available to sample user2 and automatically reload the annotation once the mutex is acquired.
- Now reopen the annotation with sample user and try to turn of `othersMayEdit`. This should not be possible, as sample user2 currently has the mutex.
- close the annotation with sample user2
- wait 3 mins and sample user should now be able to edit the annotation and the `othersMayEdit` setting again.#
- and of cause: Play around with it the way you want to discover uncovered edge cases :D

### Issues:
- fixes #6800

------
- [x] Ready for review


conf/application.conf
```diff
@@ -59,6 +59,7 @@ webKnossos {
   cache {
     user.timeout = 3 minutes
   }
+  annotation.mutex.expiryTime = 2 minutes
```
MichaelBuessemeyer, line 62: @fm3 As this is a setting that might be changed, I think it would be a good idea to expose it to the frontend. Thus the frontend can adjust to the timeout duration so it requests the mutex in time.

If that setting is not designed to be changed, ignore my post 📰

MichaelBuessemeyer:
@fm3 I don't know whether that's needed :)

> [ ] Needs datastore update after deployment

frontend/javascripts/oxalis/model/sagas/annotation_saga.ts
```diff
@@ -159,4 +178,75 @@ export function* watchAnnotationAsync(): Saga<void> {
...
+        if (doesHaveMutex === true || isInitialRequest) {
+          onMutexStateChanged(false, null);
+          yield* put(setAnnotationAllowUpdateAction(false));
+        }
+      }
+      isInitialRequest = false;
+      yield* call(delay, ONE_MINUTE);
+    }
+  }
+  if (shallTryAcquireMutex) {
```
philippotto, line 236: I think it would be fine to remove the condition, since the forked saga itself has the same condition as a while-loop anyway.

frontend/javascripts/oxalis/model/sagas/annotation_saga.ts
```diff
@@ -159,4 +178,75 @@ export function* watchAnnotationAsync(): Saga<void> {
...
+        blockedByUser != null
+          ? messages["annotation.acquiringMutexFailed"]({
+              userName: `${blockedByUser.firstName} ${blockedByUser.lastName}`,
+            })
+          : messages["annotation.acquiringMutexFailed.noUser"];
+      Toast.error(message, { sticky: true, key: "MutexCouldNotBeAcquired" });
+    }
+  }
+
+  function* tryAcquireMutex(): Saga<void> {
```
philippotto, line 213: i would rename it to something like `tryAcquireMutexContinuously` to signal that this is acquiring continuously.

frontend/javascripts/oxalis/model/sagas/annotation_saga.ts
```diff
@@ -159,4 +178,75 @@ export function* watchAnnotationAsync(): Saga<void> {
...
+  }
+  if (shallTryAcquireMutex) {
+    yield* fork(tryAcquireMutex);
+  }
+  function* reactToOthersMayEditChanges({
+    othersMayEdit,
+  }: SetOthersMayEditForAnnotationAction): Saga<void> {
+    shallTryAcquireMutex = othersMayEdit;
+    if (shallTryAcquireMutex) {
+      yield* call(tryAcquireMutex);
```
philippotto, line 244: I think it can happen that multiple instances of `tryAcquireMutex` are running. Imagine something like this:

- tryAcquireMutex runs because othersMayEdit is true
- the user disables othersMayEdit and re-enables it again quickly (so quickly that the while loop in the saga doesn't terminate; e.g., because it's sleeping in that period)
- now, line 244 will spawn a second instance of tryAcquireMutex

I'd suggest to store a reference of the forked/called saga and then cancel it before starting a new one. See https://redux-saga.js.org/docs/advanced/TaskCancellation for reference.

frontend/javascripts/oxalis/model/sagas/annotation_saga.ts
```diff
@@ -159,4 +178,75 @@ export function* watchAnnotationAsync(): Saga<void> {
...
+        yield* put(setAnnotationAllowUpdateAction(canEdit));
+        if (canEdit !== doesHaveMutex || isInitialRequest) {
+          doesHaveMutex = canEdit;
+          onMutexStateChanged(canEdit, blockedByUser);
+        }
+      } catch (error) {
+        console.error("Error while trying to acquire mutex.", error);
+        if (doesHaveMutex === true || isInitialRequest) {
+          onMutexStateChanged(false, null);
+          yield* put(setAnnotationAllowUpdateAction(false));
```
philippotto, line 229: if the mutex acquisition fails because i'm offline, I'll immediately lose it. how about remembering the timestamp of the last acquisition and comparing it to the timeout constant? So, if I acquired the mutex one minute ago, I should still have it, right? maybe just add a retry around the acquisition request?

philippotto: also, with the current behavior I would have expected that the catch clause sets `doesHaveMutex` to false.

frontend/javascripts/oxalis/model/sagas/annotation_saga.ts
```diff
@@ -159,4 +178,75 @@ export function* watchAnnotationAsync(): Saga<void> {
...
+  let isInitialRequest = true;
+  let doesHaveMutex = false;
+  let shallTryAcquireMutex = othersMayEdit;
+
+  function onMutexStateChanged(canEdit: boolean, blockedByUser: APIUserCompact | null | undefined) {
+    if (canEdit) {
+      Toast.close("MutexCouldNotBeAcquired");
+      if (!isInitialRequest) {
+        Toast.success(messages["annotation.acquiringMutexSucceeded"]);
+        location.reload();
```
philippotto, line 200: I don't think that this should happen automatically. I'd add a button to the toast so that the users workflow isn't interrupted (maybe they were just inspecting things).

philippotto (changes requested):
Great stuff :+1: The code looks very good. I only left some nitpicks/edge cases.

Kudos for writing the tests :100: Didn't review those yet, though. I'll test after the feedback is addressed :)

frontend/javascripts/messages.tsx
```diff
@@ -259,6 +259,13 @@ instead. Only enable this option if you understand its effect. All layers will n
   "annotation.undoFinish.confirm": "Are you sure you want to reopen your old task?",
   "annotation.undoFinish.content":
     "If you reopen your old annotation, the current annotation will not be finished or cancelled. Instead, it will remain open and you can find it in the dashboard to continue annotating.",
+  "annotation.acquiringMutexFailed": _.template(
+    "This annotation is currently being edited by <%- userName %>. To avoid conflicts, you can only view it. If you want to edit it, please ask <%- userName %> to finish their work first.",
+  ),
+  "annotation.acquiringMutexFailed.noUser":
+    "This annotation is currently being edited. To avoid conflicts, you can only view it. The server did not provide any information about the user who is currently editing it. You should check your connection to the server.",
```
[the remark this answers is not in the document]
MichaelBuessemeyer: 👍 It only happens in case the request fails due to e.g. lost connection

frontend/javascripts/oxalis/model/sagas/annotation_saga.ts
fm3: Would it help if I added a get route for the mutex that only yields info about its current state, without trying to acquire it?

frontend/javascripts/oxalis/model/sagas/annotation_saga.ts
MichaelBuessemeyer: Ah nice, thanks for spotting that and also directly providing a solution 🙏 🚀

frontend/javascripts/oxalis/model/sagas/annotation_saga.ts
MichaelBuessemeyer: I don't think so. At that point, the user already has acquired the mutex and the frontend will try to keep it. I don't see where such an info route would be useful 🤔

frontend/javascripts/oxalis/model/sagas/annotation_saga.ts
MichaelBuessemeyer: Good catch. I just changed to code so that it retries up to one minute and that `doesHaveMutex` will be set to false

frontend/javascripts/oxalis/model/sagas/annotation_saga.ts
MichaelBuessemeyer: Do you think I should add another test testing the cancelation mechanism? I am asking because this might take some more time as I already invested quite some time into the tests

philippotto: > Do you think I should add another test testing the cancelation mechanism? I am asking because this might take some more time as I already invested quite some time into the tests

I think, it's not necessary. If each fork/call invocation is accompanied by a cancel call, this seems quite robust to me :)

frontend/javascripts/oxalis/model/sagas/annotation_saga.tsx
```diff
@@ -188,67 +193,84 @@ export function* acquireAnnotationMutexMaybe(): Saga<void> {
...
         if (canEdit !== doesHaveMutex || isInitialRequest) {
           doesHaveMutex = canEdit;
           onMutexStateChanged(canEdit, blockedByUser);
         }
       } catch (error) {
-        console.error("Error while trying to acquire mutex.", error);
-        if (doesHaveMutex === true || isInitialRequest) {
-          onMutexStateChanged(false, null);
-          yield* put(setAnnotationAllowUpdateAction(false));
+        const ifGotCanceled = yield* cancelled();
```
philippotto, line 243: `wasCanceled` (or `gotCanceled`) seems more idiomatic to me.

frontend/javascripts/oxalis/model/sagas/annotation_saga.tsx
```diff
@@ -188,67 +193,84 @@ export function* acquireAnnotationMutexMaybe(): Saga<void> {
...
+        if (!ifGotCanceled) {
+          console.error("Error while trying to acquire mutex.", error);
+          if (doesHaveMutex === true || isInitialRequest) {
```
philippotto, line 246:
```suggestion
          if (doesHaveMutex || isInitialRequest) {
```

app/oxalis/telemetry/SlackNotificationService.scala
```diff
@@ -48,4 +48,10 @@ class SlackNotificationService @Inject()(rpc: RPC, config: WkConf) extends LazyL
       title = "Task creation with base",
       msg = s"$numberOfTasks tasks with BaseAnnotation for TaskTypes ${taskType.mkString(", ")} have been created"
     )
+
+  def noticeSlowRequest(msg: String): Unit =
```
jstriebel, line 52: This was already used before this PR, how is this possible? :thinking:

conf/evolutions/100-annotation-mutexes.sql
```diff
@@ -0,0 +1,18 @@
+START TRANSACTION;
+
+do $$ begin ASSERT (select schemaVersion from webknossos.releaseInformation) = 99, 'Previous schema version mismatch'; end; $$ LANGUAGE plpgsql;
```
jstriebel, line 3: :heart:

app/models/annotation/AnnotationMutexService.scala
```diff
@@ -0,0 +1,119 @@
...
+    logger.info("Cleaning up expired annotation mutexes...")
+    annotationMutexDAO.deleteExpired()
+    ()
+  }
+
+  private val defaultExpiryTime = wkConf.WebKnossos.Annotation.Mutex.expiryTime
+
+  def tryAcquiringAnnotationMutex(annotationId: ObjectId, userId: ObjectId)(
+      implicit ec: ExecutionContext): Fox[MutexResult] =
+    this.synchronized {
```
jstriebel, line 46: I was assuming that the mutex is handled via an sql transaction instead of local locking, but probably also fine :+1:

app/controllers/AnnotationController.scala
```diff
@@ -557,6 +561,7 @@ class AnnotationController @Inject()(
     sil.SecuredAction.async { implicit request =>
       for {
         annotation <- provider.provideAnnotation(typ, id, request.identity)
+        _ <- bool2Fox(annotation.typ == AnnotationType.Explorational || annotation.typ == AnnotationType.Task) ?~> "annotation.othersMayEdit.onlyExplorationalOrTask"
```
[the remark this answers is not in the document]
fm3: There are no other kinds of annotations in the database. All others are just the Compound annotations, which don’t exist in postgres. So no need to express this in postgres explicitly, this is just for a nicer error message

app/oxalis/telemetry/SlackNotificationService.scala
fm3: There is also a DSSlackNotificationService and a TSSlackNotificationService. This could be made DRYer by using a trait, but since the classes are small, I think it’s fine for the moment.
The wk-side one didn’t have this function before.

app/models/annotation/AnnotationMutexService.scala
fm3: Yes, that would be the next iteration of this feature. Because of this, I added the noticeSlowRequest with a tight timeout, so that we will know when this variant blocks too many requests. My guess is that it is not a problem, and we can upgrade to an sql-only version later. It would probably be a bit harder to write and understand, as it would need to implement more of the logic in sql, be a full test-and-set, depending on whether the locking user is the requesting one etc.

jstriebel (commented):
Backend LGTM, leaving the approval for @philippotto

philippotto:
I just tested the workflow and noticed two things:

- If I open an annotation in two tabs (with different users) and let one mutex expire (e.g., by disabling its network connection and waiting for 2 minutes (I probably should have checked it out locally and tweaked the timeouts ^^)), the second tab will acquire the mutex successfully. **However**, if I **don't** use the "reload" button in the toast (by the way, I think, the toast should be sticky), I can still edit the annotation. The save button will always show a checkmark in that state (as if I was in sandbox mode). That state should never be reached to avoid data loss. Not sure what's happening exactly, but I could reproduce it twice, so it's probably not flaky.
- Only a minor thing: "This annotation is currently being edited by Sample User. To avoid conflicts, you can only view it. If you want to edit it, please ask Sample User to finish their work first." should be a warning instead of an error toast.

Other than that, the feature looks quite good :+1:

frontend/javascripts/oxalis/model/sagas/annotation_saga.ts
MichaelBuessemeyer: Ok, ss the cancellation is mixed into the saga, the test broke due to this change. Therefore I had to fix the tests and thus also test the cancellation mechanism anyway.
So the canceling is now tested anyway 🤷

frontend/javascripts/oxalis/model/sagas/annotation_saga.tsx
```diff
@@ -234,29 +233,31 @@ export function* acquireAnnotationMutexMaybe(): Saga<void> {
           acquireAnnotationMutex,
           annotationId,
         );
-        yield* put(setAnnotationAllowUpdateAction(canEdit));
+        if (isInitialRequest && canEdit) {
+          // Only set allow update to true in case the first try to get the mutex succeeded.
+          yield* put(setAnnotationAllowUpdateAction(true));
+        } else {
+          yield* put(setAnnotationAllowUpdateAction(false));
+        }
```
philippotto, line 241:
```suggestion
        // Only set allow update to true in case the first try to get the mutex succeeded.
        yield* put(setAnnotationAllowUpdateAction(isInitialRequest && canEdit));
```

frontend/javascripts/messages.tsx
```diff
@@ -259,6 +259,13 @@ instead. Only enable this option if you understand its effect. All layers will n
   "annotation.undoFinish.confirm": "Are you sure you want to reopen your old task?",
   "annotation.undoFinish.content":
     "If you reopen your old annotation, the current annotation will not be finished or cancelled. Instead, it will remain open and you can find it in the dashboard to continue annotating.",
+  "annotation.acquiringMutexFailed": _.template(
+    "This annotation is currently being edited by <%- userName %>. To avoid conflicts, you can only view it. If you want to edit it, please ask <%- userName %> to finish their work first.",
+  ),
+  "annotation.acquiringMutexFailed.noUser":
+    "This annotation is currently being edited by someone else. To avoid conflicts, you can only view it at the moment.",
+  "annotation.acquiringMutexSucceeded":
+    "This annotation is not being edited available for editing. Reload the page to see its newest version and to edit it.",
```
philippotto, line 268:
```suggestion
    "This annotation is not being edited anymore and available for editing. Reload the page to see its newest version and to edit it.",
```

frontend/javascripts/test/sagas/annotation_saga.spec.ts
```diff
@@ -0,0 +1,188 @@
...
+    othersMayEdit,
+    annotationId: "1234",
+  },
+});
+
+const { wkReadyAction } = mockRequire.reRequire("oxalis/model/actions/actions");
+const { acquireAnnotationMutexMaybe } = mockRequire.reRequire("oxalis/model/sagas/annotation_saga");
+
+test.serial(
+  "A annotation with allowUpdate = false should not try to acquire the annotation mutex.",
```
philippotto, line 31:
```suggestion
  "An annotation with allowUpdate = false should not try to acquire the annotation mutex.",
```

philippotto: also there are some other occurrences of that in the rest of the module.

frontend/javascripts/oxalis/model/sagas/annotation_saga.tsx
philippotto: while my suggestion should be equal to the current code, I just noticed during testing that the logic itself is faulty. at the moment, on each iteration that is not the first, the boolean will be false and thus disabling the "is allowed" property. thus, I cannot edit my own annotation even when I got the mutex.

philippotto:
another thing I just noticed: If I'm the owner of an annotation, but don't have the mutex currently, I cannot disable the "may others edit" property (because the annotation is read-only for me). one could argue that this is fine, because the owner shouldn't disturb another user that is currently working on it, but I feel like this can be annoying if a user decides that they want to regain full control.

@fm3 @MichaelBuessemeyer do you have thoughts on that (regardless of implementation)?
implementation-wise it might only be a front-end thing (not sure, though).

philippotto (changes requested):
see my one comment. I think, the logic is a bit broken now so that the mutex cannot be kept. I wonder why the tests didn't catch that?

frontend/javascripts/oxalis/model/sagas/annotation_saga.tsx
MichaelBuessemeyer: > at the moment, on each iteration that is not the first, the boolean will be false and thus disabling the "is allowed" property. thus, I cannot edit my own annotation even when I got the mutex.

Ah, I see what the problem is. Sorry, that I missed that. I did not test manually for long enough for this to happen 😅 .

The code now remembers whether the mutex could be acquired from the beginning. Now the code only sets the "is allowed" to false, if the mutex could wasn't acquired from the beginning or `canEdit` sent by the server is false. This should be the correct logic right?

The new logic should now be:
=> Keep `allowUpdate` true in case the mutex was acquired initially and the server sends back `canEdit = true`
=> once the server sends `canEdit = false` `allowUpdate` will never be set to true again until the user reloads the page.
=> If the mutex wasn't acquired initially, the `allowUpdate` will always be false until the user reloads the page.

This should cover all cases now, right? And blocking all updates from the annotation once the mutex was lost only a single time should be correct, right? Because there could be another potential user that had the mutex in the meantime and made some small changes to the annotation.

frontend/javascripts/oxalis/model/sagas/annotation_saga.tsx
```diff
@@ -159,4 +182,106 @@ export function* watchAnnotationAsync(): Saga<void> {
...
+
+export function* acquireAnnotationMutexMaybe(): Saga<void> {
+  yield* take("WK_READY");
+  const allowUpdate = yield* select((state) => state.tracing.restrictions.allowUpdate);
+  const annotationId = yield* select((storeState) => storeState.tracing.annotationId);
+  if (!allowUpdate) {
+    return;
+  }
+  const othersMayEdit = yield* select((state) => state.tracing.othersMayEdit);
+  const ONE_MINUTE = 1000 * 60;
```
philippotto, line 194: I'd rather give this variable a semantic meaning (e.g., acquireMutexInterval).

philippotto (changes requested):
Works excellent now :+1: Only left one renaming suggestion..

UX-wise one thing caught my attention: If I'm viewing my own shared annotation, but don't have the mutex, the navbar switches to the read-only mode which is fine per se. However, (apart from the toast which can be closed), the UI doesn't tell me anything about this state. The "Copy to my account" even suggests that this annotation isn't mine (even though it is). Do you think the navbar could be tweaked a bit in that state? I think, you could check for `annotation.owner == activeUser && isReadOnly`. In that case, the copy-button could simply say "Duplicate" and an antd tag could be next to the avatar saying "Locked by ..." (with a tooltip which explains a bit more).

MichaelBuessemeyer:
@philippotto Thanks for the suggestions. Using a tag as a component doesn't seem very visually appealing to me, as it does not fit into the style of the navbar. But I cannot think of another component that could be used here instead :/

I additionally added a tooltip to the tag that uses the same message as the toast. 

Visually it looks like this now:
image
I did not know where exactly to put it. Here is the other variation.
image
I personally prefer the second one.
I could also put the tag as the right-most item, but I think that the avatar should always be the right-most item. To I left out this possibility.

Another thing I noticed is that these messages frequently pop up:
image
Should I add the possibility to disable them or something like that? I think they are quite annoying. And I'd guess if another person is actively editing the annotation, this notification will nearly always pop up and annoy the user that is just waiting or viewing the annotation.


And this is an example image of the renaming of the "copy to my account" button:
image


I'll fix the tests now 😅

philippotto:
Awesome!

> @philippotto Thanks for the suggestions. Using a tag as a component doesn't seem very visually appealing to me, as it does not fit into the style of the navbar. But I cannot think of another component that could be used here instead :/

> Visually it looks like this now: image I did not know where exactly to put it. Here is the other variation. image I personally prefer the second one.

I prefer the second version, too, and think it looks alright :+1: 

> Another thing I noticed is that these messages frequently pop up: image Should I add the possibility to disable them or something like that? I think they are quite annoying. And I'd guess if another person is actively editing the annotation, this notification will nearly always pop up and annoy the user that is just waiting or viewing the annotation.

Hm, yeah, you're probably right. Maybe adapt the saga to not display the warnings if the "others may edit" setting is true (in that case one has either locked the annotation --> no problem. OR: someone else has locked it --> navbar already shows what's up).

> And this is an example image of the renaming of the "copy to my account" button: image

:+1:

MichaelBuessemeyer:
And this is how it looks, when the user acquired the mutex and only needs to reload:
image

philippotto (approved):
Awesome, lets ship it :)

frontend/javascripts/navbar.tsx
```diff
@@ -706,6 +767,11 @@ function Navbar({ activeUser, isAuthenticated, isInAnnotationView, hasOrganizati
       menuItems.push(getTimeTrackingMenu(collapseAllNavItems));
     }
 
+    if (othersMayEdit && !allowUpdate) {
+      trailingNavItems.push(
+        <AnnotationLockedByUserTag blockedByUser={blockedByUser} activeUser={activeUser} />,
```
MichaelBuessemeyer, line 772: Looks like I forgot to add a unique key attribute here 🙈

PR #6831, opened 2023-02-08
Loki batched insert

normanrz:
Batches the logs ingested into loki by 1000 messges.

------
- [x] Ready for review


app/models/voxelytics/LokiClient.scala
```diff
@@ -28,7 +28,8 @@ class LokiClient @Inject()(wkConf: WkConf, rpc: RPC, val system: ActorSystem)(im
 
   private val POLLING_INTERVAL = 1 second
   private val LOG_TIME_BATCH_INTERVAL = 1 days
-  private val LOG_ENTRY_BATCH_SIZE = 5000L
+  private val LOG_ENTRY_QUERY_BATCH_SIZE = 5000L
```
fm3, line 31:
```suggestion
  private val LOG_ENTRY_QUERY_BATCH_SIZE = 5000
```
(I guess there is no real reason to have this be a Long? I’d change it for consistency with the new constant)

fm3 (approved):
LGTM, just a minor suggestion

PR #6833, opened 2023-02-08
label loki logs with pid

normanrz:
Group log entries for loki by pid. This helps to split up the streams into contiguous streams.
Already deployed for mhlab.

- [x] Ready for review


fm3 (approved):
Code LGTM. I do not, however, understand what this is for 😅

PR #6835, opened 2023-02-08
fixes for loki batch limits

normanrz:
Another fix wrt the `limit` when querying loki.
Already deployed on mhlab. It works, but I'm happy to receive code-level suggestions.

------
- [x] Ready for review


fm3 (approved):
I don’t really have a good way of improving this.
In terms of style, it may be a good idea to extract some of the functionality in this file into smaller functions, some are getting quite long. A candidate may be the parsing of the json results.

PR #6840, opened 2023-02-10
Add TS types for search inputs

hotzenklotz:
Small PR. I added and fixed some TS type for event callbacks of the `<Search>` input. We are using a live, search-on-every-change approach so every `onPressEnter` event handler was unused anyway. However, binding `onPressEnter` and `onChange` events to the same handler function messes with typing. So I removed the unused `onPressEnter` events.

- https://___.webknossos.xyz

### Steps to test:
- Run type check
- Go to one of the admin pages and make sure that search still works.

------
- [x] Ready for review


frontend/javascripts/dashboard/explorative_annotations_view.tsx
```diff
@@ -819,8 +815,8 @@ function TopBar({
         width: 200,
         float: "right",
       }}
-      onPressEnter={handleSearchPressEnter}
-      onChange={handleSearch}
+      onSearch={handleOnSearch}
```
hotzenklotz, line 818: This component is the only one that does something on "ENTER". I switched to the `<Search>`input `onSearch` props instead since that also handles clicks to the little search icon button, ENTER key etc

philippotto (approved):
Splendid :monocle_face:

PR #6844, opened 2023-02-13
Update style of login, signup views

hotzenklotz:
I updated the style of the login, signup, password reset, and change password pages as suggested by Albane. Mostly I just added a colorful, flashy background with some neurons. 

The included SVG is not the smallest images. About 280kb gzipped. I tried to "optimize" it with [SVGOmg](https://jakearchibald.github.io/svgomg/) but the savings were neglectable.

Tested with Chrome, Safari, Firefox.

image

image

- https://___.webknossos.xyz

### Steps to test:
- Open all the 4 pages and check styling
- Use dev tools to try responsive styling
- Try dark mode

### Issues:
- None

------


fm3:
I tried my hand at optimizing the SVG for file size, got from 429kb down to 39kb uncompressed. It does not look exactly identical (how important is it that the left neuron has stronger lines than the left one? /cc @Albane-dlv 
Find it here (note that it will look all-white on the white github background)

background_neurons_drawn_optimized

philippotto:
Awesome :+1: Looks very cool!

Regarding @fm3's optimized SVG: If we want a less pronounced version, we could apply a blur(1px) filter. However, I'm fine with the optimized version as is :+1: 

> password reset, and change password pages

When I navigated to these pages, my initial thought was that it feels like I'm in the "logged-out area" now. Not sure whether that's just me or whether this flashy-background-theme is usually associated with being logged out. Apart from that, these pages look incorrect with the dark theme. See here:

image

I'd be fine with not styling these pages, as they are not used frequently.

fm3:
>  Regarding @fm3's optimized SVG: If we want a less pronounced version, we could apply a blur(1px) filter. 

Or just tweak the alpha values again. I chose a compromise between the left and right neuron. (In my version, they have to be identical, because one is just a transformed clone of the other, with the svg <use> tag)

hotzenklotz:
> I agree with Philipp that it might be less confusing to skip the change-password page in this redesign 🤔

I reverted the design for that site.

philippotto (approved):
Really loving this :+1: :man_dancing:

PR #6847, opened 2023-02-14
Readable error message if team name is taken, even if other team is deleted

### LINKED ISSUE #6787: duplicate team name yields raw sql error
Saw `ERROR: duplicate key value violates unique constraint "teams_name__organization_key"` – It should be checked before if the name is still available, giving a readable error message to the user

fm3:
Names are, unfortunately, also blocked by deleted entities. However, this assertion only checked non-deleted ones. It should count all, though, otherwise the SQL constraint error is passed to the user in raw form.

### Steps to test:
- Create team, delete it
- Create new team of same name
- Should fail with readable error message (no raw SQL error)

### Issues:
- fixes #6787


jstriebel (approved):
:+1:

Issue #6696, opened 2022-12-12
Extend JS API with mapping and mesh functionality


philippotto:
Something like this:

```js
/**
* Returns a list of segments that are all mapped to the same ID by the specified mapping/ agglomerate file.
*
 * For the `segment_to_agglomerate` array [0, 1, 2, 1, 1, 2, 3, 4] in the agglomerate file `agglomerate_view_80.hdf5`:
* @example
* const mappedID = 1;
* const mappingName = "agglomerate_view_80";
* const mappingType = "HDF5";
* fragments = api.data.getAllMappedFragments(mappedID, mappingName, mappingType);
*
 * `fragments` is equal to the array [1, 3, 4].
*
 * If `mappingName` is not specified, the active mapping is used.
*
 */
getAllMappedFragments(
    mappedID: number,
    mappingName?: string,
    mappingType?: MappingType = "HDF5",
    ): Array<number>
 
/**
* Returns the resulting mapping value for the given segment ID with respect to the specified mapping/ agglomerate file.
 *
 * For the `segment_to_agglomerate` array [0, 1, 2, 1, 1, 2, 3, 4] in the agglomerate file `agglomerate_view_80.hdf5`:
* @example
* const segmentId = 5;
* const mappingName = "agglomerate_view_80";
* const mappingType = "HDF5";
* mappedId = api.data.getMappedId(segmentId, mappingName, mappingType)
*
 * `mappedId` is equal to 2
*
 */
getMappedId(
    segmentId: number,
    mappingName?: string,
    mappingType?: MappingType = "HDF5"
    ): number

/**
* Gets the RGB color of a 3D mesh
*/
getMeshColor(
    meshId: number,
): Array<number>
 
/**
* Sets the RGB color of a 3D mesh
*/
setMeshColor(
    meshId: number,
    rgbColor: Array<number>,
): void
```

See https://scm.slack.com/archives/C5AKLAV0B/p1670510983692359 for context.

daniel-wer:
I added `getSegmentColor` and `setSegmentColor` functions in https://github.com/scalableminds/webknossos/pull/6853. I changed the naming to "Segment", because the mesh color mirrors the segment color and the segment is more general.

For the other two functions, new backend functionality is needed as far as I can see.

PR #6853, opened 2023-02-15
Add functions to get and set segment colors to the frontend API

daniel-wer:
Also, fix a bug in the segment color picker where the color that was preselected after opening the picker was wrong.
The `hexToRgb` utils function returned the color in [0, 255], but later it was assumed to be in [0, 1] and was scaled to 255 again. Because `hexToRgb` was only used for the segment color picker, I changed its return value to be in [0, 1] which we usually use in other related functions.

- https://___.webknossos.xyz

### Steps to test:
- Create an annotation on a dataset with a segmentation layer. Select a segment and compute an ad-hoc mesh for it.
- Change its color using the segments list.
  - Reopen the color picker. The preselected color should be correct.
- Use `webknossos.apiReady(3).then(async (api) => {console.log(api.data.getSegmentColor(<YOUR_SEGMENT_ID>))})` to get the color of the segment via the api -> The color should be the previously selected color in RGB in the interval [0, 1]
- Use `webknossos.apiReady(3).then(async (api) => {console.log(api.data.setSegmentColor(<YOUR_SEGMENT_ID>, [0, 1, 1]))})` to set the color of the segment (and mesh) -> The segment and mesh should change their color to turquoise

### Issues:
- contributes to #6696 

------


frontend/javascripts/oxalis/shaders/segmentation.glsl.ts
```diff
@@ -166,22 +166,22 @@ export const convertCellIdToRGB: ShaderModule = {
   `,
 };
 // This function mirrors the above convertCellIdToRGB-function.
-// Output is in [0,1] for H, S, L and A
-export const jsConvertCellIdToHSLA = (
+// Output is in [0,1] for R, G, B, and A
+export const jsConvertCellIdToRGBA = (
   id: number,
-  customColors?: Array<number> | null | undefined,
+  customColors?: Array<Vector3> | null | undefined,
```
daniel-wer, line 172: None of the callers uses this parameter which is why I changed its format. I think it's purely there to "mirror" the functionality of the `convertCellIdToRGB` shader function as pointed out in the comment above. Since the `customColors` are in RGB format in the shader as well, I figured it's fine to mirror that, here.

philippotto (approved):
100/100 :+1:

PR #6856, opened 2023-02-16
fix reversion being linked in MIGRATIONS.released.md

thopp-tudelft:
The MIGRATIONS.released.md had a significant flaw in it, the listed migrations were instead the database evolution needed to revert the migration for 94 and 95, this seems to have been missed in both 14dbb479fb and 6bedf79da4.

I have corrected the file here (and rolled back my database from backups).

### DIFF
--- a/MIGRATIONS.released.md
+++ b/MIGRATIONS.released.md
@@ -18,8 +18,8 @@ None.
 
 ### Postgres Evolutions:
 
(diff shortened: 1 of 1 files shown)


hotzenklotz (approved):

fm3:
I hope it does not bother you, but we could not (quickly) figure out how to make the CI run on your branch so we duplicated this change in PR #6859 and merged that one. Thank you again for notifying us about this error.

thopp-tudelft:
you're welcome

PR #6863, opened 2023-02-21
Fix exploring datasets via http (without s)

fm3:
The FileSystemProviders already support this, there was just some checks in the frontend and the createCredential codepath that didn’t handle http without s.

@philippotto Do you think the front-end fix is ok? I did not add http as a selectedProtocol, but rather reuse https in that case.

### Steps to test:
- Insert http (not https) uri in the Add Remote Dataset view
- Should work


philippotto (approved):
LGTM :)

PR #6867, opened 2023-02-22
Ensure that segments are added to segment list when using quick select...

### LINKED ISSUE #6858: Quick-Selecting voxels does not add them to segments list
When using the quick-select tool with a new segment ID or empty volume layer, then the resulting segment is not listed in the "Segments" tab in the right-hand side panel. It looks like we overlooked this step in the implementation of the quick select tool.

## Expected Behavior
Quick selecting new voxel should add them to the segments list

## Current Behavior
Quick selecting new voxel does not add them to the segments list

## Steps to Reproduce the bug
1. Create a new volume annotation
2. Without having labelled any segments (=segement list is empty), switch to quick select tool and annotated something
3. Check the "Segments" tab in the right-hand side panel. It remains empty

## Your Environment for bug

- Browser name and version: Tested in Chrome & Safari 
- Operating System and version: Mac
- Version of WebKnossos (Release or Commit): Version 21688

philippotto:
… or flood fill to create new segments

- https://___.webknossos.xyz

### Steps to test:
- use quick select and floodfill to create new segments --> should be added to segment list
- floodfill can be tested best by using it to relabel an existing segment (and 2D mode)

### Issues:
- fixes #6858

------


hotzenklotz (approved):
LGTM. Works like a charm.

PR #6871, opened 2023-02-23
Add evolutions 99,100 to migration guide

fm3:
Someone (me) missed adding these to the migration guide. Luckily, they did not yet make it into releases, so this should not have caused harm


philippotto (approved):
:+1:

philippotto:
> Luckily, they did not yet make it into releases, so this should not have caused harm

The release process includes a step to check that no evolutions are missing in the migration guide. Therefore, I would have (hopefully) spotted that :)

PR #6874, opened 2023-02-24
Followups for OME-TIFF export

normanrz:
- Fixes download filename
- Fixes backend error when requesting larger bounding boxes (wasn't scaled by mag)

### Steps to test:
- Set a smaller `exportTiffMaxVolumeMVx`
- Try to download some data of a large bbox (e.g. full layer) in a high mag
- Should download the file with a proper filename: `date__dataset__layer.ome.tiff`

------

### DIFF
--- a/frontend/javascripts/admin/admin_rest_api.ts
+++ b/frontend/javascripts/admin/admin_rest_api.ts
@@ -985,6 +985,27 @@ export function convertToHybridTracing(
   });
 }
 
+export async function downloadWithFilename(downloadUrl: string) {
+  const { buffer, headers } = await Request.receiveArraybuffer(downloadUrl, {
+    extractHeaders: true,
+  });
+
+  // Using headers to determine the name and type of the file.
+  const contentDispositionHeader = headers["content-disposition"];
+  const filenameStartingPart = 'filename="';
+  const filenameStartingPosition =
+    contentDispositionHeader.indexOf(filenameStartingPart) + filenameStartingPart.length;
+  const filenameEndPosition = contentDispositionHeader.indexOf('"', filenameStartingPosition + 1);
+  const filename = contentDispositionHeader.substring(
+    filenameStartingPosition,
+    filenameEndPosition,
+  );
+  const blob = new Blob([buffer], {
+    type: headers["content-type"],
+  });
+  saveAs(blob, filename);
+}
+
 export async function downloadAnnotation(
   annotationId: string,
   annotationType: APIAnnotationType,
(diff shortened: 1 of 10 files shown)


util/src/main/scala/com/scalableminds/util/geometry/BoundingBox.scala
```diff
@@ -55,6 +55,8 @@ case class BoundingBox(topLeft: Vec3Int, width: Int, height: Int, depth: Int) {
   def *(that: Vec3Int): BoundingBox =
     BoundingBox(topLeft * that, width * that.x, height * that.y, depth * that.z)
 
+  def /(that: Vec3Int): BoundingBox = BoundingBox(topLeft / that, width / that.x, height / that.y, depth / that.z)
```
philippotto, line 58: will this floor? should it maybe ceil to avoid that voxels get lost at the border?

frontend/javascripts/admin/admin_rest_api.ts
```diff
@@ -1005,24 +1013,7 @@ export async function downloadAnnotation(
...
-    contentDispositionHeader.indexOf(filenameStartingPart) + filenameStartingPart.length;
-  const filenameEndPosition = contentDispositionHeader.indexOf('"', filenameStartingPosition + 1);
-  const filename = contentDispositionHeader.substring(
-    filenameStartingPosition,
-    filenameEndPosition,
-  );
-  const blob = new Blob([buffer], {
-    type: headers["content-type"],
-  });
-  saveAs(blob, filename);
```
philippotto, line 1025: Nice clean up :+1:

philippotto (commented):
Frontend lgtm :+1: Only tested the (named) download of annotations (without export) due to the dev instance not having a worker set up. I assume that you did test it locally.

util/src/main/scala/com/scalableminds/util/geometry/BoundingBox.scala
```diff
@@ -55,6 +56,9 @@ case class BoundingBox(topLeft: Vec3Int, width: Int, height: Int, depth: Int) {
   def *(that: Vec3Int): BoundingBox =
     BoundingBox(topLeft * that, width * that.x, height * that.y, depth * that.z)
 
+  def /(that: Vec3Int): BoundingBox =
```
fm3, line 59: could you add a comment here explaining why ceilDiv was chosen?

fm3 (commented):
Backend LGTM :) I just added two small comments

fm3 (approved):

PR #6879, opened 2023-02-27
adds dedicated explore method for zarr datasets with a datasource-properties.json

normanrz:
- Added support for remote Zarr datasets with a `datasource-properties.json` as created by the webknossos-libs.

### Steps to test:
- Import `https://example.cloud.scm.io/l4_sample`

------

### DIFF
--- a/app/models/binary/explore/WebknossosZarrExplorer.scala
+++ b/app/models/binary/explore/WebknossosZarrExplorer.scala
@@ -20,9 +20,9 @@ class WebknossosZarrExplorer extends RemoteLayerExplorer {
       zarrLayers <- Fox.serialCombined(dataSource.dataLayers) {
         case l: ZarrSegmentationLayer =>
           for {
-            zarrLayers <- ngffExplorer.explore(remotePath.resolve(l.name), credentialId)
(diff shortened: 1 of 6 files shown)


app/models/binary/explore/WebknossosZarrExplorer.scala
```diff
@@ -0,0 +1,49 @@
...
+
+  override def explore(remotePath: Path, credentialId: Option[String]): Fox[List[(ZarrLayer, Vec3Double)]] =
+    for {
+      dataSourcePropertiesPath <- Fox.successful(remotePath.resolve("datasource-properties.json"))
+      dataSource <- parseJsonFromPath[DataSource](dataSourcePropertiesPath)
+      ngffExplorer = new NgffExplorer
+      zarrLayers <- Fox.serialCombined(dataSource.dataLayers) {
+        case l: ZarrSegmentationLayer =>
+          for {
+            zarrLayers <- ngffExplorer.explore(remotePath.resolve(l.name), credentialId)
```
fm3, line 23: I’d suggest renaming one of the zarrLayers variables. How about zarrLayersFromNgff for the inner or zarrLayersAdapted for the outer?

fm3 (approved):
Backend LGTM, I just added a small naming suggestion

PR #6882, opened 2023-02-28
When merging volume tracings, also merge segment lists

### LINKED ISSUE #6877: When merging volume annotations, also merge segment list
I merged two explorational annotations into one, using the merge dialog, with the checkbox »merge into active« _unchecked_. Merging skeleton and volume data worked as expected, but the segment list only showed the ids of one of the annotations.

fm3:
When merging volume annotations, their segment lists are now also merged. This needs information from the actual data merging (now passed as MergedVolumeStats) because the segment ids are changed during the merge process.

The PR also adresses a bug where initialDataMultiple failed, because it tried looking up the resolution list from uploaded zipfiles, which failed due to a match error on path regex, and then it tried to look up the resolution list remotely, which failed because postgres does at that time not yet have an entry for the currently created annotation.

Also slipped in here, sorry for the small unrelated change: GlobalAccessContext in JobsController to yield a readable error message when a user tries to run a globalizeFloodfill job on a dataset of a different organization.

- https://mergevolumesegmentlist.webknossos.xyz

### Steps to test:
- Create two volume annotations for the same dataset (without fallback layer is easier to test)
- Draw some segments
- From one of the annotations, merge the other in (menu→merge, select annotation id, UNCHECK the »merge into active« checkbox), result should have segment list from both annotations, ids should be uniqueified
- Download two volume annotations for the same dataset, reupload both together (drag’n’drop in dashboard)
- Result should show data, but no segment list, compare #6887

### Issues:
- fixes #6877

------

### DIFF
--- a/CHANGELOG.unreleased.md
+++ b/CHANGELOG.unreleased.md
@@ -16,6 +16,8 @@ For upgrade instructions, please check the [migration guide](MIGRATIONS.released
 ### Changed
 
 ### Fixed
+- Fixed a bug where merging multiple volume annotations would result in inconsistent segment lists. [#6882](https://github.com/scalableminds/webknossos/pull/6882)
+- Fixed a bug where uploading multiple annotations with volume layers at once would fail. [#6882](https://github.com/scalableminds/webknossos/pull/6882)
 
 ### Removed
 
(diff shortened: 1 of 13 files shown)


fm3:
@frcroth I realize that this touches a lot of code that you are not yet familiar with. If you feel lost in the review, maybe we could go through the code together :)

frcroth (approved):

PR #6890, opened 2023-03-02
Support n5 end-chunks with chunksize differing from metadata chunksize

### LINKED ISSUE #6888: Support reading N5 datasets where end-chunks have size smaller than that specified in metadata
N5 datasets can have end-chunks with sizes smaller than that specified in the n5 metadata.

Our chunk reader fails with an assertion in this case. It should instead be padded with zeros, like we do for neuroglancer precomputed datasets.

Compare https://scm.slack.com/archives/C02H5T8Q08P/p1677744181121229 for a concrete reproduction example.

fm3:
Turns out N5 chunks report their own size, which may differ from the chunk size specified in metadata for end-chunks (like, at the edges of the dataset). This PR introduces code into the wk chunk reading process that handles that.

First I tried reusing the chunkSizeAtIndex from our Neuroglancer Precomputed code, but it turns out that for N5 the bounding box is not a reliable source, chunks may or may not be truncated, so the chunk size mentioned in each chunk header should be used as the source of truth here.

Additionally, I renamed the TypedChunkReader to ChunkTyper, because its job is not actually reading anything, but converting a previously read byte array to a typed multi array.

- https://nendchunks.webknossos.xyz

### Steps to test:
- Open an N5-streamed dataset, e.g. via the datasource-properties.json found at https://scm.slack.com/archives/C02H5T8Q08P/p1677765260865059?thread_ts=1677744181.121229&cid=C02H5T8Q08P
- Data should load fine (note: may be pretty slow, note2: is very dark, see histogram, use brightness sliders and gamma correction)
- Zoom in to mag 1 and check dataset borders as well.

### Issues:
- fixes #6888

------

### DIFF
--- a/CHANGELOG.unreleased.md
+++ b/CHANGELOG.unreleased.md
@@ -16,6 +16,7 @@ For upgrade instructions, please check the [migration guide](MIGRATIONS.released
 ### Changed
 
 ### Fixed
+- Fixed a bug where N5 datasets reading with end-chunks that have a chunk size differing from the metadata-supplied chunk size would fail for some areas. [#6890](https://github.com/scalableminds/webknossos/pull/6890)
 
 ### Removed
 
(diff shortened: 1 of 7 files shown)


frcroth (approved):
LGTM

PR #6894, opened 2023-03-02
Avoid spinner when switching tabs in dashboard

philippotto:
I noticed that there was an incorrectly spinner every time one switches the active tab in the dashboard (which replaced the entire table).

- https://___.webknossos.xyz

### Steps to test:
- open dashboard and switch tabs --> table should **not** be replaced by a spinner for a brief moment

### Issues:
- follow-up to #6509

### DIFF
--- a/frontend/javascripts/admin/admin_rest_api.ts
+++ b/frontend/javascripts/admin/admin_rest_api.ts
@@ -2053,6 +2053,8 @@ export async function getPricingPlanStatus(): Promise<APIPricingPlanStatus> {
   return Request.receiveJSON("/api/pricing/status");
 }
 
+export const cachedGetPricingPlanStatus = _.memoize(getPricingPlanStatus);
+
 // ### BuildInfo webknossos
 export function getBuildInfo(): Promise<APIBuildInfo> {
   return Request.receiveJSON("/api/buildinfo", {
--- a/frontend/stylesheets/_utils.less
+++ b/frontend/stylesheets/_utils.less
@@ -122,3 +122,7 @@ td.nowrap * {
 .min-height-0 {
   min-height: 0;
 }
+
+.text-center {
+  text-align: center;
+}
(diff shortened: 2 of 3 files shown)


frontend/stylesheets/_utils.less
```diff
@@ -122,3 +122,7 @@ td.nowrap * {
 .min-height-0 {
   min-height: 0;
 }
+
+.text-center {
```
hotzenklotz, line 126: This is not used in the PR. Do you still want to merge it?

philippotto: I noticed that the spinner in question was centered by using a `text-center` class which didn't exist (which didn't work obviously). That's why, I added this class.

frontend/javascripts/admin/admin_rest_api.ts
```diff
@@ -2053,6 +2053,8 @@ export async function getPricingPlanStatus(): Promise<APIPricingPlanStatus> {
   return Request.receiveJSON("/api/pricing/status");
 }
 
+export const cachedGetPricingPlanStatus = _.memoize(getPricingPlanStatus);
```
hotzenklotz, line 2056: Caching is not without its own problems. The status might update (e.g. new people will join, more datasets will be uploaded taking up space etc...) 
How does this cache live? Until the next page refresh, right? So, the above points might not be a such a big issue after all. 

The approach is fine by me.

hotzenklotz (approved):
LGTM.

frontend/stylesheets/_utils.less
hotzenklotz: Ah. Good point. Maybe a Bootstrap leftover: https://getbootstrap.com/docs/5.0/utilities/text/#text-alignment

frontend/javascripts/admin/admin_rest_api.ts
philippotto: > Until the next page refresh, right?

Yes, exactly. When visiting the orga page, the non-cached version is used. So, only the dashboard is affected. If outdated data is noticed, the user will likely hit refresh to fix it.

PR #6899, opened 2023-03-06
WIP: New file system abstraction

frcroth:
- https://___.webknossos.xyz

### Steps to test:
- abc

### Issues:
- fixes #

------

### DIFF
--- a/webknossos-datastore/app/com/scalableminds/webknossos/datastore/storage/FileSystemsHolder.scala
+++ b/webknossos-datastore/app/com/scalableminds/webknossos/datastore/storage/FileSystemsHolder.scala
@@ -5,6 +5,13 @@ import com.google.cloud.storage.StorageOptions
 import com.google.cloud.storage.contrib.nio.{CloudStorageConfiguration, CloudStorageFileSystem}
 import com.scalableminds.util.cache.AlfuFoxCache
 import com.scalableminds.util.tools.Fox
+import com.scalableminds.webknossos.datastore.remotefilesystem.{
+  GoogleCloudRemoteFileSystem,
+  HttpsRemoteFileSystem,
+  RemoteFileSystem,
+  RemotePath,
+  S3RemoteFileSystem
+}
 import com.scalableminds.webknossos.datastore.s3fs.{S3FileSystem, S3FileSystemProvider}
 import com.scalableminds.webknossos.datastore.storage.httpsfilesystem.HttpsFileSystem
 import com.typesafe.scalalogging.LazyLogging
@@ -25,11 +32,18 @@ object FileSystemsHolder extends LazyLogging {
   private val fileSystemsCache: AlfuFoxCache[RemoteSourceDescriptor, FileSystem] =
     AlfuFoxCache(maxEntries = 100)
 
+  private val remotePathCache: AlfuFoxCache[RemoteSourceDescriptor, RemotePath] =
+    AlfuFoxCache(maxEntries = 100)
+
   def isSupportedRemoteScheme(uriScheme: String): Boolean =
     List(schemeS3, schemeHttps, schemeHttp, schemeGS).contains(uriScheme)
 
   def getOrCreate(remoteSource: RemoteSourceDescriptor)(implicit ec: ExecutionContext): Fox[FileSystem] =
     fileSystemsCache.getOrLoad(remoteSource, create)
+
+  def getOrCreateRemote(remoteSourceDescriptor: RemoteSourceDescriptor)(
+      implicit ec: ExecutionContext): Fox[RemotePath] =
+    remotePathCache.getOrLoad(remoteSourceDescriptor, createRemote)
 
   def create(remoteSource: RemoteSourceDescriptor)(implicit ec: ExecutionContext): Fox[FileSystem] = {
     val scheme = remoteSource.uri.getScheme
@@ -40,6 +54,28 @@ object FileSystemsHolder extends LazyLogging {
         getAmazonS3FileSystem(remoteSource)
       } else if (scheme == schemeHttps || scheme == schemeHttp) {
         getHttpsFileSystem(remoteSource)
+      } else {
+        throw new Exception(s"Unknown file system scheme $scheme")
+      }
+      logger.info(s"Successfully created file system for ${remoteSource.uri.toString}")
+      Fox.successful(fs)
+    } catch {
+      case e: Exception =>
+        val msg = s"get file system errored for ${remoteSource.uri.toString}:"
+        logger.error(msg, e)
+        Fox.failure(msg, Full(e))
+    }
+  }
+
+  def createRemote(remoteSource: RemoteSourceDescriptor)(implicit ec: ExecutionContext): Fox[RemotePath] = {
+    val scheme = remoteSource.uri.getScheme
+    try {
+      val fs: RemotePath = if (scheme == schemeGS) {
+        GoogleCloudRemoteFileSystem.create(remoteSource)
+      } else if (scheme == schemeS3) {
+        S3RemoteFileSystem.create(remoteSource)
+      } else if (scheme == schemeHttps || scheme == schemeHttp) {
+        HttpsRemoteFileSystem.create(remoteSource)
       } else {
         throw new Exception(s"Unknown file system scheme $scheme")
       }
--- a/app/models/binary/explore/ExploreRemoteLayerService.scala
+++ b/app/models/binary/explore/ExploreRemoteLayerService.scala
@@ -163,8 +163,7 @@ class ExploreRemoteLayerService @Inject()(credentialService: CredentialService) 
                                                             requestingUser._organization)
       remoteSource = RemoteSourceDescriptor(uri, credentialOpt)
       credentialId <- Fox.runOptional(credentialOpt)(c => credentialService.insertOne(c)) ?~> "remoteFileSystem.credential.insert.failed"
-      fileSystem <- FileSystemsHolder.getOrCreate(remoteSource) ?~> "remoteFileSystem.setup.failed"
-      remotePath <- tryo(fileSystem.getPath(FileSystemsHolder.pathFromUri(remoteSource.uri))) ?~> "remoteFileSystem.getPath.failed"
+      remotePath <- FileSystemsHolder.getOrCreateRemote(remoteSource) ?~> "remoteFileSystem.setup.failed"
       layersWithVoxelSizes <- exploreRemoteLayersForRemotePath(
         remotePath,
         credentialId.map(_.toString),
--- a/app/models/binary/explore/RemoteLayerExplorer.scala
+++ b/app/models/binary/explore/RemoteLayerExplorer.scala
@@ -5,6 +5,7 @@ import com.scalableminds.util.io.ZipIO
 import com.scalableminds.util.tools.{Fox, FoxImplicits, JsonHelper}
 import com.scalableminds.webknossos.datastore.dataformats.MagLocator
 import com.scalableminds.webknossos.datastore.models.datasource.{DataLayer, ElementClass}
+import com.scalableminds.webknossos.datastore.remotefilesystem.RemotePath
 import net.liftweb.util.Helpers.tryo
 import play.api.libs.json.Reads
 
@@ -25,7 +26,10 @@ trait RemoteLayerExplorer extends FoxImplicits {
 
   protected def parseJsonFromPath[T: Reads](path: Path): Fox[T] =
     for {
-      fileBytes <- tryo(ZipIO.tryGunzip(Files.readAllBytes(path))) ?~> "dataSet.explore.failed.readFile"
+      fileBytes <- path match {
+        case path: RemotePath => tryo(ZipIO.tryGunzip(path.get(""))) ?~> "dataSet.explore.failed.readFile"
+        case _                => tryo(ZipIO.tryGunzip(Files.readAllBytes(path))) ?~> "dataSet.explore.failed.readFile"
+      }
       fileAsString <- tryo(new String(fileBytes, StandardCharsets.UTF_8)).toFox ?~> "dataSet.explore.failed.readFile"
       parsed <- JsonHelper.parseAndValidateJson[T](fileAsString)
     } yield parsed
--- a/webknossos-datastore/app/com/scalableminds/webknossos/datastore/datareaders/FileSystemStore.scala
+++ b/webknossos-datastore/app/com/scalableminds/webknossos/datastore/datareaders/FileSystemStore.scala
@@ -1,6 +1,7 @@
 package com.scalableminds.webknossos.datastore.datareaders
 
 import com.scalableminds.util.io.ZipIO
+import com.scalableminds.webknossos.datastore.remotefilesystem.RemotePath
 import com.scalableminds.webknossos.datastore.storage.httpsfilesystem.ByteChannelOptions
 import net.liftweb.common.{EmptyBox, Full}
 import net.liftweb.util.Helpers.tryo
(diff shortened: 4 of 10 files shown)


webknossos-datastore/app/com/scalableminds/webknossos/datastore/storage/FileSystemsHolder.scala
```diff
@@ -22,24 +15,24 @@ object FileSystemsHolder extends LazyLogging {
   val schemeHttp: String = "http"
   val schemeGS: String = "gs"
 
-  private val fileSystemsCache: AlfuFoxCache[RemoteSourceDescriptor, FileSystem] =
+  private val vaultCache: AlfuFoxCache[RemoteSourceDescriptor, VaultPath] =
```
fm3, line 18: I find it a bit weird to cache VaultPaths here, when semantically what we want to cache is the Vaults themselves. How do you feel about this?
Maybe it makes more sense to have DataVault.create return a DataVault, not a path. That can be cached here.

The public method here could still return a path. Maybe something like

```
getVaultPath(remoteSourceDescriptor: RemoteSourceDescriptor): Fox[VaultPath] =
for {
valut <- vaultCache.getOrLoad(remoteSourceDescriptor, create)
} yield (new VaultPath(remoteSourceDescriptor.uri, vault)
```
if that makes sense?

webknossos-datastore/app/com/scalableminds/webknossos/datastore/storage/FileSystemsHolder.scala
```diff
@@ -1,18 +1,11 @@
...
+import com.scalableminds.webknossos.datastore.datavault.{GoogleCloudDataVault, HttpsDataVault, VaultPath, S3DataVault}
 import com.typesafe.scalalogging.LazyLogging
 import net.liftweb.common.Full
 
-import java.io.ByteArrayInputStream
-import java.net.URI
-import java.nio.file.FileSystem
 import scala.concurrent.ExecutionContext
 
 object FileSystemsHolder extends LazyLogging {
```
fm3, line 11:
```suggestion
object DataVaultsHolder extends LazyLogging {
```

webknossos-datastore/app/com/scalableminds/webknossos/datastore/datavault/VaultPath.scala
```diff
@@ -0,0 +1,76 @@
...
+import java.net.URI
+import java.nio.file.{FileSystem, LinkOption, Path, Paths, WatchEvent, WatchKey, WatchService}
+
+/*
+VaultPath implements Path so that a drop in replacement is possible while continuing to use Paths for local storage.
+This class does not implement all relevant methods and it might be a good idea to remove the inheritance on Path in the
+future.
+ */
+
+class VaultPath(uri: URI, dataVault: DataVault, fileSystemCredentialOpt: Option[FileSystemCredential]) extends Path {
```
fm3, line 16: Isn’t the fileSystemCredentialOpt already stored in the DataVault? If so, I’d not store it again here, but rather provide an accessor if needed, that forwards this to dataVault.fileSystemCredentialOpt

webknossos-datastore/app/com/scalableminds/webknossos/datastore/datavault/NullDataVault.scala
```diff
@@ -0,0 +1,8 @@
+package com.scalableminds.webknossos.datastore.datavault
+
+class NullDataVault extends DataVault {
```
fm3, line 3: I see that this is used for the FileSystemVaultPath – would it make sense to name it FileSystemDataVault? If so, the `get` implementation using `Files.newByteChannel` could also live here

app/models/binary/explore/PrecomputedExplorer.scala
```diff
@@ -19,6 +19,7 @@ class PrecomputedExplorer extends RemoteLayerExplorer {
 
   override def explore(remotePath: Path, credentialId: Option[String]): Fox[List[(PrecomputedLayer, Vec3Double)]] =
     for {
+      _ <- Fox.bool2Fox(remotePath.toUri.toString.last == '/') ?~> "Remote dataset explore path must be a directory (indicated by trailing slash)"
```
fm3, line 22: Would it be possible to automatically add a trailing slash in this case so the user does not get this error?

webknossos-datastore/app/com/scalableminds/webknossos/datastore/datavault/VaultPath.scala
```diff
@@ -0,0 +1,76 @@
...
+
+  def get(key: String, range: Option[Range] = None): Array[Byte] = dataVault.get(key, this, range)
```
fm3, line 18: I’m a bit confused about the `key` – do we need this distinction between the Path’s uri and the key? I would have assumed that the path _is_ the key.
then path.get(key) would instead be (path / key).get()
Are there problems with that?

fm3: Also, I’d suggest naming the function `readBytes` sinde `get` is very overloaded in the scala world already. (I assume that you took the name from the GET concept in http, which does make sense, but I think it’s a bit confusing here)

webknossos-datastore/app/com/scalableminds/webknossos/datastore/datavault/VaultPath.scala
```diff
@@ -0,0 +1,76 @@
...
+
+  def tryGet(key: String, range: Option[Range] = None): Option[Array[Byte]] =
```
fm3, line 20: Is there any use case where we want to use get and not tryGet? If not, let’s just make tryGet the default and the other private.

Long-term, my guess is that we want to use Foxes for all of the error handling, but that does not have to happen now. I wrote follow-up issue https://github.com/scalableminds/webknossos/issues/6906

webknossos-datastore/app/com/scalableminds/webknossos/datastore/datavault/HttpsDataVault.scala
```diff
@@ -0,0 +1,104 @@
...
+
+import java.net.URI
+import scala.concurrent.duration.DurationInt
+import sttp.client3._
+import sttp.model.Uri
+
+class HttpsDataVault(credential: Option[FileSystemCredential]) extends DataVault {
+
+  private val connectionTimeout = 5 seconds
+  private val readTimeout = 1 minute
```
fm3, line 18: let’s go higher, I’d say 1 minute and 10 minutes respectively. You wouldn’t believe how slowly some servers answer :D

webknossos-datastore/app/com/scalableminds/webknossos/datastore/datavault/HttpsDataVault.scala
```diff
@@ -0,0 +1,104 @@
...
+  private lazy val backend = HttpClientSyncBackend(options = SttpBackendOptions.connectionTimeout(connectionTimeout))
+
+  def getBasicAuthCredential: Option[HttpBasicAuthCredential] =
+    credential match {
+      case Some(c) => {
```
fm3, line 23:
```suggestion
    credential.map{ c =>
```

webknossos-datastore/app/com/scalableminds/webknossos/datastore/datavault/HttpsDataVault.scala
```diff
@@ -0,0 +1,104 @@
...
+
+  private val headerInfoCache = scala.collection.mutable.Map[URI, (Boolean, Int)]()
+
+  private def getHeaderInformation(uri: URI) =
+    headerInfoCache.get(uri) match {
+      case Some((b, i)) => (b, i)
+      case None => {
+        val response: Identity[Response[Either[String, String]]] = backend.send(headRequest(uri))
+        val acceptsPartialRequests = response.headers.find(_.is("Accept-Ranges")).exists(_.value == "bytes")
+        val dataSize = response.headers.find(_.is("Content-Length")).map(_.value.toInt).getOrElse(0)
```
fm3, line 52:
```suggestion
        val dataSize = response.headers.find(_.is("Content-Length")).map(_.value.toLong).getOrElse(0L)
```
I guess files on the server may be bigger than 4GB.

fm3: Hmm, this means we should also probably take `NumericRange[Long]` instead of the Int-specific `Range` in the Vault signatures, so that we can read chunks that are beyond the 4GB limit.

webknossos-datastore/app/com/scalableminds/webknossos/datastore/datavault/S3Utilities.scala
```diff
@@ -0,0 +1,83 @@
...
+
+class S3Utilities {}
+
+class AnonymousAWSCredentialsProvider extends AWSCredentialsProvider {
+  override def getCredentials = new AnonymousAWSCredentials
+
+  override def refresh(): Unit = {}
+}
+
+object S3Utilities {
```
fm3, line 23: I think both classes are small enough that it would be acceptable to move this stuff into S3DataVault directly (and make it private there), so that we don’t have to import all this there.

webknossos-datastore/app/com/scalableminds/webknossos/datastore/datavault/S3Utilities.scala
```diff
@@ -0,0 +1,83 @@
...
+    }
+
+  def getAWSCredentials(props: Properties) =
+    new BasicAWSCredentials(props.getProperty(ACCESS_KEY), props.getProperty(SECRET_KEY))
+
+  def getCredentialsProvider(props: Properties): AWSCredentialsProvider = {
+    if (props.getProperty(ACCESS_KEY) == null && props.getProperty(SECRET_KEY) == null)
+      return new AnonymousAWSCredentialsProvider
+    new AWSStaticCredentialsProvider(getAWSCredentials(props))
+  }
```
fm3, line 74: Could we get rid of the Properties passing and instead build the BasicAWSCredentials directly from our S3AccessKeyCredential?

fm3 (changes requested):
Amazing stuff, props for getting through this so quickly, and even including tests :) Very much enjoying the diff line count as well :laughing: This already feels a lot more designed and less patched together, compared to using bits and pieces of the NIO providers.

I added some suggestions. Most are fairly small, but one thing I think is interesting: whether or not to pass an additional `key` to the actual read calls, or whether the key can itself be mapped to the VaultPath instances. Maybe you could read through my coments and have a look at that. I’m open to discussing this, it is well possible that I missed some cases that make the current version necessary.

webknossos-datastore/app/com/scalableminds/webknossos/datastore/datavault/HttpsDataVault.scala
frcroth: These were the values from the original httpsfilesystem (https://github.com/scalableminds/webknossos/blob/6b5143a94e657c624370d0de85e36d6e64eeb97d/webknossos-datastore/app/com/scalableminds/webknossos/datastore/storage/httpsfilesystem/HttpsSeekableByteChannel.scala) but sure can be changed

webknossos-datastore/app/com/scalableminds/webknossos/datastore/datavault/HttpsDataVault.scala
frcroth: Doesn't work for me, since the inner match may return none which is not possible in the map which expects a full result.

fm3: maybe flatMap, then? 🤔 Feel free to ignore if that is also not a good fit

frcroth: Yes flatMap it is!

webknossos-datastore/app/com/scalableminds/webknossos/datastore/datavault/NullDataVault.scala
frcroth: Actually that does not work that well, since the readBytes implementation in FileSystemDataVault would still need a real nio Path, but the interface provides a VaultPath, which does not allow it to call Files.readAllBytes(path) because of the missing filesystem

fm3: I see, then let’s consider this again later. I think, eventually there should be no more nio Paths used in the same places as VaultPaths. I created the follow-up issue https://github.com/scalableminds/webknossos/issues/6929

app/models/binary/explore/PrecomputedExplorer.scala
fm3: If this is specific to the PrecomputedExplorer, I’d say it should happen right here, something like 
`remotePathNormalized <- addTrailingSlashIfMissing(remotePath)` – what do you think?

test/backend/DataVaultTestSuite.scala
```diff
@@ -0,0 +1,68 @@
...
+          val uri = new URI("gs://neuroglancer-fafb-data/fafb_v14/fafb_v14_orig")
+          val vaultPath = new VaultPath(uri, GoogleCloudDataVault.create(RemoteSourceDescriptor(uri, None)))
+          val bytes = (vaultPath / s"fafb_v14_orig/$dataKey").readBytes().get
+
+          assert(bytes.length == dataLength)
+          assert(bytes.take(10).sameElements(Array(-1, -40, -1, -32, 0, 16, 74, 70, 73, 70)))
+        }
+      }
+    }
+  }
```
fm3, line 67: Could you add a few tests on the behavior of `vaultPath./`? I’m not certain what uri.resolve does in different cases, especially concerning slashes. What happens on
 - existing path has trailing slash vs no trailing slash
 - param has starting slash vs no trailing slash
 - the combinations thereof
 - param has no trailing slash
 - can consecutive slashes be produced? do we want them?
 - can a path be accidentally constructed that has no slash at a spot where the user would have expected it? path / "hi" / "there"?
 - what happens if you pass ".." or "."?

test/backend/DataVaultTestSuite.scala
```diff
@@ -0,0 +1,68 @@
...
+          assert(bytes.length == dataLength)
+          assert(bytes.take(10).sameElements(Array(-1, -40, -1, -32, 0, 16, 74, 70, 73, 70)))
+        }
+      }
+
+      "with Google Cloud Storage Vault" should {
+        "return correct response" in {
+          val uri = new URI("gs://neuroglancer-fafb-data/fafb_v14/fafb_v14_orig")
+          val vaultPath = new VaultPath(uri, GoogleCloudDataVault.create(RemoteSourceDescriptor(uri, None)))
+          val bytes = (vaultPath / s"fafb_v14_orig/$dataKey").readBytes().get
```
fm3, line 60:
```suggestion
          val bytes = (vaultPath / "fafb_v14_orig" / dataKey).readBytes().get
```

fm3 (commented):
Very nice cleanup! We are getting very close :)

test/backend/DataVaultTestSuite.scala
frcroth: Actually, until now the slash only did resolve, but yes, it makes way more sense if the slash adds a slash. Will rework and then write tests.

fm3: I’d say automatically remove the starting slash of the parameter

app/models/binary/explore/PrecomputedExplorer.scala
frcroth: No longer needed since "/" now adds a trailing slash if necessary.

PR #6903, opened 2023-03-06
Store editable mappings in multiple fossildb columns+keys

### LINKED ISSUE #6797: Measure + Improve server-side performance for supervoxel proofreading
a.k.a. editable mappings.

Also compare #6288

Ideas:
  - [ ] Optimize Roundtrip?
  - [ ] Use low-level typed data processing (no UnsignedInteger)
  - [ ] Is the size limit for agglomerate skeletons too tight?
  - [ ] save editable mapping in distributed keys in fossildb (agglomerateToGraph)

fm3:
- https://editablemappingsdistributed.webknossos.xyz

### TODO
 - [x] on multiple update actions, use temporary store, flush to fossil on commit
 - [x] migration
 - [x] problem: newest version may not be materialized, do we apply update actions in migration? or should the code be able to apply lazily?
 - [x] test perf
 - [x] how to apply migration in production? how much downtime is needed?
 - [x] two-phase migration to lower downtime?
 - [x] include relevant changes from #6829
 - [x] finalize migration (args, camel_case)
 - [ ] Reproduce philipps failed mesh requests → cannot reproduce, let’s defer.
 - [x] Front-end should not send fallback layer requests if volume tracing bucket requests fail
 - [x] should action with segment id 0 also become no-op? yes

### Steps to test:
- Create annotation with volume layer with segmentation fallback (that has an agglomerate file. on the dev instance I suggest `l4dense_motta_et_al_dev_v2`)
- Select an agglomerate mapping and use the proofreading tool to split and merge some cells
- refresh the page, should still be look right
- To test the migration locally, repeat above steps on master, then switch to this branch, start the fossildb (yarn start does it), then run the migration `python tools/migrate-editable-mappings/migrate-editable-mappings.py -vw localhost:7155` , then reload the same annotation and see that everything is there and still works as expected

### Issues:
- fixes #6797

------


MIGRATIONS.unreleased.md
```diff
@@ -9,5 +9,7 @@ User-facing changes are documented in the [changelog](CHANGELOG.released.md).
 [Commits](https://github.com/scalableminds/webknossos/compare/23.04.1...HEAD)
 
  - The config key features.isDemoInstance was renamed to features.isWkorgInstance (only needs to be adapted for the main wkorg instance). [#6941](https://github.com/scalableminds/webknossos/pull/6941/files)
+ - For instances with existing editable mapping (a.k.a supervoxel proofreading) annotations: To keep those annotations alive, a python migration has to be run with access to your tracingstore’s FossilDB. It is recommended to do this during a webknossos downtime to avoid data loss. It needs python 3.8+ and the pip packages installable by `pip install grpcio-tools grpcio-health-checking`. Run it with `python tools/migrate-editable-mappings/migrate-editable-mappings.py -v -w -o localhost:7155`. Omit -o for a faster migration but no access to older versions of the editable mappings. The migration is idempotent.
+ - Note that FossilDB needs to be opened with new additional column families editableMappingsInfo, editableMappingsAgglomerateToGraph, editableMappingsSegmentToAgglomerate.
```
philippotto, line 13: I think "Note that" can be omitted for brevity. Also, I would swap this line with the previous one, as it seems non-optional whereas the other point is somewhat "optional".

tools/migrate-editable-mappings/migrate-editable-mappings.py
```diff
@@ -1,5 +1,28 @@
 #!/usr/bin/env python3
 
+# Migrating Editable Mappings a.k.a. proofreading annotations
+# 2023-04 https://github.com/scalableminds/webknossos/pull/6903
+#
+## Required python packages:
+# pip install grpcio-tools grpcio-health-checking
+#
+## Development
+# Note: needs both old and new format editable mapping proto definitions
```
philippotto, line 10: these are already checked in, right? might be worth mentioning this again here.

fm3:
Hmm, I’m not sure about that. The chunk data request did not change in this PR. However, the chunk list request may have changed (since it has to list all chunks of the segments currently mapped to that particular agglomerate id) so there may be invalid byte offsets there? However, I don’t see how that would not lead in a 4xx or 5xx but this strange behavior of your screenshot :thinking: I could not reproduce this behavior.

fm3:
This was also my strategy locally. In production, we would have to continue opening the full list of column families. This already works for the dev deployment fossildb (which is shared by all dev deployments)

fm3:
> I think I found an unrelated problem, where I cannot perform a particular split thinking I’ll have a look at that

It turned out to be caused by mag1 not being present. The agglomerate file had a node very close to the edge of the agglomerate. The voxel there was eaten by downsampling (the other segment id won at that position). That meant that looking up the segment id by the position from the agglomerate file lead to the wrong segment id, essentially making the split a no-op since the edge was not actually present in the agglomerate. I’d say that this is acceptable since it only happens in case of a mismatch between the present segmentation data and the agglomerate file. I fear there is no good way to guard against this. I added some warning logging in case an edge is not present and a split becomes a no-op.

fm3:
@philippotto do you think we can go ahead with this? Can you still reproduce the issue with mesh loading? Maybe you could show me in person?

philippotto:
> @philippotto do you think we can go ahead with this? Can you still reproduce the issue with mesh loading? Maybe you could show me in person?

As discussed in person, I think we can go ahead with this. I also fixed the incorrect fallback fetching case.

philippotto (approved):
:100:

PR #6910, opened 2023-03-09
[WIP] Add wk logo to screenshots

MichaelBuessemeyer:
- https://___.webknossos.xyz

### Steps to test:
- abc

### Issues:
- fixes #

------


CHANGELOG.unreleased.md
```diff
@@ -11,6 +11,7 @@ For upgrade instructions, please check the [migration guide](MIGRATIONS.released
 [Commits](https://github.com/scalableminds/webknossos/compare/23.03.1...HEAD)
 
 ### Added
+- WEBKNOSSOS screenshots now include a logo in the lower left corner. [#6910](https://github.com/scalableminds/webknossos/pull/6910)
```
daniel-wer, line 14: If you add the possibility to disable the logo, please add a sentence here on who can disable it.

frontend/javascripts/oxalis/view/rendering_utils.ts
```diff
@@ -106,6 +117,26 @@ export async function downloadScreenshot() {
...
     const inputCatcherElement = document.querySelector(`#inputcatcher_${planeId}`);
+    const scalebar = inputCatcherElement?.querySelector(".scalebar") as HTMLElement;
+    let drawImageIntoCanvasCallback = null;
+    if (scalebar) {
+      drawImageIntoCanvasCallback = (ctx: CanvasRenderingContext2D) => {
+        const scalebarDistanceToRightBorder =
+          scalebar.offsetParent != null
+            ? scalebar.offsetParent.clientWidth - scalebar.offsetLeft - scalebar.clientWidth
+            : 0;
+        const scalebarDistanceToTopBorder = scalebar.offsetTop;
```
daniel-wer, line 128: Nitpicky, but I was wondering why the top of the logo and scale bar don't exactly align and I think it's caused by the following: `offsetTop` is relative to the parent's [content-box](https://developer.mozilla.org/en-US/docs/Web/CSS/box-sizing#content-box), but the inputcatchers use [box-sizing: border-box](https://developer.mozilla.org/en-US/docs/Web/CSS/box-sizing#border-box) with a 2 px wide border (`OUTER_CSS_BORDER`). These 2 px need to be added here, to be exact :)

frontend/javascripts/oxalis/view/rendering_utils.ts
```diff
@@ -106,6 +117,26 @@ export async function downloadScreenshot() {
     const buffer = renderToTexture(planeId, null, null, false, clearColor);
 
     const inputCatcherElement = document.querySelector(`#inputcatcher_${planeId}`);
+    const scalebar = inputCatcherElement?.querySelector(".scalebar") as HTMLElement;
+    let drawImageIntoCanvasCallback = null;
+    if (scalebar) {
```
daniel-wer, line 122: Since the scale bar and logo need to be toggled independently, I would propose to change the `right` and `bottom` position of the scale bar from the relative 1% to an absolute value like 10 px or so and add that to the constants. Then, the scalebar is no longer needed, here, and the position calculations are more straightforward.

daniel-wer (commented):
Cool stuff, please see my suggestions :)

I'll have another look once you implemented the logo toggle as discussed, internally.

frontend/javascripts/messages.tsx
```diff
@@ -45,6 +45,7 @@ export const settings: Partial<Record<keyof RecommendedConfiguration, string>> =
   gpuMemoryFactor: "Hardware Utilization",
   overwriteMode: "Volume Annotation Overwrite Mode",
   useLegacyBindings: "Classic Controls",
+  renderWatermark: "Include Watermark",
```
daniel-wer, line 48: I'm not entirely sure about the naming/description here. I would propose "Watermark on Screenshots" to make it clear that the watermark refers to screenshots. Also, watermark might have a slightly negative connotation so we could also refer to it as "Logo on Screenshots" but I might be overthinking it :thinking: @philippotto any opinion?

For reference, this is what it currently looks like (with the description being "Show a watermark in the lower-left corner in each screenshot.")

image

frontend/javascripts/messages.tsx
```diff
@@ -72,6 +73,7 @@ export const settingsTooltips: Partial<Record<keyof RecommendedConfiguration, st
   mouseRotateValue: "Rotation speed when using the mouse to drag the rotation.",
   zoom: "Zoom in or out in the data viewports",
   displayScalebars: "Show a scale in the lower-right corner of each viewport",
+  renderWatermark: "Show a watermark in the lower-left corner in each screenshot.",
```
daniel-wer, line 76:
```suggestion
  renderWatermark: "Show a watermark in the lower-left corner of each screenshot.",
```

frontend/javascripts/oxalis/view/scalebar.tsx
```diff
@@ -74,8 +74,9 @@ function Scalebar({ zoomValue, dataset, viewportWidthInPixels, viewportHeightInP
       <div
         style={{
           position: "absolute",
-          bottom: "1%",
-          right: "1%",
+          // TODO use this constant in the rendering utils and fix the offset of the wk watermark logi
```
daniel-wer, line 77:
```suggestion
```

frontend/javascripts/test/fixtures/dummy_organization.ts
```diff
@@ -0,0 +1,19 @@
...
+import { PricingPlanEnum } from "admin/organization/pricing_plan_utils";
+
+const dummyOrga: APIOrganization = {
+  id: "dummy_orga_id",
+  name: "dummy_orga",
+  additionalInformation: "more information",
+  displayName: "Dummy Orga",
+  pricingPlan: PricingPlanEnum.Basic,
+  enableAutoVerify: true,
+  newUserMailingList: "dummy@orga.com",
```
daniel-wer, line 11:
```suggestion
  newUserMailingList: "dummy@example.com",
```

I know it's not used but this is [best practice](https://www.iana.org/domains/reserved).

daniel-wer (approved):
Well done, worked like a charm during testing :+1: 

Please see my suggestions, nothing major.

frontend/javascripts/messages.tsx
philippotto: Yes, I would change "watermark" to "logo", too. In the tooltip, it could even say "WEBKNOSSOS logo".

>  so we could also refer to it as "Logo on Screenshots"

Maybe "Logo **in** Screenshots"?

fm3: I am also in favor of saying “logo” instead of “watermark” :)

frontend/javascripts/test/fixtures/dummy_organization.ts
MichaelBuessemeyer: ah, good to know, thanks :pray:

PR #6914, opened 2023-03-13
Fix race condition which could disable saving

philippotto:
This fixes a race condition which would affect saving when annotation-sharing was activated. Especially on slow computers, this could lead to a faulty initialization which would mean that nothing was saved.

- https://fixsavingracecondition.webknossos.xyz

### Steps to test:
- the PR adds a sleep to provoke the original bug (without the fix the bug was reliably happening)
- Create an annotation and enable others-may-edit
- reload the annotation and add a tree
- save and reload -> tree should be there

### Issues:
- follow-up to #6819 
- context: https://scm.slack.com/archives/C02H5T8Q08P/p1678178937732359

------


daniel-wer (approved):
LGTM, I did some monkey testing and couldn't provoke any errors or unsaved data.

PR #6916, opened 2023-03-14
Fix styling for login pages for dark mode users

### LINKED ISSUE #6885: Login Screen has no input hints in dark mode
## Context
Light/Dark mode has an effect on the login/signup/pw reset pages. There were all designed & tested for light mode, the WEBKNOSSOS default. Yet, after logging out of dark mode, the theme persists and caused issues.

Bug introduced by PR #6844 

## Expected Behavior
Input text boxes should have visible text hints and icons in dark mode

## Current Behavior
Input text boxes have white text hints and icons on a white background. Hence the appear to be missing

## Steps to Reproduce the bug
1. Log in
2. Switch to Dark Mode
3. Log out
4. Log in screen has no icon and no input field hints

## Your Environment for bug

- Browser name and version: all
- Operating System and version: MacOS
- Version of WebKnossos (Release or Commit): Version: 21905

hotzenklotz:
This PR fixes some styling issues for dark mode users after logging out. The default theme for WEBKNOSSOS is light, especially when logged out. This fix, restores light mode on each log out to fix the UI. Theme switching in this case is not persisted as a user choice, though, in order to return to dark mode on the next login for dark mode users.

- https://___.webknossos.xyz

### Steps to test:
1. Log in 
2. Switch to dark mode
3. Log out.
4. Login page should look correct, i.e. have hints for each input field
5. Log in again. You should be in dark mode once more.

### Issues:
- fixes #6885 

------


philippotto:
> This PR fixes some styling issues for dark mode users after logging out.

In Firefox, I have the styling problem simply by opening the log in page (without having logged in before). See screenshot. I think, the problem is that the default theme is "auto" and my Firefox uses a dark theme which then propagates to wK using the dark theme.

Maybe it would be easier to give the entire log in view a css class à la "light-container" and then override the input color via `.light-container { input { color: black; } }` or something like that. With the current approach I would also suspect that logging out and directly logging in again could lose the theme setting (not tested, though).

image

philippotto (approved):
:+1:

PR #6920, opened 2023-03-15
Implement viewing sharded neuroglancer precomputed datasets

### LINKED ISSUE #6823: Support sharding for remote datasets
Both Neuroglancer Precomputed and Zarr (v3) can have sharded data formats. When served from a file server, range requests should be used to fetch only individual chunks out of the shards.

This needs to 
 - understand the sharding metadata of the datasets to figure out byte offsets and lengths of the chunks to request
 - adapt the NIO file systems to support such partial reads
 - adapt the data  reading code to _use_ those partial reads (both in local and remote case)

Possibly related: 
https://stackoverflow.com/questions/35745403/java-how-to-read-part-of-file-from-specified-position-of-bytes
My assumption is that in the NIO implementation this will use SeekableByteChannel, which is implemented by the different NIO File System implementations. The HttpsFileSystem currently reads the whole file before returning a SeekableByteChannel, needs to be adapted to use http range requests. I don’t know how the S3 and GCS File System adapters behave.

frcroth:
- https://sharding.webknossos.xyz

### Steps to test:
- Sharded datasets to explore and view:
	- gs://h01-release/data/20210601/4nm_raw
	- See [neuroglancer](https://neuroglancer-demo.appspot.com/#!%7B%22dimensions%22:%7B%22x%22:%5B4e-9%2C%22m%22%5D%2C%22y%22:%5B4e-9%2C%22m%22%5D%2C%22z%22:%5B3.3e-8%2C%22m%22%5D%7D%2C%22position%22:%5B499195.875%2C364427.59375%2C2646.5%5D%2C%22crossSectionScale%22:1339.4307643944185%2C%22projectionScale%22:1048576%2C%22layers%22:%5B%7B%22type%22:%22image%22%2C%22source%22:%22precomputed://gs://h01-release/data/20210601/4nm_raw%22%2C%22tab%22:%22source%22%2C%22name%22:%224nm_raw%22%7D%5D%2C%22selectedLayer%22:%7B%22visible%22:true%2C%22layer%22:%224nm_raw%22%7D%2C%22layout%22:%224panel%22%7D) for comparison

### Issues:
- fixes #6823 

------


webknossos-datastore/app/com/scalableminds/webknossos/datastore/datareaders/precomputed/PrecomputedArray.scala
```diff
@@ -69,4 +76,182 @@ class PrecomputedArray(relativePath: DatasetPath,
...
+    Range.Long(0, end, 1)
+  }
+
+  private def getShardIndex(shardPath: VaultPath)(implicit ec: ExecutionContext): Fox[Array[Byte]] =
+    shardIndexCache.getOrLoad(shardPath, readShardIndex)
+
+  private def readShardIndex(shardPath: VaultPath)(implicit ec: ExecutionContext): Fox[Array[Byte]] =
+    for {
+      bytes <- Fox.option2Fox(shardPath.readBytes(Some(shardIndexRange)))
+    } yield bytes
```
fm3, line 133:
```suggestion
    Fox.option2Fox(shardPath.readBytes(Some(shardIndexRange)))
```
should be identical

webknossos-datastore/app/com/scalableminds/webknossos/datastore/datareaders/DatasetArray.scala
```diff
@@ -82,21 +82,32 @@ class DatasetArray(relativePath: DatasetPath,
...
+        chunkData: Array[Byte] <- readShardedChunk(chunkIndex)
+        chunkShape = header.chunkSizeAtIndex(chunkIndex)
+        multiArray <- chunkReader.parseChunk(chunkData, chunkShape)
+      } yield multiArray
+    } else {
+      val chunkFilename = getChunkFilename(chunkIndex)
+      val chunkFilePath = relativePath.resolve(chunkFilename)
+      val storeKey = chunkFilePath.storeKey
+      val chunkShape = header.chunkSizeAtIndex(chunkIndex)
+      chunkReader.read(storeKey, chunkShape)
```
fm3, line 109: I’m wondering whether we can get the two branches here into one. As I understand, currently, both branches have their own version of reading from the store, decompressing, then typing. Could it be unified? Maybe the sharding implementation could just return the chunk path plus byte range to be passed to the existing chunkReader? (With non-sharding returning None for the range)

fm3 (commented):
Cool stuff! Nice dealing with the sharding metadata!
I added a question dealing with unifying the sharded vs non-sharded reading code, otherwise this already looks very good :)

fm3 (approved):
Works for me :) Thanks for addressing the feedback

PR #6838, opened 2023-02-10
Adds OME-TIFF export

normanrz:
- Adds OME-Tiff export
- Adds mag selection in the export modal
- Gets rid of the specialized bbox export modal (in favor of the regular download modal)
- The download modal polls for the export job status and immediately begins download, when complete
- Cleans up some job parameters in the backend 

### Steps to test:
- Checkout the `ome-tiff-export` branch of vx
- Start the worker
- Start wk
- Try to export an OME-TIFF
- Validate that the exported data is right with Fiji

------
- [x] Updated [(unreleased) changelog](../blob/master/CHANGELOG.unreleased.md#unreleased)
- [x] Updated [(unreleased) migration guide](../blob/master/MIGRATIONS.unreleased.md#unreleased) if applicable
- [x] Updated [documentation](../blob/master/docs) if applicable
- [x] Needs worker update after deployment
- [x] Ready for review

PR #6931, opened 2023-03-20
Fix download button for annotations when tiff export is disabled

philippotto:
- https://___.webknossos.xyz

### Steps to test:
- I created an annotation locally and tried to download it (non worker configured)
- without the change, the ok button was disabled.
- with the change, it works.

### Issues:
- follow-up for #6838 

------


fm3 (approved):
LGTM :)

Issue #4194, opened 2019-07-18
Blend modes for layers


philippotto:
For some use cases it would be helpful to have Photoshop-like blending modes for each layer. This would require that the order of the layers can be changed.

See [here](https://docs.gimp.org/en/gimp-concepts-layer-modes.html) for an overview of blending modes in GIMP including the mathematical formulars.

Things to take care of:
- [ ] the changes should be backwards compatible (existing datasets should render exactly as before, even though the order of the layers is taken into account <-- maybe make the layers "additive" by default?)
- [ ] use case 1: blend high res and low res data in a "sensible way"
- [ ] use case 2: allow to achieve something like an alpha channel (e.g., everything which is black should be invisible by using a "Screen" blending mode)

PR #6936, opened 2023-03-21
Add "merge" blend mode

MichaelBuessemeyer:
- https://coverblendmode.webknossos.xyz/

#### TODOs
- [x] check all layer types
  - [x] uint24 doesn't work :/
    - [x] patched it, but need to talk with @philippotto about this "fix". (why the other packing degrees have a default alpah value of `10.0` and but the uint24 case does not. 
  - [x] Layer with `packingDegree = 2` / `byteCount = 2` is still untested. This is not included the data layers dataset.
- [x] build blend mode switch in frontend
  - [x] discuss name of this blend mode. GIMP calls it "merge" as it merges images together where layer 1 has one part of the image and the other layer has the rest of the image. But I think we agreed on calling this blend mode "cover" as the lower layers cover the layers above in the layer list.

### Steps to test:
- Open a dataset with multiple layers (e.g. the data layers dataset) in view mode
- Make all layers visible -> the data should be pretty bright
- Got to the settings tab to the data rendering settings section and change the render mode to cover.
- Now only the last layer in the list should be rendered.
- Turn all layers off besides the uint24 layer. This layer should also be rendered correctly.
- Then play a little around with what layers are set to visible and what rendering mode you select.
- Please notice, that both float layers are slightly darker than the other layers. This is likely due to the layers / data of the layers themselves (as this behaviour is already present on the master).

### Issues:
- fixes [#4194](https://github.com/scalableminds/webknossos/issues/4194)

------


MichaelBuessemeyer:
As discussed, the code now decides whether a color is valid by `alpha < 0.0` and not `alpha <= 0.0`.

frontend/javascripts/oxalis/shaders/filtering.glsl.ts
```diff
@@ -115,11 +115,8 @@ export const getMaybeFilteredColorOrFallback: ShaderModule = {
     ) {
       vec4 color = getMaybeFilteredColor(layerIndex, d_texture_width, packingDegree, worldPositionUVW, suppressBilinearFiltering, supportsPrecomputedBucketAddress);
 
-      if (color.a < 0.0) {
-        // Render gray for not-yet-existing data
-        color = fallbackColor;
-      }
-
+      // if alpha === 0 -> No data
+      color = mix(color, fallbackColor, float(color.a < 0.0));
```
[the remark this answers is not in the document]
MichaelBuessemeyer: outdated

MichaelBuessemeyer:
> As discussed, the code now decides whether a color is valid by alpha < 0.0 and not alpha <= 0.0.

This is now also outdated. The cover blend mode now applies [Alpha_blending](https://en.wikipedia.org/wiki/Alpha_compositing#Alpha_blending). This automatically discards colors with an invalid alpha value.

MichaelBuessemeyer:
@philippotto Do you want to take over the pr review? Or should I ask daniel?

frontend/javascripts/oxalis/shaders/main_data_shaders.glsl.ts
```diff
@@ -222,14 +226,33 @@ void main() {
...
+        vec4 additive_color = data_color + layer_color;
+        // Cover blendMode == 0: Applying alpha blending to merge the layers where the top most layer has priority.
+        // See https://en.wikipedia.org/wiki/Alpha_compositing#Alpha_blending for details.
+        float mixed_alpha_factor = (1.0 - data_color.a) * layer_color.a;
+        float mixed_alpha = mixed_alpha_factor + data_color.a;
+        vec3 cover_color_rgb = data_color.a * data_color.rgb + mixed_alpha_factor * layer_color.rgb;
+        // Catching edge case where mixed_alpha is 0.0 and therefore the cover_color would have nan values.
+        float is_mixed_alpha_zero = float(mixed_alpha == 0.0);
+        vec4 cover_color = vec4(cover_color_rgb / (mixed_alpha + is_mixed_alpha_zero), mixed_alpha);
+        cover_color = mix(cover_color, vec4(0.0), is_mixed_alpha_zero);
```
MichaelBuessemeyer, line 243: @daniel-wer I diverged from the original approach with the if statement to guard against the division by zero. Instead I used a neat tick that was already used a few code lines above where a simple `1.0` is added to the potentially dangerous divisor in case the divisor (`mixed_alpha`) is equal to zero.

I simply saw how it was handled in the code lines above and think that it is even easier to ready than the if expression.

MichaelBuessemeyer: Oh, sadly I cannot comment the lines I mean in the comment above. The trick is already applied in the lines 218 - 228.

docs/tracing_ui.md
```diff
@@ -116,6 +116,12 @@ Note, not all control/viewport settings are available in every annotation mode.
...
 - `Loading Strategy`: You can choose between two different loading strategies. When using "best quality first" it will take a bit longer until you see data, because the highest quality is loaded. Alternatively, "Progressive quality" can be chosen which will improve the quality progressively while loading. As a result, initial data will be visible faster, but it will take more time until the best quality is shown.
+- `Blend Mode`: You can switch between two modes of blending the color layer. The default (Additive) simply sums up all color values of all visible layers. The cover mode renders all color layers on top of each other. Thus the top most layer covers the layers below. This blend mode is especially useful for datasets using multi modality layers. Here is an example for such a dataset by Bosch et al. [1]:
+
+|Additive Blend Mode        | &nbsp;&nbsp;&nbsp;&nbsp;Cover Blend Mode &nbsp; &nbsp;|
```
MichaelBuessemeyer, line 121: The `&nbsp;` are needed to artificially increase the width of the header cell so that both header cells have the same width and the images below will be the same size.

Other possibilities would be using html & css, but I thought this was the easiest way to get the wanted outcome. Other solutions to this problem can be found here: https://stackoverflow.com/questions/36121672/set-table-column-width-via-markdown

CHANGELOG.unreleased.md
```diff
@@ -12,6 +12,7 @@ For upgrade instructions, please check the [migration guide](MIGRATIONS.released
 
 ### Added
 - Added segment groups so that segments can be organized in a hierarchy (similar to skeletons). [#6966](https://github.com/scalableminds/webknossos/pull/6966)
+- Added a new "cover" blend mode which renders the visible layers on top of each other. The new blend mode can be selected in the Data Rendering settings in the right settings tab. [#6936](https://github.com/scalableminds/webknossos/pull/6936)
```
philippotto, line 15:
```suggestion
- Added a new "cover" blend mode which renders the visible layers on top of each other. The new blend mode can be selected in the Data Rendering settings in the settings tab of the left side bar. [#6936](https://github.com/scalableminds/webknossos/pull/6936)
```

docs/tracing_ui.md
```diff
@@ -116,6 +116,12 @@ Note, not all control/viewport settings are available in every annotation mode.
 #### Data Rendering
 - `Hardware Utilization`: Adjusts the quality level used for rendering data. Changing this setting influences how many data is downloaded from the server as well as how much pressure is put on the user's graphics card. Tune this value to your network connection and hardware power. After changing the setting, the page has to be refreshed.
 - `Loading Strategy`: You can choose between two different loading strategies. When using "best quality first" it will take a bit longer until you see data, because the highest quality is loaded. Alternatively, "Progressive quality" can be chosen which will improve the quality progressively while loading. As a result, initial data will be visible faster, but it will take more time until the best quality is shown.
+- `Blend Mode`: You can switch between two modes of blending the color layer. The default (Additive) simply sums up all color values of all visible layers. The cover mode renders all color layers on top of each other. Thus the top most layer covers the layers below. This blend mode is especially useful for datasets using multi modality layers. Here is an example for such a dataset by Bosch et al. [1]:
```
philippotto, line 119: Is it true that this only affects color layers? Segmentation layers are handled differently?

frontend/javascripts/messages.tsx
```diff
@@ -73,6 +74,8 @@ export const settingsTooltips: Partial<Record<keyof RecommendedConfiguration, st
   mouseRotateValue: "Rotation speed when using the mouse to drag the rotation.",
   zoom: "Zoom in or out in the data viewports",
   displayScalebars: "Show a scale in the lower-right corner of each viewport",
+  blendMode:
+    "Set the blend mode for the dataset. The additive mode (default) adds the data values of all layers. The cover mode only shows the data of the top most layer in the list.",
```
philippotto, line 78:
```suggestion
    "Set the blend mode for the dataset. The additive mode (default) adds the data values of all layers. In cover mode, layers are rendered on top of each other so that the data values of lower layers are hidden by values of higher layers.",
```

also: all layers or only color layers?

frontend/javascripts/oxalis/geometries/materials/plane_material_factory.ts
```diff
@@ -567,6 +568,19 @@ class PlaneMaterialFactory {
...
     );
+    this.storePropertyUnsubscribers.push(
+      listenToStoreProperty(
+        (storeState) => storeState.datasetConfiguration.blendMode,
+        (blendMode) => {
+          if (blendMode === BLEND_MODES.Additive) {
+            this.uniforms.blendMode.value = 1.0;
+          } else {
+            this.uniforms.blendMode.value = 0.0;
+          }
```
philippotto, line 579:
```suggestion
          this.uniforms.blendMode.value = blendMode === BLEND_MODES.Additive ? 1 : 0;
```

frontend/javascripts/oxalis/shaders/filtering.glsl.ts
```diff
@@ -113,14 +113,18 @@ export const getMaybeFilteredColorOrFallback: ShaderModule = {
       vec4 fallbackColor,
       bool supportsPrecomputedBucketAddress
     ) {
+      vec4[2] returnValue;
       vec4 color = getMaybeFilteredColor(layerIndex, d_texture_width, packingDegree, worldPositionUVW, suppressBilinearFiltering, supportsPrecomputedBucketAddress);
 
+      vec4 usedFallbackColor = vec4(0.0);
```
philippotto, line 119: the semantics of this var is a simple boolean, right? in that case, maybe define a struct with a vec4 and a bool?

frontend/javascripts/oxalis/shaders/main_data_shaders.glsl.ts
```diff
@@ -202,8 +203,11 @@ void main() {
             false,
             fallbackGray,
             !<%= name %>_has_transform
-          ).xyz;
-
+          );
+          float used_fallback = maybe_filtered_color_and_used_fallback[1].r;
+          // For uint24 the alpha channel is always 0.0 :/
```
philippotto, line 208: is this still true?

frontend/javascripts/oxalis/shaders/main_data_shaders.glsl.ts
```diff
@@ -202,8 +203,11 @@ void main() {
...
+          vec4 maybe_filtered_color = maybe_filtered_color_and_used_fallback[0];
+        color_value = maybe_filtered_color_and_used_fallback[0].rgb;
```
philippotto, line 210:
```suggestion
        color_value = maybe_filtered_color.rgb;
```

frontend/javascripts/oxalis/shaders/main_data_shaders.glsl.ts
```diff
@@ -222,14 +226,33 @@ void main() {
         color_value = abs(color_value - <%= name %>_is_inverted);
         // Catch the case where max == min would causes a NaN value and use black as a fallback color.
         color_value = mix(color_value, vec3(0.0), is_max_and_min_equal);
-        // Multiply with color and alpha for <%= name %>
-        data_color += color_value * <%= name %>_alpha * <%= name %>_color;
+        color_value = color_value * <%= name %>_alpha * <%= name %>_color;
+        // Marking the color as invalid by setting alpha to 0.0 if the fallback color has been used
+        // so that it does not cover other colors.
+        vec4 layer_color = vec4(color_value, used_fallback == 1.0 ? 0.0 : maybe_filtered_color.a * <%= name %>_alpha);
```
philippotto, line 232: I think, I'd simply rename `color_value` to `layer_value` and then don't define a new variable (as it's not necessary to have both at the same time, I think).

frontend/javascripts/oxalis/shaders/main_data_shaders.glsl.ts
```diff
@@ -222,14 +226,33 @@ void main() {
         color_value = abs(color_value - <%= name %>_is_inverted);
         // Catch the case where max == min would causes a NaN value and use black as a fallback color.
         color_value = mix(color_value, vec3(0.0), is_max_and_min_equal);
-        // Multiply with color and alpha for <%= name %>
-        data_color += color_value * <%= name %>_alpha * <%= name %>_color;
+        color_value = color_value * <%= name %>_alpha * <%= name %>_color;
```
philippotto, line 229: @MichaelBuessemeyer can you give this entire code block a bit of love? variable naming could be more consistent and I'd also hope that the code for the different blend mode could be extracted into helper functions?

philippotto (commented):
I left some code feedback :) My biggest question would be about segmentation layers: Did you discuss that these should be handled separately with the cover-mode?

philippotto:
> @philippotto Do you want to take over the pr review? Or should I ask daniel?

I reviewed it now, but I'm not sure whether the two of you already discussed some details I'm missing now.

Independent from that, I think, it would be good to show a new dev instance to the team to keep these in the loop who gave feedback last time.

MichaelBuessemeyer (commented):
Thanks for the feedback philipp. I think I covered everything. Let's wait for the CI and then ask for some further testing feedback. I hope everything should work now as expected.

docs/tracing_ui.md
MichaelBuessemeyer: Update: We already discussed a solution: 
1. Keep the segmentation layer special right now.
2. Add another pr that enabled editing the layer order
3. Add another pr that enables setting the blend mode for every single layer including the segmentation layer. The default for segmentation layer will be cover, the default for color layers will be additive.

philippotto (approved):
Looks great to me :) Looking forward to the future iterations with layer ordering and layer-specific blend modes.

PR #6939, opened 2023-03-23
Use new zip.js version to allow zip64 uploads

daniel-wer:
I wasn't able to upload the recent negative_float test dataset, getting the message that the zip file was invalid, even after re-zipping. I investigated and found out that the 4 year old repackaged version of zip.js we used doesn't support zip64 (which is used if the files > 4 GB are zipped). In the meantime, zip.js supports webpack and also supports zip64 which is why I decided to use the official version instead. Review ignoring white space changes.

- https://___.webknossos.xyz

### Steps to test:
- Upload a dataset and check that the uploader still is able to find wkw and nml files inside the zip

------


philippotto:
Very cool! Since you just looked into the topic: Any chance we could ditch the jszip library, too? It's currently used to let the user import a hybrid annotation as a zip into the skeleton tab via drag and drop. The code around it is a bit more complex, but maybe the interface didn't change much? If it's too much of a hassle, feel free to keep it as is :)

daniel-wer:
@philippotto Good suggestion! Was fairly easy to replace :)

philippotto (approved):
Yay! Works great for me :)

PR #6944, opened 2023-03-27
[WIP] Remove segment from list and add undo/redo for segments

philippotto:
… removing a segment

- https://___.webknossos.xyz

### Steps to test:
- abc

### Issues:
- fixes #

------

### DIFF
--- a/frontend/javascripts/libs/diffable_map.ts
+++ b/frontend/javascripts/libs/diffable_map.ts
@@ -1,12 +1,15 @@
 const defaultItemsPerBatch = 1000;
 let idCounter = 0;
-const idSymbol = Symbol("id"); // DiffableMap is an immutable key-value data structure which supports fast diffing.
+const idSymbol = Symbol("id");
+
+// DiffableMap is an immutable key-value data structure which supports fast diffing.
 // Updating a DiffableMap returns a new DiffableMap, in which case the Maps are
 // derived from each other ("dependent").
 // Diffing is very fast when the given Maps are dependent, since the separate chunks
 // can be compared shallowly.
 // The insertion order of the DiffableMap is not guaranteed.
-// Stored values may be null. However, `undefined` is equal to "does not exist".
+// Stored values may be null. However, be careful when dealing with `undefined`, as
+// it's interpreted as "does not exist", but can still be listed during enumeration.
 
 class DiffableMap<K extends number, V> {
   chunks: Array<Map<K, V>>;
--- a/frontend/javascripts/oxalis/model/actions/volumetracing_actions.ts
+++ b/frontend/javascripts/oxalis/model/actions/volumetracing_actions.ts
@@ -35,6 +35,7 @@ export type ImportVolumeTracingAction = ReturnType<typeof importVolumeTracingAct
 export type SetLargestSegmentIdAction = ReturnType<typeof setLargestSegmentIdAction>;
 export type SetSegmentsAction = ReturnType<typeof setSegmentsAction>;
 export type UpdateSegmentAction = ReturnType<typeof updateSegmentAction>;
+export type RemoveSegmentAction = ReturnType<typeof removeSegmentAction>;
 export type SetMappingIsEditableAction = ReturnType<typeof setMappingIsEditableAction>;
 
 export type ComputeQuickSelectForRectAction = ReturnType<typeof computeQuickSelectForRectAction>;
@@ -61,6 +62,7 @@ export type VolumeTracingAction =
   | SetContourTracingModeAction
   | SetSegmentsAction
   | UpdateSegmentAction
+  | RemoveSegmentAction
   | AddBucketToUndoAction
   | ImportVolumeTracingAction
   | SetLargestSegmentIdAction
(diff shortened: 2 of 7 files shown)


daniel-wer (commented):
Code looks very good to me :+1: 

- [x] During testing I encountered some inconsistent behavior related to branching off from an earlier undo state. In that case the undo stack doesn't seem to be correctly pruned and one version of the old branch remains on the stack instead of the version that was branched from. See this gif, but let me know if you can't follow what's happening:
segments_undo_redo_confusion
- [ ] When removing a segment from the list which had an ad-hoc or precomputed mesh loaded and then undoing that, the mesh check mark is restored but the mesh is not. Maybe the mesh loading can be retriggered in that case? Otherwise, the check mark shouldn't be checked.
- [x] [Low-Pri] Unrelated to this PR, but maybe it's quick to fix. The cross hair and brush icon next to the segment ID are in italics as well, looking slightly weird.

philippotto:
Thanks for your review & testing! I fixed the issues you found and also refactored the code a bit which is why I would recommend to review the changes commit-by-commit (one of them just extracts a large chunk of code into a new module). I hope the comments are self-explanatory, otherwise, let me know! Took me quite a bit of diving today to get into the save saga again :diving_mask:

frontend/javascripts/oxalis/model/sagas/undo_saga.ts
```diff
@@ -386,6 +390,7 @@ export function* manageUndoStates(): Saga<never> {
         if (!action) {
           throw new Error("Unexpected action");
         }
+        shouldClearRedoState = true;
```
daniel-wer, line 393: Why does the redo state always need to be cleared, even for actions that are not added to the undo stack? :thinking: Should it have been moved to line 397 instead (so into the `if (addToUndo)` block)?

daniel-wer (commented):
Please see my one question :) I'll test again once you say you're done with the PR.

frontend/javascripts/oxalis/model/sagas/undo_saga.ts
philippotto: Yes, valid question! I added a comment for that. Let me know if you agree with the reasoning. The comment:

```
// Even if addToUndo is false, we will clear the redo state. Otherwise,
// selecting a segment while being in an earlier state and then doing a redo operation
// would forget about the newly registered segment which is weird.
// Clearing the redo state after an explicit action (such as selecting a segment)
// seems more intuitive.
// Also note that a segment update as a consequence of an annotation operation
// is already accompanied by an empty redo stack (because the volume operation
// will have cleared that stack). So, in that case shouldClearRedoState is true,
// but essentially nothing will be cleared, anyway.
```

daniel-wer: Thank you for explaining. I agree with the reasoning, but am having some additional thoughts. I've actually wondered about this during the review, too, but shouldn't these other explicit segment update actions be undoable/redoable, then, too? It's a bit weird that something that clears the redo stack isn't undoable.
However, I also see the issue that the undo stack might get "crowded", but as long as only explicit user actions end up on the undo stack I would argue that's fine. Maybe in the future we can find a way to "highlight" where changes were undone or show a small description of what was just undone - which could help with that issue.

I don't think this needs to be fully discussed for this issue. The current state is fine to be merged.

philippotto: > I agree with the reasoning, but am having some additional thoughts. I've actually wondered about this during the review, too, but shouldn't these other explicit segment update actions be undoable/redoable, then, too? It's a bit weird that something that clears the redo stack isn't undoable.

Ok, convinced :) Especially, the fact that changing a segment color wasn't undoable disturbed me. I just pushed a commit which takes care of this. I think it makes the code more explicit / easier, too. Could you do test the feature again? I did so, too, obviously, but I don't know whether I'm not seeing the forest for the trees after all this :)

frontend/javascripts/components/color_picker.tsx
```diff
@@ -47,19 +47,26 @@ export function ChangeColorMenuItemContent({
...
     if (isDisabled) {
       return;
     }
     const colorRgb = Utils.hexToRgb(colorStr);
     const newColor = Utils.map3((component) => component / 255, colorRgb);
-    onSetColor(newColor);
+
+    // Only create a new undo state on the first color change event.
+    // All following color change events should mutate the most recent undo
+    // state so that the undo stack is not filled on each mouse movement.
```
daniel-wer, line 65: This works because the color picker is unmounted when it is closed (and remounted when it's opened again), right?

daniel-wer (approved):
Works wonderfully for me :tada:

frontend/javascripts/components/color_picker.tsx
philippotto: Exactly! At first, I wanted to hook into the visibility-has-changed prop of the antd popover, but it turned out that wasn't even triggered, since the entire popover was simply unmounted when the menu closed. So, it's definitely unmounted :)

Issue #6821, opened 2023-02-07
Support adding multiple layers with different voxel sizes separately


frcroth:
When exploring [storage.googleapis.com/neuroglancer-fafb-data/fafb_v14/fafb_v14_orig](https://storage.googleapis.com/neuroglancer-fafb-data/fafb_v14/fafb_v14_orig/) with voxel size 4,4,40 and then adding [storage.googleapis.com/neuroglancer-fafb-data/fafb_v14/fafb_v14_clahe](https://storage.googleapis.com/neuroglancer-fafb-data/fafb_v14/fafb_v14_clahe/) with voxel size 8,8,40 a warning occurs and the resulting data is messed up. What should happen is that on the second dataset, the mags are adjusted so that it starts at mag 2,2,1. 

#6670 already implemented this when exploring a data source with multiple layers of different voxel sizes (e.g. segmentation and image), however this does not work when adding one layer and then adding and exploring another layer.

philippotto:
With #6748 the restriction of uniform voxel sizes might be lifted (might take a few iterations). Therefore, I would advise to not tackle this issue here before it's clear whether the other issue will automatically solve it.

normanrz:
With #6748 and https://github.com/scalableminds/webknossos/pull/6943 merged, we should be able to tackle this now?

philippotto:
Yes, this should be possible now. Instead of the originally suggested "renaming" of the mag layers, the mag linking could be 1:1. Then, the layers with differing voxel sizes would need an affine `coordinateTransformations` property which does the scaling. The scaling itself can generalize over arbitrary voxel sizes (doesn't depend on power-of-two-factors).

We still have the concept of a singular "voxel size" property for the entire dataset. I think, the first layer that is added could dictate that voxel size and all other layers get a transformation if their voxel sizes differ.

frcroth:
@philippotto I think the problem is that the frontend calls exploreRemoteDatasource, but the backend has no knowledge if this is for a new dataset or if it is a new layer for an explored dataset. Thus the backend does not know the voxel size (for which coordinate transformations could then be applied). I think the frontend needs to pass a preferred voxel size on the exploreRemoteDataset API call.

PR #6947, opened 2023-03-28
Implement compressed segmentation

### LINKED ISSUE #6820: Support compressed segmentation for Neuroglancer precomputed datasets
#6716 does not support compressed segmentation: https://github.com/google/neuroglancer/tree/master/src/neuroglancer/sliceview/compressed_segmentation.

This format is used in all (?) segmentations, so currently segmentations can not be viewed.

There are two implementations: TS (https://github.com/google/neuroglancer/tree/master/src/neuroglancer/sliceview/compressed_segmentation) and C++ (https://github.com/seung-lab/compressedseg) with Python and Go bindings. There is no Java/Scala library.

frcroth:
- https://compressedsegmentation.webknossos.xyz

### Steps to test:
- Explore and load a segmentation
- gs://fafb-ffn1-20200412/segmentation (compare with [neuroglancer](https://neuroglancer-demo.appspot.com/#!%7B%22dimensions%22:%7B%22x%22:%5B4e-9%2C%22m%22%5D%2C%22y%22:%5B4e-9%2C%22m%22%5D%2C%22z%22:%5B4e-8%2C%22m%22%5D%7D%2C%22position%22:%5B140935.484375%2C58285.6875%2C3622.5%5D%2C%22crossSectionScale%22:36.59823444367803%2C%22projectionScale%22:262144%2C%22layers%22:%5B%7B%22type%22:%22segmentation%22%2C%22source%22:%22precomputed://gs://fafb-ffn1-20200412/segmentation%22%2C%22tab%22:%22source%22%2C%22name%22:%22segmentation%22%7D%5D%2C%22selectedLayer%22:%7B%22visible%22:true%2C%22layer%22:%22segmentation%22%7D%2C%22layout%22:%224panel%22%2C%22selection%22:%7B%7D%7D)) 
- gs://neuroglancer-janelia-flyem-hemibrain/v1.2/segmentation
I did not find segmentation datasets with uint32 as data type

### Current problems:
- ~~Segment ids are not correct. That is, the shapes that result are correct, but they do not have the right ids. Also, some segments are missing.~~
- ~~Performance is pretty bad right now~~
- ~~Ensure License compliance~~
- ~~Update docs: Remove wkconnect?~~

### Issues:
- fixes #6820 

------


webknossos-datastore/app/com/scalableminds/webknossos/datastore/datareaders/precomputed/compressedsegmentation/CompressedSegmentation.scala
```diff
@@ -0,0 +1,123 @@
+package com.scalableminds.webknossos.datastore.datareaders.precomputed.compressedsegmentation
+
+import java.nio.{ByteBuffer, ByteOrder}
+import scala.reflect.ClassTag
+
+// Based on https://github.com/janelia-flyem/compressedseg, see LICENSE.txt
+// Original author: Stephen Plaza (plazas@janelia.hhmi.org)
```
normanrz, line 7:
```suggestion
// Implementation of compressed segmentation as specified in:
// https://github.com/google/neuroglancer/tree/master/src/neuroglancer/sliceview/compressed_segmentation
// For a reference implementation, see also:
// https://github.com/janelia-flyem/compressedseg
```

webknossos-datastore/app/com/scalableminds/webknossos/datastore/datareaders/precomputed/compressedsegmentation/LICENSE.txt
```diff
@@ -0,0 +1,30 @@
...
+OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+(INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+Author: Stephen Plaza (plazas@janelia.hhmi.org)
+ Written as part of the FlyEM Project at Janelia Farm Research Center.
```
normanrz, line 30:
```suggestion
```

webknossos-datastore/app/com/scalableminds/webknossos/datastore/datareaders/precomputed/PrecomputedCompressorFactory.scala
```diff
@@ -1,16 +1,28 @@
...
-      case "raw"                     => nullCompressor
-      case "jpeg"                    => new JpegCompressor
-      case "compressed_segmentation" => throw new NotImplementedError
-      case _                         => throw new IllegalArgumentException(s"Chunk encoding '$encoding' not supported.")
+      case "raw"  => nullCompressor
+      case "jpeg" => new JpegCompressor
+      case "compressed_segmentation" =>
+        new CompressedSegmentationCompressor(dataType,
+                                             chunkSize,
+                                             compressedSegmentationBlockSize.getOrElse(Array(8, 8, 8)))
```
fm3, line 24: Is this block size of 8,8,8 documented somewhere as the default? Maybe add a link to the spec?

fm3: edit: found the 8,8,8 again as privat val in CompressedSegmentation.scala – maybe make it not private and access it here? Or pass None here, and then default to the default in the implementation? The duplication may lead to problems

webknossos-datastore/app/com/scalableminds/webknossos/datastore/datareaders/precomputed/compressedsegmentation/CompressedSegmentation.scala
```diff
@@ -0,0 +1,123 @@
...
+
+  val typeWidthBytes: Int
+
+  def readValue(input: Array[Int], position: Int): T
+
+  def initializeArray(length: Int): Array[T]
+
+  private def decompressChannel(input: Array[Int], volumeSize: Array[Int], blockSize: Array[Int]): Array[T] = {
+    assert(blockSize.length == 3)
+    val numElements = volumeSize(0) * volumeSize(1) * volumeSize(2)
```
fm3, line 22: is the order of volumeSize elements fixed? can we rely on 0,1,2 being x,y,z, and 3 being c? Could there be a time axis as well?

webknossos-datastore/app/com/scalableminds/webknossos/datastore/datareaders/precomputed/compressedsegmentation/CompressedSegmentation.scala
```diff
@@ -0,0 +1,123 @@
...
+    } else {
+      volumeSize
+    }
+    val input32 = new Array[Int](encodedBytes.length / 4)
+    ByteBuffer.wrap(encodedBytes).order(ByteOrder.LITTLE_ENDIAN).asIntBuffer().get(input32)
+    val values = decompressChannels(input32, vs, blockSize)
+    values.flatMap(v => longToBytes(valueAsLong(v)))
+  }
+
+  def longToBytes(l: Long): Array[Byte] = {
```
fm3, line 90:
```suggestion
  private def longToBytes(l: Long): Array[Byte] = {
```

normanrz:
Probably a different issue: I just tried to import `gs://fafb-ffn1-20200412/segmentation` and then `gs://neuroglancer-fafb-data/fafb_v14/fafb_v14_clahe`. The explore feature did not pick up that the `clahe` layer actually started in mag 2 (8x8x40nm).

fm3:
> Probably a different issue: I just tried to import `gs://fafb-ffn1-20200412/segmentation` and then `gs://neuroglancer-fafb-data/fafb_v14/fafb_v14_clahe`. The explore feature did not pick up that the `clahe` layer actually started in mag 2 (8x8x40nm).

Yes, that’s https://github.com/scalableminds/webknossos/issues/6821

webknossos-datastore/app/com/scalableminds/webknossos/datastore/datareaders/precomputed/PrecomputedCompressorFactory.scala
frcroth: There is no word in the standard about a default block size. It is the default in the janelia-flyem implementation "typically (8,8,8)". The default parameter in CompressedSegmentation/decompress could be removed, as it is not used anyway. However I would like to have some value here and not an option. This is a case that can not happen in validly formed json files anyway as block size is a required field, so I would semantically like to use `get` here, which is of course verboten.

webknossos-datastore/app/com/scalableminds/webknossos/datastore/datareaders/precomputed/compressedsegmentation/CompressedSegmentation.scala
frcroth: The specification only describes 3d volume data, so I think it can be left like this.

webknossos-datastore/app/com/scalableminds/webknossos/datastore/datareaders/Compressor.scala
```diff
@@ -268,16 +295,39 @@ class JpegCompressor() extends Compressor {
...
   }
 }
+
+class CompressedSegmentationCompressor(dataType: PrecomputedDataType, volumeSize: Array[Int], blockSize: Array[Int])
+    extends Compressor {
+  override def getId: String = "compressedsegmentation"
+
+  override def toString: String = s"compressor=$getId/dataType=${dataType.toString}"
+
+  private def blockSizeVec = Vec3Int(x = blockSize(0), y = blockSize(1), z = blockSize(2))
```
fm3, line 319: could blockSize be Vec3Int everywhere? (even in case class PrecomputedScale). Then the json validator would already reject it if it does not match, and we wouldn’t need the conversion here

fm3 (commented):
Thanks for addressing the feedback! I added just one more small suggestion

webknossos-datastore/app/com/scalableminds/webknossos/datastore/datareaders/Compressor.scala
frcroth: oh I did not know Vec3Int could be automatically parsed.

fm3 (approved):
:tada: :rocket:

PR #6950, opened 2023-03-28
Log all details on deleting annotation layer

fm3:
- https://deletelayerlogdetails.webknossos.xyz

### Steps to test:
- delete annotation layer
- backend logging should show full details needed to later restore it
- (I tested it, should be fine)


daniel-wer (approved):
:+1:

Issue #5633, opened 2021-07-23
Remove or fix persistent mesh import of "ordinary" STL files


philippotto:
In addition to computing meshes ad-hoc and loading meshes from precomputed files, we also have the support to import STL files into a tracing view. The import handles two cases:

(1) Import an STL which was previously downloaded via webKnossos (that file encodes the segment id in the header and can be restored as if the mesh was just recomputed).
(2) Import an STL which was created by another program. For such an STL we have very little information (i.e., don't know which segment is represented).

For (2) we once built the feature to store the mesh on the back-end. Also, properties, such as the position and scale, of the mesh could be edited in the UI. This feature has always been kind of obscure and probably wasn't used much for that reason.

I noticed that the support for ordinary STL files is broken currently. Importing such an STL adds the mesh to the scene, but it's not listed in the sidebar. I couldn't find my dummy files in the scene, so I'm not even sure, there are rendered properly.

We should discuss whether the support for persisted and "ordinary" STL should be repaired or simply removed. I'd vote for the latter.

Remove in:
- [X] frontend
- [ ] backend

philippotto:
Some users reported that they used the STL feature to add some shapes, such as planes:

![image](https://user-images.githubusercontent.com/2486553/144221800-d8f00ce1-d650-4958-ab19-8d7022fe037d.png)

We might want to reconsider the support for that.

PR #6960, opened 2023-04-03
[WIP] Isolate meshes by layer and remove old mesh-storage code

philippotto:
- https://___.webknossos.xyz

### Steps to test:
- abc

### Issues:
- fixes #

------


philippotto:
@fm3 What do you think about removing the mesh-storage related code in the back-end in this PR (see #https://github.com/scalableminds/webknossos/issues/5633)? When I started rebuilding the code for proper isolation, I ran into the old, (mostly) dead code and removed it. Do you have capacities for this? Or would you like to postpone it?

fm3:
The last post in the issue stated that we might want to reconsider if we really want to remove that feature (?)

Either way, I’d defer removing the backend code in order not to block this PR until I have time. But I will gladly do so in a follow-up PR

philippotto:
> The last post in the issue stated that we might want to reconsider if we really want to remove that feature (?)

Yes, but I'd put the emphasis on "might" :see_no_evil: The last feedback we got on that topic was over a year ago (and in total it was by one user). The feature was already effectively removed, anyway, and I don't really see big reasons to reintroduce it. For the mentioned use case, one could also find other ways (e.g., create a volume annotation to get the desired shape as an isosurface or allow to create a plane with a skeleton).

> Either way, I’d defer removing the backend code in order not to block this PR until I have time. But I will gladly do so in a follow-up PR

Ok, cool :+1:

frontend/javascripts/oxalis/controller/mesh_controller.ts
```diff
@@ -0,0 +1,167 @@
...
+import app from "app";
+import { mergeVertices } from "libs/BufferGeometryUtils";
+import _ from "lodash";
+import type { Vector3 } from "oxalis/constants";
+import CustomLOD from "oxalis/controller/custom_lod";
+import { getSegmentColorAsHSLA } from "oxalis/model/accessors/volumetracing_accessor";
+import { NO_LOD_MESH_INDEX } from "oxalis/model/sagas/isosurface_saga";
+import Store from "oxalis/store";
+
+export default class MeshController {
```
MichaelBuessemeyer, line 13: I would rename this into "IsosurfaceController" or "IsosurfaceMeshController". Because there are still other meshes in the `SceneController` like the bounding boxes and more (see `SceneController.createMeshes`) that are not handled by this class.

MichaelBuessemeyer (approved):
Thanks for this bug fixing & refactoring 🥇 🦖 .

I only left some small suggestions (renaming & moving a method into another class). As they are not much work / shouldn't be risky changes I already gave you an approve :) I hope that's alrighty.

I did not check, whether you removed every single method that is not needed anymore is removed. I hope that's ok. If you want me to also check this, just ping me :).

frontend/javascripts/oxalis/controller/mesh_controller.ts
philippotto: good point! I renamed it to `SegmentMeshController`, since I feel like we should get rid of the term "isosurface" in the long run (the UI already uses meshes mostly).

frontend/javascripts/oxalis/controller/mesh_controller.ts
```diff
@@ -0,0 +1,167 @@
...
+        }
+      },
+    );
+    delete this.isosurfacesGroupsPerSegmentationId[layerName][segmentationId];
+  }
+
+  getIsosurfaceGeometryInBestLOD(segmentId: number, layerName: string): THREE.Group {
+    const bestLod = Math.min(
+      ...Object.keys(this.isosurfacesGroupsPerSegmentationId[layerName][segmentId]).map((lodVal) =>
+        parseInt(lodVal),
```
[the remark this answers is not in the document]
philippotto: I think, the case in which the look ups would fail are unexpected (e.g., the UI shouldn't expose setIsosurfaceVisibility if there is no corresponding mesh) which is why they should fail. Adding a defensive check only to throw another error doesn't feel valuable to me. Doing a noop could be the alternative, but I'd rather throw an exception here. Since this will likely be in an eventhandler, wk shouldn't crash anyway.

PR #6972, opened 2023-04-06
Avoid that an unsaved state is shown when loading an annotation

### LINKED ISSUE #5882: Creating new annotation immediately leads to "unsaved" state
Not really a bug, but also not nice either. Reproduce:

- create new annotation
- --> save button shows unsaved state

The save queue holds:

```
[
    {
        ...
        "actions": [
            {
                "name": "updateTracing",
                "value": {
                    "activeSegmentId": 2504698,
                    "editPosition": [
                        3584,
                        3584,
                        1024
                    ],
                    "editRotation": [
                        0,
                        0,
                        0
                    ],
                    "largestSegmentId": 2504697,
                    "zoomLevel": 1
                }
            },
            {
                "name": "updateTdCamera",
                "value": {}
            }
        ],

        ...
    {
        ...
        "actions": [
            {
                "name": "updateTdCamera",
                "value": {}
            }
        ],
        ...
]
```

which should not be necessary.

daniel-wer:
The functionality to "hide" some changes to the 3d viewport from time tracking (so from the save queue) was already there, but in this spot the boolean was not set. Since this piece of code is only triggered when changing the size of the viewports or when changing the zoom (which triggers an update action, anyways) or clipping distance, it's not relevant for time tracking.

I've also added that the save queue entries won't be compressed in the dev environment. This eases debugging and one cannot forget to revert the change before committing any longer. Let me know if you think that's an issue.

- https://___.webknossos.xyz

### Steps to test:
- Open or create an annotation. The save button shouldn't show any unsaved changes.
- Move, zoom or rotate in the 3D viewport -> There should be unsaved changes.
- Other changes should still show unsaved changes and be saved across reloads.

### Issues:
- fixes #5882 

------

### DIFF
--- a/CHANGELOG.unreleased.md
+++ b/CHANGELOG.unreleased.md
@@ -23,6 +23,7 @@ For upgrade instructions, please check the [migration guide](MIGRATIONS.released
 
 ### Fixed
 - Fixed incorrect initial tab when clicking "Show Annotations" for a user in the user list. Also, the datasets tab was removed from that page as it was the same as the datasets table from the main dashboard. [#6957](https://github.com/scalableminds/webknossos/pull/6957)
+- Fixed that unsaved changes were shown when opening an annotation, although there weren't any. [#6972](https://github.com/scalableminds/webknossos/pull/6972)
 
 ### Removed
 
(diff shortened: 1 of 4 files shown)


philippotto (approved):
Awesome, works great :)

I agree with your compress suggestion :+1: One could even think about having a variable exposed in `window` (or GET parameter) so that we can switch the behavior on prod instances ad-hoc. However, I think, this is something for a future PR (I have some other ideas for such debug-related use cases).

PR #6977, opened 2023-04-12
If uploaded dataset needs conversion, tag user only after conversion success

### LINKED ISSUE #6672: Hand out mailchimp tag HasUploadedOwnDataset only after successful conversion
This tag triggers an email to the user, which reads like the dataset is now ready to use. This means this tag should only be sent out after we are sure that the dataset is actually ready to use (if a conversion job was needed, only send out the tag after the job was successful)

fm3:
Hand out mailchimp tag HasUploadedOwnDataset (which triggers an onboarding email) only after successful conversion

### Steps to test:
(I tested this locally, I think for a review you only need to have a look at the code)
- Add logging to mailchimp tagging (e.g. `  def tagUser(user: User, tag: MailchimpTag): Unit = {logger.info(s"tagging as ${tag.toString}")` … in `MailchimpClient.scala`, change mail.logToStdout to true in `application.conf`
- Enable jobs, set up worker
- Upload wkw dataset, tagging should happen immediately after upload
- Uplaod tiff dataset, tagging should happen only after successful conversion job (should come together with the job success email)

### Issues:
- fixes #6672

------


conf/webknossos.latest.routes
```diff
@@ -100,7 +100,7 @@ PUT           /datastores/:name/datasource
 PUT           /datastores/:name/datasources                                             controllers.WKRemoteDataStoreController.updateAll(name: String, key: String)
 PATCH         /datastores/:name/status                                                  controllers.WKRemoteDataStoreController.statusUpdate(name: String, key: String)
 POST          /datastores/:name/reserveUpload                                           controllers.WKRemoteDataStoreController.reserveDataSetUpload(name: String, key: String, token: String)
-POST          /datastores/:name/reportDatasetUpload                                     controllers.WKRemoteDataStoreController.reportDatasetUpload(name: String, key: String, token: String, dataSetName: String, dataSetSizeBytes: Long)
+POST          /datastores/:name/reportDatasetUpload                                     controllers.WKRemoteDataStoreController.reportDatasetUpload(name: String, key: String, token: String, dataSetName: String, dataSetSizeBytes: Long, needsConversion: Boolean)
```
frcroth, line 103: Doesn't this need some front end change?

frcroth (approved):

conf/webknossos.latest.routes
fm3: Fortunately not, this route is only called by the datastore, not the frontend

PR #6978, opened 2023-04-12
Fix incorrect rotation matrix after selecting nodes

philippotto:
Nodes that are created in arbitrary mode have an attached non-default rotation. Selecting that node will center their position and change apply their rotation. In orthogonal mode, this shouldn't happen as that rotation is used for bucket picking.

- https://fixchangingrotation.webknossos.xyz

### Steps to test:
- switch to flight mode with non-default angles (e.g., 45**3)
- create some nodes
- switch back to ortho mode and select nodes
- data should be rendered without problems

### Issues:
- second fix for #5288 

------

### DIFF
--- a/frontend/javascripts/oxalis/model/reducers/flycam_reducer.ts
+++ b/frontend/javascripts/oxalis/model/reducers/flycam_reducer.ts
@@ -166,9 +166,7 @@ export function setDirectionReducer(state: OxalisState, direction: Vector3) {
 }
 
 export function setRotationReducer(state: OxalisState, rotation: Vector3) {
-  // Don't change rotation in orthogonal mode, because this would lead to incorrect
-  // buckets being selected.
-  if (state.dataset != null && state.temporaryConfiguration.viewMode !== "orthogonal") {
+  if (state.dataset != null) {
     const [x, y, z] = rotation;
     let matrix = resetMatrix(state.flycam.currentMatrix, state.dataset.dataSource.scale);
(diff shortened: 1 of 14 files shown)


frontend/javascripts/oxalis/model/reducers/flycam_reducer.ts
```diff
@@ -164,8 +164,11 @@ export function setDirectionReducer(state: OxalisState, direction: Vector3) {
     },
   });
 }
+
 export function setRotationReducer(state: OxalisState, rotation: Vector3) {
-  if (state.dataset != null) {
+  // Don't change rotation in orthogonal mode, because this would lead to incorrect
+  // buckets being selected.
```
philippotto, line 170:
```suggestion
  // buckets being selected for rendering.
```

daniel-wer (approved):
Code LGTM :+1:

daniel-wer:
I followed the testing instructions and the issue still occured :thinking: 

image

philippotto:
> I followed the testing instructions and the issue still occured thinking

Argh, I think I had "prepared" nodes with rotations and after building the fix only ensured that selecting these nodes didn't cause rendering bugs.

> Edit: The issue actually already occurs when switching back to orthogonal, no node selection needed. Maybe, the additional condition prevents the rotation reset that's triggered when switching to orthogonal?

Yes, that was exactly it. Thanks for catching this :pray: :pray: :pray: 

Could you have another look for a final thumbs up?

daniel-wer (approved):
Nice, works well :+1:

Issue #6969, opened 2023-04-06
Unify backend-internal caching


fm3:
We have a couple of different approaches to caching inside of the backend. Should be unified.

 - ✓ First step: fill out the table below
 - Check which features are needed where
 - Design API (how high-level is possible?)
 - Build + use unified solution
 - Test

|                               | Temporary Store    | LRU Concurrent Cache | Redis Temporary Store | ALFU cache |
|---|---|---|---|---|
| catch async requests          | ✗                  | ✗                    | ✗                     | ✓          |
| removal callback              | ✗                  | ✓                    | ✗                     | ✓          |
| persist across restarts       | ✗                  | ✗                    | ✓                     | ✗          |
| timeToLive                    | ✓                  | ✗                    | ✓                     | ✓          |
| timeToIdle                    | ✗                  | ✗                    | ✗                     | ✓          |
| clear conditional             | ✓                  | ✓                    | ✓                     | ✓          |
| clear all                     | ✓                  | ✓                    | ✗                     | ✓          | 
| max elements                  | ✗                  | ✓                    | ✗                     | ✓          |
| element weight                | ✗                  | ✗                    | ✗                     | ✓          |
| thread safety                 | map. synchronized  | map. synchronized    | client. synchronized  | caffeine   |
| unexpected removal            | only on wk restart | LRU                  | never                 | LFU        |
| key                           | Any                | Any                  | String                | Any        |
| returns                       | Option             | Option               | Fox[Option]           | Fox        |
| getOrLoad                     | ✗                  | ✓                    | ✗                     | ✓          |
| usable get                    | ✓                  | ✓                    | ✓                     | ✗          |
| clear returns count           | ✗                  | ✓                    | ✗                     | ✓          |

fm3:
I’d say that the PRs # #6630, #6980, #7007, #7067 have pushed this as far as is needed for the time being. The remaining use cases do (somewhat) justify the different implementations. We can do another round of refactoring if we change wkw shard handling to no longer use mmap

PR #6980, opened 2023-04-13
Replace usercache by AlfuFoxCache

fm3:
This is the one remaining usage of the play cache api (the others are replaced in #6903 ).
So this PR contributes to unifying the caching usages.

- https://nomoreplaysynccache.webknossos.xyz

### Steps to test:
- Add logging to cached user request, load some pages, see that there is only one cache miss.

### Issues:
- contributes to #6969


app/models/user/UserService.scala
```diff
@@ -78,11 +81,8 @@ class UserService @Inject()(conf: WkConf,
       _ <- bool2Fox(userBox.isEmpty) ?~> "organization.alreadyJoined"
     } yield ()
 
-  def findOneById(userId: ObjectId, useCache: Boolean)(implicit ctx: DBAccessContext): Fox[User] =
-    if (useCache)
-      userCache.findUser(userId)
-    else
-      userCache.store(userId, userDAO.findOne(userId))
+  def findOneCached(userId: ObjectId, useCache: Boolean)(implicit ctx: DBAccessContext): Fox[User] =
```
frcroth, line 84:
```suggestion
  def findOneCached(userId: ObjectId)(implicit ctx: DBAccessContext): Fox[User] =
```

fm3: Good catch, looks like I hit request review a bit too early

frcroth (approved):

PR #6984, opened 2023-04-17
Fix that hiding all layers would crash wk

philippotto:
- https://___.webknossos.xyz

### Steps to test:
- open a dataset
- hide all layers
- should not crash

### Issues:
- follow-up for #6943 

------
- [x] ~Updated [changelog](../blob/master/CHANGELOG.unreleased.md#unreleased)~ Since this is a follow-up for 43a119861620ff8412995f44e904ae2250f36719 which was merged today, I think a changelog entry is not necessary


fm3 (approved):
Works for me :)

PR #6988, opened 2023-04-18
Discourage rest api use in docs

fm3:
In preparation for a new API versioning protocol between python client and server, I’d like to deprecate users interacting with the REST API directly.
Any opinions?


docs/rest_api.md
```diff
@@ -1,5 +1,11 @@
 # WEBKNOSSOS REST API
 
+## WARNING: Interacting with the REST API directly is deprecated
+
+As of 2023, please use the WEBKNOSSOS python client to interact with WEBKNOSSOS programmatically. The API endpoints described in this document are now deprecated and after a grace period will no longer be held stable.
```
philippotto, line 5: how about linking the docs for that?

philippotto (approved):
Makes sense :+1: See my one comment.

PR #6996, opened 2023-04-20
Add folders to dataset table

### LINKED ISSUE #6883: Display folders next to datasets in middle view
When opening a folder in the dataset dashboard, the table only lists the datasets in that folder. It would be useful to also list the folders in that same table. Some columns would remain empty then because they don't make sense for the folders:

- no tags
- no creation date (probably not known?)
- Actions could simply show "Settings" (same as rightclicking a folder in the sidebar then)

- [x] To make it easier to distinguish folders from datasets in the table, the table rows should have a icon at the front (same ones that are used in the right sidebar).

Things to keep in mind when implementing this:
- [x] context menu for folders in the table
- [x] double click for folders in table (should open that folder then)
- [x] single click selection (not sure whether we want to display the folder details similar to how the infos of the current folder are displayed right now; I could imagine that this might be confusing)
- [ ] drag-and-drop (maybe don't support this for the 1st iteration?)
  - for folders into sidebar
  - for datasets in table to folder in table
  - for folders and datasets (mixed) into sidebar

MichaelBuessemeyer:
This PR adds the child folders of the currently active folder in the dashboard to the dataset table. This pr should include all changes mentioned in [#6883](https://github.com/scalableminds/webknossos/issues/6883), except for the drag & drop feature. I disabled this for the folders as suggested to maybe not include this in the first iteration. Although if you think this should already be included, I can do it. I do not have an opinion on this except that leaving this feature away makes this pr earlier merge ready

- https://___.webknossos.xyz

### TODOs: 
- [x] Open the folder settings upon clicking on some icon in the actions column for folders
- [x] Implement context menu support for the folder rows
- [x] disable drag & drop for the folders
- [x] Decrease the width of the types column
- [x] The settings actions for a folder should only be available to an admin or such.

Optionally:
- ~~[ ] allow dragging dataset rows into a folder row~~
  - Not so important IMO: one can still drag the rows to the left sidebar and drop them there into the respective folder.

### Bugs
- [x] The expand icon should not be rendered for the folders
- [x] The details for the folders seem broken

### Steps to test:
- Open the dataset table in the dashboard
- Select the root folder in the left folder tree view
- Right-click on the root folder and create a new folder. The new folder should be listed in the table in the center.
- creating another folder should also result in the folder appearing in the table.
- Right-clicking on a folder in the table should open up the same context menu as if one would click on the folder in the folder hierarchy view on the left.
- single clicking a folder should show its details on the right (another single click should deselect it)
- selecting multiple folders should not work
- drag & drop of folders in the table should not work
- double-clicking a folder should open that folder's contents in the table
- If logged in as an unprivileged user (non admin / dataset manager) manipulating any folder setting should not work / not be available in the UI or the UI should be disabled.
- Please look at [#6883](https://github.com/scalableminds/webknossos/issues/6883) for more details on how the folder rows should behave. (But I think I covered everything in the list above)

### Issues:
- fixes #6883

------


MichaelBuessemeyer (commented):
I think everything should be working. Additionally, I looked through the code and it should be review ready.

@philippotto do you have time to take on this review? The pr is quite big/medium-sized. Or should I ask daniel?

philippotto:
> @philippotto do you have time to take on this review?

Yes, I think so :)

MichaelBuessemeyer:
Oh btw:
A thought that I had: In this pr, there is the problem that the table needs to handle both dataset-objects and folder-objects as a potential row. Thus there are quite a lot of ifs and else structs added in the code. IMO this is not ideal.

My suggestion to tackle this would be to write a class for datasets and one for folders that have the same interface but depending on their implementation behave/render differently:
e.g.  in the dataset class, there would be a function to render the name column:
```
   renderNameColumn(){
             return (
                  <>
                    <Link
                      to={`/datasets/${datasetOrFolder.owningOrganization}/${datasetOrFolder.name}/view`}
                      title="View Dataset"
                      className="incognito-link"
                    >
                      {datasetOrFolder.name}
                    </Link>
                    <br />

                    {context.globalSearchQuery != null ? (
                      <BreadcrumbsTag parts={context.getBreadcrumbs(datasetOrFolder)} />
                    ) : null}
                  </>
                );
     }
```
And the folder class would look like:
```
 renderNameColumn(){
  return datasetOrFolder.name;
 }
 ``` 
 
 And then we could just render the name column with:
```
   <Column
            title="Name"
            /*.........*/
            render={(_name: string, datasetOrFolder: DatasetOrFolder) => {
              return datasetOrFolder.renderNameColumn();
            }}
          />
```

But as we usually do not create classes for such cases I did not implement this in the first version and decided to simply propose this first. IMO this would make the code cleaner and also easier to read and in case another type of row is added, this would be easier to extend upon.

What do you think @philippotto?

**Comment by philippotto** [comment]
> But as we usually do not create classes for such cases I did not implement this in the first version and decided to simply propose this first. IMO this would make the code cleaner and also easier to read and in case another type of row is added, this would be easier to extend upon.

I agree in principle, but I'm a bit hesitant, since this would be a new pattern in this code base. However, I think, that this would be alright as long as those instances don't live in the react state. Otherwise, this would seem weird to me since these instances would only be dumb wrappers with some rendering behavior living in react state (I wouldn't want the caching context to be responsible for constructing these instances).

I imagine something like this:

```js
class DatasetRendering {
   renderNameColumn() {
    return ...;
   }
}

class FolderRendering {
   renderNameColumn() {
    return ...;
   }
}

function getRenderer(datasetOrFolder) {
  if (datasetOrFolder ... is dataset) {
    return DatasetRendering(datasetOrFolder);
  } else {
    ...
  }
}

renderSomething() {
  return getRenderer(datasetOrFolder).renderNameColumn();
}
```

I'd hope that the browser can optimize these short lived objects away. But maybe you could make a micro benchmark to test that this only has a negligible impact when being compared to a simple if-else.

What do you think?

philippotto (commented):
Awesome! Only commented some smaller stuff :) Also one other small thing: The "New Folder" in the Actions column might be clearer as "New Subfolder".

I know that the original issue outlined that dnd-features should come in another iteration. However, I noticed one particular dnd use case (that I just added to the issue) which is something I immediately tried out because it seemed intuitive. Namely: dragging a dataset from the dataset table to a folder within the same table. Could you have a look whether this is easy to achieve? I think, it's easier than the other dnd-features, because the dragged items (the datasets) are already implemented. the drag-target definition in the sidebar might be easy to reuse for the drag targets in the dataset table. if it takes more than 30 minutes, feel free to defer this to another PR :)

frontend/javascripts/dashboard/dataset/dataset_collection_context.tsx
```diff
@@ -179,6 +184,10 @@ export default function DatasetCollectionContextProvider({
     return breadcrumbs;
   };
 
+  const getActiveSubfolders = () => {
+    return folderHierarchyQuery.data?.itemById[activeFolderId ?? ""].children || [];
```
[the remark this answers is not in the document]
MichaelBuessemeyer: I just checked in the browser console using a basic example: 
```
$: const a = {c: 1, b: {a: 2}}
undefined
$: a.b.a
2
$: a?.d.asda
VM503:1 Uncaught TypeError: Cannot read properties of undefined (reading 'asda')
    at <anonymous>:1:5
(anonymous) @ VM503:1
$: a?.d?.asda
undefined
```
I don't know why TS did not catch this :thinking:. I'll fix it

philippotto:
> My preferred version would be working with objects:

Yes, I like that one, too :+1:

MichaelBuessemeyer:
## TODO for me:

- [ ] Found a :bug:: A folder with only subfolders and no datasets as children renders and empty folder / an empty table :/

frontend/javascripts/dashboard/folders/details_sidebar.tsx
```diff
@@ -202,7 +204,10 @@ function FolderDetails({
 }) {
   let message = getMaybeSelectMessage(datasetCount);
   if (!displayedFolderEqualsActiveFolder) {
-    message = datasetCount > 0 ? "Double-click the folder to list these datasets." : "";
+    message =
+      datasetCount > 0
+        ? `Double-click the folder to list these ${pluralize("dataset", datasetCount)}.`
```
philippotto, line 209: "these" still needs to be adapted to the datasetCount.

frontend/javascripts/dashboard/advanced_dataset/dataset_table.tsx
```diff
@@ -420,12 +509,17 @@ class DatasetTable extends React.PureComponent<Props, State> {
             .reverse()
             .value()
         : dataSourceSortedByRank;
+    const sortedDataSourceRenderer: RowRenderer[] = sortedDataSource.map((record) =>
```
philippotto, line 512:
```suggestion
    const sortedDataSourceRenderers: RowRenderer[] = sortedDataSource.map((record) =>
```

frontend/javascripts/dashboard/advanced_dataset/dataset_table.tsx
```diff
@@ -445,8 +539,10 @@ class DatasetTable extends React.PureComponent<Props, State> {
         />
         <FixedExpandableTable
           childrenColumnName="notUsed"
-          dataSource={sortedDataSource}
-          rowKey="name"
+          dataSource={sortedDataSourceRenderer}
+          rowKey={(record: RowRenderer) =>
+            isRecordADataset(record.data) ? record.data.name : record.data.key
```
philippotto, line 544: couldn't this be a method on the renderer, too?

philippotto (commented):
great to see the refactoring worked out :)

MichaelBuessemeyer:
Hi @philippotto,
I addressed all your feedback. This include that it is now possible to drop the datasets into the folders listed in the table.

A short discussion with daniel resulted in only highlighting the folders in the dataset table if the user drags datasets over them and not all the time as done in the left folder hierarchy view. We did this, because in the artificial scenario, where I had only 2 datasets but 70 folders due to the benchmark mentioned above, having all folders highlighted looked confusing. Additionally, it is not ideal that the "drop area highlight color" is the same color used to mark datasets in the table as selected.
Maybe my scenario with 70 folders and only two datasets was too artificial and the regular use case looks also alright when the folders in the table are highlighted all the time. Because highlighting the folder hierarchy view all the time but the folders in the dataset table only once they are hovered is a mismatch in behaviour, which might not be ideal.
Do you have any preference or idea on how to improve this? react-dnd seems not to support custom cursors on drag and drop to show indicate whether something can be dropped or not. There only seem to exist hacky solutions.

philippotto:
Awesome, works great :+1: 

> A short discussion with daniel resulted in only highlighting the folders in the dataset table if the user drags datasets over them and not all the time as done in the left folder hierarchy view. We did this, because in the artificial scenario, where I had only 2 datasets but 70 folders due to the benchmark mentioned above, having all folders highlighted looked confusing. Additionally, it is not ideal that the "drop area highlight color" is the same color used to mark datasets in the table as selected.
> Maybe my scenario with 70 folders and only two datasets was too artificial and the regular use case looks also alright when the folders in the table are highlighted all the time. Because highlighting the folder hierarchy view all the time but the folders in the dataset table only once they are hovered is a mismatch in behaviour, which might not be ideal.
> Do you have any preference or idea on how to improve this? react-dnd seems not to support custom cursors on drag and drop to show indicate whether something can be dropped or not. There only seem to exist hacky solutions.

I think the current way is absolutely fine :+1: The left sidebar is only highlighted during drag, so that the user knows that they can use the sidebar as a drop target (since the table and the sidebar are two visually disconnected components I found this sensible when I implemented this). Dragging a DS from the table to a folder *in the table* feels way more intuitive, since source and target are in the same component. Therefore, there's no need to highlight all valid drag targets in my opinion. Thus, only highlighting a folder while dragging a DS over it feels like the right approach to me :+1:

philippotto (approved):
Excellent as usual :+1: :)

PR #7000, opened 2023-04-21
Fix that changing a segment color could lead to a crash

philippotto:
(if no mesh was loaded yet for that segment)

- https://___.webknossos.xyz

### Steps to test:
- open an annotation
- change the color of a segment

### Issues:
- follow-up for #6960 

------


daniel-wer (approved):
LGTM :+1:

PR #7001, opened 2023-04-21
Allow batch reading of mesh chunks

### LINKED ISSUE #6752: Optimize loading of precomputed meshes of oversegmentation
If a meshfile exists for the oversegmentation, webknossos can load meshes for any agglomerate mapping, by fetching the oversegmentation meshes for all segments of an agglomerate and giving them all the same color. The back-end side for this is implemented in `listMeshChunksForSegmentForVersion` and `readMeshChunkForVersion` in `DSMeshController.scala`.

We had a couple of ideas for optimizing the performance
 - the front-end currently sends one request per chunk per segment id
 - instead it could send just one request per chunk, including a list of meshFile/byteOffset/byteSize tuples for all relevant segments, reducing the number of http requests
 - the back-end could sort this by byte offset to optimize for spinning-disk based reading of the hdf5 file
 - optional: similar to bucket loading, we can think about loading more than one chunk per http request

Proposal: request contains json list, as described above. Reply is binary, consisting of
 - Length of jumptable (one long)
 - The jumptable (start offsets for each segment content)
 - the concatenated draco contents

frcroth:
- https://optimizepremeshesoverseg.webknossos.xyz/

### Steps to test:
- open l4dense_motta_et_al_dev_v2 in view mode
- enable mapping 80
- load a precomputed mesh for a segment (e.g., at 2905, 4566, 1786)

### TODO
- [X] Adapt front end
- [X] Use new format on other route
- [X] Tweak performance so that it generalizes over different mesh files
- ~[ ] Use uint32 instead of uint64 for jump table so that Uint32Array can be used in JS~
- [x] ensure proper error handling when decoding fails (independent of this PR, strictly speaking)
- [ ] debug why some requests cannot be decoded (independent of this PR, strictly speaking)

### Issues:
- fixes #6752 

------


frcroth:
@philippotto Backend should be ready (though not completely tested)

webknossos-datastore/app/com/scalableminds/webknossos/datastore/services/MeshFileService.scala
```diff
@@ -368,24 +377,46 @@ class MeshFileService @Inject()(config: DataStoreConfig)(implicit ec: ExecutionC
...
+      dataSorted = data.sortBy(d => d._3)
+      _ <- Fox.bool2Fox(data.map(d => d._2).toSet.size == 1) // Ensure same encoding for all responses
+      encoding = data.map(d => d._2).head
+      jumpTable = dataSorted
+        .map(d => d._1.length)
+        // The jump table contains n + 1 elements (that way, for each requested
+        // chunk i , jumpTable[i] and jumpTable[i + 1] define the byte range).
+        // The jumptable contains the offsets in bytes (encoded as Longs).
+        .scanLeft((8 * (data.length + 1)).toLong)((a, b) => a + b)
+        .flatMap(l => longToBytes(l))
```
philippotto, line 416: @frcroth I adapted the jump table so that its entries are in byte and not in elements (in other words, I added the `8 * ...` part in line 415). Probably the "magic" 8 can be expressed by retrieving the byte count of a `Long`? Could you tidy this up?
Also, I have another wish: Could you adapt the jump table so that its elements are not Longs, but instead unsigned 32-bit integers. 64-bit numbers are a bit tedious to handle in JS and since we won't request 4 GB mesh data in one request, 32-bit should be enough, I assume.

philippotto:
I just noticed that `THREE.DRACOLoader: Unexpected geometry type.` is thrown on the dev instance for some chunk (most of the mesh loads fine). This is already the case on master, too. See the testing steps in the PR description which trigger this.

The front-end doesn't handle this correctly currently (the spinner just keeps spinning). I'll add proper error handling and then we should be able to debug this further.

webknossos-datastore/app/com/scalableminds/webknossos/datastore/services/MeshFileService.scala
philippotto: @frcroth nevermind! the front-end can compute the jump table on its own, because it already knows the byteSizes of the individual chunks. I'll remove the jump table code from the back-end next week.

frontend/javascripts/libs/DRACOLoader.ts
```diff
@@ -1,7 +1,15 @@
 // @ts-nocheck
 /* eslint-disable */
-// Copied from node_modules/three/examples/jsm/loaders/DRACOLoader.js to fix ERR_REQUIRE_ESM error.
-import { BufferAttribute, BufferGeometry, FileLoader, Loader } from "three";
+// Copied from https://github.com/mrdoob/three.js/pull/25475 / DRACOLoader.js to fix ERR_REQUIRE_ESM error.
```
philippotto, line 3: copied from threejs

philippotto:
I think, the PR is ready for review. @daniel-wer It can make sense to review the PR by commit range (I did some refactoring of the isosurface saga were I simply extracted/moved some code). Also, I upgraded the draco libs.

@frcroth @fm3 See my previous comment regarding an error where a mesh chunk cannot be decoded. I added better error handling and reporting, but I couldn't really figure out what's happening. I only found one segment where one mesh request fails and that is:

```
Requested chunk: {position: [1024, 5120, 1024], byteOffset: 3920109208, byteSize: 1818}
Error: {type: 'error', id: 9592, error: 'THREE.DRACOLoader: Unexpected geometry type.'}
```

It happens with `l4dense_motta_et_al_dev_v2/segmentation/meshes/meshfile_mapping0.hdf5` with mesh id 355 and mapping 80. The mesh file is 13 GB big..

Maybe you have an idea what could be causing this. However, I think, it's not blocking this PR as this is already an issue on master.

philippotto:
Bug can be seen isolated on this dev instance: https://provokemeshdecodingerror.webknossos.xyz/datasets/sample_organization/l4dense_motta_et_al_dev_v2/view#2816,4353,1792,0,1.3 (enable mapping 80 and load a mesh).

webknossos-datastore/app/com/scalableminds/webknossos/datastore/services/MeshFileService.scala
```diff
@@ -368,24 +377,38 @@ class MeshFileService @Inject()(config: DataStoreConfig)(implicit ec: ExecutionC
...
+                  .uint8()
+                  .readArrayBlockWithOffset("neuroglancer",
+                                            meshChunkDataRequest.byteSize,
+                                            meshChunkDataRequest.byteOffset)
+              (data, meshFormat, requestAndIndex._2)
+            } ?~> "mesh.file.readData.failed"
+          })
+          .toList)
+      dataSorted = data.sortBy(d => d._3)
+      _ <- Fox.bool2Fox(data.map(d => d._2).toSet.size == 1) // Ensure same encoding for all responses
```
fm3, line 408: Bare assertions are hard to identify when something goes wrong in production. Please add an error message with `?~>`

webknossos-datastore/app/com/scalableminds/webknossos/datastore/services/MeshFileService.scala
```diff
@@ -368,24 +377,38 @@ class MeshFileService @Inject()(config: DataStoreConfig)(implicit ec: ExecutionC
...
+      encoding = data.map(d => d._2).head
```
fm3, line 409:
```suggestion
      encoding <- data.map(d => d._2).headOption
```
(will create a Fox.failure rather than an exception if something goes wrong)

webknossos-datastore/app/com/scalableminds/webknossos/datastore/services/MeshFileService.scala
```diff
@@ -368,24 +377,38 @@ class MeshFileService @Inject()(config: DataStoreConfig)(implicit ec: ExecutionC
...
-        cachedMeshFile.reader
-          .uint8()
-          .readArrayBlockWithOffset("neuroglancer", meshChunkDataRequest.byteSize, meshChunkDataRequest.byteOffset)
-      (data, meshFormat)
-    } ?~> "mesh.file.readData.failed"
-  }
+                      meshChunkDataRequests: MeshChunkDataRequestV3List,
+  ): Fox[(Array[Byte], String)] =
+    for {
+      data: List[(Array[Byte], String, Int)] <- Fox.combined(
```
fm3, line 383: Could you test if Fox.serialCombined slows this down significantly on your machine? The current implementation with Fox.combined and map may parallelize over all entries, which could actually be slower on a spinning disk.

fm3 (commented):
Works for me :) Great speed increase! I added a note on parallelization.

@philippotto could you write an issue for the draco decoding problem?

philippotto:
> @philippotto could you write an issue for the draco decoding problem?

Done! https://github.com/scalableminds/webknossos/issues/7044

webknossos-datastore/app/com/scalableminds/webknossos/datastore/services/MeshFileService.scala
fm3: Thanks for looking into this! Yes, I’d say for the moment we should optimize for spinning disk (which is also what the sorting-by-byte-offset step is for)

CHANGELOG.unreleased.md
```diff
@@ -14,6 +14,7 @@ For upgrade instructions, please check the [migration guide](MIGRATIONS.released
 - In addition to drag and drop, the selected tree(s) in the Skeleton tab can also be moved into another group by right-clicking the target group and selecting "Move selected tree(s) here". [#7005](https://github.com/scalableminds/webknossos/pull/7005)
 
 ### Changed
+- Loading of precomputed meshes got significantly faster (especially when using a mesh file for an oversegmentation with an applied agglomerate mappings). [#7001](https://github.com/scalableminds/webknossos/pull/7001)
```
daniel-wer, line 17:
```suggestion
- Loading of precomputed meshes got significantly faster (especially when using a mesh file for an oversegmentation with an applied agglomerate mapping). [#7001](https://github.com/scalableminds/webknossos/pull/7001)
```

frontend/javascripts/oxalis/model/sagas/isosurface_saga.ts
```diff
@@ -667,6 +663,7 @@ function* loadPrecomputedMeshForSegmentId(
     return;
   }
 
+  console.time("_getLoadChunksTasks");
```
daniel-wer, line 666: Can likely be removed before merging

daniel-wer (approved):
Works very well, awesome speed boost!

public/wasm/draco_decoder.wasm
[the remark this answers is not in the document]
philippotto: No, it was checked in and used :) However, its size increased from 275 KB to 280 KB which is why github puts the "+4.17 KB (100%)" label. I was confused by this, too, since the 100% suggests that it's completely new. However, it probably means that 100% of the changes are additions? :sweat_smile:

webknossos-datastore/app/com/scalableminds/webknossos/datastore/services/MeshFileService.scala
```diff
@@ -368,24 +377,34 @@ class MeshFileService @Inject()(config: DataStoreConfig)(implicit ec: ExecutionC
...
-          .uint8()
-          .readArrayBlockWithOffset("neuroglancer", meshChunkDataRequest.byteSize, meshChunkDataRequest.byteOffset)
-      (data, meshFormat)
-    } ?~> "mesh.file.readData.failed"
-  }
+                      meshChunkDataRequests: MeshChunkDataRequestV3List,
+  ): Fox[(Array[Byte], String)] =
+    for {
+      data: List[(Array[Byte], String, Int)] <- Fox.serialCombined(
+        meshChunkDataRequests.requests.zipWithIndex.sortBy(requestAndIndex => requestAndIndex._1.byteOffset).toList
```
fm3, line 384: could you add a comment that the sorting is done to optimize for spinning disk access?

public/wasm/draco_decoder.wasm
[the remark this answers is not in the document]
daniel-wer: Ah, yes that's slightly confusing :sweat_smile:

philippotto:
@frcroth just a small ping in case this PR slipped through your inbox :)

frcroth:
> @frcroth just a small ping in case this PR slipped through your inbox :)

@philippotto I'm a bit out of the loop on this one. What should I do?

webknossos-datastore/app/com/scalableminds/webknossos/datastore/services/MeshFileService.scala
frcroth: Don't think anything can go wrong because of the statument in the line above, but I guess it is better style

philippotto:
> > @frcroth just a small ping in case this PR slipped through your inbox :)
> 
> @philippotto I'm a bit out of the loop on this one. What should I do?

Only the PR feedback was missing, but I guess this is good now :+1: @fm3 Do you want to leave a final approval?

fm3 (approved):
Boom :rocket:

PR #7005, opened 2023-04-24
Context action to move tree to group

### LINKED ISSUE #6973: "Move selected trees/groups into this group" as context action
When having large tree hierarchies, drag-and-drop can get really clunky (backed by user feedback). Especially, scrolling while dragging is very tedious and easy to cause frustration.

<details>
  <summary>Initial cut/paste suggestion</summary>
I was thinking that a cut/paste workflow might be a good alternative. It would work like this:

- select the tree or group you want to move (multi-selection is also thinkable)
- ctrl+x or rightclick > cut
- select group into which the elements should be moved
- ctrl+v or rightclick > paste

Naming might need tuning to make it more intuitive. Also, the state in which trees are "in the clipboard" might be highlighted somehow (maybe similar to how multi-selection is conveyed).

Implementation-wise I think that we could do it entirely without the clipboard API, but maybe it has benefits to use the clipboard API? Maybe users expect to get the list of tree/group names when pasting into a text editor.

Alternatively, one could phrase it like this:
- cut == "Mark to move"
- paste == "Move marked trees/elements here"
</details>

See comment below for an easier way.

dieknolle3333:
### Steps to test:
- Go to any annotions view of a dataset and switch to the `Skeleton` tab
- With any tree active, right-click a group and make sure there is a context action called `Move active tree here`
- Repeat this step having selected multiple trees. Make sure the label of the context action is adjusted to "trees".
- Actually move a single and multiple trees to a group and see that it works.
- It is not yet possible to move a group to another group via context action. Thus the context action shouldnt appear if a group is active and you want to move it to another group.
- After having selected multiple trees, clear the selection. Right-click a group and make sure there is no context action called `Move active [group|trees?] here` without an active component.

### Issue:
- fixes #6973 

------


philippotto (changes requested):
Works great :tada: Code looks also good, but I left a couple of smaller suggestions.

Also, I noticed that the context menu doesn't close automatically. This is also the case for other entries, but I think, we should change that. Can you add

`this.handleTreeDropdownMenuVisibility(tree.treeId, false);`

in the context menu item handlers where it makes sense so that the menu closes after the action? for shuffle-tree-colors, the menu shouldn't close automatically, so that users can shuffle repeatedly until they are satisfied with the new color.

dieknolle3333:
> Works great tada Code looks also good, but I left a couple of smaller suggestions.
> 
> Also, I noticed that the context menu doesn't close automatically. This is also the case for other entries, but I think, we should change that. Can you add
> 
> `this.handleTreeDropdownMenuVisibility(tree.treeId, false);`
> 
> in the context menu item handlers where it makes sense so that the menu closes after the action? for shuffle-tree-colors, the menu shouldn't close automatically, so that users can shuffle repeatedly until they are satisfied with the new color.

Thank you for the quick review!

Regarding the automatic closing of the context menu, are you referring only to group context actions or to tree context actions aswell? For group context actions, I understand that I need to close the dropdown in cases where it's useful. 
For tree actions, the menu closes automatically if I delete the tree or hide/show other trees and remains open for all other actions afaict. Should this also be adjusted in any way?

frontend/javascripts/oxalis/view/right-border-tabs/tree_hierarchy_view.tsx
```diff
@@ -589,7 +605,10 @@ class TreeHierarchyView extends React.PureComponent<Props, State> {
             },
             {
               key: "measureSkeleton",
-              onClick: () => this.handleMeasureSkeletonLength(tree.treeId, tree.name),
+              onClick: () => {
+                this.handleMeasureSkeletonLength(tree.treeId, tree.name);
```
philippotto, line 609:
```suggestion
```

philippotto (changes requested):
> Regarding the automatic closing of the context menu, are you referring only to group context actions or to tree context actions aswell? For group context actions, I understand that I need to close the dropdown in cases where it's useful.
For tree actions, the menu closes automatically if I delete the tree or hide/show other trees and remains open for all other actions afaict. Should this also be adjusted in any way?
[Edit:] I will add it for tree actions aswell and see what you say :)

Yes, I think, your changes make sense :+1: The only thing is that this doesn't work well for "Change tree color", because that menu item opens a color picker and with your newest changes the color picker closes immediately after one has changed the color for the very first time. I added two code suggestion which revert that and then, the PR should be good to go :)

frontend/javascripts/oxalis/view/right-border-tabs/tree_hierarchy_view.tsx
```diff
@@ -589,7 +605,10 @@ class TreeHierarchyView extends React.PureComponent<Props, State> {
...
+                this.handleTreeDropdownMenuVisibility(tree.treeId, false);
```
philippotto, line 608:
```suggestion
```

frontend/javascripts/oxalis/view/right-border-tabs/tree_hierarchy_view.tsx
philippotto: Oops, I deleted the wrong line :see_no_evil: I simply wanted to remove the auto-close behavior for the color-changing menu.

dieknolle3333: okay, I did that!

PR #7007, opened 2023-04-24
Remove vault cache when reloading dataset

### LINKED ISSUE #6990: Reloading a dataset layer should cache-clear the corresponding data vaults
The vaultCache entries for a layer should be removed when the layer reload button is used.

frcroth:
- https://___.webknossos.xyz

### Steps to test:
- Reload layer, see log

### Issues:
- fixes #6990

------

### DIFF
--- a/webknossos-datastore/app/com/scalableminds/webknossos/datastore/controllers/DataSourceController.scala
+++ b/webknossos-datastore/app/com/scalableminds/webknossos/datastore/controllers/DataSourceController.scala
@@ -484,7 +484,7 @@ Expects:
           organizationName)
         for {
           _ <- Fox.runIf(layerName.isDefined)(
-            dataSourceRepository.invalidateVaultCache(organizationName, dataSetName, layerName.getOrElse("")))
+            dataSourceService.invalidateVaultCache(reloadedDataSource, layerName.getOrElse("")))
           _ <- dataSourceRepository.updateDataSource(reloadedDataSource)
         } yield Ok(Json.toJson(reloadedDataSource))
       }
--- a/webknossos-datastore/app/com/scalableminds/webknossos/datastore/controllers/DataSourceController.scala
+++ b/webknossos-datastore/app/com/scalableminds/webknossos/datastore/controllers/DataSourceController.scala
@@ -483,6 +483,8 @@ Expects:
           dataSourceService.dataBaseDir.resolve(organizationName).resolve(dataSetName),
           organizationName)
         for {
+          _ <- Fox.runIf(layerName.isDefined)(
+            dataSourceRepository.invalidateVaultCache(organizationName, dataSetName, layerName.getOrElse("")))
           _ <- dataSourceRepository.updateDataSource(reloadedDataSource)
         } yield Ok(Json.toJson(reloadedDataSource))
       }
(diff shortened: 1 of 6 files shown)


webknossos-datastore/app/com/scalableminds/webknossos/datastore/services/DataSourceService.scala
```diff
@@ -227,4 +232,22 @@ class DataSourceService @Inject()(
...
 
+  def invalidateVaultCache(dataSource: InboxDataSource, dataLayerName: String): Fox[Unit] =
+    for {
+      genericDataSource <- dataSource.toUsable
+      dataLayer <- genericDataSource.getDataLayer(dataLayerName)
+      magsOpt = dataLayer match {
+        case layer: N5Layer          => Some(layer.mags)
+        case layer: PrecomputedLayer => Some(layer.mags)
+        case _: SegmentationLayer    => None
+        case _: WKWLayer             => None
```
fm3, line 243: Is there a reason to have these here? Shouldn’t they match to the `_` below? Also, SegmentationLayers may be ZarrLayers at the same time, I think

webknossos-datastore/app/com/scalableminds/webknossos/datastore/services/DataSourceService.scala
```diff
@@ -227,4 +232,22 @@ class DataSourceService @Inject()(
...
+        case layer: ZarrLayer        => Some(layer.mags)
+        case _                       => None
+      }
+      _ = magsOpt match {
+        case Some(mags) => mags.map(mag => dataVaultService.clearVaultPath(mag))
```
fm3, line 248:
```suggestion
        case Some(mags) => mags.map(mag => dataVaultService.removeVaultFromCache(mag))
```

webknossos-datastore/app/com/scalableminds/webknossos/datastore/controllers/DataSourceController.scala
```diff
@@ -484,7 +484,7 @@ Expects:
           organizationName)
         for {
           _ <- Fox.runIf(layerName.isDefined)(
-            dataSourceRepository.invalidateVaultCache(organizationName, dataSetName, layerName.getOrElse("")))
+            dataSourceService.invalidateVaultCache(reloadedDataSource, layerName.getOrElse("")))
```
fm3, line 487:
```suggestion
            dataSourceService.invalidateVaultCache(reloadedDataSource, layerName))
```
Would it be possible to do all layers if no layer name is passed to the route?

fm3 (commented):
Thanks for having a look at this! I think the structure is good already :) Please see my two comments. Also: could this return the number of removed keys? Then we could add that to the logging in the controller.

webknossos-datastore/app/com/scalableminds/webknossos/datastore/services/DataSourceService.scala
frcroth: This was the automatic IntelliJ exhaustive match, will be adjusted.

webknossos-datastore/app/com/scalableminds/webknossos/datastore/controllers/DataSourceController.scala
frcroth: > Thanks for having a look at this! I think the structure is good already :) Please see my two comments. Also: could this return the number of removed keys? Then we could add that to the logging in the controller.

For the number of keys? Do we need to check if it exists before removing or would it be fine to assume it exists, so counting every remove call as one removed vault. In that path, the number will just be the number of mags

fm3: I guess the number of mags will do for the moment. I added a note to https://github.com/scalableminds/webknossos/issues/6969 that the cache API’s remove should return the number of removed items (0 or 1). If we add that there later we can make this more exact here

fm3 (approved):

PR #7017, opened 2023-04-25
Use zarr string fill values

### LINKED ISSUE #6989: Zarr data loading: double-check fill values
- [ ] double-check that fill values is used if a chunk does not exist/cannot be loaded
  - [ ] support string-encoded fill values (custom json reads?) NaN should become 0 in webknossos.

compare https://zarr.readthedocs.io/en/stable/spec/v2.html#fill-value-encoding

It may make sense to locally create a zarr array for testing

frcroth:
- https://___.webknossos.xyz

### Steps to test:
- Remove one chunk
- Request buckets with that chunk

### Issues:
- fixes #6989 

Not sure if converting infinity to max value is useful (and the other direction).

------


frcroth:
image
image

webknossos-datastore/app/com/scalableminds/webknossos/datastore/datareaders/ArrayDataType.scala
```diff
@@ -19,4 +19,32 @@ object ArrayDataType extends ExtendedEnumeration {
...
+      case ArrayDataType.u1 => Math.pow(2, 1 * 8).toLong - 1
+    }
+
+  def minValue(dataType: ArrayDataType): Number =
+    dataType match {
+      case ArrayDataType.f8 => Double.MinValue
+      case ArrayDataType.f4 => Float.MinValue
+      case ArrayDataType.i8 => Long.MinValue
+      case ArrayDataType.u8 => 0
+      case ArrayDataType.i4 => Int.MaxValue
```
fm3, line 43:
```suggestion
      case ArrayDataType.i4 => Int.MinValue
```

fm3 (commented):
Good stuff! Thanks for going through the data types! I guess for i4 it should also be MinValue :)

fm3 (approved):

Issue #6906, opened 2023-03-07
Error handling in DatasetArray: move away from Exceptions, use Fox/flatMap


fm3:
DataReader, and the jzarr code it arose from, still use exception handling and `@throws` annotations. Instead we should wrap exception-throwing stuff as far inside the reading process as we can, with tryos and map over the results, like everywhere else

PR #7019, opened 2023-04-25
Fix voxel offset for Neuroglancer Precomputed datasets

### LINKED ISSUE #7006: Neuroglancer Precomputed voxel_offset is not respected for sharded datasets
When loading Neuroglancer Precomputed datasets, the header can have a `voxel_offset`, compare  https://github.com/google/neuroglancer/blob/master/src/neuroglancer/datasource/precomputed/volume.md#info-json-file-specification

webknossos reads this voxel_offset, and holds it as `lazy val voxelOffset: Array[Int]`, but it is not used when loarding sharded datasets. Should be used when loading data etc.

One approach might also be to use the new transformations feature, compare #6748 (not sure if it is the easiest way, though)

fm3:
- use neuroglancer precomputed “voxel offset” field to set bounding box when exploring
- use the same voxel offset when loading data
- adapts computeChunkIndices to handle negative total offsets
- note that the offset in chunk can still become negative, but copyRange already handles that :sunglasses: 
- change some of the error handling from exceptions to Fox.failure
- catch exceptions during chunk decompression (related: #6906)

### Steps to test:
- Explore minnie dataset (segmentation and em, talk to us for dataset uri)
- Load data, should look plausible, layers should match

### Issues:
- fixes #7006 

------


webknossos-datastore/app/com/scalableminds/webknossos/datastore/datareaders/ChunkUtils.scala
```diff
@@ -7,35 +7,35 @@ object ChunkUtils extends LazyLogging {
...
-      start(dim) = startIdx
-      to(dim) = toIdx
-      val numChunksForDim = toIdx - startIdx + 1
+    for (dim <- 0 until nDims) {
+      val largestPossibleIndex = (arrayShape(dim) - 1) / arrayChunkSize(dim)
+      val smallestPossibleIndex = 0
+      val startIndexRaw = selectedOffset(dim) / arrayChunkSize(dim)
+      val startIndexClamped = Math.max(smallestPossibleIndex, Math.min(largestPossibleIndex, startIndexRaw))
+      val toIndexRaw = (selectedOffset(dim) + selectedShape(dim) - 1) / arrayChunkSize(dim)
+      val toIndexClampedToBbox = Math.max(smallestPossibleIndex, Math.min(largestPossibleIndex, toIndexRaw))
```
frcroth, line 20: Don't know if *to* matches *start*, perhaps use *end* in names here?

webknossos-datastore/app/com/scalableminds/webknossos/datastore/datareaders/ChunkUtils.scala
```diff
@@ -7,35 +7,35 @@ object ChunkUtils extends LazyLogging {
...
-    val currentIdx = start.clone
+    val currentIndex = start.clone
     for (i <- chunkIndices.indices) {
-      chunkIndices(i) = currentIdx.clone
-      var depthIdx = depth - 1
-      while ({ depthIdx >= 0 }) if (currentIdx(depthIdx) >= to(depthIdx)) {
-        currentIdx(depthIdx) = start(depthIdx)
-        depthIdx -= 1
+      chunkIndices(i) = currentIndex.clone
+      var dimIdx = nDims - 1
```
frcroth, line 32:
```suggestion
      var dimIndex = nDims - 1
```
Is also changed above, should be changed here as well

webknossos-datastore/app/com/scalableminds/webknossos/datastore/services/BinaryDataService.scala
```diff
@@ -91,13 +91,16 @@ class BinaryDataService(val dataBaseDir: Path,
           logger.warn(
             s"Caught internal error: $msg while loading a bucket for layer ${request.dataLayer.name} of dataset ${request.dataSource.id}")
           Fox.failure(e.getMessage)
+        case Failure(msg, _, _) =>
+          /*logger.warn(
+            s"Bucket loading for layer ${request.dataLayer.name} of dataset ${request.dataSource.id} at ${readInstruction.bucket} failed with error: $msg")*/
```
frcroth, line 96: Remove

frcroth (approved):

PR #7025, opened 2023-04-26
Fix download of public annotation, include access ctx in user cache key

### LINKED ISSUE #7016: Sometimes, api/annotation/download, called by libs, returns 404
…and an ambiguous “this resource could not be found”. We should find a way to reliably reproduce and debug this. Possibly with more error chaining in production.

fm3:
Fetching the user object to write the NML failed for logged-out users, since it was missing the GlobalAccessContext modifier. However, the user cache was independent of the access context, making this work sometimes nonetheless (except if the cache was expired). Instead, the user cache should take the access context as cache key, avoiding leaking user info in unexpected contexts.

### Steps to test:
- Set annotation and its dataset to public
- Download route should work even when logged out and no other requests have been sent in the past 3 minutes (use wget directly on `/api/annotations/:id/download`

### Issues:
- fixes #7016

------

### DIFF
--- a/app/controllers/AnnotationIOController.scala
+++ b/app/controllers/AnnotationIOController.scala
@@ -411,7 +411,7 @@ Expects:
           skeletonAnnotationLayer =>
             tracingStoreClient.getSkeletonTracing(skeletonAnnotationLayer, skeletonVersion)
         } ?~> "annotation.download.fetchSkeletonLayer.failed"
-        user <- userService.findOneCached(annotation._user) ?~> "annotation.download.findUser.failed"
+        user <- userService.findOneCached(annotation._user)(GlobalAccessContext) ?~> "annotation.download.findUser.failed"
         taskOpt <- Fox.runOptional(annotation._task)(taskDAO.findOne)
         nmlStream = nmlWriter.toNmlStream(
           fetchedSkeletonLayers ::: fetchedVolumeLayers,
--- a/app/models/user/UserService.scala
+++ b/app/models/user/UserService.scala
@@ -47,7 +47,7 @@ class UserService @Inject()(conf: WkConf,
   private lazy val Mailer =
     actorSystem.actorSelection("/user/mailActor")
 
-  private val userCache: AlfuFoxCache[ObjectId, User] =
+  private val userCache: AlfuFoxCache[(ObjectId, DBAccessContext), User] =
     AlfuFoxCache(timeToLive = conf.WebKnossos.Cache.User.timeout, timeToIdle = conf.WebKnossos.Cache.User.timeout)
 
   def userFromMultiUserEmail(email: String)(implicit ctx: DBAccessContext): Fox[User] =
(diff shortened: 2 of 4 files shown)


app/models/user/UserService.scala
```diff
@@ -249,12 +257,12 @@ class UserService @Inject()(conf: WkConf,
...
   def updateLastTaskTypeId(user: User, lastTaskTypeId: Option[String])(implicit ctx: DBAccessContext): Fox[Unit] =
     userDAO.updateLastTaskTypeId(user._id, lastTaskTypeId).map { result =>
-      userCache.remove(user._id)
+      removeUserFromCache(user._id)
       result
     }
 
   def retrieve(loginInfo: LoginInfo): Future[Option[User]] =
-    userDAO.findOne(ObjectId(loginInfo.providerKey))(GlobalAccessContext).futureBox.map(_.toOption)
+    findOneCached(ObjectId(loginInfo.providerKey))(GlobalAccessContext).futureBox.map(_.toOption)
```
fm3, line 265: I noticed that this didn’t use the cache. I think it is safe for it to do so now. This method is used by silhouette when authenticating requests. This change may save us quite a few database queries

fm3:
note to self: change cache key from whole access context to its toStringAnonymous representation to avoid cache invalidation when the requesting user is mutated

app/controllers/AnnotationIOController.scala
```diff
@@ -411,7 +411,7 @@ Expects:
           skeletonAnnotationLayer =>
             tracingStoreClient.getSkeletonTracing(skeletonAnnotationLayer, skeletonVersion)
         } ?~> "annotation.download.fetchSkeletonLayer.failed"
-        user <- userService.findOneCached(annotation._user) ?~> "annotation.download.findUser.failed"
+        user <- userService.findOneCached(annotation._user)(GlobalAccessContext) ?~> "annotation.download.findUser.failed"
```
philippotto, line 414: So, this failed before (ignoring the caching issue) because the requesting user didn't have access to the owning user?
Does this mean that an annotation owned by user A can be downloaded by user B (if B has access to it) even though user B might not be allowed to "read" user A? Because user A needs to be serialized into the annotation?
I'm wondering whether this circumvents the permission system a bit?

fm3: Yes, that’s true. However, not all info about user A is serialized into the annotation. I believe we had a discussion about this some time ago and decided that it would be fair to say that a user publishing an annotation publishes their name with it. I’m open to having that discussion again, though. Also it might be worth double-checking what info is really needed for wk to show an annotation. The download only includes the name, but i think the annotation info request (which already works in the same way before this PR) also includes the email address.

philippotto: Ok, fair enough :+1:

> Also it might be worth double-checking what info is really needed for wk to show an annotation.

Yes, this is a good idea. One could also put in a dummy email value à la `anonymous`.

fm3: https://github.com/scalableminds/webknossos/issues/7029

philippotto (approved):
Works great :+1:

PR #7043, opened 2023-05-02
Team edit modal

### LINKED ISSUE #7035: Allow to add/remove users in teams view
As a follow-up to #6913, the teams view can be further extended:

- show all users for a team in a modal
- allow to remove users from that list
- allow to add users to that list

The UI could be similar to the slack UI for editing channels.

dieknolle3333:
### Steps to test:
- Go to teams list 
- Click on `Add/Remove Users`
- Wait for modal to load (demo only)
- click into modal
- remove and add users
- make sure changes are updating the next time the dropdown is opening

### Issues:
- fixes #7035 

------
~~-[ ] Updated [documentation](../blob/master/docs) if applicable~~


frontend/javascripts/admin/team/edit_team_modal_view.tsx
```diff
@@ -0,0 +1,145 @@
...
+  isOpen: boolean;
+  team: APITeam | null;
+};
+
+function EditTeamModalForm({ onCancel, isOpen, team }: Props) {
+  const [autoCompleteValue, setAutoCompleteValue] = useState("");
+  const onChange = (newValue: string) => setAutoCompleteValue(newValue);
+  const [users, setUsers] = useState<APIUser[] | null>(null);
+  const fetchUsers = async () => setTimeout(async () => setUsers(await getEditableUsers()), 5000); // TODO delete; only testing purposes
+  //const fetchUsers = async () => setUsers(await getEditableUsers());
```
dieknolle3333, line 21: :dodo: dont forget

frontend/javascripts/admin/team/edit_team_modal_view.tsx
```diff
@@ -0,0 +1,145 @@
...
+    const newUser = Object.assign({}, user, {
+      teams: newTeams,
+    });
+    const serverUser = await updateUser(newUser);
+    setUsers(users.map((oldUser) => (oldUser.id === serverUser.id ? serverUser : oldUser)));
+  };
+
+  const addTo = async (user: APIUser, team: APITeam | null) => {
+    if (team === null) return;
+    const newTeam: APITeamMembership = { id: team.id, name: team.name, isTeamManager: false }; //TODO never make manager?
```
dieknolle3333, line 38: :dodo: dont forget to delete

frontend/javascripts/admin/team/edit_team_modal_view.tsx
```diff
@@ -0,0 +1,145 @@
...
+    }
+    return (
+      <span onClick={() => removeFrom(user, team)}>
+        <MinusOutlined /> Remove from {team?.name}
+      </span>
+    );
+  };
+
+  const renderTeamMember = (user: APIUser): DefaultOptionType => ({
+    //TODO was unsure whether clicking on the name should also remove team member; same for renderUserNotInTeam
```
dieknolle3333, line 59: dont forget to delete

frontend/javascripts/admin/team/edit_team_modal_view.tsx
philippotto: Yes, I think, it's fair to add users as "normal" users without special permissions to a team. The user can use the users page to edit the roles for now.

frontend/javascripts/admin/team/edit_team_modal_view.tsx
philippotto: I think, clicking on the name itself (and not on Add/Remove) should ideally do nothing (not even closing the dropdown). However, I could imagine that the closing of the dropdown is hard to avoid. If closing the dropdown is inevitable, the add/remove action should happen in my opinion. Otherwise, the user would think that something happened (because the dropdown closed) but actually nothing was changed. Does this make sense?

dieknolle3333: yes, it does, thanks!

dieknolle3333: I think this could be done by changing the AutoComplete component to a Select component https://4x.ant.design/components/select/

frontend/javascripts/admin/team/edit_team_modal_view.tsx
```diff
@@ -0,0 +1,162 @@
...
+  const updateTeamMembership = async (user: APIUser, newTeams: APITeamMembership[]) => {
+    setPendingRequestsCounter(1);
+    setDropDownVisible(false);
+    if (users === null) return;
+    const newUser = Object.assign({}, user, {
+      teams: newTeams,
+    });
+    const serverUser = await updateUser(newUser);
+    setUsers(users.map((oldUser) => (oldUser.id === serverUser.id ? serverUser : oldUser)));
+    setTimeout(async () => await setPendingRequestsCounter(0), 2000); //TODO
```
dieknolle3333, line 37: delete me before merging but keep me while you might be testing

frontend/javascripts/admin/team/edit_team_modal_view.tsx
```diff
@@ -0,0 +1,162 @@
...
+  team: APITeam | null;
+};
+
+function EditTeamModalForm({ onCancel, isOpen, team }: Props) {
+  const [autoCompleteValue, setAutoCompleteValue] = useState("");
+  const [dropDownVisible, setDropDownVisible] = useState(false);
+  const onChange = (newValue: string) => setAutoCompleteValue(newValue);
+  const [users, setUsers] = useState<APIUser[] | null>(null);
+  const [pendingRequestsCounter, setPendingRequestsCounter] = useState(0);
+  const fetchUsers = async () => setTimeout(async () => setUsers(await getEditableUsers()), 2000); //TODO
```
dieknolle3333, line 22: delete me before merging but keep me while you might be testing

philippotto (commented):
excellent, it works really well :+1:

only one last UI suggestion: please add a small "minus" icon button next to each user in the the users list (not in the dropdown) so that one can also use that button to remove a user from a team (without having to use the dropdown).

philippotto (commented):
excellent, it works really well :+1:

only one last UI suggestion: please add a small "minus" icon button next to each user in the the users list (not in the dropdown) so that one can also use that button to remove a user from a team (without having to use the dropdown).

frontend/javascripts/admin/team/edit_team_modal_view.tsx
```diff
@@ -15,27 +15,31 @@ type Props = {
...
-  //const fetchUsers = async () => setUsers(await getEditableUsers());
+  const [pendingRequestsCounter, setPendingRequestsCounter] = useState(0);
+  const fetchUsers = async () => setTimeout(async () => setUsers(await getEditableUsers()), 2000); //TODO
   useEffect(() => {
     fetchUsers();
   }, []);
 
   if (team === null) return null;
   const updateTeamMembership = async (user: APIUser, newTeams: APITeamMembership[]) => {
+    setPendingRequestsCounter(1);
```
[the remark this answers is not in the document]
philippotto: > Initially I tried that, but it didn't work. The counter ended up being smaller than zero at the beginning of the second request, although it was zero after the first request.

Just fyi: incrementing/decrementing is a classic example for using a functional update. See the docs [here](https://legacy.reactjs.org/docs/hooks-reference.html#functional-updates). When not doing so, it can easily happen that the calculation is done on "outdated" values. This can happen because the callback that is executed is still seeing old values in its scope. The scope will only update on the next re-render, but an event handler can be kicked off by the re-rendering happens.

> Alright, I'll do that!

great :+1: the functional update won't be necessary then.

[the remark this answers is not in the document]
dieknolle3333: Very interesting, thank you for the hint!

frontend/javascripts/admin/team/team_list_view.tsx
```diff
@@ -49,6 +51,35 @@ export function renderTeamRolesAndPermissionsForUser(user: APIUser) {
...
+  return (
+    user.teams.some((userTeam: APITeamMembership) => userTeam.id === team.id) ||
+    (user.isAdmin && user.isActive)
+  );
+}
+
+export function renderUsersForTeam(
+  team: APITeam,
+  allUsers: APIUser[] | null,
+  renderAdditionalContent = Function(),
```
philippotto, line 64: The `Function()` part is quite unusual in JS. Usually, it would be used to create a function from a string. Also `new Function()` would be cleaner. To avoid `Function()`, I would suggest this:

```suggestion
  renderAdditionalContent = (_teamMember: APIUser, _team: APITeam) => {},
```

philippotto:
great! I got one last code suggestion and the todo/test comments/code can be removed now :)

frontend/javascripts/admin/team/team_list_view.tsx
dieknolle3333: Alright! This is a good reminder to double check what I'm reading on blogs :sweat_smile:

dieknolle3333: Why are you using the `_`syntax for `_teamMember` and `_team`?

philippotto (approved):
Beautiful :)

frontend/javascripts/admin/team/team_list_view.tsx
dieknolle3333: Makes sense, thanks!

PR #7050, opened 2023-05-04
Only reload affected areas when proofreading

### LINKED ISSUE #7022: During Proofreading, only reload "affected" buckets after split/merge
If segments A and B are merged or when segment C is split, we know that only the buckets will change that contain voxels A or B (for merges) or C (for splits). Therefore, only these buckets would need to be reloaded which can save quite a few resources.

The front-end could either (1) scan all buckets for the segment ids in question (if performance allows it) or (2) maintain a set of contained voxel ids per bucket.

(2) is probably preferrable, but let's test (1) first to check the impact and maybe the perf is fine.

philippotto:
Rough measurements show that ~80% of bucket reloads can be avoided that way for an average proofread operation.

- https://smartreload.webknossos.xyz

### Steps to test:
- open l4dense_motta_et_al_dev_v2
- select mapping 80
- do proofreading with
  - merge and split (without skeletons)
  - load agglomerate skeletons and split/merge/min-cut these
- only the affected areas should reload

### Issues:
- fixes #7022

------

### DIFF
--- a/frontend/javascripts/oxalis/model/bucket_data_handling/data_cube.ts
+++ b/frontend/javascripts/oxalis/model/bucket_data_handling/data_cube.ts
@@ -342,7 +342,6 @@ class DataCube {
       }
     }
 
-
     this.buckets = notCollectedBuckets;
     this.bucketIterator = notCollectedBuckets.length;
   }
--- a/frontend/javascripts/oxalis/model/bucket_data_handling/data_cube.ts
+++ b/frontend/javascripts/oxalis/model/bucket_data_handling/data_cube.ts
@@ -322,16 +322,26 @@ class DataCube {
     this.bucketIterator = (this.bucketIterator + 1) % (this.buckets.length + 1);
   }
 
-  collectAllBuckets(): void {
+  collectAllBuckets(predicateFn?: (bucket: DataBucket) => boolean): void {
+    predicateFn = predicateFn || (() => true);
     this.pullQueue.clear();
     this.pullQueue.abortRequests();
 
+    const notCollectedBuckets = [];
     for (const bucket of this.buckets) {
-      this.collectBucket(bucket);
+      // If a bucket is requested, collect it independent of the predicateFn,
+      // because the pullQueue was already cleared (meaning the bucket is in a
+      // requested state, but will never be filled with data).
+      if (bucket.state === "REQUESTED" || predicateFn(bucket)) {
+        this.collectBucket(bucket);
+      } else {
+        notCollectedBuckets.push(bucket);
+        console.log("skip because of", predicateFn);
+      }
     }
 
-    this.buckets = [];
-    this.bucketIterator = 0;
+    this.buckets = notCollectedBuckets;
+    this.bucketIterator = notCollectedBuckets.length;
   }
 
   collectBucket(bucket: DataBucket): void {
--- a/frontend/javascripts/oxalis/model/bucket_data_handling/data_cube.ts
+++ b/frontend/javascripts/oxalis/model/bucket_data_handling/data_cube.ts
@@ -322,23 +322,29 @@ class DataCube {
     this.bucketIterator = (this.bucketIterator + 1) % (this.buckets.length + 1);
   }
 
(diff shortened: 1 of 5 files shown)


frontend/javascripts/oxalis/model/bucket_data_handling/data_cube.ts
```diff
@@ -323,15 +323,31 @@ class DataCube {
...
+        this.collectBucket(bucket);
+        collectedCount++;
+      } else {
+        notCollectedBuckets.push(bucket);
+      }
     }
 
-    this.buckets = [];
-    this.bucketIterator = 0;
+    console.log(`collected: ${collectedCount} vs skipped ${notCollectedBuckets.length}`);
```
philippotto, line 347: todo: remove again

frontend/javascripts/oxalis/model/sagas/proofread_saga.ts
```diff
@@ -309,8 +309,9 @@ function* splitOrMergeOrMinCutAgglomerate(
   yield* call([Model, Model.ensureSavedState]);
 
   /* Reload the segmentation */
-
-  yield* call([api.data, api.data.reloadBuckets], layerName);
+  yield* call([api.data, api.data.reloadBuckets], layerName, (bucket) =>
+    bucket.containsValue(targetAgglomerateId),
```
philippotto, line 313: @fm3 Can you confirm that only voxels that have the the target agglomerate id will change after merge/split/min-cut? My tests confirmed this, but I want to doublecheck with you that this is really deterministic.

fm3: It is deterministic in the backend. I am not familiar with the term target agglomerate id.

Split only affects one “before” agglomerate so that should be safe. Min-Cut itself does not have side effects, instead, the front-end uses its return value to then issue a sequence of splits.

The merge udpate actions have agglomerateId1 and agglomerateId2. The segments that were agglomerateId2 will be mapped to agglomerateId1 instead. So if the frontend’s targetAgglomerateId is equal to the update action’s agglomerateId2, this should be safe.

philippotto: great, I think, everything should work then :+1:

daniel-wer (approved):
**Very** elegant solution! :star_struck:

I didn't encounter any issues during testing. I found it hard to assess how much of the latency is due to the added containment check - did you measure this somehow? In any case, I think it's definitely worth it to avoid reloading all the data! Much better UX.

philippotto:
>  did you measure this somehow?

Yes, I measured this (forgot to include it in the PR description). It takes about 1ms to compute the `Set`, which seems like a very good investment to me :)

frontend/javascripts/oxalis/model/bucket_data_handling/data_cube.ts
```diff
@@ -323,15 +323,31 @@ class DataCube {
...
 
   collectAllBuckets(): void {
+    this.collectBucketsIf(() => true);
+  }
+
+  collectBucketsIf(predicateFn: (bucket: DataBucket) => boolean): void {
     this.pullQueue.clear();
     this.pullQueue.abortRequests();
 
+    let collectedCount = 0;
```
philippotto, line 333:
```suggestion
```

frontend/javascripts/oxalis/model/bucket_data_handling/data_cube.ts
```diff
@@ -323,15 +323,31 @@ class DataCube {
...
+    const notCollectedBuckets = [];
     for (const bucket of this.buckets) {
-      this.collectBucket(bucket);
+      // If a bucket is requested, collect it independently of the predicateFn,
+      // because the pullQueue was already cleared (meaning the bucket is in a
+      // requested state, but will never be filled with data).
+      if (bucket.state === "REQUESTED" || predicateFn(bucket)) {
+        this.collectBucket(bucket);
+        collectedCount++;
```
philippotto, line 341:
```suggestion
```

PR #7053, opened 2023-05-04
Improve error logging for bucket requests

### LINKED ISSUE #7038: Allow toggling bucket request error logging for individual datasets
The datastore currently silences error logging for failed bucket requests, because it would generate a huge volume of logs. For some cases errors are expected.

Sometimes, though, we would need the error logging (incl. stacktrace) for failed bucket requests.

This should be done in BinaryDataService.handleBucketRequest.

There should be a way of turning this on for indivual datasets (possibly a route). Could expire after, say, 10 minutes.

`s"Bucket loading for layer ${request.dataLayer.name} of dataset ${request.dataSource.id} at ${readInstruction.bucket} failed with error: $msg")`

(add stack trace as well)

fm3:
- Changes the error handling in most of DatasetArray and the related classes to Fox instead of exceptions
 - Separates typing of loaded chunk bytes from creating chunks from fill value
 - Adds error logging in BinaryDataService for failed bucket requests. This logging is rate-limited, only 5 errors are logged per dataset until the next reset (interval of 60  minutes) or dataset reload/cacheclear (Note that some times a bit more logging will happen since the counting is not thread-safe, but that should be ok, and avoids slow synchronization)

### Steps to test:
- Add errors in chunk loading. I tested it by inserting `throw new Exception("blosc boom!")` in Compressor.scala L266 (in BloscCompressor.decompress) and then loading a local zarr dataset
- Load data, should see some readable error logging, but not thousands of lines
- Page reload should not lead to more error logging, only dataset cache clear (use the reload button in dataset layers tab of left sidebar)

### Issues:
- contributes to #6906 
- fixes #7038 (though I changed the logic a bit from the proposed idea there)

------


fm3:
Good catch on `wrapAndType`, I added two more tryos!

On the missing messages: Since this is never user facing, but only for our own debugging purposes, I’d say the short versions are sufficient and we need no further explanatory text. What do you think?

The best ticker interval is a good question. I’d argue that an hour is fair, to limit spam. If we want to debug a particular dataset more intensely, the cache is cleared by the reload button. Do you think there is a case where 60 minutes is too much and the reload button is no option?

PR #7057, opened 2023-05-08
Fix display of used storage in power plan

### LINKED ISSUE #6987: If the upper limit is infinity, the reported storage doesn't have a unit
## Context
- on wkorg (set isWkorgInstance=true in `application.conf`)
- organization needs to have the power pricing plan (can be changed with pgadmin for example, but should be the default locally)
- also see discussion in https://github.com/scalableminds/webknossos/pull/6893#issuecomment-1484796152

## Current behavior
The organization page (http://localhost:9000/organizations/sample_organization) shows `0.0/infinity` as the used storage even when multiple GB are in fact used. The reason for this is that apparently TB is chosen as a unit which is then rounded to 0.0.

image

## Expected behavior
There should be a (sensible) unit. In the above scenario I would expect something like `5.5 / infinity GB` or maybe `5.5 GB / infinity`?

## Hints for developing
The used storage can also be changed in the organizations sql table instead of trying to upload many datasets.

dieknolle3333:
### Steps to test:
- Having uploaded some datasets, set `datastore.reportUsedStorage.enabled=true` in `application.conf`
- If needed, edit organization_usedstorage.usedstoragebytes in pgAdmin
- Make sure you're subscribed to any but the Basic Plan (e.g. Custom)
- Go to Administration>Organization page and see whether the used storage is displayed adequately
- Up to 0.1TB of used storage, it should be displayed in GB, otherwise in TB

### Issues:
- fixes #6987 

------
~~- [ ] Updated [migration guide](../blob/master/MIGRATIONS.unreleased.md#unreleased) if applicable~~
~~- [ ] Updated [documentation](../blob/master/docs) if applicable~~


frontend/javascripts/admin/organization/organization_cards.tsx
```diff
@@ -194,9 +194,9 @@ export function PlanDashboardCard({
     organization.includedStorageBytes === Number.POSITIVE_INFINITY ? "∞" : includedStorageLabel;
 
   const usedStorageLabel =
-    organization.pricingPlan === PricingPlanEnum.Basic
-      ? `${(organization.usedStorageBytes / 10 ** 9).toFixed(1)}`
-      : `${(organization.usedStorageBytes / 10 ** 12).toFixed(1)}`;
+    organization.pricingPlan === PricingPlanEnum.Basic || organization.usedStorageBytes < 10 ** 11 //=0.1TB
```
philippotto, line 197: While reading this code, I remembered that we already have a utility function for formatting bytes. I think, this would simplify the existing code quite a bit. I imagine something like this:

```ts
  const includedStorageLabel =
    organization.includedStorageBytes === Number.POSITIVE_INFINITY
      ? "∞"
      : formatCountToDataAmountUnit(organization.includedStorageBytes);
  const usedStorageLabel = formatCountToDataAmountUnit(organization.usedStorageBytes);
```

That way, one doesn't need to tie the formatting logic to the organization plan, either.

dieknolle3333: Oh, nice!
During testing I noticed that the formatting is different than I would have implemented it. In the case of this picture, I would have rendered "0.1TB/5TB" or something, because there would be no break in `GB`. But I think it's still readable, so should I just leave it at your suggested code?
Screenshot from 2023-05-09 19-05-16

PR #7063, opened 2023-05-09
Segment to Bucket Index for Volume Annotations

### LINKED ISSUE #7033: Segments-to-chunk index
## Detailed Description
Build an index structure so that we know where (in what chunks) segments are in an annotation. This index should be versioned and maintained as part of a volume annotation.
A second step would be to use segment statistics from voxelytics (eg generated upon import) as source of information for fallback layers.

## Context
This is an important piece of infrastructure for segment statistics, improved adhoc meshing, fill tool and other features.

fm3:
Creates an index structure for volume annotations. For every segment id, the list of buckets that contain this segment is stored. This is stored separately in the fossildb and updated on volume annotation mutation.

This lays the basis for new routes calculating properties of the segment volumes, e.g. the volume in voxels. A new route already provides that one. More complex metrics (surface area, sphericity, connected components) may be added later

### TODO
 - [x] create data structure (proto vec3int list)
 - [x] add column family, put and get in fossildb
 - [x] on bucket save, get id set
 - [x] on bucket save, load previous bucket (either from fallback layer or volume), get id set
 - [x] get segment index for affected segments, mutate it, write back
 - [x] add route to compute total segment volume
 - [x] duplicate should update this index too
 - [x] downsampling should update this index too
 - [x] revert to version should update this index too (?)
 - [x] how to handle legacy volume annotations with no segment index? hasSegmentIndex → set to true in duplicate (if no fallback layer)
 - [x] perf optimization: skip previous bucket check for “duplicate”
 - [x] perf optimization: process multiple buckets at once (use local buffer + flush)
 - [x] perf optimization: avoid UnsignedInteger conversion of every voxel
    - [x] for uint32
    - [x] for the other types (ShortStream, ByteStream are not available)
 - [x] remove logging, clean up code
 - [x] merging volume tracings should update this index too
 - [x] add some explaining comments (e.g. to buffer)
 - [x] test some more (including merge)
 - [x] CI

### Follow-up
 - get fallback layer segment index if available (as computed by future versions of voxelytics?) #7132
 - perf optimization: fossildb multiget/multiput (with transaction management?), compare https://github.com/scalableminds/fossildb/issues/39

### Steps to test:
- Create volume annotation without fallback layer
- brush some, with several ids
- go to route `/tracings/volume/<tracingId>/segmentStatistics/volume?segmentId=1&mag=1&token=<token>`
- should get segment volume in vx
- merging annotations should still work (and create merged segment index if all source tracings have segment index)
- saving volume updates should not be unbearably slow (slower than before, though. I had about 100ms for 32 buckets, which is about thrice the old value)

### Issues:
- fixes #7033

------

### DIFF
--- a/CHANGELOG.unreleased.md
+++ b/CHANGELOG.unreleased.md
@@ -25,6 +25,7 @@ For upgrade instructions, please check the [migration guide](MIGRATIONS.released
 - Fixed problems which could arise when annotating volume data at negative positions (which is not supported and is properly ignored now). [#7124](https://github.com/scalableminds/webknossos/pull/7124)
 - Fixed some requests failing for streaming remote data via HTTP, which was observed when streaming data via Zarr from another WEBKNOSSOS instance. [#7125](https://github.com/scalableminds/webknossos/pull/7125)
 - Fixed that the brush preview was inaccurate in some scenarios. [#7129](https://github.com/scalableminds/webknossos/pull/7129)
+- Fixed order of decompression on neuroglancer precomputed datasets, which caused some segmentation layers to not load correctly. [#7135](https://github.com/scalableminds/webknossos/pull/7135/)
 
 ### Removed
 - Support for [webknososs-connect](https://github.com/scalableminds/webknossos-connect) data store servers has been removed. Use the "Add Remote Dataset" functionality instead. [#7031](https://github.com/scalableminds/webknossos/pull/7031)
--- a/webknossos-datastore/app/com/scalableminds/webknossos/datastore/datareaders/precomputed/PrecomputedCompressorFactory.scala
+++ b/webknossos-datastore/app/com/scalableminds/webknossos/datastore/datareaders/precomputed/PrecomputedCompressorFactory.scala
@@ -21,9 +21,9 @@ object PrecomputedCompressorFactory {
         getCompressorForShardingChunks(shardingSpecification.data_encoding)
       case None => nullCompressor
     }
-    val outerCompression = getCompressorForEncoding(header)
+    val chunkCompression = getCompressorForEncoding(header)
 
-    new ChainedCompressor(Seq(shardingCompression, outerCompression))
+    new ChainedCompressor(Seq(chunkCompression, shardingCompression))
 
   }
 


webknossos-tracingstore/app/com/scalableminds/webknossos/tracingstore/tracings/volume/VolumeSegmentIndexService.scala
```diff
@@ -0,0 +1,135 @@
...
+      implicit ec: ExecutionContext): Fox[Set[Long]] =
+    for {
+      set <- tryo(UnsignedIntegerArray.toSetFromByteArray(bytes, elementClass)).toFox
+    } yield
+      set.filter(!_.isZero).map { u: UnsignedInteger =>
+        u.toPositiveLong
+      }
+
+  private def bucketPositionVec3IntProto(bucketPosition: BucketPosition) =
+    Vec3IntProto(bucketPosition.bucketX, bucketPosition.bucketY, bucketPosition.bucketZ)
```
frcroth, line 98: Move to BucketPosition?

app/models/annotation/nml/NmlParser.scala
```diff
@@ -90,7 +90,8 @@ object NmlParser extends LazyLogging with ProtoGeometryImplicits with ColorGener
                 userBoundingBoxes,
                 organizationName,
                 segments = v.segments,
-                segmentGroups = v.segmentGroups
+                segmentGroups = v.segmentGroups,
+                hasSegmentIndex = Some(v.fallbackLayerName.isEmpty)
```
frcroth, line 94: Seems like dangerous code duplication. Why not use the property of the tracing?

webknossos-tracingstore/app/com/scalableminds/webknossos/tracingstore/tracings/volume/VolumeSegmentIndexBuffer.scala
```diff
@@ -0,0 +1,50 @@
...
+trait SegmentIndexKeyHelper {
+  protected def segmentIndexKey(tracingId: String, segmentId: Long, mag: Vec3Int) =
+    s"$tracingId/$segmentId/${mag.toMagLiteral()}"
+}
+
+// To introduce buffering for updating the segment-to-bucket index for a volume tracing
+// read provides fallback data from fossildb
+// while write is done only locally in-memory, until flush is called
+// This saves a lot of db interactions (since adjacent bucket updates usually touch the same segments)
+class VolumeSegmentIndexBuffer(tracingId: String, volumeSegmentIndexClient: FossilDBClient, newVersion: Long)
```
frcroth, line 21:
```suggestion
class VolumeSegmentIndexBuffer(tracingId: String, volumeSegmentIndexClient: FossilDBClient, version: Long)
```

webknossos-tracingstore/app/com/scalableminds/webknossos/tracingstore/tracings/volume/VolumeSegmentIndexService.scala
```diff
@@ -0,0 +1,135 @@
...
+  def canHaveSegmentIndex(tracing: VolumeTracing): Option[Boolean] = Some(tracing.fallbackLayer.isEmpty)
+
+  def updateFromBucket(segmentIndexBuffer: VolumeSegmentIndexBuffer,
+                       bucketPosition: BucketPosition,
+                       bucketBytes: Array[Byte],
+                       previousBucketBytesBox: Box[Array[Byte]],
+                       elementClass: ElementClassProto)(implicit ec: ExecutionContext): Fox[Unit] =
+    for {
+      bucketBytesDecompressed <- tryo(
+        decompressIfNeeded(bucketBytes, expectedUncompressedBucketSizeFor(elementClass), "")).toFox
```
frcroth, line 46: Remove tryo/create new tryo fox function?

util/src/main/scala/com/scalableminds/util/tools/Fox.scala
```diff
@@ -134,9 +134,25 @@ object Fox extends FoxImplicits {
...
+        for {
+          currentResult <- f(it.next())
+          results <- runNext(currentResult :: results)
+        } yield results
+      } else {
+        Fox.successful(results.reverse)
+      }
+
+    runNext(Nil)
+  }
```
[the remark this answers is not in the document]
fm3: Turned the List-taking one into `serialCombined(l.toIterator)(f)`

webknossos-tracingstore/app/com/scalableminds/webknossos/tracingstore/tracings/volume/VolumeSegmentIndexService.scala
fm3: Wrote a new issue https://github.com/scalableminds/webknossos/issues/7133 for this

frcroth (approved):

PR #7064, opened 2023-05-09
Min length for layer names is one

### LINKED ISSUE #7020: Layer name length should not be restricted to minimum 3 characters
“em” is a valid name. Even single-char layers should be allowed.

This is enforced in the dataset settings view, as well as the dataset import view (after explore/add remote).

dieknolle3333:
### Steps to test:
- Go to dataset list view > Settings  and change the dataset name in the form to less then 3 characters. There should be a form error. 
- In the same form, edit a layer name. You should be able to set a single character name.
- Add a remote dataset. Repeat steps 1 and 2 in the second form (after clicking `Add Layer`).

### Issues:
- fixes #7020

------
~~- [ ] Updated [migration guide](../blob/master/MIGRATIONS.unreleased.md#unreleased) if applicable~~
~~- [ ] Updated [documentation](../blob/master/docs) if applicable~~


philippotto:
great! :+1: 

@fm3 do you know whether the back-end also has a check for this? At least, I'm vaguely remembering this..

Also I'm wondering whether we should soften the requirement for the dataset name, too :thinking: Any thoughts, @fm3?

fm3:
The back-end does not check layer length names. [It has a couple checks, including unique names](https://github.com/scalableminds/webknossos/blob/f0208c68c0c991f6895b66593e40120316b0514c/webknossos-datastore/app/com/scalableminds/webknossos/datastore/services/DataSourceService.scala#L137), but layer name length is not among them.

As Norman stated in https://github.com/scalableminds/webknossos/issues/7020#issuecomment-1540216371, the dataset name restriction should stay in place.

philippotto (approved):
Great! :+1: 

Thanks for the clarifictions, @fm3 :pray:

PR #7066, opened 2023-05-10
Fix local to global layer index look up

philippotto:
- https://___.webknossos.xyz

### Steps to test:
- check that datasets with heterogenous mags render correctly
- for the dataset in question (see slack thread), I checked that both layer orderings render okay

### Issues:
- fixes rendering bug introduced by #6748 
- also see https://scm.slack.com/archives/C5AKLAV0B/p1683617650361979

------


frontend/javascripts/admin/admin_rest_api.ts
```diff
@@ -1375,14 +1375,18 @@ export async function getActiveDatasetsOfMyOrganization(): Promise<Array<APIData
...
 ): Promise<APIDataset> {
   const sharingTokenSuffix = sharingToken != null ? `?sharingToken=${sharingToken}` : "";
-  return Request.receiveJSON(
+  const ds = await Request.receiveJSON(
     `/api/datasets/${datasetId.owningOrganization}/${datasetId.name}${sharingTokenSuffix}`,
   );
+
+  // ds.dataSource.dataLayers = [ds.dataSource.dataLayers[1], ds.dataSource.dataLayers[0]];
+
+  return ds;
```
philippotto, line 1389: revert before merging. this is only to highlight how I tested this.

frontend/javascripts/oxalis/model/bucket_data_handling/layer_rendering_manager.ts
```diff
@@ -103,8 +103,9 @@ export function getGlobalLayerIndexForLayerName(
 ): number {
   const sanitizer = optSanitizer || _.identity;
   const dataset = Store.getState().dataset;
-  const allLayers = [...getColorLayers(dataset), ...getSegmentationLayers(dataset)];
-  const layerIndex = allLayers.findIndex((layer) => sanitizer(layer.name) === layerName);
```
philippotto, line 107: this old code is similar to how the shader code constructs the layer order. however, we need the indices from `dataset.dataSource.dataLayers` because that ordering is used via `Object.values` in multiple places to set up uniforms (e.g., for passing the active mag indices and the resolutions of the layers). for datasets where all layers are homogenous (or where the ordering matched by chance), this was not an issue.

philippotto (commented):
@fm3 I think it would be cool if the back-end would send the layer order deterministically. what do you think? in the end, it shouldn't matter much, but reducing randomness sounds like a good idea to me to reduce future debugging headaches.

fm3:
Sounds fair. Do you think `order by name` is fine? The order might then change when the properties.json is edited but I guess that’s rare. Propagating the exact order from the properties.json is not straightforward, since the index is not stored in postgres.

philippotto:
Yes, sorting by name should be enough. It doesn't need to be stable across DS properties edits. Should just be stable across different instances or reloads :)

normanrz:
> Yes, sorting by name should be enough.

I think layers should be ordered as they are in the datasource-properties.json. Eventually, they should be user-sortable.

daniel-wer (approved):
Code LGTM, I didn't test again.

Issue #6234, opened 2022-05-23
In scala backend, get rid of null entirely, use scala Options instead.


fm3:
There are some usages of null, e.g. in IsosurfaceServiceHolder and VolumeTracingService. This is a code smell, should be replaced by Scala Option/None for type safety.

fm3:
I found some in
 - LRUConcurrentCache
 - ImageCreator
 - ZipIO
 - JsonHelper
 - TextUtils
 - IntervalScheduler
 - DataServiceDataRequest
 - IsosurfaceServiceHolder
 - mappingNameForMeshFile

PR #7067, opened 2023-05-10
Shared Chunk Cache for all DatasetArrays, CacheWeight for AlfuCache

### LINKED ISSUE #6630: Share chunkContents cache across multiple arrays

fm3:
- The cache for remote dataset array contents can now have a configured size in bytes. New config option `datastore.cache.imageArrayChunks.maxSizeBytes`. Default is 2 GB, consider increasing for production.
- For this, one cache object is shared between the dataset arrays
- Switches from akka cache to Caffeine directly (akka uses caffeine internally as well), allowing to add cache item weighing (weight is now set to bytes per chunk for array chunks)
- Unified some caching usages / naming (no more distinction between AlfuCache and AlfuFoxCache)

### TODO
 - [x] use shared cache also in Precomputed + Neuroglacner classes
 - [x] include vault path in cache key
 - [x] how to clear cache for individual datasets?
 - [x] null pointer exception for loading volume annotation buckets

- https://cacheweight.webknossos.xyz

### Steps to test:
- Load some datasets, should still show
- annotate some, should still show
- use reload button and look for logging (cache chunks should be removed, but only for the reloaded layer)

### Issues:
- fixes #6630 
- contributes to #6969 

------


util/src/main/scala/com/scalableminds/util/cache/LRUConcurrentCache.scala
```diff
@@ -1,11 +1,12 @@
 package com.scalableminds.util.cache
 
+import java.util
 import scala.collection.JavaConverters._
 
 trait LRUConcurrentCache[K, V] {
   def maxEntries: Int
 
-  private val cache = new java.util.LinkedHashMap[K, V]() {
+  private val cache: util.LinkedHashMap[K, V] = new java.util.LinkedHashMap[K, V]() {
```
frcroth, line 9:
```suggestion
  private val cache: util.LinkedHashMap[K, V] = new util.LinkedHashMap[K, V]() {
```

webknossos-datastore/app/com/scalableminds/webknossos/datastore/services/BinaryDataService.scala
```diff
@@ -84,8 +87,10 @@ class BinaryDataService(val dataBaseDir: Path,
     if (request.dataLayer.doesContainBucket(bucket) && request.dataLayer.containsResolution(bucket.mag)) {
       val readInstruction =
         DataReadInstruction(dataBaseDir, request.dataSource, request.dataLayer, bucket, request.settings.version)
-      val bucketProvider = bucketProviderCache.getOrLoadAndPut(request.dataLayer)(dataLayer =>
-        dataLayer.bucketProvider(dataVaultServiceOpt))
+      val dataSourceId = if (request.dataSource != null) request.dataSource.id else DataSourceId("", "")
```
frcroth, line 90: When is this the case and should it lead to something else, since this will not use the right cache?

webknossos-datastore/app/com/scalableminds/webknossos/datastore/services/BinaryDataService.scala
```diff
@@ -20,6 +22,7 @@ class BinaryDataService(val dataBaseDir: Path,
                         val agglomerateServiceOpt: Option[AgglomerateService],
                         dataVaultServiceOpt: Option[DataVaultService],
                         val applicationHealthService: Option[ApplicationHealthService],
+                        sharedChunkContentsCache: Option[AlfuCache[String, MultiArray]],
                         datasetErrorLoggingService: Option[DatasetErrorLoggingService])
     extends FoxImplicits
     with DataSetDeleter
```
frcroth, line 28: Doesn't have to be this PR, but I noticed that you renamed DataSet stuff to Dataset. Should be done everywhere, e.g. here.

webknossos-datastore/app/com/scalableminds/webknossos/datastore/services/BinaryDataService.scala
```diff
@@ -199,19 +204,34 @@ class BinaryDataService(val dataBaseDir: Path,
...
+
     val closedAgglomerateFileHandleCount =
       agglomerateServiceOpt.map(_.agglomerateFileCache.clear(agglomerateFileMatchPredicate)).getOrElse(0)
+
     val closedDataCubeHandleCount = shardHandleCache.clear(dataCubeMatchPredicate)
-    (closedAgglomerateFileHandleCount, closedDataCubeHandleCount)
+
+    bucketProviderCache.clear(bucketProviderPredicate)
+
+    def chunkContentsPredicate(key: String): Boolean =
```
frcroth, line 228: Not totally happy with this being placed here, since the definition of the key is somewhere else. Maybe move this predicate to where the key is defined and only use it here?

frcroth (commented):
Maybe also change PR name?

webknossos-datastore/app/com/scalableminds/webknossos/datastore/services/BinaryDataService.scala
fm3: I added an explaining comment “dataSource is null and unused for volume tracings. Insert dummy DataSourceId (also unused in that case)”

Related: https://github.com/scalableminds/webknossos/issues/6234

webknossos-datastore/app/com/scalableminds/webknossos/datastore/services/BinaryDataService.scala
fm3: Yes, that’s follow-up for https://github.com/scalableminds/webknossos/issues/6053 – trouble is, we can’t just replace everything, because some parts are serialized and/or used elsewhere (e.g. routes). But for local variables, I intend to use dataset in the future as discussed in the linked issue

webknossos-datastore/app/com/scalableminds/webknossos/datastore/services/BinaryDataService.scala
fm3: Hmm, yes I see what you mean. However, the same would be the case for the other two predicate functions here. So we could either extract each of those so that there is only one place for each cache, where cache key building happens. This would lead to less distributed knowledge on how the keys work. However, I also like all of these methods being together here, since it is less distributed knowledge about the fact that these caches should be cleared dataset-wise or layer-wise. So it’s some fragmentation either way. If you’re ok with it, I’d leave it as it is for the moment.

frcroth (approved):

PR #7068, opened 2023-05-10
Combine both download modals into one

MichaelBuessemeyer:
This pr combines the two download modals by merging the python modal that was used in the dataset view mode. The only python modal was deleted afterward.
The download modal now supports tiff export in view mode but not wkw download because there is no annotation to download from view mode. 

For the review, I suggest turning off showing changes in whitespace. This will drastically decrease the number of changed lines.

- https://___.webknossos.xyz

### Steps to test:
- run `yarn enable-jobs` and start a local wk worker. ( <-- needed to test whether the tiff export works)
- choose a data and open it in view mode
- open the download modal. The modal should contain the tiff and the python download tab.
- The tiff export should work for the whole layer (as long as it is not too large). As a result, a (ome) tiff file should be downloaded after some time
- The python tab should display the python code that downloads the dataset and not the code that tries downloading an annotation
- Check what happens when the layer is too large: Go to `download_modal_view.tsx` and decrease the max limit 112 to something small by e.g. replacing the value that is compared to the variable `volume` with `1000`. Reloading and reopening the dataset should show the tiff export modal tab but the download should be disabled, as the layer is too large.
- Undo the changes from the previous step and create an annotation for that dataset
- Open the download modal and try out the wkw download and the tiff export. Both should work. 
- Last check that the python tab includes the code that downloads the annotation and not tries to download the dataset (compared to the modal in view mode)

### Issues:
- fixes #

------


daniel-wer (commented):
Well done! Code LGTM, I'll report back tomorrow after testing :)

daniel-wer (commented):
The general functionality worked very well, I also successfully tested the export :+1: 

- [ ] I noticed some issues where "annotation" appears in the "Download Dataset" modal that should be fixed:
annotation_mentions_1
annotation_mentions_2
annotation_mentions_3
- [x] In the annotation download code, the indentation depth seems wrong for some lines
wrong_indentation

frontend/javascripts/oxalis/view/action-bar/download_modal_view.tsx
```diff
@@ -507,7 +506,7 @@ function _DownloadModalView({
               >
                 {!hasVolumes ? "This is a Skeleton-only annotation. " : ""}
                 {!hasSkeleton ? "This is a Volume-only annotation. " : ""}
-                {messages["annotation.download"]}
+                {messages["download.data_is_ready_for_download"]({ typeName })}
```
daniel-wer, line 509: This tab is only ever shown for annotations, so the distinction is not needed here. The resulting text "The following **dataset data** is available for download immediately." would sound weird, too, so I'd revert that change.

daniel-wer (approved):
LGTM, only one small suggestion :)

frontend/javascripts/oxalis/view/action-bar/download_modal_view.tsx
MichaelBuessemeyer: Thanks for :cat:-ching that. I revered the change :recycle:

Issue #7069, opened 2023-05-11
Allow to identify large artifacts in voxelytics reports


daniel-wer:
## Detailed Description

The Voxelytics Report UI should be improved to easily show the size of all artifacts of a workflow run, sortable by size, possibly highlighting old versions.
The workflow list should also include the aggregated size of the artifacts.
Maybe, there could be a view showing all artifacts in a specific artifact environment.

## Context

The larger the datasets that are processed with voxelytics, the more important it becomes to archive unused or old artifacts in order to regain free disk space. The voxelytics reports (list) serves as an information hub for voxelytics runs and should help users to identify large (and possibly unused) artifacts created with voxelytics.

PR #7071, opened 2023-05-12
adds fileSize to voxelytics workflow list

normanrz:
### Steps to test:
- Run a vx report and look at the workflow list
Screenshot 2023-05-12 at 10 29 04

### Issues:
- contributes to #7069 

------


frontend/javascripts/admin/voxelytics/workflow_list_view.tsx
```diff
@@ -181,6 +181,25 @@ export default function WorkflowListView() {
...
           },
+          {
+            title: "File Size",
+            key: "fileSize",
+            width: 200,
+            render: (run: RenderRunInfo) => (
+              <Tooltip
+                overlay={
+                  <>
+                    {formatBytes(run.taskCounts.fileSize)} •{" "}
```
fm3, line 192: Is there a reason for the • char? I would have expected | or comma or dash.

Also, maybe add a disclaimer like “Note: manual changes on disk are not respected here“?

fm3: Also, I think on other places (like organization storage report) we went for powers-of-ten filesize formatting. Are there still other places where powers-of-two is used? Should we unify them? I am in favor of powers-of-ten everywhere (but showing the correct unit is of course the most important, which is already ok here).

fm3: Thanks for considering my feedback :) What do you think about my idea of a disclaimer concerning the manual changes on disk?

fm3 (approved):

PR #7076, opened 2023-05-15
Allow to disable automatic reloading of meshes during proofreading

### LINKED ISSUE #6976: [Proofreading] Allow to disable automatic reload of mesh
When proofreading, the automatic reload of meshes can be inconvenient (as users feel blocked by it). It should be optional.

dieknolle3333:
### Steps to test:
- See prerequisites in issue #6976 
- Activate proofreading mode in toolbar
- If switch button is activated, meshes should render as before when clicking a segment
- If it is deactivated, meshes should not be rendered after clicking a segment
- Setting should be stored for a user, so logging on and off shouldn't change the setting of the switch

### Issue:
- fixes #6976

------


frontend/javascripts/oxalis/default_state.ts
```diff
@@ -50,6 +50,7 @@ const defaultState: OxalisState = {
   datasetConfiguration: defaultDatasetViewConfigurationWithoutNull,
   userConfiguration: {
     autoSaveLayouts: true,
+    autoRenderMeshInProofreading: false,
```
dieknolle3333, line 53: was unsure which default makes sense

philippotto: true would be better, as this was the old default :)

dieknolle3333:
I thought your code concerning the button styling from the mentioned PR was very useful. is there any way to incorporate it into my code while keeping you as the author?
Maybe I could delete the snippet and you could commit it to this branch?

dieknolle3333:
I left the opacity out because it felt more like an adjustment if the button was disabled. I will add it back in :)

philippotto (approved):
Great, thank you :)

dieknolle3333:
@philippotto I felt like the merge conflicts were non-trivial. My new button is still working as expected, but could you check whether I destroyed any functionality of your auto-select SAM PR (or anything else)?

Issue #6519, opened 2022-09-30
Perform mergeVertices across chunks


philippotto:
/edit: Initially, the issue focused on only one use case, but in fact there are two:
- when exporting a mesh as STL
- when loading meshes in chunks, there might be visible chunk borders (depending on the compression) which could use fusing

Since meshes are computed in chunks, downloading an STL for a mesh usually has redundant vertices at the chunk borders. `three/examples/jsm/utils/BufferGeometryUtils.js` provides a `mergeVertices` function which could get rid of these. However, the function  takes one geometry as input (whereas we have multiple ones). So, it cannot be used directly.

Possible approaches:
- merge the input geometries into one before calling the function
- take the source code as an inspiration to create a new function which can deal with multiple geometries

philippotto:
Also see https://github.com/scalableminds/webknossos/pull/7077#issuecomment-1548193832 where mergeVertices was called on multiple geometries within a chunk. However, this didn't really make a noticable, visible difference. The main issue are still the chunk-borders and for that the geometries need to be merged at another code location (and so that their global offsets are taken into account).

PR #7077, opened 2023-05-15
Improve performance for large (oversegmentation) meshes

daniel-wer:
Avoid triggering the raycasting when rotating or moving the 3d viewport to avoid lags.
Merge buffer geometries per chunk when loading oversegmentation meshes to avoid lag due to too many threejs scene objects.

- https://___.webknossos.xyz

### Steps to test:
- abc

------

### DIFF
--- a/frontend/javascripts/oxalis/model/sagas/isosurface_saga.ts
+++ b/frontend/javascripts/oxalis/model/sagas/isosurface_saga.ts
@@ -4,7 +4,7 @@ import { V3 } from "libs/mjs";
 import { chunkDynamically, sleep } from "libs/utils";
 import ErrorHandling from "libs/error_handling";
 import type { APIDataset, APIMeshFile, APISegmentationLayer } from "types/api_flow_types";
-import { mergeVertices } from "libs/BufferGeometryUtils";
+import { mergeBufferGeometries, mergeVertices } from "libs/BufferGeometryUtils";
 import Deferred from "libs/deferred";
 
 import Store from "oxalis/store";
@@ -73,6 +73,7 @@ import processTaskWithPool from "libs/task_pool";
 import { getBaseSegmentationName } from "oxalis/view/right-border-tabs/segments_tab/segments_view_helper";
 import { RemoveSegmentAction, UpdateSegmentAction } from "../actions/volumetracing_actions";
 import { ResolutionInfo } from "../helpers/resolution_info";
+import { mergeGeometries } from "libs/BufferGeometryUtils";
 
 export const NO_LOD_MESH_INDEX = -1;
 const MAX_RETRY_COUNT = 5;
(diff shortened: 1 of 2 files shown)


frontend/javascripts/oxalis/model/sagas/isosurface_saga.ts
```diff
@@ -824,33 +824,53 @@ function _getLoadChunksTasks(
...
+                      const bufferGeometry = yield* call(loader.decodeDracoFileAsync, chunk.data);
+                      bufferGeometries.push(bufferGeometry);
+                    } catch (error) {
+                      errorsWithDetails.push({ error, chunk });
+                    }
                   }
+
+                  const geometry = mergeBufferGeometries(bufferGeometries);
+
+                  // If mergeBufferGeometries does not succeed, the method logs the error to the console and returns null
```
philippotto, line 853: just to doublecheck: did you deliberately test this or did mergeBufferGeometries fail sometimes? (as in, there were problems with it?)

philippotto (commented):
code looks perfect :+1: i didn't test yet, as I have one suggestion which might make sense to do before testing:

Since you implemented the merging of meshes in one chunk, the issue https://github.com/scalableminds/webknossos/issues/6519 just got even easier. I think, it would only be a matter of adding a line à la `geometry = mergeVertices(geometry);`.

Do you feel like giving this a try to see whether:
- this reduces visible mesh borders
- the performance hit is negligible

If you don't feel like tacking this on this PR, it's also fine. Just let me know; then, I'll test the PR :)

daniel-wer:
> Since you implemented the merging of meshes in one chunk, the issue https://github.com/scalableminds/webknossos/issues/6519 just got even easier. I think, it would only be a matter of adding a line à la geometry = mergeVertices(geometry);.
> 
> Do you feel like giving this a try to see whether:
> 
> - this reduces visible mesh borders
> - the performance hit is negligible

I tested that, see this log for some time measurements and how many vertices were removed for some sample chunks for lods 0 and 4:

```
console.js:13 Before 9045 lod 0
isosurface_saga.ts:863 simplification: 50.305908203125 ms
console.js:13 After 8859 lod 0
console.js:13 Before 7085 lod 0
isosurface_saga.ts:863 simplification: 29.9140625 ms
console.js:13 After 6932 lod 0
console.js:13 Before 12333 lod 0
isosurface_saga.ts:863 simplification: 73.60693359375 ms
console.js:13 After 11944 lod 0
console.js:13 Before 12137 lod 0
isosurface_saga.ts:863 simplification: 115.741943359375 ms
console.js:13 After 11763 lod 0
console.js:13 Before 12848 lod 4
isosurface_saga.ts:863 simplification: 84.89111328125 ms
console.js:13 After 10499 lod 4
console.js:13 Before 13198 lod 4
isosurface_saga.ts:863 simplification: 62.898193359375 ms
console.js:13 After 12019 lod 4
console.js:13 Before 13172 lod 4
isosurface_saga.ts:863 simplification: 62.249755859375 ms
console.js:13 After 11403 lod 4
console.js:13 Before 10787 lod 4
isosurface_saga.ts:863 simplification: 88.73193359375 ms
console.js:13 After 9325 lod 4
console.js:13 Before 11184 lod 4
isosurface_saga.ts:863 simplification: 55.5029296875 ms
console.js:13 After 9816 lod 4
```

For the large mesh I'm testing with, this would be called thousands of times so the performance impact would be noticeable I'd say. For the in-chunk borders, I don't see any difference :/

Before:
chunk_borders_before
After:
chunk_borders_after

frontend/javascripts/oxalis/model/sagas/isosurface_saga.ts
daniel-wer: This was mentioned in the documentation and I also saw it in the source code of the method. It never happened during testing.

philippotto (approved):
Awesome :tada: :tada: :tada: 

I also tested the performance and for me the FPS went from 15 to 40 when using the rotate-buttons for a scene with three large-ish meshes. I also noticed that the loading speed of the meshes was faster, but maybe this was imagination (?)

Also thank you for giving the mergeVertices idea a go :+1: I commented in the original issue that we probably need something that merges across chunks. And even that, it's probably a trade-off between performance and looks.

PR #7081, opened 2023-05-16
Allow dataset managers to see all workflow reports of their orga

fm3:
### Steps to test:
- set up voxelyticsEnabled locally, run voxelytics workflow
- promote sample user 2 to dataset manager, log in as sample user 2. should see the workflow

### Issues:
- fixes https://discuss.webknossos.org/t/access-to-voxelytics-overview-table/1786

### TODO:
 Currently, the whole administration menu is hidden for non-team-managers. @philippotto do you think it would make sense to show it to dataset managers (that are not team managers)? Otherwise, the voxelytics workflows link is not visible to them

------


normanrz:
> Currently, the whole administration menu is hidden for non-team-managers. @philippotto do you think it would make sense to show it to dataset managers (that are not team managers)? Otherwise, the voxelytics workflows link is not visible to them

We could also move the Voxelytics workflow list to the dashboard

philippotto:
> > Currently, the whole administration menu is hidden for non-team-managers. @philippotto do you think it would make sense to show it to dataset managers (that are not team managers)? Otherwise, the voxelytics workflows link is not visible to them
> 
> We could also move the Voxelytics workflow list to the dashboard

Do you mean that there would be an additional tab in the dashboard for the voxelytics page? Or only a sub menu item in the "Dashboard" menu in the navbar?

> Currently, the whole administration menu is hidden for non-team-managers. @philippotto do you think it would make sense to show it to dataset managers (that are not team managers)? Otherwise, the voxelytics workflows link is not visible to them

Yes, I think, it's okay. Any chance there is another category, too, that should be accessible by dataset managers? Then, the menu wouldn't consist of a single item.

normanrz:
> Do you mean that there would be an additional tab in the dashboard for the voxelytics page? Or only a sub menu item in the "Dashboard" menu in the navbar?

I was thinking of a new tab. But no strong opinion.

philippotto:
> @philippotto do you think it would make sense to show it to dataset managers (that are not team managers)?

I just pushed a change which does this. Additionally, voxelytics jobs are visible to dataset managers.

> > Do you mean that there would be an additional tab in the dashboard for the voxelytics page? Or only a sub menu item in the "Dashboard" menu in the navbar?
> 
> I was thinking of a new tab. It doesn't really make sense under "Administration", so it needs a new home. But no strong opinion.

I'm not against that, but I went with the simpler route for now.

PR #7082, opened 2023-05-16
Fix bug in fallback rendering when layer has missing mags

philippotto:
The computation of the representative mag (used for aligning the vertices with the buckets) effectively ignored layers for which the active mag doesn't exist. However, the fallback mags should be taken into account in that case.

- https://___.webknossos.xyz

### Steps to test:
- abc

### Issues:
- follow-up to #6085 

------


daniel-wer (approved):
I can no longer reproduce the bug, thank you for fixing this so quickly! :+1:

PR #7085, opened 2023-05-17
Fix duplicate on volumes with no fallback layer

fm3:
The RemoteFallbackLayer is only need for duplicating the editable mapping structures, but was tried to construct always

- https://duplicatewithoutfallbacklayer.webknossos.xyz

### Steps to test:
- create volume annotation with no fallback layer
- should be able to duplicate it

------


philippotto (approved):
LGTM :+1:

PR #7087, opened 2023-05-20
Show organization in dataset info tab

### LINKED ISSUE #6922: Show owning organization in dataset info tab
When looking at a dataset which belongs to another organization (could be a public dataset), there's no easy way of knowing which organization owns the dataset. This could be shown in the dataset info tab view. Probably, the same should happen for annotations.

image

dieknolle3333:
### Steps to test:
- You need to set up an additional organization locally. The easiest way to do this is to set isWkorgInstance and mail.logToStdout to true each in the application.conf. Then, log out and create a new account. An confirmation link should be printed in the console. Click the link and then finish the registration. Then, the second account will belong to a separate organization.
Then, switch back to the first account (or do this beforehand) and make a dataset public via the dataset settings. Then, open the dataset url with the second account so that the base condition is met. (credits to @philippotto for this description)
- Check this in three places: 
-- In the dataset list view, after clicking on a dataset.
-- When viewing the dataset in the dataset info tab on the right.
-- When viewing or creating an annotation, again in the dataset info tab.
- (If the setup seems to hard, you can also change the checks === to !== so that you can test it with a dataset that belongs to your organization.)

### Issues:
- fixes #6922

------
- [x] ~~Updated [documentation](../blob/master/docs) if applicable~~ doesnt seem applicable
- [x] Removed dev-only changes like prints and application.conf edits !!!
- [x] Considered [common edge cases](../blob/master/.github/common_edge_cases.md). note: also works if activeUser is null, so if no one is logged in


frontend/javascripts/dashboard/folders/details_sidebar.tsx
```diff
@@ -79,6 +84,26 @@ function getMaybeSelectMessage(datasetCount: number) {
...
+    async () => {
+      const owningOrga = await getOrganization(selectedDataset.owningOrganization);
+      return owningOrga.displayName;
+    },
+    null,
+    [],
+  );
+
+  const renderOrganization = () => {
+    if (activeUser?.organization !== selectedDataset.owningOrganization) return; //TODO invert
```
dieknolle3333, line 98: dont miss me

frontend/javascripts/oxalis/view/right-border-tabs/dataset_info_tab_view.tsx
```diff
@@ -527,6 +570,15 @@ class DatasetInfoTabView extends React.PureComponent<Props, State> {
     );
   }
 
+  maybePrintOrganization = () => {
+    const { activeUser, dataset } = this.props; // TODO can I use activeOrganization?
+    const datasetOrAnnotationOwningOrganization = this.props.isDatasetViewMode
+      ? dataset.owningOrganization
+      : this.props.tracing.annotationId;
+    if (activeUser?.organization !== datasetOrAnnotationOwningOrganization) return; // TODO invert
```
dieknolle3333, line 578: yes hello dont miss me

dieknolle3333:
@philippotto right now I left the check for equality inverted because I am not sure how include another organizations dataset in my users dataset list view. Of course, for the final check, this need to be put right. I feel like everything else is working as I expect it, so I am requesting a review already :)

frontend/javascripts/dashboard/folders/details_sidebar.tsx
```diff
@@ -79,6 +84,26 @@ function getMaybeSelectMessage(datasetCount: number) {
 function DatasetDetails({ selectedDataset }: { selectedDataset: APIDatasetCompact }) {
   const context = useDatasetCollectionContext();
   const { data: fullDataset, isFetching } = useDatasetQuery(selectedDataset);
+  const activeUser = useSelector((state: OxalisState) => state.activeUser);
+  const owningOrganizationDisplayName = useFetch(
+    async () => {
+      const owningOrga = await getOrganization(selectedDataset.owningOrganization);
```
philippotto, line 90: this request will always be triggered now when the the sidebar changes because the user clicked on a different dataset. this is done even when the organization is not needed (a) because the dataset is owned by the active organization anyway and (b) because it was already fetched before.
I think (a) is fine when we add caching anyway, since it will only mean one additional request (for the owning organization) for the whole lifetime of the tab.
Therefore, I suggest to add caching. The library react-query can be very useful for this. It would look something like this:

```js
const { data: owningOrganization } = useQuery(
    ["organizations", selectedDataset.owningOrganization],
    () => getOrganization(selectedDataset.owningOrganization),
    {
        refetchOnWindowFocus: false,
    }
);
```

the first parameter tells react-query under which key the result should be cached. let me know if you have questions.

this will only work in the details_sidebar here, I think, because the dataset info tab doesn't use functional components. however, in the dataset info tab, the fetch is not problematic, since the active dataset doesn't really change over time.

philippotto (commented):
nice :) I left some feedback and will test now.

frontend/javascripts/dashboard/folders/details_sidebar.tsx
```diff
@@ -79,6 +84,26 @@ function getMaybeSelectMessage(datasetCount: number) {
...
+      return owningOrga.displayName;
+    },
+    null,
+    [],
```
philippotto, line 94: if we would stick to useFetch, this would need to be `[selectedDataset.owningOrganization]`. this means that the fetch will be re-executed as soon as the owning organization changes. this would happen when you first select a dataset from orga A and then a dataset from orga B. since the sidebar is not re-mounted, only the selected dataset changes. currently, the sidebar would show the wrong organization for that case. you can check this by opening the network tab in the devtools and selecting different datasets (only the first selection will trigger a request, which is nice on the one hand because "caching", but on the other hand it will never re-fetch even when it's necessary ^^).

this means that my comment from above ("this request will always be triggered now when the the sidebar changes because the user clicked on a different dataset.") is not correct. however, I would still recommend the usage of react-query, since it will fix the bug and ensure caching etc.

frontend/javascripts/oxalis/view/right-border-tabs/dataset_info_tab_view.tsx
```diff
@@ -206,6 +226,29 @@ class DatasetInfoTabView extends React.PureComponent<Props, State> {
     this.props.setAnnotationDescription(newDescription);
   };
 
+  componentDidMount(): void {
+    this.fetchData();
+  }
+
+  async fetchData(): Promise<void> {
+    let owningOrga: APIOrganization;
```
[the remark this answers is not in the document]
dieknolle3333: I'll do that! I missed this one before opening the PR :sweat_smile: should probably do it right from the beginning to avoid this

frontend/javascripts/oxalis/view/right-border-tabs/dataset_info_tab_view.tsx
```diff
@@ -527,6 +570,15 @@ class DatasetInfoTabView extends React.PureComponent<Props, State> {
     );
   }
 
+  maybePrintOrganization = () => {
+    const { activeUser, dataset } = this.props; // TODO can I use activeOrganization?
```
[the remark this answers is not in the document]
dieknolle3333: I had the idea of using the display name of the active organization but that doesn't make any sense: If activeOrga and datasetOrga are equal, I don't need the organization's name :sweat_smile:

frontend/javascripts/oxalis/view/right-border-tabs/dataset_info_tab_view.tsx
```diff
@@ -206,6 +226,29 @@ class DatasetInfoTabView extends React.PureComponent<Props, State> {
...
+  }
+
+  async fetchData(): Promise<void> {
+    let owningOrga: APIOrganization;
+    if (this.props.isDatasetViewMode) {
+      owningOrga = await getOrganization(this.props.dataset.owningOrganization);
+    } else {
+      const annotation = await getAnnotationInformation(this.props.tracing.annotationId);
+      owningOrga = await getOrganization(annotation.organization);
+    }
```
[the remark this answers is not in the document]
dieknolle3333: Can the owning organization of a dataset differ from the owning organization of an annotation within it?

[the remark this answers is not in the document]
philippotto: I think, an annotation is always owned by the organization that also owns the dataset. Even if it was not the case, the semantics of the Organization label should be "Organization that owns the dataset" in my opinion.

> but I could also save a request if I dont need it anyway, if that makes sense.

yes, that would be great :+1:

frontend/javascripts/oxalis/view/right-border-tabs/dataset_info_tab_view.tsx
```diff
@@ -527,6 +570,15 @@ class DatasetInfoTabView extends React.PureComponent<Props, State> {
     );
   }
 
+  maybePrintOrganization = () => {
+    const { activeUser, dataset } = this.props; // TODO can I use activeOrganization?
+    const datasetOrAnnotationOwningOrganization = this.props.isDatasetViewMode
+      ? dataset.owningOrganization
+      : this.props.tracing.annotationId;
```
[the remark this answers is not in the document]
dieknolle3333: wow, l. 577 really doesn't make any sense like this. Okay so I will just conclude that datasetOrga and annotationOrga are always equal

frontend/javascripts/dashboard/folders/details_sidebar.tsx
dieknolle3333: makes sense, thank you!

dieknolle3333: this might be a stupid question but I am trying to understand why you propose `organizations` as the key for the cache. does the cache contain all organizations that have ever been cached? or only one at a time?

philippotto: it's not a stupid question :) they key is a composite of the string `"organizations"` and the actual id `selectedDataset.owningOrganization`. this means, all organizations are cached by using their respective ids. the string "organizations" is simply there to group all cached organizations together. the react-query cache is global and can be used for other stuff, too (e.g., datasets). that's why, we establish a "cache hierarchy" with the string.

philippotto:
Great :) Please invert the conditions where necessary and then I'll do a final test before we merge the pr.

dieknolle3333:
> Great :) Please invert the conditions where necessary and then I'll do a final test before we merge the pr.
done!

philippotto (approved):
Awesome :tada:

PR #7101, opened 2023-05-26
Brush presets

### LINKED ISSUE #5363: Dataset/Annotation view: disable hover effects while viewport is occluded
When I’m viewing a dataset or an annotation and I am in a modal or drop down menu, the mouse movements still cause cells to be highlighted / brush preview to be shown. Additionally, since presumably the mouseout events are not correctly tracked, this occasionally happens in the wrong viewport (where the mouse is not actually placed). This is rather confusing. I’d propose disabling these effects entirely when the mouse is not actually directly above a viewport.

### LINKED ISSUE #7046: Brush presets
## Detailed Description
When brushing, it can be useful to have different brush presets for quick switching between different brush sizes. For example: small brush (5 vx), medium brush (50 vx), large brush (500 vx) etc.

The brush list should be:
- user configurable (i.e., also stored per user)
- have shortcuts (could be composite ones à la `ctrl + k, 1`)

See https://scm.slack.com/archives/C5AKLAV0B/p1683097839871459 for additional context.

dieknolle3333:
### Steps to test:
- Go into brush mode when viewing an annotation
- Open the change brushsize popover
- Make sure there are buttons you can click to set the brush size to small, medium and large
- The brush itself should be visible in the center of the active viewport

### Issues:
- partly fixes #7046: 
-- creates buttons with preset brush sizes
-- renders brush in the center of the active viewport
- fixes #5363

------


dieknolle3333:
@philippotto Some thoughts and questions.
- I manage to center the brush upon opening the popover. The problem is that it is only shown in the active viewport (or in Plane_XY as a default am seeing that correctly). That also means that at first the brush is centered in YZ, but as the popover is above XY, the plane with the centered brush switches to XY. I find that to be quite confusing. Should I simply remember the active viewport upon opening the popover and display the brush there as long as the popover remains open? So the viewport where the brush is shown could only change after closing and opening the popover.
- The initial problem, that the popover is covering up the brush, can still happen (see picture) after resizing the upper right viewport. Is that just how it is or should I try to take care of that?
image

philippotto:
@dieknolle3333 I think you found a deeper problem in the current event handling. Namely, that the brush can be moved even when the mouse is an another UI (e.g, your popover which is placed above the viewport). I investigated this a bit and think that the `input.ts` module should be adapted accordingly. I just a committed the change for this on this PR. Please test whether this solves your problems. If this is the case, it should even be possible to remove the code you added that ignores the "set mouse position" event while the popover is visible (you might want to comment out the relevant condition first and if this works then the rest can be cleaned up; e.g., storing the visible flag in the store is probably not necessary anymore then).

> The initial problem, that the popover is covering up the brush, can still happen (see picture) after resizing the upper right viewport. Is that just how it is or should I try to take care of that?

I think, this edge case is okay to ignore for now. You could try to change the `placement` property of the antd popover to `right` so that the popover hides a bit less of the viewports. However, I'm not sure whether the popover will stick out of the top window edge and thus only will be partly visible (which would be bad ^^).

frontend/javascripts/oxalis/controller/combinations/move_handlers.ts
```diff
@@ -15,7 +19,15 @@ import { setMousePositionAction } from "oxalis/model/actions/volumetracing_actio
 import _ from "lodash";
 
 export function setMousePosition(position: Point2 | null | undefined): void {
-  if (position != null) {
+  console.log("setMousePos");
```
dieknolle3333, line 22: dont forget me

dieknolle3333:
@philippotto 
It works, thank you!

frontend/javascripts/oxalis/view/action-bar/toolbar_view.tsx
```diff
@@ -530,8 +532,22 @@ function CreateTreeButton() {
 }
 
 function ChangeBrushSizeButton() {
+  const dispatch = useDispatch();
   const brushSize = useSelector((state: OxalisState) => state.userConfiguration.brushSize);
-  const maximumBrushSize = useSelector((state: OxalisState) => getMaximumBrushSize(state));
+  const [isBrushSizePopoverOpen, setIsBrushSizePopoverOpen] = useState(false);
+  const maximumButtonBrushSize = useSelector((state: OxalisState) => getMaximumBrushSize(state));
+  const mediumButtonBrushSize = calculateMediumBrushSize(maximumButtonBrushSize);
+  const minimumButtonBrushSize = Math.max(userSettings.brushSize.minimum, 10); // TODO unsure whether that makes sense across the board
```
dieknolle3333, line 540: A brush smaller than 10vx is hardly visible for the dataset I used to test (test-agglomerate-file without zooming in), so I thought this makes sense but wasn't sure. See screenshots attached for visibility of brush sizes 5vx and 10vx with said settings
Screenshot from 2023-05-30 16-28-44
Screenshot from 2023-05-30 16-28-52

philippotto: > A brush smaller than 10vx is hardly visible for the dataset I used to test (test-agglomerate-file without zooming in), so I thought this makes sense but wasn't sure.

I'd vote to not add any clamping, because it can be confusing to the user. Especially, when zooming very far in, this would be very unintuitive in my opinion. If the brush size is so small, that one cannot see it, so be it :)

dieknolle3333: but would this make sense for the default sizes or should the small preset simply be the smallest size possible?

dieknolle3333:
apart from the question about the small brush preset (and the unmerged key shortcuts) this seems done to me, so I am opening the PR :)

frontend/javascripts/oxalis/default_state.ts
```diff
@@ -68,6 +68,7 @@ const defaultState: OxalisState = {
     centerNewNode: true,
     overrideNodeRadius: true,
     particleSize: 5,
+    presetBrushSizesAsc: [],
```
dieknolle3333, line 71: I felt torn between creating an object or leaving the order in the name. I am only 50% happy with this solution, but creating an object {smallBrushSize, mediumBrushSize, largeBrushSize} seemed over the top. Maybe

dieknolle3333: I also wasn't sure about "Asc" or "Ascending" but "Asc" is known because of SQL, so I kept it. Looking forward to your input! @philippotto

frontend/javascripts/oxalis/view/action-bar/toolbar_view.tsx
philippotto: I'm not sure if I understood your question correctly, but I'd change the small brush size preset from 1 to 10 vx, since that is more common in my opinion.

philippotto (changes requested):
Great! Looks already very good :+1: I have left some feedback, but nothing major :)

frontend/javascripts/oxalis/default_state.ts
dieknolle3333: alright! so I would simply create a type `BrushPresets` in the store?

philippotto: yes, that should work :+1:

frontend/javascripts/oxalis/store.ts
```diff
@@ -334,7 +341,11 @@ export type UserConfiguration = {
   readonly centerNewNode: boolean;
   readonly overrideNodeRadius: boolean;
   readonly particleSize: number;
-  readonly presetBrushSizesAsc: [number, number, number] | [];
+  readonly presetBrushSizes: {
+    small: number;
+    medium: number;
+    large: number;
+  } | null;
```
philippotto, line 348:
```suggestion
  readonly presetBrushSizes: BrushPresets | null;
```

frontend/javascripts/oxalis/view/action-bar/toolbar_view.tsx
```diff
@@ -582,6 +620,13 @@ function ChangeBrushSizeButton() {
     );
   };
 
+  const extendedKeyboardBindings = {
+    1: () => handleUpdateBrushSize(smallBrushSize),
+    2: () => handleUpdateBrushSize(mediumBrushSize),
+    3: () => handleUpdateBrushSize(largeBrushSize),
+  };
+  new InputKeyboardNoLoop({}, {}, extendedKeyboardBindings);
```
philippotto, line 628: I'd move the shortcut related code into the plane_controller.tsx module. In `getNotLoopedKeyboardControls`, `extendedControls` are defined where you can add the three shortcuts. The handle-code needs to be redefined there probably, as you need to read out the brush sizes from the store.
Now the explanation as to why this should be moved: Instantiating a new class without really using the object afterwards is usually a code smell. In this case, the instance will linger around and will never be cleaned up again (via the `destroy` method which the object offers). Even worse, on each render of this component, a new InputKeyboardNoLoop instance will be created (and all will be listening to the shortcuts).

philippotto (changes requested):
Looks great, but the keyboard handling code needs to be moved somewhere else :)

philippotto (approved):
Awesome, let's merge it :)

dieknolle3333:
I just thought, maybe I should quickly add the new shortcuts to the keyboard shortcut docu? If you agree I'll do that :) @philippotto

philippotto:
Sure, good catch!

dieknolle3333:
@philippotto can you have a quick look at what I wrote?

PR #7103, opened 2023-05-30
When adding annotation layer, assert no duplicate names

fm3:
This is already asserted in the front-end, but my guess is that users got around that by using the python library. Either way, it should be asserted in the backend as well, and not just via SQL constraints (which saved the valid state here, at least, but which don’t produce very readable error messages)

### Issues:
- fixes https://scm.slack.com/archives/CMBMU5684/p1685214765993529

------


conf/messages
```diff
@@ -204,8 +204,9 @@ annotation.timelogging.read.failed=Time
 annotation.write.failed=Could not convert annotation to json
 annotation.needsEitherSkeletonOrVolume=Annotation needs at least one of skeleton or volume
 annotation.downsample.layerNotFound=Could not downsample volume annotation (annotation layer not found)
-annotation.makeHybrid.explorationalsOnly=Could not convert annotation to hybrid annotation because it is only allowed for explorational annotations.
-annotation.makeHybrid.failed=Could not convert to hybrid.
+annotation.addLayer.explorationalsOnly=Could not add annotation layer because it is only allowed for explorational annotations.
+annotation.addLayer.nameInUse=An annotation layer with this name already exists in this annotation. Please change it to resolve duplicates.
```
frcroth, line 208: Don't know if "resolve" is the right phrasing here, since the duplicate does not exist at this point? Perhaps "prevent"?

frcroth (approved):

PR #7111, opened 2023-06-01
Remove backbone events

### LINKED ISSUE #4873: Clean up event-related code
- only use nano-events instead of backbone events
- get a good understanding of the existing events (`bucketLabeled` on bucket and on datacube; `rerender` etc) to see what can be removed/improved

philippotto:
- https://removebackbone.webknossos.xyz

### Steps to test:
- test mouse inteactions
- test switching between plane and arbitrary modes
- do some monkey testing

### Issues:
- fixes #4873


frontend/javascripts/oxalis/view/plane_view.ts
```diff
@@ -118,10 +104,9 @@ class PlaneView {
     // This prevents the GPU/CPU from constantly
     // working and keeps your lap cool
     // ATTENTION: this limits the FPS to 60 FPS (depending on the keypress update frequence)
-    if (forceRender || this.needsRerender || window.needsRerender) {
-      window.needsRerender = false;
+    if (forceRender || this.needsRerender) {
       const { renderer, scene } = SceneController;
-      this.trigger("render");
+      this.emitter.emit("render");
```
daniel-wer, line 109: What is this needed for? I'm just wondering, because in the arbitrary_view `this.trigger("render");` was removed without any replacement.

daniel-wer (approved):
What a treat :star_struck: Thanks for taking the time to clean up this bit of legacy code!

I didn't notice any issues during my monkey testing :)

frontend/javascripts/oxalis/view/plane_view.ts
philippotto: Very good observation :+1: I just compared plane view and arbitrary view and in the end both need to call SceneController.update. The arbitrary view did this directly and the plane view did this via the event emitter (the planecontroller listened to that and then did the update call). since there's not really a benefit in the extra indirection, I removed it (thus, saving ~20 additional lines). please have another look at the latest commit and give me a final ok :)

daniel-wer (approved):
LGTM :+1:

PR #7112, opened 2023-06-01
Add extended keyboard shortcut mode via ctrl + k for tool shortcuts

### LINKED ISSUE #7045: Shortcut for faster tool switching
I think (1) should be prioritized.

## 1) Fast switching to tools via dedicated shortcuts
See https://scm.slack.com/archives/C5AKLAV0B/p1683099934581439?thread_ts=1683097839.871459&cid=C5AKLAV0B
Something like:
- ctrl + k, b -> selects brush
- ctrl + k, e -> selects eraser
- ...

## 2) Fast switching between two tools via one shortcut
When annotating data, some users would like to switch between tools with a shortcut (which would be assigned to buttons on a digital pen). The existing w/shift+w shortcut is not very predictable.

I see two options:
a) have a shortcut which switches to the most recently used tool (similar to alt+tab)
b) have a shortcut which switches to the "pendant" (brush <-> brush eraser, lasso <-> lasso eraser,  $otherTools <-> move-tool)

I think (a) is the most intuitive (and generic) one, but requires a "set up" by activating the two tools between one wants to switch.

As a shortcut, I suggest "X". This is slightly similar to how Photoshop/GIMP where X switches foreground and background color.

MichaelBuessemeyer:
This PR adds a new "extended" mode to the keyboard shortcuts. This mode is activated using `ctrl + k` and gives access to more commands. Only shortcuts for switching between annotation tools are implemented, but more can be added. 
To leave the "extended" mode, the user needs to use one of the supported extended shortcuts or use `ctrl + k` again or wait 3 seconds.

Here is a list of all shortcuts for all tools. These shortcuts only work, once `ctrl + k` was pressed.
| Shortcut | Tool |
|---|---------------------|
| m | Move Tool           |
| s | Skeleton Tool       |
| b | Brush Tool          |
| e | Brush Erase Tool    |
| l | Lasso Tool          |
| r | Lasso lErase Too    |
| p | Segment Picker Tool |
| q | Quick Select Tool   |
| x | Bounding Box Tool   |
| o | Proofreading Tool   |

- https://___.webknossos.xyz

### Steps to test:
- Open any (hybrid) tracing
- Press `ctrl + k` and try to quick-access the tools above by using their respective shortcuts. This should work.

### Issues:
- fixes #7045

### Open Questions:
- [ ] Where should I add the shortcut for the proofreading tool? Into the Skeleton or volume or default extended shortcuts? Additionally, do I need to make the proofreading tool shortcut only available if the proofreading tool is allowed? Because the shortcut can still exist without any problems and will not activate the proofreading tool, in case it is disabled for the user / lab / instance / dataset e.t.c.

### Known Problems:
- One should be able to deactivate the extended mode by pressing `ctrl + k` again, but Firefox does not trigger the second press as a key event and instead opens the Google search bar.

### TODOs:
- [x] Write changelog entry and link keyboard shortcuts doc.

------


frontend/javascripts/oxalis/controller/viewmodes/plane_controller.tsx
```diff
@@ -468,16 +510,29 @@ class PlaneController extends React.PureComponent<Props> {
...
+    const extendedSkeletonControls =
+      this.props.tracing.skeleton != null ? SkeletonKeybindings.getExtendedKeyboardControls() : {};
+    const extendedVolumeControls =
+      this.props.tracing.volumes.length > 0 != null
+        ? VolumeKeybindings.getExtendedKeyboardControls()
+        : {};
+    ensureNonConflictingHandlers(extendedSkeletonControls, extendedVolumeControls);
+    const extendedAnnotationControls = { ...extendedSkeletonControls, ...extendedVolumeControls };
+    ensureNonConflictingHandlers(extendedAnnotationControls, extendedControls);
+    extendedControls = { ...extendedControls, ...extendedAnnotationControls };
```
MichaelBuessemeyer, line 522: I am not so happy about these lines as they seem not easy to read, but I thought that still ensuring no conflicts makes sense and overweights the hard to read code lines

frontend/javascripts/oxalis/controller/viewmodes/plane_controller.tsx
```diff
@@ -154,6 +167,19 @@ class VolumeKeybindings {
...
+  static getExtendedKeyboardControls() {
+    return {
+      b: () => setTool(AnnotationToolEnum.BRUSH),
+      e: () => setTool(AnnotationToolEnum.ERASE_BRUSH),
+      l: () => setTool(AnnotationToolEnum.TRACE),
+      r: () => setTool(AnnotationToolEnum.ERASE_TRACE),
+      f: () => setTool(AnnotationToolEnum.FILL_CELL),
+      p: () => setTool(AnnotationToolEnum.PICK_CELL),
+      q: () => setTool(AnnotationToolEnum.QUICK_SELECT),
+      o: () => setTool(AnnotationToolEnum.PROOFREAD),
```
MichaelBuessemeyer, line 180: @dieknolle3333 In order to define new "extended" shortcuts, just define them as regular keyboard shortcuts leaving out the `ctrl + k`. But make sure these are included in their own object that is additionally passed to the `InputKeyboardNoLoop` in the constructor. E.g. `ctrl + k, 1` becomes simply `1`.


Then you need to create the `InputKeyboardNoLoop` like this: 
```javascript
const extendedKeyboardBindings = { 1: () -> setBrushToPreset(PRESETS.1)};
new InputKeyboardNoLoop(normalKeyboardBindings, {options}, extendedKeyboardBindings);
```
This way pressing `ctrl + k, 1`  should call ` () -> setBrushToPreset(PRESETS.1)`

frontend/javascripts/libs/input.ts
```diff
@@ -113,7 +157,14 @@ export class InputKeyboardNoLoop {
...
     ];
-    KeyboardJS.bind(...binding);
+    if (isExtendedCommand) {
+      KeyboardJS.withContext("extended", () => {
+        KeyboardJS.bind(...binding);
+      });
+    }
+    KeyboardJS.withContext("default", () => {
+      KeyboardJS.bind(...binding);
+    });
```
MichaelBuessemeyer, line 167: In order to realize / implement the "extended keyboard mode" we use the [context feature of KeyboardJS](https://github.com/RobertWHurst/KeyboardJS#using-contexts). This might not be the way this feature was intended but it works :+1:

frontend/javascripts/libs/keyboard.ts
```diff
@@ -257,6 +257,7 @@
               this._paused = false;
               this.setContext("global");
               this.watch(targetWindow, targetElement, platform, userAgent);
+              this.setContext("default");
```
MichaelBuessemeyer, line 260: We need a default context as otherwise the global context is default which will always be active

MichaelBuessemeyer (commented):
@philippotto I assigned you to this review as Daniel already knows the "hacky"-way we implemented the "extended" keyboard mode using KeyboardJS. This pr should not take long to review (IMO #tm).

MichaelBuessemeyer (commented):
@philippotto I assigned you to this review as Daniel already knows the "hacky"-way we implemented the "extended" keyboard mode using KeyboardJS. This pr should not take long to review (IMO #tm).

frontend/javascripts/libs/input.ts
philippotto, line 167: why isn't the default case in an else-clause? aren't extended shortcuts now defined both in the extended and in the default context?

philippotto (commented):
great :) looks mostly good to me, but left some comments. will test after the next iteration.

frontend/javascripts/libs/input.ts
MichaelBuessemeyer: oh right, that is actually important to have the else. I somehow forgot this :see_no_evil:. It is fixed now.

docs/keyboard_shortcuts.md
```diff
@@ -112,12 +110,31 @@ Note that you can enable *Classic Controls* which won't open a context menu on r
 | Right Mouse Drag                  | Remove Voxels                                               |
 | CTRL + Right Mouse Drag           | Remove Voxels while inverting the overwrite-mode (see toolbar for overwrite-mode) |
 
+## Tool Switching Shortcuts
+
+To be able to switch to a specific tool first press CTRL + K and then the key that was assigned to that tool.
```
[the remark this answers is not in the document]
MichaelBuessemeyer: maybe like this?

philippotto (approved):
Works great! Only see my one comment. However, I think, that this is not really critical (and I'm not sure whether it's fixable in the first place)..

frontend/javascripts/libs/input.ts
```diff
@@ -67,28 +67,70 @@ function shouldIgnore(event: KeyboardEvent, key: KeyboardKey) {
...
+    if (extendedCommands != null && initialBindings[EXTENDED_COMMAND_KEYS] != null) {
+      console.warn(
+        `Extended commands are enabled, but the keybinding for it is already in use. Please change the keybinding for '${EXTENDED_COMMAND_KEYS}'.`,
+      );
+    }
+
+    if (extendedCommands) {
+      this.attach(EXTENDED_COMMAND_KEYS, this.toggleExtendedMode);
+      // Add empty callback in extended mode to deactivate the extended mode via the same EXTENDED_COMMAND_KEYS.
+      this.attach(EXTENDED_COMMAND_KEYS, _.noop, true);
```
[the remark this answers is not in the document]
MichaelBuessemeyer: No, unfortunatley, this does not work for me :/ 
I'll try a bit whether I can fix this as github seemed to have managed it. Maybe I can find a workaround.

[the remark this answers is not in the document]
MichaelBuessemeyer: I just pushed a fix that works on Firefox. I also checked for chrome, but my chrome installation is somewhat broken. Thus it would be best, if you could check this again

philippotto:
Great :+1: :)

PR #7118, opened 2023-06-02
Create bounding box by dragging with box tool

philippotto:
- https://createbboxbydrag.webknossos.xyz

### Steps to test:
- activate tool
- create a bounding box by dragging

### Issues:
- no issue

------
- [x] ~Updated [documentation](../blob/master/docs) if applicable~ not necessary, since the tool is not explained in detail in the docs (as it's quite self-explanatory)


frontend/javascripts/oxalis/controller/combinations/bounding_box_handlers.ts
```diff
@@ -221,9 +225,57 @@ export function getClosestHoveredBoundingBox(
...
+    const isMaxEdge = edgeId % 2 === 1;
+    const resizableDimension = direction === "horizontal" ? indices[1] : indices[0];
+    return {
+      boxId: nearestBoundingBox.id,
+      direction,
+      isMaxEdge,
+      edgeId,
+      resizableDimension,
+    };
+  };
```
daniel-wer, line 271: This method doesn't seem to differ from the earlier one and could be extracted if indices and nearestBoundingBox are supplied as arguments.

daniel-wer (approved):
Great improvement, works very well! :partying_face: 

Only small issue I saw is that the cursor, when not hovering over any bounding box, still indicates a move action which is no longer accurate.

daniel-wer (approved):
Nice cursor choice :sparkles:

PR #7119, opened 2023-06-02
Fix decompression of garbage data after valid gzip data causing decompression to fail

### LINKED ISSUE #7093: Decompression of sharded mag1 precomputed dataset on s3 fails
## Steps to Reproduce the bug
1. Explore s3://bossdb-open-data/iarpa_microns/minnie/minnie65/em/
2. See that there is no histogram!
3. Log shows "Reading chunk at s3://bossdb-open-data/iarpa_microns/minnie/minnie65/em/8_8_40/10b43.shard failed <~ chunk.decompress.failed <~ Garbage after a valid .gz stream Stack trace: java.io.IOException: Garbage after a valid .gz stream"

Note that for the same dataset hosted on GCS, it does not fail

frcroth:
I am not sure if this breaks any other datasets (it fixes minnie65 with s3). The true parameter was introduced with #6466, I tested the only n5 dataset known to me and it worked.

- https://___.webknossos.xyz

### Steps to test:
- Explore minnie65 via s3

### Issues:
- fixes #7093 

------


fm3 (approved):
Thanks! I tested a couple more datasets and they worked for me as well :)

PR #7121, opened 2023-06-05
Lazy load onnx and canvas2html module

### LINKED ISSUE #7084: Lazy-load onnx runtime
As a follow-up to #6991, the onnxruntime-web module should ideally be lazy-loaded so that users (or instances) that don't use the feature don't pay the overhead of the module.

philippotto:
I also extracted the html2canvas module since it caught my attention in the bundle analyzer and was simple to extract.

- https://lazyonnx.webknossos.xyz

### Steps to test:
- open an annotation and the network tab
- select the quick select tool (onnx module and ml model should be loaded)
- use the tool
- hit Q to create a screenshot (html2canvas should be loaded)
- screenshot should work

### Issues:
- fixes #7084 

------
- [ ] ~Updated [changelog](../blob/master/CHANGELOG.unreleased.md#unreleased)~ doesn't seem changelog-worthy to me


daniel-wer (approved):
Works great :tada: 

Tangential, but during testing I noticed that triggering the screenshot functionality, leads to all sorts of css, svg, woff files to be redownloaded for every single screenshot (also happens on master!). Do you understand why this is? I understand that the `logo-screenshot.svg` is downloaded, but what about all the other files :thinking:

philippotto:
> Tangential, but during testing I noticed that triggering the screenshot functionality, leads to all sorts of css, svg, woff files to be redownloaded for every single screenshot (also happens on master!). Do you understand why this is? I understand that the `logo-screenshot.svg` is downloaded, but what about all the other files thinking

html2canvas seems to copy the document into an iframe to be able to do the export. Apparently, this means that these resources are loaded again (however, if "disable cache" is not enabled, the resources are directly served from cache for me). So, I guess, we can't do much about it :thinking:

daniel-wer:
> html2canvas seems to copy the document into an iframe to be able to do the export. Apparently, this means that these resources are loaded again (however, if "disable cache" is not enabled, the resources are directly served from cache for me). So, I guess, we can't do much about it thinking

Ah, that makes sense, thanks for investigating! I agree, that there's no need to do anything about it.

PR #7124, opened 2023-06-05
Prevent 'negative' buckets from being created

### LINKED ISSUE #7117: Volume annotations should not save buckets with negative bucket position
By brushing outside of the dataset bbox in a volume annotation, save actions with negative bucket positions are generated. These should be filtered out (also the backend should assert that no such buckets are saved)

philippotto:
- https://nonegativebrushing.webknossos.xyz

### Steps to test:
- brush at negative positions (nothing should happen)
- check that no buckets are sent to the server
- note that updateSegments will still be created with negative anchor positions, but I think it's alright (the case is very unlikely anyway and shouldn't do any harm)

### Issues:
- fixes #7117

------


fm3 (approved):
Backend LGTM

daniel-wer:
@philippotto Could you explain the fix? As far as I can see, the tracing's bounding box being negative caused "negative" buckets to be created and sent to the server. How did the tracing's bounding box become negative and shouldn't this be prevented, too?

daniel-wer:
Thank you, that makes sense :+1: Does it make sense to use the lower boundary of the dataset instead, then? Otherwise, the behavior is inconsistent with the upper boundary being strict and the lower boundary always being at [0, 0, 0].

philippotto:
> Thank you, that makes sense +1 Does it make sense to use the lower boundary of the dataset instead, then? Otherwise, the behavior is inconsistent with the upper boundary being strict and the lower boundary always being at [0, 0, 0].

Good point! I also refactored the code a bit so that we have one bbox type less. please have another look :)

daniel-wer (approved):
Nice cleanup, LGTM! :+1:

PR #7125, opened 2023-06-06
Change sttp backend to HttpURLConnectionBackend

### LINKED ISSUE #7073: ZARR streaming sometimes returns empty buckets
## Steps to Reproduce the bug

1. Follow the steps from [this issue](https://github.com/scalableminds/webknossos/issues/7072)
2. Correct the mags in the datasource properties so that url and mag match
3. open the dataset
4. some buckets remain black

See screenshot:
image

The back-end replies with bucket data (missing-bucket header is set to empty), but the received data is black.

frcroth:
I don't know why, but this fixes 7073 for me. When testing that issue, I noticed that responses would sometimes be empty, even though I could manually request them and they would be full. This looked like an sttp issue, I tried changing the backend, and the empty responses disappeared. 

- https://changehttpbackend.webknossos.xyz/

### Steps to test:
- See #7073

### Issues:
- fixes #7073

------


fm3 (approved):
Works for me as well. I also tested another dataset with basicauth. Could you add a changelog entry that the dataset streaming symptom is fixed?

philippotto:
> When testing that issue, I noticed that responses would sometimes be empty, even though I could manually request them and they would be full. This looked like an sttp issue, I tried changing the backend, and the empty responses disappeared.

Great find :+1:

PR #7032, opened 2023-04-27
Retire wk-connect

hotzenklotz:
(merge this only after the remaining connect datastores have been shutdown)

PR removes wk-connect integration from WK frontend. Neuroglancer Precomputed, Zarr, and N5 dataset can now be streamed/loaded natively from the cloud by WEBKNOSSOS instead.

Changes:
- Remove forms/UI for Precomupted and BossDB dataset imports. Use "Add remote Dataset" instead
- Remove mentions to wk-connect and BossDB in docs
- Remove wk-connect from Docker file
- Merged all the backend changes from #7031 

### Steps to test:
- Data loading for regular datasets should still work as usual

### Issues:
- contributes to https://github.com/scalableminds/webknossos-connect/issues/144

------
(Please delete unneeded items, merge only when none are left open)
- [x] Updated [changelog](../blob/master/CHANGELOG.unreleased.md#unreleased)
- [x] Updated [migration guide](../blob/master/MIGRATIONS.unreleased.md#unreleased) if applicable
- [x] Updated [documentation](../blob/master/docs) if applicable
- [ ] Adapted [wk-libs python client](https://github.com/scalableminds/webknossos-libs/tree/master/webknossos/webknossos/client) if relevant API parts change
- [ ] Removed dev-only changes like prints and application.conf edits
- [x] Considered [common edge cases](../blob/master/.github/common_edge_cases.md)
- [ ] Needs datastore update after deployment

PR #7126, opened 2023-06-06
Add missing evolution to migration guide

philippotto:
Follow-up to https://github.com/scalableminds/webknossos/pull/7032.

- [X] Updated [migration guide](../blob/master/MIGRATIONS.unreleased.md#unreleased) if applicable


fm3 (approved):
good catch!

PR #7129, opened 2023-06-08
Fix true brush size for small, uneven brush sizes

### LINKED ISSUE #7128: Brush preview can be imprecise

philippotto:
- https://___.webknossos.xyz

### Steps to test:
- create annotation
- use brush with different sizes and test that the preview is equal to the actually painted area

### Issues:
- fixes #7128

------


fm3 (approved):
Works for me :) Thanks!

PR #7131, opened 2023-06-08
Add support for transformations with thin plate splines

### LINKED ISSUE #6965: Experiment with Thin-Plate Splines
Do a spike on TPS to check whether they could be used ad-hoc for multi-modality datasets.

Design doc:
https://www.notion.so/scalableminds/Design-Doc-Thin-Plate-Splines-in-WK-51a79067faf84fb198244e0e0057f574

philippotto:
Adds support for a thin-plate-spline transform which is defined as a list of correspondences (source and target points). An (inverse) TPS is found which fits these correspondences. The resulting weights are passed to the shader which then uses these to transform the coordinates ad-hoc.
The rest of the code (mostly the bucket pickers) don't use the TPS, but instead use an affine approximation which is also derived from the correspondences. The downside of this approach is that the bucket picking can get inaccurate if the affine approximation deviates from the actual TPS transformation. However, in the one known demo case, this is not an issue, which is why I'd keep it as is for now.

### Steps to test:
- open this dataset: https://tps.webknossos.xyz/datasets/sample_organization/C555_DIAMOND_2f-TPS/view#513,512,480,0,0.424
- ensure correct rendering
- you can compare it with https://tps.webknossos.xyz/datasets/sample_organization/C555_DIAMOND_2f/view#513,512,480,0,0.424 which is the same DS but with an affine-only transformation

### Issues:
- fixes #6965 

------


philippotto:
@fm3 As discussed, I imagine the following format for the new tps transform type:

```
     {
        type: "thin_plate_spline";
        correspondences: { source: [number, number, number]; target: [number, number, number] };
      }
```

Could you adapt the back-end accordingly? :slightly_smiling_face:

fm3:
> Could you adapt the back-end accordingly?

Should be done, please have a look and ping me if you need anything else :)

frontend/javascripts/libs/utils.ts
```diff
@@ -407,7 +407,12 @@ export function stringToNumberArray(s: string): Array<number> {
   const result = [];
 
   for (const e of stringArray) {
-    const newEl = parseFloat(e);
+    let newEl;
+    try {
+      newEl = eval(e);
```
philippotto, line 412: @daniel-wer I adapted the position input view so that one can also do mathematical expressions in it. This was handy during development (and I also imagine this to be helpful when inspecting datasets analytically). For example, you can change the position 1000, 1000, 1000 to 2 * 1000 + 1234, 1000, 1000 and it will immediately jump to 3234, 1000, 1000.
The implementation with `eval` should be fine security-wise, since the field can only be edited by the user (who has control over JS anyway). The position encoded in the url is **not** eval'ed.
What do you think about this feature and the implementation? Do you think it can stay or should I remove it again?

frontend/javascripts/oxalis/model/accessors/dataset_accessor.ts
```diff
@@ -647,29 +648,46 @@ export function getMappingInfoForSupportedLayer(state: OxalisState): ActiveMappi
   );
 }
 
-function _getTransformsForLayerOrNull(layer: APIDataLayer): Matrix4x4 | null {
-  if (!layer.coordinateTransformations) {
+type Transform =
+  | { type: "affine"; affineMatrix: Matrix4x4 }
+  | { type: "thin_plate_spline"; affineMatrix: Matrix4x4; tpsInv: TPS3D };
```
daniel-wer, line 653: Just noting the use of snake-case instead of camel-case. Ignore, if it was deliberate.

frontend/javascripts/test/libs/estimate_affine.spec.ts
```diff
@@ -0,0 +1,32 @@
...
+
+  const transform = (x: number, y: number, z: number) => {
+    const vec = new Matrix([[x, y, z, 1]]).transpose();
+    const transformed = affineMatrix.mmul(vec);
+    return transformed.to1DArray() as Vector3;
+  };
+
+  almostEqual(
+    t,
+    [568.1202797015036, 528.013612246682, 1622.1124501555569],
```
daniel-wer, line 21: How did you obtain these ground-truth values you are comparing to?

daniel-wer (commented):
Looks very good! I'm impressed by how relatively few changes were needed to support this :+1: 

It also worked well with the dataset you supplied. For further testing, it would be great to get a dataset with a scale other than [1, 1, 1], preferably anisotropic.

tools/check-cyclic-dependencies.js
```diff
@@ -1,6 +1,83 @@
 const madge = require("madge");
 
 const KNOWN_CYCLES = [
+  ["types/api_flow_types.ts", "admin/organization/pricing_plan_utils.ts"],
```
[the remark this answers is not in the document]
philippotto: since this is a js file and a ts extension would require a build step, I'm afraid this it is not worth it..

frontend/javascripts/test/libs/estimate_affine.spec.ts
philippotto: These stem from my reference implementation in python. The input coordinates here are also part of the correspondences, which is why I could sanity-check them easily by comparing to them the target coordinates.

frontend/javascripts/oxalis/model/accessors/dataset_accessor.ts
philippotto: It felt natural to me, since these also are in the datasource-properties.json. I'm open to changes, but it was deliberate, at least :)

frontend/javascripts/oxalis/geometries/materials/plane_material_factory.ts
```diff
@@ -399,6 +402,12 @@ class PlaneMaterialFactory {
     };
 
     this.material.side = THREE.DoubleSide;
+    this.startListening();
+
+    this.recomputeShaders();
```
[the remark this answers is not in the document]
philippotto: You are right, this is not ideal. I pushed a commit which works around this. Still not 100% satisfied with the solution, but I don't have a better idea for now..

daniel-wer (approved):
LGTM :+1: 

Since this is a niche feature, it would also be alright with me to test the anisotropic case in production - your call.

frontend/javascripts/oxalis/geometries/materials/plane_material_factory.ts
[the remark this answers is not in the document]
philippotto: Sure, I see that the commit doesn't make this really clear and also was unnecessarily complex. I changed the code again so that basically the only remaining change is that the uniform listeners are set up before the initial shader code is generated. The reason for this is that one of the uniform listeners also mutates `this.scaledTpsInvPerLayer` and that property is necessary to compute the shader code. It's not super elegant, as I said before, but it does the trick :see_no_evil:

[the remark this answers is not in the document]
daniel-wer: Thanks, that makes sense :+1:

PR #7135, opened 2023-06-12
Reverse compression on sharded neuroglancer datasets

frcroth:
- https://__.webknossos.xyz

Seems like I misunderstood the precomputed specification with the order of decompressions, which caused problems with one dataset's sharded segmentation.

### Steps to test:
- Open sharded datasets, specifically manc-v1 was failing and works now, while the others should still work.

------


fm3 (approved):
LGTM

PR #7136, opened 2023-06-13
ND Datasets

### LINKED ISSUE #4985: Add timeseries support (4D)
There are 4D dataset where samples are captured over a period of time. Maybe we could add lightweight support for it, by swapping out the datasets based on a slider value.

philippotto:
This PR adds support for n-d datasets and annotations. Such datasets can be imported by using zarr.

See #7229 for restrictions and follow-up tasks regarding n-d support.

- https://d.webknossos.xyz

### Steps to test:
- go to https://d.webknossos.xyz/datasets/sample_organization/idr0101A-4d/view#1114,1075,17,0,0.977
  - was imported remotely with `https://uk1s3.embassy.ebi.ac.uk/idr/zarr/v0.4/idr0101A/13457227.zarr`
- use the t-slider in the toolbar to switch between different t values
- create a new annotation
- create skeleton nodes at different t values
  - edges should be half-transparent in that case
- create segments at different t values
- reload the page
- everything should be at the correct t values
- download/re-import skeletons via the skeleton tab
- download/re-upload whole annotation (not sure whether this is implemented yet?)

### Issues:
- fixes #4985

### Todo:

- [x] Backend
  - [x] accept additionalCoords as property in update actions
    - [x] skeletons. e.g., for createNode `{ "name": "createNode", "value": { "position": [...], "additionalCoords": [{ "name": "t", "value": 10 }], ...}}`
    - [x] volume. e.g., updateBucket and createSegment
      - [x] Save these in FossilDB. This requires changing Morton code calculation in https://github.com/scalableminds/webknossos-wrap ? Or only store xyz in morton code and additional coords afterward :heavy_check_mark: 
      	- Morton code is implemented currently. However, with only 64 Bits available, the coordinate bounds are very limited for high dimensions: Maximum coordinate = 2^(64/Dimensions) so with 5 dimensions already maximum coordinate value is 4096.
      	- [x] Validate that storing and getting them back actually works (did some weird regex parsing there)
      	- [ ] ~~Export/import for volume downloads/uploads~~
    - [x] allow to store additional coords in tracing (next to `editPosition`)
  - [ ] TracingStore: Move AdditionalCoordinateRequest.toProto to ProtoGeometryImplicits? Caused weird compiler errors at different places previously, maybe investigate
  - [x] Update DatasetArray
  - [x] Infer bounds of additional coordinates
  - [x] Make a clearer distinction between additionalCoordinate[request vs definition]. Maybe introduce new names?
  - [x] Import/export of nd-skeletons in NML (nodes should get properties, such as `additionalCoordinate-t="10"`)
  - [ ] ~~NML import or export of Volume annotations does not seem to work yet~~ -> change wkw
  - [x] consider writing tests (Added 13 tests :open_mouth: )
  - [x] On merging annotations / tracings etc., merge the additional coords
  - [x] Fix volume tracing bucket keys not needing migrations
- [x] frontend
  - [X] adapt data loading and rendering to higher dimensions
  - [X] adapt skeletons to higher dimensions
  - [x] test saving after backend was adapted
  - [x] encode additional coords in url
  - [x] respect in NML export
  - [x] tackle remaining todo comments
  - [x] bugs
    - [x] fix rendering of skeletons
  - [x] fix tests
  - [x] fix cyclic dependencies

------


philippotto:
@frcroth I added some todo items to the PR description. feel free to add to that.

webknossos-datastore/app/com/scalableminds/webknossos/datastore/datareaders/DatasetArray.scala
```diff
@@ -50,7 +81,8 @@ class DatasetArray(vaultPath: VaultPath,
   // Read from array. Note that shape and offset should be passed in XYZ order, left-padded with 0 and 1 respectively.
   // This function will internally adapt to the array's axis order so that XYZ data in fortran-order is returned.
   private def readAsFortranOrder(shape: Array[Int], offset: Array[Int])(implicit ec: ExecutionContext): Fox[Object] = {
-    val totalOffset: Array[Int] = (offset, header.voxelOffset).zipped.map(_ - _)
+    val totalOffset
+      : Array[Int] = (offset, header.voxelOffset).zipped.map(_ - _) // TODO: header.voxeloffset might not work for non-ngff nd-datasets
```
philippotto, line 85: todo

frontend/javascripts/oxalis/model/actions/annotation_actions.ts
```diff
@@ -275,13 +274,6 @@ export const updateCurrentMeshFileAction = (
     meshFileName,
   } as const);
 
-export const importIsosurfaceFromSTLAction = (layerName: string, buffer: ArrayBuffer) =>
```
philippotto, line 278: was unused

frontend/javascripts/oxalis/model/sagas/isosurface_saga.ts
```diff
@@ -513,13 +529,23 @@ function* _refreshIsosurfaceWithMap(
   // The isosurface should only be removed once after re-fetching the isosurface first position.
   let shouldBeRemoved = true;
 
+  // todop: store mesh with additional coordinates
```
philippotto, line 532: todo: handle this known incompatibility somehow?

frontend/javascripts/oxalis/model_initialization.ts
```diff
@@ -511,6 +524,8 @@ function setupLayerForVolumeTracing(
       // Remember the name of the original layer (e.g., used to request mappings)
       fallbackLayer: tracing.fallbackLayer,
       fallbackLayerInfo: fallbackLayer,
+      // todop
```
philippotto, line 527: todo

philippotto:
@daniel-wer I think you can already start reviewing this PR, as it's quite big and I hope that no big changes are necessary anymore :crossed_fingers:

frontend/javascripts/oxalis/model/sagas/proofread_saga.ts
```diff
@@ -715,9 +735,25 @@ function* removeOldMeshesAndLoadUpdatedMeshes(
       yield* put(removeIsosurfaceAction(layerName, targetAgglomerateId));
     }
 
-    yield* call(loadCoarseMesh, layerName, newSourceAgglomerateId, sourceNodePosition);
+    // Segmentations with more than 3 dimensions are currently not compatible
```
daniel-wer, line 738: Could this be detected in the code, showing an error toast or logging to airbrake?

frontend/javascripts/oxalis/model_initialization.ts
```diff
@@ -560,6 +576,7 @@ function determineDefaultState(
   const { viewMode } = temporaryConfiguration;
   const defaultPosition = datasetConfiguration.position;
```
daniel-wer, line 577: It should also be possible to configure the default `additionalCoordinates`. Could be part of a followup PR.

frontend/javascripts/oxalis/model_initialization.ts
```diff
@@ -569,6 +586,7 @@ function determineDefaultState(
 
   if (someTracing != null) {
     position = Utils.point3ToVector3(someTracing.editPosition);
+    additionalCoordinates = someTracing.editPositionAdditionalCoordinates;
   }
 
   if (urlStatePosition != null) {
```
daniel-wer, line 592: This should also use the `additionalCoordinates` that might have been part of the url state

frontend/javascripts/oxalis/view/components/input_component.tsx
```diff
@@ -1,47 +1,52 @@
-import { Input, InputProps, Tooltip } from "antd";
+import { Input, InputProps, InputRef, Tooltip } from "antd";
 import * as React from "react";
 import _ from "lodash";
 
-type InputComponentState = {
-  isFocused: boolean;
-  currentValue: React.InputHTMLAttributes<HTMLInputElement>["value"];
-}; // TODO Double check if we still need this once React v16 is released.
```
daniel-wer, line 8: Thank you for checking this!

frontend/javascripts/oxalis/view/context_menu.tsx
```diff
@@ -113,32 +114,12 @@ type OwnProps = {
...
-  createBranchPoint: (arg0: number, arg1: number) => void;
-  deleteBranchpointById: (arg0: number, arg1: number) => void;
-  performMinCut: (arg0: number, arg1: number | undefined) => void;
-  removeMesh: (arg0: string, arg1: number) => void;
-  hideMesh: (arg0: string, arg1: number) => void;
-  setPosition: (arg0: Vector3) => void;
-  refreshMesh: (arg0: string, arg1: number) => void;
-};
+
+type DispatchProps = ReturnType<typeof mapDispatchToProps>;
```
daniel-wer, line 122: :star_struck:

frontend/javascripts/oxalis/view/right-border-tabs/connectome_tab/connectome_view.tsx
```diff
@@ -166,6 +166,9 @@ const synapseNodeCreator = (synapseId: number, synapsePosition: Vector3): Mutabl
   timestamp: Date.now(),
   bitDepth: 8,
   interpolation: false,
+  // Don't assume any additionalCoordinates here, because no 4D connectomes are
```
daniel-wer, line 169: Should we collect which features are not prepared for >3d somewhere?

frontend/javascripts/oxalis/view/action_bar_view.tsx
```diff
@@ -59,6 +64,75 @@ type State = {
...
+        </div>
+      }
+    >
+      <ArbitraryVectorInput
+        autoSize
+        vectorLength={additionalCoordinates.length}
+        value={additionalCoordinates.map((el) => el.value)}
+        onChange={changeAdditionalCoordinatesFromVector}
+        style={{ marginLeft: 10, marginRight: 10 }}
+        addonBefore={additionalCoordinates.map((coord) => coord.name).join("")}
```
daniel-wer, line 130: Should this be joined with `", "`, instead maybe so the names are displayed similarly to the values?

frontend/javascripts/oxalis/geometries/materials/edge_shader.ts
```diff
@@ -1,20 +1,25 @@
...
 
   constructor(treeColorTexture: THREE.DataTexture) {
     this.setupUniforms(treeColorTexture);
     this.material = new THREE.RawShaderMaterial({
-      // @ts-expect-error ts-migrate(2565) FIXME: Property 'uniforms' is used before being assigned.
       uniforms: this.uniforms,
       vertexShader: this.getVertexShader(),
       fragmentShader: this.getFragmentShader(),
+      transparent: true,
+      glslVersion: THREE.GLSL3,
```
daniel-wer, line 20: Are we requiring this somewhere else already or would this be a new requirement for systems running webknossos?

philippotto: The same flag is already passed in the node shader. Also, from my understanding, WebGL 2 support always implies support for GLSL 3.0. So, it's not a new requirement :)

frontend/javascripts/oxalis/model/reducers/flycam_reducer.ts
```diff
@@ -237,6 +238,44 @@ function FlycamReducer(state: OxalisState, action: Action): OxalisState {
...
+      // In case the specified additional coordinates don't cover all existing additional
+      // coordinates, add the missing coordinates back.
+      // This *should* not happen, but in case of a bug somewhere, it's better to guard
+      // against this here (for example, if a skeleton node doesn't have the necessary
+      // additional coordinates, the UI shouldn't forget about the other additional
+      // coordinates).
+      let { values } = action;
+      if (action.values == null || action.values.length !== expectedAdditionalCoordinatesCount) {
+        values = Utils.values(unifiedAdditionalCoordinates).map(({ name, bounds }, index) => {
+          const existingadditionalCoordinates = state.flycam.additionalCoordinates;
```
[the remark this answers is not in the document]
philippotto: whoops, this happened due to a global search&replace, I guess.

frontend/javascripts/oxalis/view/action_bar_view.tsx
philippotto: I would find something like `uvw` intuitive (but maybe I'm biased) and more importantly: shorter than `u, v, w`. Therefore, I'd keep it. >= 5d datasets will be very rare, anyway.

frontend/javascripts/oxalis/model/reducers/flycam_reducer.ts
```diff
@@ -237,6 +238,44 @@ function FlycamReducer(state: OxalisState, action: Action): OxalisState {
...
+          const fallbackValue =
+            (existingadditionalCoordinates != null
+              ? existingadditionalCoordinates[index]?.value
+              : null) ?? Math.floor((bounds[1] - bounds[0]) / 2);
```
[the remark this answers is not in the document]
philippotto: Good idea :+1: In that special case, bounds[0] is still sensible value and much better to understand from a user perspective.

frontend/javascripts/oxalis/model/sagas/isosurface_saga.ts
philippotto: Good idea! I added the toast into the loadAdHocIsosurface saga.

> Probably, there won't exist any mesh files for >3d datasets. Ad-hoc mesh computation would compute the correct mesh for the current additionalCoordinates but then act like it's the correct one for all other additionalCoordinates, too, right?

yes and yes :+1:

frontend/javascripts/oxalis/model/sagas/proofread_saga.ts
philippotto: see the other comment. a toast will be shown in that unlikely event.

frontend/javascripts/oxalis/model_initialization.ts
philippotto: I added this to https://github.com/scalableminds/webknossos/issues/7229.

frontend/javascripts/oxalis/model_initialization.ts
philippotto: good catch!

frontend/javascripts/oxalis/view/right-border-tabs/connectome_tab/connectome_view.tsx
philippotto: I also added this to https://github.com/scalableminds/webknossos/issues/7229.

frontend/javascripts/oxalis/model/helpers/nml_helpers.ts
```diff
@@ -291,12 +309,17 @@ function serializeTrees(trees: Array<Tree>): Array<string> {
 function serializeNodes(nodes: NodeMap): Array<string> {
   return nodes.map((node) => {
     const position = node.position.map(Math.floor);
+    const maybeProperties =
+      (node.additionalCoordinates?.length || 0) === 0
+        ? {}
+        : { additionalCoordinates: JSON.stringify(node.additionalCoordinates) };
```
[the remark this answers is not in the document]
philippotto: very good idea! I changed it in ace858272c.

frontend/javascripts/oxalis/geometries/materials/node_shader.ts
```diff
@@ -157,6 +185,13 @@ vec3 shiftHue(vec3 color, float shiftValue) {
 }
 
 void main() {
+    <% _.each(additionalCoordinates || [], (_coord, idx) => { %>
+      if (additionalCoord_<%= idx %> != currentAdditionalCoord_<%= idx %>) {
+        return;
```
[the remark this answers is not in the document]
philippotto: I introduced the alpha blending so that edges that go from t=1 to t=2 are only half-visible (i.e., there's a linear gradient from 0% to 100%). Using the same alpha mechanism for invisible trees would very likely work, but I wouldn't want to risk potential performance hits. For example, when having a huge annotation where most skeletons are invisible, I imagine less effort the fragment shader when vertices are outside of the viewport (in constrast to being scattered everywhere with alpha=0).
I might be wrong about this, but I don't think challenging the above assumption is really worth it :see_no_evil:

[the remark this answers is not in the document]
daniel-wer: Ok, thanks for explaining, the "half-visible" part was not apparent to me. I would also assume that the shader won't become _more_ performant using alpha instead of moving the vertices outside of the viewport.

frontend/javascripts/oxalis/view/action_bar_view.tsx
daniel-wer: I was thinking about the case where the dimensions are not named with single characters, for example, time and angle. Still, I agree that it's not important for now :+1:

philippotto:
@daniel-wer The dev instance is live now. I also updated the testing instructions (but didn't execute them myself yet). If you feel adventurous and have time, you can give it a go :crossed_fingers:

frontend/javascripts/oxalis/model/helpers/nml_helpers.ts
```diff
@@ -309,10 +299,17 @@ function serializeTrees(trees: Array<Tree>): Array<string> {
...
-      (node.additionalCoordinates?.length || 0) === 0
-        ? {}
-        : { additionalCoordinates: JSON.stringify(node.additionalCoordinates) };
+    const maybeProperties = Object.fromEntries(
+      (node.additionalCoordinates || []).map((coord) => [
+        // Export additional coordinates like this:
+        // additionalCoordinate-t="10"
+        // Don't capitalize coord.name, because it it's not reversible for
+        // names that are already capitalized.
+        `additionalCoordinate-${coord.name}`,
```
daniel-wer, line 308: Since `coord.name` is user-controlled it needs to be escaped (could be done in the `serializeTag` method for all attribute names or triggered by a flag argument). I struggled to find the exaxt rules for attribute _names_, but something likes this could be a start:

> The characters that are not allowed in attribute names are space, quotation marks (", '), less than (<), greater than (>), ampersand (&), and some reserved XML characters like question mark (?) and exclamation mark (!).

We would need to test whether saxophone copes with this automatically during deserialization (as it does for attribute values), but probably not :/
It's likely fine to add this to the followup issue if the priority is to get this PR merged.

frontend/javascripts/oxalis/model/helpers/nml_helpers.ts
```diff
@@ -729,11 +714,14 @@ export function parseNml(nmlString: string): Promise<{
                 Math.trunc(_parseFloat(attr, "y")),
                 Math.trunc(_parseFloat(attr, "z")),
               ] as Vector3,
-              additionalCoordinates: _parseJSON(
-                attr,
-                "additionalCoordinates",
-                [],
-              ) as AdditionalCoordinate[],
+              // Parse additional coordinates, like additionalCoordinate-t="10"
+              additionalCoordinates: Object.keys(attr)
```
daniel-wer, line 718: The String's `startsWith` method in combination with `slice(len("additionalCoordinate-")` could be a more readable replacement here, but also fine as-is.

frontend/javascripts/test/reducers/flycam_reducer.spec.ts
```diff
@@ -152,3 +155,47 @@ test("Flycam should move by plane in ortho mode with dynamicSpaceDirection", (t)
...
+    adaptedState,
+    // t is passed, but u is missing. Instead, a superfluous
+    // q is passed.
+    FlycamActions.setAdditionalCoordinatesAction([
+      { name: "t", value: 123 },
+      { name: "q", value: 0 },
+    ]),
+  );
+  t.deepEqual(newState.flycam.additionalCoordinates, [
+    { name: "t", value: 123 },
```
daniel-wer, line 198: Is it relevant that t is now 123 but the bounds of "t" are actually only 0 to 10 according to the layer configuration?

daniel-wer (commented):
Thank you for adding a couple of frontend unit tests :pray: The code reads very well, I've only got some minor comments. Will report back after testing.

daniel-wer:
Many things are working great already! :tada: I love the fading out of skeleton edges between different `additionalCoordinates`! :sparkles: 

Still, please see my notes from testing:

- [ ] Volume annotation seems to be implemented differently in the front- and backend. When annotating in t=8 in the frontend, the annotation is only visible iff t==8. However, after reloading the annotation, the annotation is visible in all t.
- [ ] The currently active `additionalCoordinate` is not persisted in the frontend-generated nml. It is done for the backend-generated nml, though: `<editPosition x="585" y="1247" z="12" additionalCoordinate-t="9" />`. Also, the backend-generated nml contains this section:
  ```
  <additionalCoordinates>
      <additionalCoordinate name="t" index="0" min="0" max="17" />
    </additionalCoordinates>
  ```
  which is missing from the frontend nml.
- [ ] I'm not sure whether the quick select AI tool is working correctly. The weirdness might be due to the dataset, but it looks like there is an offset :shrug: 
  quick_select_weird
- [ ] In the context menu and in the status bar at the bottom, (node) positions are only shown as 3-dimensional, although there is a 4th dimension in this dataset.
- [ ] When entering an out-of-bounds t-value, the input field continues to show the out-of-bounds value after blurring, but the slider resets to some kind of default value. For the position, values outside of the bounding box are allowed, but it doesn't make sense for the additionalCoordinates, I think.
  invalid_t
- [ ] When entering a non-number t-value, the slider and data don't change, but the non-number t-value remains in the input field. The coordinate should be reset to the last good value (as is happening for the position).
- [ ] In light mode, the "t" label element sticks out a lot from the toolbar as it is white. In dark mode it looks fine.
t_light_mode
t_dark_mode
- [ ] [Followup] Since the `additionalCoordinate`s are not reflected in the standard URL, a reload when viewing a dataset possibly doesn't remember the current `additionalCoordinate`s :/
- [ ] [Followup] It might be a helpful feature to select the coordinate along which the volume interpolation is applied (i.e. to select whether it should happen along z or along an `additionalCoordinate` like time)
- [ ] [Separate Issue] During undo of volume annotations, the position is not changed so users don't see what is actually changed if they are not at the correct position by chance.

frontend/javascripts/oxalis/model/helpers/nml_helpers.ts
philippotto: Good point! I added it to #7229 since I think, it is very niche for now.

frontend/javascripts/oxalis/model/helpers/nml_helpers.ts
philippotto: Since this needs the `additionalCoordinate` literal twice, I'm leaning against that :see_no_evil:

philippotto:
Thanks for your (testing) feedback! I'll go through it now.

> [ ] Volume annotation seems to be implemented differently in the front- and backend. When annotating in t=8 in the frontend, the annotation is only visible iff t==8. However, after reloading the annotation, the annotation is visible in all t.

Oh, I thought I had tested this. However, I can reproduce this right now. @frcroth can you have a look at that? The front-end sends the additionalCoordinates parameters for save and for load. So, it sounds like a back-end problem?

frcroth:
> Thanks for your (testing) feedback! I'll go through it now.
> 
> > [ ] Volume annotation seems to be implemented differently in the front- and backend. When annotating in t=8 in the frontend, the annotation is only visible iff t==8. However, after reloading the annotation, the annotation is visible in all t.
> 
> Oh, I thought I had tested this. However, I can reproduce this right now. @frcroth can you have a look at that? The front-end sends the additionalCoordinates parameters for save and for load. So, it sounds like a back-end problem?

Should be fixed with [dbb504c](https://github.com/scalableminds/webknossos/pull/7136/commits/dbb504cda65298c8a57ca3cd9310f3e9a63d3f05)

frcroth:
Volume data can be saved, but the annotation can not be downloaded and the data reimported. Downloaded volume data is saved in [WKW format](https://github.com/scalableminds/webknossos-wrap#readme), which currently does not support additional coordinates. I do not think changing the format is in the scope of this issue? (I think storing it would be possible, but that would not be wkw anymore)

philippotto:
> Therefore, I think the best solution would be to assert that the additional coordinates are the same when merging.

Yes, I think, this is a reasonable constraint.

normanrz:
> > Volume data can be saved, but the annotation can not be downloaded and the data reimported. Downloaded volume data is saved in [WKW format](https://github.com/scalableminds/webknossos-wrap#readme), which currently does not support additional coordinates. I do not think changing the format is in the scope of this issue? (I think storing it would be possible, but that would not be wkw anymore)
> 
> ^ cc @normanrz FYI This is a relatively big restriction for n-d annotations in my opinion. Export/import via ZARR might make be a sensible next step (as follow-up issue of course).

Zarr would be an option. Let's schedule a followup issue and discuss with @fm3 once he's back.

frontend/javascripts/test/reducers/flycam_reducer.spec.ts
philippotto: Good point, I adapted the test so that both cases (valid and invalid range) are covered.

philippotto:
@daniel-wer Thanks again for your testing :pray: I think I've finished your points (however, I moved some into the follow-up issue).

> * [x]  When entering an out-of-bounds t-value, the input field continues to show the out-of-bounds value after blurring, but the slider resets to some kind of default value. For the position, values outside of the bounding box are allowed, but it doesn't make sense for the additionalCoordinates, I think.
>   invalid_t
> * [x]  When entering a non-number t-value, the slider and data don't change, but the non-number t-value remains in the input field. The coordinate should be reset to the last good value (as is happening for the position).

I fixed these two points, but the new UX can be a bit weird, since the input is already "corrected" while the user edits it. Let's say the valid range goes from 15 to 20 and the user is at 17 (but wants to change to 19). If the user deletes the 7, the new input will be 1 which is invalid and thus it is directly corrected to 15. This is unfortunate, but it turned out quite difficult to handle this correctly. I think, it's okay for the first iteration, though.

webknossos-tracingstore/app/com/scalableminds/webknossos/tracingstore/tracings/volume/VolumeTracingBucketHelper.scala
```diff
@@ -191,29 +270,30 @@ trait VolumeTracingBucketHelper
...
     val key = buildKeyPrefix(dataLayer.name)
-    new BucketIterator(key, volumeDataStore, expectedUncompressedBucketSizeFor(dataLayer), version)
+    new BucketIterator(key, volumeDataStore, expectedUncompressedBucketSizeFor(dataLayer), version, dataLayer)
   }
 
   def bucketStreamWithVersion(dataLayer: VolumeTracingLayer,
                               version: Option[Long]): Iterator[(BucketPosition, Array[Byte], Long)] = {
     val key = buildKeyPrefix(dataLayer.name)
-    new VersionedBucketIterator(key, volumeDataStore, expectedUncompressedBucketSizeFor(dataLayer), version)
+    new VersionedBucketIterator(key, volumeDataStore, expectedUncompressedBucketSizeFor(dataLayer), version, dataLayer)
```
fm3, line 279: I think I would prefer it if the Iterators only received the info about the additional coordinate axes/ranges/whatever we’ll name the, and not the entire data layer. This could hopefully give a bit less entanglement of our data structures

webknossos-datastore/app/com/scalableminds/webknossos/datastore/datareaders/DatasetArray.scala
```diff
@@ -40,6 +43,43 @@ class DatasetArray(vaultPath: VaultPath,
...
+
+    channelIndex match {
+      case Some(c) => offsetArray(axisOrder.c.getOrElse(axisOrder.x - 1)) = c
+      case None    => ()
+    }
+
+    for (additionalCoordinateRequest <- additionalCoordinateRequests) {
+      val index = additionalCoordinates(additionalCoordinateRequest.name).index
+      offsetArray(index) = additionalCoordinateRequest.value
+      // Shape for additional coordinates will always be 1e
```
fm3, line 78: 1e?

test/backend/BucketKeyTestSuite.scala
```diff
@@ -0,0 +1,80 @@
+package backend
+
+import com.scalableminds.util.geometry.Vec3Int
+import com.scalableminds.webknossos.datastore.models.{AdditionalCoordinateRequest, BucketPosition}
+import com.scalableminds.webknossos.datastore.models.datasource.AdditionalCoordinateDefinition
+import com.scalableminds.webknossos.tracingstore.tracings.volume.BucketKeys
+import org.scalatestplus.play.PlaySpec
+
+class BucketKeyTestSuite extends PlaySpec {
```
fm3, line 9:
```suggestion
class VolumeBucketKeyTestSuite extends PlaySpec {
```
(to hint that this is about volume annotations)

fm3 (commented):
Consider me impressed! This already works really well for the test dataset (I also managed to download the zarr data so I could verify that for locally saved data the performance is pretty good already)

I looked through the code, as well as I could understand it. I think the solutions look good! I added a few comments, mostly cosmetic. I also had these thoughts:

- [ ] Should volume download (and possibly tiff export via worker) be disabled, already in frontend? Volume download yields exceptions for 4d brushed data
- What is the NML format for nodes you agreed upon? The frontend nml spec file lists something like time="5" while the NML i downloaded via backend had something like additionalCoordinate-t="5". Should probably be unified? I think we have color.r color.g somewhere, so it might make sense to use a dot as separator here as well, instead of hyphen?
- I’m not super happy with the naming of additionalCoordinates in the sense of a point and in the sense of a range/definition. E.g. a DataLayer has `additionalCoordinates: Option[Seq[AdditionalCoordinateDefinition]]` – maybe the field should also be called additionalCoordinateDefinitions? Or maybe something else entirely? AdditionalAxes? AdditionalCoordinateBounds? I’m not sure what would be the best option, but currently, the name seems dangerously overloaded
- Note that everything that requires on-disk 4d segmentation layers is untested, as we don’t currently have any. (Fallback layer for volume annotations, Mappings, Editable Mappings). We can probably start working on that once we have created ourselves some segmentation layers using the wk python lib. I’d say this is ok for the moment, but let’s keep that in mind.
- There should be a follow-up issue detailing the issue of different axis orders in the underlying zarr array. If I read the comment in DatasetArray.scala correctly, nd datasets currently only work if zyx are the last three axes. Edit: I created #7280 – could you fill in some details there describing the problem?

philippotto:
> * What is the NML format for nodes you agreed upon? The frontend nml spec file lists something like time="5" while the NML i downloaded via backend had something like additionalCoordinate-t="5". Should probably be unified? I think we have color.r color.g somewhere, so it might make sense to use a dot as separator here as well, instead of hyphen?

Where did you find the time="5" part? I don't find that in the nml.spec.ts file. It should be `additionalCoordinate-t="5"` everywhere afaik.

philippotto:
> in frontend/javascripts/test/snapshots/public-test/test-bundle/test/libs/nml.spec.js.md
> 
> (does this link work?)
> 
> https://github.com/scalableminds/webknossos/pull/7136/files#diff-f121b9716d2215444a03c64d6300706b1f12b37a864702eb74b387dae788477cR151

Ah, the time property refers to the timestamp that already existed before this PR. I adapted the test so that at least one node has an additional coordinate. Also, there's already another spec which covers the serialization of additionalCoordinates.

fm3:
> Ah, the time property refers to the timestamp that already existed before this PR. I adapted the test so that at least one node has an additional coordinate. Also, there's already another spec which covers the serialization of additionalCoordinates.

I see, thanks!

frcroth:
@philippotto After consultation with @fm3 I have changed backend AdditionalCoordinateRequest to AdditionalCoordinate and AdditionalCoordinateDefinition to AdditionalAxis. To reflect this, Datalayers now longer have an additionalCoordinates key, but an additionalAxes key (also in the JSON). Can you change the frontend to reflect this naming (axis for the definition)?

philippotto:
> @philippotto After consultation with @fm3 I have changed backend AdditionalCoordinateRequest to AdditionalCoordinate and AdditionalCoordinateDefinition to AdditionalAxis. To reflect this, Datalayers now longer have an additionalCoordinates key, but an additionalAxes key (also in the JSON). Can you change the frontend to reflect this naming (axis for the definition)? (e.g. in nml and everything working with backend jsons)

Done. FYI: During testing, the bucket requests took "forever" (at some point, the requests time out then). With a bit of patience, some requests succeeded then (a single request took 20 seconds, though). Not sure whether we can do anything about that..

webknossos-datastore/app/com/scalableminds/webknossos/datastore/datareaders/zarr/NgffMetadata.scala
```diff
@@ -62,6 +62,37 @@ case class NgffAxis(name: String, `type`: String, unit: Option[String] = None) {
         case Some(unknownUnit)  => Fox.failure(s"Unknown space axis unit: $unknownUnit")
       }
 
+  def timeUnitToSFactor(implicit ec: ExecutionContext): Fox[Double] =
```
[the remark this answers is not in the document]
frcroth: Actually, this is something I haven't considered yet, but should units be supported for additionalAxes? Would be nice to view I guess, can also be followup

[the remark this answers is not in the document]
fm3: I added https://github.com/scalableminds/webknossos/issues/7297 to the list of follow-up issues

app/controllers/AnnotationController.scala
```diff
@@ -39,7 +40,8 @@ case class AnnotationLayerParameters(typ: AnnotationLayerType,
                                      autoFallbackLayer: Boolean = false,
                                      mappingName: Option[String] = None,
                                      resolutionRestrictions: Option[ResolutionRestrictions],
-                                     name: Option[String])
+                                     name: Option[String],
+                                     additionalCoordinate: Option[Seq[AdditionalCoordinateDefinition]])
```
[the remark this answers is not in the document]
fm3: I see that you renamed the class, but not the parameter name. `additionalCoordinate: Option[Seq[AdditionalAxis]]` is not ideal imo. Could it be `additionalAxes: Option[Seq[AdditionalAxis]]`? Could there be other spots like this?

fm3 (commented):
Getting close :) I added two more comments on the backend code.

Maybe you could look into the slow requests philipp mentioned. Maybe this is just due to how the test dataset is stored, but maybe we perform more requests than really needed?

Also, could you have another look at the checkboxes in my comment above? I think some of that is already resolved in the meantime.

app/controllers/AnnotationController.scala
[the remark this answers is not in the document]
frcroth: Thought I had renamed all such cases, did another pass now.

frcroth:
> Done. FYI: During testing, the bucket requests took "forever" (at some point, the requests time out then). With a bit of patience, some requests succeeded then (a single request took 20 seconds, though). Not sure whether we can do anything about that..
Do you mean automatic testing or manually looking at the dataset and changin t/browsing around? Manual testing works very well for me with the local dataset.

fm3:
I just went and did a bit more testing and found a bug:

- [ ] Volume Download + Reupload seems to be broken. I brushed some in a fresh annotation, hit download, and then uploaded the same annotation via drag+drop of the zip to the dashboard. The resulting annotation did not have any visible brushed segmentation. This still works on master. @frcroth could you have a look at what might be going on there?

fm3 (approved):
Let’s go :shipit:

PR #7137, opened 2023-06-13
Async IO for HttpsDataVault, Fox Error Handling

### LINKED ISSUE #6872: Zarr-Streaming from a datastore into itself is broken
When adding a zarr-streamed dataset served by a datastore *to that same datastore* as a remote dataset, the streaming is broken. This is not a typical use case, but still it is weird that this does not work. Should be investigated.

We saw this both on localhost and in production

### LINKED ISSUE #6906: Error handling in DatasetArray: move away from Exceptions, use Fox/flatMap

### LINKED ISSUE #7141: Should Fox have an Empty-preserving variant of `?~>`?
`?~>` wraps Failures and Emptys in a new Failure with the argument error message prepended in the chain.

Some use cases have come up when we only want to add the message in failure case, but leave Empty as is.

Should this become the default behavior? Or should there be two variants of this? We should have a look

fm3:
- VaultPath.readBytes now returns Fox[Array[Byte]]
 - Adapted usages and implementations accordingly
 - Used the actually async http client `play ws` in the HttpsDataVault
 - 404 is now propagated as Fox.empty and fill-value is inserted there. Other errors are propagated as errors (leading to neutral gray rendered by the front-end rather than the fill value’s black). Note that this only works for the HttpsDataVault at this point

### TODO
 - [x] adapt vault path and data vault array interfaces
 - [x] adapt usages
    - [x] adapt opening arrays and reading headers
    - [x] adapt reading chunks
 - [x] adapt tests
 - [x] use non-blocking http request in HttpsDataVault
   - [x] pass ws in there
   - [x] use ws to get info and data
   - [x] basic auth
   - [x] what about connect timeout?
   - [x] ensure this works with gzip and brotli encodings
 - [x] 404 should be propagated as Fox.empty and use fill value where appropriate
 - [x] merge master into this after #7067 is in there

- https://vaultreadfox.webknossos.xyz

### Steps to test:
- Explore and view some remote datasets with different stores/data types, should still show data as before
- Self-streaming should now work (create zarr link in annotation, explore that link in the same webknossos instance)

### Issues:
- fixes #6906 
- fixes #6872
- fixes #7141

------


fm3:
I was able to validate that both gzip and brotli work ok with the adapted HttpsDataVault. Looks like play ws does decode gzip itself, but it then also “eats” the encoding header, so our own decode code does not attempt to decode it again. On the other hand, when ws encounters an encoding it does not know how to decode, it keeps the header and does not decode. So that also works with our VaultPath code.

To test this I used this simple flask server:

```
from flask import Flask, make_response
import gzip
import brotli

app = Flask(__name__)
app.config["COMPRESS_REGISTER"] = False

@app.route("/")
def hello_world():
    return "Hello, World!"

@app.route("/gzip")
def test_gzip():
    s = "Hello, World!"
    b = s.encode('utf8')
    content = gzip.compress(b, 5)
    response = make_response(content)
    response.headers['Content-length'] = len(content)
    response.headers['Content-Encoding'] = 'gzip'
    return response

@app.route("/brotli")
def test_brotli():
    s = "Hello, World!"
    b = s.encode('utf8')
    content = brotli.compress(b)
    response = make_response(content)
    response.headers['Content-length'] = len(content)
    response.headers['Content-Encoding'] = 'br'
    return response
```

and the test route in commit [bdef9ce](https://github.com/scalableminds/webknossos/pull/7137/commits/bdef9cea5ddb27f13e9c6f00648bc50630fe42f5)

webknossos-datastore/app/com/scalableminds/webknossos/datastore/datareaders/zarr3/Zarr3Array.scala
```diff
@@ -1,47 +1,37 @@
...
-                   channelIndex,
-                   sharedChunkContentsCache)
-  }
-
+           sharedChunkContentsCache: AlfuCache[String, MultiArray])(implicit ec: ExecutionContext): Fox[Zarr3Array] =
+    for {
+      headerBytes <- (path / Zarr3ArrayHeader.ZARR_JSON)
+        .readBytes() ?~> s"Could not read header ${Zarr3ArrayHeader.ZARR_JSON}"
+      header <- JsonHelper.parseAndValidateJson[Zarr3ArrayHeader](headerBytes) ?~> "Could not parse array header"
+      _ <- DatasetArray.assertChunkSizeLimit(header.bytesPerChunk)
```
frcroth, line 26:
```suggestion
```
Can not be used here, since with the way sharding is implemented, zarr3 chunks can be way bigger (and still used, since the second chunking happens with the sharding codec).

webknossos-datastore/app/com/scalableminds/webknossos/datastore/datavault/Encoding.scala
```diff
@@ -1,19 +1,21 @@
...
 import com.scalableminds.util.enumeration.ExtendedEnumeration
+import net.liftweb.common.{Box, Failure, Full}
 
+// Lists the encodings supported by VaultPath.readBytes
 object Encoding extends ExtendedEnumeration {
   type Encoding = Value
-  val gzip, brotli, identity, unsupported = Value
+  val gzip, brotli, identity = Value
 
   // Header defined in https://datatracker.ietf.org/doc/html/rfc7231#section-3.1.2.2,
```
frcroth, line 11:
```suggestion
  // Header defined in https://datatracker.ietf.org/doc/html/rfc7231#section-3.1.2.2
```
just noticed that the link does not work

util/src/main/scala/com/scalableminds/util/tools/Fox.scala
```diff
@@ -246,9 +247,20 @@ object Fox extends FoxImplicits {
 class Fox[+A](val futureBox: Future[Box[A]])(implicit ec: ExecutionContext) {
   val self: Fox[A] = this
 
+  // Add error message in case of Failure and Empty (wrapping Empty in a Failure)
   def ?~>(s: String): Fox[A] =
     new Fox(futureBox.map(_ ?~! s))
 
+  // Add error message only in case of Failure, pass through Empty
+  def ?~~>(s: String): Fox[A] =
```
frcroth, line 255: Not totally happy with the name, can we find something more distinguishable?

frcroth (approved):

PR #7138, opened 2023-06-13
Quick fixes for overflow menu

dieknolle3333:
- there was a small error (wrong negotiation) when the reload button is rendered. it should be shown in the layer actions overflow menuif the dataset is editable.
- the overflow menu can now be opened just by hovering. thus I also deleted the tooltip.


hotzenklotz (approved):
LGTM

PR #7139, opened 2023-06-14
Larger play thread pool, async fossil client

### LINKED ISSUE #7140: Optimize ThreadPool usage, make more use of async IO
We should take a closer look at the threadpool configuration, compare https://www.playframework.com/documentation/2.8.x/ThreadPools

 - We should probably get rid of the Global Execution Context
 - We should probably increase the number of threads in the play thread pool, because we have some blocking IO
 - We should reduce the amount of blocking IO

What may be blocking
 - talking to postgres (though slick claims async, maybe it has its own thread pool?)
 - talking to redis (there may be alternative clients, such as lettuce or salad (I’m not making these up))
 - talking to dataset storage via https (fixed in #7137)
 - talking to google cloud storage, s3
 - talking to file system
 - talking to fossildb (wip in #7139)

fm3:
- Remove usages of ExecutionContext.global. [Using that discouraged because there are no limits to how many threads it spawns](https://stackoverflow.com/questions/30805337/plays-execution-contexts-vs-scala-global). Before this PR, we randomly used it whereever it was too hard to inject the play execution context.
 - Adds a sensible config for the play execution context. Before this PR, the default config was used, which yielded way too few threads for the amount of blocking operations we have.
 - Adapts most of the fossildb gprc client to async and adapts its error handling accordingly. With one exception, which is getMultipleKeys, because that is used in Iterators, which are harder to rewrite to async (this may be a future optimization)

### TODO
 - [x] get rid of global execution contest
 - [x] adapt akka config for some more threads
   - [x] also in datastore + tracingstore
 - [x] go for more async io usages
   - [x] async fossil client
     - [x] adapt health check
     - [x] adapt get request
     - [x] adapt other requests
     - [x] error handling (exceptions are thrown inside the future?)

- https://threadpools.webknossos.xyz

### Steps to test:
- Test that data loading for remote datasets still works
- Test that annotating (save, reload, download) still works
- To actually test the improvements in parallelism, I used the test route added in [4c96ce4](https://github.com/scalableminds/webknossos/pull/7139/commits/4c96ce4d3ffb04698460cfb8e2c205d03fcd1da2) – sending a lot of parallel requests to /tracings/test showed in the logging how the requests were answered in parallel

### Issues:
- fixes #7140

------

### DIFF
--- a/CHANGELOG.unreleased.md
+++ b/CHANGELOG.unreleased.md
@@ -20,6 +20,7 @@ For upgrade instructions, please check the [migration guide](MIGRATIONS.released
 - Optimized processing of parallel requests (new thread pool configuration, asynchronous FossilDB client), improving performance and reducing idle waiting. [#7139](https://github.com/scalableminds/webknossos/pull/7139)
 
 ### Fixed
+- Fixed rare rendering bug at viewport edge for anisotropic datasets. [#7163](https://github.com/scalableminds/webknossos/pull/7163)
 
 ### Removed
 
--- a/frontend/javascripts/oxalis/geometries/plane.ts
+++ b/frontend/javascripts/oxalis/geometries/plane.ts
@@ -12,13 +12,20 @@ import constants, {
   OrthoViewValues,
 } from "oxalis/constants";
 
-// A subdivision of 80 means that there will be 80 segments per axis
-// and thus 81 vertices per axis (i.e., the vertex shader is executed 81**2).
+// A subdivision of 100 means that there will be 100 segments per axis
+// and thus 101 vertices per axis (i.e., the vertex shader is executed 101**2).
 // In an extreme scenario, these vertices would have a distance to each other
 // of 32 voxels. Thus, each square (two triangles) would render one bucket.
-// 80**2 == 6400 buckets per plane are currently unrealistic and therefore
+// 100**2 == 10,000 buckets per plane are currently unrealistic and therefore
 // a valid upper bound.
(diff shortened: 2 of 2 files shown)


util/src/main/scala/com/scalableminds/util/xml/XMLWrites.scala
```diff
@@ -1,23 +0,0 @@
-package com.scalableminds.util.xml
-
-import javax.xml.stream.XMLStreamWriter
-
-import com.scalableminds.util.tools.Fox
-
-import scala.concurrent.ExecutionContext.Implicits.global
```
fm3, line 7: the whole file was unused

project/plugins.sbt
```diff
@@ -20,4 +20,4 @@ addSbtPlugin("com.sksamuel.scapegoat" %% "sbt-scapegoat" % "1.1.1")
 addSbtPlugin("net.vonbuchholtz" % "sbt-dependency-check" % "3.1.3")
 
 //protocol buffers
-libraryDependencies += "com.thesamet.scalapb" %% "compilerplugin" % "0.11.12"
+libraryDependencies += "com.thesamet.scalapb" %% "compilerplugin" % "0.11.13"
```
fm3, line 23: Upgrade not really needed for this, but shouldn’t hurt (I added it while debugging something which turned out to be caused elsewhere)

frcroth (approved):

PR #7151, opened 2023-06-16
Fix vault path for precomputed datasets

frcroth:
- https://___.webknossos.xyz

### Steps to test:
- Open any precomputed dataset (was broken previously)

### Issues:
- fixes #

------
- [ ] Updated [changelog](../blob/master/CHANGELOG.unreleased.md#unreleased) Necessary? Only fixes an issue introduced in the same release


normanrz (approved):
I assume you have tested this for a selection of datasets

fm3 (approved):
argh, this happend while resolving merge conflicts. sorry and thanks for the quick fix! (if this is the only usage of basePath, inlining `magPath.parent` would be nicer imo, but not important)

PR #7152, opened 2023-06-16
Add table showing artifacts disk space consumption to workflow view

MichaelBuessemeyer:
This PR adds a new table to the vx workflow view that enables users to identify the artifacts consuming the most disk space. The table is wrapped in a modal and can be shown by clicking the "Show Disk Usage of Artifacts" option in workflow view under the menu next to the "Collapse all button" in the header on the right side of the view.

- https://___.webknossos.xyz

### Steps to test:
- I suggest testing locally, as vx needs to be active on the wk instance.
- Enable voxelytics in the application.conf
- execute ` WK_URL=http://localhost:9000 WK_TOKEN=secretSampleUserToken vx voxelytics/connect/test/configs/segem_test_2d.yaml --cpu` in the working directory of vx and wait for the command to finish
- Go to `http://localhost:9000/workflows` and open the workflow.
- In the "..." menu-button select the option "Show Disk Usage of Artifacts"
- A modal should open up, that displays all artifacts with names, task name, artifact size, inode count, and path in table. Each column of the table should be sortable.
- The table should also work with meta tasks

### Issues:
- fixes [#7069](https://github.com/scalableminds/webknossos/issues/7069)

------
- [ ] Updated [documentation](../blob/master/docs) if applicable <-- is this needed?


MichaelBuessemeyer:
I'd say this pr is ready for the first review (despite the open TODO).

MichaelBuessemeyer:
I honestly don't know who should review these changes. @philippotto in case you want/know someone else that could review the changes better, please reassign the review

frontend/javascripts/admin/voxelytics/artifacts_disk_usage_list.tsx
```diff
@@ -0,0 +1,88 @@
...
+      return _.flatten(
+        taskGroup.subtasks
+          .map(taskToTableEntry)
+          .filter((artifactTableEntry) => artifactTableEntry != null),
+      );
+    }
+    return Object.entries(artifacts[taskGroup.taskName]).map(([artifactName, artifact]) => ({
+      artifactName,
+      artifactSize: formatCountToDataAmountUnit(artifact.fileSize),
+      fileSize: artifact.fileSize,
```
philippotto, line 42: isn't this the same? the first entry is formatted and the second is not?

philippotto (commented):
lgtm! only the see my one comment :)

frontend/javascripts/admin/voxelytics/artifacts_disk_usage_list.tsx
MichaelBuessemeyer: You are right: one is the formatted value and one is the unformatted raw value.
The unformatted raw value is used to be able to sort the artifact entries properly by their size.
The formatted value is simply the value that is displayed in the table.


I looked at the docs that list the `render` prop for the columns prop of a table. I updated the artifact entries to only have the file size and to format them in the render function.
This way is much better as the filesize is not contained in every entry twice. :+1: 

Thanks for mentioning this :pray:

MichaelBuessemeyer:
@philippotto Do you think I should test the artifact size table with meta tasks? If yes, I somehow need a way to get a workflow with a meta task into webknossos.
Running the vx tests with the WK_URL and WK_TOKEN being set did not make the run workflows accessible to wk :/

philippotto:
> @philippotto Do you think I should test the artifact size table with meta tasks? If yes, I somehow need a way to get a workflow with a meta task into webknossos. Running the vx tests with the WK_URL and WK_TOKEN being set did not make the run workflows accessible to wk :/

If it's too tedious to do the testing locally, I'm also fine with merging the PR and then directly having a look in production, since the chances are rather good that it just works :tm: :)

frontend/javascripts/admin/voxelytics/artifacts_disk_usage_list.tsx
philippotto: great :+1: did you push this already?

MichaelBuessemeyer: Sorry, I got interrupted a little. It is pushed now :)

philippotto (approved):
LGTM :+1:

MichaelBuessemeyer:
> If it's too tedious to do the testing locally, I'm also fine with merging the PR and then directly having a look in production, since the chances are rather good that it just works tm :)

Ok, let's just :shipit: :tm: (after the lunch break)

PR #7155, opened 2023-06-19
Second try for “Async IO for HttpsDataVault, Fox Error Handling”

fm3:
Reverts scalableminds/webknossos#7154

second try for #7137 

This time with separate DataVaultService (which does not need ds remote wk client, so it can be instantiated in the wk side even if there is no local datastore) and RemoteSourceDescriptorService (which does)

Could you have a look at the latest commit [fd179c5](https://github.com/scalableminds/webknossos/pull/7155/commits/fd179c5c5014b6ad0d96461404045d0e6590dac9) that includes this change?

I tested that this now launches even if no datastore runs (this failed previously in production) by adding `play.http.router = "noDS.Routes"` and `play.modules.disabled += "com.scalableminds.webknossos.datastore.DataStoreModule"` to application.conf, like it is in the production config. With this, I could still get a dashboard (but as expected could not load actual image data)


frcroth (approved):
how about a test?

PR #7161, opened 2023-06-19
Implement email verification

### LINKED ISSUE #7040: Ask users to confirm their email address
We should introduce a protocol to let users verify that they can receive emails via the address they supplied during signup

frcroth:
- https://___.webknossos.xyz (Does not make a lot of sense because you need to read the mails for testing)

### Steps to test:
- Sign up at webknossos.
- Notice that your mail is not yet verified (you can check this in the database)
  - If the config requires mail verification and the grace period is 0, you will notice that you can not sign in (even after the account was activated)
- Click the link in the mail (log mails to stdout)
- Your mail will be verified (you can check in db multiUser table)
- You can now log in.
- Clicking the link again results in an error.

### TODO

- [ ] Questions about the Verification flow
	- [ ] When do users get the verification mail? (Maybe only after activation by admin)
	- [ ] Add verification link to welcome mail?
	- [ ] When verification is required (settings), after the grace period, users are not allowed to sign in. Don't know if that makes sense (also resending verification link currently needs sign in, at least that should be possible)
- [x] ~~Once flow is clear, also write about it in docs~~
- [x] Write verification mail
- [x] Mark verification keys as used
- [x] What would a good default application.conf look like?
- [x] Update application.conf (Do this immediately before merging)
- [x] Frontend
- [x] Refresh snapshots (Does not work for me with `yarn refresh-all-snapshots`)

### Issues:
- fixes #7040

------


philippotto:
I implemented the front-end part so that the user can click on a verification link. In the happy path, a success message is shown. Otherwise, an error is shown with a "resend" link. In both cases, the user is directly redirected to the dashboard.

> Add verification link to welcome mail?

I think, verification links in welcome mails are more likely to be overlooked. Therefore, I'd rather send a dedicated "Please verify" email.

> When do users get the verification mail? (Maybe only after activation by admin)

No strong opinion. I think, it could happen directly after sign in. The sooner the better maybe?

> When verification is required (settings), after the grace period, users are not allowed to sign in. Don't know if that makes sense (also resending verification link currently needs sign in, at least that should be possible)

I think, this is ok, but I would simply send a new verification email automatically if a sign in fails due to missing verification. This saves the user from an additional click and makes the UI simpler, too. Maybe this needs throttling to avoid that somebody uses WK to spam another email address with verification mails?

-------------------------

Another note: In my opinion the verify route should check if the user is already verified as the very first check. Then, a 200 could simply be returned. Otherwise, it can be weird to get a "verification failed" error when clicking on an old link even though one is already verified.

conf/webknossos.latest.routes
```diff
@@ -278,6 +278,10 @@ POST          /credentials/httpBasicAuth
 POST          /credentials/s3AccessKey                                                          controllers.CredentialController.createS3AccessKeyCredential
 POST          /credentials/googleServiceAccount                                                 controllers.CredentialController.createGoogleServiceAccountCredential
 
+# Email verification
+GET           /verifyEmail/:key                                                                 controllers.EmailVerificationController.verify(key: String)
```
philippotto, line 282: this should be a POST probably.

frcroth:
@philippotto 

- I changed the route to POST
- When email verification fails (because the grace period is over), an email verification link is automatically resent. Throttling should not be necessary, since this requires pretty malicious intent. Logging was added to be able to identify if something like that were to happen
- Currently, if the user ignores the verification mail (or it ends up in spam), the user does not interact with the system again until they are no longer able to sign in. This is pretty harsh. Maybe the frontend could display a banner or a pop up on sign in which states that email verification has not happened yet? Perhaps with a button to resend the verification mail.
- When clicking an expired email verification link, the error message contains a "Resend verification mail" button. This does not work, as it requires the user to be signed in.

> Another note: In my opinion the verify route should check if the user is already verified as the very first check. Then, a 200 could simply be returned. Otherwise, it can be weird to get a "verification failed" error when clicking on an old link even though one is already verified.

This was done (although the frontend now misleadingly prints "Successfully verified your email")

philippotto:
@frcroth 

> * I changed the route to POST
> * When email verification fails (because the grace period is over), an email verification link is automatically resent. Throttling should not be necessary, since this requires pretty malicious intent. Logging was added to be able to identify if something like that were to happen

:+1: 
I didn't test it yet, but did you include an error message in that case which tells the user that a verification email was (re)sent?

> * Currently, if the user ignores the verification mail (or it ends up in spam), the user does not interact with the system again until they are no longer able to sign in. This is pretty harsh. Maybe the frontend could display a banner or a pop up on sign in which states that email verification has not happened yet? Perhaps with a button to resend the verification mail.

Yes, we can do that. How would the front-end know about this? Maybe include `isEmailVerified` in `publicWrites` of `User`?

> * When clicking an expired email verification link, the error message contains a "Resend verification mail" button. This does not work, as it requires the user to be signed in.

I changed it so that there's an error message telling the user to log in (if they tried to click resend without being logged in). If they can log in, they can click the link again. If they cannot log in, an email is resent automatically.

> > Another note: In my opinion the verify route should check if the user is already verified as the very first check. Then, a 200 could simply be returned. Otherwise, it can be weird to get a "verification failed" error when clicking on an old link even though one is already verified.
> 
> This has been implemented (although the frontend now misleadingly prints "Successfully verified your email")

I think, this is okay. The end result for the user is the same: the email is verified.

frcroth:
@philippotto 
>  I didn't test it yet, but did you include an error message in that case which tells the user that a verification email was (re)sent?
Yes
> Yes, we can do that. How would the front-end know about this? Maybe include isEmailVerified in publicWrites of User?
Added this now :+1:
> I changed it so that there's an error message telling the user to log in (if they tried to click resend without being logged in). If they can log in, they can click the link again. If they cannot log in, an email is resent automatically.

Yes, I think that works.

philippotto:
Cool, I added the verification warning now. It appears every time the active user is set (e.g., log in, registration but also on each full page load). Might be a bit annoying, but we really want that the user verifies the email. So, it's fine in my opinion.

> For possible frontend changes, do the application.conf settings need to be exposed somwhere (e.g. to say "If you don't verify in X days ...") or are they already automatically available?

Only the `features` block is automatically available afaik. However, I don't think we really need to explain the grace period. Email verification should be a no-brainer (and once lock-out happens it should be a matter of seconds to verify it since a new mail will arrive).

fm3:
Is this (backend-side) ready for review? If so, I can have a look tomorrow. I just saw that there are still quite a few unchecked boxes in the PR description so just making sure I’m not looking at this prematurely. Also, are there any open questions left? I did not yet read through the whole conversation. Should I refresh the snapshots for you?

frcroth:
@fm3 
> Is this (backend-side) ready for review? If so, I can have a look tomorrow. I just saw that there are still quite a few unchecked boxes in the PR description so just making sure I’m not looking at this prematurely. Also, are there any open questions left? I did not yet read through the whole conversation. Should I refresh the snapshots for you?

Yes backend is ready for review. The open checkboxes are questions to discuss (mostly design not implementation). I think the current design matches the expectations of @philippotto and me, but if you have differing opinions, they would be worthwhile to discuss.

conf/application.conf
```diff
@@ -200,7 +206,7 @@ proxy {
 
 # Send emails to users and admins
 mail {
-  logToStdout = false # To protect privacy, always set to false in production
+  logToStdout = true # To protect privacy, always set to false in production
```
fm3, line 209: reminder to set this back to false before merge

fm3 (commented):
Looking pretty good! I only added some minor code style comments.

I agree with philipps and your assessment on the user flow. I think the most important part is that the user always knows what is happening and gets readable error messages. Seems like you already took good care of that.

I did not test yet, I’d leave that for after the front-end review.

I’m currently refreshing the snapshots, will push when that has succeeded.

app/controllers/EmailVerificationController.scala
```diff
@@ -0,0 +1,33 @@
...
+class EmailVerificationController @Inject()(emailVerificationService: EmailVerificationService, sil: Silhouette[WkEnv])(
+    implicit ec: ExecutionContext,
+    val bodyParsers: PlayBodyParsers)
+    extends Controller
+    with FoxImplicits {
+
+  @ApiOperation(value = "")
+  def verify(key: String): Action[AnyContent] = Action.async { implicit request =>
+    for {
+      _ <- emailVerificationService.verify(key)(GlobalAccessContext, ec)
```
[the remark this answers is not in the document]
frcroth: Does not work:
```
[error] EmailVerificationController.scala:22:48: not enough arguments for method verify: (implicit ctx: com.scalableminds.util.accesscontext.DBAccessContext, implicit ec: scala.concurrent.ExecutionContext)com.scalableminds.util.tools.Fox[Unit].
[error] Unspecified value parameter ec.
[error]       _ <- emailVerificationService.verify(key)(GlobalAccessContext)
[error]
```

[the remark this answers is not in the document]
fm3: ok then I guess this is not available if one implicit parameter is passed explicitly. sorry!

frcroth:
> When changing a user's email address (Admin -> Users), we need to re-verify. (Maybe follow up issue?)

This is already implemeted. When user data is updated, the email verification is revoked (in the last commit I changed it so it is only invalidated if the email has actually changed

> I didn't like the email templates. Here is the OpenAI email in comparison: 
Agreed, I like the button and the old template was only a placeholder. However, wouldn't it make sense to include the link as well? That's something I often see "If the button does not work, use this link". I guess for non-HTML mail clients?

philippotto:
I just pushed a fix for the confusing toast when verifying the email. Let me know in case anything front-end related needs to happen.

conf/application.conf
```diff
@@ -65,6 +65,12 @@ webKnossos {
     timeTrackingPause = 60 seconds
     inviteExpiry = 14 days
     ssoKey = ""
+    emailVerification {
+      activated = true
```
fm3, line 69: should be false by default (for development)
(also reminder to reset mail.logToSstdout)

fm3 (commented):
Backend LGTM :)

Is it expected that the snapshots are different again now? (May be due to email address changes in the tests? Should I refresh the snapshots again?)

frcroth:
> Is it expected that the snapshots are different again now? (May be due to email address changes in the tests? Should I refresh the snapshots again?)

Yes, please refresh. I think it is because of changes in the initial data.

fm3:
Yes I think that would be helpful. Since tom is currently out of office, maybe @daniel-wer could have a look?

I also still see quite a few open checkboxes both in the PR on top and in some of the comments. Do you think that anything of that needs addressing @frcroth ?

daniel-wer (commented):
Frontend LGTM. Testing also went very well :+1: 

- [x] I wasn't able to check what the HTML-rendered email looks like, maybe you could upload a screenshot in this PR or confirm that someone checked that it renders correctly.

conf/application.conf
fm3: (for the record, we agreed that activated=true makes sense, and shouldn’t block anyone while required is false)

fm3 (approved):
Email looks alright to me :) 

image

Issue #5727, opened 2021-09-13
Show all meshes for dataset (or for bounding box)


philippotto:
## Detailed Description

Allow to load all precomputed meshes for a dataset or for a bounding box if the dataset is too large.

## Context

* [X] Specific to long-running jobs (set `jobsEnabled=true` in `application.conf`)
* [ ] Specific to webKnossos.org (set `isDemoInstance=true` in `application.conf`)

https://forum.image.sc/t/show-all-meshes/57316

manuel-berning:
Was just looking for this functionality!
What is the best option present right now to load all meshes in a single precomputed meshfile?
Is there anything better than loading single meshes via context or overflow menu as described here:
https://docs.webknossos.org/webknossos/mesh_visualization.html
See "Working with Meshes"

PR #7164, opened 2023-06-20
Segment group batch actions

### LINKED ISSUE #6054: Downsample volume layer modal crashes
## Context
volume annotation with only one magnification (which should be downsampled via the dialog)

## Current Behavior

image

```
console.js:13 Error: Layer with name undefined was not found.
    at y.getLayerByName (model.js:235)
    at y.getSegmentationTracingLayer (model.js:157)
    at ov.getVolumeMagsToDownsample (layer_settings_tab.js:684)
    at ov.render (layer_settings_tab.js:885)
    at rs (react-dom.production.min.js:182)
    at ts (react-dom.production.min.js:181)
    at $u (react-dom.production.min.js:263)
    at Fu (react-dom.production.min.js:246)
    at zu (react-dom.production.min.js:246)
    at Ou (react-dom.production.min.js:239)
```

## Steps to Reproduce the bug

1. create volume annotation with only one magnification (for example, temporarily remove mags from a segmentation layer; create volume annotation with fallback; readd the mags to the dataset)
2. click the orange "downsample" button in the left sidebar
3.
4.

## Your Environment for bug

### LINKED ISSUE #6954: Segments Groups
Similar to tree groups, there could be hierarchical segment groups. Group actions (via context menu) could be:

- [x] move selected segment into this group
- [ ] batch actions for groups #7164
  - [x] load/remove/refresh meshes for everything in a group 
  - [x] download all meshes in a group
  - [x] set color for all segments in group
  - [x] hide/show meshes
  - [ ] hide segments (not implemented yet)
- [ ] multi-selection

dieknolle3333:
### Steps to test:
- Go to the segments tab in any annotation
- Make sure you see some segments in the list, e.g. by clicking segments
- Either create a new group and drop segments or test the following on the root group:
-- Right-click a group and set the color
-- Revert setting the color
-- Load meshes either ad-hoc or precomputed with a meshfile. If there is no meshfile, there should only be the option to load the meshes ad-hoc
-- Hide/show meshes. The text of the option should be adjusted to the current visibility
-- Download all meshes of the group. You should receive a zip with the according stl files.

### Issues:
- fixes #6954
- Related to #5727

------


frontend/javascripts/oxalis/view/right-border-tabs/segments_tab/segments_view.tsx
```diff
@@ -975,6 +996,24 @@ class SegmentsView extends React.Component<Props, State> {
...
+                      disabled: false,
+                      label: (
+                        <ChangeColorMenuItemContent
+                          title="Change Group Color"
+                          isDisabled={false}
+                          onSetColor={(color) => {
+                            if (getVisibleSegmentationLayer == null) {
+                              return;
+                            }
+                            console.log(color);
```
dieknolle3333, line 1010: dont forget about me

frontend/javascripts/oxalis/model/sagas/isosurface_saga.ts
```diff
@@ -962,10 +964,53 @@ function* downloadIsosurfaceCellById(
   }
 }
 
+function* downloadIsosurfaceCellsAsZIP(
+  arrayOfCells: { cellName: string; segmentId: number; layerName: string }[],
```
philippotto, line 968: I'd simply call it `cells` or `segments` (so, without "array of")

frontend/javascripts/oxalis/model/sagas/isosurface_saga.ts
```diff
@@ -962,10 +964,53 @@ function* downloadIsosurfaceCellById(
...
+        return;
+      }
+      const stlDataViews = exportToStl(geometry);
+      // Encode isosurface and cell id property
+      const { isosurfaceMarker, segmentIdIndex } = stlIsosurfaceConstants;
+      isosurfaceMarker.forEach((marker, index) => {
+        stlDataViews[0].setUint8(index, marker);
+      });
+      stlDataViews[0].setUint32(segmentIdIndex, element.segmentId, true);
+      const stlDataReader = new BlobReader(new Blob(stlDataViews));
```
philippotto, line 993: This is copied from somewhere, right? Can you extract the common code into a helper function?

frontend/javascripts/oxalis/model/sagas/isosurface_saga.ts
```diff
@@ -962,10 +964,53 @@ function* downloadIsosurfaceCellById(
...
+      return zipWriter.add(`${element.cellName}-${element.segmentId}.stl`, stlDataReader);
+    });
+    yield all(promises);
+    const result = yield* call([zipWriter, zipWriter.close]);
+    saveAs(result, "mesh-export.zip");
```
philippotto, line 998:
```suggestion
    yield* call(saveAs, result, "mesh-export.zip");
```

frontend/javascripts/oxalis/view/right-border-tabs/segments_tab/segments_view.tsx
```diff
@@ -768,6 +781,365 @@ class SegmentsView extends React.Component<Props, State> {
     </>
   );
 
+  getToastForMissingLocations = (groupId: number) => {
+    const segmentsWithoutPosition = this.getSegmentsWithMissingLocation(groupId);
+    if (segmentsWithoutPosition.length > 0) {
+      console.log(`Segments with unknown positions: ${segmentsWithoutPosition}`);
+      return Toast.info(
+        <React.Fragment>
+          {segmentsWithoutPosition.length} meshes couldn't be loaded because the segment position is
```
philippotto, line 790: this will read "1 meshes couldnt..." if length === 1. you can use the function `pluralize` for that.

frontend/javascripts/oxalis/view/right-border-tabs/segments_tab/segments_view.tsx
```diff
@@ -768,6 +781,365 @@ class SegmentsView extends React.Component<Props, State> {
...
+        <ChangeColorMenuItemContent
+          title="Change Group Color"
+          isDisabled={isEditingDisabled}
+          onSetColor={(color) => {
+            if (getVisibleSegmentationLayer == null) {
+              return;
+            }
+            this.setGroupColor(id, color);
+          }}
+          rgb={[0.5, 0.5, 0.5]}
```
philippotto, line 812: how about defaulting to the color of the first segment here? especially, if all segments already have the same color, it can be weird to default to gray here.

frontend/javascripts/oxalis/view/right-border-tabs/segments_tab/segments_view.tsx
```diff
@@ -768,6 +781,365 @@ class SegmentsView extends React.Component<Props, State> {
...
+                  }
+                  this.handleLoadMeshesAdHoc(id);
+                  this.getToastForMissingLocations(id);
+                  this.handleSegmentDropdownMenuVisibility(id, false);
+                }}
+              >
+                ad hoc
+              </div>
+            ),
+          },
```
philippotto, line 845: you can extract this into a variable and reuse that in the else-branch. the differing keys can be unified (it's only important that the keys are unique in the surrounding array).

frontend/javascripts/oxalis/view/right-border-tabs/segments_tab/segments_view.tsx
```diff
@@ -768,6 +781,365 @@ class SegmentsView extends React.Component<Props, State> {
...
+  getReloadMenuItem = (groupId: number): ItemType => {
+    return this.state != null && this.doesGroupHaveAnyMeshes(groupId)
+      ? {
+          key: "reloadMeshes",
+          label: (
+            <div
+              onClick={() => {
+                if (this.props.visibleSegmentationLayer == null) {
+                  return;
+                }
```
philippotto, line 896: I suspect that this guard (here and probably in all the other onclick handlers) can be removed, because visibleSegmentationLayer is not used in the onclick handler per se. instead, the functions which are called (e.g., handleRefreshMeshes) do the same check again.

frontend/javascripts/oxalis/view/right-border-tabs/segments_tab/segments_view.tsx
```diff
@@ -768,6 +781,365 @@ class SegmentsView extends React.Component<Props, State> {
...
+      : null;
+  };
+
+  setGroupColor(groupId: number, color: Vector3) {
+    const { visibleSegmentationLayer } = this.props;
+    if (visibleSegmentationLayer == null) return;
+    const segmentGroupToChangeColor = this.getSegmentsOfGroup(groupId);
+    if (segmentGroupToChangeColor == null) return;
+
+    const actionArray = segmentGroupToChangeColor.map((segment) =>
```
philippotto, line 1015: `actionArray` -> `actions`

frontend/javascripts/oxalis/view/right-border-tabs/segments_tab/segments_view.tsx
```diff
@@ -768,6 +781,365 @@ class SegmentsView extends React.Component<Props, State> {
...
+    if (segmentGroup == null) return;
+
+    segmentGroup.forEach((segment) => {
+      if (Store.getState().localSegmentationData[layerName].isosurfaces[segment.id] != null) {
+        Store.dispatch(updateIsosurfaceVisibilityAction(layerName, segment.id, isVisible));
+      }
+    });
+
+    const copyOfState = this.state.areSegmentsInGroupVisible;
+    copyOfState[groupId] = isVisible;
```
philippotto, line 1069: this is not a copy. instead, you are only binding the variable `copyOfState` to the object `areSegmentsInGroupVisible` and mutate it then. a "copy" via variable assignment is only done for primitive variable types.

what you probably want to do is something like this (not tested):

```suggestion
    const newSegmentsInGroupVisible = {
       ...this.state.areSegmentsInGroupVisible,
       [groupId]: isVisible
    }
```

frontend/javascripts/oxalis/view/right-border-tabs/segments_tab/segments_view.tsx
```diff
@@ -768,6 +781,365 @@ class SegmentsView extends React.Component<Props, State> {
...
+      if (segment.somePosition == null || this.props.currentMeshFile == null) return;
+      this.props.loadPrecomputedMesh(
+        segment.id,
+        segment.somePosition,
+        this.props.currentMeshFile.meshFileName,
+      );
+    });
+
+    const copyOfState = this.state.areSegmentsInGroupVisible;
+    copyOfState[groupId] = true;
```
philippotto, line 1119: same here

frontend/javascripts/oxalis/view/right-border-tabs/segments_tab/segments_view.tsx
```diff
@@ -1136,6 +1511,24 @@ class SegmentsView extends React.Component<Props, State> {
...
 
+  doesGroupHaveAnyMeshes = (groupId: number): boolean => {
+    const { visibleSegmentationLayer } = this.props;
+    if (visibleSegmentationLayer == null) return false;
+    const segmentGroup = this.getSegmentsOfGroup(groupId);
+    if (segmentGroup == null) return false;
+    let bool = false;
+    segmentGroup.forEach((segment) => {
+      if (
+        Store.getState().localSegmentationData[visibleSegmentationLayer.name].isosurfaces[
```
philippotto, line 1522: you could extract `Store.getState().localSegmentationData[visibleSegmentationLayer.name].isosurfaces` into a variable (outside of the forEach loop) to save some look ups. I think, I saw the same pattern somewhere else, too..

frontend/javascripts/oxalis/view/right-border-tabs/segments_tab/segments_view.tsx
```diff
@@ -333,6 +343,9 @@ class SegmentsView extends React.Component<Props, State> {
     if (features().jobsEnabled) {
       this.pollJobData();
     }
+    this.props.segmentGroups.forEach(
+      (group) => (this.state.areSegmentsInGroupVisible[group.groupId] = false),
+    );
```
philippotto, line 348: I think you can remove that code. 1) you are directly mutating the state even though it should be updated immutable (see my other comment for that). 2) the default of `false` is not really necessary, since it doesn't matter whether the value is false or whether the key is not in the object.

frontend/javascripts/oxalis/view/right-border-tabs/segments_tab/segments_view.tsx
```diff
@@ -768,6 +781,365 @@ class SegmentsView extends React.Component<Props, State> {
...
+        Store.getState().localSegmentationData[visibleSegmentationLayer.name].isosurfaces[
+          segment.id
+        ] != null
+      ) {
+        Store.dispatch(removeIsosurfaceAction(visibleSegmentationLayer.name, segment.id));
+      }
+    });
+  };
+
+  handleChangeMeshVisibility = (layerName: string, groupId: number, isVisible: boolean) => {
```
philippotto, line 1056: the method name should reflect that it details with the meshes within a group. currently, it sounds like it would change the visibility of one mesh.

philippotto (changes requested):
very cool stuff :+1: I left a couple of notes but overall it looks pretty good. other than the  refactoring of the per-segment-call code about which we already talked, there is one thing I'm a bit hesitant about: `areSegmentsInGroupVisible`. this is state you are maintaining in your component, but it can be affected from the outside. for example, if I rightclick into the data viewport and render a mesh, the segment list won't take note of that right now. similarily, if I use the checkbox next to a mesh to hide it, the list component will also run into an outdated state.

instead, you should probably derive this information from the `IsosurfaceInformation` objects in `this.props.isosurfaces[segment.id]` (it has a `isVisible` prop). you might want to use `getDerivedStateFromProps` for that so that you can build your `areSegmentsInGroupVisible` state everytime the isosurfaces change.

dieknolle3333:
@philippotto thank you for the review!

I'll just quickly respond to the topic on `areSegmentsInGroupVisible`: 
In my view it was okay if the visibility of some meshes changed without the field noticing, because it was meant as a map for each group, to remember whether the last interaction made all meshes in a group visible or invisible. But you are totally right of course, that the state of the meshes matters if it was manipulated from another place. One example would be groups with only one mesh.

I thought about extracting the visibility from the meshes but I wasn't sure what a good metric in a mixed-visibility group is. If I use the actual isosurface data, is it fine to derive a field called `areAnySegmentsInGroupVisible`? So if any mesh in the group is visible, clicking the action hides all meshes. Or is it better the other way around? (If any mesh in the group is invisible, clicking the action makes everything visible?) I think intuitively I would prefer the second option.

philippotto:
> I thought about extracting the visibility from the meshes but I wasn't sure what a good metric in a mixed-visibility group is. If I use the actual isosurface data, is it fine to derive a field called `areAnySegmentsInGroupVisible`? So if any mesh in the group is visible, clicking the action hides all meshes. Or is it better the other way around? (If any mesh in the group is invisible, clicking the action makes everything visible?) I think intuitively I would prefer the second option.

Yes, i think it's fine to deal with mixed visibilities like you suggested. I don't have a strong opinion whether hide or show are "preferred". If you find the one variant more intuitive, let's go with that :)

frontend/javascripts/oxalis/view/right-border-tabs/segments_tab/segments_view.tsx
dieknolle3333: right, good idea, thanks!

frontend/javascripts/oxalis/view/right-border-tabs/segments_tab/segments_view.tsx
dieknolle3333: oh right, I think it's something you told me about in the very beginning. thanks for the reminder!

dieknolle3333:
The only thing that is blocking me from requesting another review is that I dont know how to properly mock zip.js.

frontend/javascripts/admin/dataset/dataset_upload_view.tsx
```diff
@@ -490,16 +490,16 @@ class DatasetUploadView extends React.Component<PropsWithFormAndRouter, State> {
 
       if (fileExtension === "zip") {
         try {
-          const reader = new ZipReader(new BlobReader(file));
+          const reader = new Zip.ZipReader(new Zip.BlobReader(file));
           const entries = await reader.getEntries();
           await reader.close();
-          const wkwFile = entries.find((entry: Entry) =>
+          const wkwFile = entries.find((entry: any) =>
```
dieknolle3333, line 496: I know this isn't pretty, but Zip.Entry doesnt work because the import of the wrapper is somehow not found

frontend/javascripts/oxalis/view/right-border-tabs/segments_tab/segments_view.tsx
```diff
@@ -798,22 +836,366 @@ class SegmentsView extends React.Component<Props, State> {
...
+      updateSegmentAction(segment.id, { color: color }, visibleSegmentationLayer.name),
+    );
+
+    Store.dispatch(batchUpdateGroupsAndSegmentsAction(actions));
+  }
+
+  handleRefreshMeshes = (groupId: number) => {
+    const { visibleSegmentationLayer } = this.props;
+    if (visibleSegmentationLayer == null) return;
+    //TODO I didnt find a better way to call the visiblesegmentationlayer. one other way would be to pass it into the function. but thats not useful for functions below-
```
hotzenklotz, line 1069: @philippotto Maybe you have another idea how to access this.

frontend/javascripts/oxalis/view/right-border-tabs/segments_tab/segments_view.tsx
```diff
@@ -798,22 +836,366 @@ class SegmentsView extends React.Component<Props, State> {
...
+      areSegmentsInGroupVisible: newSegmentsInGroupVisible,
+    });
+  };
+
+  getSegmentsWithMissingLocation = (groupId: number): number[] => {
+    const segmentGroup = this.getSegmentsOfGroup(groupId);
+    if (segmentGroup == null) return [];
+    let segmentIdsWithoutPosition: number[] = [];
+    segmentGroup.forEach((segment) => {
+      if (segment.somePosition == null) segmentIdsWithoutPosition.push(segment.id);
```
hotzenklotz, line 1133: Line 1131 - 1134 can be condensed a `reduce()` call.

frontend/javascripts/oxalis/view/right-border-tabs/segments_tab/segments_view.tsx
```diff
@@ -798,22 +836,366 @@ class SegmentsView extends React.Component<Props, State> {
...
+      [groupId]: true,
+    };
+    this.setState({
+      areSegmentsInGroupVisible: newSegmentsInGroupVisible,
+    });
+  };
+
+  downloadAllMeshesForGroup = (groupId: number) => {
+    const { visibleSegmentationLayer } = this.props;
+    if (visibleSegmentationLayer == null) return;
```
hotzenklotz, line 1167: Combine both return conditions into one conditional statement to keep it short

frontend/javascripts/oxalis/view/right-border-tabs/segments_tab/segments_view.tsx
```diff
@@ -921,6 +1303,17 @@ class SegmentsView extends React.Component<Props, State> {
...
   }
 
+  getSegmentsOfGroup = (groupId: number): Segment[] | null => {
+    const { segments, segmentGroups } = this.props;
+
+    if (segments == null || segmentGroups == null) {
+      return null;
+    }
+    const groupToSegmentsMap = createGroupToSegmentsMap(segments);
+    const segmentsOfGroup = groupToSegmentsMap[groupId] != null ? groupToSegmentsMap[groupId] : [];
```
hotzenklotz, line 1313: You can directly return line 1313

frontend/javascripts/oxalis/view/right-border-tabs/segments_tab/segments_view.tsx
```diff
@@ -1195,6 +1581,22 @@ class SegmentsView extends React.Component<Props, State> {
...
 
+  doesGroupHaveAnyMeshes = (groupId: number): boolean => {
+    const { visibleSegmentationLayer } = this.props;
+    if (visibleSegmentationLayer == null) return false;
+    const segmentGroup = this.getSegmentsOfGroup(groupId);
+    if (segmentGroup == null) return false;
+    const isosurfacesOfLayer =
+      Store.getState().localSegmentationData[visibleSegmentationLayer.name].isosurfaces;
+    let bool = false;
+    segmentGroup.forEach((segment) => {
```
hotzenklotz, line 1592: Sound like a good use for `Array.some(...)` 
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some

frontend/javascripts/oxalis/view/right-border-tabs/segments_tab/segments_view.tsx
```diff
@@ -396,16 +409,41 @@ class SegmentsView extends React.Component<Props, State> {
...
 
-      return {
+      updateStateObject = {
         groupTree: generatedGroupTree,
         searchableTreeItemList,
         prevProps: nextProps,
       };
-    } else {
+    }
+    if (prevState.prevProps?.isosurfaces !== isosurfaces) {
```
hotzenklotz, line 418: I am not sure I understand the reason for this code. Can you elaborate please. How can I test this?

frontend/javascripts/oxalis/model/sagas/isosurface_saga.ts
```diff
@@ -962,10 +957,57 @@ function* downloadIsosurfaceCellById(
   }
 }
 
+function* downloadIsosurfaceCellsAsZIP(
+  cells: { cellName: string; segmentId: number; layerName: string }[],
+): Saga<void> {
+  const { segmentMeshController } = getSceneController();
+  const zipWriter = new Zip.ZipWriter(new Zip.BlobWriter("application/zip"));
+  try {
+    const promises = cells.map((element) => {
```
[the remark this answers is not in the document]
dieknolle3333: there is a promise for each file that's added to the zipWriter. (for reference: https://gildas-lormeau.github.io/zip.js/api/index.html) so I think `addFileToZipWriterPromises` is long but correct.

frontend/javascripts/oxalis/view/right-border-tabs/segments_tab/segments_view.tsx
```diff
@@ -396,16 +409,41 @@ class SegmentsView extends React.Component<Props, State> {
...
+      segmentGroupsWithRoot.forEach((group) => {
+        let isVisible = true;
+        const segmentsOfGroup = groupToSegmentsMap[group.groupId];
+        if (segmentsOfGroup == null) return;
+        const visibilityOfSegments: number[] = segmentsOfGroup.map((segment) => {
+          const segmentIsosurface = isosurfaces[segment.id];
+          if (segmentIsosurface == null) return 1; //only care about explicitly invisible (not unloaded) meshes
+          return isosurfaces[segment.id].isVisible ? 1 : 0;
+        });
+        if (visibilityOfSegments.reduce((a, b) => a + b, 0) < segmentsOfGroup.length) {
```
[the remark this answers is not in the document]
dieknolle3333: oh, nice, thank you for the hint!

frontend/javascripts/oxalis/view/right-border-tabs/segments_tab/segments_view.tsx
```diff
@@ -798,22 +836,366 @@ class SegmentsView extends React.Component<Props, State> {
...
+              >
+                precomputed
+              </div>
+            ),
+          },
+        ],
+      };
+    } else {
+      return this.loadAdHocMenuItem(
+        "Load meshes for segments (ad-hoc)",
```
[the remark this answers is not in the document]
dieknolle3333: makes total sense, I'll do that!

frontend/javascripts/oxalis/view/right-border-tabs/segments_tab/segments_view.tsx
```diff
@@ -798,22 +836,366 @@ class SegmentsView extends React.Component<Props, State> {
...
+              {this.state.areSegmentsInGroupVisible[groupId] ? "Hide" : "Show"} all meshes
+            </div>
+          ),
+        }
+      : null;
+  };
+
+  setGroupColor(groupId: number, color: Vector3) {
+    const { visibleSegmentationLayer } = this.props;
+    if (visibleSegmentationLayer == null) return;
```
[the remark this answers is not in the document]
dieknolle3333: I thought it would be nice to avoid any further computation of the `groupToSegmentMap` if the `visibleSegmentationLayer` is null anyway. But you're saying this doesn't have much impact?

frontend/javascripts/oxalis/view/right-border-tabs/segments_tab/segments_view.tsx
dieknolle3333: I had the same thought as above so I'm looking forward to an answer!

frontend/javascripts/oxalis/view/right-border-tabs/segments_tab/segments_view.tsx
dieknolle3333: This code makes sure that changing the visibility of a mesh (e.g. by clicking the checkbox) gets noticed by the segments_view as the visibility is part of the isosurface. 
This is needed because I toggle the "Hide/Show meshes" option based on the mesh visibility in the group. 
You can test it by loading all meshes in a group. Now, the context menu should say "Hide...". If you deselect any mesh's  visibility checkbox, the context menu notices and will then read "Show all meshes".
Does that make sense?

dieknolle3333: also see philipps comment for this: https://github.com/scalableminds/webknossos/pull/7164#pullrequestreview-1507178411

frontend/javascripts/oxalis/view/right-border-tabs/segments_tab/segments_view.tsx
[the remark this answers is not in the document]
hotzenklotz: Good point. I did not think about the compute cost. I guess it does not matter much either way. Fine to leave as is.

frontend/javascripts/oxalis/view/right-border-tabs/segments_tab/segments_view.tsx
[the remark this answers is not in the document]
dieknolle3333: Loading vs. computing was the only occurence where I noticed a difference. I also changed "Change Group Color" to "Change Segment Color". Was there another action where the labeling is inconsistent?

frontend/javascripts/admin/dataset/dataset_upload_view.tsx
philippotto: I just pushed a commit (I hope the CI passes), which improves the typing a bit.

frontend/javascripts/oxalis/view/right-border-tabs/segments_tab/segments_view.tsx
```diff
@@ -406,16 +414,41 @@ class SegmentsView extends React.Component<Props, State> {
...
-      return {
+      updateStateObject = {
         groupTree: generatedGroupTree,
         searchableTreeItemList,
         prevProps: nextProps,
       };
-    } else {
+    }
+    if (prevState.prevProps?.isosurfaces !== isosurfaces) {
+      //if any segment is invisible, set visibility false, so that the first action is to make all meshes visible
```
philippotto, line 424:
```suggestion
      // Derive the areSegmentsInGroupVisible state so that we know per group
      // if it contains only visible elements. This is used to know whether "Show meshes" or
      // "Hide meshes" context item should be shown.
      // If any segment is invisible, set the visibility of the group to false, so that the preferred
      // action is to make all meshes visible.
```

frontend/javascripts/oxalis/view/right-border-tabs/segments_tab/segments_view.tsx
```diff
@@ -406,16 +414,41 @@ class SegmentsView extends React.Component<Props, State> {
...
+      const newVisibleMap: { [groupId: number]: boolean } = {};
+      const segmentGroupsWithRoot = [...segmentGroups, rootGroup];
+      segmentGroupsWithRoot.forEach((group) => {
+        let isVisible = true;
+        const segmentsOfGroup = groupToSegmentsMap[group.groupId];
+        if (segmentsOfGroup == null) return;
+        const visibilityOfSegments: number[] = segmentsOfGroup.map((segment) => {
```
philippotto, line 431: Could this be simplified to something like this?

```ts
        const isSomeSegmentLoadedAndInvisible = segmentsOfGroup.some((segment) => {
          const segmentIsosurface = isosurfaces[segment.id];
          // Only regard loaded, but invisible meshes
          return segmentIsosurface != null && !isosurfaces[segment.id].isVisible;
        });

        newVisibleMap[group.groupId] = !isSomeSegmentLoadedAndInvisible;
```

(not tested)

frontend/javascripts/oxalis/view/right-border-tabs/segments_tab/segments_view.tsx
```diff
@@ -1039,20 +1046,20 @@ class SegmentsView extends React.Component<Props, State> {
...
+    Store.dispatch(batchUpdateGroupsAndSegmentsAction(actions));
   }
 
   handleRefreshMeshes = (groupId: number) => {
     const { visibleSegmentationLayer } = this.props;
     if (visibleSegmentationLayer == null) return;
-    const segmentGroup = this.getSegmentsOfGroup(groupId);
-    if (segmentGroup == null) return;
+    //TODO I didnt find a better way to call the visiblesegmentationlayer. one other way would be to pass it into the function. but that's not useful for functions below-
+    // so my solution is a little more code duplication. putting it higher up isn't really a solution.
```
philippotto, line 1070: I'm not sure what you mean by "call the visiblesegmentationlayer" ? Accessing it via `const { visibleSegmentationLayer } = this.props;` seems fine to me if that's what you are asking about?

frontend/javascripts/oxalis/view/right-border-tabs/segments_tab/segments_view.tsx
```diff
@@ -1080,60 +1084,60 @@ class SegmentsView extends React.Component<Props, State> {
...
-    const copyOfState = this.state.areSegmentsInGroupVisible;
-    copyOfState[groupId] = isVisible;
+    const newSegmentsInGroupVisible = {
+      ...this.state.areSegmentsInGroupVisible,
+      [groupId]: isVisible,
+    };
     this.setState({
-      areSegmentsInGroupVisible: copyOfState,
+      areSegmentsInGroupVisible: newSegmentsInGroupVisible,
     });
```
philippotto: are you sure that this is necessary? Since you are computing areSegmentsInGroupVisible in getDerivedStateFromProps, I would assume that the state is automatically updated as soon as the store changes (which is done when you update the visibility of an isosurface). I'd think, that getDerivedStateFromProps is the only place where  areSegmentsInGroupVisible needs to be set. Could you test this? If this is the case, the setState code here and in the other handle* functions can be removed.

philippotto (changes requested):
great stuff :) I left some suggestions for simplifying some things further :crossed_fingers:

frontend/javascripts/admin/dataset/dataset_upload_view.tsx
dieknolle3333: nice, thank you!

frontend/javascripts/oxalis/view/right-border-tabs/segments_tab/segments_view.tsx
dieknolle3333: right, accessing would be the right word here. 
In `handlePerSegment` I do the exact check as in these lines, and I was wondering whether I could circumvent this check right here.

frontend/javascripts/oxalis/view/right-border-tabs/segments_tab/segments_view.tsx
dieknolle3333: oh right, this is so much nicer

frontend/javascripts/oxalis/view/right-border-tabs/segments_tab/segments_view.tsx
philippotto: ah I see. since the repeated lines include a return-statement which influences the control flow, I don't see a simple way to DRY this. Therefore, I think it's fine to keep it as is :) So, you can remove the todo comment.

frontend/javascripts/oxalis/view/right-border-tabs/segments_tab/segments_view.tsx
```diff
@@ -1098,41 +1096,21 @@ class SegmentsView extends React.Component<Props, State> {
...
-    const newSegmentsInGroupVisible = {
-      ...this.state.areSegmentsInGroupVisible,
-      [groupId]: true,
-    };
-    this.setState({
-      areSegmentsInGroupVisible: newSegmentsInGroupVisible,
-    });
   };
 
   getSegmentsWithMissingLocation = (groupId: number): number[] => {
```
philippotto, line 1108: I'd replace `Location` with `Position` here to keep it consistent with the `somePosition` attribute.

frontend/javascripts/libs/zipjs_wrapper.ts
```diff
@@ -0,0 +1,14 @@
+class TransFormStream {}
+
+let mockedWindow = false;
```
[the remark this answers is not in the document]
dieknolle3333: Omitting it doesn't seem to break anything, so I left it out.
Just for clarification. Technically this *could* have been used somewhere else because the scope of the variable is global, right? (https://www.w3schools.com/js/js_scope.asp)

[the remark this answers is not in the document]
philippotto: > Technically this could have been used somewhere else because the scope of the variable is global, right? (https://www.w3schools.com/js/js_scope.asp)

The linked article is a bit confusing (generally, w3schools ist not the best resource; I recommend MDN instead). However, the specific [MDN article](https://developer.mozilla.org/en-US/docs/Glossary/Scope?retiredLocale=de) is also not perfect (it mentions module scope, but doesn't really explain when it comes into play).
To answer your question: no, it could not have been used somewhere else because the variable is module-scoped. That scoping is achieved during the compilation/bundling of all the module files. To get a truly global variable, one has to extend `window` (in a browser context) or `global` (in a node context).

philippotto (approved):
Great :) Code looks good to me and it works great, too :+1:

PR #7169, opened 2023-06-22
Add distinction for light mode / dark mode icons for right sidebar where applicable

fm3:
- add latest design iteration by @albane-dlv
 - ensure the same canvas size in svg and css to avoid unwanted scaling of the icons
 - dark mode / light mode icons for mouse and key

- https://sidebariconsdarktheme.webknossos.xyz

### Steps to test:
- Look at the right sidebar icons both for annotation and dataset view mode, in light theme and dark theme
- should look pretty


hotzenklotz:
I am not a big fan of having different icons for dark and light mode. We should have a design that works for both out of the box.

public/images/icon-mouse-left-light.svg
hotzenklotz, line 1: Albane's latest design used a semi-transparent mouse button fill that works well in dark and light mode. I don't see the need for two separate icons.

image

image

PR #7173, opened 2023-06-23
Remove globalize floodfill job

### LINKED ISSUE #6126: Unify code of globalize floodfill modal and the starting jobs modals
## Detailed Description

- Currently, the code for the globalize floodfill modal (`bounding_box_tab.js`) is separate from the starting jobs modals (`starting_job_modals.js`). The idea of the `starting_job_modals.js` file is to contain all modals to start long-running jobs with. So it would be nice if the modal for the globalize floodfill could be merged into the `starting_job_modals.js` file. It would be possible to do that in #6086 but this would delay this pr being merged. Therefore I suggest doing this change separately on another branch.

### LINKED ISSUE #7172: Remove globalize floodfill job
## Detailed Description
The job is currently clunky to use (it is slow and produces a new dataset which is also not very user friendly). Instead, using the fill tool multiple times or using the proofreading tool should be the way to go.

philippotto:
- https://___.webknossos.xyz

### Steps to test:
- Use fill tool in 3D and check that no globalize UI is there anymore

### Issues:
- fixes #7172
- fixes #6126

------


philippotto:
@fm3 Could you remove the back-end code related to the globalize floodfill job? :innocent:

frontend/javascripts/oxalis/model/sagas/volumetracing_saga.tsx
```diff
@@ -463,7 +462,9 @@ export function* floodFill(): Saga<void> {
       yield* put(
         addUserBoundingBoxAction({
           boundingBox: coveredBoundingBox,
-          name: getBBoxNameForPartialFloodfill(oldSegmentIdAtSeed, activeCellId, seedPosition),
+          name: `Limits of flood-fill (source_id=${oldSegmentIdAtSeed}, target_id=${activeCellId}, seed=${seedPosition.join(
```
hotzenklotz, line 465: I am correctly assuming that this code is for the local, frontend only flood-fill operation?

hotzenklotz (approved):
LGTM

frontend/javascripts/oxalis/model/sagas/volumetracing_saga.tsx
philippotto: Exactly :+1:

PR #7174, opened 2023-06-23
Always show proofreading tool + refactoring

### LINKED ISSUE #7098: Always show proofreading tool (even when disabled)
This makes the tool more visible to the user / easier to find. If it's disabled, there should be an explanation about it and ideally a CTA to reach out to us (since the tool currently requires an editable mapping which not everybody has).

philippotto:
The proofreading tool is always shown (for hybrid annotations) so that users can find this tool better. In case of a missing agglomerate file, a tool tip tells the user to contact the WK team.

image

Additional changes:
- refactoring/improvement of the agglomerate disabled message
- tool names are always shown in tool tip (even when tool is disabled)
- improve styling of disabled buttons (e.g., on hover)

- https://___.webknossos.xyz

### Steps to test:
- try to activate the different tools in the different scenarios (e.g., brush is not available on coarser mags; proofreading requires agglomerate file etc)

### Issues:
- fixes #7098 

------

### DIFF
--- a/frontend/javascripts/messages.tsx
+++ b/frontend/javascripts/messages.tsx
@@ -237,7 +237,7 @@ instead. Only enable this option if you understand its effect. All layers will n
   "tracing.agglomerate_skeleton.no_agglomerate_file_active":
     "Loading a skeleton for a segment only works with agglomerate file mappings.",
   "tracing.agglomerate_skeleton.no_agglomerate_file_available":
-    "No agglomerate file mapping is available for this segmentation layer. Please reach out to hello@scalableminds.com to get help with generating one.",
+    "No agglomerate file mapping is available for this segmentation layer. Please reach out to hello@webknossos.org to get help with generating one.",
   "tracing.agglomerate_skeleton.no_skeleton_tracing":
     "Loading a skeleton for a segment only works in skeleton or hybrid tracings.",
   "tracing.skeletons_are_hidden_warning":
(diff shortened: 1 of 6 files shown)


philippotto:
@normanrz See screenshot in PR description. Do you think this is enough to point users into the right direction?

normanrz (approved):

frontend/javascripts/oxalis/model/sagas/mapping_saga.ts
```diff
@@ -119,23 +167,16 @@ function* maybeFetchMapping(
...
-    call(getMappingsForDatasetLayer, ...params),
-    call(getAgglomeratesForDatasetLayer, ...params),
-  ]);
+
   // Make sure the available mappings are persisted in the store if they are not already
   const areServerHdf5MappingsInStore =
     "agglomerates" in layerInfo && layerInfo.agglomerates != null;
-  if (!areServerHdf5MappingsInStore) {
-    yield* put(setLayerMappingsAction(layerName, jsonMappings, serverHdf5Mappings));
-  }
```
philippotto, line 138: @daniel-wer In this commit, I changed the logic slightly, but I'm not too sure about this code. Maybe you remember this bit better? Two things:

- the old code always fetched the mappings/agglomerates from the server (even when they already existed). I changed that so that the store is used if they already exist
- the loaded agglomerates were only updated in the store if no agglomerates existed before. I somehow kept that, but I don't think that it really makes sense, because loadLayerMappings could simply do the update if it loaded something from the server (no extra parameter required). but I wanted to ask you about this logic before I'm butchering it any further :sweat_smile: 

might make sense to have a call about it, if you prefer that :call_me_hand:

daniel-wer (commented):
My tests mostly went well :+1: I encountered these issues:

- [ ] The proofreading mode cannot be activated, because allegedly the annotation doesn't have a skeleton, but it does have one.
the_skeleton_is_a_lie
- [ ] Although the proofreading mode cannot be activated by via clicking (see previous point), the `Ctrl + K, O` does activate it.
- [ ] Activated json mappings do not seem to be persisted in an annotation across reloads (also the case on master). This might be known, already.

PR #7175, opened 2023-06-23
Implement search for segments and groups

### LINKED ISSUE #7167: Search in segment list
The segment list is virtualized, so ctrl+f does not work.

I am looking for a specific segment id in an annotation with many segments, in order to jump to its anchorPosition. Currently I have to scroll through the whole list (everything is still named only with its id). It would be great to have a search field in there.

philippotto:
Implement search for segments and groups in the segments tab (similar to the skeleton tab).

- https://___.webknossos.xyz

### Steps to test:
- open an annotation
- create many segments (e.g., by clicking on various segments on an existing segmentation layer)
- name some segments
- use the new search button to search for segments and use the jump feature

### Issues:
- fixes #7167 

------


MichaelBuessemeyer (approved):
Works very very nicely.

I like the idea of extending the type of the searchKey for the search popover.
I just added one suggestion that you may apply or ignore.

Additional thanks for getting rid of the ts error :rocket:

PR #7008, opened 2023-04-24
Added more drawings for empty dashboards

hotzenklotz:
PR adds new icons/ drawing and slight redesign for the list views in the dashboard in line with the new branding.

image

image

image

First time user:
image

### URL of deployed dev instance (used for testing):
- https://emptylistdrawings.webknossos.xyz/

### Steps to test:
- Remove all annotations, tasks, and datasets
  - The easiest way is to drop your DB (`yarn refresh-schema`) and rename the orga in your binary data folder
- Check out the empty dashboard
- Upload a dataset -> regular empty list
- Create a new folder -> empty folder list view
- Create an annotation -> regular annotation list
- Create a task -> regular task list

### Issues:
- Contributes to #6739 

------
(Please delete unneeded items, merge only when none are left open)
- [x] Updated [changelog](../blob/master/CHANGELOG.unreleased.md#unreleased)
- [ ] Updated [migration guide](../blob/master/MIGRATIONS.unreleased.md#unreleased) if applicable
- [ ] Updated [documentation](../blob/master/docs) if applicable
- [ ] Adapted [wk-libs python client](https://github.com/scalableminds/webknossos-libs/tree/master/webknossos/webknossos/client) if relevant API parts change
- [ ] Removed dev-only changes like prints and application.conf edits
- [x] Considered [common edge cases](../blob/master/.github/common_edge_cases.md)
- [ ] Needs datastore update after deployment

PR #7177, opened 2023-06-26
Fix search for single-folder organizations

philippotto:
- https://___.webknossos.xyz

### Steps to test:
- open dashboard with single folder
- use search
- should not show empty datasets placeholder

### Issues:
- follow-up to #7008

------


fm3 (approved):
Works for me :) Thanks!

PR #4040, opened 2019-04-25
Make Frontpage Mobile-Friendly

hotzenklotz:
Google Search complains, the wK.org is not mobile optimized. Therefore, I tried to improve this. What I did so far:

- add viewport `<meta>` tag
- some CSS adjustments for the header & spotlight view
- specified different grid size for large and small displays in spotlight view.

The `<meta>` tag typically restricts the browser viewport to the device width, which in turn is the basis for all the grids, media queries etc. This works well for regular web pages, but not great for wK.

Goal of the PR is not get wK ready for a great mobile experience. My focus was exclusively on the frontpage ("spotlight"), "Features", and "Pricing". Those look reasonably well. Opening a dataset is another story... 

## Discussion
1. Perhaps, we should only active the "mobile-friendly" frontpage on wk.org and not on other instances?
2. It would be great if we could have the `<meta>` tag only on the frontpage and not once you are logged in / viewing a dataset. I don't know a good way to enforce that though.

### URL of deployed dev instance (used for testing):
- https://mobile.webknossos.xyz

### Steps to test:
- Try https://mobile.webknossos.xyz on many devices. Perhaps BrowserStack?

### Issues:
- fixes #3890 
------
- [X] Ready for review

Some Screenshots:
"Pixel 2"
localhost_9000_(Pixel 2)

"iPhone X"
localhost_9000_(iPhone X)

"iPad"
localhost_9000_(iPad)

PR #2070, opened 2017-10-18
Get webknossos tablet ready

daniel-wer:
This PR makes webknossos tablet ready. It should now be possible to use it on mobile devices. It uses HammerJs for the panning and pinch gestures.
It also adds flow types to the `toast.js` file, because I was annoyed by an error that was not fixed earlier due to the missing flow annotations.

### Mailable description of changes:
 - Webknossos can be used to view datasets on mobile devices now. Panning the dataset and pinch to zoom is supported in all tracing views.

### Steps to test:
- Visit tablet-ready.webknossos.xyz and test whether panning and pinch to zoom works in all views. Pinch to zoom should always zoom to whatever position was at the center when the zooming gesture was initiated (except in arbitrary, there is only position-independent zooming) :)

### Issues:
- fixes #563

------
- [x] Ready for review

PR #7178, opened 2023-06-26
Add navigation+layout buttons for mobile

### LINKED ISSUE #7158: Mobile-friendly layout
## Detailed Description
A low-hanging fruit to make Webknossos more mobile-friendly is to collapse the sidebars on mobile browsers by default. That increases the viewport of the data and makes it possible to scroll around.
We should think about how to detect mobile browsers.

## Context
While it is not useful to do serious work in wk on a phone, it can be useful to showcase some data. The ability to scroll around a bit would be nice.

### LINKED ISSUE #7159: Navigation buttons for iPad
## Detailed Description
Using an iPad to do volume annotations works great with Webknossos. However, when maximizing one viewport, it is very hard to navigate to the next section. I would propose to add some floating buttons (e.g. in the lower left screen) for forward and backward navigation.

philippotto:
- Adds floating navigation buttons (see screenshot) for touch devices*. the buttons allow to:
  - toggle each sidebar
  - maximize a viewport (the one which was most recently touched)
  - navigate in the third dimension backwards/forwards (also depends on the most recently touched viewport). keeping the navigation buttons pressed moves the position continually (does not depend on the move-value at the moment, though)
- automatically collapses sidebars on narrow screens (the current threshold is 1080 pixel which is somewhat arbitrary, but loosely checked against https://gs.statcounter.com/screen-resolution-stats/mobile/worldwide. also it works on a Mi 8 device)

Instead of detecting touch support or mobile devices, I implemented the logic so that the floating buttons appear automatically when the user uses touch as input. Conversely, the buttons are hidden again on mouse over. This should make the UX smooth even for devices where both is possible (e.g., convertible laptops). For narrow screens (se above), the floating bar is shown automatically, but will also disappear on mouse move.

image

- https://mobiletracing.webknossos.xyz

### Steps to test:
- open wk on your laptop and check that everything works as usual
- open wk on a tablet or mobile device and check that the floating buttons appear (either automatically or on touch; depending on screen width)
- test the floating buttons itself

### Issues:
- fixes #7158
- fixes #7159

------


frontend/javascripts/components/adapt_viewport_metatag.ts
```diff
@@ -1,13 +0,0 @@
-import { useEffect } from "react";
```
philippotto, line 1: see https://github.com/scalableminds/webknossos/pull/4040 for context. I think the removal of the meta tag is not necessary anymore. I think, the original idea was so that one can zoom in and pan around WK with the "desktop dimensions", but the new mobile button bar is the better solution probably.

package.json
```diff
@@ -236,7 +236,8 @@
     "worker-loader": "^3.0.8"
   },
   "resolutions": {
-    "**/mini-store": "^1.1.0"
+    "**/mini-store": "^1.1.0",
+    "**/redux": "3.7.2"
```
philippotto, line 240: this is not strictly necessary anymore (was a byproduct of trying to get the library [use-redux-effect](https://github.com/aspect-apps/use-redux-effect) to work). however, I noticed that currently multiple versions of redux are referenced in yarn.lock which doesn't seem ideal (not sure whether they were included in the webpack bundle). so, I thought that pinning the version wouldn't hurt.

frontend/javascripts/oxalis/view/layouting/floating_mobile_controls.tsx
```diff
@@ -0,0 +1,95 @@
...
+import * as React from "react";
+import { useSelector } from "react-redux";
+import ButtonComponent from "../components/button_component";
+
+const moveForward = () => {
+  return MoveHandlers.moveW(1, true);
+};
+const moveBackward = () => MoveHandlers.moveW(-1, true);
+
+const BUTTON_STYLE = { userSelect: "none" } as const;
```
normanrz, line 17:
```suggestion
const BUTTON_STYLE = { userSelect: "none", WebkitUserSelect: "none" } as const;
```

frontend/javascripts/oxalis/view/layouting/floating_mobile_controls.tsx
```diff
@@ -0,0 +1,95 @@
...
+  const moveForwardProps = useRepeatedButtonTrigger(moveForward);
+  const moveBackwardProps = useRepeatedButtonTrigger(moveBackward);
+  const activeViewport = useSelector(
+    (store: OxalisState) => store.viewModeData.plane.activeViewport,
+  );
+
+  return (
+    <div
+      className="floating-buttons-bar"
+      style={{ position: "absolute", left: 8, bottom: 28, zIndex: 1000 }}
```
normanrz, line 30:
```suggestion
      style={{ position: "fixed", left: 8, bottom: 28, zIndex: 1000 }}
```

frontend/javascripts/oxalis/view/layouting/tracing_layout_view.tsx
```diff
@@ -169,6 +178,37 @@ class TracingLayoutView extends React.PureComponent<PropsWithRouter, State> {
...
+
+  handleMouseOver = () => {
+    if (this.lastTouchTimeStamp && Date.now() - this.lastTouchTimeStamp < 1000) {
+      // Ignore mouse move events when they are shortly after touch events because the browser
+      // emulates these events when touch is used.
+      // Also ignore the event when touch was never used, because then the mobile buttons
+      // were never shown, anyway.
+      return;
+    }
+    return this.setState({ showFloatingMobileButtons: false });
```
daniel-wer, line 211: Minor, but I would set `this.lastTouchTimeStamp = null` here as well. Otherwise, `setState` will be called on every subsequent mouseOver event if there ever was a touch event.

frontend/javascripts/libs/react_hooks.ts
```diff
@@ -90,6 +90,39 @@ export function useKeyPress(targetKey: "Shift" | "Alt" | "Control") {
   return keyPressed;
 }
 
+export function useRepeatedButtonTrigger(triggerCallback: () => void, repeatDelay: number = 150) {
+  const [isPressed, setIsPressed] = useState(false);
+
+  useEffect(() => {
+    let timerId: NodeJS.Timeout;
```
daniel-wer, line 97: Minor, but people seem to be using [this](https://stackoverflow.com/a/51040768) type instead.

daniel-wer (approved):
Very cool! I like the simple design and thoughtful UX very much :100: 

Only thing I noticed during testing was that the viewport indicator as well as the forward/backward buttons are not implemented to properly support flight and oblique mode. I think the last active viewport is memorized and the position is adapted according to that.
Depending on if you want to support it, the respective buttons could either be hidden, there, or the viewport button could be hidden and the forward/backward could be adapted to move in the proper w-direction according to the current rotation.

frontend/javascripts/libs/utils.ts
```diff
@@ -24,6 +24,10 @@ export function mod(x: number, n: number) {
   return ((x % n) + n) % n;
 }
 
+export function keys<T extends string | number | symbol>(o: Record<T, any>): T[] {
```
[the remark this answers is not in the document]
philippotto: you're right, it should only be `T extends string` (so that string literals can be inferred). this was a copy-paste error :see_no_evil:

app/views/main.scala.html
```diff
@@ -2,7 +2,10 @@
 <!DOCTYPE html>
 <html lang="en">
   <head>
-    <meta name="viewport" content="width=device-width, initial-scale=1.0, user-scalable=yes" />
+    <meta
+      name="viewport"
+      content="width=device-width, initial-scale=1.0, minimum-scale=1, maximum-scale=1, user-scalable=yes"
```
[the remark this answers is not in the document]
philippotto: yes, you are right. I didn't find a way to allow user scaling without preventing the browser from scaling the page when rotating the screen. I'm changing the user-scalable option to no to avoid confusion.

frontend/javascripts/libs/input.ts
```diff
@@ -617,7 +617,7 @@ export class InputMouse {
     if (this.lastScale != null) {
       const delta = evt.scale - this.lastScale;
       this.lastScale = evt.scale;
-      this.emitter.emit("pinch", 10 * delta);
+      this.emitter.emit("pinch", 10 * delta, this.position);
```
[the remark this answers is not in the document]
philippotto: I meant that the factor 10 feels fishy, but I don't think that it can be simply removed. I suspect that it's there for a reason :thinking: It was added in this PR: https://github.com/scalableminds/webknossos/pull/2070/files#diff-f1af7da1cff58c6ece76e97bdc89036a204b0e0f4ce57a201f5ff8a1b365ef43R455. Since I cannot really test the pinch-to-zoom feature, I'd leave it as is for now..

PR #7188, opened 2023-07-06
Make layer order editable via drag and drop

### LINKED ISSUE #7100: Layer ordering for blend modes

MichaelBuessemeyer:
This PR makes the color layer settings sortable via drag. It adds a grabbing menu icon left to the enable/disable switch. Users can use this handle to adjust the order of the color layers which will affect the rendering output in case the cover mode is active. Afaik there is no difference for the additive blending mode.
The handle is only shown if there is more than one color layer.

- https://___.webknossos.xyz

##### Test 1:

- open a dataset with only one color layer (e.g. l4_sample in view mode or with a tracing)
- There should not be a handle next to the enable/disable switch of the color layer  

##### Test 2:

- Open an annotation of a dataset with multiple color layers
- Give different colors to the color layers
- Turn on cover mode
- Reposition the color layers. It should be possible to see the changes of the repositioning by observing the changes in color due to the color layer having different configured colors.
- Reload the annotation
- The order of the layers should have persisted and rendering should still work and look as expected.

### Issues:
- fixes #7100 

### TODOs:
- [x] Persist layer order settings
- [x] Consider making segmentation layer not movable in the first version as there is currently no use case :thinking: 

## Wanted Feedback:
- Currently, I called the field that saves the order of the color layers in the dataset configuration `layerOrder`. This is not as precise as it could be, because the variable only includes color layer names. Thus maybe `colorLayerOrder` would be more accurate and more understandable. But on the other hand, we might decide to also make the segmentation layer sortable and then the name would no longer be correct. I talked to Norman and his opinion is that the segmentation layer should not be sortable in the current state, as there is no real use-case right now for it and the segmentation layer will always be rendered on top; no matter whether one of the upper layers defines the pixel as covered by selecting cover blend mode or not. Therefore the current version does not include the segmentation layers being draggable / sortable. @philippotto what do you is the best name to keep for the field? In long term `layerOrder` might be smarter but more confusing right now and `colorLayerOrder` would be good at the moment, but bad once segmentation layers are also sortable which might come in the future.
- Update: I now renamed the field to `colorLayerOrder`

### Important Note
In case anyone comes across this pr while developing the pr that makes segmentation layers sortable: There is already a working version at commit [4817a49](https://github.com/scalableminds/webknossos/pull/7188/commits/4817a49853b8b5f122cf552a52b3b56e1b4f922d)

### Bugs:
- [x] rendering first segmentation layer and then color layer is broken :broken_heart: 

------


frontend/javascripts/oxalis/shaders/main_data_shaders.glsl.ts
```diff
@@ -202,7 +203,8 @@ void main() {
 
   // Get Color Value(s)
   vec3 color_value  = vec3(0.0);
-  <% _.each(colorLayerNames, function(name, layerIndex) { %>
+  <% _.each(orderedColorLayerNames, function(name, layerIndex) { %>
```
philippotto, line 206: do you think it's possible to generate the shader code so that it doesn't need to be recompiled when the layer order changes? on my machine, the shader compilation takes a few seconds which means that the UI is blocked everytime I change the layer order :S
`orderedColorLayerNames` can probably be passed as a uniform, but I'm not sure what needs to be done additionally (if it's possible at all).

frontend/javascripts/oxalis/view/left-border-tabs/layer_settings_tab.tsx
```diff
@@ -356,6 +381,7 @@ class DatasetSettings extends React.PureComponent<DatasetSettingsProps, State> {
     layerName: string,
     elementClass: string,
     layerSettings: DatasetLayerConfiguration,
+    draggingDisabled?: boolean,
```
philippotto, line 384:
```suggestion
    draggingDisabled: boolean = true,
```

would make the condition in 472 clearer, I think.

frontend/javascripts/oxalis/view/left-border-tabs/layer_settings_tab.tsx
```diff
@@ -706,16 +733,21 @@ class DatasetSettings extends React.PureComponent<DatasetSettingsProps, State> {
...
-  getLayerSettings = (
-    layerName: string,
-    layerConfiguration: DatasetLayerConfiguration | null | undefined,
-    isColorLayer: boolean = true,
-  ) => {
+  LayerSettings = ({
+    layerName,
+    layerConfiguration,
+    isColorLayer,
+    draggingDisabled = undefined,
```
philippotto, line 740: why not a boolean initial value?

frontend/javascripts/oxalis/view/left-border-tabs/layer_settings_tab.tsx
```diff
@@ -1049,22 +1084,53 @@ class DatasetSettings extends React.PureComponent<DatasetSettingsProps, State> {
...
+      (layerName) => !getIsColorLayer(this.props.dataset, layerName),
+    );
+    const colorLayerSettings = colorLayerOrder.map((layerName, index) => {
+      const isSortingDisabled = colorLayerOrder.length < 2;
+      return (
+        <SortableLayerSettings
+          key={layerName}
+          layerName={layerName}
+          layerConfiguration={layers[layerName]}
+          isColorLayer={true}
```
philippotto, line 1117:
```suggestion
          isColorLayer
```

frontend/javascripts/oxalis/geometries/materials/plane_material_factory.ts
```diff
@@ -633,6 +633,22 @@ class PlaneMaterialFactory {
       ),
     );
 
+    let oldLayerOrder: Array<string> = [];
```
philippotto, line 636: Is this ever written to?

frontend/javascripts/oxalis/model_initialization.ts
```diff
@@ -785,6 +790,28 @@ function enforcePricingRestrictionsOnUserConfiguration(
...
+    datasetConfiguration.colorLayerOrder.length < colorLayerNames.length
+  ) {
+    return {
+      ...datasetConfiguration,
+      colorLayerOrder: colorLayerNames,
+    };
+  }
+  const onlyExistingLayers = datasetConfiguration.colorLayerOrder.filter(
+    (layerName) => colorLayerNames.indexOf(layerName) >= 0,
+  );
```
philippotto, line 811: this guards against the case in which `colorLayerOrder` contains additional names that don't exist in the dataset, right? however, if this happens, couldn't it be that `onlyExistingLayers.length < colorLayerNames.length`? How about moving this filtering over the if-block in line 800?

philippotto (changes requested):
nice stuff! DND works very nicely :+1:

I left some comments. The biggest one is with regard to the shader recompilation which I would like to avoid as much as possible.

And another thought: maybe the reordering UI should be disabled when the layer mode is additive because it won't make a difference in that case? Otherwise, I imagine this to be a bit misleading..

frontend/javascripts/oxalis/shaders/main_data_shaders.glsl.ts
philippotto: I discussed this with Tom and we think that this shouldn't block the PR as the reordering won't be used with high frequency at the moment. Maybe we revamp the rendering architecture to be multi-pass soon. Then, this issue can be solved easier probably.

MichaelBuessemeyer (commented):
@philippotto I think I addressed everything and the layer rendering is now only enabled when the cover blend mode Is turned on.

frontend/javascripts/oxalis/model_initialization.ts
```diff
@@ -799,18 +799,19 @@ function ensureDatasetSettingsHasLayerOrder(
   const colorLayerNames = _.keys(datasetConfiguration.layers).filter((layerName) =>
     isColorLayer(dataset, layerName),
   );
+  const onlyExistingLayers =
+    datasetConfiguration?.colorLayerOrder?.filter(
+      (layerName) => colorLayerNames.indexOf(layerName) >= 0,
+    ) || [];
   if (
     datasetConfiguration.colorLayerOrder == null ||
```
philippotto, line 807:
```suggestion
```

shouldn't be necessary, since onlyExistingLayers will be empty in that case.

philippotto (approved):
excellent :+1: just check my one suggestion before merging :)

Issue #6305, opened 2022-06-27
Allow selecting mapping for thumbnail


fm3:
The thumbnail.jpg route of the datastore should support selecting mappings, and the wk-side routes should also use that (let’s discuss how the user should specify what should be used there)

fm3:
may be superseded by #3937

fm3:
Note that as of #7190 the thumbnail route takes a `mappingName` parameter. The front-end does not yet make use of that, though. Not sure how the frontend would know which to choose.

fm3:
mapping name should probably become a part of the view configuration #7412 then we could take it from there.

fm3:
The mappingName is now part of the viewConfiguration, as of https://github.com/scalableminds/webknossos/pull/7858 
It should be passed to the thumbnail route called in ThumbnailService

PR #7190, opened 2023-07-07
Thumbnail improvements: persistent cache, configured intensity, float layers

### LINKED ISSUE #5089: Cache publication thumbnails across restarts
Now that we are more frequently upgrading the webknossos.org instance, the thumbnail cache gets cleared very often, leading to long load times. It would be nice if the thumbnail cache could survive back-end restarts. We could use Redis or Postgres for that, or even just save them to disk.

### LINKED ISSUE #6387: Use brightness min/max from configuration for thumbnails, support float data

fm3:
- Replaced the TemporaryStore-based caching of thumbnails by a two-level caching (in-memory via AlfuCache and in postgres via new table)
 - Moved the parameter selection for thumbnails from datastore to wk side
 - Propagate intensityRange from dataset view configuration to thumbnail generation
 - Using that, and depending on dtype, perform intensity normalization
 - Also support float dataset thumbnails (here the intensityRange is a hard requirement, since float could otherwise go from negativeInfinity to positiveInfinity)

### TODO
 - [x] use intensityRange
 - [x] test cache clear
 - [x] evolutions
 - [x] test cleanup job
 - [x] store selected metadata in postgresql (mimetype, bbox, intensityrange?)
 - [ ] adapt postgres backups to skip this table?

- https://thumbnailcache.webknossos.xyz

### Steps to test:
- Visit the thumbnail route for a few datasets `http://localhost:9000/api/datasets/sample_organization/<datasetName>/layers/<layerName>/thumbnail`
- Only the first request should yield logging in backend like »sending request to thumbnail.jpg« (this generates the thumbnail on the datastore
- Subsequent should use cache
- Adapt intensity in view configuration of the dataset (in dataset settings page)
- Thumbnail should then look like that as well
- Also test this for a float layer.

### Issues:
- fixes #5089 
- fixes #6387 
- contributes to #6305 (though the front-end does not use this parameter yet)

------


fm3:
@normanrz Looks like pg_dump can be configured with `--exclude-table-data` to skip the contents of a table (here `webknossos.dataSet_thumbnails`) – do you know where this change needs to be made for our backups?

app/models/binary/ThumbnailService.scala
```diff
@@ -0,0 +1,237 @@
...
+  private val MaxThumbnailHeight = 4000
+
+  def getThumbnailWithCache(
+      organizationName: String,
+      datasetName: String,
+      layerName: String,
+      w: Option[Int],
+      h: Option[Int],
+      mappingName: Option[String])(implicit ec: ExecutionContext, mp: MessagesProvider): Fox[Array[Byte]] = {
+    val width = com.scalableminds.util.tools.Math.clamp(w.getOrElse(DefaultThumbnailWidth), 1, MaxThumbnailWidth)
```
frcroth, line 44: Interesting that this clamp does not use the commonly used order of min, value, max which is also used in frontend utils.ts and common library functions. Made me stumble here.

frcroth (approved):

app/models/binary/ThumbnailService.scala
```diff
@@ -0,0 +1,237 @@
...
+                      height: Int,
+                      mappingNameOpt: Option[String],
+                      image: Array[Byte],
+                      mimeType: String,
+                      mag: Vec3Int,
+                      mag1BoundingBox: BoundingBox): Fox[Unit] = {
+    val mappingName = mappingNameOpt.getOrElse("") // in sql, nullable columns can’t be primary key, so we encode no mapping with emptystring
+    for {
+      _ <- run(q"""INSERT INTO webknossos.dataSet_thumbnails (
+            _dataSet, dataLayerName, width, height, mappingName, image, mimetype, mag, mag1BoundingBox, created)
```
frcroth, line 214:
```suggestion
            _dataSet, dataLayerName, width, height, mappingName, image, mimetype, sourceMag, mag1BoundingBox, created)
```

app/models/binary/ThumbnailService.scala
```diff
@@ -0,0 +1,237 @@
...
+                   VALUES($datasetId, $layerName, $width, $height, $mappingName, $image, $mimeType, $mag, $mag1BoundingBox, ${Instant.now})
+                   ON CONFLICT (_dataSet, dataLayerName, width, height, mappingName)
+                   DO UPDATE SET
+                     image = $image,
+                     mimeType = $mimeType,
+                     mag = $mag,
```
frcroth, line 220:
```suggestion
                     sourceMag = $mag,
```

frcroth (changes requested):

app/models/binary/ThumbnailService.scala
fm3: I googled around a bit and it looks like the ordering `value, min, max` is actually also pretty common across languages (e.g. [c++](https://en.cppreference.com/w/cpp/algorithm/clamp) or [c#](https://www.educative.io/answers/what-is-mathclamp-in-c-sharp), [unity3d library](https://docs.unity3d.com/ScriptReference/Mathf.Clamp.html)). On stackoverflow it’s about 50/50, compare [here](https://stackoverflow.com/questions/16656651/does-java-have-a-clamp-function).

I think I’d leave this as is for now unless you would vote otherwise?

frcroth: No strong opinion here, just thought it weird

frcroth (approved):

PR #7182, opened 2023-07-04
Implement security.txt route

frcroth:
### URL of deployed dev instance (used for testing):
- https://___.webknossos.xyz

### Steps to test:
- Visit /.well-known/security.txt

### Issues:
- fixes #7113 

For discussion on content, see slack.

------
(Please delete unneeded items, merge only when none are left open)
- [x] Updated [changelog](../blob/master/CHANGELOG.unreleased.md#unreleased)

PR #7191, opened 2023-07-07
enable security.txt route also for no-ds setup

fm3:
follow-up for #7182

in some instances, the routes file is changed to one of these (to disable the datastore and/or tracingstore routes if those modules are disabled)


frcroth (approved):
Nice! I was wondering why it didn't work on wkorg

PR #7192, opened 2023-07-07
add edge case on noDs/noTs

fm3:
add edge case to md file

### DIFF
--- a/.github/common_edge_cases.md
+++ b/.github/common_edge_cases.md
@@ -8,4 +8,4 @@ Will this also work if
  - The annotation is a compound task/project/tasktype annotation
  - The user is logged out and views a public dataset or annotation
  - User uses dark mode / light mode
- - There is no local datastore/tracingstore module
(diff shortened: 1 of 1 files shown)


.github/common_edge_cases.md
```diff
@@ -8,3 +8,4 @@ Will this also work if
  - The annotation is a compound task/project/tasktype annotation
  - The user is logged out and views a public dataset or annotation
  - User uses dark mode / light mode
+ - There is no local datastore/tracingstore module
```
frcroth, line 11: Good that it is here, but how would I test this?

fm3: I added a link to https://github.com/scalableminds/webknossos/wiki/Set-up-a-standalone-datastore-locally – I think that’s still mostly up to date. I’ll re-check that the instructions there still work soon

frcroth (approved):
The tasks seem pretty straight forward. Maybe there could even be a script to automatically start and configure a separate data store?

PR #7198, opened 2023-07-10
Distinguish between Error and Empty for GCS and S3 DataVaults

### LINKED ISSUE #7149: Distinguish between error and Empty (404) for GCS and S3 DataVaults
#7137 introduces this for HttpsDataVault. This way, we could insert the fill value on empty, while propagating errors on other Failures

frcroth:
- https://___.webknossos.xyz

### Steps to test:
- Should be done through tests ;)
- (check if GCS and S3 still work)

### Issues:
- fixes #7149 

Since it is not user facing I guess no changelog.

------


util/src/main/scala/com/scalableminds/util/tools/Fox.scala
```diff
@@ -323,14 +323,24 @@ class Fox[+A](val futureBox: Future[Box[A]])(implicit ec: ExecutionContext) {
...
+    val box = await(awaitTimeout)
     box.openOrThrowException(justification)
   }
 
+  /**
+    *
+    * Awaits the future and returns the box. Should not be used in production code (therefore marked as Deprecated).
+    */
+  @Deprecated
+  def await(awaitTimeout: FiniteDuration = 10 seconds): Box[A] =
```
fm3, line 341: I think it would be good form to force the user to pass a justification here as well, just like in `get`. This, together with the deprecation warning, hopefully ensures that nobody ever uses this in production code :)

test/backend/DataVaultTestSuite.scala
```diff
@@ -45,14 +46,35 @@ class DataVaultTestSuite extends PlaySpec {
...
           val bytes = (vaultPath / dataKey).readBytes(Some(range))(globalExecutionContext).get(openFoxJustification)
 
           assert(bytes.length == range.length)
           assert(bytes.take(10).sameElements(Array(-1, -40, -1, -32, 0, 16, 74, 70, 73, 70)))
         }
+
+        "return empty box" when {
+          "requesting a nox-existent object" in {
+            val result = (vaultPath / "non-existent-key").readBytes()(globalExecutionContext).await()
+            assert(result.isEmpty)
```
fm3, line 62: Failure boxes also have `isEmpty == true`. I think this needs a different kind of check (e.g. use `match`)

fm3 (commented):
The implementation looks good to me :) 

Props for writing tests 🎉 

The test does not currently distinguish between Failure and Empty, though. Also, could you add a test case where a Failure is expected? (e.g. Forbidden?)

I think this is actually a user-facing change, I’d add something like this into the changelog

“Correctly use configured fill-value for missing chunks of remote datasets hosted on gcs or s3.” (since there is a distinction on Failure vs Empty in ChunkReader, which can now work here)

fm3 (approved):
That’s fair then, I don’t have such a link ready either.

LGTM! :shipit:

fm3 (approved):

Issue #7199, opened 2023-07-11
Adding dataset in folder with rights while not having rights to organization root folder


frcroth:
When a user does not have access rights to the root folder of an organization but access to a different (sub)folder, they can upload a dataset (which is automatically in the root folder) but then not access it (or even complete the import procedure). (additional context: https://scm.slack.com/archives/C5AKLAV0B/p1689069626934229?thread_ts=1689064935.050179&cid=C5AKLAV0B)

## Expected Behavior
Two options to fix this:
	- Store folder directly on upload so that the dataset is not uploaded into the root directory (preferred)
	- Disallow uploading generally when a user does not have access rights to the root directory

## Steps to Reproduce the bug

1. As the admin user, create a directory and give acess to a non-admin user via team permission (the non-admin user does not have rights to upload in the root directory)
2. Make the non-admin user team manager
3. As the non-admin user, try to upload a dataset in that folder

fm3:
I agree that users with this rights configuration should be able to upload into the directories that they have write access for. The information in which folder the dataset should land should be propagated accordingly. I would have expected this to happen already, via `/datastores/:name/reserveUpload` but apparently there is some bug in that logic.

fm3:
I double-checked that this works for upload, and this issue concerned with the addRemote aka put datasource variant

PR #7200, opened 2023-07-11
Being dataset uploader does not make editable true

### LINKED ISSUE #6780: being dataset uploader should not make isEditable true
Currently, the uploader is allowed to edit the dataset, there is no way to revoke this.

Can it be removed without breaking the initial teams logic during upload? Probably, because team managers can only upload to directories that grant them access anyway. Needs to be double-checked, though.

Edit this both in sql list query and scala-based isEditable.

Compare https://github.com/scalableminds/webknossos/pull/6759#discussion_r1085098419

frcroth:
- https://___.webknossos.xyz

### Steps to test:
- Create dataset with read acess by non-admin user.
- Set uploader in db.
- Ensure that the user still can not edit it (e.g. via PATCH /datasets/org/dataset/updatePartial)

Uploading a dataset requires managing permissions in that folder (see #7199 for an issue arising from that). That way the initial uploading logic should be unaffected.

### Issues:
- fixes #6780 

Also "uploader" is not used for remote datasets?

------
- [x] ~~Updated [migration guide](../blob/master/MIGRATIONS.unreleased.md#unreleased) if applicable ???~~


fm3 (approved):
Thanks!

No migrations are necessary, I also double-checked that this detail is not mentioned in the docs, so no changes needed there either.

Could you write an additional issue for setting uploader during explore-and-add for remote datasets? I think that would make sense.

Also, good find on the other issue with folder permissions. We should have a look at that soon. (It’s not extremely pressing, though, since most people who add datasets are admin or dataset managers anyway)

PR #7205, opened 2023-07-13
Add button to save current view configuration as default in view mode

### LINKED ISSUE #5183: Add button for making view settings default
## Detailed Description
Add a button in the view mode of a dataset to persist the current view configuration (e.g. histogram settings, layer enabled, color, inverted, position, zoom) for this dataset as default view configuration.

Maybe this could be implemented alongside #4772.

## Context
Setting the view configuration of a dataset is finicky, because it needs to be expressed in a JSON format in the dataset settings view. What I typically do, is to configure the dataset optics in the view mode and then go to the settings to try to replicate these values in the JSON format.

MichaelBuessemeyer:
- https://updatedefaultlayerconfiginviewmode.webknossos.xyz

### Steps to test:
- Open a dataset in view mode (as admin or dataset manager)
- Change the view mode configuration in a recognizable way (e.g. set histogram min = 100, max = 200)
- Use the new button at the bottom of the layer configuration tab to save the new configuration
- A modal should open up for confirmation. If you cancel the action, no update should be performed.
- If you confirm the modal, a toast should indicate that the updated succeeded
- Use the crog-icon next to the dataset name at the top of the info tab to open the dataset settings
- Go to the view configurations and confirm that the update worked successfully.

## TODOs:
- [x] Test whether this pr works well together with the new layer ordering pr during cover mode for color layers. #7188

### Issues:
- fixes #5183

------


MichaelBuessemeyer:
@normanrz could you please take a better example image that has the same style as the other images from the docs once this pr is merged?

frontend/javascripts/oxalis/view/left-border-tabs/layer_settings_tab.tsx
```diff
@@ -1103,6 +1141,17 @@ class DatasetSettings extends React.PureComponent<DatasetSettingsProps, State> {
           </Row>
         ) : null}
 
+        {this.props.controlMode === ControlModeEnum.VIEW && this.props.isAdminOrManager ? (
```
daniel-wer, line 1144: Is there a reason why this is only allowed from view mode? I cannot read from the issue description that this was a requirement and was wondering why you introduced this restriction.
I guess annotations would complicate this functionality, because volume annotation layers would need to be filtered out and/or replaced by their fallback layers :thinking: 
@normanrz Do you think it's sufficient to enable this for viewed datasets, only?

frontend/javascripts/oxalis/view/left-border-tabs/layer_settings_tab.tsx
```diff
@@ -1049,6 +1052,41 @@ class DatasetSettings extends React.PureComponent<DatasetSettingsProps, State> {
...
+            position,
+            zoom,
+          });
+          await updateDatasetDefaultConfiguration(dataset, completeDatasetConfiguration);
+          Toast.success("Successfully saved the current view configuration saved as default.");
+        } catch (error) {
+          Toast.error(
+            "Failed to save the current view configuration as default. Please look at the console for more details.",
+          );
+          console.error(error);
```
daniel-wer, line 1084: Please add airbrake logging as well.

daniel-wer (commented):
Nice functionality :+1: Please see my suggestions and remarks.

I'll report back after testing :monkey:

daniel-wer:
Regarding my comment https://github.com/scalableminds/webknossos/pull/7205#discussion_r1265355156

I tested which settings are saved and restored when using the new functionality and it seems to be all-of-them (all dataset-related settings). This is in contrast to the existing configuration UI where only the values you mentioned can be modified. We need to decide whether (a) it's fine that the new functionality is more powerful, (b) the existing UI should be extended to have the same capabilities, or (c) the new functionality should be restricted to only persist the values that can be modified in the existing UI.

I made a screenshot of the settings before. Then, I modified all settings, saved using the new functionality and opened the dataset with a new user:

Before             |  After
:-------------------------:|:-------------------------:
settings_before  |  settings_after
settings_before_2  |  settings_after_2

Apart from that inconsistency, the functionality works very well :)

frontend/javascripts/oxalis/view/left-border-tabs/layer_settings_tab.tsx
```diff
@@ -1141,6 +1190,7 @@ const mapStateToProps = (state: OxalisState) => ({
   task: state.task,
   controlMode: state.temporaryConfiguration.controlMode,
   isArbitraryMode: Constants.MODES_ARBITRARY.includes(state.temporaryConfiguration.viewMode),
+  isAdminOrManager: state.activeUser != null ? Utils.isUserAdminOrManager(state.activeUser) : false,
```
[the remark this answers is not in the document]
MichaelBuessemeyer: Oh yeah, sorry I thought that the "Manager" in `isUserAdminOrManager` stands for "DatasetManager". Thanks for pointing out

frontend/javascripts/oxalis/view/left-border-tabs/layer_settings_tab.tsx
MichaelBuessemeyer: > I guess annotations would complicate this functionality, because volume annotation layers would need to be filtered out and/or replaced by their fallback layers 

Excactly that was my thought :+1: (Additionally norman mentioned to me that view mode only should be fine)

frontend/javascripts/oxalis/view/left-border-tabs/layer_settings_tab.tsx
```diff
@@ -1049,6 +1052,41 @@ class DatasetSettings extends React.PureComponent<DatasetSettingsProps, State> {
...
+    const { dataset, datasetConfiguration } = this.props;
+    Modal.confirm({
+      title: "Save current view configuration as default?",
+      content: (
+        <>
+          Do you really want to save your current view configuration as the dataset's default?
+          <br />
+          This will overwrite the current default view configuration.
+          <br />
+          This includes the current camera position, zoom level, interpolation setting and layer
```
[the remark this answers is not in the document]
MichaelBuessemeyer: So: Here are the results of my investigation: All settings you mentioned are sent to the backend and upon loading the dataset in view mode all settings are returned from the backend except for the position, zoon, and maybe rotation. But I think the default position and zoom and rotation are sent by another request to the frontend. Although the position does not seem to be applied upon loading the dataset (without the position already being defined by the query params)

Note that I did not test the `segmentationPatternOpacity` and the `rotation property.

This makes the new button more potent than the settings UI as the settings UI does not allow to configure the following settings:
```
fourBit
renderMissingDataBlack
loadingStrategy
blendMode
```

> I'm wondering whether we should make it more explicit what exactly would be saved.

For this I can add a table to the modal that tasks for confirmation. The table would list the properties saved. 

To agree upon what settings should be saved we can talk in person :)

MichaelBuessemeyer:
> The layer configuration in the dataset view settings doesn't currently allow to change: brightness, contrast, min, max
I think these should be added to the schema so they can be changed (and documentation in the table at the right should be added, too).

I am pretty sure that brightness and contrast were replaced with `min, max, intensityRange` and thus they should not be configurable. But you are right with `min` and `max`. They are already configurable via the json layer config in the UI and are now also added to the table.

As discussed I reworked the UI and made the `intensityRange` setting optional / required for color layers.
As this change required some changes in multiple files, it would be great if you could look over them to avoid accidental bugs :pray:

frontend/javascripts/types/schemas/dataset_view_configuration_defaults.ts
```diff
@@ -62,6 +63,11 @@ export const enforceValidatedDatasetViewConfiguration = (
           existingLayerConfig,
         );
         eliminateErrors(existingLayerConfig, layerErrors, layerConfigDefault);
+        if (layer.category === "segmentation") {
+          delete existingLayerConfig.intensityRange;
+        } else if (existingLayerConfig.intensityRange == null) {
+          existingLayerConfig.intensityRange = getDefaultIntensityRangeOfLayer(dataset, layer.name);
```
daniel-wer, line 69: Why is this always added for color layers? In my tests, I saved a layer configuration of 

```
{
  "color": {},
  "segmentation": {} 
}
```

and when I reloaded the page, the layer configuration was:

```
{
  "color": {
    "intensityRange": [
      0,
      255
    ]
  },
  "segmentation": {}
}
```

This was unexpected.

daniel-wer (commented):
Thank you for the newest improvements. I have a couple of small remarks, but afterwards this PR should be ready to be merged :+1:

MichaelBuessemeyer (commented):
Thanks for the review daniel. I hope everything should be fine now.

frontend/javascripts/oxalis/model/sagas/clip_histogram_saga.ts
```diff
@@ -11,11 +11,7 @@ import { api } from "oxalis/singletons";
 import { getActiveMagIndexForLayer } from "../accessors/flycam_accessor";
 
 function onThresholdChange(layerName: string, [firstVal, secVal]: [number, number]) {
-  if (firstVal < secVal) {
-    Store.dispatch(updateLayerSettingAction(layerName, "intensityRange", [firstVal, secVal]));
-  } else {
-    Store.dispatch(updateLayerSettingAction(layerName, "intensityRange", [firstVal, secVal]));
-  }
+  Store.dispatch(updateLayerSettingAction(layerName, "intensityRange", [firstVal, secVal]));
```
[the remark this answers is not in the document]
daniel-wer: Sounds good, thank you for testing!

daniel-wer (approved):
Nice, LGTM :+1:

MichaelBuessemeyer:
Sadly, (but obviously) this pr did not work with #7188 out of the box. I had to add a missing form item that enables the user to manipulate and save the order of the color layers. 
As an approach, I used the same dnd lib as used in #7188 to create a list of sortable color layer names via dnd.
Sadly this also implies that I had to add another settings item to the UI. This means by the current layout that a new row only for a single setting needed to be added.

The current version looks like this: 
image
(Below the part of the image is a divider and then the json settings input for each layer. I excluded this in the screenshot.)
As the list takes up a lot of space on the screen, I wrapped it in a collapsable and hide it as default. The collapsible is also only extendable when having more than one color layer.

 @philippotto @daniel-wer Do you think this UI is okayish enough? / Do you have suggestions?

frontend/javascripts/messages.tsx
```diff
@@ -79,6 +80,8 @@ export const settingsTooltips: Partial<Record<keyof RecommendedConfiguration, st
     "Set the blend mode for the dataset. The additive mode (default) adds the data values of all color layers. In cover mode, color layers are rendered on top of each other so that the data values of lower color layers are hidden by values of higher layers. Cover mode enables reordering of color layers.",
   renderWatermark: "Show a WEBKNOSSOS logo in the lower-left corner of each screenshot.",
   antialiasRendering: "Antialias rendering (can impact performance)",
+  colorLayerOrder:
+    "Set the order in which color layers are rendered. This setting will only affect the cover blend mode.",
```
philippotto, line 84:
```suggestion
    "Set the order in which color layers are rendered. This setting is only relevant if the cover blend mode is active.",
```

is a bit clearer in my opinion.

frontend/javascripts/dashboard/dataset/color_layer_ordering_component.tsx
```diff
@@ -0,0 +1,74 @@
...
+      const movedElement = colorLayerNames[oldIndex];
+      let newColorLayerNames = colorLayerNames.filter((_, index) => index !== oldIndex);
+      newColorLayerNames.splice(newIndex, 0, movedElement);
+      onChange(newColorLayerNames);
+    }
+  };
+
+  const isSettingEnabled = colorLayerNames && colorLayerNames.length > 1;
+  const collapsibleDisabledExplanation =
+    "This setting only makes sense if there are multiple color layers.";
```
philippotto, line 42:
```suggestion
    "The order of layers can only be configured when the dataset has multiple color layers.";
```

philippotto:
Code looks good, but I couldn't change the layer order for this DS: https://updatedefaultlayerconfiginviewmode.webknossos.xyz/datasets/sample_organization/idr0101A-4d/edit
even though it has multiple color layers?

frontend/javascripts/types/schemas/dataset_view_configuration_defaults.ts
```diff
@@ -71,6 +98,7 @@ export const enforceValidatedDatasetViewConfiguration = (
           : _.pickBy(layerConfigDefault, (value: any) => value !== null);
       }
     });
+    ensureDatasetSettingsHasLayerOrder(datasetViewConfiguration, dataset);
```
MichaelBuessemeyer, line 101: > Code looks good, but I couldn't change the layer order for this DS: https://updatedefaultlayerconfiginviewmode.webknossos.xyz/datasets/sample_organization/idr0101A-4d/edit
even though it has multiple color layers?

Thanks for finding this bug :mag:.

This line should fix the bug (according to my testing). 
The problem was I added the new property `colorLayerOrder` (see https://github.com/scalableminds/webknossos/pull/7188) and that it is expected to be always present in a validated `DatasetConfiguration`. But the backend does not send this property if it wasn't set by the frontend previously. 
Previous to my changes, the property was only enforced during model initialization (see changes in `model initialization`). But the enforcing of the property with correct values must be done every time when working with a `DatasetConfiguration`. Thus it made more sense to me, to move the enforcing code to the `enforceValidatedDatasetViewConfiguration` function. 
This covers the bug, as this function is explicitly called in `dataset_settings_view.tsx` after it is fetched from the backend. 
Additionally, the `enforceValidatedDatasetViewConfiguration` is also enforced for the model initialization as it is part of the `getDatasetViewConfiguration` API backend call in `admin_rest_api.ts` which is called during the model initialization to fetch the `DatasetConfig` from the backend.

Could you please retest the changes (once the CI is done)?

philippotto: Hm, it still doesn't work for me [here](https://updatedefaultlayerconfiginviewmode.webknossos.xyz/datasets/sample_organization/idr0101A-4d/edit)? :thinking: Does it work for you?

philippotto: Great, now it works :)

Issue #7144, opened 2023-06-15
“add segment index” button for volume annotations that don’t have one


fm3:
Similar to the downsample button, this would internally use the duplicate route and replace the tracing id in the annotation layer.

Would need a new route on webknossos backend side that forwards this to tracingstore (should be quick to do)

Note that this is currently only possible for volume annotations with no fallback segmentation layer

fm3:
unassigning myself as the backend part is done

fm3:
> Is there a way to know whether a volume tracing has a segment index?

Yes, see VolumeTracing.proto, `optional bool hasSegmentIndex`

PR #7208, opened 2023-07-17
Add route to add segment index to existing volume annotations

fm3:
Adds route to add a segment index to an existing volume annotaiton. Note that this is currently only available for volume annotaiton layer with no fallback segmentation.

### Steps to test:
(I think no manual testing is required if the code looks good, as it is pretty hard to set up, and if we see this fail in production, not much harm is done and we can fix it then)

- Create volume annotation with no segment index (this is hard to do, needs code changes, or check out a webknossos from before https://github.com/scalableminds/webknossos/pull/7063 and then migrate forwards)
- Call the new route
- Should now have segment index

### Issues:
- backend part for #7144


conf/messages
```diff
@@ -203,13 +203,16 @@ annotation.transferee.noDataSetAccess=Cannot transfer annotation to a user who h
...
 annotation.addLayer.nameInUse=An annotation layer with this name already exists in this annotation. Please change it to prevent duplicates.
 annotation.makeHybrid.failed=Could not convert annotation to hybrid.
 annotation.makeHybrid.alreadyHybrid=Could not convert annotation to hybrid annotation because it is already a hybrid annotation.
+annotation.downsample.layerNotFound=Could not downsample volume annotation (annotation layer not found)
 annotation.downsample.explorationalsOnly=Could not downsample annotation because it is only allowed for explorational annotations.
-annotation.downsample.volumeOnly=Downsampling is only possible for volume or hybrid annotations.
+annotation.downsample.volumeOnly=Downsampling is only possible for volume annotation layers.
+annotation.addSegmentIndex.layerNotFound=Could not add segment index to volume annotation (annotation layer not found)
+annotation.addSegmentIndex.explorationalsOnly=Could not add segment index to annotation because it is only allowed for explorational annotations.
+annotation.addSegmentIndex.volumeOnly=Adding segment index is only possible for volume annotation layers.
```
frcroth, line 215: `annotation.addSegmentIndex.failed` is used but does not occur here

app/models/annotation/AnnotationService.scala
```diff
@@ -393,14 +393,27 @@ class AnnotationService @Inject()(
...
       _ <- annotationLayersDAO.replaceTracingId(annotation._id, volumeAnnotationLayer.tracingId, newVolumeTracingId)
     } yield ()
 
+  def addSegmentIndex(annotation: Annotation, volumeAnnotationLayer: AnnotationLayer)(
+      implicit ctx: DBAccessContext): Fox[Unit] =
+    for {
+      dataSet <- dataSetDAO.findOne(annotation._dataSet) ?~> "dataSet.notFoundForAnnotation"
+      _ <- bool2Fox(volumeAnnotationLayer.typ == AnnotationLayerType.Volume) ?~> "annotation.downsample.volumeOnly"
+      rpcClient <- tracingStoreService.clientFor(dataSet)
+      // duplicate in tracingstore will add segment index if possible
```
frcroth, line 410: Can you point me to where that happens? Didn't immediately see it in VolumeTracingController.duplicate

fm3: https://github.com/scalableminds/webknossos/blob/26baa684225ce8e48ac0739e84f48efed3f90581/webknossos-tracingstore/app/com/scalableminds/webknossos/tracingstore/tracings/volume/VolumeTracingService.scala#L361

in combination with

https://github.com/scalableminds/webknossos/blob/26baa684225ce8e48ac0739e84f48efed3f90581/webknossos-tracingstore/app/com/scalableminds/webknossos/tracingstore/tracings/volume/VolumeTracingService.scala#L403

Not sure how to make this more discoverable. Hence the comment here

frcroth: I guess line 361 in VolumeTracingService is the surprising one, so there could be a comment there "Adding segment index on duplication if the volume tracing allows it" or something like that.

frcroth (approved):

fm3:
Email looks alright to me :) 

image

PR #7217, opened 2023-07-20
Fix bug in thumbnail size handling for coarse mags

fm3:
Switched up mag1 and targetMag in the thumbnail parameter selection :see_no_evil: 

follow-up for #7190


daniel-wer (approved):
LGTM :+1:

PR #7220, opened 2023-07-20
Use New Task Terminology in Code

### LINKED ISSUE #7218: Use new Task terminology also in code
(depending no context) Assignments → Task Instances, Project status
Open Tasks → Pending Tasks

fm3:
- https://taskterminology.webknossos.xyz

### Steps to test:
- Create some tasks, request, trace finish, look at report pages, task list view, project list view, should all still look as expected

### TODOs:
- [x] adapt schema
- [x] adapt backend code
- [x] snapshots
- [x] adapt frontend code
  - [x] open_tasks_report_view
- [x] test wk
- [x] write sql evolution
- [x] adapt python client https://github.com/scalableminds/webknossos-libs/pull/930
  - [x] discuss how much backwards compatibility is needed [→none :tada: ]
- [x] test python

### Issues:
- fixes #7218

------


app/controllers/TaskController.scala
```diff
@@ -83,7 +83,7 @@ Expects:
  - As Form data:
    - taskTypeId (string) id of the task type to be used for the new tasks
    - neededExperience (Experience) experience domain and level that selects which users can get the new tasks
-   - openInstances (int) if greater than one, multiple instances of the task will be given to users to annotate
+   - pendingInstances (int) if greater than one, multiple instances of the task will be given to users to annotate
```
philippotto, line 86: would it make sense to rename this to `initialPendingInstances`? or only `instancesCount`? I find this `pending` here a bit confusing :thinking:

conf/webknossos.latest.routes
```diff
@@ -63,8 +63,8 @@ GET           /users/:id/annotations
 GET           /teams                                                                            controllers.TeamController.list(isEditable: Option[Boolean])
 POST          /teams                                                                            controllers.TeamController.create
 DELETE        /teams/:id                                                                        controllers.TeamController.delete(id: String)
-GET           /teams/:id/openTasksOverview                                                      controllers.ReportController.openTasksOverview(id: String)
-GET           /teams/:id/progressOverview                                                       controllers.ReportController.projectProgressOverview(id: String)
+GET           /teams/:id/availableTasksReport                                                   controllers.ReportController.availableTasksReport(id: String)
```
philippotto, line 66: what is the difference pending pending and available? I'm asking because open was renamed to pending in other places.

philippotto (dismissed):
mostly looks good :+1: 

However, one navbar link is broken (statistics > available tasks points to https://taskterminology.webknossos.xyz/reports/openTasks which is a 404).

also see my one question about pending vs available.

I did not review the SQL migration, as I don't feel confident enough to spot potential errors there. can @frcroth chime in for that review maybe?

app/controllers/TaskController.scala
fm3: if create task where the only place where this was used, I’d agree, but the same class is also used when updating a task, and there pending instances makes more sense. So I’d rather leave it as is here as well

conf/webknossos.latest.routes
fm3: Fair question! We decided that this spot is a different perspective (which tasks are available for assignment for different users, versus how many tasks of a project are still pending).

I would also have been ok with using “pending” in both places, but we went for available in the report, compare discussion at https://scm.slack.com/archives/C5AKLAV0B/p1688653296691219?thread_ts=1688649830.633089&cid=C5AKLAV0B

app/controllers/TaskController.scala
philippotto: Ok, I see :+1:

conf/webknossos.latest.routes
philippotto: Ok, thanks for the clarification. Is this pending<->available difference explained somewhere in the docs or code comments? just for the sake of completeness..

frcroth (approved):
Sql could use comments, but looks consistent

PR #7221, opened 2023-07-20
Properly reserve dataset when putting datasource (addRemote)

### LINKED ISSUE #7199: Adding dataset in folder with rights while not having rights to organization root folder

### LINKED ISSUE #7211: Set uploader when adding a remote dataset

### LINKED ISSUE #7215: Supply Target Folder when adding Remote Dataset
Starting in #7176 (still WIP as of this writing) the backend supports a `folderId` field for the add route called by the frontend in `storeRemoteDataset`. The frontend should use that (and show a folder selection menu just like in the upload view).

This way the normal folder rights management is also respected when storing remote datasets. Otherwise, the user can only add remote datasets when having write access to the root folder

Related: #7199

fm3:
Datastore now calls `reserveDataSourceUpload` when adding a datasource. This sets both uploader and folderId properly, as well as performs the correct write access checks.

(Note that this is a subset of #7176 as I expect that to be blocked for a while, but we should fix this bug. I copied the changes to the two backend source files from that PR)

### Steps to test:
- Add remote dataset as non-admin user, see that this works for folders you have write access for (and not for others)
- Check that the folder is set correctly. (Also, uploader should be set in db)

### TODOs:
- [x] Frontend: pass folderId to the put route *instead of* calling updatePartial afterwards.

### Issues:
- fixes #7215
- fixes #7211
- fixes #7199

------


fm3:
@philippotto thanks! This works well when hitting the add button while navigated to a folder. However, in the upload view, there is a dropdown for specifying the folder. This is missing in the add remote view. Do you think this can be unified? I think having the menu (or at least a note saying where the dataset will go) would be good for the usability.

philippotto:
> @philippotto thanks! This works well when hitting the add button while navigated to a folder. However, in the upload view, there is a dropdown for specifying the folder. This is missing in the add remote view. Do you think this can be unified? I think having the menu (or at least a note saying where the dataset will go) would be good for the usability.

Maybe I'm missing something, but the add-remote-view has the dropdown you want. However, it only appears after one has explored the first URI. I can see why showing it earlier might make sense, but due to the current explore-configure flow, the UI appears later. Changing that would take some effort which isn't worth it in my opinion.

image

fm3:
I see, thanks! I missed it :see_no_evil: Yes, let’s keep it like that for now, then :)

frcroth (approved):
Very efficient! Closing 3 issues with one PR

PR #7231, opened 2023-07-26
Show email verification status in users table and make it filterable

### LINKED ISSUE #7228: isEmailVerified should be listed on user page
If a user email is not verified and the grace period is over, the user can no longer log in.

This should be mentioned in the user list view, so that team managers have a chance to understand what’s up with that user.

philippotto:
- https://___.webknossos.xyz

### Steps to test:
- create new account (automatically not activated and not verified)
- go to users list and check out the status column (tooltips and filters should make sense)
image
image

### Issues:
- fixes #7228


hotzenklotz (approved):
LGTM. Tested and works.

Two remarks/observations:
- When selecting nothing from the filter dropdown all users are shown. I guess this is fine.
- The filter works in `OR` mode. Initially my feeling was that the "activation status" and "email verification" are two orthogonal concept and therefore should be `and` filtered to fine tune the "search". But again, this is such niche feature that it does not matter too much and we should rather aim to ship it quickly.

philippotto:
> * When selecting nothing from the filter dropdown all users are shown. I guess this is fine.

I agree that this should be fine.

> * The filter works in `OR` mode. Initially my feeling was that the "activation status" and "email verification" are two orthogonal concept and therefore should be `and` filtered to fine tune the "search". But again, this is such niche feature that it does not matter too much and we should rather aim to ship it quickly.

I think, the OR semantic is easier to grasp and avoids weird situations (i.e., selecting both disabled and enabled would yield zero users in the AND case because the combination is impossible). The filtering UI would need more care for that semantic which is overkill, I think.

I'll merge once the CI has finished.

PR #7232, opened 2023-07-26
Don't show folders in search results

### LINKED ISSUE #7226: Message about 3 characters is blanked if folder exists
If there is a folder in the dataset structure it doesn't show that you need to enter at least three characters to search for dataset

image

philippotto:
When a dataset search is active, no folders within the current folder should be shown.

- https://___.webknossos.xyz

### Steps to test:
- go to dashboard table
- create a subfolder in the root (root should stay active)
- use the search bar
- the "you have to enter at least 3 chars" message should appear 
- no folders should be rendered in the search

### Issues:
- fixes #7226
- follow-up for regression caused by #6996 

------


MichaelBuessemeyer (approved):
Awesome :+1: works perfectly

PR #7233, opened 2023-07-26
Improve activativate-user flow

### LINKED ISSUE #7219: Clarify activate-and-configure-user flow
After activating a user, there's a modal with a cancel button. That button, however, does not cancel/undo the activation.

See https://discuss.webknossos.org/t/new-user-cancel/1797/2 for context.

Suggestion:
> right after clicking Activate user: a popup comes and tells me something like: Account is activated. Do you want to change the team permissions now? Yes - No

philippotto:
- https://improveuseractivation.webknossos.xyz

### Steps to test:
- go to users page
- deactivate a user
- activate a user
- this prompt should show up:
image
- ok/cancel should both work as expected

### Issues:
- fixes #7219


fm3:
Thanks for tackling this! 🎉 

I’d like to vote for printing a bit more info here, to avoid user confusion. What teams and permissions does the user have if you click »no«? I don’t even know for certain and I’ve worked with this a lot 😬 Maybe a single sentence could clarify that?

Also, can I wish for descriptive action buttons? yes/no buttons always require you to read the actual text fully, which, while not super long here, till feels a bit inconvenient. How about »close« and »configure teams and permissions« buttons? (Or something similar?)

philippotto:
> I’d like to vote for printing a bit more info here, to avoid user confusion. What teams and permissions does the user have if you click »no«? I don’t even know for certain and I’ve worked with this a lot

Fair enough! I adapted the message. It was a bit tricky to phrase, since the front-end doesn't know whether the user was activated before (thus, potentially having non-deault permissions). I hope the new copy is okay.

fm3:
much better, thanks! :)

daniel-wer (approved):
Works like a charm :tada:

PR #7239, opened 2023-07-27
make antd messages click-through

fm3:
Makes antd messages click-through. (A user noted that they block selecting tools during annotating, which can be annoying).

Do you think there are cases where the user wants to interact with antd messages (e.g. copy the text)? Nothing comes to my mind.

- https://messagesclickthrough.webknossos.xyz

### Steps to test:
- Brush some in a volume annotation, hit undo, try to switch tool behind the “finished undo” message
- Should be able to click right through the message, selecting tools.

------

### DIFF
--- a/frontend/stylesheets/main.less
+++ b/frontend/stylesheets/main.less
@@ -667,3 +667,7 @@ i.without-icon-margin,
     }
   }
 }
+
+.ant-message, .ant-message * {
+  pointer-events: none;
+}


philippotto (commented):
lgtm. ideally, messages shouldn't hide important parts of the UI, I'd argue. but it seems edge-case-y enough to go with this simple solution. 

Additionally, we could try to position the message a bit "lower". According to the docs, it should be possible with this:

```js
import { message } from 'antd';

message.config({top: 100})
```

Do you want to give this a try? It only needs to be executed once and should affect all future messages. I can also try this out if you want.

> Do you think there are cases where the user wants to interact with antd messages (e.g. copy the text)? Nothing comes to my mind.

:+1:

fm3:
Do we have a central spot for stuff that should be executed once on pageload? the router?

philippotto:
I think, here could be an okay spot for that:

https://github.com/scalableminds/webknossos/blob/65f15f4c3fd61c63c9d141ed6d18098807596ac0/frontend/javascripts/main.tsx#L90

fm3:
Yes, that works :) Went for 30, 100 looked a bit lost on the page

philippotto:
Sweet! yes, 100 was only an example value.

philippotto (approved):

Issue #6777, opened 2023-01-24
Active segment id doesn't respect dtype of segmentation


philippotto:
## Steps to Reproduce the bug
1. Add a dataset with segmentation layer in uint8
2. Set largest segment id to 255
3. Create an annotation
4. The current segment id is set to 256 which isn't part of uint8 (it's handled modulo 255 which makes it an eraser)

## Suggestion
- [ ] Ensure that the active segment id follows the dtype of the segmentation layer.
- [ ] Also, handle the special case where largest segment id equals the maximum value of the dtype (i.e., warn in the dataset settings and in the tracing UI since it's not possible to use a "new" id)

PR #7240, opened 2023-07-27
Prevent segmentation id from overflowing

MichaelBuessemeyer:
This PR prevents the segmentId to overflow and informs the user that overflowing the maximum possible segment id is not allowed.

- https://___.webknossos.xyz

### Steps to test:
- Go to the dataset configurations of a dataset with a segmentation layer
- Set the largestSegmentId setting of that segmentation layer to `2 ** <bit_depth> - 1`. This should show a warning that the maximum possible segmentId has been reached and thus annotations cannot create new segments. Save this change.
- Create an annotation of that dataset. The segmentation id should equal `2 ** <bit_depth> - 1`.
- Try to increase the segmentId by creating a new cell using e.g. the shortcut `c`. This should not update the segment id and show a warning toast, that the largest possible segment id must not be exceeded.
- Try to set the segmenId to something larger than `2 ** <bit_depth> - 1` by e.g. setting it manually in the footer bar. This should also show the toast and ignore the update.

### TODOs:
 ~~- [ ] Check whether it is ok to set the the `EMPYT_KEY_VALUE` of the cuckoo table to `0`? Maybe only do this for the segmentation cuckoo table.~~
  - We decided to simply accept this edge case for now where the `segmentId === EMPYT_KEY_VALUE`.

### Issues:
- fixes [#6777](https://github.com/scalableminds/webknossos/issues/6777)

------

### DIFF
--- a/frontend/javascripts/dashboard/dataset/dataset_settings_data_tab.tsx
+++ b/frontend/javascripts/dashboard/dataset/dataset_settings_data_tab.tsx
@@ -532,6 +532,17 @@ function SimpleLayerForm({
                     {
                       warningOnly: true,
                       validator: (_rule, value) =>
+                        value != null && value === 2 ** bitDepth - 1
+                          ? Promise.reject(
+                              new Error(
+                                `The largest segmentation ID has already reached the maximum possible value of 2^${bitDepth}-1. Annotations of this dataset cannot create new segments.`,
+                              ),
+                            )
+                          : Promise.resolve(),
+                    },
+                    {
+                      warningOnly: true,
+                      validator: (_rule, value) =>
                         value == null || value === ""
                           ? Promise.reject(
                               new Error(
--- a/frontend/javascripts/messages.tsx
+++ b/frontend/javascripts/messages.tsx
@@ -173,6 +173,9 @@ instead. Only enable this option if you understand its effect. All layers will n
     </span>
   ),
   "tracing.copy_cell_id": "Hit CTRL + I to copy the currently hovered segment id",
+  "tracing.segment_id_out_of_bounds": _.template(
+    "Cannot create a segment id larger than the segment layers maximum value of <%- maxSegmentId %>.",
+  ),
   "tracing.copy_maybe_mapped_cell_id":
     "Hit CTRL + I to copy the currently hovered segment id. Press CTRL + ALT + I if you want to copy the mapped id.",
   "tracing.no_more_branchpoints": "No more branchpoints",
--- a/frontend/javascripts/oxalis/model/accessors/dataset_accessor.ts
+++ b/frontend/javascripts/oxalis/model/accessors/dataset_accessor.ts
@@ -341,6 +341,11 @@ export function determineAllowedModes(settings?: Settings): {
     allowedModes,
   };
 }
+
+export function getMaximumSegmentIdForLayer(dataset: APIDataset, layerName: string) {
+  return getDefaultIntensityRangeOfLayer(dataset, layerName)[1];
+}
+
 export function getBitDepth(layerInfo: DataLayer | DataLayerType): number {
   switch (layerInfo.elementClass) {
     case "uint8":
--- a/frontend/javascripts/oxalis/model/accessors/dataset_accessor.ts
+++ b/frontend/javascripts/oxalis/model/accessors/dataset_accessor.ts
@@ -343,7 +343,8 @@ export function determineAllowedModes(settings?: Settings): {
 }
 
 export function getMaximumSegmentIdForLayer(dataset: APIDataset, layerName: string) {
-  return getDefaultIntensityRangeOfLayer(dataset, layerName)[1];
+  const valueRange = getDefaultIntensityRangeOfLayer(dataset, layerName);
+  return valueRange[1];
 }
 
 export function getBitDepth(layerInfo: DataLayer | DataLayerType): number {
--- a/frontend/javascripts/oxalis/model/accessors/dataset_accessor.ts
+++ b/frontend/javascripts/oxalis/model/accessors/dataset_accessor.ts
@@ -193,7 +193,7 @@ export function getByteCount(dataset: APIDataset, layerName: string): number {
 export function getElementClass(dataset: APIDataset, layerName: string): ElementClass {
   return getLayerByName(dataset, layerName).elementClass;
 }
-export function getDefaultIntensityRangeOfLayer(
+export function getDefaultValueRangeOfLayer(
   dataset: APIDataset,
   layerName: string,
 ): [number, number] {
@@ -343,8 +343,7 @@ export function determineAllowedModes(settings?: Settings): {
 }
 
 export function getMaximumSegmentIdForLayer(dataset: APIDataset, layerName: string) {
-  const valueRange = getDefaultIntensityRangeOfLayer(dataset, layerName);
-  return valueRange[1];
+  return getDefaultValueRangeOfLayer(dataset, layerName)[1];
 }
 
 export function getBitDepth(layerInfo: DataLayer | DataLayerType): number {
--- a/frontend/javascripts/oxalis/model/actions/volumetracing_actions.ts
+++ b/frontend/javascripts/oxalis/model/actions/volumetracing_actions.ts
@@ -10,7 +10,7 @@ import { batchActions } from "redux-batched-actions";
 
 export type InitializeVolumeTracingAction = ReturnType<typeof initializeVolumeTracingAction>;
 export type InitializeEditableMappingAction = ReturnType<typeof initializeEditableMappingAction>;
-type CreateCellAction = ReturnType<typeof createCellAction>;
+export type CreateCellAction = ReturnType<typeof createCellAction>;
 type StartEditingAction = ReturnType<typeof startEditingAction>;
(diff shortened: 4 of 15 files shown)


frontend/javascripts/oxalis/model/accessors/dataset_accessor.ts
```diff
@@ -341,6 +341,11 @@ export function determineAllowedModes(settings?: Settings): {
     allowedModes,
   };
 }
+
+export function getMaximumSegmentIdForLayer(dataset: APIDataset, layerName: string) {
+  return getDefaultIntensityRangeOfLayer(dataset, layerName)[1];
```
philippotto, line 346: I think intensity range should be renamed to value range in this function, because intensity doesn't make sense in the context of segmentations.

frontend/javascripts/oxalis/model/sagas/volumetracing_saga.tsx
```diff
@@ -130,6 +135,25 @@ function* warnOfTooLowOpacity(): Saga<void> {
   }
 }
 
+function* warnTooLargeSegmentId(): Saga<void> {
+  yield* take("INITIALIZE_VOLUMETRACING");
+  let oldSegmentId = yield* select((state) => enforceActiveVolumeTracing(state).activeCellId);
+  while (true) {
+    yield* take(["SET_ACTIVE_CELL", "CREATE_CELL"]);
+    const newSegmentId = yield* select((state) => enforceActiveVolumeTracing(state).activeCellId);
+    if (oldSegmentId === newSegmentId) {
```
philippotto, line 144: I think, this logic is flawed. 1) oldSegmentId is never updated. 2) If I select an already active segment id, the condition will eval to true, I think? So, if the warning can also appear for uncritical segment ids (e.g., 3 is selected and I select it again somehow).

What do you think about using the "requested segment id" in the actions? I think, both set-active-cell and create-cell get a segment id parameter (probably named differently). That id could be checked against `maxSegmentId` then.

frontend/javascripts/dashboard/dataset/dataset_settings_data_tab.tsx
```diff
@@ -529,6 +529,17 @@ function SimpleLayerForm({
                               ),
                             ),
                     },
+                    {
+                      warningOnly: true,
+                      validator: (_rule, value) =>
+                        value != null && value === 2 ** bitDepth - 1
+                          ? Promise.reject(
```
philippotto, line 536: Why is this an error? Maybe the segmentation has used up all available ids? Ideally, it should be a warning, I think.

philippotto (commented):
already left some comments, even though the review wasn't requested yet :see_no_evil:

frontend/javascripts/oxalis/model/sagas/volumetracing_saga.tsx
MichaelBuessemeyer: > I think, this logic is flawed. 1) oldSegmentId is never updated. 2) If I select an already active segment id, the condition will eval to true, I think? So, if the warning can also appear for uncritical segment ids (e.g., 3 is selected and I select it again somehow).

Oh thanks for catching this flaw. :pray: 

> What do you think about using the "requested segment id" in the actions? I think, both set-active-cell and create-cell get a segment id parameter (probably named differently). That id could be checked against maxSegmentId then.

:+1:

frontend/javascripts/dashboard/dataset/dataset_settings_data_tab.tsx
MichaelBuessemeyer: This is not an error. The field `warningOnly` configures the validator to show the rejected message as a warning.
image

frontend/javascripts/oxalis/model/accessors/dataset_accessor.ts
MichaelBuessemeyer: Is this what you intended? or should I rename the whole function `getDefaultIntensityRangeOfLayer` to something like `getDefaultValueRangeOfLayer`?

frontend/javascripts/oxalis/model/actions/volumetracing_actions.ts
```diff
@@ -126,11 +126,16 @@ export const initializeEditableMappingAction = (mapping: ServerEditableMapping)
...
+  // that the activeCellId is larger than the largestSegmentId. Choose the larger of the two ids increased by one.
+  const newSegmentId =
+    largestSegmentId && largestSegmentId > activeCellId ? largestSegmentId + 1 : activeCellId + 1;
+  return {
     type: "CREATE_CELL",
-    largestSegmentId,
-  } as const);
+    newSegmentId,
+  } as const;
+};
```
MichaelBuessemeyer, line 139: Here I followed your suggestion to add the desired updated `activeCellId` value to the action. This required pulling the logic of choosing between `largestSegmentId` and `activeCellId` out of the reducer and frontloading it to the code that creates the action. Or do you see another option to not frontload the logic but still be able to add the `newSegmentId` parameter to the action?

I am not perfectly happy with putting the logic of choosing the larger value here. But otherwise, the logic would need to appear before every call to the `createCellAction` action function and thus would have more code duplication. Thus I would say that this is the better solution under the given circumstances.

frontend/javascripts/oxalis/model/accessors/dataset_accessor.ts
philippotto: > or should I rename the whole function getDefaultIntensityRangeOfLayer to something like getDefaultValueRangeOfLayer?

This is what I meant. Sorry that my initial comment wasn't clear enough :see_no_evil:

frontend/javascripts/messages.tsx
```diff
@@ -173,6 +173,9 @@ instead. Only enable this option if you understand its effect. All layers will n
     </span>
   ),
   "tracing.copy_cell_id": "Hit CTRL + I to copy the currently hovered segment id",
+  "tracing.segment_id_out_of_bounds": _.template(
+    "Cannot create a segment id larger than the segment layers maximum value of <%- maxSegmentId %>.",
+  ),
```
MichaelBuessemeyer, line 178: This message shows the maximum value directly and not in a more readable format like `2^32 -1`. 
Pro:
The direct value has the advantage to show the use of the maximum allowed segment id and therefore the user does not need to calculate the maximum value of `2^32 -1` on their own using a calculator or such. 

Con: 
It looks visually not very good.

Which one do you prefer @philippotto ?

frontend/javascripts/dashboard/dataset/dataset_settings_data_tab.tsx
philippotto: Ah, great, nevermind then :)

frontend/javascripts/oxalis/model/actions/volumetracing_actions.ts
philippotto: I think your current way is currently the best trade-off :+1:

frontend/javascripts/oxalis/model/accessors/dataset_accessor.ts
philippotto: This still needs a renaming, right?

frontend/javascripts/oxalis/model/accessors/dataset_accessor.ts
```diff
@@ -341,6 +341,12 @@ export function determineAllowedModes(settings?: Settings): {
     allowedModes,
   };
 }
+
+export function getMaximumSegmentIdForLayer(dataset: APIDataset, layerName: string) {
+  const valueRange = getDefaultIntensityRangeOfLayer(dataset, layerName);
+  return valueRange[1];
+}
+
```
MichaelBuessemeyer, line 349: @philippotto To me it looks like I successfully renamed it. Maybe you overlooked the `outdated` above the code lines where you suggested to give the result of `getDefaultIntensityRangeOfLayer(dataset, layerName);` a name?

frontend/javascripts/oxalis/model/accessors/dataset_accessor.ts
MichaelBuessemeyer: see my comment below please :arrow_down:

frontend/javascripts/oxalis/model/accessors/dataset_accessor.ts
philippotto: I think, we are talking about different things :sweat_smile: My suggestion is to rename the function `getDefaultIntensityRangeOfLayer` to `getDefaultValueRangeOfLayer`. That's what I tried to express in https://github.com/scalableminds/webknossos/pull/7240#discussion_r1283122533 where I quoted you asking whether this should be done. You probably interpreted my "this is what I meant" so that I confirmed that changes you did instead of your quote?

MichaelBuessemeyer: > You probably interpreted my "this is what I meant" so that I confirmed that changes you did instead of your quote?

Yes indeed :see_no_evil: xD, sorry for the confusion

MichaelBuessemeyer: Is there anything else you noticed or is this pr ready to :shipit: ?

PR #7242, opened 2023-07-31
Segment multiselect

### LINKED ISSUE #6954: Segments Groups

dieknolle3333:
### Steps to test:
- Go to segments tab in any annotation, preferably one with a mesh file
- Press shift and select more than one segment
- Right-click to open the context menu like you are used to
- Select any action you would like to execute on all selected segments, such as loading all meshes
- Make sure this also works for meshes in different groups
- Make sure all actions are working as expected
- Make sure the visibility toggle for "Show /Hide meshes is working"
- Make sure selecting a group after having selected multiple segments opens a toast with a warning like in the skeleton tab
- After having selected meshes, open a group context menu and move all selected segments into the group

### TODOs:
- [x] self-review code before making the PR ready-for-review
- [x] fix case where group is selected after having selected multiple segments, but ctrl is not pressed (currently missing warning toast)

### Issues:
- partly fixes #6954 

------


frontend/javascripts/oxalis/view/right-border-tabs/segments_tab/segment_list_item.tsx
```diff
@@ -455,7 +457,7 @@ function _SegmentListItem({
         padding: "2px 5px",
       }}
       className={classnames("segment-list-item", {
-        "is-selected-cell": segment.id === selectedSegmentId,
+        "is-selected-cell": selectedSegmentIds?.includes(segment.id), //TODO this wont work
```
dieknolle3333, line 460: what exactly does this do? If I set it to false the segments are selected nonetheless?

dieknolle3333: (will delete the TODO comment obviously)

dieknolle3333: so this actually makes the coloring a little better, but the box thats created while hovering has a different size than the box thats created upon selection, so I will see whether I can maybe improve this.

frontend/javascripts/oxalis/view/right-border-tabs/segments_tab/segment_list_item.tsx
```diff
@@ -466,7 +465,11 @@ function _SegmentListItem({
       }}
     >
       <Dropdown
-        menu={createSegmentContextMenu()} // The overlay is generated lazily. By default, this would make the overlay
+        menu={
+          selectedSegmentIds?.includes(segment.id) && selectedSegmentIds.length > 1
```
philippotto, line 469: I'd swap the two terms:

```suggestion
          selectedSegmentIds.length > 1 && selectedSegmentIds?.includes(segment.id)
```

it will be a bit more performant, since the greater-check is faster (and if it fails, the second doesn't need to be done).

frontend/javascripts/oxalis/view/right-border-tabs/segments_tab/segments_view.tsx
```diff
@@ -1088,39 +1241,82 @@ class SegmentsView extends React.Component<Props, State> {
...
+  };
+
+  getSegmentOrGroupIdsForKeys = (segmentOrGroupKeys: Key[]) => {
+    let selectedIds: { segments: number[]; group: number | null } = { segments: [], group: null };
+    segmentOrGroupKeys.forEach((key) => {
+      const keyAsString = String(key);
+      if (keyAsString.startsWith("group-")) {
+        const regexSplit = keyAsString.split(/-(-?)/);
+        // concatenate id with negative sign for root group. for positive ids, regexSplit[2] is empty
+        const idWithSign = regexSplit[1].concat(regexSplit[2]);
```
philippotto, line 1296: I think, this can be simplified to:

```suggestion
        // Note that negative ids can be found here, which is why group- is used as a splitter
        const idWithSign = keyAsString.split(/group-/)[1];
```

frontend/javascripts/oxalis/view/right-border-tabs/segments_tab/segment_list_item.tsx
dieknolle3333: I understand! This leads to an error though because `selectedSegmentIds` may be null and if the `includes` check is done first and the ids are null, the second one isn't done (if I understand correctly)

frontend/javascripts/oxalis/view/right-border-tabs/segments_tab/segments_view.tsx
dieknolle3333: damn, you're right :D

frontend/javascripts/oxalis/view/right-border-tabs/segments_tab/segment_list_item.tsx
philippotto: Ah, my bad, then how about `(selectedSegmentIds || []).length > 1 && selectedSegmentIds?.includes(segment.id)` ?

dieknolle3333: sounds good, I will do that!

Issue #6926, opened 2023-03-16
[Follow-up] Multi modality (Limitations)


philippotto:
- [ ] f/d might use inappropriate step width in z (see `moveW`)
- [ ] ...

PR #7246, opened 2023-08-03
Allow dynamic change of natively rendered layer (by re-applying transforms)

philippotto:
- https://___.webknossos.xyz

### Steps to test:
- open [C555_DIAMOND_2f-TPS](https://toggletransforms.webknossos.xyz/datasets/sample_organization/C555_DIAMOND_2f-TPS/view#500,450,321,0,0.621) on the dev instance
- use the new transform-toggle-buttons next to the layers
- the position and zoom should adapt when using the toggle button so that one stays roughly in the same area
- create an annotation
- volume annotation should only be possible when the volume layer is not transformed

### TODOs:
- [x] test TPS and affine chaining further
- [x] adapt position when toggling transforms
- [x] adapt zoom when toggling transforms ?
- [x] todo-comments / refactor source/target name mismatch

### Issues:
- contributes to #[6926](https://github.com/scalableminds/webknossos/issues/6926)

------


philippotto:
@daniel-wer I think the PR is ready for a complete review. There was some inverted source<>target semantics in the code which was a bit annoying to clean up. I would recommend to **not** review commit-by-commit, since lots of changes cancelled out (e.g., I renamed source and target and then swapped the order in another commit + some bug fixes where it was simply wrong). I hope this doesn't complicate the review vastly :pray:

normanrz (approved):
Nice stuff!

PR #7249, opened 2023-08-04
Add UI for segment statistics (volume and bbox)

### LINKED ISSUE #7145: Use segment statistics routes in frontend
#7063 already introduced a route to query the volume of a segment in voxels.
(works only if a segment index is present, so currently only possible for tracings with no fallback segmentation)

More routes might follow.

It would be nice for the user to be able to access this information.

dieknolle3333:
#### for a single segment
- Go to annotation without fallback layer
- Brush a segment
- Right-click the segment and read its volume
- Copy the size to clipboard
#### in segments tab
- right-click on any group to open its context menu
- select `Show Segment Statistics`
- view table and download CSV

### TODOs:
- [x] handle update of size by having context menu visibility as dependency
- [x] improve icon
- [x] implement reload button with tooltip "update this statistic", that saves the annotation
- [x] mock-up of size in segment tab
- [x] get size of all successor segments of a group recursively, not only direct children
- [x] add spinner while segment statistics are loading
- [x] make sure its the latest size, so that annotation was saved recently when the modal is opened
- [x] show bounding box in columns and in CSV after the route was built
- [x] improve naming of CSV
- [x] get group name
- [x] include bounding box info in segment context menu
- [x] make group actions recursive
- [x] check whether size in nm³ and bouding boxes are actually correct
- [x] maybe improve how root group is defined by not just checking for null

### Issues:
- fixes #7145 
- contributes to #7319
------


frontend/javascripts/libs/format_utils.ts
```diff
@@ -144,6 +144,36 @@ export function formatNumberToLength(lengthInNm: number): string {
...
+  [1e6, "µm²"],
+  [1e12, "mm²"],
+  [1e18, "m²"],
+  [1e24, "km²"],
+]);
+/* TODO: or rather: 
+const nmFactorToUnit2D = new Map(
+  Array.from(nmFactorToUnit).map((entry) => [Math.pow(entry[0], 2), entry[1].concat("²")]),
+); 
+Although I think that the explicit maps are better to read and it's better to have them stored rather than computing them every time they are needed.
```
dieknolle3333, line 159: TODO

frontend/javascripts/admin/admin_rest_api.ts
```diff
@@ -958,14 +958,32 @@ export function getNewestVersionForTracing(
...
+
 export async function importVolumeTracing(
   tracing: Tracing,
   volumeTracing: VolumeTracing,
   dataFile: File,
 ): Promise<number> {
   return doWithToken((token) =>
     Request.sendMultipartFormReceiveJSON(
-      `${tracing.tracingStore.url}/tracings/volume/${volumeTracing.tracingId}/importVolumeData?token=${token}`,
+      `${tracing.tracingStore.url}/tracings/volume/${volumeTracing}/importVolumeData?token=${token}`,
```
philippotto, line 986: I guess, this change was accidental?

fm3:
@dieknolle3333 I added a draft for the route for the segment bounding box. Find it at `/volume/:tracingId/segmentStatistics/boundingBox` (same GET parameters as for volume).

I did not build the real implementation in the backend yet, so for now the route will always return this json:
`{"topLeft":[5,5,5],"width":10,"height":20,"depth":30}`

I hope to get it to return the correct numbers soon.

Note that the bounding box (just as the volume) will be in voxels *in the requested mag*.

fm3:
@dieknolle3333 as discussed, I changed both routes to POST. Example usage via CURL:

```
f@x:~$ 28,0]f@x:~$ curl -X POST 'http://localhost:9000/tracings/volume/b34f4e39-16f1-4ff0-9085-fc13a7036242/segmentStatistics/volume?token=lnSyyDm5JETXLebvgz3zug' -H "Content-Type: application/json" -d '{"mag"" -d '{"mag":[1,1,1], "segmentIds":[0,1,2]}'

[0,124928,0]

f@x:~$ curl -X POST 'http://localhost:9000/tracings/volume/b34f4e39-16f1-4ff0-9085-fc13a7036242/segmentStatistics/boundingBox?token=lnSyyDm5JETXLebvgz3zug' -H "Content-Type: application/json" -d '{"mag":[1,1,1], "segmentIds":[0,1,2]}'

[{"topLeft":[0,0,0],"width":0,"height":0,"depth":0},{"topLeft":[100,104,60],"width":324,"height":284,"depth":4},{"topLeft":[0,0,0],"width":0,"height":0,"depth":0}]

```

webknossos-tracingstore/app/com/scalableminds/webknossos/tracingstore/tracings/volume/VolumeSegmentStatisticsService.scala
```diff
@@ -30,14 +35,102 @@ class VolumeSegmentStatisticsService @Inject()(volumeTracingService: VolumeTraci
...
+
+  private def extendBouningBoxByData(tracing: VolumeTracing,
+                                     tracingId: String,
+                                     mag: Vec3Int,
+                                     mutableBoundingBox: scala.collection.mutable.ListBuffer[Int],
+                                     bucketPosition: Vec3Int,
+                                     userToken: Option[String]): Fox[Unit] =
+    for {
+      bucketData <- data(tracing, tracingId, mag, Seq(bucketPosition), userToken)
+      dataTyped: Array[UnsignedInteger] = UnsignedIntegerArray.fromByteArray(bucketData, tracing.elementClass)
```
normanrz, line 89: 😱

webknossos-tracingstore/app/com/scalableminds/webknossos/tracingstore/tracings/volume/VolumeSegmentStatisticsService.scala
```diff
@@ -30,14 +35,102 @@ class VolumeSegmentStatisticsService @Inject()(volumeTracingService: VolumeTraci
...
+                                          position: Vec3Int): Unit = {
+    mutableBoundingBox(0) = Math.min(mutableBoundingBox(0), position.x)
+    mutableBoundingBox(1) = Math.min(mutableBoundingBox(1), position.y)
+    mutableBoundingBox(2) = Math.min(mutableBoundingBox(2), position.z)
+    mutableBoundingBox(3) = Math.max(mutableBoundingBox(3), position.x)
+    mutableBoundingBox(4) = Math.max(mutableBoundingBox(4), position.y)
+    mutableBoundingBox(5) = Math.max(mutableBoundingBox(5), position.z)
+  }
+
   private def data(tracing: VolumeTracing,
```
normanrz, line 119: Just `data`?

frontend/javascripts/oxalis/view/context_menu.tsx
```diff
@@ -1107,6 +1112,46 @@ function ContextMenuInner(propsWithInputRef: Props) {
...
 
+  const segmentIdAtPosition = globalPosition != null ? getSegmentIdForPosition(globalPosition) : 0;
+  const { visibleSegmentationLayer, volumeTracing } = props;
+  const hasNoFallbackLayer =
+    visibleSegmentationLayer != null &&
+    "fallbackLayer" in visibleSegmentationLayer &&
+    visibleSegmentationLayer.fallbackLayer == null;
+  const [segmentSize, boundingBoxInfo] = useFetch(
+    async () => {
+      if (contextMenuPosition == null) return [];
```
dieknolle3333, line 1123: not sure why, but if I include this in the checks in the next line, I get the error `Can't perform a React state update on an unmounted component`

frontend/javascripts/oxalis/view/right-border-tabs/segments_tab/segment_statistics_modal.tsx
```diff
@@ -0,0 +1,171 @@
...
+  visibleSegmentationLayer,
+  relevantSegments: segments,
+  parentGroup,
+  groupTree,
+}: Props) {
+  const mag = getResolutionInfo(visibleSegmentationLayer.resolutions);
+  const nmFactorToUnit = new Map([[1, "nm³"]]);
+  const dataSource = useFetch(
+    async () => {
+      await new Promise((resolve) => setTimeout(resolve, 10000)); //TODO delete; only for testing
```
dieknolle3333, line 77: dont forget about me

frontend/javascripts/oxalis/view/right-border-tabs/segments_tab/segment_statistics_modal.tsx
```diff
@@ -0,0 +1,171 @@
...
+      await api.tracing.save();
```
dieknolle3333, line 78: TODO: maybe omit

dieknolle3333: I think this is the best solution if there is no way clear way to determine whether there are unsaved changes

dieknolle3333: (which I think there isnt because we could have used it in the segment context menu in the viewport)

frontend/javascripts/oxalis/view/right-border-tabs/segments_tab/segment_statistics_modal.tsx
```diff
@@ -0,0 +1,171 @@
...
+        const statisticsObjects = [];
+        for (let i = 0; i < segments.length; i++) {
+          // segments in request and their statistics in response are in the same order
+          const currentSegment = segments[i];
+          const currentBoundingBox = boundingBoxes[i];
+          const segmentStatObject = {
+            segmentId: currentSegment.id,
+            segmentName:
+              currentSegment.name == null ? `Segment ${currentSegment.id}` : currentSegment.name,
+            groupId: currentSegment.groupId == null ? -1 : currentSegment.groupId,
```
dieknolle3333, line 104: I know that there could be cases where the groupId is null though the segment is not in the root group. due to the lack of time today, I didn't implement a better way to find this out, but I can still add a function to find this out!

frontend/javascripts/admin/admin_rest_api.ts
```diff
@@ -958,6 +958,38 @@ export function getNewestVersionForTracing(
   );
 }
 
+export function getSegmentVolumes(
```
dieknolle3333, line 961: I really cant tell whether these numbers make any sense so it would be nice if someone could make sure they do :)

dieknolle3333: Not that I think the backend is incorrect, I just hope the frontend parsing is correct

fm3: You could test this by zooming in a lot and setting the brush size really tiny, so you can have segment volumes of 1 voxel (or 2, or 3, etc). If these match (in vx), you could also manually check the nm³ conversion

dieknolle3333: and how about the bounding boxes?

fm3: Same thing I’d say, you could use a 1vx-sized brush to draw a 2d shape like
```
xxx
  x
```
…at a known position, and then check that the volume is 4vx, and that the bbox size is 3×2×1. Of course, this can be extended in 3d as well.

CHANGELOG.unreleased.md
```diff
@@ -14,6 +14,7 @@ For upgrade instructions, please check the [migration guide](MIGRATIONS.released
 - Added disabled drag handles to volume and skeleton layers for visual consistency. These layer cannot be dragged or reordered. [#7295](https://github.com/scalableminds/webknossos/pull/7295)
 - Dataset thumbnails for grayscale layers can now be colored using the value in the view configuration. [#7255](https://github.com/scalableminds/webknossos/pull/7255)
 - OpenID Connect authorization is now compatible with Providers that send the user information in an id_token. [#7294](https://github.com/scalableminds/webknossos/pull/7294)
+- Volume and bounding box information is shown in segments' context menus aswell as in a seperate modal in the segments tab. There is also an option to export the statistics. [#7249](https://github.com/scalableminds/webknossos/pull/7249)
```
philippotto, line 17:
```suggestion
- Volume and bounding box information is shown in segments' context menus as well as in a seperate modal in the segments tab. There is also an option to export the statistics. [#7249](https://github.com/scalableminds/webknossos/pull/7249)
```

frontend/javascripts/libs/format_utils.ts
philippotto: yes, I think it's fine to leave these hardcoded.

frontend/javascripts/oxalis/view/context_menu.tsx
```diff
@@ -1093,6 +1097,7 @@ function getInfoMenuItem(
 }
 
 function ContextMenuInner(propsWithInputRef: Props) {
+  const [lastTimeSegmentInfoWasFetched, setLastTimeSegmentInfoWasFetched] = useState(new Date());
```
philippotto, line 1100: I'd rename these to `...lastTimeSegmentInfoShouldBeFetched` because when you set it, it wasn't fetched yet. Instead, you use the variable to trigger the fetch.

frontend/javascripts/oxalis/view/context_menu.tsx
philippotto: I'm not sure why, either. probably unrelated to that: you are not returning a tuple here, which could make the unpacking in line 1121 throw an error. seems fishy to me :)

frontend/javascripts/oxalis/view/context_menu.tsx
```diff
@@ -1107,6 +1112,45 @@ function ContextMenuInner(propsWithInputRef: Props) {
...
+  const segmentIdAtPosition = globalPosition != null ? getSegmentIdForPosition(globalPosition) : 0;
+  const { visibleSegmentationLayer, volumeTracing } = props;
+  const hasNoFallbackLayer =
+    visibleSegmentationLayer != null &&
+    "fallbackLayer" in visibleSegmentationLayer &&
+    visibleSegmentationLayer.fallbackLayer == null;
+  const [segmentVolume, boundingBoxInfo] = useFetch(
+    async () => {
+      if (contextMenuPosition == null) return [];
+      if (visibleSegmentationLayer != null && volumeTracing != null && hasNoFallbackLayer) {
```
philippotto: in the else-branch, nothing is returned, right?

frontend/javascripts/oxalis/view/context_menu.tsx
```diff
@@ -1107,6 +1112,45 @@ function ContextMenuInner(propsWithInputRef: Props) {
...
+          formatNumberToVolume(segmentSize),
+          `${boundingBoxTopLeftString}, ${boundingBoxSizeString}`,
+        ];
+      }
+    },
+    ["loading", "loading"],
+    // Update segment infos when opening the context menu, in case the annotation was saved since the context menu was last opened.
+    // Of course the info should also be updated when the menu is opened for another segment, or after the refresh button was pressed.
+    [contextMenuPosition, segmentIdAtPosition, lastTimeSegmentInfoWasFetched],
+  ) as [string, string];
```
philippotto: the function doesn't always return a string-tuple. if you fix that, you shouldn't need the casting.

frontend/javascripts/oxalis/view/context_menu.tsx
```diff
@@ -1107,6 +1112,45 @@ function ContextMenuInner(propsWithInputRef: Props) {
...
+    async () => {
+      if (contextMenuPosition == null) return [];
+      if (visibleSegmentationLayer != null && volumeTracing != null && hasNoFallbackLayer) {
+        const tracingId = volumeTracing.tracingId;
+        const tracingStoreUrl = Store.getState().tracing.tracingStore.url;
+        const mag = getResolutionInfo(visibleSegmentationLayer.resolutions);
+        const [segmentSize] = await getSegmentVolumes(
+          tracingStoreUrl,
+          tracingId,
+          mag.getHighestResolution(),
```
philippotto, line 1132: I think (99% sure but please test), `getHighestResolution` returns the coarsest (poorest) resolution. The wording is poor and should be renamed, since "high res" sounds like the best quality, but in fact this refers to the high numbers (e.g., mag 32) which indicates a high degree of downsampling (== information loss).
We should rename this to coarsest and finest resolution, but not in this PR. 
For now, you should use `getLowestResolution` here and in line 1137.

frontend/javascripts/test/libs/format_utils.spec.ts
```diff
@@ -0,0 +1,51 @@
+import test from "ava";
+import { formatNumberToArea, formatNumberToLength, formatNumberToVolume } from "libs/format_utils";
+import { Unicode } from "oxalis/constants";
+
+test("Format number to length", (t) => {
+  const { ThinSpace } = Unicode;
+  const lengthsInNm = [107e-3, 107, 107e3, 107e6, 107e9, 107e12];
```
philippotto, line 7: please also use some other numbers to make the tests a bit more thorough. e.g., I'd like to see the following result strings:
`1${ThinSpace}mm`
`12${ThinSpace}nm`
`0.01${ThinSpace}pm`
`1234${ThinSpace}km`

frontend/javascripts/oxalis/view/right-border-tabs/segments_tab/segments_view.tsx
```diff
@@ -932,7 +924,7 @@ class SegmentsView extends React.Component<Props, State> {
       disabled: isEditingDisabled,
       icon: (
         <i
-          className="fas fa-eye-dropper fa-sm fa-icon"
+          className="fas fa-eye-dropper fa-sm fa-icon fa-fw "
```
philippotto, line 927:
```suggestion
          className="fas fa-eye-dropper fa-sm fa-icon fa-fw"
```

what does `fa-fw` do?

frontend/javascripts/oxalis/view/right-border-tabs/segments_tab/segment_statistics_modal.tsx
```diff
@@ -0,0 +1,173 @@
...
+  "segmendId,segmentName,groupId,groupName,volumeInVoxel,volumeInNm3,boundingBoxTopLeftPositionX,boundingBoxTopLeftPositionY,boundingBoxTopLeftPositionZ,boundingBoxSizeX,boundingBoxSizeY,boundingBoxSizeZ";
+
+const { ThinSpace } = Unicode;
+
+type Props = {
+  onCancel: (...args: Array<any>) => any;
+  isOpen: boolean;
+  tracingId: any;
+  tracingStoreUrl: any;
+  visibleSegmentationLayer: any;
```
philippotto, line 23: can we get rid of the anys? ;)

frontend/javascripts/oxalis/view/right-border-tabs/segments_tab/segment_statistics_modal.tsx
```diff
@@ -0,0 +1,173 @@
...
+  boundingBoxTopLeft: Vector3;
+  boundingBoxPosition: Vector3;
+};
+
+const exportStatisticsToCSV = (
+  segmentInformation: Array<SegmentInfo>,
+  tracingId: string,
+  groupId: number,
+) => {
+  if (segmentInformation.length < 0) {
```
philippotto, line 46: did you mean `===`? the length of an array will never be negative..

frontend/javascripts/oxalis/view/right-border-tabs/segments_tab/segment_statistics_modal.tsx
```diff
@@ -0,0 +1,173 @@
...
+    return;
+  }
+  const segmentStatisticsAsString = segmentInformation
+    .map(
+      (segmentInfo) =>
+        `${segmentInfo.segmentId},${segmentInfo.segmentName},${segmentInfo.groupId},${segmentInfo.groupName},${segmentInfo.volumeInVoxel},${segmentInfo.volumeInNm3},${segmentInfo.boundingBoxTopLeft},${segmentInfo.boundingBoxPosition}`,
```
philippotto, line 52: I'd write it like this `[segmentInfo.segmentId, segmentInfo.segmentName, ...].join(",")`

dieknolle3333:
@philippotto so it would be easily possible for me to spend another day testing and reviewing my own code. we planned that I'd finish this PR by yesterday so that I can move on to the next issue, so I fixed bigger issues that occured to me. Having this in mind I am looking forward to your feedback!

frontend/javascripts/oxalis/view/right-border-tabs/segments_tab/segment_statistics_modal.tsx
```diff
@@ -0,0 +1,173 @@
...
+  formattedSize: string;
+  volumeInVoxel: number;
+  boundingBoxTopLeft: Vector3;
+  boundingBoxPosition: Vector3;
+};
+
+const exportStatisticsToCSV = (
+  segmentInformation: Array<SegmentInfo>,
+  tracingId: string,
+  groupId: number,
```
philippotto, line 44: Is this the group id for which the user selected the export? If so, I'd rename it to `groupIdToExport` or something like that.

frontend/javascripts/oxalis/view/right-border-tabs/segments_tab/segment_statistics_modal.tsx
```diff
@@ -0,0 +1,173 @@
...
+            segmentName:
+              currentSegment.name == null ? `Segment ${currentSegment.id}` : currentSegment.name,
+            groupId: currentSegment.groupId == null ? -1 : currentSegment.groupId,
+            groupName: getGroupNameForId(
+              currentSegment.groupId == null ? -1 : currentSegment.groupId,
+            ),
+            volumeInVoxel: segmentSizes[i],
+            volumeInNm3: parseInt(
+              formatNumberToUnit(segmentSizes[i], nmFactorToUnit).split(ThinSpace)[0],
+            ),
```
philippotto, line 112: this formats the number to an appropriate unit (which isn't necessarily nm³) and then strips that unit away and puts it into `volumeInNm3`? I think, that's wrong? :sweat_smile:
Also, the actual vx->nm conversion is missing (this isn't done by formatNumberToUnit). You need to multiply the size with the datasource.scale and then you are already done, I think.

frontend/javascripts/oxalis/view/right-border-tabs/segments_tab/segment_statistics_modal.tsx
```diff
@@ -0,0 +1,173 @@
...
+            formattedSize: formatNumberToVolume(segmentSizes[i]),
```
philippotto, line 113: here, the conversion from vx³->nm³ is also missing.

frontend/javascripts/oxalis/view/right-border-tabs/segments_tab/segment_statistics_modal.tsx
```diff
@@ -0,0 +1,173 @@
...
+            boundingBoxTopLeft: currentBoundingBox.topLeft,
+            boundingBoxTopLeftAsString: `(${currentBoundingBox.topLeft[0]}, ${currentBoundingBox.topLeft[1]}, ${currentBoundingBox.topLeft[2]}) `,
```
philippotto, line 115:
```suggestion
            boundingBoxTopLeftAsString: `(${currentBoundingBox.topLeft.join(", ")})`,
```

philippotto (changes requested):
nice stuff :) I already left some feedback. let me know in case you have questions :call_me_hand:

frontend/javascripts/oxalis/view/context_menu.tsx
dieknolle3333: this makes a lot more sense. the bounding boxes were off all along, and I didnt test them thoroughly yet. now they look much more realistic upon first view, but I will follow Flo's advice and make sure that they still seem to be correct upon second and third view

frontend/javascripts/oxalis/view/right-border-tabs/segments_tab/segment_statistics_modal.tsx
dieknolle3333: 🥴

frontend/javascripts/oxalis/view/right-border-tabs/segments_tab/segments_view.tsx
```diff
@@ -1036,11 +1028,39 @@ class SegmentsView extends React.Component<Props, State> {
     };
   };
 
+  getShowSegmentStatistics = (id: number): ItemType => {
+    const visibleSegmentationLayer = this.props.visibleSegmentationLayer;
+    if (
+      visibleSegmentationLayer == null ||
```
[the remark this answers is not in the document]
dieknolle3333: @philippotto do you think the entry in the context menu should be disabled if statistics arent available, or should the entry simply not appear? not sure if we want to nudge users to this feature. i could add a tooltip explaining what the user needs to do in order to get the statistics

frontend/javascripts/oxalis/view/context_menu.tsx
dieknolle3333: when testing it, it didnt throw any errors, but i agree that this should be improved!

frontend/javascripts/oxalis/view/context_menu.tsx
dieknolle3333: i think it should be resolved with the current changes!

frontend/javascripts/oxalis/view/right-border-tabs/segments_tab/segments_view.tsx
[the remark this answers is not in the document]
philippotto: I think, it's fine to hide the entry in that case, since the user won't be able to change it (because we don't support on-disk-segmentations yet).

frontend/javascripts/oxalis/view/right-border-tabs/segments_tab/segment_statistics_modal.tsx
dieknolle3333: to be fair my `nmFactorToUnit` only consisted of nm^3, but of course you're right that I lacked understanding in the actual conversion 😅

frontend/javascripts/oxalis/view/right-border-tabs/segments_tab/segments_view.tsx
dieknolle3333: it makes all the icons have the same width (https://fontawesome.com/v5/docs/web/style/fixed-width)

frontend/javascripts/oxalis/view/right-border-tabs/segments_tab/segment_statistics_modal.tsx
dieknolle3333: I had to change `tsconfig`in order to use `replaceAll`(>=only with es2021), I hop that's okay!

frontend/javascripts/oxalis/view/right-border-tabs/segments_tab/segment_statistics_modal.tsx
dieknolle3333: actually this doesnt seem to work. maybe we can look into this together once we have the time?

frontend/javascripts/test/libs/format_utils.spec.ts
dieknolle3333: I tried :) please let me know if I can still improve the tests!

webknossos-datastore/app/com/scalableminds/webknossos/datastore/helpers/ProtoGeometryImplicits.scala
```diff
@@ -1,10 +1,6 @@
 package com.scalableminds.webknossos.datastore.helpers
 
 import com.scalableminds.util.geometry.{BoundingBox, Vec3Double, Vec3Int}
-<<<<<<< HEAD
```
dieknolle3333: @fm3 as I only had to merge imports I thought I could try that on my own. Could you have a look and make sure this makes sense? (also in `VolumeTracingService.scala`)

fm3: I added a commit changing one line in each of the two scala files :)

dieknolle3333: thanks for the quick reaction!

CHANGELOG.unreleased.md
```diff
@@ -15,7 +15,8 @@ For upgrade instructions, please check the [migration guide](MIGRATIONS.released
 - Added disabled drag handles to volume and skeleton layers for visual consistency. These layer cannot be dragged or reordered. [#7295](https://github.com/scalableminds/webknossos/pull/7295)
 - Dataset thumbnails for grayscale layers can now be colored using the value in the view configuration. [#7255](https://github.com/scalableminds/webknossos/pull/7255)
 - OpenID Connect authorization is now compatible with Providers that send the user information in an id_token. [#7294](https://github.com/scalableminds/webknossos/pull/7294)
-- Volume and bounding box information is shown in segments' context menus aswell as in a seperate modal in the segments tab. There is also an option to export the statistics. [#7249](https://github.com/scalableminds/webknossos/pull/7249)
+- Volume and bounding box information is shown in segments' context menus as well as in a seperate modal in the segments tab. There is also an option to export the statistics. [#7249](https://github.com/scalableminds/webknossos/pull/7249)
```
philippotto, line 18:
```suggestion
- Volume and bounding box information is shown in segments' context menus as well as in a separate modal in the segments tab. There is also an option to export the statistics. [#7249](https://github.com/scalableminds/webknossos/pull/7249)
```

frontend/javascripts/oxalis/view/right-border-tabs/segments_tab/segment_statistics_modal.tsx
```diff
@@ -41,19 +41,32 @@ type SegmentInfo = {
...
+          row.boundingBoxTopLeft,
+          row.boundingBoxPosition,
+        ]
+          .map(String) // convert every value to String
+          .map((v) => v.replaceAll('"', '""')) // escape double quotes
+          .map((v) => (v.includes(",") || v.includes('"') ? `"${v}"` : v)) // quote it if necessary
+          .join(","), // comma-separated
     )
-    .join("\n");
+    .join("\r\n"); // rows starting on new lines
```
philippotto, line 67:
```suggestion
    .join("\n"); // rows starting on new lines
```

I think, `\r` only interesting for windows users, but we haven't used that anywhere else yet. So, I wouldn't start now unless there is a strong reason for it.

frontend/javascripts/oxalis/view/right-border-tabs/segments_tab/segments_view.tsx
```diff
@@ -1534,8 +1535,10 @@ class SegmentsView extends React.Component<Props, State> {
     }
     const groupToSegmentsMap = createGroupToSegmentsMap(segments);
     const relevantGroupIds = getGroupByIdWithSubgroups(segmentGroups, groupId);
-    const segmentIdsNested = relevantGroupIds.map((groupId) => groupToSegmentsMap[groupId]);
-    return segmentIdsNested.flat();
+    const segmentIdsNested = relevantGroupIds
+      .map((groupId) => (groupToSegmentsMap[groupId] != null ? groupToSegmentsMap[groupId] : null))
+      .filter((x) => x !== null);
+    return segmentIdsNested.filter((x) => x).flat() as Segment[];
```
philippotto, line 1541: I think, one filter operation should suffice. I'd use `x != null` (not `!==`) as a predicate. This will cover null and undefined.

frontend/javascripts/oxalis/view/right-border-tabs/segments_tab/segments_view.tsx
```diff
@@ -1874,7 +1878,7 @@ class SegmentsView extends React.Component<Props, State> {
     if (visibleSegmentationLayer == null) return false;
     const relevantSegments =
       groupId != null ? this.getSegmentsOfGroupRecursively(groupId) : this.getSelectedSegments();
-    if (relevantSegments == null) return false;
+    if (relevantSegments == null || relevantSegments?.length === 0) return false;
```
philippotto, line 1881:
```suggestion
    if (relevantSegments == null || relevantSegments.length === 0) return false;
```

The second part of the expression is only evaluated when relevantSegments is not null.

frontend/javascripts/oxalis/view/context_menu.tsx
philippotto: yes, that's fine :+1:

frontend/javascripts/oxalis/view/right-border-tabs/segments_tab/segments_view.tsx
philippotto: cool, I didn't know about that :)

frontend/javascripts/oxalis/view/right-border-tabs/segments_tab/segment_statistics_modal.tsx
philippotto: I think, it's fine, but two things:

- please add a usage for `replaceAll` in `frontend/javascripts/libs/browser_feature_check.tsx` so that we can warn users about it
- in tsconfig you used es2022. however, you wrote above that es2021 would also suffice. I'd go with 2021 to keep things as backwards-compatible as possible for now (of course, it doesn't make a difference for the compilation output now, but other incompatibilities might sneak in later)

philippotto:
The CI currently fails. One error seems related to your code. The other might be due to the changed tsconfig. I think, it would be fine to add a `@ts-ignore` to that.

frontend/javascripts/oxalis/view/right-border-tabs/segments_tab/segment_statistics_modal.tsx
dieknolle3333: this was definitely by accident, maybe VS code added this or something, I'll delete it

philippotto (approved):
Awesome :) I only changed some small details. Also, I made a performance optimization which is actually independent of your changes, but it caught my eye during testing and it was easy to realize.

Will merge as a next step :ship:

PR #7251, opened 2023-08-07
Automatically assign color and channel name for omero ngff

### LINKED ISSUE #7224: Automatically assign colors for remote multi channel dataset
Multi channel datasets can be imported whereby every channel becomes a datalayer (#6578). When viewing the dataset, these are all white. When multiple layers are added this way, they should have different colors. For ngff multi-channel dataset, the omero/channels attribute may already include colors, which can be used here (eg. https://uk1s3.embassy.ebi.ac.uk/idr/zarr/v0.4/idr0101A/13457227.zarr)

## Context

frcroth:
- https://automaticallyassigncolor.webknossos.xyz/

### Steps to test:
- Explore https://uk1s3.embassy.ebi.ac.uk/idr/zarr/v0.4/idr0101A/13457227.zarr
- Note that the layer names are automatically assigned
- View the dataset
- Note that the layers already have colors.

### Issues:
- fixes #7224

------


app/models/binary/explore/NgffExplorer.scala
```diff
@@ -48,6 +52,7 @@ class NgffExplorer(implicit val ec: ExecutionContext) extends RemoteLayerExplore
                                        remotePath: VaultPath,
                                        credentialId: Option[String],
                                        channelCount: Int,
+                                       channelAttributes: Option[Seq[(Option[Color], Option[String])]] = None,
```
fm3, line 55: Could this be Option[ChannelAttributes] with a custom case class? I think we may get confused in the future this tuple structure here

webknossos-datastore/app/com/scalableminds/webknossos/datastore/datareaders/zarr/NgffMetadata.scala
```diff
@@ -107,3 +107,13 @@ object NgffLabelsGroup {
   implicit val jsonFormat: OFormat[NgffLabelsGroup] = Json.format[NgffLabelsGroup]
   val LABEL_PATH = "labels/.zattrs"
 }
+
+case class NgffOmeroObject(channels: List[NgffChannelAttributes])
```
fm3, line 111:
```suggestion
case class NgffOmeroMetadata(channels: List[NgffChannelAttributes])
```

I’d go for this name over Object if that’s alright with you :)

util/src/main/scala/com/scalableminds/util/image/Color.scala
```diff
@@ -21,6 +22,14 @@ object Color {
     )
   }
 
+  def fromHTML(htmlCode: String): Color = {
```
fm3, line 25: This should probably be wrapped in a tryo (possibly with toOption). The usage could then use the default if no valid color is parsed.

fm3 (changes requested):
Looking good already :-) I added a few small suggestions

PR #7254, opened 2023-08-08
Remove filters for user list

### LINKED ISSUE #5891: Replace Filters implementation by specific sql queries
In the wk backend, List queries for datasets and users fetch all accessible items from the db and then filter them according to filter criteria. This means that unnecessarily many objects are fetched fromt he DB and parsed as jvm objects. Instead, we should consider composing specific sql queries depending on the selected filters.

frcroth:
- https://___.webknossos.xyz

### Steps to test:
- Create TeamB (defaultTeam, sampleAdmin and sampleNoAdmin already exist)
- Create userA (team manager in teamB, team member in defaultTeam)
- Create userB (team member in defaultTeam, team member in teamB)
-----
- Execute the query (/users with sampleNoAdmin token) -> See only yourself
- Set isEditable = true -> []
- Set isEditable=false -> sampleNoAdmin
- Set sampleNoAdmin as Team manager of defaultTeam
- Execute query -> All 4 users
- Set isEditable = true -> All 4 users
- Set isEditable = false -> []
- Set isTeamManagerOrAdmin = true -> Everyone but userB
- Set isTeamManagerOrAdmin = false -> only userB
- Set isAdmin = true -> only sampleAdmin
- Set isAdmin = false -> everyone but sampleAdmin

I don't really understand what the first filter is supposed to do, since we only get to see users we can edit anyway? It can only be used to filter out the requesting user itself when they are not team manager. Doesn't seem useful.

Also, this is behavior that could be well tested, but I am not sure how to write a test for that.

### Issues:
- fixes #5891

------
- [ ] ~~Updated [changelog](../blob/master/CHANGELOG.unreleased.md#unreleased)~~


fm3:
The point about isEditable seems very valid to me. @philippotto I see that the frontend still has a getEditableUsers function in admin_rest_api.ts, but it seems to be used only in the tests (users.e2e.ts). Do you think we can remove this functionality?

app/models/user/User.scala
```diff
@@ -127,6 +127,52 @@ class UserDAO @Inject()(sqlClient: SqlClient)(implicit ec: ExecutionContext)
...
+  def buildSelectionPredicates(isEditableOpt: Option[Boolean],
+                               isTeamManagerOrAdminOpt: Option[Boolean],
+                               isAdminOpt: Option[Boolean],
+                               requestingUser: User)(implicit ctx: DBAccessContext): Fox[SqlToken] =
+    for {
+      accessQuery <- accessQueryFromAccessQ(listAccessQ)
+      editablePredicate = isEditableOpt match {
+        case Some(isEditable) =>
+          val usersInTeamsManagedByRequestingUser =
+            q"(SELECT _user FROM webknossos.user_team_roles WHERE _team IN (SELECT _team FROM webknossos.user_team_roles WHERE _user = ${requestingUser._id}  AND isteammanager=TRUE)))"
```
fm3, line 139:
```suggestion
            q"(SELECT _user FROM webknossos.user_team_roles WHERE _team IN (SELECT _team FROM webknossos.user_team_roles WHERE _user = ${requestingUser._id} AND isTeamManager)))"
```

fm3: (I think the `=TRUE` can be omitted).

Let’s stick with the capitalization used in schema.sql (e.g. isTeamManager, isAdmin). While postgres queries are case insensitive, I’d say that yields the best readability for us.

app/models/user/User.scala
```diff
@@ -127,6 +127,52 @@ class UserDAO @Inject()(sqlClient: SqlClient)(implicit ec: ExecutionContext)
...
+          }
+        case None => q"${true}"
+      }
+      isTeamManagerOrAdminPredicate = isTeamManagerOrAdminOpt match {
+        case Some(isTeamManagerOrAdmin) =>
+          val teamManagers = q"(SELECT _user FROM webknossos.user_team_roles WHERE isteammanager=true)"
+          if (isTeamManagerOrAdmin) {
+            q"_id IN $teamManagers OR isadmin=true"
+          } else {
+            q"_id NOT IN $teamManagers AND isadmin!=true"
```
fm3, line 153:
```suggestion
            q"_id NOT IN $teamManagers AND NOT isAdmin"
```

fm3 (commented):
Looking good :) I added a few cosmetic suggestions. Let’s also wait what Philipp says about the isEditable requirement. I’ll do some testing after that.

Concerning automated tests: I think some of this is already included in `users.e2e.ts`, which is part of the snapshot tests that run in the CI against the test db defined in `test/db/` and that can be adapted with `yarn refresh-e2e-snapshots`. If you think there is something important missing from there, feel free to add it.

fm3:
CI still failing shows that the behavior seems to differ from the previous one :thinking: But since this only pertains the isEditable users, maybe that problem will go away if we ditch that

philippotto:
> The point about isEditable seems very valid to me. @philippotto I see that the frontend still has a getEditableUsers function in admin_rest_api.ts, but it seems to be used only in the tests (users.e2e.ts). Do you think we can remove this functionality?

I don't really know what the original difference between these was, but `getEditableUsers` is used in several places? See:
- task_search_form.tsx
- user_list_view.tsx
- time_line_view.tsx
- edit_team_modal_view.tsx
- team_list_view.tsx

Maybe you only searched `.ts` files? If there is no difference between getEditableUsers and getUsers, we can change these of course.

By the way, the same logic exists for teams (`getEditableTeams` vs `getTeams`). Maybe the original intention of these has something to do with (super) admins?

fm3:
Ah, you are right, i only searched .ts files, my bad.

Hmm, I think the visibility rules may have changed over time.

In theory it makes sense to distinguish between users/teams the requester may see and those they can actually edit. But in practice, if you get to see the users page at all, you are definitely allowed to edit all of the users you can see (because only team managers+admins get to that page). So in theory we could unify this. But maybe it actually makes sense to keep it separate just to make it explicit that the requests are used differently. If a normal user should somehow land on the users page, they should see an empty list (as they are not allowed to edit anybody).

The same goes for teams, you get to see your teams even if you don’t get to edit (all of) them.

So I’d actually vote to keep it in

fm3 (approved):
:tada:

PR #7258, opened 2023-08-10
Add measurement tools

### LINKED ISSUE #3693: Add Measurement Tools
wK should offer some basic measurement tools for simple scientific analysis. Check out the below GIFs for inspiration. Simple click or drag the mouse cursor. Super simple version to begin with. Nothing fancy. No saving etc.

- Length Measurement ("Ruler")
- Area Measurement

Other Considerations:
- Include the measurements in the frontend API for scripting?
- Perhaps related to #3649

MichaelBuessemeyer:
- https://addmeasurementtool.webknossos.xyz/

### Steps to test:
- Open an annotation of any dataset
- Select the new measurement tool in the toolbar (the new tool should be the right most)
- Try measuring a path. Setting path point can be done via left click. Setting the last point via double-/or right-click and terminating the measurement by hitting escape or another right click.
- Try copying the measured length from the tooltip via the copy icon.
- Start measuring and terminating it while still in the process using escape. This should not lead to any errors or unexpected behaviour. (if so, I have a regression bug :see_no_evil:)
- Select the area measurement tool. Draw an area via right click. The area should always be closed by another highlighting line to highlight that the measured area is a closed area connecting the current end and start point.
- Upon releasing the left mouse drag, the highlighted line should be gone and the connecting line should be in the normal color.
- Potential bug: see below: When hitting escape while measuring an area, the area measurement tool is not correctly reset.
- Please report any unexpected behavior. 

- Test the tracing tool. It should still work as expected as this PR fiddled around with the helper geometry used by the tracing tool.

### TODOs / Bugs:
- ~~[ ] Add fill area for area measurement tool~~ We decided against that as this is quite the hurdle and thus went for always having a highlighted connection line between the start and end of the drawn area to highlight that the user is measuring an area.
- [x] When finished marking the area the tooltip can be moved around by left clicking.

### Known issue / to discuss:
- [ ] When canceling the area measurement via pressing escape while the mouse is still pressed down and then continuing dragging the mouse, the tooltip reappears and the measurement continues. Should this be tackled or is this misuse of the tool and the user can be expected to redo the measurement? In this scenario, no exception is thrown but the measured area is not correctly canceled upon hitting escape. So keeping it as it is right now, is acceptable imo.

### Issues:
- fixes #3693

------


MichaelBuessemeyer:
### Just for information purposes:
The algorithm used for the calculation of the area of the polygon of the area measurement tool works correctly, when the point are exactly on a line: The algorithm was taken from: https://www.mathopenref.com/coordpolygonarea2.html
```javascript
function polygonArea(X, Y, numPoints) 
{ 
area = 0;   // Accumulates area 
j = numPoints-1; 

for (i=0; i<numPoints; i++)
{ area +=  (X[j]+X[i]) * (Y[j]-Y[i]); 
  j = i;  //j is previous vertex to i
}
  return area/2;
}
// The triangle reaches from 1,1 to 4,1 to 1,3 and back to 1,1. Thus the area is h*w/2 = 2*3/2 = 3
X = [1,2,3,4,2.5,1,1]
Y = [1,1,1,1,2,3,2]
polygonArea(X,Y,7)
=> -3
// The result is correct (and negative due to the points being in counter clockwise order). 
// The implementation of the algorithm in this pr is secure against this flaw by using the absolute value of the calculated area.
```

frontend/javascripts/oxalis/view/input_catcher.tsx
```diff
@@ -105,6 +105,8 @@ const cursorForTool = {
   BOUNDING_BOX: "copy",
   QUICK_SELECT: "crosshair",
   PROOFREAD: "crosshair",
+  LINE_MEASUREMENT: "pointer",
+  AREA_MEASUREMENT: "pointer",
```
MichaelBuessemeyer, line 109: Can you think of a better suitable mouse pointer attribute for the new tools?

MichaelBuessemeyer:
@philippotto This pr should now be ready for the first review. After the CI is done, I will also ask for further UX feedback in slack.

frontend/javascripts/oxalis/controller/combinations/tool_controls.ts
```diff
@@ -780,6 +784,171 @@ export class QuickSelectTool {
   static onToolDeselected() {}
 }
 
+export class LineMeasurementTool {
+  static DOUBLE_CLICK_TIME_THRESHOLD = 600;
+  static getPlaneMouseControls(_planeId: OrthoView): any {
```
philippotto, line 789: `getPlaneMouseControls` of the area measurement tool doesn't accept that parameter. maybe make these interfaces consistent?

frontend/javascripts/oxalis/controller/combinations/tool_controls.ts
```diff
@@ -780,6 +784,171 @@ export class QuickSelectTool {
...
+      }
+    };
+    const leftClick = (pos: Point2, plane: OrthoView, event: MouseEvent) => {
+      // In case the tool was reset by the user, abort measuring.
+      if (lineMeasurementGeometry.wasReset && isMeasuring) {
+        isMeasuring = false;
+        return;
+      }
+      const currentTime = Date.now();
+      if (currentTime - lastLeftClickTime <= this.DOUBLE_CLICK_TIME_THRESHOLD) {
```
philippotto, line 838: wouldn't it make sense to check that the two clicks were registered roughly at the same spot?

frontend/javascripts/oxalis/controller/combinations/tool_controls.ts
```diff
@@ -780,6 +784,171 @@ export class QuickSelectTool {
...
+    return {
+      leftClick: "Left Click to measure distance",
+      rightClick: "Finish Measurement and reset",
+    };
+  }
+
+  static onToolDeselected() {
+    const { lineMeasurementGeometry } = getSceneController();
+    lineMeasurementGeometry.reset();
+    lineMeasurementGeometry.hide();
```
philippotto, line 879: There are several places where reset and hide is called. How about adding a `resetAndHide` helper method?

frontend/javascripts/oxalis/geometries/helper_geometries.ts
```diff
@@ -61,6 +107,43 @@ export class ContourGeometry {
     mesh.geometry.attributes.position.needsUpdate = true;
     mesh.geometry.setDrawRange(0, this.vertexBuffer.getLength());
     mesh.geometry.computeBoundingSphere();
+    this.connectingLine.geometry.attributes.position.needsUpdate = true;
+    // this.connectingLine.geometry.setDrawRange(0, 2);
```
philippotto, line 111: remove?

frontend/javascripts/oxalis/geometries/helper_geometries.ts
```diff
@@ -231,3 +314,113 @@ export class QuickSelectGeometry {
...
+    this.wasReset = false;
+    this.visible = true;
+    this.currentOrthoView = initialOrthoView;
+    this.vertexBuffer.push(pos);
+    // Adding an additional point that will be modified by setTopPoint.
+    this.vertexBuffer.push(pos);
+    this.finalizeMesh();
+  }
+
+  setTopPoint(pos: Vector3) {
```
philippotto, line 369: I think I don't really understand what a "top point" is. maybe add a comment?

frontend/javascripts/oxalis/geometries/helper_geometries.ts
```diff
@@ -231,3 +314,113 @@ export class QuickSelectGeometry {
     rectangle.material.alphaMap = null;
   }
 }
+
+export class LineMeasurementGeometry {
+  color: THREE.Color;
```
philippotto, line 319: maybe add a class comment explaining that this is used for displaying "line segments" (not sure whether this is the right term?).

frontend/javascripts/oxalis/geometries/helper_geometries.ts
```diff
@@ -231,3 +314,113 @@ export class QuickSelectGeometry {
...
+
+  reset() {
+    this.line.material.color = this.color;
+    this.vertexBuffer.clear();
+    this.finalizeMesh();
+    this.wasReset = true;
+  }
+
+  setStartPoint(pos: Vector3, initialOrthoView: OrthoView) {
+    this.wasReset = false;
```
philippotto, line 360: I'd suggest to either add a comment explaining that this method expects an empty vertexBuffer or add this.vertexBuffer.clear() to this method.

philippotto (commented):
Great stuff :) Will test now!

philippotto:
Overall it feels pretty good :+1: My testing results:
- line tool: fast clicking is sometimes registered as double click (see my other suggestion to compare the clicked positions)
- area tool shows units without ² and the line tool shows them with ² :trollface: 
- the measured area seems wrong (see my other code comment)
- I'd change "Finish Measurement and reset" in the status bar to "Finish Measurement" since it's not clear what reset actually means

-----------------------

Please remind me again why the measurement geometries are hidden as soon as the user moves? Is it because it would be hard to stick the tooltip to the geometry? If this is the case, one could also think about moving the tooltip content simply into the toolbar :thinking: But I think we'd need actual user feedback for that. Right now, I imagine it to be confusing potentially (could be percepted as bug, too).
@normanrz do you have an opinion about the second paragraph?

frontend/javascripts/oxalis/geometries/helper_geometries.ts
MichaelBuessemeyer: I called it `connected line segments` :shrug:

frontend/javascripts/oxalis/geometries/helper_geometries.ts
MichaelBuessemeyer: Good catch :+1:, I went for the 2nd suggestion, as this is more bullet proof against potential bugs imo.

frontend/javascripts/oxalis/geometries/helper_geometries.ts
MichaelBuessemeyer: I added a comment and renamed the method to `updateLatestPointPosition`.

By top point I meant the latest added point to the array of points (like the latest element on a stack is called top element).

MichaelBuessemeyer (commented):
Thanks for your feedback @philippotto: I think I addressed everything now.

> Please remind me again why the measurement geometries are hidden as soon as the user moves?


you mentioned hiding the tool as soon as the camera moves [here](https://scm.slack.com/archives/C5AKLAV0B/p1692279400018379?thread_ts=1692279123.547119&cid=C5AKLAV0B) and [here](https://scm.slack.com/archives/C5AKLAV0B/p1692279604189899?thread_ts=1692279123.547119&cid=C5AKLAV0B)

Let's wait for @normanrz answer :smiley_cat:

normanrz:
> Please remind me again why the measurement geometries are hidden as soon as the user moves? Is it because it would be hard to stick the tooltip to the geometry? If this is the case, one could also think about moving the tooltip content simply into the toolbar 🤔 But I think we'd need actual user feedback for that. Right now, I imagine it to be confusing potentially (could be percepted as bug, too).
> @normanrz do you have an opinion about the second paragraph?

I think keeping the lines on move would be nice. But how would a user remove them again?

Also, Alt+Mouse Move doesn't work for me in this tool mode. That feels like a bug.

MichaelBuessemeyer:
> I think keeping the lines on move would be nice. 

I misunderstood your statement a little :sweat_smile: Now the lines are kept (unless the user moves in the z direction of the viewport) and the tooltip follows the line / area mesurement

MichaelBuessemeyer:
> But how would a user remove them again?

Via right or double left click.

> Also, Alt+Mouse Move doesn't work for me in this tool mode. That feels like a bug.

For me this works. Maybe the viewports weren't the active part of the wk webpage when you tested? :thinking:

philippotto:
The sticky tooltip is quite cool :+1: However, I ran into some bugs:

- like @normanrz, alt+mousemove doesn't work for me while using the line tool
- alt+mousemove for the area tool works sort of. it changes to the proper tool and does the moving, but the tool still registers new points around the mouse.
- I got into a weird state where the measurement tools got broken. I can reproduce this by starting measuring in XY and then adding points in YZ. see video.

minor problems:
- when moving the mouse fast, I can move it into the tooltip which is a bit weird since I cannot pinpoint the next click in the area of the tooltip. not sure, how to solve this. as a user I can always leave the tooltip and then hidden space is visible again. so, it's not a big issue, but I ran into this quite quickly, which is why I'm writing this.

I recorded a video for these issues:
- the video starts with the broken state where I cannot measure anything. I recover this be reloading the page
- then I show that alt+mousemove doesnt work for the line tool
- then I show that the mouse can get caught up in the tooltip
- then I show alt+mousemove for the area tool
- in the end, I try to reproduce the weird state again. it was a bit difficult, but at 1:40 I managed to break it. afterwards, I cannot fix it again except for reloading the page. I would expect that switching the tools, resets all the relevant internal state.

https://github.com/scalableminds/webknossos/assets/2486553/74e25107-9453-46db-a4d2-f571f2e3765a

MichaelBuessemeyer (commented):
> like @normanrz, alt+mousemove doesn't work for me while using the line tool

Ahh, ok now I understand what you mean: It is impossible to move around while measuring with a line :+1:. Previously I only tested moving with alt when I was not actively measuring a line.
The alt moving should now be fixed for both measurement tools during the measurement and also while not measuring.


> alt+mousemove for the area tool works sort of. it changes to the proper tool and does the moving, but the tool still registers new points around the mouse.

This should be fixed now. The tool should register new points while alt is pressed.

> I got into a weird state where the measurement tools got broken. I can reproduce this by starting measuring in XY and then adding points in YZ. see video.

I think this was due to an instant reset of the measurement caused by the  `distance_measurement_tooltip`. See my comment :)

>  when moving the mouse fast, I can move it into the tooltip which is a bit weird since I cannot pinpoint the next click in the area of the tooltip. not sure, how to solve this. as a user I can always leave the tooltip and then hidden space is visible again. so, it's not a big issue, but I ran into this quite quickly, which is why I'm writing this.


While measuring, the tooltip now ignores all pointer event, which mitigates this problem as the inputcatcher get the pointer event, even when the mouse hits the tooltip. Thanks to @fm3 for the solution. 
This is done by saving whether the measurement tool is measuring at the moment in the store.



I added a tooltip to the distance option for the context menu as the entry could be misunderstood. See for reference https://scm.slack.com/archives/C5AKLAV0B/p1695894445554669

philippotto (approved):
awesome! works very well :)

there are several PR comments where the code is marked as outdated. I assume these comments can be resolved? other than that, only see my one comment about a magic number.

PR #7259, opened 2023-08-11
Move swagger route

normanrz:
https://webknossos.org/swagger.json doesn't seem to work anymore (404). Moving it should resolve that issue. I also updated the wklibs in https://github.com/scalableminds/webknossos-libs/pull/938

### DIFF
--- a/conf/routes
+++ b/conf/routes
@@ -2,5 +2,5 @@
 -> /data/                        com.scalableminds.webknossos.datastore.Routes
 -> /tracings/                    com.scalableminds.webknossos.tracingstore.Routes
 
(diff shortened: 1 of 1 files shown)


fm3:
This should be fixed in #7220 [(single-line change in the request handler)](https://github.com/scalableminds/webknossos/pull/7220/files#diff-2784947069515106921e29b0927d9afab7f3024f2953e39f0df4e49ee3060c98R33)

No need to move the route. (Do you think it should be moved anyway, is it better placed in the .well-known path?)

normanrz:
No need to change, if it works.

Issue #7256, opened 2023-08-09
Sometimes volume buckets are not properly saved


philippotto:
## Context
Volume annotation. https://scm.slack.com/archives/C02H5T8Q08P/p1691565143708429

## Current Behavior
Sometimes, volume buckets are missing after reloading the page. Notably, reports mainly showed that buckets in mag > 1 are missing. So, it might be related to downsampling (test what happens if it fails for some buckets).

## Steps to Reproduce the bug
Unfortunately, we don't know how to reproduce this (yet). Might be related to race conditions or network instability. We could also test what happens when the user closes the tab forcefully during save.

## Your Environment for bug

- Browser name and version: e.g. Chrome 39
- Operating System and version: e.g. Windows 10
- Version of WEBKNOSSOS (Release or Commit):
- [ ] Specific to long-running jobs (set `jobsEnabled=true` in `application.conf`)
- [ ] Specific to webknossos.org (set `isDemoInstance=true` in `application.conf`)

philippotto:
My hunch is that this is a front-end bug, but I might be wrong about it.

Issue #6495, opened 2022-09-26
Versions of volume data should be created at "transaction" points


philippotto:
Currently, editing volume data will create new versions where each version consists of at most 32 changed buckets (see `COMPRESSING_BATCH_SIZE`). However, this means that reverting to an older version might very well restore a state where chunk borders are visible since the version was created right in the middle of saving.
Ideally, the saving mechanism should use a transaction logic (similar to how it's done for skeletons) so that each version is a consistent tracing state.

An additional benefit of ditching the 32 limit is that we reduce the amount of versions being created (one large brush stroke can already cover much more than 32 buckets).

/edit: Another problem caused by this is that version conflicts (e.g., user B saved something while user A was saving larger changes) will result in missing data.

philippotto:
Also see https://www.notion.so/scalableminds/Design-Doc-Versioning-Transactions-90f514760a87491699fffa3fe2ba602c#f39d5acdc22e4035aa96d6006b4528d4.

philippotto:
@fm3 Are the `transaction{Ids,groupIndex,groupCount}` persisted or is it only ephemeral to ensure that no "half transactions" are stored? @daniel-wer and I noticed that the version history view doesn't make use of the transaction concept currently. If the properties are stored, we could use them to show semantic versions.

This would only help for skeleton tracings, though, since the front-end doesn't really make use of transaction groups for volume tracings.

philippotto:
Just for the record: We decided to establish a 1:1 relation between transactions and versions (see #7290). That way, we don't need to store additional properties, such as the `transactionGroupIndex`.

PR #7264, opened 2023-08-14
Transactional saving of volume annotations

philippotto:
### Description:
- Introduces sensible transaction points for volume annotations (--> also means less versions which will be helpful in the version restore view :tada:). This is done by waiting for small time windows (1 second) in which no volume data was mutated **and** no dirty volume buckets need to be downloaded. This is similar to how the PushQueue operated, anyway, but it chunked buckets into 32 batches later which destroyed any transactional semantics. Reusing the state machine of the undo mechanism would be nicer, since it is strictly user-interaction based. However, the implementation of that would be far more complicated.
- Improved speed of bucket compression by using larger payloads for the webworkers (an isolated benchmark with reasonable inputs showed that the bucket compression is now 5 to 10 times faster).
- Removes AsyncTaskQueue as its sequentiality and failure-mechanism aren't needed anymore. [This abstraction was originally introduced to deal with failing network requests.](https://github.com/scalableminds/webknossos/blob/65319c4de0bb0193af4a47fc1ec746435395e9f9/app/assets/javascripts/oxalis/model/binary/layers/wk_layer.js#L96) However, that part is not handled by the AsyncTaskQueue anymore.
- The error handling in the PushQueue was hardened and unified with the error handling in the sagas (by escalating potential exceptions to the sagas).
- Fixes a small bug in the save-saga which could lead to an accumulating backlog of save queue items that would grow faster than they would be saved. This could happen when brushing a high volume of data and only relying on the auto-save mechanism. That mechanism did only send 1 request every 30 seconds (even when multiple requests would be necessary to completely save the state of that time).
- The save button now also warns when the PushQueue is starving (i.e., constantly unable to bundle up a transaction, because a bucket was edited again). Previously, this was achieved by providing a max_time parameter to _.debounce. However, since we want sensible transactions, this is not an option anymore.
- Move some modules from libs to libs/async.
- Lower saving batch constants for volume tracings, because the backend needs more time to process the update actions (see discussion [here](https://scm.slack.com/archives/C5AKLAV0B/p1692177165085429))
- Improve progress indicator during saving (the tooltip for the save button will also show when buckets need to be downloaded during saving, because this can take a noticable amount of time).

- https://___.webknossos.xyz

### Steps to test:
- do a volume annotation (with and without fallback)
- set the network speed to very slow
- reload the tab during save (large volume changes are necessary to cause multiple requests)
- ensure that the volume data looks consistent (zoom in and out to check all mags)
- additionally/alternatively:
  - brush a large area
  - save
  - look at the version restore view. there should be one version (maybe other versions for the new segment id and for updateTracing, but not more)
  - compare the above to master.webknossos.xyz where lots of versions will be created. previewing one version and zooming to different mags will show hard-cuts in the brushed data.

### TODOs:
- [X] test volume speed on dev instance (which will use prod mode) to ensure that it didn't regress
- [X] verify that new save requests don't get too large. since updateBucket actions are way larger than skeleton update actions, some tuning might be necessary here. <-- no problem, since the compression does a good job of keeping the size small

### Issues:
- hopefully fixes [#7256](https://github.com/scalableminds/webknossos/issues/7256)
- contributes to #6495
- to close the issue fully, the version history view should show the transactions (the back-end needs to store these first, though)

------


frontend/javascripts/oxalis/model/sagas/save_saga_constants.ts
```diff
@@ -11,5 +11,5 @@ export const UNDO_HISTORY_SIZE = 20;
 export const SETTINGS_RETRY_DELAY = 15 * 1000;
 export const SETTINGS_MAX_RETRY_COUNT = 20; // 20 * 15s == 5m
 
-export const maximumActionCountPerBatch = 5000;
-export const maximumActionCountPerSave = 15000;
+export const maximumActionCountPerBatch = 20;
```
philippotto, line 14: todo: revert

frontend/javascripts/oxalis/model/sagas/save_saga_constants.ts
```diff
@@ -11,5 +11,5 @@ export const UNDO_HISTORY_SIZE = 20;
 export const SETTINGS_RETRY_DELAY = 15 * 1000;
 export const SETTINGS_MAX_RETRY_COUNT = 20; // 20 * 15s == 5m
 
-export const maximumActionCountPerBatch = 5000;
-export const maximumActionCountPerSave = 15000;
+export const maximumActionCountPerBatch = 5000; //20; // 5000
+export const maximumActionCountPerSave = 15000; // 60; // 15000
```
philippotto, line 15:
```suggestion
export const maximumActionCountPerBatch = 5000;
export const maximumActionCountPerSave = 15000;
```

frontend/javascripts/oxalis/model/sagas/save_saga.ts
```diff
@@ -151,9 +159,19 @@ function getRetryWaitTime(retryCount: number) {
...
+): Saga<number> {
+  /*
+   * Saves a reasonably-sized part of the save queue (that corresponds to the
+   * tracingId) to the server (plus retry-mechanism).
+   * The saga returns the number of save queue items that were saved.
+   */
+
   const fullSaveQueue = yield* select((state) => selectQueue(state, saveQueueType, tracingId));
   const saveQueue = sliceAppropriateBatchCount(fullSaveQueue);
+  console.log("saving", saveQueue.length, "items out of", fullSaveQueue.length);
```
philippotto, line 174:
```suggestion
```

frontend/javascripts/oxalis/model/bucket_data_handling/data_cube.ts
```diff
@@ -679,6 +679,7 @@ class DataCube {
   }
 
   triggerPushQueue() {
+    console.log("triggerPushQueue");
```
philippotto, line 682:
```suggestion
```

frontend/javascripts/oxalis/model/bucket_data_handling/pushqueue.ts
```diff
@@ -77,42 +97,75 @@ class PushQueue {
...
 
-  pushImpl = async () => {
-    await this.cube.temporalBucketManager.getAllLoadedPromise();
-
-    if (!this.sendData) {
-      return;
+  pushImpl = function* (this: PushQueue) {
+    try {
+      console.log("throw in pushimpl");
+      console.log("pushImpl start");
```
philippotto, line 103:
```suggestion
```

frontend/javascripts/oxalis/model/bucket_data_handling/pushqueue.ts
```diff
@@ -77,42 +97,75 @@ class PushQueue {
...
+      this.flushAndSnapshot();
+    } catch (error) {
+      // The above code is critical for saving volume data. Because the
+      // code is invoked asynchronously, there won't be a default error
+      // handling in case it crashes due to a bug.
+      // Therefore, escalate the error manually so that the sagas will crash
+      // (notifying the user and stopping further potentially undefined behavior).
+      Store.dispatch(escalateErrorAction(error));
     }
+    console.log("pushImpl end");
```
philippotto, line 120:
```suggestion
```

frontend/javascripts/oxalis/model/bucket_data_handling/pushqueue.ts
```diff
@@ -77,42 +97,75 @@ class PushQueue {
...
 
-    while (this.pendingQueue.size) {
-      let batchSize = Math.min(COMPRESSING_BATCH_SIZE, this.pendingQueue.size);
-      const batch: DataBucket[] = [];
+  private flushAndSnapshot() {
+    this.waitTimeStartTimeStamp = null;
+    // Flush pendingQueue. Note that it's important to do this synchronously.
+    // If other actors could add to queue concurrently, the front-end could
+    // send an inconsistent state for a transaction.
+    console.log("Flush pending queue with size:", this.pendingQueue.size);
```
philippotto, line 128:
```suggestion
```

frontend/javascripts/oxalis/model/bucket_data_handling/pushqueue.ts
```diff
@@ -1,74 +1,91 @@
...
-  constructor(cube: DataCube, sendData: boolean = true) {
+
+  // The pendingQueue contains all buckets that should be:
+  // - snapshotted,
+  // - put into one transaction and then
+  // - saved
+  // That queue is flushed in a debounced manner so that the time of the
+  // snapshot should be suitable for a transaction (since neither WK nor the
+  // user edited the buckets in a certain time window).
+  private pendingQueue: Set<DataBucket>;
```
normanrz, line 29: Why is the queue a `Set`?

normanrz (commented):
This code is a piece of art! Thanks for writing all the descriptive comments. I think I understand what is roughly going on and look forward to the walkthrough.

frontend/javascripts/oxalis/model/bucket_data_handling/pushqueue.ts
philippotto: The order of the buckets doesn't matter within a transaction, which is why this is a set. However, the mismatch is obviously still an eye sore. I renamed it to `pendingBuckets` now. The class `PushQueue` might also be ripe for a new name, but then the `PullQueue` loses its "name-sibling" :sweat_smile: Also, I'm not sure how to call it. `BucketTransactionBundler`? Might be okay to keep the PushQueue as is.

normanrz (approved):

PR #7267, opened 2023-08-16
Fetch server public key during OIDC auth

fm3:
OpenID Connect authorization now fetches the server’s public key automatically. The config keys `singleSignOn.openIdConnect.publicKey` and `singleSignOn.openIdConnect.publicKeyAlgorithm` are now unused. [7267](https://github.com/scalableminds/webknossos/pull/7267)

### Steps to test:
- Set up OIDC as described in https://github.com/scalableminds/webknossos/pull/6534#issuecomment-1285177420 and #7263
- Should still work now even with publicKey not set in the config.
- I also tested that if a different publicKey is used, the signed token is not accepted (I added `.add(BigInteger.ONE)` to the parsed publicKey modulus

------


normanrz (approved):
Cool stuff!

PR #7284, opened 2023-08-17
Add banner for scheduled maintenance

### LINKED ISSUE #7257: Maintenance banner
## Detailed Description
To inform users about upcoming or ongoing maintenance work, it would be great to add a maintenance banner to Webknossos. This would also be interesting for our dedicated instances that may have maintenance work on their local datastores.
I would imagine a simple banner above the top navigation bar that has some configurable text (via application.conf). We can discuss whether the banner should be user-dismissible or not.

fm3:
- https://___.webknossos.xyz

### Steps to test:
Backend:
- Create new maintenance with `curl -X POST 'http://localhost:9000/api/maintenances' -H 'X-Auth-Token: secretSampleUserToken' -H "Content-Type: application/json" -d '{"startTime":1692272319000, "endTime":1692272320000, "message":"a message!"}'` (adapt timestamps to something in the future or current, use https://www.epochconverter.com/ for millisecond unix timestamps)
- delete and update routes, as well as readOne and listAll are also only available for super users
  - `curl -X DELETE 'http://localhost:9000/api/maintenances/64de02abee0100c103ab66b6' -H 'X-Auth-Token: secretSampleUserToken' -H` (replace id)
  - `curl -X PUT 'http://localhost:9000/api/maintenances/64de0390ee0100a006ab66cc' -H 'X-Auth-Token: secretSampleUserToken' -H "Content-Type: application/json" -d '{"startTime":1692272319000,  "message":"a message!!!"}'`
- You should see maintenances at `GET api/maintenances/listCurrentAndUpcoming` (this is what the frontend uses) 
Frontend:
- open wk without any maintenances -> should look as always
- create an upcoming maintenance -> wk should show this in a banner. the banner should be closable. a page refresh should not show that banner anymore.
- create a current maintennance -> wk should show this banner, but it should not be closable.
- scrolling with the banner should behave intuitively.
- in the annotation view, touching the banner with the mouse should move the banner to the bottom of the page (and vice versa)

### TODOs:
- [x] Backend
  - [x] schema
  - [x] clarify permissions
  - [x] CRUD routes
- [x] Frontend
  - [x] Adapt to new format of maintenance rest api route (`GET api/maintenances/listCurrentAndUpcoming`)
  - [x] Banner for closest upcoming maintenance “Maintenance starting in 3h” with tooltip showing more details
  - [x] Banner for current maintenance “Current Maintenance” with tooltip showing more details

### Issues:
- fixes #7257

------


fm3:
@normanrz I am unsure about the permission system here. Who should be allowed to create/update scheduled maintenances in the database?

 - Only superusers? Then we would have to promote the users who want to use this feature to superuser on their wk instances
 - All organization admins? But maintenances are not currently scoped to organizations, so that would give a lot of users on webknossos.org this permission
 - Or do we add an entry in the config to enable/disable this for “normal admins”?

normanrz:
Either we do it just for superadmins, which seems fine to me, or we scope the maintenance banners to orgs.

dieknolle3333:
I have some more questions :)
- If there is a maintenance currently running, I'm assuming that only the "Current maintenance" banner should be shown? Or should the next upcoming maintenance still be displayed?
- I am assuming that "more details" are the duration and the maintenance's message?

normanrz:
> * If there is a maintenance currently running, I'm assuming that only the "Current maintenance" banner should be shown? Or should the next upcoming maintenance still be displayed?

Yes, either the current or the next maintenance should be shown. No more than 1 banner.

> * I am assuming that "more details" are the duration and the maintenance's message?

I think the message should be kept short to fit in the banner.

dieknolle3333:
this should be a basic frontend implementation without storing the maintenance info in the redux store (moved to new branch). 
some useful timestamps: 1696024800000 is 30.09.23 00:00; 1695938400000 is 29.09.23 00:00; 1693778400000 is 4.9.23 00:00;

dieknolle3333:
@fm3 could you help me with the merge conflicts in `MIGRATIONS.unreleased.md`? I'm not sure how to resolve them

frontend/javascripts/maintenance_banner.tsx
```diff
@@ -0,0 +1,106 @@
...
+  const saveUserClosedMaintenanceInfo = (closestUpcomingMaintenance: MaintenanceInfo) => {
+    if (activeUser == null) return;
+    const [nextMaintenanceAcknowledged] = updateNovelUserExperienceInfos(activeUser, {
+      latestAcknowledgedMaintenanceInfo: closestUpcomingMaintenance.id,
+    });
+    Store.dispatch(setActiveUserAction(nextMaintenanceAcknowledged));
+  };
+
+  const toggleTopOrBottomPosition = () => {
+    setPosition(isTop ? { top: "48px" } : { bottom: "20px" });
```
dieknolle3333, line 40: TODO: unify style to write position

frontend/stylesheets/main.less
```diff
@@ -110,6 +110,14 @@ body {
...
 }
 
+.ant-alert-message {
+  margin-left: 5px;
+}
+.ant-alert-banner {
+  position: absolute;
+  z-index: 1000;
+  width: 100%;
+}
```
philippotto, line 120: These are antd classes, right? Please move these (and the rule in line 121f) to `antd_overwrites.less`.

frontend/javascripts/admin/admin_rest_api.ts
```diff
@@ -2072,10 +2072,31 @@ export function getExistingExperienceDomains(): Promise<ExperienceDomainList> {
...
 export async function isInMaintenance(): Promise<boolean> {
-  const info = await Request.receiveJSON("/api/maintenance", {
-    doNotInvestigate: true,
-  });
-  return info.isMaintenance;
+  const allMaintenances: Array<MaintenanceInfo> = await Request.receiveJSON(
+    "/api/maintenances/listCurrentAndUpcoming",
+  );
+  const currentEpoch = Date.now();
+  const currentMaintenance = allMaintenances?.find(
```
philippotto, line 2079: why the `?`? according to the type in line 2075, it's always an array.

frontend/javascripts/admin/admin_rest_api.ts
```diff
@@ -2072,10 +2072,31 @@ export function getExistingExperienceDomains(): Promise<ExperienceDomainList> {
...
+  );
+  return currentMaintenance != null;
+}
+
+export type MaintenanceInfo = {
+  startTime: number;
+  endTime: number;
+  id: string;
+  message: string;
+};
```
philippotto, line 2090: can be moved to api_flow_types.ts

frontend/javascripts/admin/admin_rest_api.ts
```diff
@@ -2072,10 +2072,31 @@ export function getExistingExperienceDomains(): Promise<ExperienceDomainList> {
...
+
+export async function listCurrentAndUpcomingMaintenances(): Promise<Array<MaintenanceInfo>> {
+  const allMaintenances: Array<MaintenanceInfo> = await Request.receiveJSON(
+    "/api/maintenances/listCurrentAndUpcoming",
+  );
+  const currentEpoch = Date.now();
+  return allMaintenances
+    ?.filter((maintenance) => maintenance.endTime > currentEpoch)
```
philippotto, line 2098: 1) see my other comment regarding the `?`
2) why is this filter necessary? shouldn't the backend only return current and upcoming entries (according to the name), anyway?

frontend/javascripts/maintenance_banner.tsx
```diff
@@ -0,0 +1,106 @@
...
+import { setActiveUserAction } from "oxalis/model/actions/user_actions";
+import { Store } from "oxalis/singletons";
+import React, { useState } from "react";
+
+export function MaintenanceBanner() {
+  const { activeUser } = Store.getState();
+  const [currentAndUpcomingMaintenances, setcurrentAndUpcomingMaintenances] = useState<
+    Array<MaintenanceInfo>
+  >([]);
+  const [position, setPosition] = useState<Object>({ top: 48 });
```
philippotto, line 18: please import `navbarHeight` from navbar.tsx and use it in this module.

frontend/javascripts/maintenance_banner.tsx
```diff
@@ -0,0 +1,106 @@
...
+  const [isTop, setIsTop] = useState(true);
+  useInterval(
+    async () => {
+      if (activeUser == null) return;
+      setcurrentAndUpcomingMaintenances(await listCurrentAndUpcomingMaintenances());
+    },
+    10000, //TODO find value for production
```
philippotto, line 25: I think, once a minute would be okay. should be extracted into a top-level constant then.

frontend/javascripts/maintenance_banner.tsx
```diff
@@ -0,0 +1,106 @@
...
+export function MaintenanceBanner() {
+  const { activeUser } = Store.getState();
+  const [currentAndUpcomingMaintenances, setcurrentAndUpcomingMaintenances] = useState<
+    Array<MaintenanceInfo>
+  >([]);
+  const [position, setPosition] = useState<Object>({ top: 48 });
+  const [isTop, setIsTop] = useState(true);
+  useInterval(
+    async () => {
+      if (activeUser == null) return;
```
philippotto, line 22: wouldn't maintenances also be relevant for logged out users that view public datasets? you probably did that because you cannot store the `latestAcknowledgedMaintenanceInfo` for logged-out users, right? however, for running maintenances, this property isn't used anyway. I'd suggest to show running maintenances for all users and upcoming maintenances only for logged-in users.

frontend/javascripts/maintenance_banner.tsx
```diff
@@ -0,0 +1,106 @@
...
+import { Alert } from "antd";
+import FormattedDate from "components/formatted_date";
+import { useInterval } from "libs/react_helpers";
+import { setActiveUserAction } from "oxalis/model/actions/user_actions";
+import { Store } from "oxalis/singletons";
+import React, { useState } from "react";
+
+export function MaintenanceBanner() {
+  const { activeUser } = Store.getState();
+  const [currentAndUpcomingMaintenances, setcurrentAndUpcomingMaintenances] = useState<
```
philippotto, line 15: should be `setCurrentAndUpcomingMaintenances`

frontend/javascripts/maintenance_banner.tsx
philippotto: 1) it's not necessary to pass a string with px. you can just pass a number
2) use variables for both values

frontend/javascripts/maintenance_banner.tsx
```diff
@@ -0,0 +1,106 @@
...
+  const toggleTopOrBottomPosition = () => {
+    setPosition(isTop ? { top: "48px" } : { bottom: "20px" });
+    setIsTop(!isTop);
+  };
+
+  const getClosestUpcomingMaintenanceBanner = () => {
+    const currentTime = Date.now();
+    const closestUpcomingMaintenance = currentAndUpcomingMaintenances
+      ?.filter((maintenance) => maintenance.startTime > currentTime)
+      .sort((a, b) => a.startTime - b.startTime)[0];
```
philippotto, line 48: isn't this redundant to the rest admin module? at least the sorting. I understand, that a running maintenance can become finished while this component is updating.

frontend/javascripts/maintenance_banner.tsx
```diff
@@ -0,0 +1,106 @@
...
+    //if (closestUpcomingMaintenance == null || activeUsersLatestAcknowledgedMaintenance === closestUpcomingMaintenance.id) return; //TODO use this after testing and remove next 3 lines
+    if (closestUpcomingMaintenance == null) return;
+    if (activeUsersLatestAcknowledgedMaintenance === closestUpcomingMaintenance.id)
+      return <Alert message="Latest maintenance already acknowledged" type="success" />;
```
philippotto, line 52: this is probably only used for testing?

frontend/javascripts/maintenance_banner.tsx
```diff
@@ -0,0 +1,106 @@
...
+        banner
+        onMouseEnter={() => {
+          toggleTopOrBottomPosition();
+        }}
+        style={position}
+      />
+    );
+  };
+
+  if (currentAndUpcomingMaintenances?.length === 0 || currentAndUpcomingMaintenances == null)
```
philippotto, line 98: same with the `?`

philippotto (commented):
code looks quite good already, but I still have some feedback :)

frontend/stylesheets/main.less
dieknolle3333: okay! when does something count as overwriting? 
there are multiple CSS rules for antd in main.less, e.g. in l. 72 and I am not sure I understand when to write antd rules where

frontend/javascripts/admin/admin_rest_api.ts
dieknolle3333: oh no, that slipped through!

frontend/javascripts/maintenance_banner.tsx
dieknolle3333: actually I just thought that this info wouldnt be important for logged-out users, but what I didnt have in mind is that they could actively be viewing datasets, so I think you're right that this info matters for everyone

frontend/javascripts/maintenance_banner.tsx
dieknolle3333: yes, hence the comment :) can I already delete this?

frontend/javascripts/maintenance_banner.tsx
dieknolle3333: my thought was that I didn't want to depend on the rest admin module returning a sorted array (maybe a refactoring might change this), so I'll just do it again. But in the admin rest API I thought that it would be nice to return the maintenances sorted. Then I think I'll just remove the sorting from the admin rest module

frontend/javascripts/maintenance_banner.tsx
philippotto: yes, I'd say so :)

philippotto:
Nice! Works well :+1: However, there's an oddity about the banner when the user scrolls in the dashboard. See these screenshots:

image

image


I see two possible solutions:
- use sticky instead of absolute so that the scrolling doesn't interfere
- don't use the absolute positioning, but normal positioning (as it was in an earlier iteration) for all pages except for the tracing layout view. the "switch position to the bottom of the screen" would only be needed for the tracing layout view.

The first option is probably easier. I hope that it behaves well then. If not, the second option might be the better approach (since the layout in the tracing view should forbid scrolling anyway).

dieknolle3333:
with the first option I faced the problem that the lower position wasn't taken into account, so the switching didnt work any more. thus I implemented the second version. this leads to some views (at least Statistics > Overview) scrolling, not sure whether that is a problem. other than that it seems to be working fine!

frontend/stylesheets/main.less
philippotto: You are right, it's not very clear right now. I moved another rule, but the others rules mostly use antd class names in combination with other descriptors (e.g., `#main-container > section > .ant-layout-header`). This means that only specific occurrences of antd components are changed. In the overwrites module, on the other hand, the rules are mutating all occurrences of the mentioned antd components. So, this is can be used as a rule of thumb :)

frontend/javascripts/maintenance_banner.tsx
```diff
@@ -4,24 +4,41 @@ import {
...
 import React, { useState } from "react";
+import { useSelector } from "react-redux";
 import { MaintenanceInfo } from "types/api_flow_types";
 
+const INITIAL_DELAY = 5000;
 const INTERVAL_TO_FETCH_MAINTENANCES_MS = 60000;
 
 export function MaintenanceBanner() {
-  const { activeUser, uiInformation } = Store.getState();
+  const activeUser = useSelector((state: OxalisState) => state.activeUser);
```
philippotto, line 21: @dieknolle3333 just fyi, I fixed the store access here. When rendering react components, it's always important to properly hook up with the store so that the component updates when the store updates. This listener can be set up with `connect` (from `react-redux`) or in case of functional components `useSelector` (which is better in my opinion since it's less verbose). Your access to `Store.getState()` is fine in event handlers (e.g., onClick handler in a button) when one doesn't need the synchronization. However, during rendering this would mean that the component gets outdated at some point.

philippotto (approved):
Great :) I fixed some smaller stuff and now think that the PR should be ready to go :rocket:

frontend/javascripts/maintenance_banner.tsx
dieknolle3333: oh, I understand!

Issue #7116, opened 2023-06-02
Support all valid values for transpose codec configuration in zarr v3


frcroth:
The codec spec (https://zarr-specs.readthedocs.io/en/latest/v3/codecs/transpose/v1.0.html) states that the "order" configuration key must be one of:

>     An array of integers specifying a permutation of 0, 1, …, n-1, where n is the number of dimensions in the decoded chunk representation provided as input to this codec.
> 
>     The string "C", equivalent to specifying the identity permutation 0, 1, …, n-1. This makes the codec a no-op.
> 
>     The string "F", equivalent to specifying the permutation n-1, …, 1, 0.

Currently, only C and F are supported.

fm3:
In https://github.com/scalableminds/webknossos/pull/7288 I add the possibility to specify an IntArray transpose codec. However, it will still only work correctly for C and F order (either expressed by string or by 4,3,2,1,0).

frcroth:
C and F are now no longer supported in the spec.

fm3:
I learned today that this does not touch the AxisOrder or the AdditionalCoordinates order. Instead it is only applied when reading a chunk.

So for this we should change the ArrayOrder (not AxisOrder) from the C/F switch to arbitrary permutations.

This would then need adaptation for getting the view on the source chunk `axisOrderXYZView` (the flip boolean needs to instead become an additional permutation using the transpose codec)

PR #7288, opened 2023-08-17
Zarr3 download for volume annotation, supporting nd

### LINKED ISSUE #7273: Support download+upload of ND volume annotations
Design a format for volume annotations for ND data. This might, in fact, change all volume downloads.
We might want to use zarr3 (no sharding) for that.

Note that the libs should be able to support this new kind of volume zip as well.

Upload will have to support both cases.

fm3:
- https://ndvolumedownload.webknossos.xyz

### Steps to test:
- Create annotation on 3d dataset, see download modal, try different options
- Create annotation on nd dataset, see download modal, try different options
- Download from other places like dashboard should still use wkw for the moment (untill the libs support zarr)
- Reupload of a zarr volume annotation should yield the same annotation
- Zarr metadata and datasource-properties.json in data.zip should look right.
- Unpacking a zarr data.zip and moving it to the a folder in binaryData should yield a usable webknossos dataset (this does not yet work in nd case because only specific axis orders are supported with additional coordinates)

### TODOs:
- [x] Backend
  - [x] add parameter selecting download as zarr
  - [x] write zarr chunks
  - [x] support nd in writing zarr chunks, (what axis order? should channels be included?)
  - [x] write zarr header (how too persist axis order?)
    - [x] Q: should the shape be the full layer shape? (starting at 0,0,0)? A: yes
    - [x] Q: should the fill_value be set? (0 is kind of wrong due to fallback layer?) A: use 0 anyway
    - [x] Q: should we set any additional attributes?
    - [ ] Get rid of the auto-generated field for the codecs → I don’t know how :(
  - [x] save compressed buckets
  - [x] save compression info in the zarr.json
  - [x] make parameter enum instead
  - [x] include dataZipFormat as tag in NML
  - [x] support upload
    - [x] in upload, distinguish between volume formats
       - [x] Q: should this be auto-detected from the contents or taken from the NML? Auto-detect for now
    - [x] get bucket position with additional coordinates from paths
    - [x] in upload, include assertions for the right zarr format
    - [x] support nd
    - [x] support merging volumes when uploading multiple
  - [x] include datasource-properties.json in data.zip to enable using this as a normal dataset
  - [x] include channel axis
    - [x] why does wk no longer display the layer as normal dataset? (even in 3d+c case)
  - [x] in zarr.json metadata, use array-of-int transpose codec instead of "F".
- [x] Frontend
  - [x] Select zarr3 download format for annotations with additional coordinates
  - [x] Say so in the download modal
  - [x] In places where downloads are started without the modal (i.e. dashboard), keep wkw for now until the libs support the zarr-based format, compare https://github.com/scalableminds/webknossos-libs/issues/951

### Issues:
- fixes #7273
- contributes to  #7116

------


MichaelBuessemeyer:
I took care of the frontend part of this pr:
The download modal now gives the user the option to choose between wkw and zarr for annotations with volume data. The UI looks like this in the current version:
image
In case one of the volume layers is nd, wkw is disabled as a download format and only zarr is available.

The other cases where annotations are downloaded (like tasks or projects), nd datasets are not supported for download right now.

frontend/javascripts/admin/admin_rest_api.ts
```diff
@@ -1035,21 +1035,28 @@ export async function downloadAnnotation(
   annotationType: APIAnnotationType,
   showVolumeFallbackDownloadWarning: boolean = false,
   versions: Versions = {},
+  volumeDataZipFormat: string = "wkw",
```
philippotto, line 1038: the type could be `"zarr3" | "wkw"` instead of string, I think.

frontend/javascripts/oxalis/view/action-bar/download_modal_view.tsx
```diff
@@ -268,10 +278,13 @@ function _DownloadModalView({
     (state: OxalisState) => state.temporaryConfiguration.isMergerModeEnabled,
   );
   const hasVolumeFallback = tracing.volumes.some((volume) => volume.fallbackLayer != null);
+  const isVolumeNDimensional = tracing.volumes.some((tracing) => tracing.additionalAxes.length > 0);
+  const hasVolumes = hasVolumeTracings(tracing);
+  const initialFilesToDownload = hasVolumes ? (isVolumeNDimensional ? "zarr" : "wkw") : "nml";
```
philippotto: I'd rename `initialFilesToDownload` and `filesToDownload` to `[initial]fileFormatToDownload`. otherwise, it sounds like the vars would contain arrays.

frontend/javascripts/oxalis/view/action-bar/download_modal_view.tsx
```diff
@@ -547,46 +555,42 @@ function _DownloadModalView({
                 }}
               >
                 Select the data you would like to download.
+                <Hint style={{ marginTop: 12 }}>
+                  An NML file will always be included with any download.
```
philippotto, line 559: will the NML always contain the skeletons? if not, then it's not possible to download volume + skeleton at the same time? and if yes, then it's not possible to download volume without skeletons? either way, it seems as this gives less control to the user? is this intended?

fm3: Indeed, skeletons are always included. This was already the case before this PR. We might want to change that in the future, but that’s unrelated to this change.

philippotto: Ah, I see, then never mind :)

webknossos-tracingstore/app/com/scalableminds/webknossos/tracingstore/tracings/volume/VolumeDataZipHelper.scala
```diff
@@ -31,21 +58,84 @@ trait VolumeDataZipHelper extends WKWDataFormatHelper with ByteUtils {
...
+              val data = compressor.decompress(dataCompressed)
+              block(bucketPosition, data)
+            }
+          }
+      }
+    } yield ()
+
+  private def parseZarrChunkPath(path: String, zarr3ArrayHeader: Zarr3ArrayHeader): Option[BucketPosition] = {
+    val dimensionNames = zarr3ArrayHeader.dimension_names.getOrElse(Array("x", "y", "z"))
+    val additionalAxesNames: Seq[String] = dimensionNames.toSeq.drop(1).dropRight(3) // drop channel left, and xyz right
```
frcroth, line 85: Seems somewhat brittle to me, especially if 2d datasets are also supported. Why not filter the dimensionNames by name like done when exploring?

webknossos-tracingstore/app/com/scalableminds/webknossos/tracingstore/tracings/volume/VolumeDataZipHelper.scala
```diff
@@ -31,21 +58,84 @@ trait VolumeDataZipHelper extends WKWDataFormatHelper with ByteUtils {
...
+
+    // assume additionalAxes,x,y,z
+    val chunkPathRegex = s"(|.*/)(\\d+-\\d+-\\d+)/c\\.(.+)".r
+
+    path match {
+      case chunkPathRegex(_, magStr, dimsStr) =>
+        val dims: Seq[String] = dimsStr.split("\\.").toSeq
+        val additionalAxesDims = dims.drop(1).dropRight(3) // drop channel left, and xyz right
+        val additionalCoordinates: Seq[AdditionalCoordinate] = additionalAxesNames.zip(additionalAxesDims).map {
+          case (name, coordinateValue) => AdditionalCoordinate(name, coordinateValue.toInt)
```
frcroth, line 95: See above

frcroth (approved):

app/controllers/AnnotationIOController.scala
```diff
@@ -491,7 +518,10 @@ Expects:
       project <- projectDAO.findOne(projectIdValidated) ?~> Messages("project.notFound", projectId) ~> NOT_FOUND
       _ <- Fox.assertTrue(userService.isTeamManagerOrAdminOf(user, project._team)) ?~> "notAllowed" ~> FORBIDDEN
       annotations <- annotationDAO.findAllFinishedForProject(projectIdValidated)
-      zip <- annotationService.zipAnnotations(annotations, project.name, skipVolumeData)
+      zip <- annotationService.zipAnnotations(annotations,
```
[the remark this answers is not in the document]
fm3: The Task system can be used to ask users to annotate certain areas of datasets in certain ways. Tasks are organized in projects. Users can then request a task and webknossos will automatically assign them.

These routes here allow the organization’s admins/team managers to download the result of such task annotations (or collections of such task annotations, e.g. grouped by project).

Eventually we want to enable ND datasets for this as well. However, since the UI does not currently supply format selection methods for these download routes, and since not every client can handle the new format yet, we decided that for now these features should stay with the WKW format, even if that means that they cannot be used productively with ND data.

PR #7289, opened 2023-08-17
Enable layer order manipulation in additive blend mode

### LINKED ISSUE #7286: Allow manipulation of layer order even when cover blend mode is turned off
## Detailed Description

It is unintuitive that a user cannot change the layer order when additive blend mode is turned on. Therefore it should always be possible to change the layer order.

MichaelBuessemeyer:
- https://___.webknossos.xyz

### Steps to test:
- Open a dataset with multiple color layers
- Turn on additive blend mode. You should be able to manipulate the layer order.
- Turn on cover blend mode. You should still be able to manipulate the layer order.
- Reloading should preserve the set color layer order.

### Issues:
- fixes #7286

------

### DIFF
--- a/CHANGELOG.unreleased.md
+++ b/CHANGELOG.unreleased.md
@@ -21,6 +21,7 @@ For upgrade instructions, please check the [migration guide](MIGRATIONS.released
 
 ### Changed
 - Small messages during annotating (e.g. “finished undo”, “applying mapping…”) are now click-through so they do not block users from selecting tools. [7239](https://github.com/scalableminds/webknossos/pull/7239)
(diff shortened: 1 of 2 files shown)


philippotto (approved):
lgtm :+1: I assume the dataset settings don't need to be changed, since they always allowed re-ordering (as long as multiple color layers exist)?

MichaelBuessemeyer:
> lgtm 👍 I assume the dataset settings don't need to be changed, since they always allowed re-ordering (as long as multiple color layers exist)?

Jup, pretty sure that's the case. Whether the order could be manipulated or not was simply determined by `const isSortingDisabled = colorLayerOrder.length < 2;`

PR #7290, opened 2023-08-17
One Version per Transaction

### LINKED ISSUE #6495: Versions of volume data should be created at "transaction" points

fm3:
Adapts the transaction management when saving updates to an annotation. Now, the version number is not increased inside of a transaction. The goal is a cleaner history, especially for volume annotations. Every version that shows up in the version history/restore view, should have a consistent state across mags, and complete brush strokes.

- https://___.webknossos.xyz

### Steps to test:
- brush a high volume of data
- save
- look at version history and check that each version has consistent data

### TODOs:
- [x] Backend
  - [x] assume constant version for all groups of a transactions
  - [x] no longer assert increasing versions inside of a transaction
  - [x] instead assert increasing versions compared to the latest committed transaction
  - [x] also assert that the transactionGroupIndex is increasing monotonically when storing uncommitted update groups for a transaction
  - [x] transaction info is no longer optional in an update group, the frontend always sends this
  - [x] store uncommitted updates for more than 20 minutes, what if the user is offline for that time?
- [x] Frontend
  - [x] send constant version number for all groups of a transaction

### Issues:
- fixes #6495

------


fm3:
Well possible that something’s wrong in the backend. I did this pretty quickly this afternoon. I can have a closer look on monday, or maybe we can debug together

philippotto:
I looked at it with a fresh mind and found a logic error in my approach. However, the new code still runs into 409 errors. It's probably best to debug this together :) Maybe I misunderstand the new protocol. For the record, this is what the front-end is currently sending when brushing a single voxel:

```
[
  {version: 64, transactionId: "aiauvz190a", transactionGroupCount: 2, transactionGroupIndex: 0,…},
  {version: 64, transactionId: "aiauvz190a", transactionGroupCount: 2, transactionGroupIndex: 1,…},
  {version: 65, transactionId: "u2p7dhx3mw", transactionGroupCount: 1, transactionGroupIndex: 0,…}
]
```

The previous version was at 63. The above request fails. the version numbers look correct to me, though.

fm3:
@philippotto I think I found the problem, could you try again?

Before we merge this, the backend should do one more assertion (see checkbox above), which I will add in the coming days.

philippotto:
> @philippotto I think I found the problem, could you try again?

Awesome, works for me :+1:

normanrz (approved):

fm3:
I think before we merge this we need to think about the retry logic again. In the old backend code there was a »failUnlessAlreadyHandled« method which suggests that this is about silently ignoring retries of saved actions. We probably need something like that again :thinking:

philippotto:
> The version list is still a bit noisy. From a user's pov, it is not clear what the difference between "updated segmentation" and "modified annotation" is. Also, I think it is confusing that updating the segment list is a separate action. Screenshot 2023-08-29 at 13 47 02

Yes, you are right. This is still not ideal, but at least the volume data should be consistent at all times now. I created a potential follow-up here: https://github.com/scalableminds/webknossos/issues/7300 (however, I see the priority far lower compared to the current PR).

@fm3 Shall we merge? :)

fm3:
> @fm3 Shall we merge? :)

As stated above, I think I have to re-visit the backend part once more, with respect to the retry logic. I’ll let you know once I think I restored it.

philippotto:
> > @fm3 Shall we merge? :)
> 
> As stated above, I think I have to re-visit the backend part once more, with respect to the retry logic. I’ll let you know once I think I restored it.

Oh, I missed your comment. Probably had the tab open for too long.. Makes sense what you wrote :+1:

fm3:
@philippotto I re-added the logic to ignore retries of requests that were already handled in the backend. I tested this locally and it seems to work. However, as discussed in person, let’s show this code to more people before merging, in order to avoid any subtle mistakes in this important part of the code.

fm3:
@philippotto I’m not sure how to resolve the conflicts in the changelog. Looks like you edited multiple lines, but it also looks like they are in master (released changelog) already? Could you have a quick look?

philippotto:
> @philippotto I’m not sure how to resolve the conflicts in the changelog. Looks like you edited multiple lines, but it also looks like they are in master (released changelog) already? Could you have a quick look?

I resolved the conflicts now. This PR added three lines and the rest was moved to the released file by the most recent release pr.

philippotto:
For the record: @MichaelBuessemeyer and I discussed the PR, too, so the front-end is now reviewed multiple times and should be good to go.

frcroth (approved):
Issue for race conditions?

fm3:
Thanks for going through this with me :tada: I created this follow-up issue https://github.com/scalableminds/webknossos/issues/7307

PR #7291, opened 2023-08-17
Fix rendering when switching visibility between multiple volume layers

philippotto:
- https://___.webknossos.xyz

### Steps to test:
- create an annotation for a dataset that has a volume layer
- add another volume layer
- switch visibility between these two volume layers
- ensure that the rendering works

### Issues:
- follow-up to #7100
- also see https://scm.slack.com/archives/C5AKLAV0B/p1692284081691689

------

### DIFF
--- a/frontend/javascripts/oxalis/geometries/materials/plane_material_factory.ts
+++ b/frontend/javascripts/oxalis/geometries/materials/plane_material_factory.ts
@@ -922,7 +922,7 @@ class PlaneMaterialFactory {
     // The third parameter returns the number of globally available layers (this is not always equal
     // to the sum of the lengths of the first two arrays, as not all layers might be rendered.)
     const state = Store.getState();
-    const sanitizedOrderedColorLayerNames =
+    const allSanitizedOrderedColorLayerNames =
       state.datasetConfiguration.colorLayerOrder.map(sanitizeName);
     const colorLayerNames = getSanitizedColorLayerNames();
     const segmentationLayerNames = Model.getSegmentationLayers().map((layer) =>
@@ -938,7 +938,7 @@ class PlaneMaterialFactory {
       return [
         colorLayerNames,
         segmentationLayerNames,
-        sanitizedOrderedColorLayerNames,
+        allSanitizedOrderedColorLayerNames,
         globalLayerCount,
       ];
     }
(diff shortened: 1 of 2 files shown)


frontend/javascripts/oxalis/geometries/materials/plane_material_factory.ts
```diff
@@ -971,11 +971,12 @@ class PlaneMaterialFactory {
       names,
       ({ isSegmentationLayer }) => !isSegmentationLayer,
     ).map((layers) => layers.map(({ name }) => sanitizeName(name)));
+    const colorNameSet = new Set(sanitizedColorLayerNames);
 
     return [
       sanitizedColorLayerNames,
       sanitizedSegmentationLayerNames,
-      sanitizedOrderedColorLayerNames,
+      allSanitizedOrderedColorLayerNames.filter((name) => colorNameSet.has(name)),
```
MichaelBuessemeyer, line 979: Does this ensure that `allSanitizedOrderedColorLayerNames` does not include null values and names of segmentation layer?

If that's the case this sure fixes the bug, but does not tackle the actual root cause of the problem. Imo `allSanitizedOrderedColorLayerNames` should only have valid sanitized color layer names in a specific order in it. If that's not the case here, maybe something around line 925 should be changed :thinking: 

If that's not the case, I do not understand what causes the bug.

MichaelBuessemeyer (approved):
Thanks for fixing this.:pray:
I tried to reproduce the bug locally and on a dev instance and both seemed fine to me -> bug seems fixed. :heavy_check_mark: 
As the bug is fixed, I'll leave an approve but I still need some clarification what actually caused the bug (please see my comment in the code). The approve should make things easier to merge as this bug seems pretty urgent to me.

frontend/javascripts/oxalis/geometries/materials/plane_material_factory.ts
philippotto: The problem is that often not all layers can be rendered. In that case, only a subset of layers is rendered (this method is responsible for choosing these). `sanitizedColorLayerNames` only contains the layers that should be rendered, but `sanitizedOrderedColorLayerNames` contained more in the error case. However, the arrays should be the same length. Due to this mismatch, the shader tried to refer to some layers that the shader shouldn't know of.
Maybe we overhaul the rendering approach at the next hackathon. So, I just merged the PR now :)

PR #7292, opened 2023-08-21
Hint about restricted magnifications

philippotto:
when disabling volume tools due to zoom level, tell user how to solve this situation by restricting resolutions. also simplify code a bit.

### Steps to test:
- looks like this:
image

### Issues:
- only few users know about this feature and get annoyed when they cannot annotate stuff. this change should help them.


philippotto:
@normanrz ping :)

normanrz (approved):

Issue #7089, opened 2023-05-22
Allow to make edges of a tree invisible


philippotto:
## Detailed Description
As an alternative to the existing soma-clicking mode, one could make the edges of a tree invisible. That way nodes are still grouped loosely but they look like they are independent. The main benefit is the reduced tree count (which is good for small NML files, a short tree list, performance etc).

## Context
https://scm.slack.com/archives/C5AKLAV0B/p1684748763542109

PR #7296, opened 2023-08-24
Add option to hide tree edges

MichaelBuessemeyer:
This PR makes it possible for users to hide the edges of tree and also persist this setting.

- https://___.webknossos.xyz

### Steps to test:
- Open a tracing
- Create two trees
- Enable/Disable the edge visibility of the trees using the context menu in the skeleton tab
- The nodes should always be visible (unless the whole tree is turned invisible)

### TODOs:
- [x] Adapt Tracing Store to changes. I need your help here @fm3 or maybe @frcroth 
- ~~[ ] Maybe also adapt wklibs client (as the trees have a new property)~~ decided to defer that
- ~~[ ] Adapt NML reading & writing~~ decided to defer that

### Issues:
- fixes [#7089](https://github.com/scalableminds/webknossos/issues/7089)

------


MichaelBuessemeyer:
@fm3 (or @frcroth) I am struggling with the tracing store as the newly added skeleton update action `updateTreeEdgesVisibility` still needs to be added. The same goes for the new property each tree now has (if it does not yet exist the frontend sets a default for skeletons retrieved by the server).

The tracingstore currently throws the following error:
image


Could you please help me out here and add the necessary changes?

And maybe you know whether wklibs needs to be adjusted to comply with the newly added property.

fm3:
Sure, we can add that. Just to make sure I’m understanding this correctly: There is a new update action that sets for a specific tree the `hideEdges` boolean? So I would adapt the skeleton proto definition to reflect that, and add the update action in the backend?

I’m wondering if the boolean’s naming is ideal. Since trees also have `isVisible`, maybe `edgesAreVisible = true`? That would also match the name “updateTreeEdgesVisibility”. But I don’t have a strong opinion here.

MichaelBuessemeyer:
> Sure, we can add that. Just to make sure I’m understanding this correctly: There is a new update action that sets for a specific tree the hideEdges boolean? So I would adapt the skeleton proto definition to reflect that, and add the update action in the backend?

I hope so, yes :sweat_smile:. But we can talk about my changes a little before you start so you only make the necessary changes. You text me when you want to start :)

> I’m wondering if the boolean’s naming is ideal. Since trees also have isVisible, maybe edgesAreVisible = true? That would also match the name “updateTreeEdgesVisibility”. But I don’t have a strong opinion here.

I noticed the same thing in the middle of coding these changes. Thus I simply applied your suggestion and renamed the property to `edgesAreVisible`.

> Related: Should we also adapt the NML format? In this case, we could indeed also adapt the python libs to get/set this boolean. But I’d say adapting the libs not super important/urgent.

Oh yeah, that's right. I think I already adjusted the reading nml in the frontend. But the writer still needs to be adjusted :+1:. I added the point to the ToDos.

fm3:
@MichaelBuessemeyer I added the backend changes, please see if they work for you as intended :)

Note that we decided against adapting NML (and thus the python libs) for the moment, compare https://scm.slack.com/archives/C5AKLAV0B/p1693224632337739?thread_ts=1684748763.542109&cid=C5AKLAV0B

frontend/javascripts/oxalis/model/helpers/nml_helpers.ts
```diff
@@ -721,6 +722,7 @@ export function parseNml(nmlString: string): Promise<{
               isVisible: _parseFloat(attr, "color.a") !== 0,
               groupId: groupId >= 0 ? groupId : DEFAULT_GROUP_ID,
               type: _parseTreeType(attr, "type", TreeTypeEnum.DEFAULT),
+              edgesAreVisible: _parseBool(attr, "edgesAreVisible", true),
```
philippotto, line 725: this is only for parsing right? what about serialization? can be invoked via the trees-tab ("save selected trees as nml" or sth like that)

frontend/javascripts/oxalis/geometries/materials/node_shader.ts
```diff
@@ -198,7 +198,8 @@ void main() {
     ));
 
     color = texture(treeColors, treeIdToTextureCoordinate).rgb;
-    bool isVisible = texture(treeColors, treeIdToTextureCoordinate).a == 1.0;
+    float alpha = texture(treeColors, treeIdToTextureCoordinate).a;
+    bool isVisible = alpha == 1.0 || alpha == 0.5 ;
```
philippotto, line 202: please add a comment that 0.5 is the value for invisible edges.

philippotto (commented):
looks good to me overall :+1: why didn't you need to adapt the edge_shader, though? don't the edges support alpha and 0.5 is then interpreted as 0, anyway? I think, you will need to change this, since the #7136 PR added alpha support for edges (because these can get half-transparent when going from one t value to another).

And another note: We have a generic update-tree-action which could have been used for updating the edgesAreVisible parameter. However, now that we have the dedicated update action, it's also fine :)

philippotto:
image

The icon doesn't look good on dark screen :see_no_evil:

frontend/javascripts/oxalis/model/helpers/nml_helpers.ts
MichaelBuessemeyer: You are right, I did not take care of serialization. Internally we agreed to postpone the support of the new `edgesAreVisible` field in nmls. Thus I did not implement any NML support besides what is needed for typescript. This line is needed because the created object `currentTree` in line `712`, because the field `edgesAreVisible` is required by this object. Alternatively, I could set the value to always be true, if you prefer this variation over already trying to parse a value that is not yet serialized, I would also be fine with that :)

philippotto: Ah, I see, makes sense :)

MichaelBuessemeyer:
> why didn't you need to adapt the edge_shader, though? don't the edges support alpha and 0.5 is then interpreted as 0, anyway? 

Exactly :+1: 

> I think, you will need to change this, since the https://github.com/scalableminds/webknossos/pull/7136 PR added alpha support for edges (because these can get half-transparent when going from one t value to another).

Ok, I'll find a way :+1: 

> And another note: We have a generic update-tree-action which could have been used for updating the edgesAreVisible parameter. However, now that we have the dedicated update action, it's also fine :)

:thinking: I was following the way the visibility of the whole tree is updated. This has its own action. But this is potentially due to multiple of these update actions later being combined into a single action.
I can still undo these changes and use the regular tree update action, if you think, that's cleaner. This shouldn't take up much time, right?

fm3:
I think there is no harm in the dedicated update action.

I’m not certain anymore, but I think the update action for the tree visibility was introduced because of backwards compatibility. There are old update actions stored in fossildb, as they are applied lazily in the backend. When applying those, we would have to interpret the missing field, and consider whether it should mean the visibility is not changed, or is set to true. While introducing that logic would be fine by me as well, the dedicated action is fine for me as well. The only downside might be a busier history. Philipp, if you think it would be better to unify this into the update tree action, we can certainly do it as well.

MichaelBuessemeyer:
> I think, you will need to change this, since the https://github.com/scalableminds/webknossos/pull/7136 PR added alpha support for edges (because these can get half-transparent when going from one t value to another).

This pr works perfectly together with the changes of the nd dataset pr. There were no changes needed from my side and it still works :100:

> The icon doesn't look good with the dark mode 🙈

I also fixed the icon in dark mode and increased the size of the rectangles a little.
image
image

philippotto:
> I think there is no harm in the dedicated update action.

Yes, I agree.



> > I think, you will need to change this, since the #7136 PR added alpha support for edges (because these can get half-transparent when going from one t value to another).
> 
> This pr works perfectly together with the changes of the nd dataset pr. There were no changes needed from my side and it still works 💯
> 
> > The icon doesn't look good with the dark mode 🙈
> 
> I also fixed the icon in dark mode and increased the size of the rectangles a little. image image

Awesome :+1: Will test now.

philippotto (approved):
Awesome, works great for me :)

PR #7308, opened 2023-09-05
Sanitize additional axis name

### LINKED ISSUE #7272: Guard against special characters in ND axis names
When importing datasets, additionalCoordinate names should be validated/adapted

frcroth:
- https://___.webknossos.xyz

I tested it against the local 4d dataset.

### Issues:
- fixes #7272

------
- [x] Updated [changelog](../blob/master/CHANGELOG.unreleased.md#unreleased) **Probably not necessary?**


fm3 (commented):
Would it be possible to have an error message in case the name becompes empty or non-unique after this? (Probably needs to change the return type to Box)?

fm3 (approved):

PR #7318, opened 2023-09-06
Fix default filter in users table

philippotto:
... so that only activated users are shown by default

- https://___.webknossos.xyz

### Steps to test:
- deactivate a user
- open users tab (don't use page refresh but instead open the url in a new tab, since the history state is used to remember the active filter)
- ensure that only enabled users are shown (use the filter UI in the column to verify that too)

### Issues:
- fixes regression caused by #7231 

------


hotzenklotz (approved):
LGTM

PR #7324, opened 2023-09-07
overwrite font in tab menu

fm3:
Is this how you do it?

image


philippotto:
thank you for taking the time to do this :) I moved the code to another module which is already doing some flexlayout overrides. I also changed the font-size to 14 since that was used already in the module..

philippotto (approved):

PR #7326, opened 2023-09-08
Fix too tight assertions in annotation saving

fm3:
Fixing regression introduced in #7290 

### Steps to test:
- Save skeleton with just a few clicks, should work
- Save skeleton with imported large nml, should work
- Save skeleton with a few clicks PLUS an import of a large nml, should work (this was the error case before)

------


normanrz:
So the problem was that all transactions with just 1 group would fail?

philippotto (approved):

fm3:
> So the problem was that all transactions with just 1 group would fail?

Only in requests with single-group transactions *and* bigger transactions

PR #7328, opened 2023-09-11
[WIP] Encode additional coordinates in url

philippotto:
- https://___.webknossos.xyz

### Steps to test:
- abc

### TODOs:
- [ ] ...

### Issues:
- fixes #

------

### DIFF
--- a/frontend/javascripts/oxalis/model/helpers/nml_helpers.ts
+++ b/frontend/javascripts/oxalis/model/helpers/nml_helpers.ts
@@ -340,16 +340,24 @@ function serializeNodes(nodes: NodeMap): Array<string> {
   });
 }
 
+export function additionalCoordinateToKeyValue(coord: AdditionalCoordinate) {
+  return [
+    // Export additional coordinates like this:
+    // additionalCoordinate-t="10"
+    // Don't capitalize coord.name, because it it's not reversible for
+    // names that are already capitalized.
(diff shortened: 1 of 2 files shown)


philippotto:
@normanrz What do you think about the encoding? If we want to go for brevity, we could change
`additionalCoordinate-t=123` to `aC-t=123` in the URL.

normanrz:
> @normanrz What do you think about the encoding? If we want to go for brevity, we could change
> `additionalCoordinate-t=123` to `aC-t=123` in the URL.

I like the key=value encoding. Concise and extendable. We should've done that from the start :D
I'd prefer a shorter key, such as `pos-t` or `aC-t`.

frontend/javascripts/oxalis/model/helpers/nml_helpers.ts
```diff
@@ -340,16 +340,36 @@ function serializeNodes(nodes: NodeMap): Array<string> {
   });
 }
 
+function getAdditionalCoordinateLabel(useConciseStyle: boolean) {
```
normanrz, line 343: Why do we need the non-concise style?

philippotto: this is used in the NMLs. do you think the NMLs should strive for brevity instead of clarity, too?

PR #7329, opened 2023-09-12
Fix transaction management bug

fm3:
Saved multiple update groups at the same version in the committed update store, which led to updates being overwritten.
This PR changes the logic so that the update groups are saved into one before stored to the committed store under that version number.

### Steps to test:
- Create large skeletons / volumes, save, reload, should all look right

------


philippotto (approved):
lgtm :+1: I also disabled the mapping-related transaction chunking (as discussed) and tested this locally.

PR #7331, opened 2023-09-13
Link Previews

### LINKED ISSUE #7061: Add link previews containing more information (dataset/annotation name/page title)
Whenever I share a webknossos link in slack, I get a little preview where it states what webknossos is. It would be nice to include more info about the concrete link (page title, dataset name/annotation name/etc) to quickly discern them.

fm3:
- https://reacthelmet.webknossos.xyz

### Steps to test:
- Copy links to annotations and datasets to slack
- If public or with sharing token, it should show name and sensible text and thumbnail preview
- Should also work for shortlinks
- Should not show relevant information in case of non-public datasets (if no sharing token is supplied)

### TODOs:
- [x] Use tokens for permission checks (this should only work on public datasets/annotations or when a token is supplied)
  - [x] ctx for annotations+datasets
  - [x] default depending on page type
  - [x] include thumbnail with token (not only for public)
  - [x] resolve short links
  - [x] Access Check for workflows?
  - [x] test that
- [x] Remove react-based attempts (would have been the nicer code, but sadly the remote sites don’t execute any javascript)

### Issues:
- fixes #7061 

------


app/oxalis/opengraph/OpenGraphService.scala
```diff
@@ -0,0 +1,152 @@
...
+      case _ => Fox.failure("not a matching uri")
+    }
+
+  private def thumbnailUri(dataset: Dataset,
+                           layerOpt: Option[DataLayerLike],
+                           organization: Organization,
+                           token: Option[String]): Option[String] =
+    layerOpt.map { layer =>
+      val tokenParam = token.map(t => s"&sharingToken=$t").getOrElse("")
+      s"${conf.Http.uri}/api/datasets/${organization.name}/${dataset.name}/layers/${layer.name}/thumbnail?w=1000&h=300$tokenParam"
```
normanrz, line 136: the size could live in a constant somewhere

app/oxalis/opengraph/OpenGraphService.scala
```diff
@@ -0,0 +1,152 @@
+package oxalis.opengraph
```
normanrz, line 1: would it be a lot of effort to rename oxalis --> webknossos?

fm3: Nope, I’ll do that in a separate PR. Truth be told, the package structure doesn’t make that much sense even with the renaming, but it would certainly be an easy improvement

PR #7333, opened 2023-09-15
disabled some actions when segmentation layer is visible but not edit…

### LINKED ISSUE #7207: Some segment actions are not properly disabled for non-editable segmentation layers
## Context
Annotation for a dataset with a segmentation layer, but where the segmentation layer is not editable.
https://scm.slack.com/archives/C5AKLAV0B/p1689585308901369?thread_ts=1688991494.182429&cid=C5AKLAV0B

## Expected Behavior
- "create new group" should be disabled
- user should not be able to rename segments

## Current Behavior
- create new group can be clicked but nothing happens
- segments can be renamed (on reload, this change will be lost)

## Steps to Reproduce the bug

1. create annotation for a dataset that has a segmentation layer but don't create an editable segmentation layer (e.g., only skeleton annotation layer or create a new/empty segmentation layer)
2. make sure the non-empty segmentation layer is visible
3. click on some segments to fill the segment list
4. use context menu on the root group and try to rename segments

dieknolle3333:
…able

### Steps to test:
- see issue

### TODOs:
- [x] fix var naming

### Issues:
- fixes #7207 

------


frontend/javascripts/oxalis/view/right-border-tabs/segments_tab/segments_view.tsx
```diff
@@ -1221,7 +1222,7 @@ class SegmentsView extends React.Component<Props, State> {
                 this.handleChangeMeshVisibilityInGroup(
                   this.props.visibleSegmentationLayer.name,
                   groupId,
-                  !this.areSelectedSegmentsMeshesVisible(),
+                  !areGroupOrSelectedSegmentMeshesVisible,
```
dieknolle3333, line 1225: not related to the bug, but saw that this lead to wrong behaviour

hotzenklotz (approved):
Code looks good. I still need to test.

hotzenklotz:
Tested and works like a charm. 

One little thing: I can still move segments into groups even when I am not supposed to. This was not directly part of the issue but perhaps it easy to tackle as well. Drag and drop moving a segment into a group should probably be prevented. If this is easy, please fix. Otherwise we merge as is.

dieknolle3333:
if it was just about drag and drop, the fix was easy. :)

fm3 (approved):
:shipit:

PR #7335, opened 2023-09-18
Update password hashing library

frcroth:
- https://___.webknossos.xyz

### Steps to test:
- Create users with passwords before applying this merge, ensure that you can still login afterward.
- ~~Check if front-end still works with CSP @philippotto perhaps you can take a look at the CSP? I had to enable unsafe-inline and unsafe-eval to get WK to run, which significantly weakens security. Don't know if the code can be adjusted to make it work without them. If inline scripts need to be supported but don't change often/dynamically, nonces or hashes can be used to enable only specific inline scripts.~~ -> Content Security Policy moved to #7345 

------


fm3:
@philippotto do you know if we need to whitelist any other script sources? How about google analytics, airbrake, etc?

philippotto:
I checked-in the olvy script so that we don't need to link it in a script tag. Couldn't test it, though, since the back-end didn't compile for me locally.

> @philippotto do you know if we need to whitelist any other script sources? How about google analytics, airbrake, etc?

Airbrake is imported via NPM. GA is a good point. It probably needs an explicit exclusion. It is pulled from `https://www.googletagmanager.com`. I guess, an exclusion for that domain is okay, since it's less likely to be compromised than `app.olvy.co`?

frcroth:
image

frcroth:
These script tags are causing the unsafe-inline problem:
image
And this is eval
image

conf/application.conf
```diff
@@ -24,9 +24,10 @@ play {
   }
   filters.headers {
     # Unset some of the security filters enabled in datastore.Filters
-    contentSecurityPolicy = null
+    contentSecurityPolicy = "script-src 'self'"
```
fm3, line 27: the same should probably be added to standalone-datastore.conf and standalone-tracingstore.conf

Note that you can already test this in production before merging by briefly setting this setting in the configmap. Talk to me or norman on how to do this

fm3 (approved):

PR #7342, opened 2023-09-18
Fix nightly

philippotto:
I pinned the view configuration for the nightly so that changes to the dataset default settings in prod don't affect the CI. I had to change the checked-in screenshot, since I didn't find out the originally used zoom value and didn't feel like experimenting with it.
Additionally, I cleaned up some typing.

### Steps to test:
- see succesful run at https://github.com/scalableminds/webknossos/runs/16893635127

### Issues:
- fixes failing nightly


hotzenklotz (approved):
LGTM. Thanks.

PR #7349, opened 2023-09-19
Support 2d ome zarr

### LINKED ISSUE #6740: Allow import of 2D OME-Zarr data
## Detailed Description
If a dataset only has 2 space dimensions, wk should be able to import it. We even have optimized ui for datasets with depth=1.

frcroth:
- https://___.webknossos.xyz

Tested for one dataset with channels and without.

### Steps to test:
- Explore and read a 2d ome zarr dataset

### Issues:
- fixes #6740 

------


webknossos-datastore/app/com/scalableminds/webknossos/datastore/datareaders/AxisOrder.scala
```diff
@@ -1,8 +1,16 @@
...
+import play.api.libs.json.{JsError, JsNumber, JsObject, JsResult, JsValue, Json, OFormat, Reads, Writes}
+
+sealed abstract class AxisOrder() {
+  def x: Int
+  def y: Int
+  def z: Int
+
+  def c: Option[Int]
+
+  def hasZAxis: Boolean = false
```
fm3, line 12: should not have default value in the abstract class. Also is there a reason not to make it a trait? Is this about the json reads again? :see_no_evil:

webknossos-datastore/app/com/scalableminds/webknossos/datastore/datareaders/DatasetArray.scala
```diff
@@ -30,11 +30,35 @@ class DatasetArray(vaultPath: VaultPath,
 
   protected lazy val chunkReader: ChunkReader = new ChunkReader(header)
 
+  // Helper variables to allow reading 2d datasets as 3d datasets with depth 1
```
fm3, line 33: Could you also add a comment in the header definition mentioning this? We might in the future want to use a value from the header class, and should instead look here.

webknossos-datastore/app/com/scalableminds/webknossos/datastore/datareaders/DatasetArray.scala
```diff
@@ -90,9 +114,9 @@ class DatasetArray(vaultPath: VaultPath,
   // Read from array. Note that shape and offset should be passed in XYZ order, left-padded with 0 and 1 respectively.
   // This function will internally adapt to the array's axis order so that XYZ data in fortran-order is returned.
   private def readAsFortranOrder(shape: Array[Int], offset: Array[Int])(implicit ec: ExecutionContext): Fox[Object] = {
-    val totalOffset: Array[Int] = offset.zip(header.voxelOffset).map { case (o, v) => o - v }
-    val chunkIndices = ChunkUtils.computeChunkIndices(axisOrder.permuteIndicesReverse(header.datasetShape),
-                                                      axisOrder.permuteIndicesReverse(header.chunkSize),
+    val totalOffset: Array[Int] = offset.zip(header.voxelOffset).map { case (o, v) => o - v }.padTo(offset.length, 0)
```
fm3, line 117: why is this necessary now?

fm3 (commented):
Thanks for fighting your way through this!

Do you think it would be a lot of effort to change it to a single case class AxisOrder with an optional z, instead of the two classes with the abstract one on top?

Then, wherever you have the match pattern now, we could have `if (axisOrder.hasZAxis) {} else {}`. Also, we would get the json format for free.

fm3:
Does readBytesWithAdditionalCoordinates also have to be adapted? :thinking:

webknossos-datastore/app/com/scalableminds/webknossos/datastore/datareaders/DatasetArray.scala
frcroth: For 2d datasets header.voxelsize will be 2d, but we need a 3d totaloffset, which zip of different lengths does not provide (it will only have a length of 2)

frcroth:
> Does readBytesWithAdditionalCoordinates also have to be adapted? 🤔

Looking at it I guess not, is this in scope of this PR?

fm3 (approved):
:+1:

PR #7350, opened 2023-09-19
WIP: Rename isosurface in to adHocMesh

### LINKED ISSUE #7243: Rename “isosurface“ to “adHocMesh” in code
Ad-hoc meshing may be a better term, backend still uses "isosurface", which is a quite general term (MATLAB legacy?)

frcroth:
- https://___.webknossos.xyz

### Steps to test:
- abc

### TODOs:
- [ ] ...

### Issues:
- fixes #7243

------


frcroth:
@philippotto I have also done the refactoring in the frontend now, there is a CI error which I don't understand.
I tested creating an adhoc mesh once and it seemed to work, but something else might be broken, since this refactoring was done without understanding what was refactored.

hotzenklotz (commented):
@philippotto Three questions for the frontend part: 
1. Do you know if there is a difference in terminology between `mesh` (pre-computed) and `mesh` (ex isosurface) that will be blurred by this renaming?
2. Are there any caches/local storage that rely on this naming? (object property name / keys)
3. What about URL encoding of the selected meshes? Does this reference "isosurface"?

MIGRATIONS.unreleased.md
```diff
@@ -5,6 +5,8 @@ See `MIGRATIONS.released.md` for the migrations which are part of official relea
 This project adheres to [Calendar Versioning](http://calver.org/) `0Y.0M.MICRO`.
 User-facing changes are documented in the [changelog](CHANGELOG.released.md).
 
+The "datastore/isosurface" configuration key was renamed to "datastore/adHocMesh".
```
philippotto, line 8: this should be moved into below the `## Unreleased\n[Commits](...)` lines.

philippotto:
@frcroth Very cool to this this clean up :+1: 

------------

@hotzenklotz You ask good question. Luckily, we should be safe on all these points, though.

> 1. Do you know if there is a difference in terminology between `mesh` (pre-computed) and `mesh` (ex isosurface) that will be blurred by this renaming?

No, I think, this should be fine.

> 2. Are there any caches/local storage that rely on this naming? (object property name / keys)

No, I don't think so. Should be fine.

> 3. What about URL encoding of the selected meshes? Does this reference "isosurface"?

I just checked this and this uses a boolean `precomputed` property. So, no issues here.

------------

The PR itself looks good. Will test now.

philippotto (commented):
Testing went well, too :+1: 

@frcroth Please update the changelog and mark the PR as ready :ship:

MIGRATIONS.unreleased.md
```diff
@@ -8,4 +8,6 @@ User-facing changes are documented in the [changelog](CHANGELOG.released.md).
 ## Unreleased
 [Commits](https://github.com/scalableminds/webknossos/compare/23.10.2...HEAD)
 
+The "datastore/isosurface" configuration key was renamed to "datastore/adHocMesh".
```
fm3, line 11:
```suggestion
The `datastore.isosurface` configuration key was renamed to `datastore.adHocMesh`.
```

webknossos-datastore/app/com/scalableminds/webknossos/datastore/controllers/BinaryDataController.scala
```diff
@@ -251,22 +251,22 @@ class BinaryDataController @Inject()(
...
   @ApiOperation(hidden = true, value = "")
-  def requestIsosurface(token: Option[String],
-                        organizationName: String,
-                        dataSetName: String,
-                        dataLayerName: String): Action[WebKnossosIsosurfaceRequest] =
-    Action.async(validateJson[WebKnossosIsosurfaceRequest]) { implicit request =>
+  def requestAdHocMesh(token: Option[String],
+                       organizationName: String,
+                       dataSetName: String,
+                       dataLayerName: String): Action[WebKnossosAdHocMeshRequest] =
```
fm3, line 260:
```suggestion
                       dataLayerName: String): Action[WebknossosAdHocMeshRequest] =
```

While whe’re here :sweat_smile: The camel case spelling is deprecated

webknossos-datastore/app/com/scalableminds/webknossos/datastore/services/AdHocMeshingService.scala
```diff
@@ -23,69 +23,69 @@ import scala.concurrent.duration._
...
+      sender() ! Failure("Unexpected message sent to AdHocMeshingActor.")
   }
 }
 
-class IsosurfaceService(binaryDataService: BinaryDataService,
-                        mappingService: MappingService,
-                        actorSystem: ActorSystem,
-                        isosurfaceTimeout: FiniteDuration,
-                        isosurfaceActorPoolSize: Int)(implicit ec: ExecutionContext)
+class AdHocMeshingService(binaryDataService: BinaryDataService,
```
fm3, line 52:
```suggestion
class AdHocMeshService(binaryDataService: BinaryDataService,
```

I think I’d prefer it without the “ing” since that would be the only place where the service would be named with a conjucated form.

fm3 (commented):
Great! I really enjoy how we’re moving towards more consistent terminology.

webknossos-datastore/app/com/scalableminds/webknossos/datastore/services/AdHocMeshingService.scala
frcroth: I removed all ings

fm3 (approved):

PR #7352, opened 2023-09-22
Fix auto save interval

philippotto:
Fixes a regression introduced in https://github.com/scalableminds/webknossos/commit/e719bc38c31ada5320c30ba79562786a9c9b3c1b which changed the auto-save interval from 30s to 300s.


hotzenklotz (approved):

hotzenklotz:
🙈

PR #7355, opened 2023-09-25
fix statistics bug and rename resolution methods to coarsest and finest

dieknolle3333:
I missed to places where `lowest`and `highest` resolution were mixed in the segment statistics modal. I fixed that and renamed the methods.
It also occured to us that the conversion from voxel to nm^3 was faulty, so this is also fixed for all segment stats.

### Steps to test:
- Create empty volume layer
- Brush
- Look at segment statistics modal and make sure the right volumes are displayed

------

### DIFF
--- a/frontend/javascripts/oxalis/view/context_menu.tsx
+++ b/frontend/javascripts/oxalis/view/context_menu.tsx
@@ -106,6 +106,8 @@ import { getSegmentBoundingBoxes, getSegmentVolumes } from "admin/admin_rest_api
 import { useFetch } from "libs/react_helpers";
 import { AsyncIconButton } from "components/async_clickables";
 import { type AdditionalCoordinate } from "types/api_flow_types";
+import { voxelToNm3 } from "oxalis/model/scaleinfo";
+import { getBoundingBoxInMag1 } from "oxalis/model/sagas/volume/helpers";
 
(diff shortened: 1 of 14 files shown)


frontend/javascripts/oxalis/view/context_menu.tsx
```diff
@@ -1122,13 +1122,13 @@ function ContextMenuInner(propsWithInputRef: Props) {
         const [segmentSize] = await getSegmentVolumes(
           tracingStoreUrl,
           tracingId,
-          mag.getLowestResolution(),
+          mag.getFinestResolution(),
```
fm3, line 1125:
```suggestion
          magInfo.getFinestResolution(),
```

I’ve also seen resolutionInfo for this thing elsewhere. Also fine by me.

I think it would be important to distinguish that this object contains information about all the available mags (resolutions)

fm3 (commented):
Fixes the bug for me and LGTM. Thanks for looking into this so quickly!

I added one comment on naming. While this was not introduced in this PR, I think it should be changed to avoid future confusion.

dieknolle3333:
thank you for the very fast review!

fm3 (approved):

fm3 (approved):
Nice! :sunglasses: 

Could you adapt the PR description and changelog to briefly mention the faulty computation that this fixes?
Then this is good to go from my side

PR #7367, opened 2023-10-05
Define CSP

### LINKED ISSUE #3085: Investigate and determine a reasonable Content-Security-Policy to prevent potential future XSS issues
Originally part of #1685.
See https://content-security-policy.com/ for documentation on CSP.

### LINKED ISSUE #7345: Define Content Security Policy
See https://github.com/scalableminds/webknossos/pull/7335 for context. We will split-off the CSP part into a separate PR, though.

Known problems for a strict CSP are:
- [ ] google analytics (maybe import a package from NPM instead of using a script-tag?)
- [ ] olvy (check-in their script; see https://github.com/scalableminds/webknossos/pull/7335/commits/63a68257072148a26cb80945950b7ffb4001c679)
- [ ] webpack sourcemaps (these use eval; maybe something like [this](https://github.com/webpack/webpack/issues/5627#issuecomment-374386048) helps, but it should be thoroughly tested, since I remember source map problems in the past with different modes)
- [ ] inline script in main.scala.html (potentially fixable with a hash)

Also test:
- [ ] sandbox mode in combination with connectome viewer

MichaelBuessemeyer:
This PR introduces a CSP. The main changes are:
- All parameters used in the inline scripts in the main.html were put into meta tags
- Inline scripts were moved into js/ts files where possible (loading the initial style sheet).
- The inline scripts read the parameters from the meta tags. Thus the code inside the meta tags should be static and therefore the hash of the code should also be static. But this needs testing! The hashes of these inline scripts are included in the CSP to allow them.

- https://definecsp.webknossos.xyz

### Steps to test:
- This PR defines a reasonably narrow CSP for wk :).
- Check whether the CSP works on a dev branch.
- Check whether the CSP works with the connectome viewer. I tried this in my local dev env and it seemed to work.

### TODOs:
- [x] Internal Diskussion
- [x] Check connectome viewer
  - I checked this locally and everything seemed to work. But after merging we should recheck whether the connectome viewer still works.

### Issues:
- fixes #7345
- fixes #3085

------
- [x] Updated migrations log
- [ ] After merging and deploying check whether the connectome viewer still works


app/views/main.scala.html
```diff
@@ -27,42 +35,13 @@
...
-          'rel="stylesheet" ' +
-          'type="text/css" ' +
-          'media="screen" ' +
-          'href="/assets/bundle/' +
-          initialTheme +
-          '.css?nocache=@(webknossos.BuildInfo.commitHash)" ' +
-          "onload=\"document.documentElement.style.display = ''\" " +
-          "/>",
-      );
-    </script>
```
MichaelBuessemeyer, line 53: This script was moved into `main.tsx` and the `selectedTheme` and the commit hash are accessed via the meta tags

webpack.config.js
```diff
@@ -185,7 +185,7 @@ module.exports = function (env = {}) {
       },
     },
     // See https://webpack.js.org/configuration/devtool/
-    devtool: env.production ? "source-map" : "eval-source-map",
+    devtool: "source-map",
```
MichaelBuessemeyer, line 188: The problem with this change is that according to the docs, `"eval-source-map"` is faster but should only be used in production (as we did before). Thus, keeping these changes might slow down our frontend rebuilding time upon changes to the frontend code in dev mode.

app/views/main.scala.html
```diff
@@ -27,42 +35,13 @@
...
       data-airbrake-project-key="@(conf.Airbrake.projectKey)"
       data-airbrake-environment-name="@(conf.Airbrake.environment)"
     ></script>
     <script src="/assets/bundle/vendors~main.js?nocache=@(webknossos.BuildInfo.commitHash)"></script>
     <script src="/assets/bundle/main.js?nocache=@(webknossos.BuildInfo.commitHash)"></script>
-    <script type="text/javascript" src="https://app.olvy.co/script.js" defer="defer"></script>
-    <script>
-      // Suppress warning emitted by Olvy because it tries to eagerly initialize
-      window.OlvyConfig = null;
-    </script>
```
MichaelBuessemeyer, line 65: The `window.OlvyConfig = null;` was moved to the end of the `ovly.js` file.

conf/application.conf
```diff
@@ -24,7 +24,7 @@ play {
   }
   filters.headers {
     # Unset some of the security filters enabled in datastore.Filters
-    contentSecurityPolicy = null
+    contentSecurityPolicy = "default-src 'self'; script-src 'self' 'unsafe-eval' www.googletagmanager.com 'sha256-NA81o2FV8jh0UizSPsBz2qEGpXfap54Eghuo7vxrH8g='; style-src 'self' 'sha256-DOu86drLfwUr1Wcsx/wxfqAogK7tFvJGjVmF/300H/M=' 'sha256-47DEQpj8HBSa+/TImW+5JCeuQeRkm5NMpJWZG3hSuFU='; connect-src 'self' ws://localhost:9002 app.ovly.co; frame-src webknossos.olvy.co; img-src 'self' data:"
```
MichaelBuessemeyer, line 27: The hashes are necessary for the inline scripts of the `main.scala.html` file. The hashes should not change afaik as the code should be static.
The variables were moved into meta tags and are loaded in the scripts by accessing the meta tags

conf/application.conf
```diff
@@ -24,7 +24,7 @@ play {
   }
   filters.headers {
     # Unset some of the security filters enabled in datastore.Filters
```
daniel-wer, line 26: The comment is no longer needed, I'd say

conf/application.conf
daniel-wer, line 27: There seems to be a typo in the `connect-src`, it should be `app.olvy.co`.

daniel-wer, line 27: It would be helpful to document which hash belongs to which script and how to recompute the hash if the script will ever be changed.

daniel-wer, line 27: What is `ws://localhost:9002` needed for?

daniel-wer: Ah, I see it will be removed before merging :+1:

daniel-wer (commented):
I did some preliminary testing:

- This CSP style-src error comes up frequently:
  ```
  dynamicCSS.js:81 Refused to apply inline style because it violates the following Content Security Policy directive:
    "style-src 'self' 'sha256-DOu86drLfwUr1Wcsx/wxfqAogK7tFvJGjVmF/300H/M=' 'sha256-47DEQpj8HBSa+/TImW+5JCeuQeRkm5NMpJWZG3hSuFU='".
  Either the 'unsafe-inline' keyword, a hash ('sha256-6Fdhs60adtuoY86RfyIYLvprzz7BHDBBVch0HlpsPhk='), or a nonce ('nonce-...') is required to enable inline execution.
  ```
- Clicking on the What's New bell in the top right corner doesn't load the olvy integration :/
- Adding a user script (e.g. https://gist.github.com/daniel-wer/0ffd533ae6cc8c5ea48df7a4e3d77a43) in the scripts admin view and then trying to load it in an annotation fails with;
  ```
  fetch.js:21 Refused to connect to 'https://api.github.com/gists/d5ff5d490ab381af9405c9078096c723' because it violates the following Content Security Policy directive: "connect-src 'self' ws://localhost:9002 app.ovly.co".
  fetch.js:21 Refused to connect to 'https://api.github.com/gists/d5ff5d490ab381af9405c9078096c723' because it violates the document's Content Security Policy.
  ```
  User scripts are only allowed to be github gists.

I'll try to think of more cases we should test.

frontend/javascripts/libs/olvy.js
```diff
@@ -0,0 +1,520 @@
+// Downloaded on 2023-10-05 from https://app.olvy.co/script.js and adjusted to comply with our linting rules
```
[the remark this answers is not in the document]
MichaelBuessemeyer: Yes, this is possible to allow olvy to be loaded externally. The downside is that this script inline CSS rules. As these rules are added by the code using a hash is not possible (afaik). Or else when the CSS changes in the olvy script, the hash would be invalidated and olvy cannot be loaded properly. Therefore `unsafe-inline` needs to be enabled for `style-src`. But I think this ok ok for now.

MichaelBuessemeyer:
I am currently left with one problem:
webpack does some kind of injection in function `injectScriptSync`. This is only done in firefox as the comment states:
>  // We run scripts on the page via the service worker (backgroud.js) for
// Manifest V3 extensions (Chrome & Edge).
// We need to inject this code for Firefox only because it does not support ExecutionWorld.MAIN
// https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/scripting/ExecutionWorld
// In this content script we have access to DOM, but don't have access to the webpage's window,
// so we inject this inline script tag into the webpage (allowed in Manifest V2).

But I think this is only an issue for local development. Thus, for the local development we could also simply enable `unsafe-inline` or so for scripts. This might fix the issue.

conf/application.conf
```diff
@@ -23,8 +23,8 @@ play {
     """
   }
   filters.headers {
-    # Unset some of the security filters enabled in datastore.Filters
-    contentSecurityPolicy = "default-src 'self'; script-src 'self' 'unsafe-eval' www.googletagmanager.com 'sha256-NA81o2FV8jh0UizSPsBz2qEGpXfap54Eghuo7vxrH8g='; style-src 'self' 'sha256-DOu86drLfwUr1Wcsx/wxfqAogK7tFvJGjVmF/300H/M=' 'sha256-47DEQpj8HBSa+/TImW+5JCeuQeRkm5NMpJWZG3hSuFU='; connect-src 'self' ws://localhost:9002 app.ovly.co; frame-src webknossos.olvy.co; img-src 'self' data:"
+    # The sha256 hash in the script-src refers to the google analytics configuration script in main.scala.html. 
+    contentSecurityPolicy = "default-src 'self'; script-src 'self' 'unsafe-inline' 'unsafe-eval' *.googletagmanager.com  app.olvy.co webknossos.olvy.co; style-src 'self' 'unsafe-inline'; connect-src 'self' ws://localhost:9002 app.olvy.co api.github.com gist.github.com *.google-analytics.com api.airbrake.io; frame-src webknossos.olvy.co; img-src 'self' data:"
```
daniel-wer, line 27: Note to myself: As far as I understood, unsafe-inline shouldn't be allowed as a script source, but only as a style-src, because of olvy.

```suggestion
    contentSecurityPolicy = "default-src 'self'; script-src 'self' 'unsafe-eval' *.googletagmanager.com  app.olvy.co webknossos.olvy.co; style-src 'self' 'unsafe-inline'; connect-src 'self' ws://localhost:9002 app.olvy.co api.github.com gist.github.com *.google-analytics.com api.airbrake.io; frame-src webknossos.olvy.co; img-src 'self' data:"
```

frontend/javascripts/main.tsx
```diff
@@ -27,8 +27,9 @@ import Model from "oxalis/model";
 import { setupApi } from "oxalis/api/internal_api";
 import { setActiveOrganizationAction } from "oxalis/model/actions/organization_actions";
 import checkBrowserFeatures from "libs/browser_feature_check";
-// @ts-ignore
-import _olvy from "libs/olvy.js";
+
+// Suppress warning emitted by Olvy because it tries to eagerly initialize
+window.OlvyConfig = null;
```
daniel-wer, line 32: Note to myself: Double-check that this is working. Last time I tested, I saw the warning, anyways.

conf/application.conf
daniel-wer: According to [this comment](https://github.com/scalableminds/webknossos/pull/7367#issuecomment-1757717225) unsafe-inline is needed as a script-src in firefox to use the react dev tools. Shouldn't be used in production, though.

frontend/javascripts/main.tsx
daniel-wer: After checking, it doesn't seem to be needed any longer. The warning I saw was "[Olvy] - Target element not found. Call Olvy.show() to open your releases widget manually", but that is printed on master, too. The cause is that we are using a dummy target element for olvy. Relevant comment

```javascript
// This target needs to be defined (otherwise, Olvy crashes when using .show()). However,
// we don't want Olvy to add any notification icons, since we do this on our own. Therefore,
// provide a dummy value here.
target: "#unused-olvy-target",
```

daniel-wer (approved):

PR #7376, opened 2023-10-11
Add Segment Index for Existing Volumes Without Losing History

fm3:
- Add route to migrate volume annotation layers to have segment indices. 
   - This creates a new version for the volume annotation layers.
   - Should be called during a downtime (compare migration guide)
- Also adds a little optimization to volume annotation downloads (skip empty buckets)

### Steps to test:
- Create some volume annotations
- For testing I created annotations without segment index by setting `hasSegmentIndex = Some(false),` in NmlParser.scala (line 102)

### TODOs:
- [x] check wether this adds to annotation time/statistics
- [x] test and measure time → took 70s for 60k buckets. extrapolation is 40min for 2.3M buckets (largest known volume layer)
- [x] create script/route to do this for all volume annotation layers
  - [x] option to restart with offset, in case of failure halfway
  - [x] cache access checks (it’s the webknossos token. tracingstore could send specific request for 
that first)
  - [x] parallelization
  - [x] log failures
- [x] ci
- [x] remove testing changes (unzip only first 100k files, no segment index for uploaded)
- [x] check how this looks in version history

### Issues:
- fixes https://github.com/scalableminds/webknossos/issues/7144

------


philippotto:
> Should be called during a downtime (compare migration guide)
How exactly would this work? I'm asking because WK has to run so that the migration route can be called. Do we need to cut off wk from inbound requests and then make the migration call from within our server?

> test and measure time → took 70s for 60k buckets. extrapolation is 40min for 2.3M buckets (largest known volume layer)

Do you have an estimate for how long it will take to migrate wkorg?

philippotto:
> > Do you have an estimate for how long it will take to migrate wkorg?
> 
> Unfortunately I don’t. There aren’t currently any routes to determine how big the existing volume annotation layers are. I could of course build routes to count that now, but I don’t know how the priority for that would be.

Let's assume the worst case for now and consider the case where the migration would take several days. Would we be able to cancel the migration and then continue it at a later point. For example, so that we could run it at night on multiple days? Or could this create inconsistent states? If the migration was "resumable", I don't think that the runtime estimation is essential.

fm3:
That should be possible, yes. Re-starting will trigger another run with all tracing ids from the DB, but tracings that are already successfuly migrated will notice that and it will be a no-op for them

fm3:
We could also do a dry-run (without flush to db so it wouldn’t need downtime) to get a fairly good estimate on the runtime. Would do this with a bit less parallelity to not impair normal operation

normanrz:
I think a dry-run would be good. 👍

frcroth (approved):

PR #7377, opened 2023-10-11
Fix segment stats for layers without segment index

dieknolle3333:
### Steps to test:
- make sure you have a visible annotation layer without a fallback layer and without segment index, e.g. by viewing an old annotation (created before segment index was introduced) or by setting `hasSegmentIndex` to false in `volumetracing_reducer`
- brush and right-click a segment in the viewport. make sure there are not segment stats shown. also make sure that there is no `Show Segment Statistics` action in the context menu in the right border segment tab.
- click anywhere in the viewport where there is no segment and make sure the segment stats are not shown either

### Issues:
- fixes https://scm.slack.com/archives/C5AKLAV0B/p1696856119961309

------
- [x] Considered [common edge cases](../blob/master/.github/common_edge_cases.md) I tried, I swear :sweat_smile:


hotzenklotz (approved):
LGTM

PR #7379, opened 2023-10-11
Fix "Import Agglomerate Skeleton" tooltip

daniel-wer:
Ensure layer mappings are actually loaded if context menu tooltip is triggered to avoid showing the loading message forever. The same is done when hovering the proofreading tool button in the action bar.

See [this](https://scm.slack.com/archives/C5AKLAV0B/p1696854134360809) slack thread for context.

I also wanted to advertise (link to) https://webknossos.org/services/automated-segmentation in case no agglomerate or connectome file was found, but I didn't find a good solution. Linking in the tooltip is weird, because all our tooltip usually don't receive pointer events and are click-through. Changing that leads to tooltips in the context menu blocking other context menu entries. Adding a question mark with a link in the context menu entry itself also doesn't work since the disabled context menu entry is not clickable. Not sure how important this is and whether @hotzenklotz or @philippotto have any other ideas?

- https://fixagglomerateskeletontooltip.webknossos.xyz

### Steps to test:
- Open an annotation, open the context menu and hover over the "Import Agglomerate Skeleton" entry - It should not show "Checking for agglomerate files..." but a more specific reason for why it is disabled.

### Issues:
- fixes https://scm.slack.com/archives/C5AKLAV0B/p1696854134360809

------


philippotto:
> I also wanted to advertise (link to) https://webknossos.org/services/automated-segmentation in case no agglomerate or connectome file was found, but I didn't find a good solution. Linking in the tooltip is weird, because all our tooltip usually don't receive pointer events and are click-through. Changing that leads to tooltips in the context menu blocking other context menu entries. Adding a question mark with a link in the context menu entry itself also doesn't work since the disabled context menu entry is not clickable. Not sure how important this is and whether @hotzenklotz or @philippotto have any other ideas?

Would a popover (that includes the link) work that opens on hover?

daniel-wer:
> Would a popover (that includes the link) work that opens on hover?

It would probably work, but also block other context menu entries which could get annoying :thinking: While I experimented with the tooltip, I also tried to show them left or right of the context menu entries to avoid the blocking, but that looked rather weird.

Another idea I just had was to enable the context menu entries, anyways, and then show a small modal advertising our automated segmentation in case no agglomerate or connectome file is available.

hotzenklotz:
For the nav menu and pricing plan checks / infos, we face them same issue with tooltips and hovering. There we used popovers like Philipp mentioned:

https://github.com/scalableminds/webknossos/blob/cc6532aa2c3fc215266425bc2019318b618a6a46/frontend/javascripts/navbar.tsx#L184-L185

philippotto:
> > Would a popover (that includes the link) work that opens on hover?
> 
> It would probably work, but also block other context menu entries which could get annoying 🤔 

> Another idea I just had was to enable the context menu entries, anyways, and then show a small modal advertising our automated segmentation in case no agglomerate or connectome file is available.

Both options sound fair to me. If the first one feels annoying (not sure how often it would be triggered by accident; it's probably a bit more likely than for the pricing enforcers that @hotzenklotz mentioned :thinking:), the second shouldn't pose problems hopefully..

daniel-wer:
> If the first one feels annoying (not sure how often it would be triggered by accident; it's probably a bit more likely than for the pricing enforcers that @hotzenklotz mentioned 🤔)

Since it's part of the context menu and would be triggered on hover, it would be triggered quite often, I'm afraid. I'll try the second method and show the popover on click.

daniel-wer:
I switched to a modal since clicking usually lets the context menu disappear, making it harder to show a popover. Since the respective menu entries are always enabled now, a warning icon is shown next to them if the functionality doesn't actually work. What do you think?

automated-segmentation-services-ad

Also, I tweaked the style of the shortcut images in the context menu a bit to avoid uneven line heights and spacing.

philippotto:
I like that :)

> Also, I tweaked the style of the shortcut images in the context menu a bit to avoid uneven line heights and spacing.

awesome :tada:

daniel-wer:
@hotzenklotz This is ready for review, then :)

hotzenklotz (approved):
Tested and works. Code LGTM.

some high-level feedback:
- The modal does not have a title, right? This leads to weird offset between the top of the icon and the first paragraph. Consider adding a title. "Ups, something is missing?" or "Get more out webknossos" or "webknossos service" etc
  Screenshot 2023-10-18 at 13 16 13
- I am not a fan of the yellow warning symbols. They will be visible to the majority of our user base once they open the right click context menu. Such a strong warning should not be the visually-strongest element for day-to-day work
- I am still confused why both menu entry talk about "Agglomerates". Why can't the second entry be something about just "Synapses" or "Synapse + Connectome" or something.

daniel-wer:
@hotzenklotz Thank you for your feedback!

I'll add a title :)

Do you have any suggestions for the latter two points?
Regarding 2) I wanted to include a visual indicator that the functionality is not actually available, because it is no longer disabled in that case :thinking: 
Regarding 3) The entry is currently named after what it does. It loads the agglomerate skeleton of the segment that's clicked on as well as its synapses.

hotzenklotz:
Regarding:

2) Mhm, we could use the same font as for disabled items and the lucky few who still click on the entry will get the modal. But that is also somewhat misleading. 
We could add new sub menus. Something about "Advanced" or "Services" or something? 

3) In that case it could also be called "Load connectome file" something like that. It seems to me that the agglomerate part of the name is an implementation detail. Something about synapse or connectome describes it purpose better.

daniel-wer:
1) I added a title
image

2) I toned down the color of the icons so they grab way less attention now
image

3) I'm not convinced by the suggestion as the button doesn't load "the connectome file". It loads an agglomerate as well as its synapses _from_ the connectome file. Why is the agglomerate part an implementation detail? The button does the same as the button above, but also loads the synapses.
An alternative proposal at the time was "Import from connectome" - do you think that's better?

hotzenklotz:
If the intention is to advertise this feature more, then "synapse" or "connectome" should be front and center. "Agglomerate skeleton" is lesser-known term.
The fact that loading the connectome / synpases will show the agglomeration skeleton (i.,e. instead of a mesh) is still a implementation detail for me.

philippotto:
I think, @daniel-wer's point is that "connectome" is not the correct term for what kind of data is imported when the user clicks that button. The term connectome refers to the overall structure of **all** neurons. However, the button only adds a part of that to the annotation, namely "one neuron" or in our lingo "one agglomerate" plus the corresponding synapses, if I'm not mistaken.

I'm not sure about the best way here, but thinking out loud, here are some suggestions:

- Import/visualize/load neuron and synapses (from connectome file)
- Import neuron as skeleton
- Nested entry à la "Use/from connectome file" > "Load neuron and synapses". However, nesting with only one child element is a bit nonsensical.

If "Import Agglomerate Skeleton" should be renamed as well, it could be "Import/visualize/load segment (as skeleton) (from mapping)". Note that this menu entry does not use the connectome file, but instead the agglomerate mapping. For that reason, I would not use the word "neuron", since we don't know what kind of cells are at hand in that case.

daniel-wer:
I talked to @hotzenklotz and we decided to rename "Import Agglomerate and Synapses" to "Import Synapses" to better distinguish it from "Load Agglomerate Skeleton"

PR #7386, opened 2023-10-15
Make organization name its id, drop the mongo id

### LINKED ISSUE #7689: Make the organization name its id, drop the mongo id
Proposing this change to the organization structure. This would also match the terminology already introduced in the webknossos-libs

display name → name
name → id
id - remove the mongoid completely

Blocked by voxelytics logs in loki. 

https://github.com/scalableminds/webknossos/pull/7687 should be deployed everywhere where voxelytics logs are stored for at least 12 weeks. Then we can proceed with this issue. From what I can tell that means after July 8, 2024 we should be safe

fm3:
The old organizationId (which was a mongoId/ObjectId) is dropped. The old organizationName is renamed to organizationId. The old organizationDisplayName is renamed to organizationName.

- https://organizationidvsname.webknossos.xyz

### Steps to test:
- [x] Test annotating, annotation download+reupload, signup, login
- [x] Test task system
- [x] Test libs access
- [x] Test migrations

### TODOs:
- [x] update in sql schema
- [x] update in scala code
- [x] the mongoid is used as key in loki?
- [x] update in routes
- [x] update in proto
- [x] update in frontend
- [x] checks on name in sql and create organization code (allow `[A-Za-z0-9\-_. ]+` in sql)
- [x] test around
- [x] sql migration, careful with foreign keys
  - [x] to make rollback easier, keep old ids, rename to nullable unused column?
  - [x] migrate analytics events json in postgres
    - [x] for foreign-webknossos analytics events it may be that the id cannot be looked up. talk to norman on how to handle these. 
- [x] make sure that analytics events those from old webknossoses can still be ingested
- [x] make sure that analytics events those from new webknossoses can still be ingested
- [x] bump api version, provide adapters → only for routes used by python client. e.g. dataset_list
- [x] todos in code
- change worker job arguments from organization_name to organization_id: Follow-up #8038
- adapt vx runs moving script in notion, make it more robust -> can be done later
- [x] What do to with `OrganizationCreationParameters` which has the `organizationDisplayName` prop? -> renaming to `organizationName`? -> its fine, just do it
- [x] check whether frontend events eventProperties need to be migrated
- [x] Mention in reversion that the newly created orgas first need a manual valid `_id_old` entry for the reversion to work.
- [x] The update route for organizations was changed. (The param `displayName` was renamed to `name`). Should a versioned API route be created for this? No need, that route is only used by our frontend.
- [x] Test wk thoroughly

### Issues:
- fixes #7689

- follow-up: #8039
- follow-up: #8038
- follow-up (libs): https://github.com/scalableminds/webknossos-libs/issues/1178

------
- Adapted [wk-libs python client](https://github.com/scalableminds/webknossos-libs/tree/master/webknossos/webknossos/client) if relevant API parts change → Follow-up: https://github.com/scalableminds/webknossos-libs/issues/1178


tools/postgres/schema.sql
```diff
@@ -318,8 +318,7 @@ CREATE TABLE webknossos.timespans(
 
 CREATE TYPE webknossos.PRICING_PLANS AS ENUM ('Basic', 'Team', 'Power', 'Team_Trial', 'Power_Trial', 'Custom');
 CREATE TABLE webknossos.organizations(
-  _id CHAR(24) PRIMARY KEY,
-  name VARCHAR(256) NOT NULL UNIQUE,
+  _id VARCHAR(256) PRIMARY KEY CHECK (_id ~* '^[A-Za-z0-9\-_. ]+$'),
   additionalInformation VARCHAR(2048) NOT NULL DEFAULT '',
   logoUrl VARCHAR(2048) NOT NULL DEFAULT '',
   displayName VARCHAR(1024) NOT NULL DEFAULT '',
```
MichaelBuessemeyer, line 324: still needs to be renamed to display name

and keep old _id column somehow for potential rollbacks

MichaelBuessemeyer:
Frontend analytics events do not include any references to organizations in their event properties. Thus, they do not need to be mirgated :tada:

tools/postgres/schema.sql
```diff
@@ -685,9 +686,9 @@ CREATE TABLE webknossos.analyticsEvents(
   created TIMESTAMPTZ NOT NULL DEFAULT NOW(),
   sessionId BIGINT NOT NULL,
   eventType VARCHAR(512) NOT NULL,
-  eventProperties JSONB NOT NULL,
+  eventProperties JSONB NOT NULL, -- TODO migrate user json from old to new organization id
```
fm3, line 689: @MichaelBuessemeyer is this TODO still open?

fm3 (commented):
LGTM :)

Two questions: 
 - There is an open TODO comment in schema.sql, I think it is actually resolved. Can you double-check that?
 - I think I remember you talking with Norman about the idea of a full backup of the analyticsEvents table. That seems to not be included in the evolution. Does that plan still stand? If so, do we want to do this manually or do we want to add it to the evolution?

I fixed a couple of small things in my last commits.

I can’t hit approve since I was the original author. Do you want to do so?

tools/postgres/schema.sql
MichaelBuessemeyer: nope, should be resolved

MichaelBuessemeyer (approved):
LGTM checked newest changes together with @fm3

PR #7389, opened 2023-10-16
Allow Exploring Datasets on Local File System

### LINKED ISSUE #7387: ExploreAndAdd route: support datasets on local file system
The explore route should also take paths on the local file system.
 - There should be a whitelist in the config for which directories wk may do reads (outside of its working directory)
 - Data loading should also work with this kind of path

fm3:
- Unifies FileSystemDataVault with other DataVaults, removes FileSystemVaultPath. VaultPaths now use  `file://` uris with absolute paths for the local file system.
- Allows exploring datasets in the local file system, in directories listed in the new config key `datastore.localFolderWhitelist`
- Note that this introduces `file://` uris also in the resulting datasource-properties.jsons. Those will not be sensibly explorable by the `WebknossosZarrExplorer` but I guess that isn’t a use case anyway

### Steps to test:
- Add a directory to config setting `datastore.localFolderWhitelist`
- Put dataset directory in there (or into its children/descendents)
- Explore that dataset in the Add Datasets view, prepend `file://` to the absolute path (so there are then three consecutive slashes)
- Should work
- Should not work is the directory is not in the whitelist
- Other datasets with mag locators (everything that isn’t wkw) should still work (with relative path + with remote path)

### TODOs:
- [x] Unify FileSystemDataVault with other DataVaults, remove FileSystemVaultPath
- [x] Allow exploring locally
- [x] Add config option to whitelist directories in which webknossos may read
  - [x] Use in datastore when reading from data vault
  - [x] Use in webknossos when exploring
- [x] Test that this does not break existing datasource-properties.jsons
  - [x] with local relative path
  - [x] with remote path

### Issues:
- fixes #7387

------


app/models/dataset/credential/CredentialService.scala
```diff
@@ -43,6 +43,8 @@ class CredentialService @Inject()(credentialDAO: CredentialDAO) {
           secret <- credentialSecret
           secretJson <- tryo(Json.parse(secret)).toOption
         } yield GoogleServiceAccountCredential(uri.toString, secretJson, userId.toString, organizationId.toString)
+      case _ =>
+        None
```
fm3, line 47: this was where the match error originated in case of an unknown uri scheme. That shouldn’t happen.

webknossos-datastore/app/com/scalableminds/webknossos/datastore/dataformats/BucketProvider.scala
```diff
@@ -15,19 +14,19 @@ trait BucketProvider extends FoxImplicits with LazyLogging {
   def remoteSourceDescriptorServiceOpt: Option[RemoteSourceDescriptorService]
 
   // To be defined in subclass.
-  def loadFromUnderlying(readInstruction: DataReadInstruction)(implicit ec: ExecutionContext): Fox[DataCubeHandle] =
+  def openShardOrArrayHandle(readInstruction: DataReadInstruction)(implicit ec: ExecutionContext): Fox[DataCubeHandle] =
```
fm3, line 17: just some renamings here

normanrz (approved):

PR #7390, opened 2023-10-16
Fix cursor jumping bug in tree and comment renaming

### LINKED ISSUE #7369: Cursor jumps while typing comments for node
## Context
When adding a comment to a skeleton node, the typing cursor jumps around/to the front of the text input while typing. This makes it extremely hard to type meaningful comments / long texts. Likely a React issues with change events / state and constantly re-rendering. 

I suspect a regression in the `InputComponent`. A recent change was introduced as part of the ["ND"-PR](#7136 ):
https://github.com/scalableminds/webknossos/commit/978dde5b48cce5e19b40ee5055c67782b56da11f#diff-6ad6009d91cd481c4afb4d0f83e258a9a1a61374935cfc36592454b5890d9329

image

See also [Slack](https://scm.slack.com/archives/C5AKLAV0B/p1696001322918929)

## Expected Behavior
I should be able to type a full sentence into the input text field without weird jumps.

## Current Behavior
The text cursor in the input text field makes weird jumps while typing.

## Steps to Reproduce the bug

1. Create a skeleton with at least one node
2. Switch to Comments tab
3. Start typing a comment. Jumpy McJump

## Your Environment for bug

- Browser name and version: Chrome 117.0.5938.149
- Operating System and version: MacOS
- Version of WEBKNOSSOS (Release or Commit): 24973

philippotto:
#7136 removed a workaround for cursor jumping, because I thought that react v16 would have fixed this. During testing, I assumed that all inputs would suffer from this bug, but it turns out that only controlled inputs that store their state in the store (and not in a local state) are affected. I think, this is only true for the tree name and comment inputs which is likely why I missed this back then.
I re-introduced the bug fix and removed the `isFocused` handling, since that assumed that the input's content should not update when the input is changed from the outside while it's focussed. However, the dataset position view supports changing the input via the arrow-keys now which is not compatible with the isFocused check. I think the change should not cause any issues :tm: 

- https://___.webknossos.xyz

### Steps to test:
- create a new tree and try to rename the tree
- create a new node and try to add a comment for it and rename it
- in both cases, do some typical text editing (e.g., also insert something in the middle of the string). the cursor should behave normally (no jumping)
- focus the position input in the navbar and use the arrow keys to increment/decrement the coordinates -> should also work as expected

### Issues:
- fixes #7369

------


daniel-wer (approved):
LGTM :+1:

PR #7394, opened 2023-10-17
Make ad-hoc meshes compatible with ND segmentations

### LINKED ISSUE #7275: Support ad-hoc meshing for ND segmentations and volume annotations
The ad-hoc meshing routes should support specifying the additionalCoordinates for ND volume annotations.

frcroth:
- https://___.webknossos.xyz

@dieknolle3333 I think backend should be completed, though I have not tested it yet.

### Steps to test:
- open an annotation for a ND dataset
- create some segments
- load meshes ad hoc
- make sure all actions for single meshes, groups and multi-selects are working
- change nd-parameter(s) and make sure it's still working.
- when moving around the nth dimension, all loaded meshes should be shown.
- when removing a mesh in one nd-position, it should be gone everywhere.

### TODOs:
- [x] show only meshes of current `additionalCoordinates`
- [x] fix color changing
- [x] fix remove meshes
- [x] fix update meshes
- [x] fix context menu for single mesh in segment view
- [x] fix visibility toggle
- [x] fix segment multiselect
- [x] fix segment context menu in viewport
- [x] implement saga with while-true-loop that saves the previous additional coordinates
- [x] make sure segment view adjusts mesh actions
- [x] move `getAdditionalCoordinatesString` to method, e.g. in flycam_accessor
- [x] find a nicer way to access store in mesh saga
- [x] make sure its working for meshes without additional params
- [x] fix yaml.lock
- [x] fix case where segment in not present in current additional coordianate. see case where I brush new segment and then erase all pixels.
- [x] fix download in group menus
- [x] investiagte whether getMeshesForAdditionalCoordinates can be used in other places
- [x] fix that some meshes arent shown, e.g. after changing coordinates
- [x] only remove mesh in current add coord 

### Issues:
- fixes #7275 

------


dieknolle3333:
When testing the code myself, it occured to me that setting `somePosition` for segments may be a little tricky. E.g. in the segment view tab on the right, clicking on a segment will lead you back to the last position of the segment that was clicked in the viewport, even if the segment has voxels in the current time (/ other add. coord.). 
@philippotto I decided to wait for your review to see what you say. Furthermore I'm not completely sure that this would be in the scope of this issue.

fm3 (commented):
Backend LGTM

I didn’t test this now, as I assume the frontend reviewer will 😇

frontend/javascripts/oxalis/controller/segment_mesh_controller.ts
```diff
@@ -111,53 +158,105 @@ export default class SegmentMeshController {
...
-        this.meshesLODRootGroup.removeLODMesh(meshGroup, lodNumber);
-      } else {
-        this.meshesLODRootGroup.removeNoLODSupportedMesh(meshGroup);
+    for (const additionalCoordinatesString of additionalCoordinatesToRemoveMeshes) {
+      if (
+        this.meshesGroupsPerSegmentationId[additionalCoordinatesString] == null ||
+        this.meshesGroupsPerSegmentationId[additionalCoordinatesString][layerName] == null ||
+        this.meshesGroupsPerSegmentationId[additionalCoordinatesString][layerName][
+          segmentationId
+        ] == null
```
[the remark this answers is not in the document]
philippotto: on a second thought: can we add get/set methods for the `meshesGroupsPerSegmentationId` related stuff in this class? e.g.,
- `this.getMeshGroupsByLOD(additionalCoordiantes, layerName, segmentationId)`
- `this.getMeshGroups(additionalCoordiantes, layerName, segmentationId, lod)`
- setter variants, too, probably

these functions could convert additionalCoordiantes to a string, too.

frontend/javascripts/oxalis/model/reducers/flycam_reducer.ts
```diff
@@ -267,11 +271,34 @@ function FlycamReducer(state: OxalisState, action: Action): OxalisState {
...
+
+      const updatedLocalSegmentationState = update(updatedFlycamState, {
+        localSegmentationData: {
+          [visibleSegmentationLayer.name]: {
+            meshes: {
+              [additionalCoordinateString]: { $set: [] },
+            },
+          },
+        },
+      });
```
philippotto, line 299: I assume that this was added so that access to `localSegmentationData[layerName].meshes[coords][...]` does not fail due to missing initialization. I think, it would be better to not have to rely on this (but instead to change the code that reads from the store so that is robust over missing keys).
You could change the type of the nested localSegmentationData object to something like:

```ts
localSegmentationData: Record<
  string,
  Record<
    string,
    Record<...> | undefined
  > | undefined
>
```

That way, each look up would need to check for undefined (which you already do at most cases, anyway) and typescript should enforce this.

Maybe you could give this a try. If it turns out to be too tedious, I'm also fine with keeping it as is for now :)

frontend/javascripts/oxalis/view/context_menu.tsx
```diff
@@ -1162,7 +1162,8 @@ function ContextMenuInner(propsWithInputRef: Props) {
         contextMenuPosition == null ||
         volumeTracing == null ||
         !hasNoFallbackLayer ||
-        !volumeTracing.hasSegmentIndex
+        !volumeTracing.hasSegmentIndex ||
+        props.additionalCoordinates != null // TODO change once statistics are available for nd-datasets
```
philippotto, line 1166: maybe link to this todo-comment already inside the other PR (and add a checkbox to the GH comment) so that we don't forget about it.

frontend/javascripts/oxalis/view/context_menu.tsx
```diff
@@ -1300,7 +1301,13 @@ function ContextMenuInner(propsWithInputRef: Props) {
     </Tooltip>
   );
 
-  if (hasNoFallbackLayer && volumeTracing?.hasSegmentIndex && isHoveredSegmentOrMesh) {
+  const areSegmentStatisticsAvailable =
+    hasNoFallbackLayer &&
+    volumeTracing?.hasSegmentIndex &&
+    isHoveredSegmentOrMesh &&
+    props.additionalCoordinates == null; // TODO change once statistics are available for nd-datasets
```
philippotto, line 1308: same here

frontend/javascripts/oxalis/view/right-border-tabs/segments_tab/segment_list_item.tsx
```diff
@@ -230,14 +234,17 @@ function _MeshInfoItem(props: {
...
+    dispatch(updateMeshVisibilityAction(layerName, id, isVisible, mesh?.seedAdditionalCoordinates));
   };
 
   const { segment, isSelectedInList, isHovered, mesh } = props;
 
-  if (!mesh) {
+  if (
+    !mesh ||
+    getAdditionalCoordinatesAsString(mesh.seedAdditionalCoordinates || null) !==
+      getAdditionalCoordinatesAsString(Store.getState().flycam.additionalCoordinates)
```
philippotto, line 246: code that renders JSX should not access `Store.getState()`. Please use `useSelector((state: OxalisState) => state.flycam....` at the top of the rendering-function instead.

frontend/javascripts/oxalis/view/right-border-tabs/segments_tab/segments_view.tsx
```diff
@@ -164,12 +167,22 @@ const mapStateToProps = (state: OxalisState): StateProps => {
...
+  let meshesForCurrentAdditionalCoordinates = EMPTY_OBJECT;
+  if (visibleSegmentationLayer != null) {
+    const meshRecords = state.localSegmentationData[visibleSegmentationLayer?.name].meshes;
+    meshesForCurrentAdditionalCoordinates =
+      meshRecords != null &&
+      Object.keys(meshRecords).length > 0 &&
+      Object.keys(meshRecords[addCoordString]).length > 0
+        ? meshRecords[addCoordString]
+        : EMPTY_OBJECT;
+  }
```
philippotto, line 181: I think this could be extracted into an `getMeshesForAdditionalCoordinates()` function that could be put into the volumetracing_accessor.ts module.

frontend/javascripts/oxalis/view/right-border-tabs/segments_tab/segments_view.tsx
```diff
@@ -1069,7 +1088,8 @@ class SegmentsView extends React.Component<Props, State> {
       visibleSegmentationLayer == null ||
       !("fallbackLayer" in visibleSegmentationLayer) ||
       visibleSegmentationLayer.fallbackLayer != null ||
-      !this.props.activeVolumeTracing?.hasSegmentIndex
+      !this.props.activeVolumeTracing?.hasSegmentIndex ||
+      this.props.flycam.additionalCoordinates != null // TODO change once statistics are available for nd-datasets
```
philippotto, line 1092: also link this in the other pr

philippotto (changes requested):
Kudos for pulling through this :) Looking forward to use this soon!
I didn't review the mesh_saga and also haven't tested yet. However, I already left some feedback. I think, we can refactor the code a bit to make it more concise (via getters/setters) and maybe also more robust (by marking the types in the store as `| undefined`; see my one comment for this).

frontend/javascripts/oxalis/controller/segment_mesh_controller.ts
```diff
@@ -84,16 +113,34 @@ export default class SegmentMeshController {
...
+    if (
+      this.meshesGroupsPerSegmentationId[additionalCoordinatesString][layerName][segmentationId][
+        lod
+      ] == null
+    ) {
       const newGroup = new THREE.Group();
-      this.meshesGroupsPerSegmentationId[layerName][segmentationId][lod] = newGroup;
+      this.meshesGroupsPerSegmentationId[additionalCoordinatesString][layerName][segmentationId][
+        lod
+      ] = newGroup;
```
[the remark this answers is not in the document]
dieknolle3333: oh wow, this is really nice, thanks for showing me this.

frontend/javascripts/oxalis/model/sagas/mesh_saga.ts
```diff
@@ -111,9 +118,17 @@ export function isMeshSTL(buffer: ArrayBuffer): boolean {
...
+  layerName: string,
+  segmentId: number,
+  additionalCoordinates?: AdditionalCoordinate[] | null,
+): ThreeDMap<boolean> {
+  let additionalCoordinatesString = getAdditionalCoordinatesAsString(additionalCoordinates || null);
+  adhocMeshesMapByLayer[additionalCoordinatesString] =
+    adhocMeshesMapByLayer[additionalCoordinatesString] || new Map();
+  adhocMeshesMapByLayer[additionalCoordinatesString][layerName] =
+    adhocMeshesMapByLayer[additionalCoordinatesString][layerName] || new Map();
+  const meshesMap = adhocMeshesMapByLayer[additionalCoordinatesString][layerName];
```
philippotto, line 131: similar refactoring can probably be done here as in the mesh controller.

frontend/javascripts/oxalis/model/sagas/mesh_saga.ts
```diff
@@ -494,30 +523,42 @@ function* refreshMeshes(): Saga<void> {
   // By that we avoid to remove cells that got annotated during reloading from the modifiedCells set.
   const currentlyModifiedCells = new Set(modifiedCells);
   modifiedCells.clear();
+
+  const additionalCoordinatesObject = yield* select((state) => state.flycam.additionalCoordinates);
+  const additionalCoordinates = getAdditionalCoordinatesAsString(additionalCoordinatesObject);
```
philippotto, line 528: I'd rename additionalCoordinatesObject to additionalCoordinates, so that it is consistent with the store property.
additionalCoordinates could be renamed to additionalCoordinatesString or additionalCoordinateKey. if you want to be concise additionalCoordKey could be an option.

frontend/javascripts/oxalis/model/sagas/mesh_saga.ts
```diff
@@ -494,30 +523,42 @@ function* refreshMeshes(): Saga<void> {
...
+      threeDMap,
+      segmentationLayer.name,
+      additionalCoordinatesObject,
+    );
   }
 }
 
 function* refreshMesh(action: RefreshMeshAction): Saga<void> {
+  const additionalCoordinatesObject = yield* select((state) => state.flycam.additionalCoordinates);
+  let additionalCoordinates = getAdditionalCoordinatesAsString(additionalCoordinatesObject);
```
philippotto, line 556: same here

frontend/javascripts/oxalis/model/sagas/mesh_saga.ts
```diff
@@ -585,7 +626,11 @@ function* _refreshMeshWithMap(
     );
     shouldBeRemoved = false;
   }
-
+  // see comment in l. 292
```
philippotto, line 629: line numbers can get quickly out of sync. I'd extract the common code into a helper function and put the comment there.

frontend/javascripts/oxalis/model/sagas/mesh_saga.ts
```diff
@@ -1082,33 +1139,105 @@ function* handleRemoveSegment(action: RemoveSegmentAction) {
...
+
+export function* handleAdditionalCoordinateUpdate(): Saga<void> {
+  yield* take("WK_READY");
+
+  let previousAdditionalCoordinates = yield* select((state) => state.flycam.additionalCoordinates);
+
+  while (true) {
+    const action = (yield* take(["SET_ADDITIONAL_COORDINATES"]) as any) as FlycamAction;
+    //satisfy TS
+    if (action.type === "SET_ADDITIONAL_COORDINATES") {
```
philippotto, line 1169:
```
    if (action.type !== "SET_ADDITIONAL_COORDINATES") {
      throw new Error("Unexpected action type")
    }
    // unindent the code from below
```

That way, we avoid one indent-step which makes the code a bit easier to grasp in my opinion.

frontend/javascripts/oxalis/model/sagas/mesh_saga.ts
```diff
@@ -1082,33 +1139,105 @@ function* handleRemoveSegment(action: RemoveSegmentAction) {
...
+      const { segmentMeshController } = yield* call(getSceneController);
```
philippotto, line 1170: this can be done outside of the loop.

frontend/javascripts/oxalis/model/sagas/mesh_saga.ts
```diff
@@ -1082,33 +1139,105 @@ function* handleRemoveSegment(action: RemoveSegmentAction) {
...
+      const meshRecords = yield* call({
+        context: segmentMeshController,
+        fn: () => segmentMeshController.meshesGroupsPerSegmentationId,
+      });
```
philippotto, line 1174: since `meshesGroupsPerSegmentationId` is not a function, but a property, you should be able to access it directly, i think.

```suggestion
      const meshRecords = segmentMeshController.meshesGroupsPerSegmentationId;
```

frontend/javascripts/oxalis/model/sagas/mesh_saga.ts
```diff
@@ -1082,33 +1139,105 @@ function* handleRemoveSegment(action: RemoveSegmentAction) {
...
+
+      if (action.values == null || action.values.length === 0) break;
+      const newAdditionalCoordinates = getAdditionalCoordinatesAsString(action.values);
+
+      for (const additionalCoordinates of [action.values, previousAdditionalCoordinates]) {
+        const currentAdditionalCoordinatesAsString =
+          getAdditionalCoordinatesAsString(additionalCoordinates);
+        const shouldBeVisible = currentAdditionalCoordinatesAsString === newAdditionalCoordinates;
+        const recordsOfLayers = meshRecords[currentAdditionalCoordinatesAsString];
+        if (recordsOfLayers != null) {
```
philippotto, line 1184:
```suggestion
        const recordsOfLayers = meshRecords[currentAdditionalCoordinatesAsString] || {};
        // no if-anymore
        ...
```

due to the `|| {}` the for-loop below will simply execute zero times.

frontend/javascripts/oxalis/model/sagas/mesh_saga.ts
```diff
@@ -1082,33 +1139,105 @@ function* handleRemoveSegment(action: RemoveSegmentAction) {
...
 function* handleMeshVisibilityChange(action: UpdateMeshVisibilityAction): Saga<void> {
-  const { id, visibility, layerName } = action;
+  const { id, visibility, layerName, additionalCoordinates } = action;
   const { segmentMeshController } = yield* call(getSceneController);
-  segmentMeshController.setMeshVisibility(id, visibility, layerName);
+  segmentMeshController.setMeshVisibility(id, visibility, layerName, additionalCoordinates);
+}
+
+export function* handleAdditionalCoordinateUpdate(): Saga<void> {
+  yield* take("WK_READY");
```
philippotto, line 1162: please add a brief comment that explains what the purpose of this saga is.

philippotto (commented):
Here's my feedback for the saga module :)

frontend/javascripts/oxalis/controller/segment_mesh_controller.ts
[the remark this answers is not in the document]
dieknolle3333: AFAICT, the setters are each only needed once, in `addMeshFromGeometry`, and nowhere outside of `segment_mesh_controller`. They have their own methods to set values (`CustomLOD.addNoLODSupportedMesh()`, `CustomLOD.addLODMesh()` and `Record.add()`)  so these functions would be one-liners. 
thus my intuition would be to leave the setters out for now. what do you think?

frontend/javascripts/oxalis/model/reducers/flycam_reducer.ts
dieknolle3333: In this code I tried to make sure that entries for new additional coordinates are added. I moved this code so that new entries are only created if actual meshes are present. Other than that I also changed the types, which really was a little tedious, but the code seems to handle errors better now, so I hope it was worth it. :)

frontend/javascripts/oxalis/api/api_latest.ts
```diff
@@ -2257,8 +2258,15 @@ class DataApi {
...
     ).name;
 
-    if (Store.getState().localSegmentationData[effectiveLayerName].meshes[segmentId] != null) {
+    if (
+      Store.getState().localSegmentationData[effectiveLayerName].meshes != null &&
+      Store.getState().localSegmentationData[effectiveLayerName].meshes![segmentId] != null
+    ) {
       Store.dispatch(updateMeshVisibilityAction(effectiveLayerName, segmentId, isVisible));
+    } else {
+      throw new Error(
```
dieknolle3333, line 2267: create error for both

frontend/javascripts/oxalis/controller/segment_mesh_controller.ts
```diff
@@ -84,80 +102,106 @@ export default class SegmentMeshController {
...
-    }
-    if (this.meshesGroupsPerSegmentationId[layerName][segmentationId] == null) {
-      return;
+  removeMeshById(
+    segmentationId: number,
+    layerName: string,
+    additionalCoordinates: AdditionalCoordinate[] | null | undefined,
+  ): void {
+    // either remove a mesh for specific additional coordinates, or remove all meshes for a segment per default by passing null as additionalCoordinates.
+    let additionalCoordinatesToRemoveMeshes;
```
dieknolle3333, line 140: this is unnecessary because this method is only ever called with the current add. coords -> remove param and get addCoords in method

frontend/javascripts/oxalis/model/reducers/annotation_reducer.ts
```diff
@@ -260,19 +266,36 @@ function AnnotationReducer(state: OxalisState, action: Action): OxalisState {
...
+        stateWithCurrentAddCoords = update(state, {
+          localSegmentationData: {
+            [layerName]: {
+              meshes: {
+                [additionalCoordKey]: { $set: [] },
+              },
+            },
+          },
+        });
+      }
```
philippotto, line 292: wouldn't the same code be needed for the `ADD_PRECOMPUTED_MESH` case? if so, I'd move this into a helper function and call it from both sites.

frontend/javascripts/oxalis/store.ts
```diff
@@ -570,7 +570,7 @@ export type OxalisState = {
   readonly localSegmentationData: Record<
     string, //layerName
     {
-      readonly meshes: Record<string, Record<number, MeshInformation>>; //string represents additional coordinates, number is the segment ID
+      readonly meshes: Record<string, Record<number, MeshInformation> | undefined> | undefined; //string represents additional coordinates, number is the segment ID
```
philippotto, line 573: please add a comment that explains that `undefined` was added to enforce stricter null checks when using these structures.

frontend/javascripts/oxalis/model/sagas/mesh_saga.ts
```diff
@@ -1087,21 +1106,25 @@ function* downloadMeshCellsAsZIP(
...
   try {
     const addFileToZipWriterPromises = segments.map((element) => {
-      const geometry = segmentMeshController.getMeshGeometryInBestLOD(
-        element.segmentId,
-        element.layerName,
-        additionalCoordinates,
-      );
+      if (
+        segmentMeshController.hasMesh(element.segmentId, element.layerName, additionalCoordinates)
+      ) {
```
philippotto, line 1111: why is this check needed? couldn't getMeshGeometryInBestLOD return null if hasMesh is false? currently, the user wouldn't see any toast, if this outer if evaluates to false which is not optimal.

frontend/javascripts/oxalis/model/sagas/mesh_saga.ts
```diff
@@ -1133,23 +1156,27 @@ function* downloadMeshCells(action: TriggerMeshesDownloadAction): Saga<void> {
...
 
 function* handleRemoveSegment(action: RemoveSegmentAction) {
-  // The dispatched action will make sure that the mesh entry is removed from the
-  // store **and** from the scene. Otherwise, the store will still contain a reference
-  // to the mesh even though it's not in the scene, anymore.
-  yield* put(removeMeshAction(action.layerName, action.segmentId));
+  const additionalCoordinates = yield* select((state) => state.flycam.additionalCoordinates);
+  const additionalCoordKey = getAdditionalCoordinatesAsString(additionalCoordinates);
+  const { layerName, segmentId } = action;
+  if (adhocMeshesMapByLayer[additionalCoordKey]?.[layerName]?.get(segmentId) != null) {
```
philippotto, line 1162: is this if needed? ideally, the action could simply be dispatched and if the mesh didn't exist in the first place, nothing would happen (no error either).

philippotto (changes requested):
Great refactoring :tada: The code looks way more concise now :) I left a couple of comments, but mostly only small stuff and some questions for clarifications. The finish line is visible :checkered_flag: :smiley:

frontend/javascripts/oxalis/api/api_latest.ts
```diff
@@ -2257,8 +2258,12 @@ class DataApi {
       layerName,
     ).name;
 
-    if (Store.getState().localSegmentationData[effectiveLayerName].meshes[segmentId] != null) {
+    if (Store.getState().localSegmentationData[effectiveLayerName].meshes?.[segmentId] != null) {
       Store.dispatch(updateMeshVisibilityAction(effectiveLayerName, segmentId, isVisible));
+    } else {
+      throw new Error(
+        `Mesh for segment ${segmentId} was not found in OxalisState.localSegmentationData.`,
```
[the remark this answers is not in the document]
dieknolle3333: The error was a little cryptic before introducing the check. It failed in the annotation reducer, where the update method threw an error about `object` being undefined. So I thought it would be clearer this way.

frontend/javascripts/oxalis/model/sagas/mesh_saga.ts
dieknolle3333: Before this, nothing would be downloaded, because `getMeshGeometryInBestLOD` used to throw an error and nothing was downloaded. So if one of the segments doesn't have a mesh, there should be a toast, and the available should be downloaded, right?

dieknolle3333: I am realizing that I was mixing up bugs. The malfunctioning mesh download still stems from the CSP. I'll add a null check in `getMeshGeometryInBestLOD` and revert the other changes

dieknolle3333:
one last thing... the meshes are always requested for the same position, even if the additional coordinates change. I think sometimes when creating annotation, this can lead to a bug. If I create an annotation and create segments in multiple dimensions, and I load a mesh for segment A in dimension 1, I will have to click on the segment in dimension 2 before being able to load the mesh in that dimension. However, this occurs only in the beginning of creating the dataset and I wasn't able to reproduce it afterwards. It seems to be a separate bug that has to do with the segment's position, so I feel like it's a separate issue.

philippotto:
Excellent work! The code looks very good now and testing went well, too :) :man_dancing: Let's ship it! Kudos for pulling this through the finish line :)

> It seems to be a separate bug that has to do with the segment's position, so I feel like it's a separate issue.

Let's iron it out later then :)

philippotto (approved):

philippotto (changes requested):
Sorry :see_no_evil: I found at least one error when reloading a pre-computed mesh:

image

dieknolle3333:
Good catch, thanks! This didn't occur in my test dataset (`test-agglomerate-file`), but with L4 I was able to reproduce it. 
I think it was only a very small bug and I hope to have fixed it now.

philippotto (approved):
Thank you for fixing this! :smiley: I pushed a tiny commit, because I noticed that two methods had an implicit `any` return type. Feel free to merge now :muscle:

PR #7395, opened 2023-10-17
Allow to compose datasets without transforms or with bigwarp/NML landmarks

### LINKED ISSUE #7269: Import of big warp transformations
## Detailed Description

Big Warp can align datasets with each other using thin-plate-splines. Exporting those transforms produces CSV that wk (or wk-libs) should be able to import.

## Context

philippotto:
image

This PR adds the ability to compose a new dataset from existing dataset layers. There are three possibilities to do this:
- combine datasets without any transforms
- combine datasets with a BigWarp CSV that contains landmarks. Transforms are automatically created.
- combine datasets with two WK NMLs that contain landmarks. Transforms are automatically created.

The user is guided through these options in a wizard. Transforms are only generated for a pair of datasets (i.e., it is not yet possible to compose from more than two datasets).

[Slack discussion](https://scm.slack.com/archives/C5AKLAV0B/p1697184159482769)

- https://composedatasets.webknossos.xyz

### Steps to test:
- use the wizard to test the individual options
- no transforms:
  - simply pick one or more datasets
  - edit the layers afterwards bit
  - create the dataset and view it
- for the other two modes:
  - download [this archive](https://www.notion.so/scalableminds/Test-Datasets-c0563be9c4a4499dae4e16d9b2497cfb?pvs=4#8fb4f9bab6064b3c8ea074e9dc6e4e19).
  - import the datasets by moving the dataset folders into your binaryData folder
  - then compose a new dataset each by providing the landmark files
  - in the CSV case, you need to pass the dataset names manually (`ROI2017_wkw_tiffs_1_wkw` and `ROI2017_wkw_tiffs_2_wkw`)

### Issues:
- fixes #7269

------
- [X] Updated [documentation](../blob/master/docs) if applicable
- [X] Considered [common edge cases](../blob/master/.github/common_edge_cases.md)


webknossos-datastore/app/com/scalableminds/webknossos/datastore/services/ComposeService.scala
```diff
@@ -0,0 +1,169 @@
...
+    sourceName: String,
+    newName: String,
+    transformations: Seq[CoordinateTransformation]
+)
+
+object ComposeLayer {
+  implicit val composeLayerFormat: OFormat[ComposeLayer] = Json.format[ComposeLayer]
+}
+
+class SymlinkHelper(dataSourceService: DataSourceService)(implicit ec: ExecutionContext) extends FoxImplicits {
```
frcroth, line 52: Here a trait with parameters would be nice (scala 3 feature). I could also remove this file and merge the content into UploadService, which would elimate this common trait/class.

frontend/javascripts/admin/dataset/dataset_add_compose_view.tsx
```diff
@@ -0,0 +1,423 @@
...
+    setIsLoading(true);
+    try {
+      await createDatasetComposition(datastoreToUse.url, {
+        newDatasetName,
+        targetFolderId: form.getFieldValue(["targetFolderId"]),
+        organizationName: "sample_organization",
+        scale: linkedDatasets[1].dataSource.scale,
+        layers,
+      });
+      await Utils.sleep(3000); // wait for 3 seconds so the server can catch up / do its thing
```
philippotto, line 201: @frcroth This code is similar to what is already happening after a dataset upload. After the dataset is imported, the user is asked whether they want to view the dataset. Without the delay, the dataset is usually "not there yet" and cannot be found. For the new compose case, I ran into the 404 error even with this 3s delay. I don't know what exactly causes this delay (maybe the communication with the datastore who reports things?).
Ideally, the upload/compose/add route would only finish once the dataset can also be found when viewing it. Maybe you could have a look at that? Don't know whether it's complicated.

cc @fm3

fm3: That should be possible, let me know if you need help debugging what’s going on there.

philippotto: nice, seems to work now :tada:

frontend/javascripts/admin/dataset/dataset_add_view.tsx
```diff
@@ -100,11 +110,12 @@ function DatasetAddView({ history }: RouteComponentProps) {
...
   const defaultActiveTabFromHash = location.hash.substring(1);
-  const defaultActiveKey = Object.values(DatasetAddViewTabs).includes(
-    defaultActiveTabFromHash as DatasetAddViewTabs,
+  const defaultActiveKey = Object.values(DatasetAddType).includes(
+    defaultActiveTabFromHash as DatasetAddType,
   )
-    ? (defaultActiveTabFromHash as DatasetAddViewTabs)
-    : DatasetAddViewTabs.UPLOAD;
+    ? (defaultActiveTabFromHash as DatasetAddType)
+    : // todo: revert
```
philippotto, line 117: todo

frontend/javascripts/admin/dataset/dataset_components.tsx
```diff
@@ -138,3 +142,60 @@ export function DatastoreFormItem({
     </FormItem>
   );
 }
+
+export function AllowedTeamsFormItem({
```
philippotto, line 146: this code was simply moved here.

frontend/javascripts/admin/dataset/dataset_add_compose_view.tsx
philippotto: @frcroth Could you also add the `dataSourceService.checkInbox` code once the `/data/datasets/finishUpload` route finishes? Then, I can also remove the sleep in the upload case.

frontend/javascripts/admin/dataset/dataset_add_view.tsx
```diff
@@ -124,8 +140,28 @@ function DatasetAddView({ history }: RouteComponentProps) {
...
+          datastores={datastores}
+          onAdded={handleDatasetAdded.bind(null, DatasetAddType.REMOTE)}
+        />
+      ),
+    },
+    {
+      label: (
+        <span>
+          <CopyOutlined />
+          Compose from existing datasets
```
daniel-wer, line 155: The other tabs use title case

frontend/javascripts/admin/dataset/dataset_add_compose_view.tsx
```diff
@@ -0,0 +1,58 @@
...
+    datasets: [],
+    sourcePoints: [],
+    targetPoints: [],
+  });
+  const { currentWizardStep } = wizardContext;
+  const CurrentWizardComponent = WIZARD_STEPS[currentWizardStep].component;
+
+  return (
+    <div style={{ padding: 5 }}>
+      <CardContainer title="Compose a dataset from existing dataset layers">
```
daniel-wer, line 48: In the other tabs, title case is used in this spot

frontend/javascripts/admin/dataset/composition_wizard/02_upload_files.tsx
```diff
@@ -0,0 +1,192 @@
...
+      }
+      setWizardContext((oldContext) => ({
+        ...oldContext,
+        ...newContextPartial,
+      }));
+    } catch (exception) {
+      if (exception instanceof SoftError) {
+        Toast.warning(exception.message);
+      } else {
+        Toast.error(
```
daniel-wer, line 49: Should this case be reported to airbrake as well?

frontend/javascripts/admin/dataset/composition_wizard/02_upload_files.tsx
```diff
@@ -0,0 +1,192 @@
...
+      }
+    }
+  };
+
+  return (
+    <div>
+      <p>
+        {wizardContext.composeMode === "BIG_WARP"
+          ? "Please upload one CSV file that was exported by BigWarp."
+          : "Please upload two NML files that contain landmarks that you created with WEBKNOSSOS."}
```
daniel-wer, line 62: At least for me it would be helpful to get more information on how the two NML files should be structured. One or multiple trees, which nodes are matched?

frontend/javascripts/admin/dataset/composition_wizard/02_upload_files.tsx
```diff
@@ -0,0 +1,192 @@
...
+        <p>
+          Landmark files ({wizardContext.composeMode === "BIG_WARP" ? "1 CSV file" : "2 NML files"}
+          ):
+        </p>
+        <Upload.Dragger
+          name="files"
+          fileList={fileList}
+          onChange={handleChange}
+          beforeUpload={() => false}
+          maxCount={2}
```
daniel-wer, line 74:
```suggestion
          maxCount={wizardContext.composeMode === "BIG_WARP" ? 1 : 2}
```

frontend/javascripts/admin/dataset/composition_wizard/02_upload_files.tsx
```diff
@@ -0,0 +1,192 @@
...
+  const { trees: trees1, datasetName: datasetName1 } = await parseNml(nmlString1);
+  const { trees: trees2, datasetName: datasetName2 } = await parseNml(nmlString2);
+
+  if (!datasetName1 || !datasetName2) {
+    throw new SoftError("Could not extract dataset names.");
+  }
+
+  const nodes1 = Array.from(
+    values(trees1)
+      .map((tree) => Array.from(tree.nodes.values())[0])
```
daniel-wer, line 160: This reads like only the first node of each tree is used. It would be user-friendly to show a warning once a tree with more than one node is encountered.

frontend/javascripts/admin/dataset/composition_wizard/02_upload_files.tsx
```diff
@@ -0,0 +1,192 @@
...
+      .values(),
+  );
+  const nodes2 = Array.from(
+    values(trees2)
+      .map((tree) => Array.from(tree.nodes.values())[0])
```
daniel-wer, line 165: See my comment above.

frontend/javascripts/admin/dataset/composition_wizard/03_select_datasets.tsx
```diff
@@ -0,0 +1,70 @@
...
+  useEffect(() => {
+    setDatasetValues(wizardContext.datasets.map((ds) => ({ value: ds.name, label: ds.name })));
+  }, []);
+
+  // When not using any transforms,
+  let isDatasetCountValid = true;
+  if (wizardContext.composeMode === "WITHOUT_TRANSFORMS") {
+    isDatasetCountValid = datasetValues.length > 0;
+  } else {
+    isDatasetCountValid = datasetValues.length === 2;
```
daniel-wer, line 45: Do you think a use case could also be to transform two layers of the same dataset using BigWarp or wk NMLs?

frontend/javascripts/admin/dataset/composition_wizard/04_configure_new_dataset.tsx
```diff
@@ -0,0 +1,285 @@
...
+    form.setFieldsValue({ layers: newLinks });
+  };
+
+  useEffect(() => {
+    handleTransformImport(wizardContext.sourcePoints, wizardContext.targetPoints);
+  }, []);
+
+  const handleSubmit = async () => {
+    if (activeUser == null) {
+      throw new Error("Cannot upload dataset without being logged in.");
```
daniel-wer, line 88:
```suggestion
      throw new Error("Cannot create dataset without being logged in.");
```

frontend/javascripts/admin/dataset/composition_wizard/04_configure_new_dataset.tsx
```diff
@@ -0,0 +1,285 @@
...
+    }
+    const layers = form.getFieldValue(["layers"]);
+
+    const uploadableDatastores = props.datastores.filter((datastore) => datastore.allowsUpload);
+    const datastoreToUse = uploadableDatastores[0];
+    if (!datastoreToUse) {
+      Toast.error("Could not find datastore that allows uploading.");
+      return;
+    }
```
daniel-wer, line 97: Is this correct? Can datasets only be created (using this wizard) on data stores that allow upload?

frontend/javascripts/admin/dataset/composition_wizard/04_configure_new_dataset.tsx
```diff
@@ -0,0 +1,285 @@
...
+
+    const newDatasetName = form.getFieldValue(["name"]);
+    setIsLoading(true);
+    try {
+      await createDatasetComposition(datastoreToUse.url, {
+        newDatasetName,
+        targetFolderId: form.getFieldValue(["targetFolderId"]),
+        organizationName: "sample_organization",
```
daniel-wer, line 105: This should probably be the user's orga

frontend/javascripts/admin/dataset/composition_wizard/04_configure_new_dataset.tsx
```diff
@@ -0,0 +1,285 @@
...
+        scale: linkedDatasets[1].dataSource.scale,
```
daniel-wer, line 106: From this line and line 76 I get that dataset[0] is transformed to the coordinate system of dataset[1]. Should this be pointed out somewhere or do you think it's irrelevant?

frontend/javascripts/admin/dataset/composition_wizard/04_configure_new_dataset.tsx
```diff
@@ -0,0 +1,285 @@
...
+      <Form form={form} layout="vertical" onFinish={handleSubmit}>
+        <Row gutter={8}>
+          <Col span={12}>
+            <DatasetNameFormItem activeUser={activeUser} />
+          </Col>
+          <Col span={12}>
+            <AllowedTeamsFormItem
+              isDatasetManagerOrAdmin={isDatasetManagerOrAdmin}
+              selectedTeams={selectedTeams}
+              setSelectedTeams={(selectedTeams) => setSelectedTeams(selectedTeams)}
```
daniel-wer, line 129:
```suggestion
              setSelectedTeams={setSelectedTeams}
```

frontend/javascripts/admin/dataset/composition_wizard/04_configure_new_dataset.tsx
```diff
@@ -0,0 +1,285 @@
...
+              formRef={formRef}
+            />
+          </Col>
+        </Row>
+
+        <FormItemWithInfo
+          name="targetFolderId"
+          label="Target Folder"
+          info="The folder into which the dataset will be uploaded. The dataset can be moved after upload. Note that teams that have access to the specified folder will be able to see the uploaded dataset."
```
daniel-wer, line 138:
```suggestion
          info="The folder in which the dataset will be created. The dataset can be moved after creation. Note that teams that have access to the specified folder will be able to see the created dataset."
```

frontend/javascripts/admin/dataset/composition_wizard/04_configure_new_dataset.tsx
```diff
@@ -0,0 +1,285 @@
...
+
+        <FormItem
+          style={{
+            marginBottom: 0,
+          }}
+        >
+          <Button onClick={onPrev}>Back</Button>
+
+          <Button type="primary" htmlType="submit" loading={isLoading} style={{ marginLeft: 8 }}>
+            Upload
```
daniel-wer, line 191:
```suggestion
            Create Dataset
```

daniel-wer (commented):
Well done! The code mostly looks very good and was easy to read :+1: I got stuck during testing and will continue once that's fixed.

- I cannot progress further than the dataset selection. I chose no transforms in the first step and then regardless of which datasets I select, I get the message that the datasets don't exist.
not_find_datasets

- The dataset selection behaves a little un-intuitive: (1) When first clicking into the selection, no dataset names are shown. (2) After typing, results that match the typed string are shown. (3) After deleting the typed string, some but not all datasets are shown (I think all of the datasets that matched a typed string at some point). - (2) is good, but for (1) and (3) maybe a list of all available dataset names could be shown?

webknossos-datastore/app/com/scalableminds/webknossos/datastore/controllers/DataSourceController.scala
```diff
@@ -144,6 +145,7 @@ class DataSourceController @Inject()(
                                                       urlOrHeaderToken(token, request)) {
             for {
               (dataSourceId, datasetSizeBytes) <- uploadService.finishUpload(request.body)
+              _ <- dataSourceService.checkInbox(false)
```
fm3, line 148: It should not be necessary here to do a full scan of the inbox. Instead, we should report the newly created dataset to the wk side individually. 

And in fact, I think this already happens in uploadService.finishUpload (it calls `dataSourceRepository.updateDataSource(dataSource)` which does exactly that)

webknossos-datastore/app/com/scalableminds/webknossos/datastore/controllers/DataSourceController.scala
```diff
@@ -439,6 +441,22 @@ class DataSourceController @Inject()(
...
+      val userToken = urlOrHeaderToken(token, request)
+      accessTokenService.validateAccess(UserAccessRequest.administrateDataSources(request.body.organizationName), token) {
+        for {
+          _ <- Fox.serialCombined(request.body.layers.map(_.datasetId).toList)(
+            id =>
+              accessTokenService.assertUserAccess(
+                UserAccessRequest.readDataSources(DataSourceId(id.name, id.owningOrganization)),
+                userToken))
+          _ <- composeService.composeDataset(request.body, userToken)
+          _ <- dataSourceService.checkInbox(verbose = false)
```
fm3, line 455: this should also use `dataSourceRepository.updateDataSource(dataSource)` instead of `checkInbox`. (either return the dataSource out of the composeService or call that in there)

webknossos-datastore/app/com/scalableminds/webknossos/datastore/services/ComposeService.scala
fm3: Could it be an injectable service? Then it could hold a proper dependency on the DataSourceService and be usable like the others?

webknossos-datastore/app/com/scalableminds/webknossos/datastore/services/ComposeService.scala
```diff
@@ -0,0 +1,177 @@
...
+    targetFolderId: String,
+    organizationName: String,
+    scale: Vec3Double,
+    layers: Seq[ComposeLayer]
+)
+
+object ComposeRequest {
+  implicit val composeRequestFormat: OFormat[ComposeRequest] = Json.format[ComposeRequest]
+}
+case class ComposeLayer(
```
fm3, line 40:
```suggestion
case class ComposeRequestLayer(
```

(Maybe this would make it clearer that this class is not an intermediate representation of a whole layer, but is just used in the request to identify one and specify what is to happen with it)

fm3 (commented):
Backend looking good already :) Just added a few small comments.

I did not test yet, let’s see if Daniels test case will be resolved

frontend/javascripts/admin/dataset/composition_wizard/04_configure_new_dataset.tsx
philippotto: I think so, because the datastore will create a new folder with symlinks. This is probably only possible if uploads are allowed. Maybe @fm3 can confirm this?

frontend/javascripts/admin/dataset/composition_wizard/03_select_datasets.tsx
philippotto: I think, this usecase is imaginable, but not the default one. One could argue that two layers should only be put into the same DS if they are aligned with each other. If they still need alignment, they should be put into separate datasets and then be combined using this new workflow. Of course, it will happen that this is not the case, but I wouldn't want to design for this (at least not in this iteration).

frontend/javascripts/admin/dataset/composition_wizard/04_configure_new_dataset.tsx
fm3: Kinda. The datastore at least needs write access on its file system. This is usually the case for those datastores with allowUpload=true. There may be cases where this is too strict (some datastores could allow this, but we don’t want users to upload data there), but let’s introduce more fine-grained logic when we find that it is needed.

philippotto:
> * I cannot progress further than the dataset selection. I chose no transforms in the first step and then regardless of which datasets I select, I get the message that the datasets don't exist.
>   not_find_datasets

Hmm, this is weird. It works for me. Can you check the network tab and understand why the dataset can't be found? Happy to debug this together in a call :call_me_hand: 

> * The dataset selection behaves a little un-intuitive: (1) When first clicking into the selection, no dataset names are shown. (2) After typing, results that match the typed string are shown. (3) After deleting the typed string, some but not all datasets are shown (I think all of the datasets that matched a typed string at some point). - (2) is good, but for (1) and (3) maybe a list of all available dataset names could be shown?

I made the component an async one to avoid that all datasets have to be fetched. Instead, the same search api is used as in the dashboard. To reduce the confusion I changed two things:

- I adapted the placeholder to read: `Type to search and select dataset(s)...`
- I added code that clears the suggestions after selecting an entry. This means that clicking into the field won't show old suggestions. Thus, the user shouldn't assume that these are all available datasets. Instead, they hopefully type again.

It's not perfect, but I can't think of a better way right now without fetching possibly thousands of datasets in this UI. I hope this is okay?

frontend/javascripts/admin/dataset/composition_wizard/04_configure_new_dataset.tsx
philippotto: Good point. I added this info into the "Upload files" view as this is the place where the user might want to influence this be re-ordering the files (or the contents).

webknossos-datastore/app/com/scalableminds/webknossos/datastore/controllers/DataSourceController.scala
frcroth: @philippotto You requested this change, did this do anything for you?

webknossos-datastore/app/com/scalableminds/webknossos/datastore/services/ComposeService.scala
frcroth: Move to another file?

webknossos-datastore/app/com/scalableminds/webknossos/datastore/controllers/DataSourceController.scala
philippotto: The front-end code in `master` has a 3 seconds delay after a DS upload which I wanted to get rid of. After you did the change here, I removed the delay and uploading worked fine. However, it could be that the delay already was superfluous for some time. I temporarily removed the line locally and a test upload worked for me without problems.
So, if you think, that this line doesn't do anything and a finished upload-request implies that the DS is available (no race conditions exist), we can remove `checkInbox` again and keep the front-end-delay removed, too.
If you are not confident, I'd keep the 3s delay. I definitely want to avoid that a user uploads something, clicks on "view" (in a success-modal; not in the dashboard) and gets greeted with a 404.

webknossos-datastore/app/com/scalableminds/webknossos/datastore/services/ComposeService.scala
fm3: sounds good to me :)

frcroth: Did something a little more radical :wink:

fm3 (commented):
Backend LGTM :)

daniel-wer:
@philippotto Thanks for addressing my feedback! The added format instructions will be very helpful to users I would imagine.

> Hmm, this is weird. It works for me. Can you check the network tab and understand why the dataset can't be found? Happy to debug this together in a call 🤙

This turned out to be an sql issue of the dev deployment. I wrongly assumed I was the first to deploy the branch at the time, but it was deployed earlier and the schemaVersion became incompatible. After a re-deployment, this works well :+1: 

> I added code that clears the suggestions after selecting an entry. This means that clicking into the field won't show old suggestions. Thus, the user shouldn't assume that these are all available datasets. Instead, they hopefully type again.
It's not perfect, but I can't think of a better way right now without fetching possibly thousands of datasets in this UI. I hope this is okay?

This mostly works well and addresses my issue, but a small nuisance now is that typing and then clicking anywhere but on a dataset entry will clear the search box. I'm not sure whether that's important, but if you see a quick fix, feel free.

One optional suggestion would be to use toggles to disable layers in the layer selection screen, because it would allow users to change their mind (going back one screen and next again, resets all values including the dataset name).

@fm3 Feel free to test as well and/or approve if you are happy with this PR :)

philippotto:
> This mostly works well and addresses my issue, but a small nuisance now is that typing and then clicking anywhere but on a dataset entry will clear the search box. I'm not sure whether that's important, but if you see a quick fix, feel free.

Hmm, this seems to be the default behavior in the select component from antd (see [doc examples](https://4x.ant.design/components/select/)). I didn't find a way to avoid this (I tried `autoClearSearchValue` and `onClear`). So, there's not much we can do for now, I'm afraid.

> One optional suggestion would be to use toggles to disable layers in the layer selection screen, because it would allow users to change their mind (going back one screen and next again, resets all values including the dataset name).

Good idea! However, I'd like to postpone this to another iteration :) We will certainly get more feedback when this feature is available in production and then we can priotize.

frontend/javascripts/admin/dataset/composition_wizard/02_upload_files.tsx
philippotto: I changed this so that all nodes are taken into account and that it is asserted that all tree pairs have the same amount nodes (this is done by checking that no pair is null during the `zip` iteration). I think, this is way more intuitive and flexible for users.

fm3 (approved):
Works for me :)

I think the first page of the wizard is a bit cumbersome. It contains three explanations, then a radio group with three buttons that mirror those, and then a next button. Instead it could be three clickable cards, where each starts one of the workflows, and which contain the explanations. To my mind that would be a way clearer UI.

However, this does not need to block this PR

fm3:
@frcroth One more question: I see that the new case class `DataLayerId` used inside of `ComposeRequestLayer` has `name` and `owningOrganization`. Isn’t this the same as DataSourceId? Should that just be used instead? I think this thing shouldn’t be called `DataLayerId` as it does not actually identify a layer, but rather a dataset. Maybe another way could be found here

fm3:
Definitely better :) :shipit:

philippotto:
> @frcroth One more question: I see that the new case class `DataLayerId` used inside of `ComposeRequestLayer` has `name` and `owningOrganization`. Isn’t this the same as DataSourceId? Should that just be used instead? I think this thing shouldn’t be called `DataLayerId` as it does not actually identify a layer, but rather a dataset. Maybe another way could be found here

Should I wait with merging this PR?

fm3:
you’re right, I now wrote https://github.com/scalableminds/webknossos/issues/7560

PR #7397, opened 2023-10-17
Allow uploading zarr datasets

### LINKED ISSUE #6205: Allow uploading ZARR files
One of the following would probably make sense to have for better zarr support:

- Support arbitrary zarr files (probably should be converted with wkcuber so that a mag hierarchy and meta data exist)
- Support zarr files which already follow our format

frcroth:
- https://___.webknossos.xyz

### Steps to test:
- Upload zarr/ngff datasets as zip
- Upload a zip containing multiple zarr layers

### TODOs:
- [x] Upload of multiple layers in the same zip
- [x] Frontend

### Issues:
- fixes #6205 

------


app/controllers/DatasetController.scala
```diff
@@ -74,6 +75,7 @@ class DatasetController @Inject()(userService: UserService,
                                   analyticsService: AnalyticsService,
                                   mailchimpClient: MailchimpClient,
                                   exploreRemoteLayerService: ExploreRemoteLayerService,
+                                  exploreLayerService: ExploreLayerService,
```
fm3, line 78: seems to be unused in DatasetController

webknossos-datastore/app/com/scalableminds/webknossos/datastore/services/UploadService.scala
```diff
@@ -224,12 +228,50 @@ class UploadService @Inject()(dataSourceRepository: DataSourceRepository,
...
       for {
-        _ <- addLayerAndResolutionDirIfMissing(unpackToDir).toFox
-        _ <- addSymlinksToOtherDatasetLayers(unpackToDir, layersToLink.getOrElse(List.empty))
-        _ <- addLinkedLayersToDataSourceProperties(unpackToDir, dataSourceId.team, layersToLink.getOrElse(List.empty))
+        isZarr <- looksLikeZarrArray(unpackToDir).toFox
+
+        _ <- Fox.runIf(isZarr)(exploreLocalDatasource(unpackToDir, dataSourceId))
+        p <- Fox.runIf(!isZarr)(tryExploringMultipleZarrLayers(unpackToDir, dataSourceId))
+        _ <- Fox.runIf(!isZarr && p.flatten.isEmpty)(
+          postProcessUploadedWKWDataSource(unpackToDir, dataSourceId, layersToLink))
```
fm3, line 236: It would be great if the layersToLink feature would also work with uploaded zarr datasets. This adds entries to layers from other datasets as specified by the layersToLink parameter

util/src/main/scala/com/scalableminds/util/tools/Fox.scala
```diff
@@ -242,6 +242,23 @@ object Fox extends FoxImplicits {
 
     failure.msg + formatStackTrace(failure) + formatChain(failure.chain)
   }
+
+  def firstSuccess[T](foxes: Seq[Fox[T]])(implicit ec: ExecutionContext): Fox[T] = {
```
fm3, line 246: :+1:

webknossos-datastore/app/com/scalableminds/webknossos/datastore/services/UploadService.scala
```diff
@@ -224,12 +228,50 @@ class UploadService @Inject()(dataSourceRepository: DataSourceRepository,
       Fox.successful(())
     else {
       for {
-        _ <- addLayerAndResolutionDirIfMissing(unpackToDir).toFox
-        _ <- addSymlinksToOtherDatasetLayers(unpackToDir, layersToLink.getOrElse(List.empty))
-        _ <- addLinkedLayersToDataSourceProperties(unpackToDir, dataSourceId.team, layersToLink.getOrElse(List.empty))
+        isZarr <- looksLikeZarrArray(unpackToDir).toFox
```
fm3, line 231: Do you think we could detect existing datasource-properties.jsons and in this case skip the explorer (or possibly, just make the paths relative)?

I tried uploading the sharded zarr3 dataset from [here](https://www.notion.so/scalableminds/Test-Datasets-c0563be9c4a4499dae4e16d9b2497cfb?pvs=4#fea01b638d2c4c689565adb72a79630a), which webknossos can just serve as is. However, the upload was rejected. Since this (or something close) is to become our standard format, which will also be created by the wkcuber, uploading this should also work.

fm3 (changes requested):
Cool stuff, code looks pretty good to me, and I tested a bunch, and most cases worked well. I added a couple of comments, especially about datasets with an existing datasource-properties.json file

fm3 (commented):
Looking good :) Please see my question about guessed WKW vs Zarr

fm3 (approved):
Nice trick with the maxDepth ;) (could you call it by name, with `maxDepth=2` though? That way we may stumble less when reading this later)

Works for me!

PR #7404, opened 2023-10-24
Fix mag restrictions when task type uses not-existing mags

philippotto:
Instead of trying to look up a resolution by the provided mag spec (e.g., task type defines max=512 and then it is assumed that mag 512 exists), the logarithm of the mag spec is used to index into the known zoom ranges. 

- https://___.webknossos.xyz

### Steps to test:
- create a new task type with mag restrictions (e.g., 2 to 512)
- create a task for a dataset that does not have mags until 512
- get the task and ensure it doesn't crash
- zoom in to mag 1 -> a toast should appear to warn you that the mag restriction is violated

### Issues:
- fixes https://scm.slack.com/archives/C5AKLAV0B/p1698139256299479

------


daniel-wer (approved):
Nice, works well :+1:

PR #7410, opened 2023-10-25
Speed up annotation list

### LINKED ISSUE #6779: Speed up Annotation List Request
Similar to the dataset list request that is being tackled in #6759 the Annotation list request could also be sped up by moving some of its logic to SQL.

However, there the priority for this change is a bit lower, since the annotation list is paginated and loading one page currently takes about 3s, as opposed to the 20s we saw for some dataset list requests.

frcroth:
- https://___.webknossos.xyz

### Steps to test:
- test the annotations list in the dashboard
- view an annotation
- download an annotation from the list
- some monkey-testing

### TODOs:
- [X] adapt frontend @philippotto

### Issues:
- fixes #6779 

------


philippotto:
@frcroth In the front-end, we already distinguished between the types [`APIAnnotation`](https://github.com/scalableminds/webknossos/blob/43af02da93e0d003e2d1bbcd68961f20d7ddc687/frontend/javascripts/types/api_flow_types.ts#L507C24-L507C24) (e.g., returned by `/api/annotations/${annotationId}/info`) and [`APIAnnotationCompact`](https://github.com/scalableminds/webknossos/blob/43af02da93e0d003e2d1bbcd68961f20d7ddc687/frontend/javascripts/types/api_flow_types.ts#L423) (e.g., returned by `/api/users/${userId}/annotations`).

How is this distinguished/named in the back-end? Should I handle it like this in the front-end?
- keep `APIAnnotation`
- rename `APIAnnotationCompact` to `APIAnnotationInfo`
- add `APIAnnotationInfoCompact`

Or am I getting this wrong?

frcroth:
Yes, it seems there are now 3 degrees of compactness, full, somewhat compact from list and compact from list query with compact query set.


>     * keep `APIAnnotation`
> 
>     * rename `APIAnnotationCompact` to `APIAnnotationInfo`
> 
>     * add `APIAnnotationInfoCompact`

This seems sensible to me.

philippotto:
Also, `statistics` should be named `stats` (as it's done in `APIAnnotationInfo`).

philippotto:
And I'm afraid, we need the `tags` property, as well. The good news is that `formattedHash` seems to be unused in the front-end. So, that could be removed altogether.

philippotto:
Sorry for the peu á peu messaging. `state` is also required. I hope this is the last missing piece.

Also, when downloading an annotation, we currently check whether the annotation has a volume layer. This information could be found out via an additional request, though. So, no need to change anything for now.

frcroth:
@philippotto I added state, tags and the correct teams. Is the non-compact list still useful? otherwise I could also remove that code path.

philippotto:
> @philippotto I renamed stats and added state, tags and the correct teams.

Great :+1: I integrated it and the dashboard works now.

> Is the non-compact list still useful? otherwise I could also remove that code path.

Yes, it can be removed and I think we should do this to reduce complexity. These routes should then return the compact annotation version, too:

- /api/user/annotations
- /api/users/${userId}/annotations

After removal, the compact GET parameter shouldn't exist, anymore.

frcroth:
@philippotto The annotation list query returns a "X-Total-Count" header when not compact. Is this used?

philippotto:
> Yes, it can be removed and I think we should do this to reduce complexity.

We need to check compatibility with wklibs, too. @fm3 can give insights probably, since he just reworked the client in wklibs.

fm3:
Good point! That’s true, this would be a breaking change for the libs. https://github.com/scalableminds/webknossos-libs/pull/948 makes the libs use versioned api routes. This means that we can support the old case by supplying the older version in the LegacyApiController. However, many users might not yet have adopted to that new libs version. Let’s wait some time before introducing this kind of breaking change. I’d say in this PR it would be best to have this be opt-in only.

frcroth:
> Good point! That’s true, this would be a breaking change for the libs. [scalableminds/webknossos-libs#948](https://github.com/scalableminds/webknossos-libs/pull/948) makes the libs use versioned api routes. This means that we can support the old case by supplying the older version in the LegacyApiController. However, many users might not yet have adopted to that new libs version. Let’s wait some time before introducing this kind of breaking change. I’d say in this PR it would be best to have this be opt-in only.

I think it wouldn't be difficult to adapt this so that everything that was returned in the previous non-compact way is now also part of the compact path (and still use only one query). In that case, the old way could be removed, couldn't it?

fm3:
sounds fair, yes

philippotto:
> I think it wouldn't be difficult to adapt this so that everything that was returned in the previous non-compact way is now also part of the compact path (and still use only one query). In that case, the old way could be removed, couldn't it?
I think the only relevant things missing are annotation layers and org.

So, you would re-add annotation layers and org? Do you have any benchmarks for the speed improvement introduced by this PR? If the annotation layers and org are not a considerable hit (thanks to a single query), I'm fine with it. If so, I would revert my front-end changes, since the interface wouldn't change at all, right?.

philippotto:
> @philippotto For my small test instance the non-compact route takes 35ms while the compact call takes 15ms.

I'm not sure to what exactly these numbers refer. Is it single-sql-statement-with-layers-and-org VS single-sql-statement-without-layers-and-org ? This would be the comparison I'm interested in to decide whether we should simply keep layers and org in it.
Or does the non-compact route mean "the old implementation with multiple sql queries"?

frcroth:
> I'm not sure to what exactly these numbers refer. Is it single-sql-statement-with-layers-and-org VS single-sql-statement-without-layers-and-org ? This would be the comparison I'm interested in to decide whether we should simply keep layers and org in it. Or does the non-compact route mean "the old implementation with multiple sql queries"?

It is old implementation with multiple sql queries vs new with only one

philippotto:
Ok, then, I'd suggest to do the comparison between single-sql-statement-with-layers-and-org and single-sql-statement-without-layers-and-org to decide whether it's okay to always include layers and organization.

frcroth:
> Ok, then, I'd suggest to do the comparison between single-sql-statement-with-layers-and-org and single-sql-statement-without-layers-and-org to decide whether it's okay to always include layers and organization.

Just tested it, no measurable difference for my test instance.

frcroth:
> > > Ok, then, I'd suggest to do the comparison between single-sql-statement-with-layers-and-org and single-sql-statement-without-layers-and-org to decide whether it's okay to always include layers and organization.
> > 
> > 
> > Just tested it, no measurable difference for my test instance (of course I only have one organization etc).
> 
> Ok, great 👍 Then, let's keep it simple and just go with one type. I cleaned up the front-end and undid the unnecessary changes. I took the opportunity and renamed the APIAnnotationCompact type to APIAnnotationInfo to match the back-end's nomenclature better. Let me know in case you need something else from the front-end.

Nice! I think you could assign someone for frontend review then if you believe the changes are non-trivial.

app/models/annotation/Annotation.scala
```diff
@@ -268,22 +298,120 @@ class AnnotationDAO @Inject()(sqlClient: SqlClient, annotationLayerDAO: Annotati
...
-    for {
-      accessQuery <- accessQueryFromAccessQ(listAccessQ)
-      r <- run(q"""select $columns from $existingCollectionName
-                   where typ = ${AnnotationType.Explorational} and $stateQuery and $accessQuery
-                   order by _id desc limit $limit offset ${pageNumber * limit}""".as[AnnotationsRow])
-      parsed <- parseAll(r)
-    } yield parsed
-  }
+  private def parseObjectIdArray(objectIdArray: String): Seq[ObjectId] =
+    Option(objectIdArray).map(_.split(",").map(id => ObjectId(id))).getOrElse(Array[ObjectId]()).toSeq
```
fm3, line 302: Could this use `ObjectId.fromString`? That one also checks valid value.

Also, why the `Option` wrapper? When would this become None?

app/models/annotation/Annotation.scala
```diff
@@ -268,22 +298,120 @@ class AnnotationDAO @Inject()(sqlClient: SqlClient, annotationLayerDAO: Annotati
...
+      query = q"""
+          SELECT
+          a._id,
+          a.name,
+          a.description,
+          a._user,
+          u.firstname,
+          u.lastname,
+          a.othersmayedit,
+          string_agg(t._id, ',') as team_ids,
```
fm3, line 325:
```suggestion
          STRING_AGG(t._id, ',') AD team_ids,
```
(if it also works like that, I’d prefer all SQL features to be uppercase for easier distinguishability)

app/models/annotation/Annotation.scala
```diff
@@ -268,22 +298,120 @@ class AnnotationDAO @Inject()(sqlClient: SqlClient, annotationLayerDAO: Annotati
...
+          AnnotationCompactInfo(
+            id = r._1,
+            name = r._2,
+            description = r._3,
+            ownerId = r._4,
+            ownerFirstName = r._5,
+            ownerLastName = r._6,
+            othersMayEdit = r._7,
+            teamIds = parseObjectIdArray(r._8),
+            teamNames = Option(r._9).map(_.split(",")).getOrElse(Array[String]()).toSeq,
```
fm3, line 393: Again I don’t understand the Option wrapper

app/models/annotation/AnnotationService.scala
```diff
@@ -966,39 +964,56 @@ class AnnotationService @Inject()(
...
+    val tracingType: String = {
+      val skeletonPresent = annotationInfo.annotationLayerTypes.contains(AnnotationLayerType.Skeleton.toString)
+      val volumePresent = annotationInfo.annotationLayerTypes.contains(AnnotationLayerType.Volume.toString)
+      if (skeletonPresent && volumePresent) {
+        "hybrid"
+      } else if (skeletonPresent) {
+        "skeleton"
+      } else {
+        "volume"
+      }
```
fm3, line 994: Should be extracted to a function. Maybe this already exists somewhere?

conf/webknossos.latest.routes
```diff
@@ -167,7 +167,6 @@ GET           /annotations/:typ/:id/loggedTime
 
 GET           /annotations/source/:accessTokenOrId                                              controllers.AnnotationPrivateLinkController.annotationSource(accessTokenOrId: String, userToken: Option[String])
 
-GET           /annotations/shared                                                               controllers.AnnotationController.sharedAnnotations()
```
fm3, line 170: I guess you talked to @philippotto about whether this was used somewhere?

fm3 (commented):
Nice :sunglasses: Please see my small comments

app/models/annotation/Annotation.scala
frcroth: ObjectId.fromString is Fox which is inconvenient here

frcroth: This is null when the column has no content (rather than "")

conf/webknossos.latest.routes
frcroth: Yes (see above)

app/models/annotation/AnnotationService.scala
frcroth: It exists in a different form on the Annotation object, which makes it hard to use here.

MichaelBuessemeyer (approved):
Frontend LGTM and I did not find any bugs during testing with the dev instance.

MichaelBuessemeyer:
>  I just installed the new version as a dev instance :)

Thanks :pray:.
I did not find any issues :rocket:

fm3 (approved):

PR #7411, opened 2023-10-25
Add additional coordinates to segment index

### LINKED ISSUE #7276: Segment Index + Stats for ND volume annotations
Segment Index that is stored in FossilDB should include additionalCoordinates. We should consider adding these to the key as well (just like the mag) or design a more elaborate lookup structure, depending on the queries we might get.

My guess is that the typical query is for a single spot along the additionalCoordinates axes.

frcroth:
### Steps to test:
- Open nd dataset and create some segments
- Open segment context menu in viewport and see BB and volume info
- In the segments tab, click on a group and see the segment statistics. Also test CSV export.
- Make sure everything is working for 3D datasets, such as export
- Make sure ad-hoc mesh rendering and its download is still working

### TODOs:
- [x] Frontend
	- [x]  Supply additional coordinates on segment statistics routes
	- [x] Use the new POST        /volume/:tracingId/segmentIndex/:segmentId  route (instead of GET), which also supports additional coordinates
	- [x] take care of and remove TODO comments in `frontend/javascripts/oxalis/view/context_menu.tsx`, ll. 1166 and 1308
	- [x] fix l. 1098 in https://github.com/scalableminds/webknossos/pull/7394/files/e971675c54c92a55e8c37de12835c1efa4435b76#r1404463304
	- [x] remove condition in segment_mesh_saga > loadAdHocMesh: `usePositionsFromSegmentStats`
	- [x] test mesh download (currently blocked by local CSP)

### Issues:
- fixes #7276

------


fm3:
@frcroth do you know what happens if I created an ND volume annotation on master, brush some, and then upgrade to this change? Will everything explode or will this be handled somehow? It might be too much to ask to add another migration for this case, but if things get into inconsistent states, maybe we should just disable the segment index features for such annotations?

frcroth:
> @frcroth do you know what happens if I created an ND volume annotation on master, brush some, and then upgrade to this change? Will everything explode or will this be handled somehow? It might be too much to ask to add another migration for this case (there aren’t many ND annotations in production yet), but if things get into inconsistent states, maybe we should just disable the segment index features for such annotations?

I validated that the annotations can still be viewed, but the segment index won't work anymore. Also I noticed that the front end does not provide additional coordinates for segment volume calculations leading to unhandled exceptions whenever one right clicks segments in the canvas.

dieknolle3333:
I seem to be blocked, see here: https://scm.slack.com/archives/C5AKLAV0B/p1703082065289029?thread_ts=1702477845.669909&cid=C5AKLAV0B

frontend/javascripts/oxalis/model/accessors/flycam_accessor.ts
```diff
@@ -281,11 +281,13 @@ export function hasAdditionalCoordinates(
...
 export function getAdditionalCoordinatesAsString(
   additionalCoordinates: AdditionalCoordinate[] | null | undefined,
+  separator: string = ";",
 ): string {
   if (additionalCoordinates != null && additionalCoordinates.length > 0) {
     return additionalCoordinates
       ?.map((coordinate) => `${coordinate.name}=${coordinate.value}`)
-      .reduce((a: string, b: string) => a.concat(b, ";"), "") as string;
+      .reduce((a: string, b: string) => a.concat(b, separator), "")
+      .slice(0, -1 * separator.length);
```
philippotto, line 290:
```suggestion
      .join(separator);
```

probably does what you want, if I understand the code correctly.

frontend/javascripts/oxalis/view/right-border-tabs/segments_tab/segment_statistics_modal.tsx
```diff
@@ -1,21 +1,34 @@
...
 import { voxelToNm3 } from "oxalis/model/scaleinfo";
 import { getBoundingBoxInMag1 } from "oxalis/model/sagas/volume/helpers";
+import { useSelector } from "react-redux";
+import {
+  getAdditionalCoordinatesAsString,
+  hasAdditionalCoordinates,
+} from "oxalis/model/accessors/flycam_accessor";
+
+const MODAL_ERROR_MESSAGE =
+  "Segment statistics could not not fetched. Check the console for more details.";
```
philippotto, line 22:
```suggestion
  "Segment statistics could not be fetched. Check the console for more details.";
```

frontend/javascripts/oxalis/view/right-border-tabs/segments_tab/segment_statistics_modal.tsx
```diff
@@ -1,21 +1,34 @@
...
+const CONSOLE_ERROR_MESSAGE =
+  "Segment statistics could not not fetched due to the following reason:";
```
philippotto, line 24:
```suggestion
  "Segment statistics could not be fetched due to the following reason:";
```

frontend/javascripts/oxalis/view/right-border-tabs/segments_tab/segment_statistics_modal.tsx
```diff
@@ -1,21 +1,34 @@
...
+const ERROR_CODE = CONSOLE_ERROR_MESSAGE;
 
 const SEGMENT_STATISTICS_CSV_HEADER =
   "segmendId,segmentName,groupId,groupName,volumeInVoxel,volumeInNm3,boundingBoxTopLeftPositionX,boundingBoxTopLeftPositionY,boundingBoxTopLeftPositionZ,boundingBoxSizeX,boundingBoxSizeY,boundingBoxSizeZ";
 
+const ADDITIONAL_COORDS_COLUMN = "addtionalCoordinates,";
```
philippotto, line 30:
```suggestion
const ADDITIONAL_COORDS_COLUMN = "additionalCoordinates,";
```

philippotto: I'm not fond of the trailing comma in this string, because it's easy to miss and a bit unexpected (at least for me). I'd rather see the comma when concatting the column names.

frontend/javascripts/oxalis/view/right-border-tabs/segments_tab/segment_statistics_modal.tsx
```diff
@@ -59,13 +78,18 @@ const exportStatisticsToCSV = (
           ...row.boundingBoxTopLeft,
           ...row.boundingBoxPosition,
         ]
+          .filter((el) => el != null && el.toString().length > 0)
```
philippotto, line 81: is this meant to remove additional coordinates that could be null? I think, this filtering is dangerous as it can easily remove other columns (for example, a segment name could be an empty string) which would result in an invalid CSV.
How about conditionally adding `row.additionalCoordinates` to the values array above if `hasAdditionalCoords` is true?

frontend/javascripts/oxalis/view/right-border-tabs/segments_tab/segment_statistics_modal.tsx
```diff
@@ -87,65 +111,84 @@ export function SegmentStatisticsModal({
...
+              segmentId: currentSegment.id,
+              segmentName:
+                currentSegment.name == null ? `Segment ${currentSegment.id}` : currentSegment.name,
+              groupId: currentGroupId,
+              groupName: getGroupNameForId(currentGroupId),
+              volumeInVoxel: currentSegmentSizeInVx,
+              volumeInNm3,
+              formattedSize: formatNumberToVolume(volumeInNm3),
+              boundingBoxTopLeft: boundingBoxInMag1.topLeft,
+              boundingBoxTopLeftAsString: `(${boundingBoxInMag1.topLeft.join(", ")})`,
```
philippotto, line 175: `boundingBoxTopLeftAsString` and `boundingBoxPositionAsString` are written here, but they are never used, are they? at least, I don't see it currently.

frontend/javascripts/oxalis/view/right-border-tabs/segments_tab/segment_statistics_modal.tsx
```diff
@@ -35,21 +50,25 @@ type SegmentInfo = {
...
   volumeInVoxel: number;
   boundingBoxTopLeft: Vector3;
+  boundingBoxTopLeftAsString: string;
   boundingBoxPosition: Vector3;
+  boundingBoxPositionAsString: string;
 };
 
 const exportStatisticsToCSV = (
-  segmentInformation: Array<SegmentInfo>,
+  segmentInformation: Array<SegmentInfo> | string,
```
philippotto, line 59: `segmentInformation` will be a string only in the error case, right? I would try to avoid calling this function in the first place if an error occurred. see my other comment for details.

frontend/javascripts/oxalis/view/right-border-tabs/segments_tab/segment_statistics_modal.tsx
```diff
@@ -35,21 +50,25 @@ type SegmentInfo = {
...
   tracingId: string,
   groupIdToExport: number,
+  hasAdditionalCoords: boolean,
 ) => {
-  if (segmentInformation.length === 0) {
+  if (segmentInformation.length === 0 || segmentInformation instanceof String) {
```
philippotto, line 64: this if-statement could disappear then completely. I think, it's alright to emit an empty CSV if the table is empty. feels a bit more consistent than doing nothing at all.

frontend/javascripts/oxalis/view/right-border-tabs/segments_tab/segment_statistics_modal.tsx
```diff
@@ -87,65 +111,84 @@ export function SegmentStatisticsModal({
...
+  const hasAdditionalCoords = hasAdditionalCoordinates(additionalCoordinates);
+  const additionalCoordinateStringForModal = getAdditionalCoordinatesAsString(
+    additionalCoordinates,
+    ", ",
+  );
   const dataSource = useFetch(
     async () => {
       await api.tracing.save();
-      const segmentStatisticsObjects = await Promise.all([
+      const segmentStatisticsObjects: Array<SegmentInfo> | string = await Promise.all([
```
philippotto, line 126: I would change this to
```suggestion
      const segmentStatisticsObjects: Array<SegmentInfo> | null = await Promise.all([
```

with `null` being the error case. the error message is clear in that case (it should be ERROR_CODE) and doesn't need to be contained here.

frontend/javascripts/oxalis/view/right-border-tabs/segments_tab/segment_statistics_modal.tsx
```diff
@@ -87,65 +111,84 @@ export function SegmentStatisticsModal({
...
+              ] as Vector3,
+              boundingBoxPositionAsString: `(${boundingBoxInMag1.width}, ${boundingBoxInMag1.height}, ${boundingBoxInMag1.depth})`,
+            };
+            statisticsObjects.push(segmentStateObject);
+          }
+          return statisticsObjects;
+        },
+        (error) => {
+          console.log(CONSOLE_ERROR_MESSAGE, error);
+          return ERROR_CODE;
```
philippotto, line 189:
```suggestion
          return null;
```

frontend/javascripts/oxalis/view/right-border-tabs/segments_tab/segment_statistics_modal.tsx
```diff
@@ -190,17 +233,38 @@ export function SegmentStatisticsModal({
     return potentialGroupNode?.name == null ? "" : potentialGroupNode.name;
   };
 
+  const isErrorCase = dataSource === ERROR_CODE;
```
philippotto, line 236: instead of this variable, you can now use `dataSource != null` everywhere directly. then, TS should also understand that dataSource will be an array (and not a string). so, the casting won't be necessary anymore.

frontend/javascripts/oxalis/view/right-border-tabs/segments_tab/segment_statistics_modal.tsx
```diff
@@ -87,65 +111,84 @@ export function SegmentStatisticsModal({
...
+  const dataSetScale = useSelector((state: OxalisState) => state.dataset.dataSource.scale);
+  const additionalCoordinates = useSelector(
+    (state: OxalisState) => state.flycam.additionalCoordinates,
+  );
+  const hasAdditionalCoords = hasAdditionalCoordinates(additionalCoordinates);
+  const additionalCoordinateStringForModal = getAdditionalCoordinatesAsString(
+    additionalCoordinates,
+    ", ",
+  );
   const dataSource = useFetch(
```
philippotto, line 123: should probably be renamed to `segmentStatisticsObjects` or something like this :smile:

frontend/javascripts/oxalis/view/right-border-tabs/segments_tab/segment_statistics_modal.tsx
```diff
@@ -190,17 +233,38 @@ export function SegmentStatisticsModal({
...
-        <Table dataSource={dataSource} columns={columns} style={{ whiteSpace: "pre" }} />
+      <Spin spinning={dataSource?.length === 0}>
+        {isErrorCase ? (
+          MODAL_ERROR_MESSAGE
+        ) : (
+          <>
+            {hasAdditionalCoords && (
+              <Alert
+                className="segments-stats-info-alert"
+                message={`These statistics are only valid for the current additional coordinates ${additionalCoordinateStringForModal}.`}
```
philippotto, line 256:
```suggestion
                message={`These statistics only refer for the current additional coordinate(s) ${additionalCoordinateStringForModal}.`}
```

philippotto (commented):
The front-end looks great mostly :+1: Testing also went smooth :) See my couple of comments for the final polish.

webknossos-tracingstore/app/com/scalableminds/webknossos/tracingstore/tracings/volume/VolumeTracingBucketHelper.scala
```diff
@@ -64,20 +64,31 @@ trait VolumeBucketCompression extends LazyLogging {
   }
 }
 
-trait BucketKeys extends WKWMortonHelper with WKWDataFormatHelper with LazyLogging {
+trait AdditionalCoordinateKey {
```
fm3, line 67: nice extraction :)

webknossos-tracingstore/app/com/scalableminds/webknossos/tracingstore/tracings/volume/VolumeTracingService.scala
```diff
@@ -137,11 +135,11 @@ class VolumeTracingService @Inject()(
                   ))
               case a: RevertToVersionVolumeAction =>
                 revertToVolumeVersion(tracingId, a.sourceVersion, updateGroup.version, tracing)
-              case a: DeleteSegmentDataVolumeAction =>
+              /*case a: DeleteSegmentDataVolumeAction =>
                 if (!tracing.getHasSegmentIndex) {
                   Fox.failure("Cannot delete segment data for annotations without segment index.")
                 } else
-                  deleteSegmentData(tracingId, tracing, a, segmentIndexBuffer, updateGroup.version) ?~> "Failed to delete segment data."
+                  deleteSegmentData(tracingId, tracing, a, segmentIndexBuffer, updateGroup.version) ?~> "Failed to delete segment data."*/
```
fm3, line 142: How difficult would it be to implement this feature for ND? If it’s not straightforward, I’d vote to add an assertion here, and also disable it in the frontend for ND datasets. Commenting it out is probably not the solution :D

webknossos-datastore/app/com/scalableminds/webknossos/datastore/models/datasource/AdditionalAxis.scala
```diff
@@ -29,6 +29,15 @@ object AdditionalAxis {
       p => AdditionalAxis(p.name, Array(p.bounds.x, p.bounds.y), p.index)
     )
 
+  def fromProtoAsOpt(additionalAxisProtos: Seq[AdditionalAxisProto]): Option[Seq[AdditionalAxis]] = {
```
fm3, line 32:
```suggestion
  def fromProtoSeqAsOpt(additionalAxisProtos: Seq[AdditionalAxisProto]): Option[Seq[AdditionalAxis]] = {
```
(or something similar,) to make it clear that this is about multiple AdditionalAxis elements (since the caller only sees `AdditionalAxis.fromProtoAsOpt` which looks a lot like singular)

fm3 (commented):
Backend code looks good :) I added some comments about minor things.
I also tested it and looked good :) Also still seems to work for non-ND datasets

frontend/javascripts/oxalis/model/accessors/flycam_accessor.ts
dieknolle3333: oh nice, thanks!

webknossos-tracingstore/app/com/scalableminds/webknossos/tracingstore/controllers/VolumeTracingController.scala
```diff
@@ -470,24 +475,29 @@ class VolumeTracingController @Inject()(
...
+    Action.async(validateJson[GetSegmentIndexParameters]) { implicit request =>
       accessTokenService.validateAccess(UserAccessRequest.readTracing(tracingId), urlOrHeaderToken(token, request)) {
         for {
-          magParsed <- Vec3Int.fromMagLiteral(mag, allowScalar = true).toFox ?~> "dataLayer.invalidMag"
-          cubeSizeParsed <- Vec3Int.fromUriLiteral(cubeSize).toFox ?~> "Parsing cube size failed. Use x,y,z format."
+          magParsed <- Vec3Int.fromMagLiteral(request.body.mag, allowScalar = true).toFox ?~> "dataLayer.invalidMag"
+          cubeSizeParsed <- Vec3Int
+            .fromUriLiteral(request.body.cubeSize)
+            .toFox ?~> "Parsing cube size failed. Use x,y,z format."
+          tracing <- tracingService.find(tracingId) ?~> "tracing.notFound"
```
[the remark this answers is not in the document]
frcroth: Since this is far from the only occurence of cubeSize: Maybe move renaming to a new PR?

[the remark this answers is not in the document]
fm3: fair enough. but still, could you describe what it does here in this context, maybe in a code comment?

frontend/javascripts/oxalis/view/right-border-tabs/segments_tab/segment_statistics_modal.tsx
dieknolle3333: Both are used in the columns of the table in the modal, but referenced via `dataIndex`

webknossos-tracingstore/app/com/scalableminds/webknossos/tracingstore/tracings/volume/VolumeTracingService.scala
frcroth: Don't know why I commented this originally, but I uncommented it now. @dieknolle3333 I think the delete segment data action currently does not provide additional coordinates.

frontend/javascripts/oxalis/view/right-border-tabs/segments_tab/segment_statistics_modal.tsx
dieknolle3333: ah, because the default value is `[]` and not `null`. thank you!

frontend/javascripts/oxalis/view/right-border-tabs/segments_tab/segment_statistics_modal.tsx
philippotto: yes

frontend/javascripts/oxalis/view/right-border-tabs/segments_tab/segment_statistics_modal.tsx
dieknolle3333: I think it won't be called in an error case because the ok button is disabled in that case

philippotto: ok, then the function (and its interface) should not have to deal with the `string` case even more :)

webknossos-tracingstore/app/com/scalableminds/webknossos/tracingstore/tracings/volume/VolumeTracingService.scala
dieknolle3333: Do you want me to disable this context-menu-action in the frontend? Or adjust the action that it provides additional coords? Do we all agree that this should delete the segments data in the current additional coord tuple (and not everywhere)?
I looked into it shortly and I think it would be straight-forward to implement, though I would prefer to do it in a quick and separate PR and disable it for now.
@philippotto what's your opinion on this?

philippotto: > Do we all agree that this should delete the segments data in the current additional coord tuple (and not everywhere)?

Hm, I think it would make more sense to delete it everywhere (it's not called "delete_segment_in_current_additional_axes"). Can the back-end do this in one go (i.e., not requiring one request per coord tuple)? If so, I'd prefer this.

**If not**, the following sounds like a good plan. The user should be warned then that the deletion is only done in the current coords.

> I looked into it shortly and I think it would be straight-forward to implement, though I would prefer to do it in a quick and separate PR and disable it for now.

frontend/javascripts/oxalis/view/right-border-tabs/segments_tab/segment_statistics_modal.tsx
```diff
@@ -56,38 +56,33 @@ type SegmentInfo = {
...
+        row.groupName,
+        row.volumeInVoxel,
+        row.volumeInNm3,
+        ...row.boundingBoxTopLeft,
+        ...row.boundingBoxPosition,
+      ]
+        .map(String) // convert every value to String
+        .map((v) => v.replaceAll('"', '""')) // escape double quotes
+        .map((v) => (v.includes(",") || v.includes('"') ? `"${v}"` : v)); // quote it if necessary
+      return maybeAdditionalCoords.concat(segmentInfoRow).join(",");
```
philippotto, line 80: instead of `concat`, you can simply use `...maybeAdditionalCoords` within the definition of `segmentInfoRow`.

frontend/javascripts/oxalis/view/right-border-tabs/segments_tab/segment_statistics_modal.tsx
```diff
@@ -56,38 +56,33 @@ type SegmentInfo = {
 };
 
 const exportStatisticsToCSV = (
-  segmentInformation: Array<SegmentInfo> | string,
+  segmentInformation: Array<SegmentInfo> | null,
```
philippotto, line 59: Please change it like this:

```suggestion
  segmentInformation: Array<SegmentInfo>,
```

and then adapt the caller so that the function is not called in the null case.

frontend/javascripts/oxalis/view/right-border-tabs/segments_tab/segment_statistics_modal.tsx
```diff
@@ -56,38 +56,33 @@ type SegmentInfo = {
...
   tracingId: string,
   groupIdToExport: number,
   hasAdditionalCoords: boolean,
 ) => {
-  if (segmentInformation.length === 0 || segmentInformation instanceof String) {
-    return;
-  }
   const segmentStatisticsAsString = (segmentInformation as Array<SegmentInfo>)
```
philippotto, line 64: this cast in this code version is now wrong (because `segmentInformation` might be null according to the functions typing) and should be removed. if `null` is not a valid parameter type anymore, it shouldn't be necessary, anymore.

frontend/javascripts/oxalis/view/right-border-tabs/segments_tab/segment_statistics_modal.tsx
```diff
@@ -233,33 +228,38 @@ export function SegmentStatisticsModal({
...
+                  "coordinate",
+                  additionalCoordinates?.length || 0,
+                )} ${additionalCoordinateStringForModal}.`}
                 type="info"
                 showIcon
               />
             )}
             <Table
-              dataSource={dataSource as Array<SegmentInfo>}
+              dataSource={segmentStatisticsObjects as Array<SegmentInfo>}
```
philippotto, line 262: please try to avoid (such) casts as much as possible, since they are weakinging TS ability to warn about problems. I think, in this particular case you need to inline the value of `isErrorCase` in line 246. Then, TS should understand that that `segmentStatisticsObjects` won't be null in this if-else-branch.

philippotto (commented):
Getting there :)

webknossos-tracingstore/app/com/scalableminds/webknossos/tracingstore/controllers/VolumeTracingController.scala
```diff
@@ -39,13 +43,45 @@ import com.scalableminds.webknossos.tracingstore.{
...
+    mag: Vec3Int,
+    cubeSize: Vec3Int, // Use the cubeSize parameter to map the found bucket indices to different size of cubes (e.g. reducing granularity with higher cubeSize)
+    additionalCoordinates: Option[Seq[AdditionalCoordinate]]
+)
+
+object GetSegmentIndexParameters {
+  implicit object GetSegmentIndexParametersFormat extends Format[GetSegmentIndexParameters] {
+    override def reads(json: JsValue): JsResult[GetSegmentIndexParameters] =
+      for {
+        magString <- (json \ "mag").validate[String]
```
fm3, line 63: Could the frontend send number arrays here? Then we wouldn’t need a custom reads? Or is there a reason to use the stringified version? (same for CubeSize)

fm3:
Good stuff! I added a comment about the json format.

Also, do you know what happens with nd volume annotations created before this PR? Do they have a broken segment index? Or don’t they have one? Do we need a migration for them, similar to the original https://github.com/scalableminds/webknossos/pull/7376 ?

philippotto (approved):
front-end looks good to me :) testing also went well.

frcroth:
> Also, do you know what happens with nd volume annotations created before this PR? Do they have a broken segment index? Or don’t they have one? Do we need a migration for them, similar to the original #7376 ?
Well the segment index previously was invalid since all additional coordinates were mapped to the same bucket key. With this PR the previous segment index can't be used since the keys have changed. So I think if there are nd annotations with segment indices, it would be best to recalculate the segment index.

fm3:
As discussed in person, we found that no productively used volume annotations on nd datasets exist yet. So let’s add an sql migration archiving (state='Finished') any (experimental) existing annotations on nd datasets.

fm3 (approved):
:rocket:

PR #7421, opened 2023-11-01
Move maintenance banner to top of the screen

hotzenklotz:
This PR moves the maintenance banner to the very top of the screen as part of the navbar so it does not interfere with other menus / buttons / tabs etc.

The CSS remains mostly untouched. The trick was to use real CSS variables for the navbar height instead of just a LESS variable. These CSS variables can be modified in (JS) code and then the rest of the styles just adapt.
Additionally the current navbar height is also saved in the stored and can be referenced in different downstream methods that used a constant instead.

I also refactored the maintenance banner component a little bit to more clearly separate the data/state from the rendering of the component. 

image

image

image

- https://___.webknossos.xyz

### Steps to test:
- Create new scheduled maintenances
  - e.g. Edit "maintenances" table
```
INSERT INTO "webknossos"."maintenances"("_id", "_user", "starttime", "endtime", "message", "created", "isdeleted") VALUES('653baadfef0100310a01ec94', '6538f9e0ef0100ef0101ec6d', '2023-11-02 13:03:15.125+01', '2023-11-02 13:50:00.452+01', 'Test Maintenance Banner Foobar Test Maintenance Banner Foobar Test Maintenance Banner Foobar Test Maintenance Banner Foobar Test Maintenance Banner Foobar Test Maintenance Banner Foobar Test Maintenance Banner Foobar ', '2023-10-27 14:19:43.343+02', FALSE) RETURNING "_id", "_user", "starttime", "endtime", "message", "created", "isdeleted";
```
- Try with a currently active maintenance and a maintenance scheduled in the future

### Issues:
- fixes https://scm.slack.com/archives/C5AKLAV0B/p1698066526401089

------

### DIFF
--- a/frontend/javascripts/maintenance_banner.tsx
+++ b/frontend/javascripts/maintenance_banner.tsx
@@ -4,8 +4,7 @@ import {
 } from "admin/admin_rest_api";
 import { Alert } from "antd";
 import FormattedDate from "components/formatted_date";
-import { useFetch, useInterval } from "libs/react_helpers";
-import { sleep } from "libs/utils";
+import { useInterval } from "libs/react_helpers";
 import constants from "oxalis/constants";
 import { setActiveUserAction } from "oxalis/model/actions/user_actions";
 import { Store } from "oxalis/singletons";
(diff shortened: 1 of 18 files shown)


frontend/javascripts/maintenance_banner.tsx
```diff
@@ -4,123 +4,141 @@ import {
...
 import { Store } from "oxalis/singletons";
 import { OxalisState } from "oxalis/store";
-import React, { useState } from "react";
+import React, { useEffect, useState } from "react";
 import { useSelector } from "react-redux";
 import { MaintenanceInfo } from "types/api_flow_types";
 
 const INITIAL_DELAY = 5000;
-const INTERVAL_TO_FETCH_MAINTENANCES_MS = 60000;
+const INTERVAL_TO_FETCH_MAINTENANCES_MS = 600000; // 10min
```
hotzenklotz, line 19: I intentionally increased this duration to 10min. I don't think we need to check this minute by minute. Maintenance occur rather infrequently.

philippotto: It's true that the polling won't return anything for 99% of the time. However, in the worst case, the maintenance starts and the user is only notified about it 10 minutes later (they probably closed the heads-up several days earlier and forgot about it). That might be 10 minutes of thinking WK is broken, which is why I'd avoid this. I see two options for that:
- introduce another interval which merely checks whether the next-upcoming maintenance just turned into the `current` state (no request necessary)
- stick to the 1 minute interval. this small request pales in comparison to what WK requests for data rendering.

I'd be fine with both options.

hotzenklotz: Fair point. I only considered a maintenance ending and not one starting.

```suggestion
const INTERVAL_TO_FETCH_MAINTENANCES_MS = 60000; // 1min
```

frontend/javascripts/maintenance_banner.tsx
```diff
@@ -4,123 +4,141 @@ import {
...
   }
-  const upcomingMaintenanceBanners = getClosestUpcomingMaintenanceBanner();
-  return upcomingMaintenanceBanners == null ? null : upcomingMaintenanceBanners;
+
+  useEffect(() => {
+    if (currentMaintenance || closestUpcomingMaintenance) {
+      setNavbarHeight(constants.DEFAULT_NAVBAR_HEIGHT + constants.MAINTENANCE_BANNER_HEIGHT);
+    }
+
+    if (currentMaintenance === undefined && closestUpcomingMaintenance === undefined) {
```
philippotto, line 122:
```suggestion
    if (currentMaintenance == null && closestUpcomingMaintenance == null) {
```

frontend/javascripts/maintenance_banner.tsx
```diff
@@ -4,123 +4,141 @@ import {
...
+      // Reset Navbar height if maintenance is over
+      setNavbarHeight(constants.DEFAULT_NAVBAR_HEIGHT);
+    }
+  }, [currentMaintenance, closestUpcomingMaintenance]);
+
+  // Do an initial fetch of the maintenance status so that users are notified
+  // quickly in case of ongoing maintenances.
+  setTimeout(pollMaintenances, INITIAL_DELAY);
```
philippotto, line 130:
```suggestion
  useEffect(() => {
    // Do an initial fetch of the maintenance status so that users are notified
    // quickly in case of ongoing maintenances.
    setTimeout(pollMaintenances, INITIAL_DELAY);
  }, [])
```

otherwise, a new timeout would be kicked off on each render.

frontend/javascripts/oxalis/view/plane_view.ts
```diff
@@ -248,6 +249,11 @@ class PlaneView {
     this.resize();
     this.animate();
     window.addEventListener("resize", this.resizeThrottled);
+    listenToStoreProperty(
+      (storeState) => storeState.uiInformation.navbarHeight,
+      () => this.resizeThrottled(),
+      true,
+    );
```
philippotto, line 256:
```suggestion
    this.unsubscribeFunctions.push(listenToStoreProperty(
      (storeState) => storeState.uiInformation.navbarHeight,
      () => this.resizeThrottled(),
      true,
    ));
```

that way, the listener gets removed again on `stop()` (e.g., when switching to arbitrary mode).

philippotto (commented):
Great stuff :tada: I like the CSS solution and also the refactoring of the component.
Besides my few comments, I noticed that the `ArbitraryView` didn't get a listener (like the `PlaneView` did) even though it uses `getGroundTruthLayoutRect`. Probably needs to be added?

hotzenklotz:
>  I noticed that the ArbitraryView didn't get a listener (like the PlaneView did) even though it uses getGroundTruthLayoutRect. Probably needs to be added?

Good call. Thanks I overlooked this one and will add it.

philippotto (approved):
Works very well :+1: Thank you for investing time in this :1st_place_medal:

PR #7424, opened 2023-11-06
Answer descriptive error when requesting too high API version

### LINKED ISSUE #7414: Create speaking error message when a route with a too-new api version is requested
Assume v5 is the current api version. The request handler could check incoming api requests if they request v6,v7,v8 and if so, reply with something like "This WEBKNOSSOS instance does not yet support this API version"
A comment should be added in webknossos.versioned.routes to remind us of updating the code in the request handler when a new version is introduced

frcroth:
- https://___.webknossos.xyz

### Steps to test:
- Requests a URI with higher API version

### Issues:
- fixes #7414

------


fm3 (changes requested):
Works well for me :)

Could this be rewritten to avoid the `return`? Compare https://github.com/scalableminds/webknossos/issues/6356

I think it would be more readable and scala-esque to go with another if-else. You could also extract the handling of this case into a separate method to keep this if-else cascade clean.

I’d say it is fair to do two regex matchings if needed (one for the if/else and one in the extracted function to get the api version for the error message I guess)

PR #7429, opened 2023-11-08
Don't delete dataset in DB when annotations exist

### LINKED ISSUE #6038: Deleting datasets triggers SQL foreign-key violation
Users deleting datasets for which annotations exist triggers an SQL foreign-key violation
`Detail: Key (_id)=(61fd361a010000e600f41f29) is still referenced from table "annotations".`
We should discuss what the behavior should be here (can the annotations somehow stay downloadable?).

frcroth:
- https://___.webknossos.xyz

### Steps to test:
- Create an annotation for a dataset
- Delete that dataset
- Download the annotation (it still works)
- Opening annotation or dataset (via old URL) leads to error

### TODOs:
- [x] Fix compact writes of annotations
- [x] do use the deleted dataset?
- [ ] ~~Frontend: Disable the view button for annotations with dataset status = "Deleted by user."~~

### Issues:
- fixes #6038

------


fm3:
I’ve given this some more thought, and I think maybe we should reuse the `status` field of datasets to store this information. Have a look at `deactivateUnreportedDataSources`, which is called when updating the dataset list as triggered by an update from a datastore server. If a datastore is deleted on disk (not by webknossos, but by someone with disk access), this is how webknossos "deactivates" it, implicitly setting isActive to false by setting a status message. That is a string (because it can also include different error messages from the datastore), but it has a couple of standard possible values, one of which is "No longer available on datastore." (Compare the uses of the status field throughout the code). We could add one status message like "Deleted by user" for this feature. What do you think?

Sorry that I haven’t said this sooner, I see you have already started implementing it with a new column. But maybe unifying this would be nice.

philippotto:
I just looked into this, but I think two things are missing:

- I don't see a dataset status in the annotation objects (only on the DS object) which is why I cannot change the UI when rendering annotations for deleted DS. 
- opening the annotation (with a deleted dataset) rightfully throws an error. however, opening the DS directly in view mode (e.g., via an old URL) yields a 200 status. rendering crashes later for other reasons, but I would expect an error response similar to the case where an annotation is opened.

fm3:
How do these things work for datasets that are "No longer available on the datastore"?

philippotto:
Good point. I checked that now and the error is controlled by the front-end. It only cares for the existence of `dataset.dataSource.dataLayers`. If that is undefined, the error is shown. Now, the error is properly shown for the same DS for which it didn't work before. I assume that it was a caching issue? possibly bound to the rescan interval of the disk?

By the way, annotations with the "No longer available on the datastore" case are normally shown in the annotations list. The error is then shown when clicking "open" which is fine, I guess.

So, from my point of view, the front-end wouldn't need any change. Avoiding outdated `dataset.dataSource.dataLayers` would be nice for the back-end (if my theory from above is correct). Alternatively, the front-end could start checking the status string, but I'm not sure whether that is actually helps?

fm3:
Good to hear! Yes, I think that behavior is ok.

If I understood it correctly, the new delete action should immediately delete the layers in the postgres database, so there should be no backend-internal caching for `dataset.dataSource.dataLayers`. Maybe you could have another look at that Felix, and see if you can reproduce the issue Philipp describes

app/controllers/WKRemoteDataStoreController.scala
```diff
@@ -189,11 +191,16 @@ class WKRemoteDataStoreController @Inject()(
...
           .findOneByNameAndOrganizationName(datasourceId.name, datasourceId.team)(GlobalAccessContext)
           .futureBox
+
         _ <- existingDataset.flatMap {
           case Full(dataset) =>
-            datasetDAO
-              .deleteDataset(dataset._id)
-              .flatMap(_ => usedStorageService.refreshStorageReportForDataset(dataset))
+            for {
+              annotations <- annotationDAO.findAllByDataset(dataset._id)(GlobalAccessContext)
```
fm3, line 198:
```suggestion
              annotations <- annotationDAO.countAllByDataset(dataset._id)(GlobalAccessContext)
```
No need to create objects for all the relevant annotations, if the only operation then is nonEmpty

app/models/dataset/Dataset.scala
```diff
@@ -291,13 +291,15 @@ class DatasetDAO @Inject()(sqlClient: SqlClient, datasetLayerDAO: DatasetLayerDA
...
-                                       includeSubfolders: Boolean)(implicit ctx: DBAccessContext): Fox[SqlToken] =
+  private def buildSelectionPredicates(
+      isActiveOpt: Option[Boolean],
+      isUnreported: Option[Boolean],
+      organizationIdOpt: Option[ObjectId],
+      folderIdOpt: Option[ObjectId],
+      uploaderIdOpt: Option[ObjectId],
+      searchQuery: Option[String],
+      includeSubfolders: Boolean,
+      excludeRemovedOnDisk: Boolean = true)(implicit ctx: DBAccessContext): Fox[SqlToken] =
```
fm3, line 302: This means the datasets marked as deleted are not included in the list query at all? I’d say they should still be included, just like the "no longer available on datastore" ones. I think there are already filters used by the frontend for active/inactive datasets. My guess is that the isUsable column is used for that. Also could you have another look at the `deactivateUnreported` logic in DatasetService? I’d guess that this will somehow interact. the deletedByUserStatus should probably be included in `inactiveStatusList`?

fm3 (commented):
Great stuff, I think this help create less confusion for users :)

I added some comments, I think this needs to be integrated with the existing deactivateUnreported logic a bit more.

app/models/dataset/Dataset.scala
frcroth: I think I implemented this now, but I noticed that the status "Deleted by user" is actually never (or only for a short amount of time) present in the DB, because I trigger the rechecking of the inbox which checks the directories and the dataset is actually deleted on disk, so it will be marked as "No longer available on datastore.". Is this sufficient or should there be a more explicit status? Otherwise I could delete the "Deleted by user" status again.

fm3: Hmm, I see. I think it would be valuable to preserve the information that it was in fact deleted by the user. Could you adapt the existing functions so that "Deleted by user" is not overwritten by "No longer available on datastore"?

Also, I do not yet fully see why this added check inbox trigger is needed now. Wouldn’t its effects already have happened on the wk side (delete layers, set status, etc)? Maybe I’m missing something though. It’s not like the new trigger should hurt much (it does potentially take some seconds, though, depending on IO speed)

frcroth: Fixed it now. This behavior was caused by a typo in an SQL query, which meant the status was never actually assigned. This is fixed now. A problem that persists is that the deleted dataset is still displayed in the dataset list. This status should be probably filtered (via a modified request).

fm3: Cool! Hmm, I would have expected that these datasets have isActive=false, so the current query would skip them together with the "unreported" ones.

Maybe this has something to do with the `isUsable` bool? Just noticed that. Looks somewhat redundant to be honest, but I haven’t looked into it that much. Maybe you need to change that value when updating the status

frcroth: Not sure that would do anything. On my instance the only query parameters are folderId and compact, there are datasets with isActive being returned and they are also shown in the list.

fm3: @philippotto could you comment on this? How does the frontend decide which datasets to show depending on the selected filters? is it the isActive property?

image

philippotto: > @philippotto could you comment on this? How does the frontend decide which datasets to show depending on the selected filters? is it the isActive property?

It uses the `isUnreported` property of each dataset.

frcroth: I think deleted datasets should not be shown on any filter? That would at least be unexpected for me. So they should be marked with isUnreported?

fm3: Not certain what the best UX here would be. I would have said isUnreported (meaning show them if the "missing" filter is active) is good. That is at least some explanation why the name is still reserved. But maybe you prefer something different? cc @philippotto

philippotto: > I would have said isUnreported (meaning show them if the "missing" filter is active) is good. That is at least some explanation why the name is still reserved.

I agree with that (for "deleted" datasets where annotations exist). To avoid any potential misunderstandings, let me write out how I'd imagine this overall:

- deleting a DS for which no annotation exists should completely delete the DS so that it cannot be found via the UI (also not via any filters)
- deleting a DS for which annotations exist, should tell the user "we deleted the dataset, but since annotations exist for it, the dataset will still be listed under "unreported". if you want to delete the dataset completely, do ...".
  - the perceived behavior would be equal to when the user deletes a dataset from disk.

Does this make sense?

fm3: Sounds good to me. On "if you want to delete the dataset completely, do ..." – leider gibt es aktuell keinen user-exposed weg hierfür. Das wäre ein follow-up issue. Daher würd ich den Part in der Message erstmal weglassen :sweat_smile:

philippotto: Ok, fair :)

frcroth: Deleted by user is now also a cause for datasets to be unreported.

philippotto:
> Looking good! As discussed in person, I’d ask for additional explanations in the frontend for the user,
> 
> * before deletion: that the name may still be blocked
> * after deletion inside of "show error", that this was still referenced and is kept to not break things elsewhere

Done!

frontend/javascripts/dashboard/advanced_dataset/dataset_action_view.tsx
```diff
@@ -169,7 +169,17 @@ class DatasetActionView extends React.PureComponent<Props, State> {
           onClick={() =>
             Modal.error({
               title: "Cannot load this dataset",
-              content: dataset.status,
+              content: (
+                <div>
+                  <p>{dataset.status}</p>
+                  {dataset.status === "Deleted by user." ? (
+                    <p>
+                      Even though, this dataset was deleted by a user, it is still shown here,
```
fm3, line 177: no comma after even though 🙈

frontend/javascripts/dashboard/advanced_dataset/dataset_action_view.tsx
```diff
@@ -169,7 +169,17 @@ class DatasetActionView extends React.PureComponent<Props, State> {
...
+                      because it is referenced by at least one annotation.
```
frcroth, line 178: Right now it should be "it **was** referenced" since it will not deleted if all associated annotations are deleted eventually. Follow up issue?

philippotto: you're right. "was" is clearer. I changed it. A follow-up issue for automatically "completing the deletion" once all annotations are deleted is not necessary imo.

frcroth:
@philippotto Does this need frontend review?

philippotto:
> @philippotto Does this need frontend review?

I don't think so. The front-end diff is mostly changing some words. @fm3 had a look at this, I think :)

PR #7430, opened 2023-11-08
Add stats filter to task list view

hotzenklotz:
PR does two things:
- Prioritized the stats column in the table by moving it towards the beginning of the table instead of the last position which is often cut off
- Added a filter to the stats column to quickly filter for tasks with "Prending", "In-Progress" or "Finished" instances

image

- https://___.webknossos.xyz

### Steps to test:
- Create new tasks
- Go to Task List View
- Filter away

### Issues:
- None

------


fm3 (commented):
Works for me :)

Could you replace the wording "in progress" by "active", in order to use the same term everywhere rather than introduce another synonym?

fm3 (approved):

PR #7437, opened 2023-11-13
Read Segment Index Files, Segment Stats for Annot. w/ FallbackL.

### LINKED ISSUE #7132: segment-to-bucket index for volume annotations *with fallback segmentation*
Follow-up for #7063 with fallback segmentation layer. This requires segment index metadata for the fallback layer to be available. Format to be determined.

First step would be to read the segment index file in the datastore (similar to the meshfile).
Second step is to then enable the segment index feature also for volume annotations with fallback layer, mixing the info from datastore and tracingstore.

We need to make sure the delete segment action does not explode by tackling hundreds of thousands of buckets

frcroth:
- https://___.webknossos.xyz

### Steps to test:
- Test the datastore routes segmentIndex, segmentVolume and boundingBox (test with l4_sample)
- Test the fallback behavior: When a volume tracing has a fallback, it will by default return the segment file index. Once something changed in the corresponding buckets by annotating, all affected segments will be queried from the mutable segment index

### TODOs:
- [x] Test Scenario 1
	- [x] Create a volume tracing annotation for a dataset with segmentation layer using that as fallback layer
	- [x] The tracing has "hasSegmentIndex" true
	- [x] When annotating a new segment, fossilDB is filled with content
	- [x] When querying segment volume for the new segment, the correct/a sensible value is returned
	- [x] When querying a segment volume for a segment in the file, the correct value is returned
	- [x] Select an existing segment id, annotate
	- [x] Now the segment volume is the one you just annotated
- [x] A lot of things are marked as TODO
- [x] The delete segment data action should be limited to segments smaller than x buckets -> Followup?
- [x] Handle ds without mag 1
- [x] Bounding Box
- [x] additional coords -> Followup?
- [x] Create simple test dataset
- [x] Speed up saving

### TODOs frontend:
- [x] use datastore host if no volume annotation layer is visible (cases: no annotation or skeleton-only annotation)
- [x] remove code duplication by finding out where to put helper method (context menu and stats modal)
- [x] make sure the right conditions are tested
- [x] find out whether meshes should use index file in dataview mode

### Followups
- Remove await in nml parsing (#7551)
- #7554
- #7556
- #7615
- #7618

### Issues:
- fixes #7132 

------- 


webknossos-tracingstore/app/com/scalableminds/webknossos/tracingstore/tracings/editablemapping/RemoteFallbackLayer.scala
```diff
@@ -15,6 +17,10 @@ case class RemoteFallbackLayer(organizationName: String,
                                layerName: String,
                                elementClass: ElementClassProto)
 
+object RemoteFallbackLayer extends ProtoGeometryImplicits {
+  def fromDataLayerAndDataSource(dataLayer: DataLayerLike, dataSource: DataSourceId): RemoteFallbackLayer =
```
fm3, line 21: good idea =)

webknossos-tracingstore/app/com/scalableminds/webknossos/tracingstore/tracings/volume/VolumeSegmentIndexBuffer.scala
```diff
@@ -47,4 +54,20 @@ class VolumeSegmentIndexBuffer(tracingId: String, volumeSegmentIndexClient: Foss
...
+      fossilDbData <- volumeSegmentIndexClient
+        .get(key, Some(version), mayBeEmpty = Some(true))(fromProtoBytes[ListOfVec3IntProto])
+        .map(_.value)
+        .fillEmpty(ListOfVec3IntProto.of(Seq()))
+      layerData <- fallbackLayer match {
+        case Some(layer) if fossilDbData.length == 0 =>
+          remoteDatastoreClient.querySegmentIndex(layer, segmentId, mag, mappingName, userToken)
+        case _ => Fox.successful(Seq.empty)
+      }
+      combined = fossilDbData.values.map(vec3IntFromProto) ++ layerData
```
fm3, line 69: Am I reading this correctly that combined will in the end have either fossilDbData values or layerData values but not both? In this case I’m not super sure about the `++` operator. It should at least be explained in a comment. Maybe a getOrElse cascade can be used instead, to make the intent clearer

fm3 (changes requested):
Great stuff, seems to work :) I did not yet test all cases, but I did get valid-looking values in an annotation, including after I brushed some.

Some thoughts:
- [x] frontend should offer segment index functionality in dataset view mode if segment file exists.
   - needs to query if a segment index file exists (talk to @frcroth about the API for that) -> use GET /datasets/:organizationName/:dataSetName/layers/:dataLayerName/segmentIndex  which returns a boolean
   - use datastore routes then instead of tracingstore routes (`/datasets/:organizationName/:dataSetName/layers/:dataLayerName/segmentStatistics/volume`, `/datasets/:organizationName/:dataSetName/layers/:dataLayerName/segmentStatistics/boundingBox`) (can the selected mapping be supplied here? the files are always for oversegmentation, right?)
- [x] frontend should offser segment index functionality for annotations with hasSegmentIndex=true even if fallback layer exists
  - [x] in segmentation layer context menu, in segment list context menu, in 3d viewport
  - [x] I saw a request when clicking a mesh in 3d viewport send a request `{"mag":[1,1,1],"segmentIds":[0]}` (segment id 0 shouldn’t be requested, the hovered id should be)
- [x] Frontend: slow saving should not lead to »A newer version of this annotation was found on the server. Your current changes to this annotation cannot be saved anymore.« Race codition in frontend?
- [x] saving volume data is way too slow, even got a socket hang up/connection reset. We should do some profiling there
- [x] brushing/saving with a fresh id fails (as already reported)
  - [ ] same still seems to happen in removeBucket code. created a couple new cells in close proximity and brushed around
- [x] do we need a way to convert an annotation to have a segment index in case the file was created after the annotation was?
- [x] Segment Stats data is outdated when opening context menu when latest changes are not saved yet.
  - [x] This can also lead to errors if a segment id is requested that the backend doesn’t yet know about (only in the new code paths?)
  - [x] We should discuss how to handle this. Maybe the context menu isn’t the right place for this information. Maybe we should move it to a more explicit dialog that can only be shown after ensureSavedState? cc @philippotto 
- [x] TODO for myself after next iteration: test upload(single,multiple),duplicate,merge. Test with applied mappings
- [ ] Create Follow-up issues
  - [ ] Support for proofreading annotations
  - [x] Support for ND datasets
  - [ ] (added by charlie): improve context menu by creating submenu for segment volume and BB
- [x] Keep in mind that this will have to be compatible with https://github.com/scalableminds/webknossos/pull/7411 – not necessarily work for ND datasets but possibly e.g. make the datastore routes POST too (let’s see which PR is merged first…)
	- [x] change segmentIndex/:segmentId to use POST in datastore

I also added a few comments to the code.

app/models/dataset/DataStore.scala
```diff
@@ -77,6 +78,9 @@ class DataStoreService @Inject()(dataStoreDAO: DataStoreDAO)(implicit ec: Execut
       result <- block(dataStore)
     } yield result).getOrElse(Forbidden(Json.obj("granted" -> false, "msg" -> Messages("dataStore.notFound"))))
 
+  def clientFor(dataStore: DataStore): WKRemoteDataStoreClient =
+    new WKRemoteDataStoreClient(dataStore, rpc)
```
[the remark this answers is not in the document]
frcroth: All other occurrences do not use DataStoreService. Could add the service there or remove this method since it is only used once.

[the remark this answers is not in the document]
fm3: then maybe remove the method, if it is easy to access rpc in the caller

webknossos-datastore/app/com/scalableminds/webknossos/datastore/services/Hdf5Utils.scala
```diff
@@ -0,0 +1,14 @@
+package com.scalableminds.webknossos.datastore.services
+
+import org.apache.commons.codec.digest.MurmurHash3
+
+import java.nio.ByteBuffer
+
+trait Hdf5Utils {
```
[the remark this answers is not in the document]
frcroth: There are other traits in the services directory as well, which are only implemented by services, I think the location is fine.

webknossos-datastore/app/com/scalableminds/webknossos/datastore/controllers/DataSourceController.scala
```diff
@@ -529,4 +532,91 @@ class DataSourceController @Inject()(
...
+          magParsed <- Vec3Int.fromMagLiteral(mag, allowScalar = true).toFox ?~> "dataLayer.invalidMag"
+          cubeSizeParsed <- Vec3Int.fromUriLiteral(cubeSize).toFox ?~> "Parsing cube size failed. Use x,y,z format."
+          (topLefts, fileMag) <- segmentIndexFileService.readSegmentIndex(organizationName,
+                                                                          datasetName,
+                                                                          dataLayerName,
+                                                                          segmentId.toLong)
+          bucketPositions = topLefts
+            .map(_.scale(DataLayer.bucketLength)) // map indices to positions
+            .map(_ / (magParsed / fileMag))
+            .map(_ / cubeSizeParsed) // map positions to cube indices
```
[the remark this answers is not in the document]
frcroth: No such attribute

```
GROUP "/" {
   ATTRIBUTE "artifact_schema_version" {
      DATATYPE  H5T_STD_I64LE
      DATASPACE  SCALAR
   }
   ATTRIBUTE "hash_function" {
      DATATYPE  H5T_STRING {
         STRSIZE H5T_VARIABLE;
         STRPAD H5T_STR_NULLTERM;
         CSET H5T_CSET_UTF8;
         CTYPE H5T_C_S1;
      }
      DATASPACE  SCALAR
   }
   ATTRIBUTE "mag" {
      DATATYPE  H5T_STD_I64LE
      DATASPACE  SIMPLE { ( 3 ) / ( 3 ) }
   }
   ATTRIBUTE "mapping_name" {
      DATATYPE  H5T_STRING {
         STRSIZE H5T_VARIABLE;
         STRPAD H5T_STR_NULLTERM;
         CSET H5T_CSET_UTF8;
         CTYPE H5T_C_S1;
      }
      DATASPACE  SCALAR
   }
   ATTRIBUTE "n_buckets" {
      DATATYPE  H5T_STD_I64LE
      DATASPACE  SCALAR
   }
   DATASET "bucket_offsets" {
      DATATYPE  H5T_STD_U64LE
      DATASPACE  SIMPLE { ( 23490 ) / ( 23490 ) }
   }
   DATASET "buckets" {
      DATATYPE  H5T_STD_U64LE
      DATASPACE  SIMPLE { ( 17617, 3 ) / ( 17617, 3 ) }
   }
   DATASET "top_lefts" {
      DATATYPE  H5T_STD_U16LE
      DATASPACE  SIMPLE { ( 264905, 3 ) / ( 264905, 3 ) }
   }
}


```

webknossos-datastore/app/com/scalableminds/webknossos/datastore/controllers/DataSourceController.scala
```diff
@@ -529,4 +532,91 @@ class DataSourceController @Inject()(
...
+
+  def getSegmentVolume(token: Option[String],
+                       organizationName: String,
+                       datasetName: String,
+                       dataLayerName: String): Action[SegmentStatisticsParameters] =
+    Action.async(validateJson[SegmentStatisticsParameters]) { implicit request =>
+      accessTokenService.validateAccess(UserAccessRequest.readDataSources(DataSourceId(datasetName, organizationName)),
+                                        urlOrHeaderToken(token, request)) {
+        for {
+          _ <- Fox.box2Fox(segmentIndexFileService.getSegmentIndexFile(organizationName, datasetName, dataLayerName)) ?~> "segmentIndexFile.notFound" //TODO: Dont use head, get all volumes
```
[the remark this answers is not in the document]
frcroth: Already exists, just forgot to use it here

webknossos-tracingstore/app/com/scalableminds/webknossos/tracingstore/tracings/volume/VolumeSegmentIndexBuffer.scala
frcroth: Made it a little clearer now

webknossos-tracingstore/app/com/scalableminds/webknossos/tracingstore/tracings/volume/VolumeTracingService.scala
```diff
@@ -453,22 +521,33 @@ class VolumeTracingService @Inject()(
...
                             sourceTracing: VolumeTracing,
                             destinationId: String,
-                            destinationTracing: VolumeTracing): Fox[Unit] =
+                            destinationTracing: VolumeTracing,
+                            userToken: Option[String]): Fox[Unit] =
     for {
       isTemporaryTracing <- isTemporaryTracing(sourceId)
       sourceDataLayer = volumeTracingLayer(sourceId, sourceTracing, isTemporaryTracing)
       buckets: Iterator[(BucketPosition, Array[Byte])] = sourceDataLayer.bucketProvider.bucketStream()
+      // Duplicate segment index if it exists instead of building a new one?
```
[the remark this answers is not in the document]
frcroth: https://github.com/scalableminds/webknossos/issues/7554

webknossos-datastore/app/com/scalableminds/webknossos/datastore/controllers/DataSourceController.scala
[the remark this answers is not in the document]
fm3: Ok then I’d say we should go for DataLayer.bucketLength here, and keep the API simple by omitting the cubeSize parameter

app/models/annotation/AnnotationUploadService.scala
```diff
@@ -15,48 +15,65 @@ import net.liftweb.common.{Box, Empty, Failure, Full}
...
+  def extractFromNml(
+      inputStream: InputStream,
+      name: String,
+      overwritingDataSetName: Option[String],
+      isTaskUpload: Boolean,
+      userToken: Option[String],
+      basePath: Option[String] = None)(implicit m: MessagesProvider, ec: ExecutionContext): NmlParseResult =
+    NmlParser
+      .parse(name, inputStream, overwritingDataSetName, isTaskUpload, basePath, None, userToken)
+      .await("NMLParser/parse was changed to return Fox in #7437. Removing this await is tracked in #7551") match {
```
fm3, line 50: What is the timeout of the await? Parsing large NML files can take some time. Maybe it would be possible to move the await into the parser so that only the thing that really has to be async (looking up the segment index file existence) is awaited?

frcroth: Moved it; the timeout is 5 seconds currently

dieknolle3333:
@fm3 thank you for your thorough testing! I tried to make the code more robust with my current changes.
>This can also lead to errors if a segment id is requested that the backend doesn’t yet know about (only in the new code paths?)

This was a conscious choice, because I wanted to avoid saving every time the context menu is opened. As far as I can tell, the backend simply returns 0 if it doesn't know the segment ID yet. And if you hit the "update this statistic" button, the annotation is saved and the right number is displayed. Please let me know if I am missing anything.

> Maybe the context menu isn’t the right place for this information. Maybe we should move it to a more explicit dialog that can only be shown after ensureSavedState? 

@philippotto and I agreed that this could be solved by moving volume and BB to a submenu which needs to be actively opened, so it wouldn't be a problem to save the annotation right then. This would be done in a separate PR though.

fm3:
sounds fair!

frontend/javascripts/oxalis/model/sagas/dataset_saga.ts
```diff
@@ -149,3 +152,31 @@ export function* watchZ1Downsampling(): Saga<void> {
     maybeShowWarning,
   );
 }
+
+export function* ensureSegmentIndexIsLoaded(): Saga<void> {
```
dieknolle3333, line 156: I was wondering whether it makes sense to have a big dataset saga (or to start one), but all the other sagas in the module were rather separate so I put mine separately, too.

dieknolle3333:
I have a somewhat basic question. Is it true that segment stats are available <-> a segment index exists?
And am I right that technically for ND annotations, there are no segment stats if there is a fallback layer?
I am not checking for the latter atm.

fm3:
> Is it true that segment stats are available <-> a segment index exists

Yes, this is true. If the layer (be it an on-disk segmentation layer or a volume annotation layer) has a segment index, the segment stats can be computed by the backend.

> And am I right that technically for ND annotations, there are no segment stats if there is a fallback layer?

This is currently also true, because we cannot compute segment index files for ND datasets yet. However, no special check should be needed for that in the front-end, as the absence of the segment index file will already be reported in the relevant routes.

webknossos-datastore/app/com/scalableminds/webknossos/datastore/services/Hdf5Utils.scala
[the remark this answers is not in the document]
frcroth: Had a look at extraction again, did not seem to be easy enough to be wort the effort for me.

frcroth:
@fm3 Many of your points are things to discuss. How do we proceed on them?

fm3:
@normanrz I would like to discuss the point »do we need a way to convert an annotation to have a segment index in case the segment index index file for the fallbak layer was created after the annotation was?« – I see the following options
 - (1) existing annotations for this dataset have no segment index, duplicate helps
 - (2) button in each annotation to add segment index
 - (3) button to migrate all annotations of a dataset

(or something else)

What do you think?

frontend/javascripts/oxalis/view/context_menu.tsx
```diff
@@ -1179,37 +1186,46 @@ function ContextMenuInner(propsWithInputRef: Props) {
...
+  // segmentIdAtPosition is only set if a segment is hovered in one of the xy, xz, or yz viewports.
+  // maybeClickedMeshId is only set, when a mesh is hovered in the 3d viewport.
+  // Thus the segment id is always unambiguous / clearly defined.
+  const clickedSegmentOrMeshId =
+    maybeClickedMeshId != null ? maybeClickedMeshId : segmentIdAtPosition;
+  const isHoveredSegmentOrMesh = clickedSegmentOrMeshId > 0;
+
+  const { dataset, tracing, flycam } = useSelector((state: OxalisState) => state);
+  useEffect(() => {
+    Store.dispatch(ensureSegmentIndexIsLoadedAction(visibleSegmentationLayer?.name));
```
philippotto, line 1200: you pass a name here, but the saga doesn't make use of the name.

frontend/javascripts/oxalis/model/accessors/dataset_accessor.ts
```diff
@@ -827,3 +827,11 @@ export function getEffectiveIntensityRange(
 
   return layerConfiguration.intensityRange || defaultIntensityRange;
 }
+
+// Note that this prop is lazy-loaded. Call an ensureSegmentIndexIsLoadedAction to make sure this info was fetched.
+export function getMaybeSegmentIndexAvailability(
+  dataset: APIDataset,
+  layerName: string | undefined,
```
philippotto, line 834: I find it a bit weird that this function accepts undefined and then uses it to compare it with the layerName. I'd recommend to do one of the two options:
- don't accept undefined and adapt the caller to avoid this case (since react hooks cannot be called within ifs, you would need to do something like this: ¹)
- add a `if (visibleSegmentationLayer == null) { return false }` at the top of this function here

¹:
```
useSelector((state: OxalisState) =>
    visibleSegmentationLayer != null ? getMaybeSegmentIndexAvailability(state.dataset, visibleSegmentationLayer.name) : false,
  );
```

frontend/javascripts/oxalis/view/context_menu.tsx
```diff
@@ -1179,37 +1186,46 @@ function ContextMenuInner(propsWithInputRef: Props) {
...
-  const hasNoFallbackLayer =
-    visibleSegmentationLayer != null && !hasFallbackLayer(visibleSegmentationLayer);
+
+  // Currently either segmentIdAtPosition or maybeClickedMeshId is set, but not both.
+  // segmentIdAtPosition is only set if a segment is hovered in one of the xy, xz, or yz viewports.
+  // maybeClickedMeshId is only set, when a mesh is hovered in the 3d viewport.
+  // Thus the segment id is always unambiguous / clearly defined.
+  const clickedSegmentOrMeshId =
+    maybeClickedMeshId != null ? maybeClickedMeshId : segmentIdAtPosition;
+  const isHoveredSegmentOrMesh = clickedSegmentOrMeshId > 0;
```
philippotto, line 1196: can we rename this to `wasSegmentOrMeshClicked`? I know it's been like this before, but I find it confusing that the content depends on a click and not on a hover.

frontend/javascripts/oxalis/view/context_menu.tsx
```diff
@@ -1223,24 +1239,21 @@ function ContextMenuInner(propsWithInputRef: Props) {
           formatNumberToVolume(volumeInNm3),
           `${boundingBoxTopLeftString}, ${boundingBoxSizeString}`,
         ];
+      } catch (_error) {
+        const notFetchedMessage = "could not be fetched";
```
philippotto, line 1243: Would a capital C ("Could not ...") look better in the UI?

frontend/javascripts/oxalis/view/context_menu.tsx
```diff
@@ -1179,37 +1186,46 @@ function ContextMenuInner(propsWithInputRef: Props) {
...
+
+  const { dataset, tracing, flycam } = useSelector((state: OxalisState) => state);
+  useEffect(() => {
+    Store.dispatch(ensureSegmentIndexIsLoadedAction(visibleSegmentationLayer?.name));
+  }, [visibleSegmentationLayer]);
+  const isSegmentIndexAvailable = useSelector((state: OxalisState) =>
+    getMaybeSegmentIndexAvailability(state.dataset, visibleSegmentationLayer?.name),
+  );
+  const isLoadingVolumeAndBB = ["loading", "loading"];
   const [segmentVolume, boundingBoxInfo] = useFetch(
```
philippotto, line 1206: I'd rename these vars to `segmentVolumeLabel` and `boundingBoxInfoLabel`. That way, it is a bit more expected that the labels can contain strings, such as "loading" or "could not be fetched".

frontend/javascripts/oxalis/view/right-border-tabs/segments_tab/segments_view.tsx
```diff
@@ -44,8 +44,8 @@ import { getSegmentIdForPosition } from "oxalis/controller/combinations/volume_h
 import {
   getMappingInfo,
   getResolutionInfoOfVisibleSegmentationLayer,
+  getMaybeSegmentIndexAvailability as getMaybeSegmentIndexAvailability,
```
philippotto, line 47:
```suggestion
  getMaybeSegmentIndexAvailability,
```

frontend/javascripts/oxalis/view/right-border-tabs/segments_tab/segments_view.tsx
```diff
@@ -1536,21 +1538,14 @@ class SegmentsView extends React.Component<Props, State> {
   getSegmentStatisticsModal = (groupId: number) => {
     const segments = this.getSegmentsOfGroupRecursively(groupId);
     const visibleSegmentationLayer = this.props.visibleSegmentationLayer;
-    const hasNoFallbackLayer =
-      visibleSegmentationLayer != null && !hasFallbackLayer(visibleSegmentationLayer);
-    if (hasNoFallbackLayer && this.props.hasVolumeTracing && segments != null) {
-      const state = Store.getState();
-      const tracingId = this.props.activeVolumeTracing?.tracingId;
-      if (tracingId == null) return null;
-      const tracingStoreUrl = state.tracing.tracingStore.url;
```
philippotto, line 1545: nice :sunglasses:

frontend/javascripts/oxalis/view/right-border-tabs/segments_tab/segments_view_helper.tsx
```diff
@@ -35,6 +39,52 @@ export function getBaseSegmentationName(segmentationLayer: APIDataLayer) {
...
+  const tracingStoreHost = tracing?.tracingStore.url;
+
+  const maybeVolumeTracing =
+    "tracingId" in visibleSegmentationLayer && visibleSegmentationLayer.tracingId != null
+      ? getVolumeTracingById(tracing, visibleSegmentationLayer.tracingId)
+      : null;
+  // For non-segmentation layers and for viewing datasets, we'll always use the datastore URL
+  const shouldUseDataStore = maybeVolumeTracing == null;
+  return shouldUseDataStore ? dataStoreUrl : `${tracingStoreHost}/tracings/volume/${tracingId}`;
+}
```
philippotto, line 63: I think, this function could be simpler. How about this?

```suggestion
export function getVolumeRequestUrl(
  dataset: APIDataset,
  tracing: HybridTracing | null,
  tracingId: string | undefined,
  visibleSegmentationLayer: APISegmentationLayer | APIDataLayer,
) {
  if (tracing == null || tracingId == null) {
    return `${dataset.dataStore.url}/data/datasets/${dataset.owningOrganization}/${dataset.name}/layers/${visibleSegmentationLayer.name}`;
  } else {
    const tracingStoreHost = tracing.tracingStore.url;
    return `${tracingStoreHost}/tracings/volume/${tracingId}`;
  }
}
```

Your current implementation is quite pessismistic and doesn't trust the provided tracingId. However, I'd argue that that ID can be trusted (it's typically passed as `volumeTracing?.tracingId` anyway). If the caller provides a wrong id, that's a bug that should be fixed then.

philippotto (commented):
Front-end looks good mostly. Left some smaller suggestions :) Also, I didn't test yet. I'd do this in a later iteration.

frontend/javascripts/oxalis/view/context_menu.tsx
dieknolle3333: Personally I think lower-case looks better, but let me know if you disagree!

dieknolle3333: image
image
for reference

philippotto: okay, then let's stick to lower case :+1:

frontend/javascripts/oxalis/model/actions/dataset_actions.ts
```diff
@@ -5,12 +5,14 @@ type SetLayerTransformsAction = ReturnType<typeof setLayerTransformsAction>;
...
 >;
+type SetLayerHasSegmentIndexAction = ReturnType<typeof setLayerHasSegmentIndexAction>;
 
 export type DatasetAction =
   | SetDatasetAction
   | SetLayerMappingsAction
   | SetLayerTransformsAction
-  | EnsureLayerMappingsAreLoadedAction;
+  | EnsureLayerMappingsAreLoadedAction
+  | SetLayerHasSegmentIndexAction;
```
philippotto, line 15: `EnsureSegmentIndexIsLoadedAction` is missing here (also needs to be defined probably).

fm3 (commented):
I added a segment index file for this dataset https://www.notion.so/scalableminds/Test-Datasets-c0563be9c4a4499dae4e16d9b2497cfb?pvs=4#4f27c86bc8fc485e9d04b0dc1b1488c9

Unfortunately, it looks like the frontend does not yet send the mapping name in the volume/bbox routes.

The backend seems to take that parameter, but when I set it in a custom request, I did not get a changed volume, which is surprising.

Maybe you could have a look there as well.

frontend/javascripts/oxalis/model/accessors/dataset_accessor.ts
```diff
@@ -827,3 +827,16 @@ export function getEffectiveIntensityRange(
...
 }
+
+// Note that `hasSegmentIndex` needs to be loaded first (otherwise, the returned
+// value will be undefined). Dispatch an ensureSegmentIndexIsLoadedAction to make
+// sure this info is fetched.
+export function getMaybeSegmentIndexAvailability(
+  dataset: APIDataset,
+  layerName: string | undefined,
+) {
+  if (layerName == null || layerName.length < 1) {
```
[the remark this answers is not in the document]
dieknolle3333: I think of a specific case, I just thought it would be smart to check that so that we dont need to iterate through the layers. but I think you're right, right now that shouldn't happen and if it does, the behaviour is still correct

frontend/javascripts/oxalis/model/actions/dataset_actions.ts
dieknolle3333: actually I did this on purpose because this action is not used in a reducer and so I thought it wouldn't be necessary to include this in the union types. will add this so that the action types remain complete

dieknolle3333:
@fm I addressed the issues you brought up, but didn't test them thoroughly yet, especially because I didn't have a segment index file for the dataset you mentioned (but also because I lacked the time this week). I will test it first thing next week, but didn't want to block you from moving on with this PR.

frcroth:
@fm3 Fixed niow

webknossos-datastore/app/com/scalableminds/webknossos/datastore/services/SegmentIndexFileService.scala
```diff
@@ -0,0 +1,189 @@
...
+  def getSegmentIndexFile(organizationName: String, datasetName: String, dataLayerName: String): Box[Path] =
+    for {
+      _ <- Full("")
+      layerDir = dataBaseDir.resolve(organizationName).resolve(datasetName).resolve(dataLayerName)
+      segmentIndexDir = layerDir.resolve(this.segmentIndexDir)
+      files <- PathUtils.listFiles(segmentIndexDir, silent = true, PathUtils.fileExtensionFilter(hdf5FileExtension))
+      file <- files.headOption
+    } yield file
+
+  def readSegmentIndex(organizationName: String,
```
fm3, line 50: here, too, I think a comment explaining which system the returned Vec3Ints are in would help future readers

webknossos-datastore/app/com/scalableminds/webknossos/datastore/services/SegmentIndexFileService.scala
```diff
@@ -0,0 +1,181 @@
...
+        _ <- Fox.successful(())
+        buckets = segmentIndex.reader
+          .uint64()
+          .readMatrixBlockWithOffset("hash_buckets", (bucketEnd - bucketStart + 1).toInt, 3, bucketStart, 0)
+        bucketLocalOffset = buckets.map(_(0)).indexOf(segmentId)
+        _ <- bool2Fox(bucketLocalOffset >= 0)
+        topLeftStart = buckets(bucketLocalOffset)(1)
+        topLeftEnd = buckets(bucketLocalOffset)(2)
+        topLefts = segmentIndex.reader
+          .uint16() // Read datatype from attributes?
```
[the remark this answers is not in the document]
fm3: @markbader ping

webknossos-tracingstore/app/com/scalableminds/webknossos/tracingstore/TSRemoteDatastoreClient.scala
```diff
@@ -111,6 +117,52 @@ class TSRemoteDatastoreClient @Inject()(
...
+  def querySegmentIndex(remoteFallbackLayer: RemoteFallbackLayer,
+                        segmentId: Long,
+                        mag: Vec3Int,
+                        mappingName: Option[String],
+                        userToken: Option[String]): Fox[Seq[Vec3Int]] =
+    for {
+      remoteLayerUri <- getRemoteLayerUri(remoteFallbackLayer)
+      positions <- rpc(s"$remoteLayerUri/segmentIndex/$segmentId")
+        .addQueryStringOptional("token", userToken)
+//        .silent
```
fm3, line 138: line should probably be re-added

webknossos-tracingstore/app/com/scalableminds/webknossos/tracingstore/TSRemoteDatastoreClient.scala
```diff
@@ -111,6 +117,52 @@ class TSRemoteDatastoreClient @Inject()(
...
+        .postJsonWithJsonResponse[GetSegmentIndexParameters, Seq[Vec3Int]](
+          GetSegmentIndexParameters(mag,
+                                    cubeSize = Vec3Int.ones, // Don't use the cubeSize parameter here
```
fm3, line 141: could you explain why?

webknossos-tracingstore/app/com/scalableminds/webknossos/tracingstore/tracings/volume/VolumeTracingService.scala
```diff
@@ -211,12 +226,17 @@ class VolumeTracingService @Inject()(
         Fox.serialCombined(additionalCoordinateList)(additionalCoordinates => {
           val mag = vec3IntFromProto(resolution)
           for {
+            fallbackLayer <- getFallbackLayer(tracingId)
             bucketPositionsRaw <- volumeSegmentIndexService.getSegmentToBucketIndexWithEmptyFallbackWithoutBuffer(
+              fallbackLayer,
               tracingId,
               a.id,
               mag,
+              None, //TODO: ???
```
fm3, line 235: yes, None means the latest version here, which is the only one that this operation makes sense for

fm3 (commented):
Kudos for pushing this so far! I added a couple comments to the code, but nothing big. I’ll do some more testing as well.

fm3:
I think I ran into another bug: I created two annotations for a dataset with no existing segmentation layer. In both I created one brush stroke each (in different locations), each with segment id 1. I then used the merge modal described above to merge one into the other annotation. The merging operation now finishes successfully and the buckets look good. However, The displayed segment volume in the merged annotation for the segment that was merged in now has a different than in the source annotation. Maybe something is wrong there still. This might have something to do with the id remapping that happens during the merging. Not sure if this problem is introduced in this PR or if it already exists in master. Can you reproduce this?

fm3:
Looks like something may already be wrong with my source annotation. If I download + reupload it, the value also changes. Also, the value seems way too low (I get 32 nm³ even though I made a good brush stroke with the default brush size and voxel size is 1×1×1 nm³/vx).

So maybe merging works fine but the value for my source annotation is already wrong

app/models/annotation/AnnotationUploadService.scala
```diff
@@ -149,23 +172,36 @@ class AnnotationUploadService @Inject()(tempFileService: TempFileService) extend
...
 
+  def getRemoteDatastoreClientForDatasetNameAndOrg(datasetName: String, organizationName: String)(
+      implicit ec: ExecutionContext): Option[WKRemoteDataStoreClient] = {
+    val fox = for {
+      _ <- Fox.successful(())
+      organizationObjectId <- organizationDAO.findIdByName(organizationName)(GlobalAccessContext)
+      dataset <- datasetDAO.findOneByNameAndOrganization(datasetName, organizationObjectId)(GlobalAccessContext)
+      dataStore <- datasetService.dataStoreFor(dataset)(GlobalAccessContext)
+    } yield new WKRemoteDataStoreClient(dataStore, rpc)
+    fox.await("No fox context in AnnotationUploadService, see #7551").toOption
```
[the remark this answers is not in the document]
frcroth: 10 seconds by default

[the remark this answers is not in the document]
frcroth: Changed it to 1 minute now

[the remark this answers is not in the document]
fm3: Yeah, let’s increase it a bit, then, say 1 minute. This should not fail just because there is a lot going on at the same time

webknossos-datastore/app/com/scalableminds/webknossos/datastore/controllers/DataSourceController.scala
```diff
@@ -554,4 +561,109 @@ class DataSourceController @Inject()(
...
+          bucketPositionsForCubeSize = bucketPositions
+            .map(_.scale(DataLayer.bucketLength)) // bucket positions raw are indices of 32³ buckets
+            .map(_ / request.body.cubeSize)
+            .distinct // divide by requested cube size to map them to larger buckets, select unique
+            .map(_ * request.body.cubeSize) // return positions, not indices
+        } yield Ok(Json.toJson(bucketPositionsForCubeSize))
+      }
+    }
+
+  def querySegmentIndex(token: Option[String],
```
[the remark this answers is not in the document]
frcroth: Added a comment

app/models/annotation/nml/NmlParser.scala
```diff
@@ -531,4 +550,19 @@ object NmlParser extends LazyLogging with ProtoGeometryImplicits with ColorGener
...
+    for {
+      canHaveSegmentIndex <- fallbackLayerName match {
+        case Some(layerName) =>
+          remoteDatastoreClient match {
+            case Some(rdc) => rdc.hasSegmentIndexFile(organizationName, datasetName, layerName)
+            case None      => Fox.successful(false)
+          }
+        case None => Fox.successful(true)
+      }
+    } yield Some(canHaveSegmentIndex)
```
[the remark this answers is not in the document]
frcroth: Changed it.

webknossos-datastore/app/com/scalableminds/webknossos/datastore/services/SegmentIndexFileService.scala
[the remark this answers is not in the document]
fm3: Ok we found out that it is indeed the metadata field `dtype_bucket_entries`. @frcroth please add an assertion that the value of that metadata field is indeed uint16 so that we can get a readable error message otherwise, and not undefined behavior by reading the wrong bytes.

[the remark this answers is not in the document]
fm3: I changed my mind, this can be a quick follow-up for later https://github.com/scalableminds/webknossos/issues/7666 as there are currently no non-uint16 files and we should merge this PR ASAP. I’ll do some more testing all around tomorrow.

frontend/javascripts/admin/admin_rest_api.ts
```diff
@@ -949,42 +949,48 @@ export function getNewestVersionForTracing(
...
+
 export function getSegmentVolumes(
   requestUrl: string,
   mag: Vector3,
   segmentIds: Array<number>,
   additionalCoordinates: AdditionalCoordinate[] | undefined | null,
+  mappingName: string | null | undefined,
 ): Promise<number[]> {
+  let mappingNameNoEmptyString = mappingName;
+  if (mappingName != null && mappingName.length < 1) mappingNameNoEmptyString = null;
```
philippotto, line 973:
```suggestion
  const mappingNameNoEmptyString = mappingName != null && mappingName.length < 1 ? null : mappingName;
```

do you know in which case the mapping name was an empty string? I would have thought that this should not happen :thinking:

frontend/javascripts/admin/admin_rest_api.ts
```diff
@@ -949,42 +949,48 @@ export function getNewestVersionForTracing(
...
-
 export function getSegmentBoundingBoxes(
   requestUrl: string,
   mag: Vector3,
   segmentIds: Array<number>,
   additionalCoordinates: AdditionalCoordinate[] | undefined | null,
+  mappingName: string | null | undefined,
 ): Promise<Array<{ topLeft: Vector3; width: number; height: number; depth: number }>> {
+  let mappingNameNoEmptyString = mappingName;
+  if (mappingName != null && mappingName.length < 1) mappingNameNoEmptyString = null;
```
philippotto, line 990: same here

frontend/javascripts/oxalis/view/right-border-tabs/segments_tab/segment_statistics_modal.tsx
```diff
@@ -130,18 +131,30 @@ export function SegmentStatisticsModal({
     async () => {
       await api.tracing.save();
       if (requestUrl == null) return;
+      const maybeVolumeTracing =
+        tracingId != null ? getVolumeTracingById(tracing, tracingId) : null;
+      const maybeGetMappingName = () => {
+        if (maybeVolumeTracing?.mappingName != null) return maybeVolumeTracing?.mappingName;
```
philippotto, line 137:
```suggestion
        if (maybeVolumeTracing?.mappingName != null) return maybeVolumeTracing.mappingName;
```

the second ? should not be necessary due to the first ?.

philippotto (commented):
front-end looks good :) only left some minor comments.

frontend/javascripts/admin/admin_rest_api.ts
dieknolle3333: no, I don't. I can leave it out if we can rely on it, @fm3 just told me that it could cause problems if it was an empty string

philippotto: I think an empty string should not happen. if it does, it's a bug we should fix :) so, let's assume for now that an empty string won't happen.

fm3:
I noticed that there are a few things that don’t fully work yet. I’ll try to add some commits fixing things as far as I can this week.

 - volume annotation uploads, that bug also existed on master (I think I fixed that now. Some futures weren’t awaited leading to too-parallel processing which does not work with the mutable buffer)
 - mappings (the mapped data is loaded, but the bucket positions are not fetched for the mapped positions)

I’ll try to fix the mapping support this week too.

I’m dumping my test list and todos below:

TODO
 - Frontend: send mappingName also to datastore
 - Backend: fix upload parallelism (await future in withBucketsFromZip)
 - Backend DS: respect mapping in getBucketPositions
 - Backend DS: respect mapping in querySegmentIndex + getSegmentIndex
 - [done] Backend: fix duplicate in fallback layer case
 - [done] Backend: fix awaiting bucket storing fox in zip unpacking for single file upload
 - [done] Backend: fix awaiting bucket storing fox in zip unpacking for multi-file upload
 - test that also with zarr3

TEST

- view mode:
  - with file:
     - without mapping:
       - query volume + bbox 
        - [x] from context menu
        - [x] from segments tab
     - with mappping:
       - query volume + bbox 
         - [ ] from context menu
        → frontend does not send mappingName param
        → backend getBucketPositions does not seem to respect mapping
     - [ ] from segments tab
        → frontend does not send mappingName param
   - without file
     - [x] options should not be there


annotation without fallback layer:
 - [x] brush single id across multiple buckets
 - [x] overwrite it completely in one of them
 - [x] test with ND
 - [x] duplicate
 - [ ] merge by uploading multiple
 - [x] download + reupload
 - [ ] download + reupload (zarr)
 - [x] merge by modal
 - [x] delete segment action + re-brush that id

annotation with fallback layer:
 - [x] brush single id across multiple buckets
 - [x] overwrite it completely in one of them
 - [x] duplicate
     → had to fix hasSegmentIndex call in TSRemoteDataStoreClient
 - [ ] download + reupload
    → got wrong numbers, probably the same problem as without fallback layer
 - [ ] merge by modal
      → ok merging with fallback layer seems broken anyway due to id remapping
 - [ ] merge by uploading multiple
      → ok merging with fallback layer seems broken anyway due to id remapping
   with static mapping:
     - [ ] query id before changing it
         → Looks like querySegmentIndex does not resect passed mappingName
     - [ ] brush single id across multiple buckets
     - [ ] overwrite it completely in one of them
  with editable mapping:
     - [ ] query id before changing it
     - [ ] brush single id across multiple buckets
     - [ ] overwrite it completely in one of them

frontend/javascripts/oxalis/model/sagas/dataset_saga.ts
```diff
@@ -154,22 +157,27 @@ export function* watchZ1Downsampling(): Saga<void> {
...
 export function* ensureSegmentIndexIsLoaded(): Saga<void> {
-  function* maybeFetchHasSegmentIndex(): Saga<void> {
-    const visibleSegmentationLayer = yield* select((state) => getVisibleSegmentationLayer(state));
-    const maybeIsSegmentIndexAvailable = yield* select((state) =>
-      getMaybeSegmentIndexAvailability(state.dataset, visibleSegmentationLayer?.name),
+  function* maybeFetchHasSegmentIndex(action: EnsureSegmentIndexIsLoadedAction): Saga<void> {
+    const { layerName } = action;
+    const dataset = yield* select((state) => state.dataset);
+    if (layerName == null) return;
+    const concernedSegmentationLayer = yield* call(getLayerByName, dataset, layerName);
```
philippotto, line 164: I'd rename it to just `segmentationLayer`. "Concerned" sounds a bit concerning :sweat_smile:

philippotto (approved):
front-end looks good, too :) only left a comment about one variable name..

PR #7449, opened 2023-11-21
Fix ad-hoc meshing with agglomerates and cumsum.json

fm3:
Ad-Hoc meshing can request data from bboxes that are not aligned with buckets, and thus also not aligned with the segmentation chunks described in the cumsum.json.

This lead to an edge case where the cumsum-json-based agglomerate mapping had to gather the correct segmentation chunks that the requested bbox intersects. This is supported, but there was a bug in that logic, which is now fixed in this PR. The currDimensions used to advance to the next chunk bbox in each dimension has to be reset before each inner loop starts, so that the correct topleft for the next box can be found.

Other small cleanups: 
 - converting the requested box to mag1 performed unnecessary arithmetic, with potential rounding errors
 - some renamings.
 - removed `.toLong` on things that are already Long

### Steps to test:
- Request ad-hoc mesh with agglomerate mapping (with a cumsum.json) for a segment that goes across multiple segmentation chunks in the oversegmentation. Should not fail

------


daniel-wer (approved):
Very nice, worked well during my tests :+1:

PR #7450, opened 2023-11-21
CSP Followup

daniel-wer:
Followup for https://github.com/scalableminds/webknossos/pull/7367 (see context in [Slack](https://scm.slack.com/archives/C5AKLAV0B/p1700484578206959)).

Use new configuration syntax: https://www.playframework.com/documentation/2.8.x/CspFilter#Deprecation-of-SecurityHeaders.contentSecurityPolicy.
Add comments for why certain policies are included and which values need to be added/removed in production.
Disable CSP for weblium proxied pages.

- https://cspfollowup.webknossos.xyz

### Steps to test:
- Check that the Content-Security-Policy header is included in the response headers
- Test functionality that loads scripts, images, fonts, etc. or that connects to other web pages

### TODOs:
- [x] Enable CSP configuration. @fm3 Currently, the header is not actually set, although I've followed the docs. My best guess is this: https://stackoverflow.com/questions/76627420/play-framework-play-filters-enabled-cannot-work-if-play-http-filters-is-set. Since play.http.filters is set to "com.scalableminds.webknossos.datastore.Filters", play.filters.enabled seems to be ignored. Could you advise what the best way to fix that would be or maybe even fix it yourself?
- [x] Disable CSP headers for weblium proxied pages
- [x] Fix new issues in browser

### Issues:
- fixes #

------


conf/application.conf
```diff
@@ -17,21 +17,40 @@ play {
...
-    # Development Content Security Policy
-    # In addition to the production policies, the development policy contains the script-src 'unsafe-inline' (to make the React Dev Tools work in Firefox) and the connect-src 'ws://localhost:9002' (webpack dev-server).
-    contentSecurityPolicy = "default-src 'self'; script-src 'self' 'unsafe-inline' 'unsafe-eval' *.googletagmanager.com  app.olvy.co webknossos.olvy.co *.gstatic.com; style-src 'self' 'unsafe-inline' *.gstatic.com; connect-src 'self' ws://localhost:9002 app.olvy.co api.github.com gist.github.com *.google-analytics.com api.airbrake.io; frame-src webknossos.olvy.co; img-src 'self' data:"
-    contentTypeOptions = null
-    frameOptions = null
-    xssProtection = null # Not supported on modern browsers
+  filters {
+    # Note that the CSP filter is not enabled. We manually add this header ourselves for the
+    # main view, using the following config (note that currently only directives are used, not nonce or hashes)
```
MichaelBuessemeyer, line 26: > not nonce or hashes

But line 32 contains a hash?! This way of setting the CSP is simply more structured as it breaks the CSP into is directive parts that writing everything at once. Or am I misunderstanding something? :thinking:

conf/application.conf
```diff
@@ -17,21 +17,40 @@ play {
...
+      style-src = "'self' 'unsafe-inline' *.gstatic.com"
+      # In production ws://localhost:9002 should be removed as it is only needed to make the webpack dev-server work.
+      # api.github.com and gist.github.com are needed for the user script functionality.
+      # All datastore URLs need to be included here, for example: *.webknossos.org demo.data1-brain.esc.rzg.mpg.de
+      connect-src = "'self' ws://localhost:9002 app.olvy.co api.github.com gist.github.com *.google-analytics.com api.airbrake.io"
+      frame-src = "'self' webknossos.olvy.co"
+      # *.voxelytics.com is needed for images embedded in Voxelytics reports.
+      img-src = "'self' data: *.googletagmanager.com *.voxelytics.com"
+      media-src = "*.webknossos.org"
+      font-src = "'self' data:"
```
MichaelBuessemeyer, line 43: LGTM

I assume, that these two directives were missing / causing errors in production?

MichaelBuessemeyer (commented):
Thanks for this follow up pr @daniel-wer and @fm3.

I have a few questions / remarks regarding the changes.
- This pr only enables the CSP for the `WkorgProxyController`, correct? What about dedicated hostet instances? Do these also use this controller or will there be no CSP defined for these wk instances?

- I am no not very familiar with the Scala code / Scala at all. But to me, the code might read as only the main webpage of wk being hosted with the CSP headers. But we would want this on every js / creating a new context. Maybe this is already how it is configured and it simply seems wrong to me. I wanted to mention this nonetheless. -> To me it seems that the webworkers do not have a CSP set. Afaik they are loaded with the main bundle but the request route of that bundle doesn't have a CSP set. According to the post https://stackoverflow.com/questions/37819249/what-does-it-mean-to-set-a-content-security-policy-on-response-to-a-non-html-req/37819481#37819481, the CSP should be set for every route that creates a new context and the main bundle loads webworkers and executes them. This is why I think that the CSP might also be needed for the main bundle. But maybe I am wrong here. I'll simply test this soon by requesting a denied domain from a web worker context.

I will continue the review on Wednesday.

conf/application.conf
fm3: There is a dedicated setting for hashes https://www.playframework.com/documentation/2.8.x/CspFilter#Configuring-CSP-Hashes which we don’t use here. However, it looks like only using the `directives` setting results in the header being set as we wanted, including the hash in line 32.

fm3:
Thanks for having a look!

> This pr only enables the CSP for the WkorgProxyController, correct? What about dedicated hostet instances? Do these also use this controller or will there be no CSP defined for these wk instances?

Yes, this Controller is always used to serve the main view. When the domain is not wkorg, it skips the proxying and serves the main view always. (I’m open for better naming suggestions 😅 )

> But we would want this on every js / creating a new context.

I think you’re right, this is missing. I’ll add that.

fm3:
I added the CSP header now also for webworker routes. Note that this will not work in development mode (as the assets are served by webpack there, not play). Should be testable in a dev deployment, though.

@daniel-wer could you comment on the discussion about directives vs hashes? Are there any changes needed there?

daniel-wer:
> @daniel-wer could you comment on the discussion about directives vs hashes? Are there any changes needed there?

As far as I can see, the hashes key is just a slightly more descriptive and reusable way of declaring those hashes. Our method of directly including them in the respective directives works just as well.

However, during a test of the dev instance I noticed that every request has a CSP header (again? - I think this worked at some point, but I'm not 100% certain), even assets that are not workers or api requests, so I'm afraid the weblium pages would fail again :/

conf/application.conf
daniel-wer: Yes, the media-src is for embedded videos from docs.webknossos.org and the font-src is needed for the dark theme (on Mac only for some reason).

fm3:
> As far as I can see, the hashes key is just a slightly more descriptive and reusable way of declaring those hashes. Our method of directly including them in the respective directives works just as well.

Great, then I guess the changes here are complete. Could you maybe update the migration guide, mentioning the new config keys? Then this should be good to go (or be further reviewed by @MichaelBuessemeyer )

MichaelBuessemeyer (approved):
> Yes, this Controller is always used to serve the main view. When the domain is not wkorg, it skips the proxying and serves the main view always. (I’m open for better naming suggestions 😅 )

I am not familiar with the code. But from what you explained maybe "ProxyWkorgOnlyController" could highlight, that the wkorg instance is only proxied?  

> I added the CSP header now also for webworker routes. Note that this will not work in development mode (as the assets are served by webpack there, not play). Should be testable in a dev deployment, though.

Thanks a lot for fixing this :+1:. I can also confirm, that it works. :rocket: 

I'd say this pr is ready to merge :)

MichaelBuessemeyer:
> I'd say this pr is ready to merge :)

Don't know whether you want someone to check the backend changes @fm3

PR #7348, opened 2023-09-19
Add Render Animation Worker Job

fm3:
This PR adds the frontend and backend feature for starting a animation job of the current annotation / dataset.

image

### URL of deployed dev instance (used for testing):
- https://___.webknossos.xyz

### Steps to test:
- Start a worker with animation job (https://github.com/scalableminds/webknossos/issues/7338)
- Test in View mode and for an annotation
- For "Menu" in navbar select "Create Animation"
- Play around with available options to create various animations
- Restrict your orga to "Basic" plan and check plan enforcement

### Issues:
- fixes #7338

------
- [x] Updated [changelog](../blob/master/CHANGELOG.unreleased.md#unreleased)
- [x] Removed dev-only changes like prints and application.conf edits
- [x] Considered [common edge cases](../blob/master/.github/common_edge_cases.md)

PR #7452, opened 2023-11-22
Enable video animations for all users

hotzenklotz:
This PR enables the animations feature for all users and add a small example video to the modal.

image

### Steps to test:
- None

### Issues:
- Contributes to #7348 

# DO NOT MERGE YET
------


normanrz (commented):
Code looks good :+1:
Since we only allow 2 jobs to be run in parallel and these videos take up to 12h to render. I am a bit worried that the animations will starve out other more important jobs such as data import conversion. Same problem with infer_neurons.
I think we might need to restrict these heavy jobs to only one concurrent and open up more slots for import. cc @fm3

fm3:
Yeah, makes sense. That is now tracked in #7458

Issue #7339, opened 2023-09-18
SAM for squiggly lines


philippotto:
## Detailed Description
Add a new tool (or adapt an existing one) so that the user can use SAM with "squiggly lines" (e.g., an outline) to annotate a segment.

I think, @valentin-pinkau can provide details about how to use the embedding to achieve the above. This issue should be frontend-only.

## Context

* [ ] Specific to long-running jobs (set `jobsEnabled=true` in `application.conf`)
* [ ] Specific to webknossos.org (set `isDemoInstance=true` in `application.conf`)

MichaelBuessemeyer:
Note: It looks like https://github.com/facebookresearch/segment-anything/blob/main/notebooks/onnx_model_example.ipynb under Example mask input shows how to do the inference with a mask

PR #7459, opened 2023-11-27
Add Free Area selection to AI Quick Select (SAM squiggly lines)

MichaelBuessemeyer:
This PR adds a new free area selection mode to the AI quick selection. It uses a free line / area draw tool as already used by the area measurement or trace selection tool. The selected area is then transformed into a mask and input into the SAM prediction. An example of how the mask should work can be found in this [notebook](https://github.com/facebookresearch/segment-anything/blob/main/notebooks/onnx_model_example.ipynb) under "Example mask input". 

It does not explicitly state what area the mask covers of the embedding put into the prediction. Still, I figured that the area covered by the mask is the area between the input coordinates (in the notebook `input_point` and in our code in `quick_select_ml_saga.ts` the coordinates are called `onnxCoord`). This seemed to work in the first free tries.

- https://samsquigglylines.webknossos.xyz/

### Steps to test:
- ~~I suggest testing locally as the SAM setup is easier this way:~~
  - ~~In the `application.conf` configure `segmentAnything` correctly to setup the SAM backend and set `segmentAnythingEnabled` to true.~~
- Feel free to test the quick select on the dev instance. SAM is enabled on these instances by default :tada: 
- Open any annotation
- Select the quick select tool
- A new button group should open up once the AI version is selected. Select the second "free area" version.
- Draw around like with the area measurement or trace tool. The tool should predict the encircled area and not a rectangular one.
- Play a little around with the tool and also test the rectangle version to confirm it is still working as expected.  

### TODOs:
- [x] Only show the tool switch buttons when the AI version is selected
- [x] Adjust icons of tool switch buttons
- [x] Keep the main tool button active when sub (are-) variation is selected.
    - [x] Also do this for the new measurement tool.  
- [x] Fix tests :/
- [ ] Test for edge cases
- [x] Add a better mouse pointer for the area select variation of the quick select

### Suggested Refactorings:
- [x] Rename `VolumeLayer` to `TemporaryVolumeAnnotation`.

### Issues:
- fixes [#7339](https://github.com/scalableminds/webknossos/issues/7339)

------


frontend/javascripts/oxalis/controller/combinations/tool_controls.ts
```diff
@@ -785,6 +792,139 @@ export class QuickSelectTool {
...
+        if (!this.isDrawingBounds || !this.volumeLayer) {
+          return;
+        }
+        if (
+          quickSelectAreaGeometry.vertexBuffer.getLength() <= 2
+          // || state.uiInformation.quickSelectState === "inactive"
+          // Dont know why this check is here ^
+        ) {
+          cleanUp();
+          return;
```
MichaelBuessemeyer, line 879: @philippotto Sorry, but I do not clearly understand the purpose of this check. I copied this from the rectangle variation of the tool. Is this check necessary here? What does the state save / what for is `quickSelectState`?

MichaelBuessemeyer:
@philippotto Just FYI: SAM is already enabled by default on the dev instances (which I didn't know) -> Thus testing on the dev instance is easily possible :tada:.

frontend/javascripts/oxalis/controller/combinations/tool_controls.ts
philippotto: Yes, this is not super clear. I also had too look it up.
Quoting from the store:
```
readonly quickSelectState:
    | "inactive"
    | "drawing" // the user is currently drawing a bounding box
    | "active"; // the quick select saga is currently running (calculating as well as preview mode)
```

So, in general, it should not be `inactive` when the mouse button is lifted, since mouseDown should have made the state not inactive.
However, I think, the var might be `inactive` in the following cases:
- the tool is canceled with `escape` while the mouse button is pressed. then, it is lifted.
- the mouse is pressed in another tool and while it's being pressed, the tool is changed via a keyboard shortcut. then, the mouse button is lifted.
- a bug/error occurs somewhere.

Feel free to add a comment to the code to make it easier to understand :)

MichaelBuessemeyer:
Using the mask and getting reasonably good results seems to be hard. Discussion and thoughts about this can be found in this issue: https://github.com/facebookresearch/segment-anything/issues/169

philippotto:
> Using the mask and getting reasonably good results seems to be hard. Discussion and thoughts about this can be found in this issue: [facebookresearch/segment-anything#169](https://github.com/facebookresearch/segment-anything/issues/169)

Interesting. Did you read this?

> As a trick to bypass this I extract a grid of points from the mask and pass it to SAM instead - the results are much better than the minor changes provided by using the mask_input.

Sounds like an easy thing to try? However, I'm wondering whether these points are interpreted as "known to be within the segment". In that case the user should paint the squiggly lines always within the segment.
Maybe it would make more sense to have a brush tool then and the user's task would be to paint inside the segment instead outlining it. However, we wouldn't need it for testing the grid-approach.

cc @valentin-pinkau in case he's not already reading here.

MichaelBuessemeyer:
> Maybe it would make more sense to have a brush tool

Valentin already suggested this. I agree, that the first version does not need the brush to test the viability of this approach.

> However, I'm wondering whether these points are interpreted as "known to be within the segment". In that case, the user should paint the squiggly lines always within the segment.

I think it is also fine to draw the outlines of the segment and even a little outside the borders of the segment when a distance transform is used for sampling points within and outwards. This is already the approach taken by micro sam, when the outlines are supported by points (which is not enabled by default, instead a bounding box is used as additional input to sam besides the mask)

MichaelBuessemeyer:
I did some slight testing / comparing the current results with micro sam & napari. In summary, they preform equally good / bad:

Napari: The drawn outlines of the cells I annotated with naparis lasso tool are still surrounded by green lines.
image

WK:
image

Both images show that all cells except for the top right and lower left one are segmented pretty good. But wk and napari fail to correctly segment the other two cells and have some outside areas annotated as well :/. I drew about the same outlines in wk as shown the napari screenshot.

MichaelBuessemeyer:
Closing as this feature is not noticeably better than the already existing rectangular SAM quick-select

PR #7460, opened 2023-11-27
Disable unused-import warnings in sbt console REPL

### LINKED ISSUE #7454: Don't show Unused warnings in sbt console REPL
When debugging something using the sbt shell / console, if something is imported this will result in unused warnings for every subsequent command entered, which is very annoying because it unnecessarily expands the output.

Instead, hide these warnings for the REPL.

fm3:
- fixes #7454


fm3:
@frcroth does this work for you? :)

frcroth (approved):

PR #7461, opened 2023-11-27
Release 23.12.0

philippotto:


fm3 (commented):
Maybe we should wait for the CSP changes of #7450

Otherwise, LGTM

philippotto:
Are you sure? https://github.com/scalableminds/webknossos/commit/71d59f75f02b7d5cb1a5b0fd8e1a7d27317c1578 and https://github.com/scalableminds/webknossos/commit/4487180545dcb1a2e31ac077ab1f675f2a02e989 didn't sound like bug fixes to me. #7469 just fixes deprecation warnings (while doing so it switches to another antd interface which might pose a bigger risk than a win when playing the devil's advocate).

fm3 (approved):
Ok fair enough. I kinda thought of the two backend PRs as bugfixes, but it depends on the perspective, I guess. But yeah, feel free to go ahead with this :)

PR #7462, opened 2023-11-28
Make jobs readable for admins+teamManagers+superusers

### LINKED ISSUE #7447: Allow admins to download exported files from all jobs of orga
Currently, users only get to download their own jobs’ exports.

Admins should be allowed to download all jobs’ exports of their orga, if they have the links/ids.

Optional: maybe superusers should even get to download exports across orgas.

fm3:
The job list should still only show the jobs of the requesting user. But if an admin or teamManager or superUser hast the id/link to a job by one of their managed users, they should be able to view its properties (GET api/jobs/:id) and access exported files.

### Steps to test:
- Enable jobs (`yarn enable-jobs`), set up different user/admin combinations
- As the user, run some jobs
- Check that everyone can see exactly those jobs they are supposed to see
- Also check export download links (for this you need a webknossos-worker. I did test it, though, and this uses the same access check as the simple GET api/jobs/:id, which can be tested without a worker)

### Issues:
- fixes #7447

------


app/models/job/Job.scala
```diff
@@ -126,11 +126,28 @@ class JobDAO @Inject()(sqlClient: SqlClient)(implicit ec: ExecutionContext)
...
+      OR
+      ((SELECT u._organization FROM webknossos.users_ u WHERE u._id = _owner) in (select _organization from webknossos.users_ where _id = $requestingUserId and isAdmin))
+      OR
+      ($requestingUserId in
+        (
+          select u._id
+          from webknossos.users_ u join webknossos.multiUsers_ m on u._multiUser = m._id
+          where m.isSuperUser
+        )
+      )
```
frcroth, line 142: Please be consistent on capitalization of sql operators

app/models/job/Job.scala
```diff
@@ -126,29 +126,46 @@ class JobDAO @Inject()(sqlClient: SqlClient)(implicit ec: ExecutionContext)
     }
 
   override protected def readAccessQ(requestingUserId: ObjectId) =
+    q"""
+      _owner = $requestingUserId
+      OR
+      (_owner IN (select _user FROM webknossos.user_team_roles WHERE _team IN (SELECT _team FROM webknossos.user_team_roles WHERE _user = $requestingUserId AND isTeamManager)))
```
frcroth, line 132:
```suggestion
      (_owner IN (SELECT _user FROM webknossos.user_team_roles WHERE _team IN (SELECT _team FROM webknossos.user_team_roles WHERE _user = $requestingUserId AND isTeamManager)))
```

frcroth (approved):

PR #7463, opened 2023-11-28
Finer Control Over Which Worker Does What

### LINKED ISSUE #7458: Finer-grained maxparalleljobs
## Detailed Description
Workers have a `maxparalleljobs` attribute in the database. Very long-running jobs (e.g. render_animation or infer_neurons) could potentially block the workers for a long time. This is not desirable as other classes of jobs are more high priority (e.g. convert_to_wkw or export_tiff). It would be good to have some priority system, for example, 1 maxparalleljob for low-pri and 3 maxparalleljobs for high-pri jobs.

fm3:
The Worker postgres table now has new columns to control job scheduling:
```
maxParallelHighPriorityJobs INT NOT NULL DEFAULT 1,
maxParallelLowPriorityJobs INT NOT NULL DEFAULT 1,
supportedJobCommands VARCHAR(256)[] NOT NULL DEFAULT array[]::varchar(256)[]
```
This means that in a multi-worker setup, you can select which worker should run which jobs. Also, you can allow low-pri jobs to be scheduled with less parallelity than high-pri jobs.

### Steps to test:
- Compare worker setup instructions at https://github.com/scalableminds/voxelytics/tree/master/voxelytics/worker
- With the new schema, adapt your postgres workers table to support different job commands
- Should be distributed to your workers accordingly

### TODOs:
- [x] support parallelity for high and low priority jobs
- [x] support list of supported job commands per worker
- [x] add possibility for frontend to query which job commands are available for a datastore
- [x] sql migration

### Follow-Up
 - #7485

### Issues:
- fixes #7458

------


app/models/job/Job.scala
```diff
@@ -254,14 +266,15 @@ class JobDAO @Inject()(sqlClient: SqlClient)(implicit ec: ExecutionContext)
...
-    for {
-      _ <- run(
-        query.withTransactionIsolation(Serializable),
-        retryCount = 50,
-        retryIfErrorContains = List(transactionSerializationError)
-      )
-    } yield ()
-  }
+      for {
+        _ <- Fox.successful(logger.info("reserve next job"))
```
frcroth, line 270: Intentional?

app/models/job/JobCommand.scala
```diff
@@ -5,6 +5,15 @@ import com.scalableminds.util.enumeration.ExtendedEnumeration
...
+   * - Decide if it should be a highPriority job
+   * - Add it to the dbtool.js command enable-jobs so it is available during development
+   * - Add it to the migration guide (operators need to decide which workers should provide it)
+   */
+
   val compute_mesh_file, compute_segment_index_file, convert_to_wkw, export_tiff, find_largest_segment_id,
   globalize_floodfills, infer_nuclei, infer_neurons, materialize_volume_annotation, render_animation = Value
+
+  val highPriorityJobs: Set[Value] = Set(convert_to_wkw, export_tiff)
+  val lowPriorityJobs: Set[Value] = values.diff(highPriorityJobs)
```
frcroth, line 18: This is general and should not be configurable?

app/models/job/Worker.scala
```diff
@@ -37,29 +40,34 @@ class WorkerDAO @Inject()(sqlClient: SqlClient)(implicit ec: ExecutionContext)
...
       parsed <- parseFirst(r, "key")
     } yield parsed
 
-  def findOneByDataStore(dataStoreName: String): Fox[Worker] =
+  def findAllByDataStore(dataStoreName: String): Fox[List[Worker]] =
     for {
       r: Seq[WorkersRow] <- run(
-        q"select $columns from $existingCollectionName where _dataStore = $dataStoreName".as[WorkersRow])
-      parsed <- parseFirst(r, "dataStoreName")
+        q"SELECT $columns FROM $existingCollectionName where _dataStore = $dataStoreName".as[WorkersRow])
```
frcroth, line 69:
```suggestion
        q"SELECT $columns FROM $existingCollectionName WHERE _dataStore = $dataStoreName".as[WorkersRow])
```

app/models/job/Job.scala
fm3: nope, good catch, removed it now :)

app/models/job/JobCommand.scala
fm3: I’d say for now, yes

frcroth (approved):

hotzenklotz:
@fm3 If I see this correctly then the PR only adds the `jobsSupportedByAvailableWorkers` info to the endpoint `/api/datasets/<organization>/<dataset_name>`. 

What about the `/api/features` route? Some worker jobs can only be triggered in admin menus, e.g. dataset upload/conversion or finding the max segment ID.

fm3:
Good point!

I now pushed that commit on your branch of PR https://github.com/scalableminds/webknossos/pull/7591 

In there I moved both fields `jobsEnabled` and `jobsSupportedByAvailableWorkers` from the dataset json to the dataStore json.

I cannot change the `/api/features` route content since that is directly from the application.conf, whereas this information is worker (and thus datastore) specific.

Note that the dataset json does contain a full copy of the datastore json (dataset.dataStore) so all points that previously accessed `dataset.jobsEnabled` can now, as far as I understand, instead access `dataset.dataStore.jobsEnabled`.

In dataset upload, you can use the info from the selected datastore.

When finding the largestSegmentId, I would have expected that you can use the dataset-specific route, as this concerns a specific dataset.

Hope this helps!

PR #7466, opened 2023-11-29
Improve speed of user list query

### LINKED ISSUE #5269: Speed up user list request
The user list request is slow, presumably becuase of many individual database requests during publicWrites. We should analyze that and look at what we can cache or combine there.

Related: #5268

frcroth:
- https://___.webknossos.xyz

For the two test users this results in about 30% faster requests for me.

### Steps to test:
- Functionality should be tested with automatic tests

### TODOs:
- [x] Remove compact parameter as non-compact could be deleted.

### Issues:
- fixes #5269 

------


app/models/user/User.scala
```diff
@@ -168,11 +225,77 @@ class UserDAO @Inject()(sqlClient: SqlClient)(implicit ec: ExecutionContext)
...
+  implicit def GetResultUserCompactInfo(implicit e0: GetResult[String],
+                                        e1: GetResult[java.sql.Timestamp],
+                                        e2: GetResult[Boolean],
+                                        e3: GetResult[Option[String]]): GetResult[UserCompactInfo] = GetResult { prs =>
+    import prs._
+    // format: off
+    UserCompactInfo(<<[String],<<[String],<<[String],<<[String],<<[String],<<[String],<<[Boolean],<<[Boolean],
+      <<[Boolean],<<[Boolean],<<[String],<<[String],<<[String],<<[String], <<[String],<<[java.sql.Timestamp],<<[String],
+      <<[String],<<[String],<<[String],<<[java.sql.Timestamp],<<?[String],<<[Boolean],<<[Boolean]
+    )
```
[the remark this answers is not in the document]
fm3: Hmm, since this variant also needs to list all the data types, it does not appear to me to bring that much benefit. So I’d vote to keep it unified and use what we had before. Do you agree?

fm3 (changes requested):
I think it would be good to get that isEditableBy logic into SQL after all. Maybe add some comments to ensure we keep those in sync in the future. But 1 or n queries for n users will make a bigger difference for a longer user list (I assume you didn’t have more than 10 users when testing?)

Also, could you make use of `ARRAY_AGG` and the `parseArrayLiteral` function that was introduced in #7479 ? (Also for the other recently added compact queries). Then we don’t have this `null`able handling code with `Option(x)` distributed all over our codebase.

app/models/user/User.scala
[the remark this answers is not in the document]
frcroth: Unfortunately, I can not just use the one from e.g. listing Annotations, because in Scala tuples can only have 22 elements (here 24 are needed) and thus the row can not be parsed as a tuple. :(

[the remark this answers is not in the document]
fm3: Ah well ok then let’s keep your variant :D

app/models/user/UserService.scala
```diff
@@ -368,6 +372,48 @@ class UserService @Inject()(conf: WkConf,
     }
   }
 
+  def publicWritesCompact(user: User, requestingUser: User, userCompactInfo: UserCompactInfo): Fox[JsObject] =
+    for {
+      _ <- Fox.successful(())
+      teamsJson = parseArrayLiteral(userCompactInfo.team_ids).indices.map(
+        idx =>
+          Json.obj(
+            "id" -> parseArrayLiteral(userCompactInfo.team_ids)(idx),
```
fm3, line 381: Could the parseArrayLiteral calls be moved into `GetResultUserCompactInfo`?

frcroth: Not sure how. The << thing does not seem to be very flexible, not allowing me to just store it somewhere. What would be possible is to create another "raw" case class for storing all the array literal and returning a "parsed" case class so the service does not use parseArrayLiteral

fm3: I see. Then let’s keep it like you did it for now (just one case class) but I’d ask to rename the fields to make this obvious for the usages. `userCompactInfo.teamIdsAsNullableArrayLiteral` or something like that. What do you think?

Later we can try to experiment if we can get implicit conversions in there, similar to GetInstant and GetObjectId from SqlTypeImplicits.scala

frcroth: Agree, though it shouldn't be nullable anymore with the array remove stuff parseArrayLiteral

fm3 (approved):

PR #7477, opened 2023-12-05
Fix that failing bucket requests marked buckets as missing on server

### LINKED ISSUE #5387: Retrying failed bucket requests does not work consistently
## Expected Behavior
When being offline, data loading won't succeed. Going back online, should make the data appear.

## Current Behavior
When being offline, data loading won't succeed. After going back online, the black data spots stay. Moving to a new area, will load data for the new position.

## Steps to Reproduce the bug
1. Open a dataset
2. Go offline via the devtools
3. Move somewhere else. Viewports will remain black.
4. Go back online. Viewports will remain black, but should load the data instead.

## Solution
I think, the front-end should be able to figure out that it is currently offline. In that case, the pullqueue could simply block and do nothing. As soon as we recognize, that we are online again, the pullqueue could unblock. I think, this approach makes more sense then simply retrying each request N times (this is done currently, but when one is offline for one minute, the retries are likely to be exhausted). Alternatively, one could use exponential back-off for the retry, but this will probably be less "immediate" in its UX behavior (e.g., after going back online you have to wait for the next interval to expire).

philippotto:
This PR fixes an incorrect handling of failed bucket requests. #4740 added a try-catch around the bucket fetching logic for better error reporting. The reporting itself is fine, but the catch-clause used null-values as a return value. However, null values are reserved for buckets that don't exist on the server. Thus, if bucket requests failed for some reason, the front-end marked these buckets as `missing` and assumed that a re-try is not necessary. This had (at least) two implications:

- Failing bucket requests were not re-tried at all (only when manually reloading a layer or moving somewhere else and then moving back (usually activates GC)) and just remained black (or rendered with fallback data)
- When annotating volume data on buckets for which requests failed, the bucket was assumed to be empty/missing on the server. Thus, existing not-zero voxels (e.g., from fallback or from previous annotations) were lost and overwritten with 0.

Additionally, this PR makes the (now working) retry mechanism a bit less aggressive. This is more of a bandaid fix, since the actual architecture could use a slight refactoring. However, I think this is out of scope for this PR.

- https://fixincorrectmissingbucketstate.webknossos.xyz

### Steps to test:
- scenario 1:
  - view a dataset
  - make tab offline (via devtools)
  - move somewhere else
  - data should be invisible
  - make tab online again
  - move a tiny bit -> buckets should now be requested again
    - ideally, the moving should not be necessary, but I think, this is not critical for this bug fix
- scenario 2:
  - create an annotation with fallback data
  - make tab offline
  - reload the segmentation layer --> should be empty now
  - brush something
  - variant 1:
    - click save --> everything should turn red because the tab is offline
    - make tab online again
    - wait --> saving should recover
    - reload annotation --> everything should look as expected
  - variant 2:
    - click undo
    - make tab online again and follow steps from variant 1
  - variant 3:
    - click undo
    - click redo
    - make tab online again and follow steps from variant 1

### TODOs:
- [ ] re-enable commented lines

### Issues:
- fixes #5387

------


frontend/javascripts/oxalis/model/bucket_data_handling/temporal_bucket_manager.ts
```diff
@@ -26,12 +26,12 @@ class TemporalBucketManager {
     this.loadedPromises.push(this.makeLoadedPromise(bucket));
   }
 
-  pullBucket(bucket: DataBucket): Array<Promise<void>> {
```
philippotto, line 29: the promises weren't used.

frontend/javascripts/oxalis/view/left-border-tabs/layer_settings_tab.tsx
```diff
@@ -983,8 +983,8 @@ class DatasetSettings extends React.PureComponent<DatasetSettingsProps, State> {
   };
 
   reloadLayerData = async (layerName: string): Promise<void> => {
-    await clearCache(this.props.dataset, layerName);
-    this.props.reloadHistogram(layerName);
+    // await clearCache(this.props.dataset, layerName);
```
philippotto, line 986: should be re-added after testing

frontend/javascripts/oxalis/model/bucket_data_handling/pullqueue.ts
```diff
@@ -118,13 +122,41 @@ class PullQueue {
...
+            this.isRetryScheduled = false;
+            this.pull();
+          });
+        }
+      }
     }
   }
 
+  private getRetryDelay(): number {
+    const exponentialBackOff = 25 * 2 ** (this.consecutiveErrorCount / 10);
```
philippotto, line 156: this was just a gut feeling and turned okay during testing, I think. I'm open to suggestions, though.

daniel-wer (approved):
Great find! Worked very well for me :+1: 

> Additionally, this PR makes the (now working) retry mechanism a bit less aggressive. This is more of a bandaid fix, since the actual architecture could use a slight refactoring. However, I think this is out of scope for this PR.

What are you thinking of there? :)

philippotto:
> > Additionally, this PR makes the (now working) retry mechanism a bit less aggressive. This is more of a bandaid fix, since the actual architecture could use a slight refactoring. However, I think this is out of scope for this PR.
> 
> What are you thinking of there? :)

The PullQueue is responsible for fetching buckets, but it relies on being triggered (via `pull`) from other components (but also does it itself in the retry-case). This means that it's hard to predict at what frequency `pull` is called. I think, it would be nice if the PullQueue would be responsible for calling `pull`. It could do so in reaction to `add()` and then take care of throttling etc.

I created https://github.com/scalableminds/webknossos/issues/7480 for this :)

PR #7502, opened 2023-12-13
fixes Uri encoding in OpenGraphService

normanrz:
Don't decode the URL before constructing the URI in OpenGraphService.

### Steps to test:
- Create a sharing link
- Add a `%20` in the database in the longLink (e.g. in the orga part)
- The sharing link should still be resolved ok (but point to a non-existing dataset)

------


philippotto (approved):
works for me :+1:

fm3 (approved):

PR #7510, opened 2023-12-22
Fix that the Segment Statistics were not available in context menus

philippotto:
... of segment groups

- https://___.webknossos.xyz

### Steps to test:
- use the context menu of a segment group in an annotation and check that the statistics are available for:
  - a new annotation without fallback (should have an index) for a 3d (or 2d) dataset
- it should not be shown for nd annotations or annotations with fallback layer

### Issues:
- follow-up to #7394

------


hotzenklotz (approved):
Works.

PR #7512, opened 2023-12-23
Synchronize list focus with active segment

### LINKED ISSUE #7311: Segment list focus is not in sync with active segment
## Detailed Description
When clicking on an entry in the segment list a blueish overlay appears with "No mesh loaded" to highlight your are interacting/focussing with this segment. When clicking on a different segment in one of the viewports focus overlay is not updated accordingly. Neither is the old segment disfocused nor is the new entry highlight.

I constantly get lost because the two are out of sync.

image

## Context

dieknolle3333:
- https://___.webknossos.xyz

### Steps to test:
- create a lot of segments so that one can scroll in the segments view
- click on a segment in the data viewport and ensure that the segments view scrolls and highlights to the corresponding item

### TODOs:
- [x] create context menu entry
- [x] fix issue that only available segments are stored as selected (e.g. after deleting a segment, see commented-out-code). test this by removing a selected segment
- [x] highlight difference between `jump to segment` and `Select segment` in viewport segment dropdown
- [x] make sure the "click segment" action from context menu doesnt harm anything
- [x] actually focus segment in list, so scroll to it via `this.tree.current.scrollTo({ key: selectedElement.key });` with the key being `segment-<id>`
    - [x] fix that scrolling doesnt work for newly added segments 
    -> may be bug
- [x] fix drag and drop in segment list, e.g. to groups (seperate issue: #7527)
- [x] remove logging and read through the rest of the code

### Issues:
- fixes #7311

------


frontend/javascripts/oxalis/view/context_menu.tsx
```diff
@@ -981,12 +1009,13 @@ function getNoNodeContextMenuOptions(props: NoNodeContextMenuProps): ItemType[]
                 },
                 label: (
                   <>
-                    Select Segment ({segmentIdAtPosition}){" "}
+                    Activate Segment ({segmentIdAtPosition}){" "}
```
dieknolle3333, line 1012: adjusting the wording to the context menu in the segments view tab

dieknolle3333:
@philippotto I decided that a multiselection of segment should not be overwritten by a click on a segment. Because both the context menu entry and the segment click are causing the same action, the context menu entry doesnt overwrite it either. Thats probably counterintuitive. Obviously I could build around that, but I just wanted to make sure that we are keeping both ways to focus the segment :) So we are keeping the context menu entry and the click, right? Or should I ask for more opinions first?

dieknolle3333:
there may be edge cases where this context menu entry is displayed although it doesnt make sense, but I think that applies for more than this context menu entry. see #7538. I didnt know what edge exactly I found, but if its easy I will of course fix it in this PR :)

frontend/javascripts/oxalis/view/right-border-tabs/segments_tab/segments_view.tsx
```diff
@@ -209,6 +218,45 @@ const mapStateToProps = (state: OxalisState): StateProps => {
   };
 };
 
+const getOrInitializeSelectedSegmentsOrGroup = (
```
philippotto, line 221: I'd rename this function to `getCleanedSelectedSegmentsOrGroup`, since I think the cleaning up part is more important than the initialization part.

frontend/javascripts/oxalis/view/right-border-tabs/segments_tab/segments_view.tsx
```diff
@@ -209,6 +218,45 @@ const mapStateToProps = (state: OxalisState): StateProps => {
...
+  );
+  const currentGroupIds = currentGroupIdsWithoutRootGroup.add(MISSING_GROUP_ID);
+  const selectedIds = {
+    segments: segments.filter((id) => currentSegmentIds.has(id)),
+    group: group != null && currentGroupIds.has(group) ? group : null,
+  };
+  const haveSegmentsOrGroupsBeenRemovedFromList =
+    selectedIds.segments.length !== segments.length || selectedIds.group !== group;
+  if (haveSegmentsOrGroupsBeenRemovedFromList) {
+    Store.dispatch(
```
philippotto, line 249: cool, that you found a way to keep this logic incoporporated :+1: I think, it would be even cleaner if the potential removal of selected ids would happen in the reducer. However, the current way probably works okay, too.

frontend/javascripts/oxalis/view/context_menu.tsx
```diff
@@ -785,6 +786,28 @@ function getNoNodeContextMenuOptions(props: NoNodeContextMenuProps): ItemType[]
...
+    const layerName = visibleSegmentationLayer.name;
+    if (clickedSegmentId === 0) {
+      Toast.info("No segment found at the clicked position");
+      return;
+    }
+    const selectedSegmentsOrGroup = state.localSegmentationData[layerName].selectedIds;
+    const additionalCoordinates = state.flycam.additionalCoordinates;
+    const numberOfSelectedSegments = selectedSegmentsOrGroup.segments.length;
+    if (numberOfSelectedSegments < 2) {
+      // This action is dispatched because the behaviour is identical click on a segment.
```
philippotto, line 803:
```suggestion
      // This action is dispatched because the behaviour is identical to a click on a segment.
```

frontend/javascripts/oxalis/model/sagas/volumetracing_saga.tsx
```diff
@@ -782,6 +787,26 @@ function* updateHoveredSegmentId(): Saga<void> {
...
+  // If one or zero segments are selected, update selected segments in store
+  // Otherwise, the multiselection is kept.
+  const { segmentId } = action;
+  let layerName: string | undefined;
+  const segmentationLayer = yield* call([Model, Model.getVisibleSegmentationLayer]);
+  layerName = segmentationLayer?.name;
+  if (layerName == null) return;
+  const clickedSegmentId = segmentId;
+  const selectedSegmentsOrGroup = yield* select(
+    (state) => state.localSegmentationData[layerName as string]?.selectedIds,
```
philippotto, line 802: is `as string` really necessary? i wouldn't think so because of the == null check above.

frontend/javascripts/oxalis/view/context_menu.tsx
```diff
@@ -785,6 +786,28 @@ function getNoNodeContextMenuOptions(props: NoNodeContextMenuProps): ItemType[]
...
+    const clickedSegmentId = getSegmentIdForPosition(globalPosition);
+    const layerName = visibleSegmentationLayer.name;
+    if (clickedSegmentId === 0) {
+      Toast.info("No segment found at the clicked position");
+      return;
+    }
+    const selectedSegmentsOrGroup = state.localSegmentationData[layerName].selectedIds;
+    const additionalCoordinates = state.flycam.additionalCoordinates;
+    const numberOfSelectedSegments = selectedSegmentsOrGroup.segments.length;
+    if (numberOfSelectedSegments < 2) {
```
philippotto, line 802: this logic already happens in the new saga (and also doesn't happen when invoking a simple click). therefore, I'd think that it can be removed here, right?

philippotto (changes requested):
great stuff :) I left some feedback, but overall it's already looking pretty good :+1:

frontend/javascripts/oxalis/model/sagas/volumetracing_saga.tsx
dieknolle3333: If I omit it, the typechecker warns that `Type 'undefined' cannot be used as an index type.`

dieknolle3333: I could also use `layerName!`

frontend/javascripts/oxalis/view/right-border-tabs/segments_tab/segments_view.tsx
dieknolle3333: and also the initialization moved to dataset_reducer

frontend/javascripts/oxalis/view/right-border-tabs/segments_tab/segments_view.tsx
```diff
@@ -408,6 +454,25 @@ class SegmentsView extends React.Component<Props, State> {
...
       );
     }
+
+    // Scroll to selected segment.
+    // The selection of the newly added segment, that wasn't in the segment list before,
+    // triggers this function. Technically the segment is now present in the tree due
+    // to the design in the volumetracing_saga, and it can also be found while debugging
+    // this class. But in the scrollTo function, the new segment isn't found right away
+    // in the tree data, thus we need the timeout.
+    setTimeout(() => {
```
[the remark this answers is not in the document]
dieknolle3333: No, you are right, it was working back then. But sometimes it's not working after all, thus I added the time-out again.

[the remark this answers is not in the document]
dieknolle3333: I think it has to do with the tree component, because it is again not found there, although the new segment is found in the tree of this class

philippotto (approved):
Works great for me :)

PR #7517, opened 2024-01-03
Create Organization Directory When Inserting Initial Data

### LINKED ISSUE #7509: Double check that binaryData dir is created on first datastore startup
Logging from #7508 indicates that the binaryData directory was not created on startup. We should double-check that this happens if you follow the normal steps from DEV_INSTALL.md

fm3:
- When inserting initialData, webknossos now creates the organization directory (with parents), so usually `binaryData/sample_organization`.
- Moved around some code to avoid circular dependencies.
- Renamed Folder to Directory in one more place, in order to distinguish it from the in-postgres folder concept.

### Steps to test:
- delete/backup your binaryData directory or clone a fresh webknossos
- refresh postgres schema
- start webknossos with `yarn start --ignore-engines`
- `binaryData/sample_organization` should be created, datasets put there should be served

### Issues:
- fixes #7509

------
- [x] Needs datastore update after deployment **IMPORTANT: update all datastores**

### DIFF
--- a/app/models/organization/OrganizationService.scala
+++ b/app/models/organization/OrganizationService.scala
@@ -4,7 +4,6 @@ import com.scalableminds.util.accesscontext.{DBAccessContext, GlobalAccessContex
 import com.scalableminds.util.tools.{Fox, FoxImplicits, TextUtils}
 import com.scalableminds.webknossos.datastore.rpc.RPC
 import com.typesafe.scalalogging.LazyLogging
-import controllers.InitialDataService
 
 import javax.inject.Inject
 import models.dataset.{DataStore, DataStoreDAO}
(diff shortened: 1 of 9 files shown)


app/models/organization/OrganizationService.scala
```diff
@@ -113,10 +121,9 @@ class OrganizationService @Inject()(organizationDAO: OrganizationDAO,
       _ <- folderDAO.insertAsRoot(organizationRootFolder)
       _ <- organizationDAO.insertOne(organization)
       _ <- teamDAO.insertOne(organizationTeam)
-      _ <- initialDataService.insertLocalDataStoreIfEnabled()
     } yield organization
 
-  def createOrganizationFolder(organizationName: String, dataStoreToken: String): Fox[Unit] = {
+  def createOrganizationDirectory(organizationName: String, dataStoreToken: String): Fox[Unit] = {
     def sendRPCToDataStore(dataStore: DataStore) =
       rpc(s"${dataStore.url}/data/triggers/newOrganizationFolder")
```
frcroth, line 128: Would it be possible to change this route "directory" as well?

frcroth (approved):
Did not test locally, but looks good

PR #7505, opened 2023-12-15
Update antd to version 4.24.15

hotzenklotz:
PR upgrades antd to latest 4.x release as a prerequisite to antd 5. The changelog between the current 4.24.8 and 4.24.15 version is rather unremarkable and did not hint at any major changes. https://4x.ant.design/changelog

I tested most admin view and the tracing interface by clicking around and monkey testing. Nothing came up. 

### URL of deployed dev instance (used for testing):
- https://___.webknossos.xyz

### Steps to test:
- Monkey test

### Issues:
- contributes to #6861 

------
(Please delete unneeded items, merge only when none are left open)
- [x] Updated [changelog](../blob/master/CHANGELOG.unreleased.md#unreleased)
- [ ] Updated [migration guide](../blob/master/MIGRATIONS.unreleased.md#unreleased) if applicable
- [ ] Updated [documentation](../blob/master/docs) if applicable
- [ ] Adapted [wk-libs python client](https://github.com/scalableminds/webknossos-libs/tree/master/webknossos/webknossos/client) if relevant API parts change
- [ ] Removed dev-only changes like prints and application.conf edits
- [x] Considered [common edge cases](../blob/master/.github/common_edge_cases.md)
- [ ] Needs datastore update after deployment

PR #6579, opened 2022-10-21
Virtualized log for vx reports

normanrz:
Adds a virtualized list for rendering logs in vx reports. This is necessary because logs can get up to 100,000s entries long.

### Steps to test:
- Enable voxelytics and log streaming in application.conf (including setting up elasticsearch/opensearch locally)
- Stream some logs to wk
- Load the log tab in the workflow report, scroll around

------
- [x] Updated [(unreleased) changelog](../blob/master/CHANGELOG.unreleased.md#unreleased)
- [x] Ready for review

PR #5597, opened 2021-07-01
Show converting datasets

youri-k:
### URL of deployed dev instance (used for testing):
- https://showconversiondatasets.webknossos.xyz

### Steps to test:
- upload dataset that needs conversion
- go to dashboard
- => see blue alert box which shows the uploaded dataset
- => when upload is complete, alert box should be removed

### Issues:
- fixes #5482 

------
- [x] Updated [(unreleased) changelog](../blob/master/CHANGELOG.unreleased.md#unreleased)
- ~~[x] Updated [(unreleased) migration guide](../blob/master/MIGRATIONS.unreleased.md#unreleased) if applicable~~
- ~~[ ] Updated [documentation](../blob/master/docs) if applicable~~
- ~~[ ] Adapted [wk-connect](https://github.com/scalableminds/webknossos-connect) if datastore API changes~~
- ~~[ ] Needs datastore update after deployment~~
- [ ] Ready for review

PR #7521, opened 2024-01-03
Remove several unused JS dependencies

hotzenklotz:
While debugging the library version conflicts for PR #7505 I ran a check for unused libs with `depcheck`. While it flagged many false positive libs I think it correctly identified these 5 libs as unused:

#### *[Name]*: *[PR where the lib was introduced]*
- `ndarray-convolve`: [Quick Select Tool PR](https://github.com/scalableminds/webknossos/pull/6542)
- `react-sortable-tree-theme-file-explore`:    [Dataset Folder PR](https://github.com/scalableminds/webknossos/pull/6591)
- `strip-ansi`: [Virtual VX Logs](https://github.com/scalableminds/webknossos/pull/6579)
- `@use-it/interval`: [Recent Datasets conversion on Dashboard PR](https://github.com/scalableminds/webknossos/pull/5597)
- `cross-fetch`: [PR to Fix airbrake-js](https://github.com/scalableminds/webknossos/pull/3859)

Hopefully this decreases our bundle size a little bit and helps to avoid version conflicts in the future.

Command to detect unused libs.
```
npx depcheck . --specials=eslint,webpack,prettier,typescript,less,esbuild
```

- https://___.webknossos.xyz

### Steps to test:
- Delete `node_modules` and `yarn install`
- I did tests to verify that:
  - That throwing an Error will trigger a Airbrake notifications
  - That quick select works in AI and non-AI mode
  - That VX logs can still be viewed
  - That skeletons & segment tree components looked "normal"

### Issues:
- None

------


package.json
```diff
@@ -166,7 +165,6 @@
     "classnames": "^2.2.5",
     "color-hash": "^2.0.1",
     "comlink": "^4.3.0",
-    "cross-fetch": "^3.1.5",
```
hotzenklotz, line 169: This lib was introduced to added a polyfill for `airbrake-js`. We have switched to `@airbrake/browser` in the meantime, which seems to work without polyfills.

hotzenklotz:
@philippotto @daniel-wer I already tried to verify as much as possible that we really do not use these libs anywhere in our code or tooling. Yet, I didn't remember why we added some of this in the first-place or why it has become unused. Maybe some of the lib names mean more to you and you can verify that this is safe to remove. Thanks.

package.json
```diff
@@ -154,7 +154,6 @@
     "@types/pixelmatch": "^5.2.4",
     "@types/pngjs": "^6.0.1",
     "@types/three": "^0.142.0",
-    "@use-it/interval": "^1.0.0",
```
hotzenklotz, line 157: We have implemented this ourselves in `libs/react_helpers`

philippotto:
I didn't test this PR (your performed testing looks very good :+1:), but your explanations sound very reasonable and I agree that these packages seem to be unused.

> Hopefully this decreases our bundle size a little bit [...]

I would have hoped that webpack wouldn't include unused bits, but removing unused packages is still a good idea, obviously.

hotzenklotz:
@daniel-wer Do these libs mean anything to you? Do you see any red flags?

daniel-wer:
> @daniel-wer Do these libs mean anything to you? Do you see any red flags?

I don't see any red flags. Thanks for the detailed description and researching where these libs were introduced!

philippotto (approved):
cool, then, let's do it :tada:

Issue #5236, opened 2021-03-04
Show logs for long-running jobs


fm3:
When executing long-running jobs, the user can currently only see the state as one of Pending, Running, Success, Failed.
It would be cool if the user could get access to the job logs, to understand both progress and possible issues.

Things to check: 
- [ ] Don’t include anything secret in these logs
- [ ] should we serve them continuously as they are written? How do kube/circleci/github do this? Or is serving the text-file statically good enough for a first iteration?

fm3:
@normanrz also suggested to just show a high-level error message, especially for the convert_to_wkw job (e.g. "unsupported format" / "runtime error" (e.g. OOM) / "other error")

PR #7523, opened 2024-01-04
No longer show failed conversion jobs as Manual

### LINKED ISSUE #7499: Remove “Manual Pass” for wkw conversion jobs
The conversions fail way less frequently now. We decided to not automatically offer a “manual pass” anymore.

The frontend code should be adapted to always show the job state as is (manualState still has precedence over state, but no more FAILURE→MANUAL replacements.

fm3:
The conversions fail way less frequently now. We decided to not automatically offer a “manual pass” anymore.

Note that the manualState mechanism still works, so if the manualState is set in the database, it is still used.

### Steps to test:
- Set up jobs and a worker
- Upload data that cannot be converted to wkw
- job should be shown as Failure (not “Manual”) in the job list view

### Issues:
- fixes #7499

Follow-up: https://github.com/scalableminds/webknossos/issues/5236


philippotto (approved):
great, lgtm :+1:

PR #7534, opened 2024-01-09
Fix crash in proofreading when no meshes existed before

philippotto:
- https://fixcrashinproofreading.webknossos.xyz

### Steps to test:
- open an annotation with an agglomerate mapping
- select a mapping
- select proofreading tool
- disable automatic mesh loading in toolbar
- merge two segments

### Issues:
- follow-up to #7394

------


daniel-wer (approved):
LGTM :+1:

PR #7536, opened 2024-01-09
Add multi segment drag and drop

### LINKED ISSUE #7527: Support drag and drop of multiple segments
Drag and drop of multiple segments does not work yet since only one of the selected segments is moved e.g. to a new group.

see https://scm.slack.com/archives/C5AKLAV0B/p1704750514217749

dieknolle3333:
fixes #7527 

### Steps to test:
- Go to segments tab and play around with drag and drop.
- Explore the following cases:
    - No segment is selected
    - Multiple segments are selected
    - One segment is selected
    - Selected segment is dragged
    - Not selected segment is dragged

------


philippotto (approved):
great :)

PR #7548, opened 2024-01-12
Show Annotation Segment Counts in Dashboard

### LINKED ISSUE #7431: Add segment counts to dashboard
## Detailed Description
Similar to the skeleton statistics (node count, tree count etc) we should show the segment count of an annotation/task in the various lists and views:

- Dashboard > Annotations
- Admin > Tasks > Sub-View for the annotations per Task 
- (Maybe in the right-hand side info panel of an annotation)

This requires a little bit of frontend work but mostly backend. The segment counts could be cached using a similar structure as for the skeleton nodes.

## Context

fm3:
- https://savesegmentstats.webknossos.xyz

### Steps to test:
- Create annotation, annotate some, dashboard should show tree + segment count
 
### TODOs:
- [x] Frontend
  - [x] When saving volume layer updates, send stats object with current segmentCount
  - [X] Adapt dashboard to aggregate stats for annotation from annotation layers
    - [x] How to display segment count? new icon? do we need more space?
    - ~[ ] Show infos for all layers in tooltip?~ Philipp: Not important for now
- [x] Backend
  - [x] Store stats per annotation layer
  - [x] SQL migration
  - [x] Include stats in annotation layer json
  - [x] How to encode unknown status? empty object?
  - [x] Initialize stats
    - [x] as unknown (`{}` after merge, create from task, upload)
    - [x] as zeroed when creating empty annotation layer

### Issues:
- fixes #7431

------


philippotto:
@fm3 The frontend should send the segmentCount now. The dashboard receives `stats: {}` for the annotations now, but I guess the backend needs to be adapted.

fm3:
Great, thanks! Yes, I moved the stats to the annotation layers inside of the annotation json. They should :tm: already be filled, but I did not test this end-to-end yet.

fm3:
Yeah, seems to work :)

image

Note that the stats in the layers may also still be `{}` for “unknown” (I plan to initialize it with zeroes but I think it’s not easy to get the real values after upload or merge, so I probably won’t do those in this PR).

Do you think the dashboard frontend code can work with this new format?

philippotto:
Cool, I adapted the dashboard now. What's up with the old `stats` property in the actual annotation? Is that always empty now? Should we remove it?

fm3:
Yes, it’s always empty now. Removing it completely may break old libs clients. Feel free to remove it from the frontend types. I plan to remove it from the json in a later version.

fm3:
@philippotto thanks for taking care of the frontend!

Do you think it would be possible to hide the segments if the annotation has no volume layer, and hide the tree count if the annotation has no skeleton layer? This way the user could notice at a glance what kind of annotation they have? (I’m also thinking that we might want to remove the skeleton/volume/hybrid tag from the dashboard in the future, because especially “hybrid” is explained nowhere)

I think the backend is ready for review, feel free to request reviews when you think this is done :)

philippotto:
> Do you think it would be possible to hide the segments if the annotation has no volume layer, and hide the tree count if the annotation has no skeleton layer? This way the user could notice at a glance what kind of annotation they have? (I’m also thinking that we might want to remove the skeleton/volume/hybrid tag from the dashboard in the future, because especially “hybrid” is explained nowhere)

Yes, good points :+1: It's done now.

frontend/javascripts/oxalis/model/accessors/annotation_accessor.ts
```diff
@@ -11,3 +14,112 @@ export function mayEditAnnotationProperties(state: OxalisState) {
...
+}
+
+export function getCombinedStats(tracing: Tracing): CombinedTracingStats {
+  const aggregatedStats: {
+    treeCount?: number;
+    nodeCount?: number;
+    edgeCount?: number;
+    branchPointCount?: number;
+    segmentCount?: number;
+  } = {};
```
MichaelBuessemeyer, line 69: Maybe make this an own (local) type, to avoid repetition in lines 98-104?

MichaelBuessemeyer (commented):
The frontend code looks very good :+1:. I came up with one suggestion to make the code DRYer.
My testing also showed that the feature seems to work seamlessly.
I'd say, the frontend is merge ready :ship: 

@frcroth Could you please approve this PR once everything looks good in the backend part?

frcroth (approved):
LGTM

Issue #3356, opened 2018-10-15
[Refactoring] Disable volume updates for read-only volume/hybrid tracings in central spot


daniel-wer:
Currently the code is littered with checks like this:
```javascript
const allowUpdate = yield* select(state => state.tracing.restrictions.allowUpdate);
if (!allowUpdate) return;
```

Ideally, we would have a central spot where this condition is checked and all volume modifying behavior is enabled/disabled.

PR #7549, opened 2024-01-12
Lock mapping once the volume annotation was changed by the user

### LINKED ISSUE #5431: Lock Mapping for Volume Annotations
## Detailed Description
When a mapping is available for a segmentation layer, the segment IDs change.
This interacts with volume annotations that use fallback segmentations.

Currently, the mapping can freely be selected at any point in time. When the user brushes, the brushed buckets are copied into the volume annotation as they currently are shown. When the user disables the mapping later, the brushed buckets stay “mapped”. When another mapping is activated, bug #5427 even hides all brushed changes and shows the fallback layer only.

We should consider how we can ensure consistency between the fallback layer data and volume annotation data.
A possibility could be to pin the selected mapping in a volume annotation once you start brushing.
Separating the volume annotation from the selected mapping could remove all so-far brushed changes, just like unlinking the fallback layer does.

## Context
Volume annotations do make sense in the context of mappings. It is currently far too easy to create inconsistent states there, though

MichaelBuessemeyer:
- https://pinmapping.webknossos.xyz

### Steps to test:
- Open a tracing of a dataset with agglomerate mappings like `test-agglomerate-file`
- Select the hdf5 mapping (not the json mapping as json mapping persistence is not currently supported)
- Brush a bit around. The mapping selection so instantly disable itself not allowing to turn off the mapping or switching it.
- Save the tracing and reload the page. The previously used mapping should be pinned, loaded and applied from the beginning and the mapping setting should be disabled.
- Changing the mapping state & selected mapping should not be possible via the UI

### TODOs
- [x] Fix frontend tests
- [x] Write frontend test that expects the pinning of the mapping

### Issues:
- fixes  #5431

------


MichaelBuessemeyer:
@fm3 Here is the PR for the Pin Mapping functionality. The frontend part should work with the ROI dataset and its mappings but still needs some clean up after the backend part is done. Would you please implement the backend part? :pray:

fm3:
@MichaelBuessemeyer I added the backend part, seems to work now :) Please ping me again if you need anything else

MichaelBuessemeyer:
@daniel-wer could you please have a look at the frontend part and @frcroth could you please review the backend part? :pray:

MichaelBuessemeyer:
@fm3 How do you think we should handle `JSON` Mappings? Because the code says that changes to these mappings are not persisted? Should they also be pinned (this is how it is currently implemented in the frontend)

fm3:
Yes, probably. Did you test if that works as expected? I haven’t worked with JSON mapings in a while

webknossos-tracingstore/app/com/scalableminds/webknossos/tracingstore/controllers/VolumeTracingController.scala
```diff
@@ -327,11 +327,13 @@ class VolumeTracingController @Inject()(
           for {
             tracing <- tracingService.find(tracingId)
             tracingMappingName <- tracing.mappingName ?~> "annotation.noMappingSet"
+            _ <- bool2Fox(tracing.mappingIsPinned.getOrElse(false)) ?~> "annotation.mappingIsPinned"
```
frcroth, line 330: Why is the default here false, which leads to a failure, when before it would work? Maybe I don't get this line, but the error "annotation.mappingIsPinned" suggests that a problem would occur be if the mapping were pinned, even though this line requires the mapping to be pinned.

fm3: Excellent catch, I forgot a "not" here. I changed it now and tested that the behavior is as expected.

@MichaelBuessemeyer could you disable the whole proofreading tool in the frontend once the mapping is pinned (unless it is *already* a proofreading annotation i.e. `mappingIsEditable==true`)? Proofreading actions will fail after the first brush stoke as well.

fm3:
General question: should we offer a way to undo this? If the user brushes without knowing that this will disable some stuff, they might want to undo that to restore the prior state. Currently, the mapping would stay pinned. Do you think we should support this? I guess that the restore older version view would do the trick.

MichaelBuessemeyer:
> General question: should we offer a way to undo this? If the user brushes without knowing that this will disable some stuff, they might want to undo that to restore the prior state.

Yeah, I think we should support this. Maybe we should make the whole process of pinning the mapping more transparent? Once the user starts a volume annotation, we could open up a modal to inform the user, that the mapping will be pinned from now on, in case the user accepts the brushing action.

@philippotto and maybe @fm3: On second thought I think I skipped rare cases where the user is able to do an volume annotation without the mapping getting pinned via the copy to next layer functionality: e.g.
1. user annotates something in layer X
2. user enables a mapping
3. user moves to layer X+1
4. user uses the "copy layer" action.

I think there are more possible cases that do a volume annotation but are not caught by the current mechanism implemented in volumetracing_saga.tsx.
Should I consider / try to also catch these rare cases?

frontend/javascripts/oxalis/model/accessors/tool_accessor.ts
```diff
@@ -112,13 +112,48 @@ function _getDisabledInfoWhenVolumeIsDisabled(
...
+    } else if (!isProofReadingToolAllowed) {
+      explanation = getFeatureNotAvailableInPlanMessage(
+        PricingPlanEnum.Power,
+        activeOrganization,
+        activeUser,
+      );
+    } else {
+      explanation = disabledSkeletonExplanation;
+    }
+  }
```
MichaelBuessemeyer, line 141: > @MichaelBuessemeyer could you disable the whole proofreading tool in the frontend once the mapping is pinned (unless it is already a proofreading annotation i.e. mappingIsEditable==true)? Proofreading actions will fail after the first brush stoke as well.


I hope this covers all cases you mentioned @fm3? In short: The code checks whether there is a pinned but not an editable mapping. If that's the case, the proofreading tool is disabled with the message `"Proofreading actions are not allowed with a pinned and modified mapping."`

MichaelBuessemeyer (commented):
The automatic pinning of an active mapping is now also done in case a volume interpolation action, a floodfill action or a quick select action is done.

Moreover, a toast informs the user about the fact that the mapping is now pinned due to the volume annotation action. The same message is also logged to the console, in case the user did not pay enough attention to the toast.

MichaelBuessemeyer:
Besides that, I just had a discussion with @fm3 about where to implement the mapping pinning check. Currently, this is done in the respective sagas by calling the `ensureMaybeActiveMappingIsPinned` function /saga. But @fm3 suggested to maybe do this upon volume bucket changes.

Here are pros & cons about each approach:
Within respective sagas:
Pros:
- The change is more local and at a more expected location

Con:
- Code repetition (WET)
- Writing a new saga / tool / feature that annotates the volume might also need to add the call to `ensureMaybeActiveMappingIsPinned` 

Within bucket update function:
Pros:
- Automatically covers all cases of volume manipulation (DRY)

Cons:
- Unexpected effect (I would expect such a pinning change, when I simply change the data of a volume bucket)


Another proposal that came into my mind while writing:
Maybe put this into the function that up/down samples all annotation to the not annotated magnifications (`applyLabeledVoxelMapToAllMissingResolutions`). This might also catch all volume annotation cases. But on second thought it also seems not to fit as the mapping getting pinned would also be an unexpected side effect in my opinion.

What do you think @daniel-wer?

frontend/javascripts/oxalis/model/accessors/tool_accessor.ts
fm3: Sounds good :) How about something like "Proofreading actions are not supported after modifying the segmentation."? Since it’s the brushing that’s the problem, not the having a mapping.

philippotto:
In my opinion, the logic for ensuring a pinned mapping should live in the sagas, as that gives us the most flexibility (e.g., think about a potential flow where the user is asked to confirm the pinning). Also, it's the natural place for such side-effects (in contrast to the actual bucket mutation).

> Con:
> 
> * Code repetition (WET)
> * Writing a new saga / tool / feature that annotates the volume might also need to add the call to `ensureMaybeActiveMappingIsPinned`

If we would be able to get the approach DRY, both cons would disappear. We have a similar problem for the write-permission check (see https://github.com/scalableminds/webknossos/issues/3356). Ideally, we would extend/unify something like `createVolumeLayer` (which is already mostly used for mutation¹) and then return (within that function) a `{ allowed: false, reason: "not pinned" }` object or something like this if pinning is necessary. Then, all callers would be forced to deal with that return type. the same interface could be used if the annotation is read-only.

However, this refactoring does not need to block this PR imo.

¹) I think the one notable exception is the floodfill which could be changed

daniel-wer:
I agree that in theory, the bucket mutation would be _the_ spot to check for this, but that code is not intended to have these kinds of side effects (mapping pinning and user prompt). As Philipp states, we should be able to refactor the bucket mutation code to avoid that one could forget to pin the mapping/check for write permissions before mutating buckets :+1:

daniel-wer:
> Moreover, a toast informs the user about the fact that the mapping is now pinned due to the volume annotation action. The same message is also logged to the console, in case the user did not pay enough attention to the toast.

Would it be complicated to let users confirm the mapping pinning, instead of only informing them? I understood @philippotto that he preferred that solution and at least I do, too. Since we don't offer any method to revert the mapping pinning, asking for confirmation seems more user-friendly.

daniel-wer (changes requested):
Nice, I'm happy that we're finally tackling this and avoid unexpected behavior :tada: 

Happy path testing went well, but I encountered one serious and two slightly obscure issues, and have one clarifying question:

- [x] When activating an HDF5 mapping, then disabling it, and then brushing, the non-activated mapping is pinned. The pinning needs to check that the "active" mapping is actually enabled (or pin that no mapping is activated, see the last point in this list).
- [x] When activating a JSON mapping, and then brushing, it is pinned regardless (although the PR description states the pinning is only supported for HDF5 mappings). When reloading the annotation the mapping is disabled, but pinned, and cannot be activated or changed - which is unexpected. I'd avoid pinning JSON mappings in the first place if it's not properly supported and leads to weird behavior.
- [ ] When activating a JSON mapping and copying the sharing link, then switching to an agglomerate view mapping and brushing (thereby pinning the mapping) and then opening the previously copied sharing link, the mapping ID switch is in the same disabled but pinned state as described in the previous point. This might be a bug already existing for editable mappings, but maybe you could have a look or create a separate issue for that.
- [x] Thinking about this issue, shouldn't changing the volume annotation when no mapping is activated also pin that state and forbid future mapping activations? Imo, the reasoning is the same for why to do it when a mapping is activated :thinking: If I brush when no mapping is activated and then activate one, the modified buckets show the unmapped fallback layer instead of the mapped one. /cc @philippotto @fm3

fm3:
>  shouldn't changing the volume annotation when no mapping is activated also pin that state and forbid future mapping activations

Yes, I think you’re right! We didn’t think about this case before 🤔 If my understanding is correct, this should also be expressable with the available update actions

MichaelBuessemeyer:
> Thinking about this issue, shouldn't changing the volume annotation when no mapping is activated also pin that state and forbid future mapping activations? Imo, the reasoning is the same for why to do it when a mapping is activated 🤔 If I brush when no mapping is activated and then activate one, the modified buckets show the unmapped fallback layer instead of the mapped one. /cc @philippotto @fm3

For documentation purposes: Here is a part of discussion regarding this: https://scm.slack.com/archives/C5AKLAV0B/p1707745523065049

MichaelBuessemeyer (commented):
Thank you for your valuable feedback @daniel-wer :pray:.

I think I addressed everything you mentioned.

Could you please do another review of these changes?

Please note the following test cases that are now additionally required:
- Create a new annotation. Brush around. This should instantly pin the "no mapping is active" state and should no longer be changeable. In that case the merger mode should also be disabled.
- Create a new annotation. Activate some mapping. Deactivate it & brush a bit. Reloading should result in an annotation with the "no mapping is active" state and that should be pinned (not changeable)
- The bugs you described in your review should be fixed now.

frontend/javascripts/messages.tsx
```diff
@@ -254,6 +254,10 @@ instead. Only enable this option if you understand its effect. All layers will n
     'All trees are currently hidden. You can disable this by toggling the "Skeleton" layer in the layer settings in the left sidebar.',
   "tracing.invalid_json_url_hash":
     "Cannot parse JSON URL hash. More information was printed to the browser's console.",
+  "tracing.pin_mapping_info":
+    "The active volume annotation layer has an active mapping. Once you start mutating the layer, the mapping will be permanently enabled. This can only be undone by restoring an older version of this annotation. Are you sure you want to continue?",
```
daniel-wer, line 258:
```suggestion
    "The active volume annotation layer has an active mapping. By mutating the layer, the mapping will be permanently pinned and can no longer be changed or disabled. This can only be undone by restoring an older version of this annotation. Are you sure you want to continue?",
```

Maybe it's easier to understand like this?

frontend/javascripts/messages.tsx
```diff
@@ -254,6 +254,10 @@ instead. Only enable this option if you understand its effect. All layers will n
...
+  "tracing.pin_mapping_confirmed": (mappingName: string) =>
+    `The mapping ${mappingName} is from now on permanently enabled for this annotation.`,
```
daniel-wer, line 260:
```suggestion
    `The mapping ${mappingName} is now permanently pinned for this annotation and can no longer be changed or disabled.`,
```

frontend/javascripts/oxalis/model/accessors/tool_accessor.ts
```diff
@@ -138,6 +148,9 @@ function _getDisabledInfoForProofreadTool(
     } else {
       explanation = disabledSkeletonExplanation;
     }
+  } else {
+    explanation =
+      "A mapping that does not support proofreading actions is pinned to this annotation.";
```
daniel-wer, line 153:
```suggestion
      "A mapping that does not support proofreading actions is pinned to this annotation. Most likely, the annotation layer was modified before (e.g. by brushing).";
```

Just a suggestion to make this more clear to users.

frontend/javascripts/oxalis/model/reducers/volumetracing_reducer.ts
```diff
@@ -391,18 +392,27 @@ function VolumeTracingReducer(
...
 
       const { mappingName, mappingType } = action;
       return setMappingNameReducer(state, volumeTracing, mappingName, mappingType);
     }
 
     case "SET_MAPPING_IS_EDITABLE": {
       // Editable mappings cannot be disabled or switched for now
-      if (volumeTracing.mappingIsEditable) return state;
+      // An editable mapping is always pinned.
+      if (volumeTracing.mappingIsEditable && volumeTracing.mappingIsPinned) return state;
```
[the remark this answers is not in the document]
daniel-wer: > Just as a sidenote: Yes, but this goes also for mappings that are not pinned and not editable.

Yes, that's exactly the one case we want to allow.

> If our code somehow falsely triggers the SET_MAPPING_IS_EDITABLE action. Your suggestion now only covers the case where our code accidentally calls the action for a pinned mapping that should not be editable.

Of the four possible cases, the condition now does nothing for three cases, and only changes both booleans to true if they were both false before. As far as I understand, this is exactly the case where we want to allow to set a mapping to be editable.

frontend/javascripts/oxalis/model/sagas/saga_helpers.ts
```diff
@@ -39,4 +46,70 @@ export function* takeEveryUnlessBusy(
...
+          resolve({ isMappingPinnedIfNeeded: true, reason: "User confirmed." });
+        } else {
+          // Having an active mapping without a name should be impossible. Therefore, no further error handling is done.
+          reject({ isMappingPinnedIfNeeded: false, reason: "No mapping name." });
+        }
+      };
+      Modal.confirm({
+        title: "Should the active Mapping be pinned?",
+        content: messages["tracing.pin_mapping_info"],
+        okText: "Enable Mapping Permanently",
```
daniel-wer, line 78:
```suggestion
        okText: "Pin Mapping Permanently",
```

frontend/javascripts/oxalis/model/sagas/saga_helpers.ts
```diff
@@ -39,4 +46,70 @@ export function* takeEveryUnlessBusy(
...
+        cancelText: "Abort Annotation Action",
+        width: 600,
+        onOk: pinMapping,
+        onCancel: async () => {
+          reject({ isMappingPinnedIfNeeded: false, reason: "User aborted." });
+        },
+      });
+    } else {
+      resolve({ isMappingPinnedIfNeeded: true, reason: "No active mapping." });
```
daniel-wer, line 87: If I'm reading the code correctly, the reason here is rather that the mapping is already pinned, isn't it?

```suggestion
      resolve({ isMappingPinnedIfNeeded: true, reason: "Mapping is already pinned." });
```

However, this is already checked in `ensureMaybeActiveMappingIsPinned` so maybe the whole condition can be removed?

frontend/javascripts/oxalis/model/sagas/saga_helpers.ts
```diff
@@ -39,4 +46,70 @@ export function* takeEveryUnlessBusy(
...
+  }
+  const activeMappingByLayer = yield* select(
+    (state) => state.temporaryConfiguration.activeMappingByLayer,
+  );
+  const isSomeMappingActive =
+    volumeTracing.tracingId in activeMappingByLayer &&
+    activeMappingByLayer[volumeTracing.tracingId].mappingStatus === MappingStatusEnum.ENABLED;
+  const isHDF5Mapping =
+    volumeTracing.tracingId in activeMappingByLayer &&
+    activeMappingByLayer[volumeTracing.tracingId]?.mappingType === "HDF5";
```
daniel-wer, line 106:
```suggestion
    activeMappingByLayer[volumeTracing.tracingId].mappingType === "HDF5";
```

since you check for that before or does typescript complain (I would think it doesn't, because it also works in line 103).

frontend/javascripts/oxalis/model/sagas/volumetracing_saga.tsx
```diff
@@ -703,9 +710,11 @@ export function* diffVolumeTracing(
       prevVolumeTracing.mappingName !== volumeTracing.mappingName ||
       prevVolumeTracing.mappingIsPinned !== volumeTracing.mappingIsPinned
     ) {
+      // Once the first volume action is performed on a volume layer, the mapping state is pinned.
+      // In case no mapping is active, this is denoted by setting the mapping name to null.
       const action = updateMappingName(
-        volumeTracing.mappingName,
-        volumeTracing.mappingIsEditable,
+        volumeTracing.mappingName || null,
+        volumeTracing.mappingIsEditable || null,
```
daniel-wer, line 717: Are the `|| null` needed, because the backend expects `null` instead of `undefined` in that case?

frontend/javascripts/oxalis/view/left-border-tabs/mapping_settings_view.tsx
```diff
@@ -175,27 +176,36 @@ class MappingSettingsView extends React.Component<Props, State> {
       this.props.mapping &&
       this.props.hideUnmappedIds != null;
     const isDisabled = this.props.isEditableMappingActive || this.props.isMappingPinned;
+    const disabledMessage = this.props.isEditableMappingActive
+      ? "Editable mappings from proofreading actions cannot be disabled"
```
daniel-wer, line 180:
```suggestion
      ? "The mapping has been edited through proofreading actions and can no longer be disabled or changed."
```

frontend/javascripts/oxalis/view/left-border-tabs/mapping_settings_view.tsx
```diff
@@ -175,27 +176,36 @@ class MappingSettingsView extends React.Component<Props, State> {
...
+      : this.props.mapping
+      ? "This mapping has been pinned to this annotation and thus cannot be disabled."
```
daniel-wer, line 182:
```suggestion
      ? "The mapping has been pinned, because the segmentation was modified while it was active. It can no longer be disabled or changed."
```

frontend/javascripts/oxalis/view/left-border-tabs/mapping_settings_view.tsx
```diff
@@ -175,27 +176,36 @@ class MappingSettingsView extends React.Component<Props, State> {
...
+      : "This annotation was started with no mapping enabled. To ensure a consistent state, mappings cannot be enabled.";
```
daniel-wer, line 183:
```suggestion
      : "The segmentation was modified while no mapping was active. To ensure a consistent state, mappings can no longer be enabled.";
```

daniel-wer (commented):
Nice work! I mostly made some phrasing suggestions to keep the explanations more consistent and make it easier for users to understand what's happening. Use your own judgement on whether you want to apply those :)

I didn't retest yet, but will do so and drop my notes here afterwards.

daniel-wer (commented):
Notes from testing:

- [ ] When canceling a brush action in the confirmation modal, the saga crashes (same for other tools, although sometimes only an error on the console and a toast is shown)
image
- [ ] This was not resolved for me:  When activating a JSON mapping and copying the sharing link, then switching to an agglomerate view mapping and brushing (thereby pinning the mapping) and then opening the previously copied sharing link, the mapping ID switch is in the same disabled but pinned state as described in the previous point. This might be a bug already existing for editable mappings, but maybe you could have a look or create a separate issue for that.
image

frontend/javascripts/test/sagas/volumetracing/volumetracing_saga.spec.ts
```diff
@@ -394,3 +420,72 @@ test("VolumeTracingSaga should ignore brush action when busy (saga test)", (t) =
...
+  const jsonDummyMapping = { ...dummyActiveMapping, mappingType: "JSON" };
+  const saga = ensureMaybeActiveMappingIsPinned(volumeTracing);
+  saga.next();
+  expectValueDeepEqual(
+    t,
+    saga.next({ [volumeTracing.tracingId]: jsonDummyMapping }),
+    put(VolumeTracingActions.setMappingIsPinnedAction()),
+  );
+  t.true(saga.next().done);
+});
```
MichaelBuessemeyer, line 491: A still open question to me is how the pinning should behave in case the user has an active JSON mapping. As JSON mappings cannot be manipulated (changes to it are not persisted), should the state that it is active be pinned, or should "not active mapping" be pinned? For consistency reasons, the state of whether a mapping and which mapping is pinned should be determined by the first volume annotation action.

Here are the conversational bits of this pr discussing this:
> @fm3 How do you think we should handle JSON Mappings? Because the code says that changes to these mappings are not persisted? Should they also be pinned (this is how it is currently implemented in the frontend)

> Yes, probably. Did you test if that works as expected? I haven’t worked with JSON mapings in a while

> When activating a JSON mapping, and then brushing, it is pinned regardless (although the PR description states the pinning is only supported for HDF5 mappings). When reloading the annotation the mapping is disabled, but pinned, and cannot be activated or changed - which is unexpected. I'd avoid pinning JSON mappings in the first place if it's not properly supported and leads to weird behavior.

My thoughts on this would be to ensure that a mapping state is pinned, that the state of "no mapping is active" should be pinned to the annotation once the user annotates in some way while a JSON Mapping is active.

=> One open TODO for this (in my opinion) would be, to also disable / deactivate the JSON mapping once any volume annotation is done. Because if it stays active and the user annotates a little and then reloads, the user seems different results as the JSON mapping was previously still active but it cannot be enabled anymore after the reload.

@daniel-wer @philippotto What are your opinions on this?

frontend/javascripts/oxalis/model/sagas/saga_helpers.ts
```diff
@@ -39,4 +46,74 @@ export function* takeEveryUnlessBusy(
...
+    activeMappingByLayer[volumeTracing.tracingId].mappingStatus === MappingStatusEnum.ENABLED;
+  const isHDF5Mapping =
+    volumeTracing.tracingId in activeMappingByLayer &&
+    activeMappingByLayer[volumeTracing.tracingId].mappingType === "HDF5";
+  if (isSomeMappingActive && isHDF5Mapping) {
+    try {
+      return yield* call(askUserForPinningActiveMapping, volumeTracing, activeMappingByLayer);
+    } catch (error: any) {
+      return error as EnsureMappingIsPinnedReturnType;
+    }
```
MichaelBuessemeyer, line 110: > When canceling a brush action in the confirmation modal, the saga crashes (same for other tools, although sometimes only an error on the console and a toast is shown)

Thanks for finding this. I simply forgot to catch the rejected promise properly here. It should be fixed now :)

frontend/javascripts/messages.tsx
daniel-wer: I was also thinking about that which is why I added the "and can no longer be changed or disabled" explaining bit. To me "permanently enabled" sounds somewhat ambiguous and doesn't make clear that the mapping can no longer be changed.
So there's only one possible solution, I asked our future AI overlords which came up with these:

```
"By mutating the layer, the mapping will be permanently fixed in place (and can no longer be changed or disabled)."
"By mutating the layer, the mapping will be permanently locked (and can no longer be changed or disabled)."

These alternatives may help ensure that the meaning is clear to a broader audience.
```

which I both like. @philippotto What do you think?

fm3: I think »locked« is good. We could even consider using that word in code as well instead of pinned? Then we don’t have this disparity between the UI and the code. That would be an undertaking, though 😅

frontend/javascripts/test/sagas/volumetracing/volumetracing_saga.spec.ts
daniel-wer: I see two main options to make the user experience consistent:

1) Pin the JSON mapping the same way that agglomerate mappings are pinned. Reactivate the mapping after a reload, make it unchangeable, etc.

2) Forbid volume annotation actions while a JSON mapping is active. (Disabling the mapping once any volume annotation is done is potentially confusing imo)

Since JSON mappings are a very niche use case, I'd go with the option that's the easiest to implement.

frontend/javascripts/messages.tsx
philippotto: I'm also in favor of locked :+1: 

> We could even consider using that word in code as well instead of pinned? Then we don’t have this disparity between the UI and the code. That would be an undertaking, though 😅

I think, in the front-end it would be a simple search&replace. So, it's your call if it's worth the effort in the back-end :)

fm3: Yeah, it’s not that much in the backend either. If we want to do it, we should do it in this PR, though, because changing the json update actions in the fossildb later would not be fun.

frontend/javascripts/test/sagas/volumetracing/volumetracing_saga.spec.ts
philippotto: Just as a side note: there are also dynamic mappings (e.g., by merger mode). I'm not sure whether these are interpreted as JSON mappings in some cases. Locking these forever would be unusual in the context of the merger mode, I think.
I'd probably prefer (2).

frontend/javascripts/messages.tsx
MichaelBuessemeyer: Ok done, I hope I renamed every occurrence of "pinned" / "pin".

frontend/javascripts/test/sagas/volumetracing/volumetracing_saga.spec.ts
MichaelBuessemeyer: :+1: I implemented (2) and it was pretty easy to integrate this.

.gitignore
```diff
@@ -76,3 +76,4 @@ coverage-ts
 .bsp
 temp-webknossos-schema**
 conf/application.conf-e
+.env
```
MichaelBuessemeyer, line 79: I want to avoid pushing my .env file :see_no_evil:

frontend/javascripts/messages.tsx
```diff
@@ -254,10 +254,10 @@ instead. Only enable this option if you understand its effect. All layers will n
     'All trees are currently hidden. You can disable this by toggling the "Skeleton" layer in the layer settings in the left sidebar.',
   "tracing.invalid_json_url_hash":
     "Cannot parse JSON URL hash. More information was printed to the browser's console.",
-  "tracing.pin_mapping_info":
-    "The active volume annotation layer has an active mapping. By mutating the layer, the mapping will be permanently enabled. This can only be undone by restoring an older version of this annotation. Are you sure you want to continue?",
-  "tracing.pin_mapping_confirmed": (mappingName: string) =>
-    `The mapping ${mappingName} is now permanently enabled for this annotation and can no longer be changed or disabled.`,
+  "tracing.locked_mapping_info":
+    "The active volume annotation layer has an active mapping. By mutating the layer, the mapping will be permanently locked (and can no longer be changed or disabled). This can only be undone by restoring an older version of this annotation. Are you sure you want to continue?",
```
daniel-wer, line 258: I wanted the parentheses to indicate that this part would be optional. I think it's more clear to include it, but would remove the parentheses, then :)

```suggestion
    "The active volume annotation layer has an active mapping. By mutating the layer, the mapping will be permanently locked and can no longer be changed or disabled. This can only be undone by restoring an older version of this annotation. Are you sure you want to continue?",
```

frontend/javascripts/oxalis/model/actions/volumetracing_actions.ts
```diff
@@ -360,7 +360,7 @@ export const setMappingIsEditableAction = () =>
     type: "SET_MAPPING_IS_EDITABLE",
   } as const);
 
-export const setMappingIsPinnedAction = () =>
+export const setMappingIsLockedAction = () =>
   ({
     type: "SET_MAPPING_IS_PINNED",
```
daniel-wer, line 365: I'd rename this to `LOCKED` too

frontend/javascripts/oxalis/model/sagas/quick_select_saga.ts
```diff
@@ -32,13 +32,13 @@ export default function* listenToQuickSelect(): Saga<void> {
     "COMPUTE_QUICK_SELECT_FOR_RECT",
     function* guard(action: ComputeQuickSelectForRectAction) {
       try {
-        // As changes to the volume layer will be applied, the potentially existing mapping should be pinned to ensure a consistent state.
         const volumeTracing: VolumeTracing | null | undefined = yield* select(
           getActiveSegmentationTracing,
         );
         if (volumeTracing) {
+          // As changes to the volume layer will be applied, the potentially existing mapping should be locked to ensure a consistent state.
           const { isMappingPinnedIfNeeded } = yield* call(
```
daniel-wer, line 40: It looks like the search and replace was not exhaustive, could you check again for remaining occurrences of "pinned"? You could also use the "Files Changed" tab here on github.

daniel-wer (commented):
Thank you for incorporating the feedback! Code mostly LGTM, but I still encountered some issues during testing:

- [x] When aborting a flood fill in the lock confirmation modal, the flood fill is performed anyways (and the mapping is not locked).
- [x] When locking a mapping by brushing with an activated mapping and then reloading the page, the mapping no longer seems to be locked. Is this a regression due to the renaming?
- [x] Might be because of the previous point, but this is still not resolved for me: "When activating a JSON mapping and copying the sharing link, then switching to an agglomerate view mapping and brushing (thereby pinning the mapping) and then opening the previously copied sharing link, the JSON mapping gets activated."

MichaelBuessemeyer:
Hi @daniel-wer,

thank you for your feedback and thorough testing. I hope I found every occurrence of "pin / pinned" and renamed it accordingly.

> When locking a mapping by brushing with an activated mapping and then reloading the page, the mapping no longer seems to be locked. Is this a regression due to the renaming?

Did you maybe not wait long enough to reach a save state / for all update actions being sent to the server to persist in the locked state? For me, the locked state is working after clicking the save annotation button, waiting, and then reloading.

=> Maybe I should enforce that the annotation reaches a saved state after the mapping is locked? 

> Might be because of the previous point, but this is still not resolved for me: "When activating a JSON mapping and copying the sharing link, then switching to an agglomerate view mapping and brushing (thereby pinning the mapping) and then opening the previously copied sharing link, the JSON mapping gets activated."

Same here as above. Or maybe the dev deployment was simply not in a good state due to the renaming :thinking:?

What do you think about this suggestion?
=> Maybe I should enforce that the annotation reaches a saved state after the mapping is locked?

daniel-wer (approved):
Thanks for pushing this through, works beautifully :tada: 

I'm pretty sure I made sure the annotation was saved when previously testing and even removed/reinstalled the dev deployment, but who knows what happened there. Now it worked like a charm.

I don't think it's needed to ensure that a saved state is reached after a mapping was locked. The mapping lock update action will be the first in the save queue and if that one doesn't reach the server, the bucket modification actions following afterwards won't reach the server either, so the state will remain consistent if users decide not to save.

PR #7550, opened 2024-01-15
Make isValidNewName route always return 200

### LINKED ISSUE #7399: dataset isValidNewName route should always return 200, with true/false body
provide a version of `isValidNewName` that returns 200 even for invalid names (and has a response body saying if the name was valid)

the current version is more of an `assertValidNewName`.

Note that since the libs use this route, the change should introduce a new api version (adapt webknossos.versioned.routes to support the old variant in api v5)

frcroth:
- https://___.webknossos.xyz

### Steps to test:
- click on "Add Dataset" in dashboard
- type a name into the "Dataset Name" field that already exists for another ds -> red label should appear
- change name to something unique -> error should disappear
- no error toast should appear during the above

### TODOs:
- [x] Since it is not user facing, no Changelog, only migration?

### Issues:
- fixes #7399

------


fm3 (commented):
Backend LGTM :) @philippotto could you have a look at the frontend (and fix it if needed)? My guess is that showErrorToast: false is no longer needed, since in normal operation there should no longer be an error 400 here.

philippotto (approved):
sure! I pushed a small commit. testing went well :rocket:

fm3 (approved):

PR #7553, opened 2024-01-15
Implement index_location setting for Zarr3 Datasets

### LINKED ISSUE #7436: Implement zarr3 sharding index_location
compare https://github.com/zarr-developers/zarr-specs/pull/280

frcroth:
- https://___.webknossos.xyz

### Steps to test:
- Test datasets with permutations of the configuration options
	- With crc32c codec / without
	- index_location start / end

### TODOs:
- [ ] ...

### Issues:
- fixes #7436 

------


webknossos-datastore/app/com/scalableminds/webknossos/datastore/datareaders/zarr3/Zarr3Array.scala
```diff
@@ -122,8 +119,22 @@ class Zarr3Array(vaultPath: VaultPath,
       parsed = parseShardIndex(shardIndexRaw)
     } yield parsed
 
+  private lazy val checkSumLength =
```
fm3, line 122: rename to make clear this is about sharding

fm3 (approved):

webknossos-datastore/app/com/scalableminds/webknossos/datastore/datareaders/zarr3/Zarr3Array.scala
```diff
@@ -122,8 +119,22 @@ class Zarr3Array(vaultPath: VaultPath,
       parsed = parseShardIndex(shardIndexRaw)
     } yield parsed
 
+  private lazy val shardIndexChecksumLength =
```
normanrz, line 122: I guess this works for most datasets, but not for all. Although, it doesn't make a lot of sense, it is spec-compliant to have multiple crc32c checksums. In the future, there might be other codecs that change the size of the shard index.

A more correct way would be to compute the size of the index based on the entire index_codecs pipeline. zarrita uses for that the [`compute_encoded_size` methods](https://github.com/scalableminds/zarrita/blob/async/zarrita/codecs.py#L545).

normanrz: At least, there should be some assertions that only supported index configurations are usable.

webknossos-datastore/app/com/scalableminds/webknossos/datastore/datareaders/zarr3/Codecs.scala
```diff
@@ -294,8 +287,16 @@ final case class ShardingCodecConfiguration(chunk_shape: Array[Int],
...
                                             index_location: IndexLocationSetting.IndexLocationSetting =
                                               IndexLocationSetting.end)
-    extends CodecConfiguration {
+    extends CodecConfiguration with BoxImplicits {
   override def name: String = ShardingCodecConfiguration.name
+// index_codecs.size <= 2 && index_codecs.count(_.name == "bytes") == 1 && index_codecs.count(_.name == "crc32c") <= 1
+  def isSupported: Box[Unit] =
+  for {
+    _ <- bool2Box(index_codecs.size <= 2) ?~! "Maximum of 2 index codecs supported"
+    _ <- bool2Box(index_codecs.count(_.name == "bytes") == 1) ?~! "Exactly one bytes codec supported"
```
fm3, line 296: can we get something like "Only exactly one bytes codec is supported, got 3"? (for all of these?)

fm3 (approved):

fm3:
Approving this now based on normans previous feedback. @normanrz if you have other wishes, let’s add them as a follow-up issue :) that way this PR isn’t blocke

webknossos-datastore/app/com/scalableminds/webknossos/datastore/datareaders/zarr3/Zarr3Array.scala
frcroth: #7574

PR #7561, opened 2024-01-17
Use orga and project owner emails to notify

### LINKED ISSUE #7455: Send project overtime time limit emails to project owner
## Detailed Description
Each project can define a time limit for working on an annotation to keep some sort of checks and balances. In case of breaching the time limit, it seems that email notifications are only send to the annotator but not the project owner.
On other words, if you are a slow annotator you might be informed by the system but any project manager will be none the wiser.

Informing the project owner about these irregularities makes sense to me.

See https://github.com/scalableminds/webknossos/blob/c6e16633e13e9d6536be11b17bc8b9ec2701561b/app/models/user/time/TimeSpanService.scala#L171

## Context

### LINKED ISSUE #7486: if newUserMailingList and overtimeMailingList are empty, use orga owner email
Also, the newUserMailingList should also be notified if a signup is done through an invite.

Related: notify project owner on overtime, that would be #7455

frcroth:
- https://___.webknossos.xyz

### Steps to test:
- Set logToStdOut in application.conf to true
- create an invite and accept it. -> The org owner will get an email (even though none is configured)
- Not sure how to test the overtime stuff

### TODOs:
- [ ] ...

### Issues:
- fixes #7486
- fixes #7455 

------


app/mail/DefaultMails.scala
```diff
@@ -16,29 +16,30 @@ class DefaultMails @Inject()(conf: WkConf) {
...
   private val newOrganizationMailingList = conf.WebKnossos.newOrganizationMailingList
 
-  def registerAdminNotifyerMail(name: String,
+  def registerAdminNotifierMail(name: String,
                                 email: String,
                                 brainDBResult: Option[String],
                                 organization: Organization,
-                                autoActivate: Boolean): Mail =
+                                autoActivate: Boolean,
+                                receiver: String): Mail =
```
fm3, line 24:
```suggestion
                                recipient: String): Mail =
```

app/models/organization/OrganizationService.scala
```diff
@@ -144,4 +144,23 @@ class OrganizationService @Inject()(organizationDAO: OrganizationDAO,
...
+  private def fallbackOnOwnerEmail(possiblyEmptyEmail: String, organization: Organization)(
+      implicit ctx: DBAccessContext): Fox[String] =
+    if (possiblyEmptyEmail.nonEmpty) {
+      Fox.successful(possiblyEmptyEmail)
+    } else {
+      for {
+        owner <- userDAO.findOwnerByOrg(organization._id)
+        ownerMultiUser <- multiUserDAO.findOne(owner._multiUser)
+        ownerEmail = ownerMultiUser.email
+      } yield ownerEmail
```
fm3, line 157: could this use userService.emailFor? (Or does that cause circular imports or something?)

fm3 (commented):
Mostly LGTM, just added some comments about cosmetics

app/mail/DefaultMails.scala
frcroth: Removed other occurences as well

fm3 (approved):

PR #7562, opened 2024-01-17
Upgrade to Play 3

### LINKED ISSUE #7492: Upgrade to Play 3 and Pekko
Play 3 moved from akka to pekko due to license changes.

Play 2.9 will be maintained in parallel to 3, but we may want to do the same switch. We use akka for a couple things ourselves (IntervalScheduler, cleanup job scheduling, mailer actor, …), my assumption would be that it would be fairly simple to migarte those to pekko.

play-silhouette 10 is compatible with play 3 (while we currently use the play-2.9-specific silhouette 9)

frcroth:
- https://___.webknossos.xyz

### Steps to test:
Anything specific that should be tested? Going through some common features (e.g. loading datasets remote/local...) works.

### TODOS:
- [x] Investigate test failure
- [x] Migration guide

### Issues:
- fixes #7492 

------


fm3 (commented):
Sweet! I’ll do some testing on some more things that come to mind
 - [x] scheduled clean up jobs
 - [ ] requesting ad-hoc meshes

Could you add a migration guide entry noting the changed config keys in the `.conf` files?

fm3 (approved):
:tada:

PR #7563, opened 2024-01-17
Fix listing other users’ annotations

fm3:
- https://___.webknossos.xyz

### Steps to test:
- abc

### TODOs:
- [ ] ...

### Issues:
- fixes #

------

### DIFF
--- a/app/models/annotation/Annotation.scala
+++ b/app/models/annotation/Annotation.scala
@@ -248,14 +248,17 @@ class AnnotationDAO @Inject()(sqlClient: SqlClient, annotationLayerDAO: Annotati
     accessQueryFromAccessQWithPrefix(listAccessQ, q"")(ctx)
 
   override protected def readAccessQ(requestingUserId: ObjectId): SqlToken =
+    readAccessQWithPrefix(requestingUserId, q"")
+
+  protected def readAccessQWithPrefix(requestingUserId: ObjectId, prefix: SqlToken): SqlToken =
     q"""(
-              visibility = ${AnnotationVisibility.Public}
-           or (visibility = ${AnnotationVisibility.Internal}
-               and (select _organization from webknossos.teams where webknossos.teams._id = _team)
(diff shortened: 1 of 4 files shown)


frontend/javascripts/admin/admin_rest_api.ts
```diff
@@ -553,15 +553,6 @@ export function deletePrivateLink(linkId: string): Promise<{
 }
 
 // ### Annotations
-export function getAnnotationInfos(
```
hotzenklotz, line 556: Am I assuming correctly, that this route does not exist anymore?

app/models/annotation/Annotation.scala
```diff
@@ -304,11 +307,14 @@ class AnnotationDAO @Inject()(sqlClient: SqlClient, annotationLayerDAO: Annotati
   def findAllListableExplorationals(
       isFinished: Option[Boolean],
       forUser: Option[ObjectId],
+      // In dashboard, list only own + explicitly shared annotations. When listing those of another user, list all of their annotations the viewer is allowed to see
+      isForOwnDashboard: Boolean,
```
frcroth, line 311:
```suggestion
      // In dashboard, list only own + explicitly shared annotations. When listing those of another user, list all of their annotations the viewer is allowed to see
      isForOwnDashboard: Boolean,
```

I find the naming here to be too closely linked with the intent and not with what actually happens. Also surprising behavior when the function is called findAllListable and the parameter gives no hint on that the result may change. Maybe more like filterOwnedAndExplicitlyShared

hotzenklotz (approved):
Tested and works.

frcroth (commented):
Had this review pending, forgot to submit

app/models/annotation/Annotation.scala
fm3: I agree, wrote https://github.com/scalableminds/webknossos/issues/7565 to track that

Issue #6613, opened 2022-11-07
Allow to rename datasets


normanrz:
## Detailed Description
Datasets should be renamable. For that, we should decouple the URL from the dataset name in Notion-style (i.e. `<name>-<id>`, but only the `id` part is actually used). I would suggest to add a new dataset name field that is separate from the name that is used for locating the dataset on-disk (maybe rename that to "path"?). Not sure, whether we should reuse the `displayName` field. We should check the current occurrences and impose the same restrictions as on dataset names. For backwards compatibility, we resolve the name in the URL on a best-effort basis.

fm3:
unassigning myself for the moment, since sprint priorities have changed

fm3:
Note: In this context we may also create a way of deleting datasets, without breaking their annotations (i.e. still allowing to list + download them)

fm3:
I don’t think we can easily reuse the displayName, because several existing values contain spaces and parentheses etc. Unless we decide that we can just drop/convert these.

If we want to keep that option, we will then need

id
nameOnDisk (or path? directory? note that it’s not the full path)
name (for use in uris, strict restrictions, but no longer unique)
displayName (no restrictions, free text field, not unique. as before)

fm3:
Another thought: since we still need the nameOnDisk anyway, could we not use that for most APIs as well? e.g. in datastore and worker jobs.

We could just make the display name (or maybe another name field) a more prominent feature :thinking: 
And switch to the id-based URI format, with fallback to the old format.

I get the feeling I don’t yet know well enough what the desired outcome is here. Depending on that we may have to either change almost all APIs or only a few.

normanrz:
> since we still need the nameOnDisk anyway, could we not use that for most APIs as well? e.g. in datastore and worker jobs.

I don't think that is a good idea. Mid-term I'd like to make datasets virtual, ie. they don't need a folder on disk. For that, we need id-based URIs. Doesn't seem wise to refactor to nameOnDisk when we want to change that in a few months to ids.

normanrz:
Keeping both displayName and name going forward seems unnecessary

fm3:
Thoughts
 - dataset id also needs to be serialized to NMLs
 - heuristic to select dataset from ambiguous names: take the oldest (hypothesis: old NMLs/URIs were generated _before_ the renaming was ok, so probably old dataset)

MichaelBuessemeyer:
Notes from my talk with @normanrz:

[...]

PR #7573, opened 2024-01-22
Delete Projects,Teams,TaskTypes with name suffix to allow reusing names

### LINKED ISSUE #5388: Allow re-using names of deleted entities
Currently, the names of deleted tasktypes, teams, and datasets (but not projects) stay reserved. We should discuss which of those should be allowed to be reused.

Also, if this constraint is released, the uniqueness of non-deleted ones should be checked in SQL, not just in scala. This could be done with trigger functions.

fm3:
Allow re-using names of projects, teams, tasktypes after deleting them.

- https://reuseteamnames.webknossos.xyz

### Steps to test:
- Create some projects, teams, tasktypes, delete them, create new with the same name

### TODOs:
- [x] discuss if we need a special char as delimiter

### Issues:
- fixes #5388 

------


frcroth (approved):
Followup issue for deleting datasets?

fm3:
> Followup issue for deleting datasets?

Part of #6613 I’d say

For datasets this is harder since they are referenced by name in tracing proto objects in FossilDB

fm3:
@frcroth I decided to add an SQL migration to rename all previously deleted projects,teams and tasktypes to free up their names too. Could you have another look please?

frcroth (approved):

Issue #7541, opened 2024-01-11
Expand Skeletons with SAM


philippotto:
## Detailed Description
Given a skeleton, SAM could be used to annotate volume data for all nodes and the edges between the nodes. It remains to clarify how SAM should be invoked:

- simply invoke SAM with a point for each node
- should a bbox be passed to SAM? if so, how large?
- should the user label the sections where the nodes are and then the feature is basically interpolating along the edges (similar to #7540)?

The first option is certainly the simplest. Let's try this first.

## Context

* [ ] Specific to long-running jobs (set `jobsEnabled=true` in `application.conf`)
* [ ] Specific to webknossos.org (set `isDemoInstance=true` in `application.conf`)

MichaelBuessemeyer:
We decided against this feature as SAM doesn'T handle point input so well and the feature not offering enough value

PR #7575, opened 2024-01-22
First version of SAM skeleton prediction

MichaelBuessemeyer:
- https://___.webknossos.xyz

### Steps to test:
- Create an annotation and trace some part of some cell with a skeleton. A node or multiple nodes should be present for each few slices. Multiple nodes represent multiple positive input points for SAM.
- Hover one of the nodes and open the context menu. In the menu select the "Annotate Nodes with SAM" option.
- From there follow the instructions. The instructions should be sufficient to guide the user through the rest of the feature. If anything seems unclear, please include this in the feedback to improve the UX as good as possible.
- Please review the text of the modal & tooltip :)

### Bugs:
- [x] Trying an undo action after a skeleton sam prediction results in an error

### TODOs:
- [x] Support for all OrthoViewports
- [x] Make the whole prediction a single volume undo step
- [x] Make the interpolation a single volume undo step
- [x] Add a progress bar to the modal
- [x] Add a loading indication during the interpolation step
- [x] Ensure that the UI blocking animation is shown before the first interpolation block the UI

### Bugs:
- [x] Interpolation is executed regardless of whether the user closes the toast or clicks its button

### Concerns:
- [ ] After the prediction when the user corrects the prediction, this creates further annotation manipulation actions that are pushed onto the undo stack. As the stack has a max of 20 undo steps and navigating the prediction via "jump to next node" creates a new annotation action and therefore an undo stack entry, the undo step for the prediction might be pushed out of the stack pretty fast. Therefore, undoing a bad prediction might be a hassle once a few node navigations are done by the user. -> Not the best UX in my opinion...
  - Consider this in a follow up PR 

### Issues:
- fixes [#7541](https://github.com/scalableminds/webknossos/issues/7541)

------


MichaelBuessemeyer:
Closing as this feature does not provide enough value.

PR #7576, opened 2024-01-22
Include all-zero buckets in volume annotation downloads

### LINKED ISSUE #7532: Volume Annotation Download: keep all-zero chunks to support merging with fallback layer
Volume annotation downloads currently skip all-zero chunks. For annotations with fallback segmentations layers, this means the information whether a chunk was drawn over using all black, or whether it is untouched, is lost.

Original report: https://github.com/scalableminds/webknossos-libs/issues/974

I think just removing the `!isAllZero(data)` in `Zarr3BucketStreamSink` and `WKWBucketStreamSink` might not be enough, I have vague memories of a separate mechanism to indicate all-overwritten buckets, possibly also something with version restores. We should look into this in more detail.

Consider only including all-zero chunks in case of a fallback layer being present

fm3:
Fixed a bug where all-zero chunks/buckets were omitted when downloading volume annotation even in case of a fallback segmentation layer, where ther zeroed-bucket information is actually needed. Otherwise, it can’t be distinguished if the area should be black or should render the fallback layer data again.

Only “reverted” buckets (marked by a single byte with value zero) are still omitted even in case of fallback segmentation layers.

- https://downloadvolumezeroedbuckets.webknossos.xyz

### Steps to test:
- Create volume annotation on a dataset with an existing segmentation layer (e.g. l4_sample)
- With a large brush size erase a volume comprising multiple buckets (i.e in at least 32 sections)
- Download that volume annotation, reupload
- The transparent hole should still be there, and not look like this (old behavior)

### Issues:
- fixes #7532

------


CHANGELOG.unreleased.md
```diff
@@ -55,6 +55,7 @@ For upgrade instructions, please check the [migration guide](MIGRATIONS.released
 - Fixed the initialization of the mapping list for agglomerate views if json mappings are present. [#7537](https://github.com/scalableminds/webknossos/pull/7537)
 - Fixed a bug where uploading ND volume annotations would lead to errors due to parsing of the chunk paths. [#7547](https://github.com/scalableminds/webknossos/pull/7547)
 - Fixed a bug where listing the annotations of other users would result in empty lists even if there are annotations and you should be allowed to see them. [#7563](https://github.com/scalableminds/webknossos/pull/7563)
+- Fixed a bug where all-zero chunks/buckets were omitted when downloading volume annotation even in case of a fallback segmentation layer, where ther zeroed-bucket information is actually needed. [#7576](https://github.com/scalableminds/webknossos/pull/7576)
```
frcroth, line 58:
```suggestion
- Fixed a bug where all-zero chunks/buckets were omitted when downloading volume annotation even in case of a fallback segmentation layer, where their zeroed-bucket information is actually needed. [#7576](https://github.com/scalableminds/webknossos/pull/7576)
```

frcroth (approved):

PR #7577, opened 2024-01-22
Fix "Download Meshes" functionality

daniel-wer:
Avoid that zip.js dynamically creates a web worker using new Blob(...) which violates the CSP.
The z-worker.js seems to be requested for each mesh :shrug: but it's cached so it shouldn't be an issue.

- https://___.webknossos.xyz

### Steps to test:
- Load a couple of meshes
- Use "Download Meshes" in the context menu of a group in the segments tab (this fails on master due to CSP)
- Check that the single mesh download still works

------


philippotto (approved):
Thanks for fixing this! :pray: Changes LGTM. I didn't test it (CI is not ready yet), but I trust that you did so and wanted to leave an approval already.

dieknolle3333 (approved):
works great for me, thanks for looking into this :)

PR #7578, opened 2024-01-22
Support uploading neuroglancer precomputed and N5

### LINKED ISSUE #7544: Support uploading n5 and precomputed datasets
## Detailed Description
N5 and neuroglancer precomputed can already be added as remote datasets. However, some users would like to use the upload functionality.

## Context

frcroth:
- https://uploadmoredatatypes.webknossos.xyz

### Steps to test:
- Upload these datasets: https://drive.google.com/drive/folders/1Yp7FRQ0_sG8Fetd-MwCRjaXmaDMkLw5M?usp=drive_link

### TODOs:
- [x] Precomputed Multilayer
- [x] N5

### Issues:
- fixes #7544

------


webknossos-datastore/app/com/scalableminds/webknossos/datastore/explore/ExploreLocalLayerService.scala
```diff
@@ -42,25 +49,67 @@ class ExploreLocalLayerService @Inject()(dataVaultService: DataVaultService)
...
       dataSource = new DataSource(dataSourceId, relativeLayers, voxelSize)
     } yield dataSource
 
-  private def makePathsRelative(layers: Seq[ZarrDataLayer]): Seq[DataLayer] =
-    layers.map(l => l.copy(mags = l.mags.map(m => m.copy(path = m.path.map(_.split("/").takeRight(2).mkString("/"))))))
+  private def makeZarrDataLayerPathRelative(layer: ZarrDataLayer): DataLayer =
+    layer.copy(mags = layer.mags.map(m => m.copy(path = m.path.map(_.split("/").takeRight(2).mkString("/")))))
+
+  private def makeZarrSegmentationLayerPathRelative(layer: ZarrSegmentationLayer): DataLayer =
+    layer.copy(mags = layer.mags.map(m => m.copy(path = m.path.map(_.split("/").takeRight(2).mkString("/")))))
```
fm3, line 112: what is this doing? :thinking: 

Also, should probably be extracted to makeMagPathRelative since this is used twice here.

For N5/precomputed this is not needed?

webknossos-datastore/app/com/scalableminds/webknossos/datastore/services/uploading/UploadService.scala
```diff
@@ -354,27 +371,51 @@ class UploadService @Inject()(dataSourceRepository: DataSourceRepository,
...
+    } else if (looksLikeN5Multilayer(dataSourceDir).openOr(false)) {
+      UploadedDataSourceType.N5_MULTILAYER
+    } else if (looksLikeN5Layer(dataSourceDir).openOr(false)) {
+      UploadedDataSourceType.N5
     } else {
       UploadedDataSourceType.WKW
     }
 
-  private def looksLikeZarrArray(dataSourceDir: Path, maxDepth: Int): Box[Boolean] =
+  private def hasFileName(fileNames: List[String], dataSourceDir: Path, maxDepth: Int): Box[Boolean] =
```
fm3, line 386:
```suggestion
  private def containsMatchingFile(fileNames: List[String], dataSourceDir: Path, maxDepth: Int): Box[Boolean] =
```

fm3 (commented):
Looking good :)

Do you think it would help to introduce a trait LayerWithMagLocators that implements a withMags method (and does the switch case over the different implementations there?). That way we wouldn’t have to do all of this switch-casing everywhere in the business logic? Did not test if this would work but maybe it’s worth trying.

Also, thanks for going through the docs! :clap:

webknossos-datastore/app/com/scalableminds/webknossos/datastore/explore/ExploreLocalLayerService.scala
frcroth: Is used for zarr because of the different directory format (for ngff and zarr, there is an additional directory e.g. color)

frcroth:
> Looking good :)
> 
> Do you think it would help to introduce a trait LayerWithMagLocators that implements a withMags method (and does the switch case over the different implementations there?). That way we wouldn’t have to do all of this switch-casing everywhere in the business logic? Did not test if this would work but maybe it’s worth trying.

Have a look now. I think it does make sense, but I am not sure I like the name. Is there something more abstract that seperates layers with maglocators from others?

fm3:
> Have a look now. I think it does make sense, but I am not sure I like the name. Is there something more abstract that seperates layers with maglocators from others?

Good question! I’m afraid, not at this point. Layers without magLocators are currently
 - annotation layers
 - dataset layers with wkw format

In fact, I think we should add magLocator support for wkw layers as well, either with a migration of the jsons or an adapter that can read WKWResolution objects and turn them into MagLocators. This would also enable remote wkw datasets. But that does not have a high priority atm and it should not block this PR.

fm3 (approved):
:tada:

Issue #6303, opened 2022-06-23
Don't label data outside of dataset bbox


philippotto:
Labeling volume data outside of the dataset's bbox does not produce visible labels in the viewports. However, ad-hoc meshing the segment shows that data is stored outside of the dataset's bbox. Ideally, this data should not be produced so that other code (such as meshing or download) does not handle this data which the user has never seen before.

philippotto:
I think, buckets which are completely outside of the DS bbox are not created, anyway. However, buckets which overlap are probably created and bbox checks would need to be introduce dwhen labeling these.

Issue #5775, opened 2021-10-12
Layer should be clipped at bounding box


jstriebel:
Currently, if a layer has data outside of its bounding box and the bounding box is not aligned with the chunks, the data of the chunk crossing the bounding box is fully shown.

E.g. see the following dataset:
[skin_segmented_2021-10-12_12-20-21.zip](https://github.com/scalableminds/webknossos/files/7330078/skin_segmented_2021-10-12_12-20-21.zip)

Currently the `volume_annotation` shows too much data:
![Screenshot from 2021-10-12 13-52-47](https://user-images.githubusercontent.com/7216331/136960027-2f2c865c-ce87-42ce-9ca8-9794459ba396.png)

It should be cropped at the `400, 100, 0, 200, 700, 1` bounding box, but the shown data starts at `384, 96, 0`, which is divisible by `32`. The source of the data is a volume annotation with the following bounding box:
![Screenshot from 2021-10-12 13-52-30](https://user-images.githubusercontent.com/7216331/136960252-12b49df7-1990-488c-b839-115ffa33c9c1.png)

I tagged this as a frontend issue for now, but it could also be solved on the backend site by sending `0`s outside the valid bounding box.

PR #7580, opened 2024-01-23
Limit bbox of volume tracings to that of fallback layer

### LINKED ISSUE #7365: Create annotation layer with fallback layer: keep layer bbox
When creating an annotation on a dataset with fallback segmentation layer, the annotation layer currently gets the whole bbox of the dataset. It should be the fallback layer’s bbox instead.

fm3:
When creating or uploading a non-task volume annotation layer with a fallback segmentation layer, the annotation layer’s bounding box will now be limited to that layer’s, instead of the whole dataset’s.

- https://volumefallbackbbox.webknossos.xyz

### Steps to test:
- On a datast with a segmentation layer that is smaller than the color layer,
- Create a volume annotation with fallback segmentation (is the default)
- bbox should be limited to fallback layer’s
- should still look the same after download + reupload
- Note that the frontend does not currently “respect” the volume tracing’s bbox. This is fixed in https://github.com/scalableminds/webknossos/pull/8602 – it might make sense to test both PRs together
 
### Issues:
- fixes #7365

- Related PRs #8535, #8551, #8602
- Related issues #5775, #6303, #8067

------


philippotto:
> If a skeleton layer exists also, the volume layer bbox is not shown and the brushing is not restricted to the volume layer bbox
> If a skeleton layer does not exist, the volume layer bbox is shown, but you can still brush outside of it, but only unto the next bucket border

Regarding the differing visibility of the volume bbox, I'd have to investigate why this is the case.
That brushing is possible outside of the bbox (until the next bucket border) is known (see https://github.com/scalableminds/webknossos/issues/6303) and requires a bit of thought. Not super complicated, but also not a 3-line-change.

> segmentation data is rendered (also in dataset view mode) unto the next bigger bucket border (I reduced the segmentation layer bbox in the datasource-properties.json but didn’t delete the actual data outside of it)

This is also known (see https://github.com/scalableminds/webknossos/issues/5775). Each layer's bbox needs to be passed into the shader and then be handled accordingly.

Both issues together might take approx. 2 hours :tm:.

fm3:
Thanks for investigating! I think it would be worth tackling these two to three issues, but it certainly does not have the highest priority. I guess I’ll let this PR lie here until someone of the frontend team has some free capacity :)

MichaelBuessemeyer (approved):
Looks good as well and seems to work (at least it did during my testing)

Issue #5840, opened 2021-11-10
Unify REST API documentation and versioning


fm3:
## Detailed Description
- Currently the rest api versioning is in place only for the routes listed at https://docs.webknossos.org/webknossos/rest_api.html
- We should have a close look at how to unify this with the swagger annotations as used by the webknossos client.
- Also, we may have to add versioning for datastore (and possibly even tracingstore) routes.
- Additionally, adding new fields to wk’s responses was so far not considered a breaking change, and thus not a new API version. However, this may lead to too-new clients (which read those new fields) not being compatible with old webKnossos instances. We should decide whether this is a relevant problem and how to guard against it

## Draft
 - Let’s reuse the approach by the old `webknossos.versioned.routes` file, but let’s drop support for all versions before v5
 - Introduce the same routes versioning approach for datastore and tracingstore as well.
 - There should only be one version number for all three modules. So if one increases, bump the others too. Assert that this matches in the CI
 - In buildinfo (or separate route), report min and max supported version number
 - Clients (other than the js frontend) should use versioned routes. 
 - Unresolved Questions: 
   - how to guard against too-new clients? We don’t want to go through the version bumping routine every time we add a new route or field. Do we need to?
   - How to adapt swagger.json?

fm3:
We have removed both swagger annotations and the rest api document.

Users are now supposed to only use the webknossos-libs python client except in special cases.

The python client uses the versioned api, so when changing any api, we need to check if it is affected. If so, we need to increment the api version and consider compatibility in both directions.

The only thing remaining of this issue is the fact that the datastore and tracingstore api don’t yet use the versioning approach. When that becomes relevant, let’s open a new issue. I’ll consider this resolved now.

PR #7581, opened 2024-01-23
Include http api versions in buildinfo

fm3:
The buildinfo route now reports the supported HTTP API versions. Also the error message if an unsupported api version was requested is now more descriptive.

- https://httpapiversionbuildinfo.webknossos.xyz

### Steps to test:
- go to `api/buildinfo` should show 1 to 6 (Any suggestions on naming the fields?)
- request `api/v13/buildinfo`, should show readable error message.

### Issues:
- contributes to #5840 

------


frcroth (approved):
I'm fine with the names as is

Issue #7270, opened 2023-08-16
Coordinate Systems for Annotation Layers


philippotto:
## Context

* [ ] Specific to long-running jobs (set `jobsEnabled=true` in `application.conf`)
* [ ] Specific to webknossos.org (set `isDemoInstance=true` in `application.conf`)

PR #7588, opened 2024-01-25
Support transforming the skeleton layer with affine and TSP transforms

philippotto:
This PR takes care of transforming skeleton nodes (in the shader and when interacting with the positions) when necessary. During saving, the original node positions are stored (i.e., without any applied transformations).
In contrast to data layers (color and segmentation), the skeleton layer doesn't have an own transforms property. Instead it is treated like a layer that has no transforms at all. If a dataset with transformed layers exist, these layers will be transformed according to their transformations while the skeleton layer has the identity transforms. When one of the layers that have transforms is toggled to be rendered natively, the inverse transform is applied to the skeleton layer. As a result, the nodes will be transformed the same way the other layers are transformed.

When using the front-end export of skeletons to NML with transforms, the currently rendered node positions are written while the active transformation is encoded in the parameters section like the following. The array notation is loosely based on the TrakEm format (that uses something like `<t2_patch oid=“33” width=“949.0” height=“901.0” transform=“matrix(1.0,0.0,-0.0,1.0,53.0,72.0)... />`).

Affine:
```xml
<parameters>
    <experiment name="C555_DIAMOND_2f" description="" organization="sample_organization" wkUrl="http://localhost:9000" />
    ...
    <!-- The node positions in this file were transformed using the following affine transform -->
    <transform type="affine" positionsAreTransformed="true" matrix="[2.3271632194519043,12.151433944702148,0.6425480246543884,-5994.43896484375,-12.078832626342773,2.238189697265625,1.314159631729126,5853.94580078125,-1.229456901550293,0.8777050375938416,-12.358572006225586,8174.52685546875,0,0,0,1]" />
    ...
</parameters>
```

TPS:
```xml
  <parameters>
    ...
    <!-- The node positions in this file were transformed using a thin plate spline that was derived from the following correspondences: -->
    <transform type="thin_plate_spline" positionsAreTransformed="true">
      <correspondence source="[570.3021454931336,404.5548993008739,502.2248079151061]" target="[568.1197424615385,528.0132258461539,1622.1119725846156]" />
      ...
  </parameters>
....
```

If we decide in the future, that we always want to store the untransformed positions in the NML, we can do so and then change `positionsAreTransformed="true"` to false.

- https://transformskeletons.webknossos.xyz/

### Steps to test:
- create a new annotation for the C555_DIAMOND_2f-TPS dataset
- note that the skeleton layer in the left sidebar will have an icon representing whether it is rendered with or without transforms
- display the C555_DIAMOND_2f layer natively --> the skeleton layer will now be rendered with transforms
- use the download trees feature in the trees tab
  - there is one normal download (without transforms) and
  - a download with transforms
- the skeleton tool should be disabled now
- test the merger mode
- also test skeleton annotation in an nd dataset (without transforms)

### TODOs:
- [x] Disable skeleton annotation when the layer is transformed. otherwise, slightly weird behavior can occur, because the clicked position can slightly differ from the stored position
  - this is because the node positions are stored as integers. this leads to inaccuracies when going from the clicked position to an untransformed position (that is truncated) and then back again to the transformed coordinate space.
  - ~[ ] alternatively, we could change the storage format of node positions to float vectors <-- let's do this only when necessary
- [x] when exporting a NML, do we want to encode which transformation system was applied during the import?
  - [x] which format? how to encode TPS?
  - [X] is it fair to not transform the coordinates when using the back-end download? yes, by default, you get the export as it's stored in the backend. if you need the transformed nodes, you can use the tree download which asks whether you want this.
- [X] how should we treat the back-end export? we could leave it as is and argue that omitted transformation metadata in the NML encode that the node positions were exported without any transforms. then, it would be semantically correct.
- ~[ ] should the node property "position" in the DB be renamed to "transformedPosition" (as the front-end did it?)~

### Issues:
- contributes to #7270

------


frontend/javascripts/oxalis/geometries/materials/edge_shader.ts
```diff
@@ -23,14 +32,35 @@ class EdgeShader {
...
+    const nativelyRenderedLayerName =
+      Store.getState().datasetConfiguration.nativelyRenderedLayerName;
+    this.uniforms["transform"] = {
+      value: M4x4.transpose(
+        getTransformsForSkeletonLayer(dataset, nativelyRenderedLayerName).affineMatrix,
+      ),
     };
+    this.uniforms["has_transform"] = {
+      value: true,
+      // !_.isEqual(
```
philippotto, line 58: todo

frontend/javascripts/oxalis/geometries/materials/node_shader.ts
```diff
@@ -106,12 +119,69 @@ class NodeShader {
...
+      Store.getState().datasetConfiguration.nativelyRenderedLayerName;
+
+    this.uniforms["transform"] = {
+      value: M4x4.transpose(
+        getTransformsForSkeletonLayer(dataset, nativelyRenderedLayerName).affineMatrix,
+      ),
+    };
+    this.uniforms["has_transform"] = {
+      value: true,
+      // !_.isEqual(
```
philippotto, line 134: todo

frontend/javascripts/oxalis/geometries/materials/node_shader.ts
```diff
@@ -135,7 +206,16 @@ uniform int isTouch; // bool that is used during picking and indicates whether t
 uniform float highlightCommentedNodes;
 uniform float viewMode;
 
-<% _.each(additionalCoordinates || [], (_coord, idx) => { %>
+uniform mat4 transform;
+// todop: unused currently
```
philippotto, line 210: todo

frontend/javascripts/oxalis/merger_mode.ts
```diff
@@ -153,9 +160,10 @@ async function onCreateNode(
     return;
   }
 
+  // todop: inverse-transform of segmentation layer on pos
```
philippotto, line 163: todo

frontend/javascripts/oxalis/merger_mode.ts
```diff
@@ -321,7 +336,7 @@ async function mergeSegmentsOfAlreadyExistingTrees(
       // The node is not in bounds of the segmentation
       return;
     }
-
+    // todop: do inverse-transform of segmentation layer on pos
```
philippotto, line 339: todo

frontend/javascripts/oxalis/model/helpers/nml_helpers.ts
```diff
@@ -316,9 +317,10 @@ function serializeTrees(trees: Array<Tree>): Array<string> {
   );
 }
 
-function serializeNodes(nodes: NodeMap): Array<string> {
+function serializeNodes(state: OxalisState, nodes: NodeMap): Array<string> {
   return nodes.map((node) => {
-    const position = node.position.map(Math.floor);
+    // todop: we should probably encode in the NML that a transform was used
```
philippotto, line 322: todo

philippotto:
@fm3 I adapted the front-end so that a node position is transformed at the latest possible time (e.g., in the shader or shortly before presenting the current position in the UI). Within the redux store, the untransformed position is stored and that is also sent to the server. To make the semantics clear, I renamed `position` to `untransformedPosition` everywhere, except when doing the communication with the back-end. That way, everything works without having to change the back-end. However, I'm wondering whether it makes sense to also adapt the back-end (since there are now some places where the conversion between position <> untransformedPosition has to be done). What do you think about this? I assume it would be complicated because of the update actions that are already stored in the DB, right?

fm3:
> I assume it would be complicated because of the update actions that are already stored in the DB, right?

That’s right. We would either have to migrate all existing update actions or make both fields optional and use getOrElse logic, with an explicit error if both fields are None. While that would be possible, I don’t know if it would be nicer than just doing the renaming as you said

philippotto:
> That’s right. We would either have to migrate all existing update actions or make both fields optional and use getOrElse logic, with an explicit error if both fields are None. While that would be possible, I don’t know if it would be nicer than just doing the renaming as you said

Do you have a preference? I think, I'm against the getOrElse logic. This leaves us with two options: 1) keep it as is, 2) rename all existing update actions. I think, 2 would be a bit cleaner (but can also be weird, since `untransformedPosition` can be a bit confusing when there are no transforms in the first place). I'd be fine with 1, too, if you are?

fm3:
Yeah, I’d also vote to keep it. It’s not a wrong name anyway. Plus fossilDB migrations come with a bit of a hassle

daniel-wer:
@philippotto Reviewing, the following came to my mind: Do you think it would be helpful to leverage typescript to prevent mixing untransformed and transformed positions? The way to do it would be nominal/branded/opaque types, see [this example](https://www.typescriptlang.org/play#example/nominal-typing).

frontend/javascripts/oxalis/model/accessors/tool_accessor.ts
```diff
@@ -76,51 +77,97 @@ export function isTraceTool(activeTool: AnnotationTool): boolean {
...
+          "Skeleton annotation is disabled because the skeleton layer is transformed. Use the left sidebar to render the skeleton layer without any transformations.",
+      },
+    };
+  }
+
+  return {
+    [AnnotationToolEnum.SKELETON]: {
+      isDisabled: false,
+      explanation: "",
+    },
```
daniel-wer, line 121: Could be replaced with `NOT_DISABLED_INFO`.

frontend/javascripts/oxalis/model/helpers/nml_helpers.ts
```diff
@@ -285,13 +299,64 @@ function serializeParameters(
...
+  );
+
+  if (transform === IdentityTransform) {
+    return [];
+  }
+
+  if (transform.type === "affine") {
+    return [
+      serializeXmlComment(
+        "The node positions in this file were transformed using the following affine transform",
```
daniel-wer, line 323:
```suggestion
        "The node positions in this file were transformed using the following affine transform:",
```

frontend/javascripts/oxalis/model/sagas/min_cut_saga.ts
```diff
@@ -176,7 +176,10 @@ function removeOutgoingEdge(edgeBuffer: Uint16Array, idx: number, neighborIdx: n
 
 export function isBoundingBoxUsableForMinCut(boundingBoxObj: BoundingBoxType, nodes: Array<Node>) {
   const bbox = new BoundingBox(boundingBoxObj);
-  return bbox.containsPoint(nodes[0].position) && bbox.containsPoint(nodes[1].position);
+  return (
+    bbox.containsPoint(nodes[0].untransformedPosition) &&
+    bbox.containsPoint(nodes[1].untransformedPosition)
```
daniel-wer, line 181: In theory if a bounding box is involved, the transformed positions should be used, right? However, since volume annotation is not allowed if the skeleton is transformed, this should be ok. Still, I see this as a potential future bug source :thinking: Should this maybe be asserted that the skeleton layer is indeed not transformed (for example, once in the beginning when the tool saga is triggered)?

frontend/javascripts/oxalis/model/sagas/proofread_saga.ts
```diff
@@ -287,8 +287,10 @@ function* splitOrMergeOrMinCutAgglomerate(
   }
   const { layerName, agglomerateFileMag, getDataValue } = preparation;
 
-  const sourceNodePosition = sourceTree.nodes.get(sourceNodeId).position;
-  const targetNodePosition = targetTree.nodes.get(targetNodeId).position;
+  // Use untransformedPosition because agglomerate trees should not have
+  // any transforms, anyway.
```
daniel-wer, line 291: I'm not sure how future-proof this assumption is. I'd rather assert this explicitly or at least show a warning and log an error to airbrake.

frontend/javascripts/oxalis/model/sagas/proofread_saga.ts
```diff
@@ -451,9 +453,11 @@ function* performMinCut(
       let firstNodeId;
       let secondNodeId;
       for (const node of sourceTree.nodes.values()) {
-        if (_.isEqual(node.position, edge.position1)) {
+        // Use untransformedPosition because agglomerate trees should not have
```
daniel-wer, line 456: See my previous comment

frontend/javascripts/oxalis/view/right-border-tabs/skeleton_tab_view.tsx
```diff
@@ -737,11 +736,20 @@ class SkeletonTabView extends React.PureComponent<Props, State> {
...
+          title: "Download Visible Trees as NML",
         },
+        this.props.isSkeletonLayerTransformed
+          ? {
+              key: "handleNmlDownloadTransformed",
+              onClick: () => this.handleNmlDownload(true),
+              icon: <DownloadOutlined />,
+              label: "Download Visible Trees (Transformed)",
+              title: "The currently active transformation will be applied to each node.",
+            }
```
daniel-wer, line 751: I'm not sure whether I would offer the two download options. Instead, I'd always use the untransformed node positions and possibly include the active transforms if the skeleton is transformed. Consumers of the nml would need to apply the transformation themselves if they'd like to. If I remember correctly this is also what TrakEM does and the positions no longer would need to be transformed/untransformed during nml export/import.

daniel-wer (commented):
Very cool that this is supported now! I'll report back after testing :test_tube:

daniel-wer (commented):
Works very well :tada: Mainly, I think the NML export/import could be improved as well as two other small issues. My notes from testing:

- [x] The icons for the skeleton layer in the layers tab are not vertically aligned
  image
- [x] The tps correspondences are serialized as `<correspondence source="[570.3021454931336,404.5548993008739,502.2248079151061]" target="[568.1197424615385,528.0132258461539,1622.1119725846156]" />`. Is the `[570.3021454931336,404.5548993008739,502.2248079151061]` part XML-idiomatic and can it be parsed without resorting to eval or custom string splitting? Otherwise I'd change it to something more idiomatic.
- [ ] As mentioned in one of the comments: Exporting a transformed skeleton and then importing it into the same annotation when the skeleton layer is no longer transformed leads to the nodes ending up in the wrong location. When fixing this, pay attention to the fact that NMLs currently can also be imported when the skeleton is transformed.
- [ ] I'm not sure whether this broke in this PR: The merger mode modal states "8	Replace the color of the current active tree and its mapped segments with a new one.", but pressing 8 doesn't change anything.

frontend/javascripts/oxalis/merger_mode.ts
```diff
@@ -114,11 +125,21 @@ async function createNodeOverwrite(
...
+  // Calculate where the node is actually rendered.
+  const transformedNodePosition = transformNodePosition(untransformedPosition, Store.getState());
+
+  // Apply the inverse of the segmentation transform to know where to look up
+  // the voxel value.
+  const inverseSegmentationTransform = getInverseSegmentationTransformer(
+    Store.getState(),
+    segmentationLayerName,
+  );
+  const segmentPosition = inverseSegmentationTransform(transformedNodePosition);
```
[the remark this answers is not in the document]
philippotto: Yes!

philippotto:
> The tps correspondences are serialized as <correspondence source="[570.3021454931336,404.5548993008739,502.2248079151061]" target="[568.1197424615385,528.0132258461539,1622.1119725846156]" />. Is the [570.3021454931336,404.5548993008739,502.2248079151061] part XML-idiomatic and can it be parsed without resorting to eval or custom string splitting? Otherwise I'd change it to something more idiomatic.

It can be parsed with `JSON.parse` which is why I find it acceptable. Still, I don't find it really great, but the alternative with nested XML tags seemed too convoluted to me and TrakEm also does it similarily..

frontend/javascripts/oxalis/model/sagas/proofread_saga.ts
philippotto: I added a check for this which cancels proofreading and notifies the user. I didn't add the airbrake logging, because I think, that users will tell us if they really need this feature.

frontend/javascripts/oxalis/model/sagas/min_cut_saga.ts
philippotto: Good point! However, I'll remove the min_cut_saga in the next couple of days, anyway. See https://scm.slack.com/archives/C5AKLAV0B/p1706087477274379.

frontend/javascripts/test/libs/nml.spec.ts
```diff
@@ -230,6 +231,7 @@ test("NML serializing and parsing should yield the same state", async (t) => {
     initialState.tracing,
     enforceSkeletonTracing(initialState.tracing),
     BUILD_INFO,
+    false,
```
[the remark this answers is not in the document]
philippotto: Do you mean a spec that uses actual transforms or simply with identity transforms as some sort of smoke test?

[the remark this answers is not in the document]
daniel-wer: I was thinking of a case where the skeleton is transformed to test that the state is equal after exporting and importing again. This will be influenced by our internal discussion about the NML format I suppose.

daniel-wer:
> For these reasons, I'd rather not do this refactoring right now. I hope you agree!

Definitely agree, thanks for thinking this through :)

> It can be parsed with JSON.parse which is why I find it acceptable. Still, I don't find it really great, but the alternative with nested XML tags seemed too convoluted to me and TrakEm also does it similarily..

Ok, convinced :+1:

frontend/javascripts/oxalis/model/helpers/nml_helpers.ts
```diff
@@ -755,7 +826,7 @@ export function parseNml(nmlString: string): Promise<{
 
             const currentNode = {
               id: nodeId,
-              position: [
+              untransformedPosition: [
```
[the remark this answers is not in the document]
philippotto: As discussed internally, applying the inverse transform here wouldn't work with the use case where one wants to import nodes into another dataset where the transformed positions are expected. Right now, the user has to be careful about what kind of NML they export/import (with or without transforms). In the long run, we will probably want to improve this, but for this iteration it's okay in my opinion.

frontend/javascripts/test/libs/nml.spec.ts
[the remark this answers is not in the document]
philippotto: As per our discussion, I don't think that such a test would add much value right now, because no automatic retransformation should happen upon import.

philippotto:
@daniel-wer Thanks again for your review and testing :pray: Now that our [discussion](https://scm.slack.com/archives/C5AKLAV0B/p1708528595737289) has settled for now, the PR is ready from my point of view. There's certainly potential for improvement in the future, but as mentioned in another comment, I'd defer this for now.

> I'm not sure whether this broke in this PR: The merger mode modal states "8 Replace the color of the current active tree and its mapped segments with a new one.", but pressing 8 doesn't change anything.

Very interesting! I did some digging and it turns out that this feature was removed a while back in 812407957e8685886fe9364102486372eca17d7e. However, the modal wasn't updated then. I simply removed the hint from the modal now, because nobody complained about this feature being gone.

daniel-wer (commented):
Code LGTM and my previous points from testing were addressed :)

- [ ] During a renewed testing I noticed that the skeleton can still be modified even when the skeleton layer is transformed either by using keyboard shortcuts (`Ctrl` + `Arrow Keys` to move nodes, for example) or by using the context menu (Rightclick - Create Node Here). I'm also wondering whether "Import Agglomerate Skeleton" or "Import Synapses" from the context menu would do the correct thing for transformed skeleton layers or whether the entries should be disabled (maybe they are already, I didn't have an agglomerate file available for a transformed dataset).

philippotto:
>  During a renewed testing I noticed that the skeleton can still be modified even when the skeleton layer is transformed either by using keyboard shortcuts (Ctrl + Arrow Keys to move nodes, for example) or by using the context menu (Rightclick - Create Node Here).

Good find! I fixed it now :)

> I'm also wondering whether "Import Agglomerate Skeleton" or "Import Synapses" from the context menu would do the correct thing for transformed skeleton layers or whether the entries should be disabled (maybe they are already, I didn't have an agglomerate file available for a transformed dataset).

To properly support this scenario, we would need the ability that a skeleton layer uses the same transforms as a volume layer. We should tackle this once the need occurs :)

daniel-wer (approved):
LGTM :tada:

Issue #7280, opened 2023-08-17
Support ND data with arbitrary axis orders


fm3:
nd datasets currently only work if zyx are the last three axes in the underlying data. should work for all axis orders

> DatasetArray -> readAsFortranOrder only supports a shape/offset with XYZ at the end. This does not really make sense if we assume that xyz and additional coordinates may have any index/axisorder. Since only ngff datasets are currently supported for additional coordinates, and they follow the convention (t)(c) ... zyx, with additional coordinates before zyx, this works for now.

Fixing this should also include uint24 rgb layer support for ND datasets

PR #7592, opened 2024-01-29
ND data loading with other axis orders

fm3:
- enables ND data loading for all data formats with AxiOrder and AdditionalCoordinates (precomputed, n5, zarr3) (previously only zarr2)
 - volume annotation data.zips should now be viewable as datasets after unpacking (fixed the indices of their additional coordinates)
 - unifies DatasetArrayHandle (previously CubeHandle, no longer a trait with multiple implementations)
 - no more distinction between reading with or without AdditionalCoordinates from DatasetArray
 - Added some more tryos for better error handling
 - FullAxisOrder is now constructed for reading with arbitrary ND axis orders, with permutations from array order to wk order (tcxyz), including one for flipping c to fortran in the process
 - Adds some tests for these permutatinos

### Steps to test:
- Unpack zarr volume download data zip (both 3d and 4d), should be viewable as dataset
- View 4d and 3d and 2d zarr datasets, should be viewable

### TODOs:
- [x] fix volume annotation metadata (additional axes have wrong indices)
- [x] fix permute (needs to take additional axes into account)
- [x] test with volume annotations
- [x] test with 2d zarr
- [x] add tests for the permutations in c and f
- [x] fix viewing nd datasets *without* the additional coordinates in the json (cxyz should still be accessible if normal axisOrder field is correct)

### Issues:
- fixes [#7280](https://github.com/scalableminds/webknossos/issues/7280)

------


webknossos-datastore/app/com/scalableminds/webknossos/datastore/datareaders/DatasetArray.scala
```diff
@@ -116,37 +120,57 @@ class DatasetArray(vaultPath: VaultPath,
...
     val totalOffset: Array[Int] = offset.zip(header.voxelOffset).map { case (o, v) => o - v }.padTo(offset.length, 0)
-    val chunkIndices = ChunkUtils.computeChunkIndices(datasetShape.map(axisOrder.permuteIndicesReverse),
-                                                      axisOrder.permuteIndicesReverse(chunkShape),
+    val chunkIndices = ChunkUtils.computeChunkIndices(datasetShape.map(fullAxisOrder.permuteIndicesArrayToWk),
+                                                      fullAxisOrder.permuteIndicesArrayToWk(chunkShape),
                                                       shape,
                                                       totalOffset)
     if (partialCopyingIsNotNeeded(shape, totalOffset, chunkIndices)) {
       for {
         chunkIndex <- chunkIndices.headOption.toFox
```
frcroth, line 157:
```suggestion
        chunkIndexInWk <- chunkIndices.headOption.toFox
```
?

frcroth (approved):
Very good job!

webknossos-datastore/app/com/scalableminds/webknossos/datastore/datareaders/DatasetArray.scala
fm3: I’m hesitant, because then we would have to annotate shape, offset, chunkIndices, sourceChunk as well. It’s true that this ordering is important to understanding this function, but it is not the only thing going on here, so I’m scared the renamings would generate too much noise. Instead, I tried to make the comment more descriptive. What do you think?

PR #7594, opened 2024-01-29
Analytics in Postgres

normanrz:
- Adds tracking of analytics events in our own postgres database
- Adds the `POST /api/analytics/ingest` route to accept events from other wk instances, just like the current event-relay does

### Steps to test:
- `fetch("/api/analytics/ingest", {method:"POST", headers:{"content-type": "application/json"}, body: JSON.stringify({"events":[{"event_type":"open_annotation","user_id":"123456789012345678901234","time":"1706510609682","user_properties":{"organization_id":"123456789012345678901234","is_organization_admin":false,"is_superuser":false,"webknossos_uri":"https://webknossos.org"},"event_properties":{"annotation_id":"123456789012345678901234","annotation_owner_multiuser_id":"123456789012345678901234","annotation_dataset_id":"123456789012345678901234"},"session_id":1706510609681}], "api_key": "dingdong"})})` (also try with other `webknossos_uri` and `api_key`)
- Click around in wk and check the postgres database

### TODOs:
- [x] Add migration

------


fm3 (commented):
Looking good. I’ll test some once the CI is green

app/models/analytics/AnalyticsService.scala
```diff
@@ -1,49 +1,66 @@
...
 class AnalyticsService @Inject()(rpc: RPC,
                                  wkConf: WkConf,
                                  analyticsLookUpService: AnalyticsLookUpService,
-                                 analyticsSessionService: AnalyticsSessionService)(implicit ec: ExecutionContext)
+                                 analyticsSessionService: AnalyticsSessionService,
+                                 analyticsDAO: AnalyticsDAO)(implicit ec: ExecutionContext)
     extends LazyLogging {
 
+  private lazy val conf = wkConf.BackendAnalytics
+  private lazy val wellKnownUris = conf.wellKnownUris.map(_.split("\\|")).map(parts => (parts(0), parts(1))).toMap
```
fm3, line 24: should probably be wrapped in a tryo to prevent exceptions during initialization of this class. When using the result, match on the box state

app/models/analytics/AnalyticsService.scala
```diff
@@ -1,49 +1,66 @@
...
+      _ <- saveInDatabase(List(analyticsJson))
+      _ <- send(Json.toJsObject(analyticsJson))
     } yield ()
     () // Do not return the Future, so as to not block caller
   }
 
+  def ingest(jsonEvents: List[AnalyticsEventJson], apiKey: String): Fox[Unit] = {
+    val allApiKeysAreOk = !jsonEvents.exists(ev => {
+      wellKnownUris.get(ev.userProperties.webknossosUri).exists(wellKnownApiKey => wellKnownApiKey != apiKey)
+    })
```
fm3, line 39:
```suggestion
    val allApiKeysAreOk = jsonEvents.forall(ev => {
      wellKnownUris.get(ev.userProperties.webknossosUri).exists(wellKnownApiKey => wellKnownApiKey == apiKey)
    })
```

I think it reads more intuitively without all the negatives

app/controllers/Application.scala
```diff
@@ -52,6 +52,16 @@ class Application @Inject()(actorSystem: ActorSystem,
...
 
+  def ingestAnalyticsEvents: Action[AnalyticsEventsIngestJson] = Action.async(validateJson[AnalyticsEventsIngestJson]) {
+    implicit request =>
+      {
+        for {
+          _ <- bool2Fox(conf.BackendAnalytics.databaseEnabled) ?~> "Database logging of events is not enabled"
+          _ <- analyticsService.ingest(request.body.events, request.body.apiKey)
+        } yield Ok
+      }
+  }
```
fm3, line 63: Maybe extract to new AnalyticsController since we now have two routes concerned with this

app/models/analytics/AnalyticsService.scala
```diff
@@ -1,49 +1,66 @@
...
       }
-      val wrappedJson = Json.obj("api_key" -> wkConf.BackendAnalytics.key, "events" -> List(analyticsEventJson))
-      rpc(wkConf.BackendAnalytics.uri).silent.postJson(wrappedJson)
+      val wrappedJson = Json.obj("api_key" -> conf.key, "events" -> List(analyticsEventJson))
+      rpc(conf.uri).silent.postJson(wrappedJson)
+    }
+    Fox.successful(())
+  }
+
+  private def saveInDatabase(events: List[AnalyticsEventJson]): Fox[Unit] = {
```
fm3, line 61:
```suggestion
  private def saveInDatabaseIfEnabled(events: List[AnalyticsEventJson]): Fox[Unit] = {
```

conf/application.conf
```diff
@@ -327,7 +327,9 @@ backendAnalytics {
   uri = ${?DEV_WK_ANALYTICS_URI}
   key = ""
   sessionPause = 30 minutes
+  databaseEnabled = true
```
fm3, line 330:
```suggestion
  saveToDatabaseEnabled = true
```

conf/application.conf
```diff
@@ -327,7 +327,9 @@ backendAnalytics {
...
   verboseLoggingEnabled = false
+  wellKnownUris = ["https://webknossos.org|dingdong"] # flat list of uri-key pairs in the format: uri|key
```
fm3, line 332:
```suggestion
  wellKnownUris = [] # flat list of uri-key pairs in the format: uri|key. Example: "https://webknossos.org|secretKey". These are allowed to send their analytics events to this server for storage.
```

(Should be empty by default, otherwise any instance with this config will accept data with this not-really-secret key)

app/models/analytics/AnalyticsService.scala
```diff
@@ -1,49 +1,66 @@
...
+  private lazy val conf = wkConf.BackendAnalytics
+  private lazy val wellKnownUris = conf.wellKnownUris.map(_.split("\\|")).map(parts => (parts(0), parts(1))).toMap
+
   def track(analyticsEvent: AnalyticsEvent): Unit = {
     for {
       sessionId <- Fox.successful(analyticsSessionService.refreshAndGetSessionId(analyticsEvent.user._multiUser))
       analyticsJson <- analyticsEvent.toJson(analyticsLookUpService, sessionId)
-      _ <- send(analyticsJson)
+      _ <- saveInDatabase(List(analyticsJson))
+      _ <- send(Json.toJsObject(analyticsJson))
```
fm3, line 31:
```suggestion
      _ <- sendIfEnabled(Json.toJsObject(analyticsJson))
```

fm3 (commented):
I tested some locally and saving the events of the own instance works.

Did you also test the ingesting? Do you have steps to test for that?

I did see some `Unsuccessful WS request to https://events-relay.webknossos.org/events (ID: 0).Status: 500.` in the logging. Did something change there?

I also added some comments on the code.

fm3:
Testing the ingest worked with the fetch you provided. It also worked with wrong api keys and wrong uris, though. I couldn’t immediately spot the error in the code. Can you reproduce this? If so, maybe you can debug why it is not strictly checked.

fm3 (approved):

PR #7595, opened 2024-01-29
Add citation file

frcroth:
As proposed here: https://scm.slack.com/archives/C5AKLAV0B/p1677581788372419

This will create a "Cite this repository" button on GitHub. Check it out here: https://github.com/scalableminds/webknossos/tree/citation

In the file, only the part at preffered-citation is relevant, the lines above are only for standard compliance and ignored by GitHub

image


CITATION.cff
```diff
@@ -0,0 +1,66 @@
+﻿# Content: Citation File Format (CFF) file for webKnossos
+cff-version: 1.2.0
+title: "WEBKNOSSOS"
+message: >-
+  If you use this software, please cite it using the
+  metadata from this file.
+type: software
+authors:
+  - family-names: Boergens
+    given-names: Kevin M
```
normanrz, line 10: Shouldn't all authors be listed here as well?

frcroth: Doesn't have an impact on the Github citation feature since only the "preferred-citation" key is used for that, but yes, it is cleaner

normanrz (approved):

PR #7599, opened 2024-01-31
Upgrade to Java 21

### LINKED ISSUE #7491: Use newer JVMs in images + installation guides
As of play the upgrade #7420 and fossildb upgrade #7440, webknossos should support Java 17 and 21 as well as 11.

We still use 11 by default. Maybe we should move to 21?

frcroth:
- https://java.webknossos.xyz

### Steps to test:
- @fm3 you can test installing java as in the DEV_INSTALL

### Issues:
- fixes #7491

------


fm3 (commented):
Works for me :) @normanrz could you also have a look? Does this make sense for mac?
Do you think there may be disadvantages of using the eclipse temurin package and not openjdk for local development?

normanrz (approved):

PR #7600, opened 2024-01-31
Fix incorrect logic when deciding whether to use segment stats for adhoc meshing

philippotto:
- https://___.webknossos.xyz

### Steps to test:
- use adhoc meshing and ensure that the positions are queried first in the network tab

### Issues:
- follow-up for #7394

------


dieknolle3333 (approved):
Makes sense, thanks for fixing it :)

PR #7608, opened 2024-02-07
In proofreading min-cut, return directed edges to keep id of selected

### LINKED ISSUE #6470: [Proofreading] Keep color of active segment when splitting/merging
The active segment (the one with the cross) should keep its color after a min-cut (or merge).

fm3:
Min-cut returns a set of edges that are then passed to the proofreading split operation. These edges need to be directed so that when splitting the expected side keeps its agglomerate id. This PR uses the min cut’s source partition to determine which way the edges should be directed.

### Steps to test:
- In proofreading annotation, perform a split one way, merge again, perform the split the other way, repeat. Always the selected (first-clicked, cross-bearing) segment should keep its agglomerate id
- For merging this should have already worked, do some testing for that too

### Issues:
- fixes #6470

------


philippotto (approved):
Great! Works as expected :+1:

PR #7609, opened 2024-02-07
fixes for analytics event ingestion

normanrz:
Fixes for analytics event ingestion

------


fm3 (approved):
Could you try with `validate[Instant]`? That should already handle both numerals and strings.

Otherwise, LGTM

PR #7616, opened 2024-02-09
fix search input for publication gallery

hotzenklotz:
Small fix to restore the layouting of the publication gallery and its search input.

Before:
image

After:
image

### Steps to test:
- None.

### Issues:
- Follow up to https://github.com/scalableminds/webknossos/pull/7522

------

### DIFF
--- a/frontend/stylesheets/_utils.less
+++ b/frontend/stylesheets/_utils.less
@@ -66,6 +66,18 @@ td.nowrap * {
   float: right !important;
 }
 
+.clearfix::before {
+  // sometimes required by .pull-right
+  display: table;
+  content: '';
+}
+.clearfix::after {
+  // sometimes required by .pull-right
+  display: table;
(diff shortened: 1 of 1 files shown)


frontend/stylesheets/_utils.less
```diff
@@ -66,6 +66,18 @@ td.nowrap * {
   float: right !important;
 }
 
+.clearfix::before {
```
hotzenklotz, line 69: I believe these `clearfix` rules were previously provided by some antd default styles/resets. I took these CSS rules directly from the "old" CSS file that we shipped with antd4.

hotzenklotz: Bootstrap has a better explanation then me: https://getbootstrap.com/docs/4.0/utilities/clearfix/

philippotto (approved):

PR #7627, opened 2024-02-14
Add option to only track time when there are significant changes

### LINKED ISSUE #7543: Improve measurement of spent time for tasks/annotations
Time spent by annotators is measured automatically by wk. TimeSpans are stored in postgres describing the work.

When a user sends an annotation update request to the Tracingstore, it sends a summary on to the  webknossos server (`remoteWebKnossosClient.reportTracingUpdates`)/ `WKRemoteTracingStoreController.handleTracingUpdateReport`).

Some users have asked that move-only updates, where no segments/nodes were modified should not be counted for the time tracking. The report already contains a `significantChangesCount`.

An application.conf parameter should be added to disable timetracking for updates with `significantChangesCount==0`.

Note for testing this: currently, only timespans of tasks are available in the UI. You can still use postgres to inspect stored timespans for explorative annotations. Compare https://github.com/scalableminds/webknossos/pull/7524

More sophisticated ideas (that we decided against for the moment) include
 - Setting per project instead of instance-wide application.conf setting
 - For each update, use only the timestamps that stem from “significant changes”
 - Also include timestamps for move-only events *except* if move-only occurs for more than x seconds

## Context
[Internal discussion.](https://scm.slack.com/archives/C5AKLAV0B/p1704804553128329)

frcroth:
Set the new application.conf option `webKnossos.timeTrackingOnlyWithSignificantChanges` to `false` to disable time tracking for changes where only camera movements happened.

Note that the timeTrackingPause is still active, so if there are “significant changes” within that duration again, the timespans will be extended.

### Steps to test:
- Set option to true, no time tracking for moving cam and saving
- Set option to false, time tracking for moving cam and saving

### Issues:
- fixes #7543 

------

### DIFF
--- a/app/controllers/WKRemoteTracingStoreController.scala
+++ b/app/controllers/WKRemoteTracingStoreController.scala
@@ -60,8 +60,7 @@ class WKRemoteTracingStoreController @Inject()(
             annotationLayerDAO.updateStatistics(annotation._id, report.tracingId, statistics)
           }
           userBox <- bearerTokenService.userForTokenOpt(report.userToken).futureBox
-          trackTime = (!wkConf.WebKnossos.User.timeTrackingOnlyWithSignificantChanges ||
-            (wkConf.WebKnossos.User.timeTrackingOnlyWithSignificantChanges && report.significantChangesCount > 0))
+          trackTime = (report.significantChangesCount > 0 || !wkConf.WebKnossos.User.timeTrackingOnlyWithSignificantChanges)
           _ <- Fox.runOptional(userBox)(user =>
             Fox.runIf(trackTime)(timeSpanService.logUserInteraction(report.timestamps, user, annotation)))
           _ <- Fox.runOptional(userBox)(user =>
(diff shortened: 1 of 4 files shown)


app/controllers/WKRemoteTracingStoreController.scala
```diff
@@ -58,7 +60,10 @@ class WKRemoteTracingStoreController @Inject()(
             annotationLayerDAO.updateStatistics(annotation._id, report.tracingId, statistics)
           }
           userBox <- bearerTokenService.userForTokenOpt(report.userToken).futureBox
-          _ <- Fox.runOptional(userBox)(user => timeSpanService.logUserInteraction(report.timestamps, user, annotation))
+          trackTime = (!wkConf.WebKnossos.User.timeTrackingOnlyWithSignificantChanges ||
+            (wkConf.WebKnossos.User.timeTrackingOnlyWithSignificantChanges && report.significantChangesCount > 0))
```
fm3, line 64: I feel like this expression can be simplified. Didn’t verify, but shouldn’t `count > 0 || (!onlySignificant)` be the same? I think it reads more nicely.

fm3 (commented):
Looking good! I don’t think we need to group the config keys just yet. Please see my comment about simplifying the boolean expression

fm3 (approved):
Nice!

I wrote a follow-up issue for proofreading annotations, since they don’t get “automatically” time-tracked anymore if this new config option is true. https://github.com/scalableminds/webknossos/issues/7637

PR #7639, opened 2024-02-20
Fix offset when creating node with mouse

philippotto:
The border around the viewports caused a slight offset when creating a node (probably, other interactions were also impacted). Debugging showed that the mouse move listeners used `getBoundingClientRect` (which contains the border) to make the clicked position relative. However, the remaining code assumes that the border has to be ignored when interacting with the div. For example, `makeRectRelativeToCanvas` subtracts the border after calling `getBoundingClientRect`. I think, it makes sense to ignore the border altogether. However, doing this in `input.ts` would mean to dynamically read the border and subtract it there.
I figured the easiest way was to move the border simply somewhere else so that the targeted viewport divs don't have any border.

@daniel-wer I suggest to test this first thoroughly. If you are fine with this, I'd remove OUTER_CSS_BORDER completely, since it shouldn't have any relevance afterwards (hopefully).

By the way, a smell of the current border was also that you could create nodes on the border. With this PR, this is not possible anymore.

- https://fixnodecreationoffset.webknossos.xyz

### Steps to test:
- test that the mouse position is respected precisely when
  - creating, moving, selecting nodes (remember that node positions are integer; so zoom out a bit to get precise positions)
  - brushing
  - moving with alt + mouse move
  - something else?

### TODOs:
- [x] fix brush preview
- [x] remove OUTER_CSS_BORDER

### Issues:
- fixes https://scm.slack.com/archives/C5AKLAV0B/p1708346079685159

------

### DIFF
--- a/frontend/javascripts/oxalis/constants.ts
+++ b/frontend/javascripts/oxalis/constants.ts
@@ -264,10 +264,6 @@ export enum TreeTypeEnum {
 export type TreeType = keyof typeof TreeTypeEnum;
 export const NODE_ID_REF_REGEX = /#([0-9]+)/g;
 export const POSITION_REF_REGEX = /#\(([0-9]+,[0-9]+,[0-9]+)\)/g;
-// The plane in orthogonal mode is a little smaller than the viewport
-// There is an outer yellow CSS border and an inner (red/green/blue) border
-// that is a result of the plane being smaller than the renderer viewport
-export const OUTER_CSS_BORDER = 0;
 const VIEWPORT_WIDTH = 376;
 export const ensureSmallerEdge = false;
 export const Unicode = {
--- a/frontend/javascripts/oxalis/constants.ts
+++ b/frontend/javascripts/oxalis/constants.ts
@@ -267,7 +267,7 @@ export const POSITION_REF_REGEX = /#\(([0-9]+,[0-9]+,[0-9]+)\)/g;
 // The plane in orthogonal mode is a little smaller than the viewport
 // There is an outer yellow CSS border and an inner (red/green/blue) border
 // that is a result of the plane being smaller than the renderer viewport
-export const OUTER_CSS_BORDER = 2;
+export const OUTER_CSS_BORDER = 0;
 const VIEWPORT_WIDTH = 376;
 export const ensureSmallerEdge = false;
 export const Unicode = {
--- a/frontend/javascripts/oxalis/view/input_catcher.tsx
+++ b/frontend/javascripts/oxalis/view/input_catcher.tsx
@@ -146,30 +146,32 @@ function InputCatcher({
   );
 
   return (
-    <div
-      className="flexlayout-dont-overflow"
-      onContextMenu={ignoreContextMenu}
-      style={{ cursor: busyBlockingInfo.isBusy ? "wait" : cursorForTool[adaptedTool] }}
-    >
+    <div className={`inputcatcher_border ${viewportID}`}>
       <div
-        id={`inputcatcher_${viewportID}`}
-        ref={(domElement) => {
-          domElementRef.current = domElement;
-        }}
-        data-value={viewportID}
-        className={`inputcatcher ${viewportID}`}
-        style={{
-          position: "relative",
-          // Disable inputs while wk is busy. However, keep the custom cursor and the ignoreContextMenu handler
-          // which is why those are defined at the outer element.
-          pointerEvents: busyBlockingInfo.isBusy ? "none" : "auto",
-        }}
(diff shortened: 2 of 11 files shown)


frontend/stylesheets/trace_view/_tracing_view.less
```diff
@@ -18,18 +18,15 @@
   position: absolute;
   width: 100%;
 }
-.inputcatcher {
+.inputcatcher_border {
```
daniel-wer, line 21: :snake:
```suggestion
.inputcatcher-border {
```

frontend/javascripts/oxalis/view/input_catcher.tsx
```diff
@@ -146,30 +146,32 @@ function InputCatcher({
   );
 
   return (
-    <div
-      className="flexlayout-dont-overflow"
-      onContextMenu={ignoreContextMenu}
-      style={{ cursor: busyBlockingInfo.isBusy ? "wait" : cursorForTool[adaptedTool] }}
-    >
+    <div className={`inputcatcher_border ${viewportID}`}>
```
daniel-wer, line 149: :snake:
```suggestion
    <div className={`inputcatcher-border ${viewportID}`}>
```

frontend/stylesheets/trace_view/_tracing_view.less
```diff
@@ -42,10 +39,15 @@
...
   }
-
-  &:hover {
-    opacity: 1;
-  }
+}
+.inputcatcher {
+  z-index: 20;
+  box-sizing: border-box;
+  -moz-box-sizing: border-box;
```
daniel-wer, line 46: This should no longer be needed since the element doesn't have a border. I'm not sure whether it should be moved to `inputcatcher_border` instead :see_no_evil:

daniel-wer:
The first thing I noticed comparing before and after is that the scale bar got a little wider with this PR:

Before
image

After
image

When setting nodes at the edges of the scalebar and measuring the distance I get the sense that the new one is too wide and the old one is correct. I'll test more thoroughly tomorrow.

frontend/stylesheets/trace_view/_tracing_view.less
philippotto: You are right. I think, it can be removed altogether, because no code is interested in measuring (or sizing) the border-only container.

philippotto:
> When setting nodes at the edges of the scalebar and measuring the distance I get the sense that the new one is too wide and the old one is correct.

Good catch! I confirm this bug. However, I don't understand it. I removed the CSS border completely and there is still an offset. I suspect that there is another hardcoded border somewhere, but I don't know where :thinking: Other than that, I don't have an explanation right now. I'll look into this again tomorrow, too.

frontend/javascripts/oxalis/constants.ts
```diff
@@ -267,7 +267,7 @@ export const POSITION_REF_REGEX = /#\(([0-9]+,[0-9]+,[0-9]+)\)/g;
 // The plane in orthogonal mode is a little smaller than the viewport
 // There is an outer yellow CSS border and an inner (red/green/blue) border
 // that is a result of the plane being smaller than the renderer viewport
-export const OUTER_CSS_BORDER = 2;
+export const OUTER_CSS_BORDER = 0;
```
philippotto, line 270: todo: remove the variable completely.

frontend/javascripts/oxalis/view/scalebar.tsx
```diff
@@ -80,7 +80,7 @@ function Scalebar({ zoomValue, dataset, viewportWidthInPixels, viewportHeightInP
             ? 16
             : `calc(${scaleBarWidthFactor * 100}% - ${Math.round(
                 ((2 * OUTER_CSS_BORDER) / constants.VIEWPORT_WIDTH) * 100,
-              )}% + ${2 * padding}px)`,
```
philippotto, line 83: I think, it was a coincidence that this worked before, because padding is 2px (exactly as the old OUTER_CSS_BORDER value).

daniel-wer (approved):
Nice find and fix! I'm glad that we can get rid of this prone-to-error piece of code :tada: I didn't notice any issues during testing.

philippotto:
> Nice find and fix! I'm glad that we can get rid of this prone-to-error piece of code 🎉 I didn't notice any issues during testing.

Great, thanks for testing! I noticed that there is still a slight change of the scalebar width (compared to master), but it's only 1px and it's hard to say whether the old or the new width is correct. Since the new code is way simpler, I'd like to assume that it's now correct =)

PR #7640, opened 2024-02-20
Make tag search precise

### LINKED ISSUE #7629: Tag selection of annotations does not work with single letter tags
## Context
I tried to tag a couple of annotations in the annotation dashboard as quickly as possible. I decided to just use a single letter "t". When I now select this tag, I still see all annotations in the dashboard

## Expected Behavior
I would expect that webknossos only shows the tagged annotations

## Current Behavior
Webknossos shows a strange subset of all annotations. I need at least 3 characters "ttt" to see only the tagged annotations.

## Your Environment for bug

- Browser name and version: Firefox 122.0.1 (64-bit)
- Operating System and version: Windows 10
- Version of WEBKNOSSOS (Release or Commit): 24.02.0

philippotto:
The current tag search produces false positives if the tag name also appears somewhere else in the annotation properties.

- https://___.webknossos.xyz

### Steps to test:
- assign a `t` tag to an annotation in the dashboard
- rename another annotation to contain a "t"
- click on the t tag and check that only the one annotation is shown

### Issues:
- fixes #7629

------


MichaelBuessemeyer (approved):
Thanks for fixing this :pray: :rocket:.

The code looks good to me :heavy_check_mark: 

Edit: Testing also worked :tada:

PR #7643, opened 2024-02-21
Do not require binaryData dir to be writable for Compose, orga dir is enough

fm3:
Not the whole binaryData dir needs to be writable.

Also removes the allowsUpload check for dataset composition.

If there is no write access at all, the request will ultimately still fail, of course.


philippotto (approved):
lgtm :+1:

[related slack discussion](https://scm.slack.com/archives/C5AKLAV0B/p1708528915861279)

webknossos-datastore/app/com/scalableminds/webknossos/datastore/services/DataSourceService.scala
```diff
@@ -53,6 +53,9 @@ class DataSourceService @Inject()(
     if (inboxCheckVerboseCounter >= 10) inboxCheckVerboseCounter = 0
   }
 
+  def assertDataDirWritable(organizationId: String): Fox[Unit] =
+    Fox.bool2Fox(Files.isWritable(dataBaseDir.resolve(organizationId))) ?~> "Datastore can not write to its data directory."
```
philippotto, line 57:
```suggestion
    Fox.bool2Fox(Files.isWritable(dataBaseDir.resolve(organizationId))) ?~> "Datastore cannot write to its data directory."
```

PR #7645, opened 2024-02-22
Fix nightly

philippotto:
Necessary due to #7639. The nightly fix for wkorg can be done once this PR is deployed.


frontend/javascripts/oxalis/view/input_catcher.tsx
```diff
@@ -146,7 +146,10 @@ function InputCatcher({
   );
 
   return (
-    <div className={`inputcatcher-border ${viewportID}`}>
```
hotzenklotz, line 149: Are these `viewportID`s used anywhere else besides the screenshot tests? In the LESS styles? Or `document.getElementById()`?

hotzenklotz (approved):

frontend/javascripts/oxalis/view/input_catcher.tsx
philippotto: If you mean the `screenshot_target_inputcatcher_*` ids, then, no, they are only used in the screenshot tests. Initially, I tried to avoid these additional IDs, but couldn't find a good way to get the correct DOM elements with puppeteer.

PR #7657, opened 2024-02-28
Add automatic dev deployment on push

frcroth:
Currently working on editing this comment, but the Github Token does not alow for that.
 
https://___.webknossos.xyz
 
 ------
 - [ ] Readd tests in action after new behavior is tested / clear what it should do


fm3:
could we have this maybe only if the PR is ready for review?

frcroth:
> could we have this maybe only if the PR is ready for review?

Done now. As discussed after the presentation, it is not exactly clear how to proceed here. i.e. require a label, what should be the default, is draft status relevant?

normanrz:
I like the idea of a `devdeploy` label. Draft state shouldn't matter. Can actions be triggered when a label has been added/removed?

frcroth:
> I like the idea of a `devdeploy` label. Draft state shouldn't matter. Can actions be triggered when a label has been added/removed?

Yes, it is possible to run a GitHub action on a label change (https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request). However, this would then be a GitHub action and not a circleCI action.
This might also make it possible to change the PR comment more easily to include a link to the dev instance.
This would require the secret for kubernetix also be present in this repo.

Another question would be when this should run:
- Immediately after the label is added and only then
- If the label is placed on the PR after the CircleCI run is done (once)
- above but every time the CI run is done and the label is still active.

normanrz:
> * above but every time the CI run is done and the label is still active.

Maybe it would be best to just check at the end of every CircleCI run whether the label is present and deploy/update accordingly. Removing deployments would still be manual.
Running it in a separate Github action seems to introduce a race condition and I wouldn't want to switch to Github actions entirely just for this auto-deploy.

normanrz:
Would it make things easier to use the `gh` CLI tool in CircleCI? It also seems straight-forward to change the PR text: https://cli.github.com/manual/gh_pr_edit. I think we already have a Github token in CircleCI for releases.

frcroth:
Waiting for https://github.com/scalableminds/webknossos/pull/8147

frcroth:
With Github CI now used, this should be simpler to do. Closing this PR for now, since this one dealt with circle CI, we can also create a new one.

PR #7659, opened 2024-02-29
Avoid left join fanout effect in annotation list query

### LINKED ISSUE #7647: Audit all queries containing LEFT JOIN for fanout effect causing duplicates
Bugfix https://github.com/scalableminds/webknossos/pull/7646 fixes this for the user list query.

There is a good chance that similar problems could occur for other queries (e.g. dataset list?)

fm3:
I went through queries that have LEFT JOIN looking for possible fanout effect.

- User list  was fixed in #7646 (subqueries with WITH)
- Annotation list is fixed here (subquery with WITH)

Most other queries should be fine
- Dataset list is already ok (inline subqueries)
- Folder access queries are ok (duplicates don’t hurt)
- Progress reports are ok (subqueries)
- Task assignment query is ok (duplicates don’t hurt)

I did not check the voxelytics queries, @normanrz maybe you could have a quick look there?

- https://avoidjoinfanout.webknossos.xyz

### Steps to test:
- For some annotations, set multiple shared teams in annotation share modal
- annotation list in dashboard should not contain duplicate teams

### Issues:
- fixes #7647 

------

### DIFF
--- a/.github/common_edge_cases.md
+++ b/.github/common_edge_cases.md
@@ -9,3 +9,7 @@ Will this also work if
  - The user is logged out and views a public dataset or annotation
  - User uses dark mode / light mode
  - There is no local datastore/tracingstore module (Compare [instructions to test this locally](https://github.com/scalableminds/webknossos/wiki/Set-up-a-standalone-datastore-locally))
+
+ Also check that
+  - Complex SQL queries have no fan out effect due to multiple left joins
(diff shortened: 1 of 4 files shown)


frcroth (approved):

fm3:
(for the record, Norman said he checked the voxelytics queries, should be fine too)

PR #7669, opened 2024-03-06
Switch to root folder if current folder cannot be loaded

### LINKED ISSUE #7658: When  dashboard localstorage points to deleted folder, go to root instead
Currently, the »could not find the requested folder« toast is shown but no datasets and no folder tree. Only after Refresh you see the root folder again. My test case is re-visiting the dashboard after `yarn refresh-schema`.

Could this situation do this refresh automatically (fetch the new folder tree, display root)?

This would greatly improve developer convenience and I think it also makes sense for users (someone else may have deleted a folder in the meantime)

philippotto:
- https://___.webknossos.xyz

### Steps to test:
- create a new folder
- yarn refresh-schema
- reload page

### Issues:
- fixes #7658

------

### DIFF
--- a/CHANGELOG.unreleased.md
+++ b/CHANGELOG.unreleased.md
@@ -25,6 +25,7 @@ For upgrade instructions, please check the [migration guide](MIGRATIONS.released
 - Added more documentation for N5 and Neuroglancer precomputed web upload. [#7622](https://github.com/scalableminds/webknossos/pull/7622)
 - Added the config key `webKnossos.user.timeTrackingOnlyWithSignificantChanges`, which when set to `true` will only track time if the user has made significant changes to the annotation. [#7627](https://github.com/scalableminds/webknossos/pull/7627)
 - Only display UI elements to launch background jobs if the (worker) backend actually supports them. [#7591](https://github.com/scalableminds/webknossos/pull/7591)
+- If the current dataset folder in the dashboard cannot be found (e.g., because somebody else deleted it), the page navigates to the root folder automatically. [#7669](https://github.com/scalableminds/webknossos/pull/7669)
 
 ### Fixed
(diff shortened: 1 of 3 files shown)


fm3:
I’m afraid this doesn’t quite work for me yet :grimacing: It looks like some folders briefly flash but then it switches to the blank version again after a few ms. Does it work for you, did you test with refresh-schema? I don’t know what’s different, maybe it’s a firefox vs chrome thing? :thinking:

philippotto:
> I’m afraid this doesn’t quite work for me yet 😬 It looks like some folders briefly flash but then it switches to the blank version again after a few ms. Does it work for you, did you test with refresh-schema? I don’t know what’s different, maybe it’s a firefox vs chrome thing? 🤔

Initially, I simply tried `localStorage.setItem("sample@scm.io-sample_organization-activeFolderId", "65e871d4ec0000dc04353aee")` and then reloaded the page. That worked even on master. Then, I tried `yarn refresh-schema` which was problematic. With my fix, a simple reload worked. However, something might have changed inbetween.
Maybe the folder hierarchy is cached internally and if that cache is outdated, the switch to the root doesn't work.
I'll look into it again :shrug: 

It's a bit annoying to test since refreshing the schema takes so long. Do you know whether there is a faster way to clear the folders?

fm3:
I’m afraid manually editing folders in the DB is not going to be simpler or faster than yarn refresh-schema :/

philippotto:
Please try again! I think it already worked before, but there was quite a bit of a delay until the switch would happen. The active folder is fetched and if that fails, it is retried three times. Due to an exponential back off in the delay time (default of react-query), this took at least 6 seconds. Now, it's down to 1.5s. I hope this wait is acceptable :)

By the way, deleting the `rm -f target/scala-2.13/src_managed/schema/com/scalableminds/webknossos/schema/Tables.scala` bit from `yarn refresh-schema` helped a bit :)

philippotto:
Unfortunately, it's not trivial :S

fm3 (approved):

PR #7671, opened 2024-03-06
Improve layer overflow menu

### LINKED ISSUE #7585: Improve overflow menu on layers in left sidebar
## Detailed Description
I find the overflow menu in the layer settings in the left-hand sidebar quite difficult to use. I think there are 2 main reasons:
- There is a delay before the menu opens when hovering. Plus, the animation takes some time, too.
- The hover target is pretty small. It is really just the 3-dots and a few pixels below and above. Moving the cursor can lead to the menu disappearing and reappearing or move around which is disorienting.

philippotto:
- make click target for overview menu bigger
- only use "hover" as trigger to avoid that another click closes the menu again

- https://___.webknossos.xyz

### Steps to test:
- use the menu

### Issues:
- fixes #7585


philippotto:
> Does this mean that the menu cannot be opened on touchscreens? 😬 (might be fair, if nothing super important hides inside it)

To be honest, I hadn't thought about it. But luckily it works, because the hover handlers are typically also triggered on touch. Just tested it, too.

PR #7680, opened 2024-03-08
Workaround for wrong menu position when clicking on ... next to a segment

philippotto:
Also, enlarge the clickable area for a segment item.
When https://github.com/react-component/trigger/pull/447 has bubbled through to antd, we can remove the `alignPoint={false}` part again.

- https://___.webknossos.xyz

### Steps to test:
- use context menu and the ... button next to a segment

### Issues:
- follow-up for #7522 

------


hotzenklotz (approved):
Great find! Thank you so much. 
Works as expected.

PR #7682, opened 2024-03-12
Manage AI Model Trainings + Inference

### LINKED ISSUE #7675: [AI models] Models in the database and CRUD routes
## Detailed Description
See https://github.com/scalableminds/webknossos/issues/7670 for a description of the models.

- Model entity
  - _id // also used as file path on datastore file system
  - _organization
  - _dataStore
  - _user
  - _training_job (NULLABLE)
  - name (UNIQUE per organization)
  - isDeleted
  - created
  - modified
  - workflow_yaml // full content
  - comment

- Model annotation entity (annotations a model should be trained with)
  - _model
  - _annotation

- Model inference entity
  - _id
  - _model
  - _dataset (NULLABLE, newly created dataset)
  - _annotation (base annotation specifying the bbox (and mask if mask_layer_name not null)
  - segmentation_layer_name (newly created, user-specified)
  - mask_layer_name (NULLABLE from the base annotation)
  - _inference_job
  - workflow_yaml

- Job entity
  - Add _voxelytics_run (NULLABLE)

Creating a model and model-inference should trigger and connect their respective worker jobs.

fm3:
- https://___.webknossos.xyz

### Steps to test:
- check out https://github.com/scalableminds/voxelytics/pull/3491
- `yarn enable-jobs`
- as super user:
  - open an annotation for a dataset
  - create at least one bounding box (sized  2×20×18 voxel)
  - annotate segments in the bounding boxes
  - click the AI Analysis button and switch to the train tab
  - train a model
  - once the training is done, click the AI Analysis button and switch to the run-a-model tab
  - enable the advanced mode at the top right
  - select the model you trained
  - run it on a bounding box
- as a non-super-user
  - the train-tab and the advanced mode in the run-model tab should not be available

### TODOs:
- [x] Requirements
  - [x] if training fails, keep the model object? should the name stay reserved? → yes
  - [x] should hard delete be possible? → delete from postgres (isDeleted=true), keep on disk
  - [x] annotation required for inferral? → no, but if there is one, store it with the inferral object
  - [x] storing models by mongo id in orga dir ok? → yes

- [x] Backend
  - [x] First version of new Tables in SQL Schema
    - [x] Comment should be optional
  - [x] models as scala case classes
    - [x] AiModel
      - [x] With trainingAnnotations
    - [x] AiInference
  - [x] publicWrites
    - [x] AiModel
    - [x] AiInference
  - [x] Relevant CRUD Routes
    - [x] Run training
    - [x] Read one aiModel
    - [x] List aiModel
    - [x] Update aiModel (name, comment)
    - [x] Run inference
    - [x] Read one inference
    - [x] List inferances
    - [x] Import model? → use manual inserts for now
    - [x] Delete aiModel (use isDeleted flag, keep on disk)
  - [x] Route to attach voxelyticsWorkflow to job from worker
  - [x] Route to attach dataset to inference run from worker
  - [x] add all required job parameters (layer names? bboxes?)
  - [x] Foreign keys
  - [x] Evolutions
  - [x] job result slack link
- [x] Worker https://github.com/scalableminds/voxelytics/pull/3491
  - [x] train_model job
  - [x] infer_with_model job 
  - [x] base yamls
  - [x] insert values
  - [x] run voxelytics
  - [x] copy out result
  - [x] attach voxelyticsWorkflow to jobs
  - [x] attach dataset to inference run
  - [x] select best-fitting mags for inference
  - [x] reduce duplication with preset-model inference jobs
  - [x] config setting for `--cpu`
- [x] Frontend
  - [x] Modals
  - [x] Use CRUD routes
  - [x] job list view
  - [x] deal with nullable trainingjob reference

### Issues:
- fixes #7675

------


fm3:
From my point of view the backend and worker side are done :) There are two more frontend checkboxes in the PR description, but I may not have thought of everything that is relevant. Other interesting things may be raw yaml input / info about training bbox sizes / a note saying that the inference bbox will be padded to not reach the dataset border because of the field of view / whatever else you may think of :innocent: 

Please let me know if any other worker or backend changes are needed

philippotto:
> raw yaml input

does the back-end support this already?

> info about training bbox sizes

what exactly are the expected sizes? then, I could put them in the modal.

----------------------

> There are two more frontend checkboxes in the PR description

I pushed the two missing things, now. Wasn't sure about the details, but if my assumptions below are true, it should be done.

>  job list view

The new job type is now listed in the jobs table, too. The table cannot display anything actionable for the user, though, right? In the actions table, I merely put a hint with "The model may now be selected from the &quot;AI Analysis&quot; button when viewing a dataset.".

> deal with nullable trainingjob reference

Currently, the front-end only uses the trainingJob property to ensure that its job state is "success". I assume, when trainingJob is null, "success" is implied because the model was placed manually, right?

fm3:
>> raw yaml input

> does the back-end support this already?

It should :tm: `workflowYaml` is an optional string param for the run training route. This yaml will be executed by the worker as is, so we need to change that again when we release this feature for non-superusers… I did not yet test this feature end-to-end. The user will also have to know what kind of paths to enter in the yaml, but that should be acceptable for the moment.

>> info about training bbox sizes

> what exactly are the expected sizes? then, I could put them in the modal.

Currently, there should be at least two boxes. The size of the boxes should be a multiple of 20×20×18 voxels in the training mag (which the worker selects based on the voxel size). I think it’s fair to just say multiple of 20×20×18 for the moment.

Your assumptions concerning the job list view are correct! If there now is a model list view, we could also link there from the training job entry :shrug:

philippotto:
I think, the PR is ready for a first round of review :)

We haven't added a changelog entry, yet. Maybe we should defer this until the features are available for non-super-users?

frontend/javascripts/admin/voxelytics/ai_model_list_view.tsx
```diff
@@ -0,0 +1,151 @@
...
+                                        Annotation {index + 1}
+                                      </a>
+                                    </li>
+                                  ),
+                                )}
+                              </ul>
+                            </div>
+                          ),
+                        });
+                      }}
```
MichaelBuessemeyer, line 128: Nitpicky: Can this be moved into a separat method? It blows up the return statement. 

But feel free to ignore, as it is not so relevant :)

frontend/javascripts/oxalis/model/accessors/volumetracing_accessor.ts
```diff
@@ -778,3 +779,12 @@ export function getMeshInfoForSegment(
...
 }
+
+export function getReadableNameOfVolumeLayer(
+  layer: APIDataLayer,
+  tracing: HybridTracing,
+): string | null {
+  return "tracingId" in layer && layer.tracingId != null
+    ? getReadableNameByVolumeTracingId(tracing, layer.tracingId)
+    : null;
+}
```
MichaelBuessemeyer, line 790: Nice refactoring :tada:

frontend/javascripts/oxalis/view/action-bar/starting_job_modals.tsx
```diff
@@ -760,14 +742,95 @@ export function MitochondriaSegmentationForm() {
...
+    } catch (err) {
+      console.error(err);
+      Toast.error("Could not load model list.");
+    } finally {
+      setIsLoading(false);
+    }
+  }
+  useEffect(() => {
+    loadData();
+  }, []);
```
MichaelBuessemeyer, line 779: Here is another candidate for `useFetch`

frontend/javascripts/oxalis/view/jobs/train_ai_model.tsx
```diff
@@ -0,0 +1,183 @@
...
+              </FormItem>
+            ),
+          },
+        ]}
+        defaultActiveKey={[]}
+      />
+
+      <FormItem hasFeedback name="dummy" label="Training Data">
+        <div>
+          {userBoundingBoxes.length} bounding boxes ({bboxesVoxelCount / 1000000} MVx)
```
MichaelBuessemeyer, line 165: Maybe the `({bboxesVoxelCount / 1000000} MVx)` part can be done without a constant conversion factor? E.g. use the format utils/ implement a rudimentary version for vx like done with `formatBytes`.

frontend/javascripts/admin/voxelytics/ai_model_list_view.tsx
```diff
@@ -0,0 +1,151 @@
...
+import { formatUserName } from "oxalis/model/accessors/user_accessor";
+import { useSelector } from "react-redux";
+import { OxalisState } from "oxalis/store";
+import { JobState } from "admin/job/job_list_view";
+import { Link } from "react-router-dom";
+
+export default function AiModelListView() {
+  const [isLoading, setIsLoading] = useState(false);
+  const [aiModels, setAiModels] = useState<Array<AiModel>>([]);
+  const activeUser = useSelector((state: OxalisState) => state.activeUser);
```
MichaelBuessemeyer, line 18: I think is it nowherer checked whether the active user is a admin / super user. But this feature should only be available to admins, correct?

Thus, this view should also only be visible to superusers.

frontend/javascripts/oxalis/view/jobs/train_ai_model.tsx
```diff
@@ -0,0 +1,183 @@
...
+        rules={[
+          {
+            required: true,
+            message: "Please select a model category.",
+          },
+        ]}
+      >
+        <Select>
+          <Select.Option value="em_neurons">EM Neurons</Select.Option>
+          <Select.Option value="em_nuclei">EM Nuclei</Select.Option>
```
MichaelBuessemeyer, line 108: Can we make this an enum? I mean the two options

MichaelBuessemeyer (changes requested):
Frontend code review done :tada: 



>  Use CRUD routes

Is there a reason why we use CRUD operations now? Because we also have routes that do everything with a single call: In this example a route would take, the dataset, the bounding box for training, the training type, the workflow ... as input to a single route and then start the training / inference from that. Why is this decoupled in this PR now?

app/controllers/AiModelController.scala
```diff
@@ -0,0 +1,224 @@
...
+          "training_annotations" -> Json.toJson(trainingAnnotations),
+          "organization_name" -> organization.name,
+          "model_id" -> modelId,
+          "workflow_yaml" -> request.body.workflowYaml
+        )
+        newTrainingJob <- jobService
+          .submitJob(jobCommand, commandArgs, request.identity, dataStore.name) ?~> "job.couldNotRunTrainModel"
+        existingAiModelsCount <- aiModelDAO.countByNameAndOrganization(request.body.name,
+                                                                       request.identity._organization)
+        _ <- bool2Fox(existingAiModelsCount == 0) ?~> "model.nameInUse"
```
MichaelBuessemeyer, line 144: To me this reads like the job is first submitted and then the it is checked whether there is a name duplication. Shouldn't the name check be before the submision?

app/controllers/AiModelController.scala
```diff
@@ -0,0 +1,224 @@
...
+  }
+
+  def runInference: Action[RunInferenceParameters] =
+    sil.SecuredAction.async(validateJson[RunInferenceParameters]) { implicit request =>
+      for {
+        _ <- userService.assertIsSuperUser(request.identity)
+        organization <- organizationDAO.findOne(request.identity._organization)
+        dataset <- datasetDAO.findOneByNameAndOrganization(request.body.datasetName, organization._id)
+        dataStore <- dataStoreDAO.findOneByName(dataset._dataStore) ?~> "dataStore.notFound"
+        // _ <- aiModelDAO.findOne(request.body.aiModelId) ?~> "aiModel.notFound"
```
MichaelBuessemeyer, line 168: Why is this check currently excluded?

app/controllers/AiModelController.scala
```diff
@@ -0,0 +1,224 @@
...
+        _ <- datasetService.assertValidDatasetName(request.body.newDatasetName)
+        _ <- datasetService.assertNewDatasetName(request.body.newDatasetName, organization._id)
+        jobCommand = JobCommand.infer_with_model
+        boundingBox <- BoundingBox.fromLiteral(request.body.boundingBox).toFox
+        commandArgs = Json.obj(
+          "organization_name" -> organization.name,
+          "dataset_name" -> dataset.name,
+          "color_layer_name" -> request.body.coloLayerName,
```
MichaelBuessemeyer, line 176: typo again :)


```suggestion
          "color_layer_name" -> request.body.colorLayerName,
```

app/controllers/AiModelController.scala
```diff
@@ -0,0 +1,224 @@
...
+          "bounding_box" -> boundingBox.toLiteral,
+          "model_id" -> request.body.aiModelId,
+          "new_segmentation_layer_name" -> request.body.newSegmentationLayerName,
```
MichaelBuessemeyer, line 179: I'd say we should also add a check to ensure `newSegmentationLayerName` follows the allowed characters for dataset layers and no `../../..` something

MichaelBuessemeyer (commented):
The backend code reads like it accepts multiple annotations as training data input for a training job. But the frontend does not seem to support this :see_no_evil: (or I am blind :eyes:)

> We haven't added a changelog entry, yet. Maybe we should defer this until the features are available for non-super-users?

Agreed :+1: 


Just finished looking at the backend part. Please find my comments below.

frontend/javascripts/admin/api/jobs.ts
```diff
@@ -0,0 +1,346 @@
+import Request from "libs/request";
```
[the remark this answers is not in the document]
philippotto: yes :)

frontend/javascripts/admin/api/jobs.ts
```diff
@@ -0,0 +1,346 @@
...
+    mergeSegments: job.commandArgs.merge_segments,
+    trainingAnnotations: job.commandArgs.training_annotations,
+    state: adaptJobState(job.state, job.manualState),
+    manualState: job.manualState,
+    result: job.returnValue,
+    resultLink: job.resultLink,
+    createdAt: job.created,
+    voxelyticsWorkflowHash: job.voxelyticsWorkflowHash,
+  };
+}
```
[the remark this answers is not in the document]
philippotto: the difference to an accessor is that the function is only used internally (in this module) to transform the response by the server immediately. the rest of the frontend is not aware of this transformation. the function is not exported, either. this is similar to `transformTask` and maybe some other occurrences.

frontend/javascripts/admin/api/jobs.ts
```diff
@@ -0,0 +1,346 @@
...
+function adaptJobState(
+  celeryState: APIJobCeleryState,
+  manualState: APIJobManualState,
+): APIJobState {
+  if (manualState) {
+    return manualState;
+  }
+
+  return celeryState || "UNKNOWN";
+}
```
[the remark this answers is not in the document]
philippotto: same argument here :see_no_evil: this is only a transformation on the response of the server and not a utility function to access state.

frontend/javascripts/admin/api/jobs.ts
```diff
@@ -0,0 +1,346 @@
...
+  mag: Vector3;
+};
+
+type RunTrainingParameters = {
+  trainingAnnotations: Array<AiModelTrainingAnnotationSpecification>;
+  name: string;
+  comment?: string;
+  aiModelCategory?: AiModelCategory;
+  workflowYaml?: string;
+};
```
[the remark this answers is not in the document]
philippotto: Since these are not exported currently, I'd like to keep them here. I'm always glad when I see definitions that are local to the module because it's less mental overhead :)

frontend/javascripts/admin/voxelytics/ai_model_list_view.tsx
```diff
@@ -0,0 +1,151 @@
...
+      console.error(err);
+      Toast.error("Could not load model list.");
+    } finally {
+      setIsLoading(false);
+    }
+  }
+
+  useEffect(() => {
+    loadData();
+  }, []);
```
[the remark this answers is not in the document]
philippotto: useFetch doesn't do any error handling and also does not expose an isLoading boolean. for that reason, I extracted a new `useGuardedFetch` utility :)

app/controllers/AiModelController.scala
```diff
@@ -0,0 +1,224 @@
...
+    dataStoreDAO: DataStoreDAO)(implicit ec: ExecutionContext, bodyParsers: PlayBodyParsers)
+    extends Controller
+    with FoxImplicits {
+
+  def readAiModelInfo(aiModelId: String): Action[AnyContent] = sil.SecuredAction.async { implicit request =>
+    {
+      for {
+        _ <- userService.assertIsSuperUser(request.identity)
+        aiModelIdValidated <- ObjectId.fromString(aiModelId)
+        aiModel <- aiModelDAO.findOne(aiModelIdValidated) ?~> "aiModel.notFound" ~> NOT_FOUND
```
MichaelBuessemeyer, line 81: No error message exists for `"aiModel.notFound"` in messages.conf

app/controllers/AiModelController.scala
```diff
@@ -0,0 +1,224 @@
...
+      } yield Ok(jsResult)
+    }
+  }
+
+  def readAiInferenceInfo(aiInferenceId: String): Action[AnyContent] = sil.SecuredAction.async { implicit request =>
+    {
+      for {
+        _ <- userService.assertIsSuperUser(request.identity)
+        aiInferenceIdValidated <- ObjectId.fromString(aiInferenceId)
+        aiInference <- aiInferenceDAO.findOne(aiInferenceIdValidated) ?~> "aiInference.notFound" ~> NOT_FOUND
```
MichaelBuessemeyer, line 92: No error message exists for `"aiInference.notFound"` in messages.conf

app/controllers/AiModelController.scala
```diff
@@ -0,0 +1,224 @@
...
+      } yield Ok(Json.toJson(jsResults))
+    }
+  }
+
+  def runTraining: Action[RunTrainingParameters] = sil.SecuredAction.async(validateJson[RunTrainingParameters]) {
+    implicit request =>
+      for {
+        _ <- userService.assertIsSuperUser(request.identity)
+        trainingAnnotations = request.body.trainingAnnotations
+        _ <- bool2Fox(trainingAnnotations.nonEmpty || request.body.workflowYaml.isDefined) ?~> "aiModel.training.zeroAnnotations"
```
MichaelBuessemeyer, line 124: No error message exists for `"aiModel.training.zeroAnnotations"` in messages.conf

app/controllers/AiModelController.scala
```diff
@@ -0,0 +1,224 @@
...
+          .serialCombined(request.body.trainingAnnotations.map(_.annotationId))(annotationDAO.findOne) ?~> "annotation.notFound"
+        modelId = ObjectId.generate
+        organization <- organizationDAO.findOne(request.identity._organization)
+        jobCommand = JobCommand.train_model
+        commandArgs = Json.obj(
+          "training_annotations" -> Json.toJson(trainingAnnotations),
+          "organization_name" -> organization.name,
+          "model_id" -> modelId,
+          "workflow_yaml" -> request.body.workflowYaml
+        )
```
MichaelBuessemeyer, line 139: Shouldn't this include a `"webknossos_token" & "user_auth_token"`? For me, the run fails because the annotation could not be downloaded by the worker due to a missing token. Here are the command args send to there worker:

```
Starting train_model job run 665586e97d0a009e0e7e59a2__train_model__1716881129997 with params ['model_id: 665586e97d0a00920e7e59a1', 'workflow_yaml: None', 'organization_name: sample_organization', "training_annotations: [{'mag': [1, 1, 1], 'annotationId': '665586357d0a003f0e7e599c', 'colorLayerName': 'color', 'segmentationLayerName': 'segmentation'}]", 'user_auth_token: ***'].
```

app/controllers/AiModelController.scala
```diff
@@ -0,0 +1,224 @@
...
+        _ <- userService.assertIsSuperUser(request.identity)
+        organization <- organizationDAO.findOne(request.identity._organization)
+        dataset <- datasetDAO.findOneByNameAndOrganization(request.body.datasetName, organization._id)
+        dataStore <- dataStoreDAO.findOneByName(dataset._dataStore) ?~> "dataStore.notFound"
+        // _ <- aiModelDAO.findOne(request.body.aiModelId) ?~> "aiModel.notFound"
+        _ <- datasetService.assertValidDatasetName(request.body.newDatasetName)
+        _ <- datasetService.assertNewDatasetName(request.body.newDatasetName, organization._id)
+        jobCommand = JobCommand.infer_with_model
+        boundingBox <- BoundingBox.fromLiteral(request.body.boundingBox).toFox
+        commandArgs = Json.obj(
```
MichaelBuessemeyer, line 173: Just as above, shouldn't this also include the `webknossos_token` and the `user_auth_token`?

MichaelBuessemeyer (commented):
I just found some more things while trying to test:
-> I was unable to execute the trainings job on the worker due to missing tokens in the args in the job passed to the worker.

I also found some messages not defined in `message.conf`

app/controllers/AiModelController.scala
fm3: The user_auth_token is added automatically to every job’s args in `JobService.parameterWrites`. It should also be enough, no jobs should need webknossos_token anymore. Did you pull the latest changes on the worker side?

conf/evolutions/114-ai-models.sql
```diff
@@ -0,0 +1,67 @@
...
+CREATE TABLE webknossos.aiInferences(
+  _id CHAR(24) PRIMARY KEY,
+  _organization CHAR(24) NOT NULL,
+  _aiModel CHAR(24) NOT NULL,
+  _newDataset CHAR(24),
+  _annotation CHAR(24),
+  _inferenceJob CHAR(24) NOT NULL,
+  boundingBox webknossos.BOUNDING_BOX NOT NULL,
+  newSegmentationLayerName VARCHAR(256) NOT NULL,
+  maskAnnotationLayerName VARCHAR(256),
```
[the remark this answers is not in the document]
fm3: That’s right, we decided to not implement this yet in this first version. If it’s ok with you we’ll leave the stubs in, and properly implement this in the next iteration.

frontend/javascripts/admin/api/jobs.ts
```diff
@@ -0,0 +1,346 @@
...
+
+type RunInferenceParameters = {
+  annotationId: string;
+  aiModelId: string;
+  datasetName: string;
+  coloLayerName: string;
+  boundingBox: Vector6;
+  newSegmentationLayerName: string;
+  newDatasetName: string;
+  // maskAnnotationLayerName?: string | null
```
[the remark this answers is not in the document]
fm3: Yes, that’s the idea. As mentioned in https://github.com/scalableminds/webknossos/pull/7682#discussion_r1616798217 we decided to not implement this yet in this first version

app/controllers/WKRemoteWorkerController.scala
```diff
@@ -76,4 +84,35 @@ class WKRemoteWorkerController @Inject()(jobDAO: JobDAO, jobService: JobService,
       } yield Ok
   }
 
+  def attachVoxelyticsWorkflow(key: String, id: String): Action[String] = Action.async(validateJson[String]) {
```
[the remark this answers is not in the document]
fm3: The `workerDAO.findOneByKey` is the permission check, as the key is secret. This way, only workers connected to this webknossos can call this route. Since the worker code is defined by us and will only attach the workflows to its own jobs, this should be enough.

app/controllers/WKRemoteWorkerController.scala
```diff
@@ -76,4 +84,35 @@ class WKRemoteWorkerController @Inject()(jobDAO: JobDAO, jobService: JobService,
...
+        _ <- existingWorkflowBox match {
+          case Full(_)    => Fox.successful(())
+          case Empty      => voxelyticsDAO.upsertWorkflow(workflowHash, "initializing worker workflow", organizationId)
+          case f: Failure => f.toFox
+        }
+        _ <- jobDAO.updateVoxelyticsWorkflow(jobIdParsed, request.body)
+      } yield Ok
+  }
+
+  def attachDatasetToInference(key: String, id: String): Action[String] =
```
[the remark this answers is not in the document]
fm3: same as https://github.com/scalableminds/webknossos/pull/7682#discussion_r1616801654

app/controllers/WKRemoteWorkerController.scala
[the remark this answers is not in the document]
MichaelBuessemeyer: Oh, sorry as I do / did not review the worker code, I didn't see that this route is used only by the worker (and also missed that it is not used by the frontend :see_no_evil:)

Thanks for clarifying :pray:

frontend/javascripts/oxalis/view/jobs/train_ai_model.tsx
```diff
@@ -0,0 +1,183 @@
...
+
+    await runTraining({
+      trainingAnnotations: [
+        {
+          annotationId: tracing.annotationId,
+          colorLayerName: values.imageDataLayer,
+          segmentationLayerName: readableVolumeName,
+          mag: getResolutionInfo(segmentationLayer.resolutions).getFinestResolution(),
+        },
+      ],
```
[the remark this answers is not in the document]
fm3: correct, that’s also for the next iteration.

app/controllers/AiModelController.scala
fm3: good catch, yes!

app/controllers/AiModelController.scala
fm3: that was only for development. un-commenting it now :)

app/controllers/AiModelController.scala
```diff
@@ -0,0 +1,224 @@
...
+          _organization = request.identity._organization,
+          _dataStore = dataStore.name,
+          _user = request.identity._id,
+          _trainingJob = Some(newTrainingJob._id),
+          _trainingAnnotations = trainingAnnotations.map(_.annotationId),
+          name = request.body.name,
+          comment = request.body.comment,
+          category = request.body.aiModelCategory
+        )
+        _ <- aiModelDAO.insertOne(newAiModel)
```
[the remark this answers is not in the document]
fm3: I decided to instead remove the aiModel_trainingAnnotations table from postgres. The training annotation ids are already in the training job. If we later decide that we want to store annotations in a more structured way or also need trainingAnnotations for job-less (e.g. manually imported) models, we can re-add it then.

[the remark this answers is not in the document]
fm3: sorry about the confusion. turns out the aiModelDAO.insertOne does that! It calls `insertTrainingAnnotationIdQueries` for all trainingAnnotations

MichaelBuessemeyer (changes requested):
I found a little buggy in the frontend:

When I try to start an inferral from view mode, the backend fails parsing the sent `annotationId`. In the backend this job parameter is marked as optional, but in case no annotation exists the frontend still sends `annotationId: ""` which leads to the parsing error.
Thus, I'd say that this is a frontend issue. But it should be easily fixable

MichaelBuessemeyer (approved):
> I found a little buggy in the frontend:

I just retested and can confirm the fix :rocket:.

As a side note: The inferral (on the newest branch version of the worker) crashed at step `🚀🚀🚀 Executing segment_affinities@compute_meshes` with error:
```
2024-05-29 09:38:22,852 INFO     6656d7a50a0100e4191c565c__infer_with_model__1716967999738 HTTP Request: GET http://localhost:9000/api/v6/annotations/662f68f9d30000bd13eb2462/download?skipVolumeData=false "HTTP/1.1 400 Bad Request"
2024-05-29 09:38:22,852 WARNING  6656d7a50a0100e4191c565c__infer_with_model__1716967999738 Could not load task from config /home/michael/ScalableMinds/webknossos/binaryData/sample_organization/.aiModels/6656d6011e01006a0b954c89/config_train_model.yaml. Ignoring the file... Caught exception: Failed to instantiate workflow for [PosixPath('/home/michael/ScalableMinds/webknossos/binaryData/sample_organization/.aiModels/6656d6011e01006a0b954c89/config_train_model.yaml')]. See above error for details.
2024-05-29 09:38:22,856 ERROR    6656d7a50a0100e4191c565c__infer_with_model__1716967999738 'compute_meshes'
Traceback (most recent call last):
  File "/home/michael/ScalableMinds/voxelytics/voxelytics/worker/worker_logic/work_polling.py", line 123, in run_job
    return_value = command_fn(run_id, job_id, **job_kwargs)  # type: ignore
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/michael/ScalableMinds/voxelytics/voxelytics/worker/jobs/infer_with_model.py", line 37, in infer_with_model
    output_dataset_name = infer_using_voxelytics_model(
                          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

But this is likely caused by not having trained the actual model used for the inferral on my machine but using the on provided to me by @fm3. => The annotation with id `662f68f9d30000bd13eb2462` (which the worker tries to download) does not exist on my machine :)

So I guess this is fine; But I wanted to mention this anyway :)

PR #7684, opened 2024-03-13
Fix loki

normanrz:
Loki expects all key-value pairs to have string values. It seems like older versions (2.7) didn't validate that strictly, though. The new versions (2.9) do. This PR sprinkles in a few `toString` calls.

### Steps to test:
- Enable voxelytics in config, set a loki uri
- Run a vx workflow against your local webknossos

------


fm3 (approved):
LGTM.

I guess we could also just replace the `.as[Long]` by `.as[String]` in these spots, but no strong opinion.

normanrz:
> I guess we could also just replace the `.as[Long]` by `.as[String]` in these spots, but no strong opinion.

I think that would expect a string in the json, but these are numbers.

fm3:
> I think that would expect a string in the json, but these are numbers.

Yeah that may be right. Feel free to merge as is then :)

PR #7685, opened 2024-03-13
adds support for more omero info keys

### LINKED ISSUE #6570: NGFF Explorer: read omero channel metadata, write layer view configuration
Many OME NGFF datasets have omero metadata, which includes layer names and intensity windowing. This could be turned into layer view configuration and included for the imported dataset.

normanrz:
Adds histogram range, isDisabled and isInverted from the `omero` info in the NGFF metadata.

### Steps to test:
- Import https://uk1s3.embassy.ebi.ac.uk/idr/zarr/v0.4/idr0062A/6001240.zarr
- Check histogram range, isDisabled and isInverted

### Issues:
- fixes #6570 

------


fm3 (approved):
LGTM. 

Did you test it with a dataset that has this info, and also with one that doesn’t have it? (to check if the empty codepath also works)?

PR #7687, opened 2024-03-14
Store voxelytics log not by organization id but organization name

fm3:
- store voxelytics log not by organization id (soon to be dropped) but organization name (soon to be renamed to id)
- query by both (will be removed after one log retention period is over after deploying this)
- also slipped in here: fix deprecation warnings in zarr3testsuite

### Steps to test:
- ?
------


normanrz (approved):
I tested locally :+1:

PR #7690, opened 2024-03-14
Rename some more instances of dataSet to dataset

fm3:
All that remains is now the occurrences in jsons that are also used by the libs. For these we would have to introduce a new api version and an adapter for the old one.
 - Annotation Json
 - Task Json
 - TaskParameters Json
 - LinkedLayerIdentifyer Json

**IMPORTANT: update all connected datastores since the internal API used by the datastores has changed**

- https://renamedatasetsomemore.webknossos.xyz

### Steps to test:
- Test dataset upload + download
- Test task creation + annotating
- Test storage measurement

------


normanrz (approved):
Does this affect the wklibs?

PR #7691, opened 2024-03-14
Remove outdated REST API docs

fm3:
These docs haven’t been maintained for some time, with users being pointed towards the libs client instead. Also, the disclaimer has been in place for some time, I’d say the grace period is over.

I’m not sure how to test if the docs are rendered correctly. Do I have to change more than just remove the file to remove it from the sidebar sitemap?


hotzenklotz (approved):
You also need to update WK docs in the libs repo in tandem with this PR. 

I.e. the REST API is linked in the docs / left navbar. Perhaps in some README files as well. 
https://docs.webknossos.org/webknossos/rest_api.html

Please also ping Albane and let her know in case we reference this on the website. E.g.: 
image

PR #7692, opened 2024-03-14
Use new mesh API for animation job

hotzenklotz:
This PR adds support for passing the required mesh information to use mesh API for animation jobs.

- https://___.webknossos.xyz

### Steps to test:
- abc

### TODOs:
- [ ] AdHoc Meshing
- [ ] Testing with mappings etc

### Issues:
- fixes #

------


docs/animations.md
```diff
@@ -9,7 +9,7 @@ A picture is worth a thousand words. In this spirit, you can use WEBKNOSSOS to c
 Creating an animation is easy:
 
 1. Open any dataset or annotation that you want to use for your animation. 
-2. Optionally, load any [pre-computed 3D meshes](./mesh_visualization.md#pre-computed-mesh-generation) for any segments that you wish to highlight.
+2. Optionally, load some [3D meshes](./mesh_visualization.md#pre-computed-mesh-generation) for any segments that you wish to highlight.
```
philippotto, line 12: should this still link to the precomputed subsection?

frontend/javascripts/types/api_flow_types.ts
```diff
@@ -1085,11 +1086,14 @@ export enum MOVIE_RESOLUTIONS {
...
 export type RenderAnimationOptions = {
-  layerName: string;
-  segmentationLayerName?: string;
-  meshFileName?: string;
-  meshSegmentIds: number[];
+  colorLayerName: string;
+  meshes: ({
+    layerName: string;
+    hasFallbackLayer: boolean;
+    adhoc_mag: Vector3;
```
philippotto, line 1093: switch to camelCase?

frontend/javascripts/oxalis/view/action-bar/create_animation_modal.tsx
```diff
@@ -80,7 +82,10 @@ export function CreateAnimationModalWrapper(props: Props) {
 
   // early stop if no color layer exists
   const colorLayers = getColorLayers(dataset);
-  if (colorLayers.length === 0) return null;
+  if (colorLayers.length === 0) {
+    console.warn("WK can not create animations for datasets without color layers.");
```
philippotto, line 86: How about using `Toast.warning`?

frontend/javascripts/oxalis/view/action-bar/create_animation_modal.tsx
```diff
@@ -171,32 +176,34 @@ function CreateAnimationModal(props: Props) {
...
-      meshFileName = currentMeshFile?.meshFileName;
-
-      if (visibleSegmentationLayer.fallbackLayerInfo) {
-        segmentationLayerName = visibleSegmentationLayer.fallbackLayerInfo.name;
-      } else {
-        segmentationLayerName = visibleSegmentationLayer.name;
-      }
-    }
+    // Submit currently visible pre-computed & ad-hoc meshes
+    const axis = "";
```
philippotto, line 180:
```suggestion
    // Currently, additional coordinates are not supported for this rendering job
    const axis = getAdditionalCoordinatesAsString([]);
```

would be a bit cleaner.

frontend/javascripts/oxalis/view/action-bar/create_animation_modal.tsx
```diff
@@ -171,32 +176,34 @@ function CreateAnimationModal(props: Props) {
...
+        .filter((meshInfo: MeshInformation) => meshInfo.isVisible)
+        .flatMap((meshInfo: MeshInformation) => {
+          const layer = getLayerByName(state.dataset, layerName) as APISegmentationLayer;
+          const hasAFallbackLayer = hasFallbackLayer(layer);
+          const fullLayerName = layer.fallbackLayerInfo?.name || layerName;
+
+          const adhoc_mag_index = getResolutionInfo(layer.resolutions).getClosestExistingIndex(
+            preferredQualityForMeshAdHocComputation,
+          );
+          const adhoc_mag = layer.resolutions[adhoc_mag_index];
```
philippotto, line 197: I think, this can be moved above `Object.values(meshInfos)`, because the code only depends on the layerName, right?

philippotto (changes requested):
didn't test yet, but looks good already :+1: added some minor comments.

docs/animations.md
hotzenklotz: Done

frontend/javascripts/oxalis/view/action-bar/create_animation_modal.tsx
hotzenklotz: Done

frontend/javascripts/oxalis/view/action-bar/create_animation_modal.tsx
hotzenklotz: Done

app/controllers/JobsController.scala
```diff
@@ -380,22 +379,20 @@ class JobsController @Inject()(
...
           command = JobCommand.render_animation
           commandArgs = Json.obj(
             "organization_name" -> organizationName,
             "dataset_name" -> datasetName,
             "export_file_name" -> exportFileName,
             "user_auth_token" -> userAuthToken.id,
-            "layer_name" -> animationJobOptions.layerName,
-            "segmentation_layer_name" -> animationJobOptions.segmentationLayerName,
+            "is_view_mode" -> animationJobOptions.isViewMode,
+            "color_layer_name" -> animationJobOptions.colorLayerName,
```
fm3, line 392: note that if the command args change, the job list view code in the frontend also needs to be adapted (currently the existing layer_name is used to render the job description. compare also getJobs in admin_rest_api

fm3: we also need to ensure backwards compatibility for this job list view rendering (or migrate old jobs in postgres)

hotzenklotz: Good catch. I wasn't aware of this connection... will undo the renaming then.

fm3 (commented):
I tested again and most cases work for me now :tada:

I noticed two more small problems

This message already crops up on pageload when viewing a dataset without color layers. It would be nice if that would happen only once you actually clicked on create animation.

image

And: sometimes `seedAdditionalCoordinates` is not included in the mesh info json (in my test this happened in a proofreading annotation). Since the python code expects this field to be present (although null is a valid option), it fails at that point. Not sure if it should be adapted here or there.

PR #7524, opened 2024-01-08
Refactor Time Tracking API, Revamp TimeTracking Statistics Page

fm3:
Changes the time-tracking overview to show times spent in annotations and tasks and filter them by teams and projects. In the linked detail view, the tracked times can also be filtered by type (annotations or tasks) and project.

### Backend Changes
 - Removed several now-unused routes (previously used either by frontend or braintracing dbs)
 - Added new route `/time/overview` to fill new user time tracking table, filtered by team, projectIds, onlyCountTasks
 - Refactoring
   - No more JodaTime dependency, use java.time utilities instead
   - Renaming of some Methods
   - Moved more Logic to SQL
   - Integrated StatisticsController into TimeController

### Steps to test:
- go to https://timetrackingapi.webknossos.xyz/
- Create several users with different rights, annotations, tasks
- Annotate some, as different users
- Time tracking overview should list all users you are allowed to see (only your orga, only yourself if you’re no team manager, only of your teams if you are team manager)
- Use Filters to select specific teams and projects, as well as switch between tasks and non-task annotations.
- Time tracking detail view should list a single user’s spent time in more detail.

### TODOs:
- [x] Backend
  - [x] merge StatisticsController with TimeController?
  - [x] split statistics routes?
  - [x] make inclusion of explorative optional for /api/time routes AND statistics routes (unify those routes altogether?)
  - [x] Move more logic to SQL
  - [x] No more JodaTime
  - [x] Clean up
  - [x] No more java.util.Calendar
  - [x] CI
  - [x] Document changes
  - [x] support selecting multiple teams in `/time/summed/userList`
  - [x] investigate report of empty response depending on team members
  - [x] fix snapshot tests
  - [x] remove unused routes (`api/time/groupedByInterval`?, `/time/allusers`, `/time/userlist`)
  - [x] make sure every user can see their own timetracking info
- [x] Frontend
  - [x] Adapt to changed api for time graph
  - [x] Remove Best/Busiest users of the week table from `webknossos.org/statistics` -> maybe delete whole view
  - [x] Add new timetracking overview table as discussed in https://scm.slack.com/archives/C5AKLAV0B/p1704359796147499
  - [x] adapt timetracking detail view
    - [x] to filter by projects
    - [x] to handle exploratives (task-related entries can now be null)
  - [x] Rethink page structure (time tracking vs statistics)

#### Charlies todos:
**for time tracking overview**
- [x] add and fill table (sortable) in new page
- [x] add CSV export
[...]

PR #7693, opened 2024-03-14
Remove integration with braintracing databases

fm3:
“braintracing” databases were external tools used by some orgas to track times and wages for annotators. We checked with the respective clients that they are no longer in use.

- Removes automatic registration of users at a configured braintracing database.
- Note that the timetracking routes that were queried only by braintracing will be removed in #7524
- Removes some dead code in user registration flow

### Steps to test:
- Register new user, should run through

------


frcroth (approved):
:fire:

PR #7448, opened 2023-11-21
Added a delete dataset action to dashboard

hotzenklotz:
Every now and again there appear old, unimported datasets on your dashboard, mostly on WK instance with file system access. Thus far there is no easy way to delete them. This PR adds a "Delete" action to this unimported datasets.

### URL of deployed dev instance (used for testing):
- https://___.webknossos.xyz

### Steps to test:
1. Duplicate any DS in your `binaryData` folder and remove `datasource-porps.json` to make it go kaputt
2. Re-scan datasets on your dashboard
3. See and press "Delete" action
4. Profit

### Issues:
- fixes #

------
(Please delete unneeded items, merge only when none are left open)
- [ ] Updated [changelog](../blob/master/CHANGELOG.unreleased.md#unreleased)
- [ ] Updated [migration guide](../blob/master/MIGRATIONS.unreleased.md#unreleased) if applicable
- [ ] Updated [documentation](../blob/master/docs) if applicable
- [ ] Adapted [wk-libs python client](https://github.com/scalableminds/webknossos-libs/tree/master/webknossos/webknossos/client) if relevant API parts change
- [ ] Removed dev-only changes like prints and application.conf edits
- [ ] Considered [common edge cases](../blob/master/.github/common_edge_cases.md)
- [ ] Needs datastore update after deployment

PR #7697, opened 2024-03-18
Remove wkw datasource suggestions

### LINKED ISSUE #7474: Unify exploring zarr and importing wkw
The settings page calls the datastore route `GET /datasets/:organizationName/:dataSetName` (`DataSourceController.suggestDatasourceJson`) which traverses the files and makes automatic suggestions for the datasource-properties.json. This works only for WKW and was originally designed also as an import step (when dataset creators didn’t write their own datasource-properties.jsons).

#7475 skips this step for Zarr and presents a warning instead of a ton of errors (that were there before because the data didn’t match the WKW format).

We should decide if we want to implement this also for other dataformats like zarr. It could reuse the explore code that is used for remote datasets. However, the suggestions might be imperfect there, as the original datasource-properties.jsons might contain more relevant information that can be inferred from just the zarr array files.

We might also want to discontinue this feature altogether, as zarr becomes more common and datasets are usually either explored remotely (uses the explore code), created via the libs (should already have good datasource-properties.json) or uploaded using the web upload (has its own version of this feature as of #7397).

frcroth:
Removes the automatic datasource-properties.json suggestions for the Dataset Settings view. They only worked for WKW and confusing errors popped up for other data formats. Also, since most dataset creation code already generates valid jsons, it was mostly unused.

At some point in the future we may want to use the explore code to bring them back, if that turns out to be a comon use case. (We’d probably have to add an Explorer for WKW).

### TODOs:
- [x] Backend
  - [x] Remove exploring route
  - [x] remove unused code
  - [x] Adapt readInboxDatasource route to support reporting the most current version from disk
- [x] Frontend
  - [x] Use readInboxDataSource in dataset settings view (note that it may contain `existingDataSourceProperties` in case of Unusable json)
  - [x] Decide what to show exactly in case of invalid json (both invalid fields and json parse error) -> an error :)
  - [x] In dashboard, always call the link “Settings”, no longer “Import”
  - [x] What should the frontend to in case the datasource-properties.json is not valid? / cannot be parsed?

### Issues:
- fixes #7474 

------


MichaelBuessemeyer:
This is how the settings view looks like, in case the datasource-properties.json is not a valid json file:

image

And like this, in case the "name" property is missing in the "id" field
image

I'd say it is okaisch, or do we want to improve the suggestion?

philippotto:
> I'd say it is okaisch

Yes, since manually edited jsons are rare, it's okay in my opinion.

frontend/javascripts/dashboard/advanced_dataset/dataset_action_view.tsx
```diff
@@ -330,7 +236,7 @@ export function getDatasetActionContextMenu({
             },
           }
         : null,
-      dataset.isEditable && dataset.isActive
```
MichaelBuessemeyer, line 333: Am I correct, that we no longer have a "not imported" state for a dataset? In that case, is it possible for a dataset to not be active? The only option would be to delete a dataset :thinking: 

My thought is that we may not need this field `isActive` anymore / do not need to consider it any longer in the frontend :thinking:

frontend/javascripts/dashboard/dataset/dataset_settings_view.tsx
```diff
@@ -243,68 +151,12 @@ class DatasetSettingsView extends React.PureComponent<PropsWithFormAndRouter, St
...
-        savedDataSourceOnServer != null && savedDataSourceOnServer.status == null;
+      const dataSource = await readDatasetDatasource(dataset);
 
       // Ensure that zarr layers (which aren't inferred by the back-end) are still
       // included in the inferred data source
-      if (
-        savedDataSourceOnServer &&
-        "dataLayers" in savedDataSourceOnServer &&
-        inferredDataSource != null
-      ) {
```
MichaelBuessemeyer, line 266: This cannot be done as there is no longer a `inferredDataSource`. So does this maybe result in errors with zarr layers?

frontend/javascripts/dashboard/advanced_dataset/dataset_action_view.tsx
philippotto: > Am I correct, that we no longer have a "not imported" state for a dataset? In that case, is it possible for a dataset to not be active?

I don't know what happens now if a datasource properties json does not exist (or is incomplete). I assume the dataset won't be active then. However, the front-end should still show the edit button. Otherwise, the dataset would not be usable/fixable as you say.

frontend/javascripts/dashboard/dataset/dataset_settings_view.tsx
```diff
@@ -243,68 +151,12 @@ class DatasetSettingsView extends React.PureComponent<PropsWithFormAndRouter, St
...
-        savedDataSourceOnServer,
-        inferredDataSource,
-      );
-      this.setState({
-        savedDataSourceOnServer,
-        inferredDataSource: inferredDataSourceWithCorrectedScale,
-        messages: dataSourceMessages,
-        dataSourceSettingsStatus,
-      });
+      // TODOM: handle case in which dataSource is null / could not be parsed by the server
```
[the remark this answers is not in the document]
MichaelBuessemeyer: please see my post below :)

MichaelBuessemeyer:
This is how it looks if there is no `datasource-properties.json` available in the backend in the dataset list view on the newest version of the branch:
image
And this in the settings view:
image


In case the `datasource-properties.json` contains errors the dataset table looks the same as above.
And the settings look like this:
image


> I'm not sure whether it's a good idea to delete this. The use case is:
 a dataset exists on disk, but wk isn't able to read it properly (e.g., because of an invalid json)

This should be handled now: The dataset table always gives the option to a user to edit the settings (as long as the user has the capabilities (`isEditable`)).

> in that case, as a user I want to be able to see the issue, be able to fix it or to delete the dataset directly from the table.

The settings view gives a warning about not being able to parse the `datasource-properties.json`  (although the warning is not easy to understand). And the settings are accessible from the dataset table.
I do not understand why one would want to delete a dataset directly from the dataset table view. This is currently not possible but I think I also misunderstand your part 
> to delete the dataset directly from the table.

here a little.

philippotto:
> I do not understand why one would want to delete a dataset directly from the dataset table view. This is currently not possible but I think I also misunderstand your part [...] here a little.

This feature was added not long ago. See [here](https://github.com/scalableminds/webknossos/pull/7448) for context. I think, it would be good to keep it :)

frontend/javascripts/dashboard/advanced_dataset/dataset_action_view.tsx
```diff
@@ -197,48 +143,8 @@ function DatasetActionView(props: Props) {
...
-        }
-      >
-        <WarningOutlined className="icon-margin-right" />
-        Show Error
-      </a>
-      {dataset.status !== "Deleted by user." ? (
-        <a onClick={() => onDeleteDataset()}>
-          <DeleteOutlined className="icon-margin-right" />
-          Delete Dataset
-        </a>
```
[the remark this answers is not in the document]
fm3: I agree, it should still be possible to fix the json, and parsing errors should be reported. Just not the automated suggestions.

MichaelBuessemeyer:
I'd say I applied all requested cases. How to we want to proceed with this @fm3? 

Maybe we can check together my changes requested by philipp and then merge this once its clear to go?

fm3:
Do you think that letting this PR lie will create many merge conflicts? If not, I’d say it’s fair to wait until Philipp returns.

philippotto (approved):
sweet, red lines :) front-end lgtm :+1:

PR #7701, opened 2024-03-19
Improve organization switching for multi-organization users

### LINKED ISSUE #7705: Speed up organization list request
For users who are in many organizations, the organization list request that happens on pageload can be pretty slow. We should either add a compact parameter that makes it return only id+name, or speed up the existing request with a complex query.

philippotto:
- Show search box for organizations in switcher when more than 10 orgas are listed
- render at most 30 orgas in the switcher to improve performance
- move scm orga to the front when it exists (this is rather a hack for our team)

image

- https://___.webknossos.xyz

### Steps to test:
- use the switcher (currently, 300 dummy organizations are rendered)

### TODOs:
- [x] remove the 300 dummy organizations

### Issues:
- fixes https://scm.slack.com/archives/C5AKLAV0B/p1710837272371279
- fixes #7705

------


fm3:
Nice! Two more little wishes if they are not complex to achieve:
 - when i click the search field, the menu closes
 - as the height of the list changes due to the search, the browser scrollbar appears and disappears again, which is a bit disorienting. Could we limit the height to something smaller so that it ends before the bottom of the screen?
image

philippotto:
> * when i click the search field, the menu closes

unfortunately, i didn't find a way to fix this. since the focus is there by default, it shouldn't be a big usability problem for us, though.

> * as the height of the list changes due to the search, the browser scrollbar appears and disappears again, which is a bit disorienting. Could we limit the height to something smaller so that it ends before the bottom of the screen?

I pushed a fix for this which limits the list's height to 75% of the window height. I hope this is a good heuristic to avoid the problem you described.

frontend/stylesheets/main.less
```diff
@@ -647,3 +647,17 @@ button.narrow {
...
   display: inherit;
 }
+
+.segments-stats-info-alert {
+  margin-bottom: 10px;
+  margin-top: -14px; // wanted margin - modal padding = 10-24
+}
+
+.organization-switch-menu > ul {
+  max-height: 75vh;
```
normanrz, line 657: This rule is not specific enough to be applied.
```suggestion
  max-height: 75vh !important;
```

Maybe you could also use `calc` like this?
```suggestion
  max-height: calc(100vh - var(--navbar-height) - 200px) !important;
```

normanrz (commented):
Nice!
I noticed that arrow key shortcuts don't work in the search box. No big deal, but would be nice if easily fixed.

normanrz (approved):

philippotto:
> Nice! I noticed that arrow key shortcuts don't work in the search box. No big deal, but would be nice if easily fixed.

Done!

PR #7706, opened 2024-03-19
Speed up JSON mappings

daniel-wer:
Also, no longer maintain the extra mappingKeys array which was only needed, because Object.keys was very slow for large objects. I benchmarked the old code, the new code, and the new code maintaining the extra mappingKeys array (all with deepFreeze disabled!) using a large JSON mapping with 1.5M entries:

Object mapping with extra keys array (old):
    Mapping creation: ~4600 ms
    Mapping texture creation: ~60 ms

Map mapping with extra keys array (intermediate):
    Mapping creation: ~1160 ms
    Mapping texture creation: ~460 ms

Map mapping (new):
    Mapping creation: ~590 ms
    Mapping texture creation: ~1070 ms
    
Overall a nice 3x speedup for larger JSON mappings. However, I mainly did this in preparation for https://github.com/scalableminds/webknossos/pull/7654 where this frontend mapping functionality will be used extensively.

- https://___.webknossos.xyz

### Steps to test:
- Open a dataset with JSON mappings (e.g. ROI2017_wkw) and check that they still work (and are faster than before)
- Test that merger mode still works as expected

### Issues:
- contributes to https://github.com/scalableminds/webknossos/pull/7654

------

### DIFF
--- a/frontend/javascripts/oxalis/api/api_v2.ts
+++ b/frontend/javascripts/oxalis/api/api_v2.ts
@@ -565,7 +565,6 @@ class DataApi {
 
     const mappingProperties = {
       mapping: _.clone(mapping),
-      mappingKeys: Object.keys(mapping).map((x) => parseInt(x, 10)),
     };
     Store.dispatch(setMappingAction(layerName, "<custom mapping>", "JSON", mappingProperties));
   }
--- a/frontend/javascripts/oxalis/api/api_v2.ts
+++ b/frontend/javascripts/oxalis/api/api_v2.ts
@@ -555,7 +555,7 @@ class DataApi {
    *
    * api.setMapping("segmentation", mapping);
    */
-  setMapping(layerName: string, mapping: Mapping) {
+  setMapping(layerName: string, mapping: Mapping | Record<number, number>) {
     const segmentationLayer = this.model.getLayerByName(layerName);
     const segmentationLayerName = segmentationLayer != null ? segmentationLayer.name : null;
 
@@ -564,7 +564,10 @@ class DataApi {
     }
 
     const mappingProperties = {
-      mapping: _.clone(mapping),
+      mapping:
+        mapping instanceof Map
+          ? new Map(mapping)
+          : new Map(Object.entries(mapping).map(([key, value]) => [parseInt(key, 10), value])),
     };
     Store.dispatch(setMappingAction(layerName, "<custom mapping>", "JSON", mappingProperties));
   }
--- a/frontend/javascripts/oxalis/model/sagas/mapping_saga.ts
+++ b/frontend/javascripts/oxalis/model/sagas/mapping_saga.ts
@@ -229,14 +229,9 @@ function* handleSetMapping(
   const fetchedMapping = fetchedMappings[mappingName];
   const { hideUnmappedIds, colors: mappingColors } = fetchedMapping;
 
-  const [mappingObject, mappingKeys] = yield* call(
-    buildMappingObject,
-    mappingName,
-    fetchedMappings,
-  );
+  const mapping = yield* call(buildMappingObject, mappingName, fetchedMappings);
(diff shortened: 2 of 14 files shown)


frontend/javascripts/oxalis/api/api_v2.ts
```diff
@@ -555,7 +555,7 @@ class DataApi {
    *
    * api.setMapping("segmentation", mapping);
    */
-  setMapping(layerName: string, mapping: Mapping) {
+  setMapping(layerName: string, mapping: Mapping | Record<number, number>) {
```
philippotto, line 558: we call this code internally, right? (e.g., in merger mode) then, it might make sense to avoid the copying in lines 567f somehow :thinking: what do you think?

frontend/javascripts/oxalis/model/sagas/mapping_saga.ts
```diff
@@ -319,14 +314,8 @@ function* fetchMappings(
...
 }
 
-function buildMappingObject(
-  mappingName: string,
-  fetchedMappings: APIMappings,
-): [Mapping, Array<number>] {
-  const mappingObject: Mapping = {};
-  // Performance optimization: Object.keys(...) is slow for large objects
-  // keeping track of the keys in a separate array is ~5x faster
-  const mappingKeys = [];
```
philippotto, line 329: Did you try to preallocate the array with the correct size here? I did a small microbenchmark and it seems to be about 5x faster to fill the array then:

```js
n = 1500000
console.time("A")
let arr = [];
for (let i = 0; i < n; i++) {
	arr.push(i)
}
console.timeEnd("A")

console.time("B")
arr = new Array(n);
for (let i = 0; i < n; i++) {
	arr[i] = i;
}
console.timeEnd("B")

output:
A: 25.44775390625 ms
B: 5.261962890625 ms
```

not sure whether this would make it worth again to keep the keys as an extra array around?

philippotto: oh, I'm not sure whether the total amount is known apriori, at all? :see_no_evil:

philippotto (commented):
awesome speed up :man_dancing: I left a couple of thoughts, but probably the code can stay as is. let me know if this is the case, then I'll test.

frontend/javascripts/oxalis/model/sagas/mapping_saga.ts
daniel-wer: Thanks for testing! It would be possible to find out the number of keys beforehand (with an extra iteration), but I don't think the time spent building the keys array is substantial enough to warrant the extra iteration and keeping the extra array around

philippotto: ok, makes sense!

frontend/javascripts/oxalis/api/api_v2.ts
philippotto: I also see think that it would be way cleaner to avoid mutating the mapping that is already active in the store. At the same time, I'm wondering whether there would be a visible bug, because the mutated mapping is set to the store afterwards, anyway. But obviously we should only do such things, when the performance win is interesting to us. For the merger mode, this is not the case, I think. Not sure whether similar thoughts can be had for your front-end-mapping PR.. However, if you feel like performance is good enough now, let's stick to that :)

philippotto (approved):
works well :100:

frontend/javascripts/oxalis/api/api_v2.ts
daniel-wer: In the frontend mapping PR, the api functionality is not used. Instead the setMappingAction is dispatched directly. I guess one could do the same in the merger mode too, but as you said, it doesn't seem performance critical for typical merger mode scenarios.

PR #7709, opened 2024-03-20
Improve task list

### LINKED ISSUE #7586: Improvements task list
## Detailed Description
- The project list should be sorted in reverse chronological order (ie. newest first). Currently, it is the other way round. That is unfortunate, because the newest project is most often the most interesting.
- Since the detail row for an task instance is so high, the "View" button could be promoted out of the "Actions" overflow menu. That would save a few clicks when interacting with tasks.
- When going through the contents of tasks, one needs to open all detail rows manually. Maybe there could be a button to open them all at once?

Screenshot 2024-01-24 at 16 23 26

dieknolle3333:
### Steps to test:
- Go to task list and search to get some results.
- Make sure tasks are sorted by project date, but from latest to earliest.
- Test "expand all" button
- Check that expanded rows look nice 

### Issues:
- fixes #7586 

------


frontend/javascripts/admin/task/task_annotation_view.tsx
```diff
@@ -192,18 +197,19 @@ class TaskAnnotationView extends React.PureComponent<Props, State> {
     return (
       <div>
         <table>
-          <tbody>
+          <tbody className="task-annotation-view">
```
philippotto, line 200: this class is now assigned to tbody as well as td, which I find a bit unusual. also, why did you add it in the first place? I don't see it mentioned in the original issue nor in the pr description :thinking:

dieknolle3333: right, I forgot to remove it from tbody, thank you!
I added the class because the alignment of the expanded rows differed from row to row and I thought it would be nice to make it consistent between all rows. If there was a bigger reason behind this or it was plain unnecessary I can remove it of course! Didnt mention it because I though its a minor change.

philippotto: > right, I forgot to remove it from tbody, thank you!

ah, ok, then it makes sense :+1: 

>  If there was a bigger reason behind this or it was plain unnecessary I can remove it of course! Didnt mention it explicitly because I thought its a minor change.

yes, that's fine! I only wanted to explain that I didn't get the change initially :)

PR #7710, opened 2024-03-21
Make Layer Name Validation in submitJob less strict

fm3:
The protection against illegal layer names being passed to jobs (for avoiding path injection) was so strict that many existing layer names could not be used.

This can be re-added as part of https://github.com/scalableminds/webknossos/issues/7711 when we actually migrate existing layers.

But for now, let’s just be less strict about the submitted layer names in the job params.


MichaelBuessemeyer (approved):
Thanks for noticing that the previously introduced validation is too strict for now.

The code looks good to me :+1:. Should I also run manual tests locally or do you think this should be fine?

PR #7721, opened 2024-03-26
Fix that teamless users were not listed

fm3:
Fixes regression from #7646 – I hadn’t been aware that users without teams were even [allowed](https://github.com/scalableminds/webknossos/pull/2801)… I think this double left join here now should not cause new fanout duplicates since there is only one team per user_team_role

- https://listteamlessusers.webknossos.xyz

### Steps to test:
- Remove all team memberships from a user
- They should still be listed in api/users

------


fm3:
@normanrz maybe you could have a look at this? :)

normanrz (approved):

PR #7722, opened 2024-03-26
fix bug that user in url was not selected

dieknolle3333:
Bug: Clicked user in overview was not preselected in detail view. The reason seems to be that the data is fetched while the state is not yet updated.

### Steps to test:
- Go to time tracking overview and click details for a specific user
- make sure this user is selected in the detail view, in the dropdown 

### Issues:
- follow up for #7524 

------


philippotto (approved):
looks good to me!

PR #7723, opened 2024-03-26
Make color layers transparent if color value is out of range

MichaelBuessemeyer:
This PR lets cover layers "shine / see through" for each voxel where the color value is out of the range configured as the histogram min and max bounds. This setting can be adjusted by the user and therefore this feature is independent of the exact color layer value range.

- https://___.webknossos.xyz

### Steps to test:
- Open a dataset with multiple color layers. Preferably one with affine transformations.
- Turn on cover blend mode in the settings tab.
- In the layer settings adjust the min & max settings of the histogram (can be turned on via the menu in the top right corner of the histogram view)
- The lower color layers should now shine through the upper color layer

### Issues:
- No issue exists for this.

------

### DIFF
--- a/frontend/javascripts/oxalis/geometries/materials/plane_material_factory.ts
+++ b/frontend/javascripts/oxalis/geometries/materials/plane_material_factory.ts
@@ -264,8 +264,15 @@ class PlaneMaterialFactory {
       this.uniforms[`${name}_color`] = {
         value: DEFAULT_COLOR,
       };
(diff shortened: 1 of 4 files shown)


MichaelBuessemeyer:
Closing as not enough datasets needing the cover mode exist to justify this feature. The related slack thread can be found here: https://scm.slack.com/archives/C5AKLAV0B/p1710950020861239

PR #7728, opened 2024-03-27
Fix loading of the first LOD mesh

daniel-wer:
The LOD loading order for the first loaded mesh always defaulted to 0, 1, 2, ... even if the actual LOD was 2 for example. This led to the mesh repeatedly disappearing: First LOD 0 was loaded, the mesh disappeared, LOD 1 was loaded, the mesh disappeared, LOD 2 was loaded.

- https://fixfirstlodmeshloading.webknossos.xyz

### Steps to test:
- @philippotto Do you happen to have a mesh file with LODs available for testing?

------


philippotto:
> * @philippotto Do you happen to have a mesh file with LODs available for testing?

Probably not :thinking: Would probably be good to have one on the dev instances, right? Can you add one?

philippotto:
@daniel-wer I see that you prepared a dev instance now. This means I can test there? I'm currently on spotty wifi, but already wanted to clarify this :)

daniel-wer:
> @daniel-wer I see that you prepared a dev instance now. This means I can test there? I'm currently on spotty wifi, but already wanted to clarify this :)

Yes you can test now. During my quick test, I wasn't able to reproduce the original issue (the flickering) on master, although the lod order was wrong (which I confirmed using the debugger) :thinking: Then, when testing this branch, the lod order was correct, but I sometimes saw some flickering :see_no_evil: I'll try to do another test today. If you test as well, not again that only the first loaded mesh is affected by the original bug.

philippotto (approved):
I cannot reproduce the original problem, either, but maybe the problem only appears for special mesh files (with more than 3 lods for example). The code itself looks good and seems to work for the dev instance. So, let's merge it :)

PR #7729, opened 2024-03-27
After merge, remove segment item of id that doesn't exist anymore

### LINKED ISSUE #7315: [Proof Reading] Remove segment from list after merge
## Context
After merging two segments with the proof reading tool there should only be one remaining entry in the "Segments Tab" after the operation. Instead the the merged segment's list entry remains as a Karteileiche. Any operation on it will crash WK.

image

## Expected Behavior
There should only be one list entry. 

## Current Behavior
There are two list entries. Right-click removing the "remainder" entry will crash the segment's tab.

## Steps to Reproduce the bug

1. Click on each segment individually with proof reading tool
2. Merge two segments 
3. Now you have two segments

## Your Environment for bug

- Browser name and version: e.g. Chrome 116
- Operating System and version: e.g. MacOS
- Version of WEBKNOSSOS (Release or Commit): 24599

philippotto:
- When merging two segments, removes the segment from the segment list.
- The active segment/item now always survives the merge operation (wasn't the case when merging agglomerate-skeleton or skeletons with a shortcut).

- https://removesegmentaftermerge.webknossos.xyz

### Steps to test:
Part 1:
- create a new annotation for a dataset that has an agglomerate mapping (e.g., test-agglomerate)
- activate proofreading mode
- click on two segments that should be merged so that they have an entry each in the segments tab
- merge them with the context menu
- the segment id that doesn't exist anymore in the data should also not exist anymore in the segments tab
- repeat the above by loading agglomerate skeletons and merging them (make one node active and rightclick another node -> create edge)
Part 2:
- go into skeleton mode
- create two trees
- merge them with the ctrl+shift+alt shortcut --> active tree should survive
- merge them with the context menu --> active tree should survive

### Issues:
- fixes #7315

------


daniel-wer (approved):
Nice one, LGTM :+1:

Issue #7725, opened 2024-03-27
Occasionally, invisible nodes can be selected


philippotto:
## Context
Skeleton nodes are only rendered when they are within the configured clipping distance. [Further context in this GH discussion](https://github.com/scalableminds/webknossos/pull/7678#discussion_r1521735546).

## Expected Behavior
Only nodes that are visible should be selectable by the user.

## Current Behavior
The clipping distance is not perfectly respected which is why you can sometimes select nodes that are not rendered.

## Steps to Reproduce the bug

- [ ] Cannot reproduce the bug anymore / needs deeper investigation.

1. create a node
2. move in z by a few slices
3. reduce the clipping distance so that it is not visible anymore
4. click at the position where the node is in XY
5. -> wk jumps to the clicked node even though it was invisible

## Your Environment for bug

- Browser name and version: e.g. Chrome 39
- Operating System and version: e.g. Windows 10
- Version of WEBKNOSSOS (Release or Commit):
- [ ] Specific to long-running jobs (set `jobsEnabled=true` in `application.conf`)
- [ ] Specific to webknossos.org (set `isDemoInstance=true` in `application.conf`)

daniel-wer:
Not sure if this is related, but I noticed that the clipping in oblique and flight mode changes when zooming. If the clipping distance were in nm (which it is supposed to be) this shouldn't happen.

philippotto:
@daniel-wer In #7732 you wrote that that PR contributes to this issue. What parts are still missing?

daniel-wer:
I wrote this, because my fix didn't explain this observation

> Not sure if this is related, but I noticed that the clipping in oblique and flight mode changes when zooming. If the clipping distance were in nm (which it is supposed to be) this shouldn't happen.

However, I did a quick test again and although the clipping changed during zooming I wasn't able to select invisible nodes, so I guess this particular issue (mismatch between what's displayed and the node picking) is resolved. What remains to be investigated is why the clipping distance in oblique/flight mode changes during zooming and doesn't seem to be in nm.

philippotto:
Great, thanks for the clarification!

hotzenklotz:
We don't support the arbitrary mode anymore. Can this be closed?

daniel-wer:
We still support flight mode, so I think this cannot be closed.

PR #7732, opened 2024-04-03
Fix node picking regression

daniel-wer:
The regression allowed to select nodes in orthogonal mode that were not visible.
Also, I fixed the far clipping distance for flight/oblique mode which allowed to select nodes that were not visible. In contrast to orthogonal mode, nodes that are clipping-distance-behind the plane cannot be seen in flight/oblique mode which needed to be done similarly during node picking.

- https://fixnodepickingregression.webknossos.xyz

### Steps to test:
- Set some nodes and move forward/backward until they cannot be seen any longer. You should not be able to select them. When moving so you can just about see them, you should be able to select them. Test in orthogonal and flight mode.

### Issues:
- fixes internal bug report
- contributes to https://github.com/scalableminds/webknossos/issues/7725

------

### DIFF
--- a/frontend/javascripts/oxalis/view/rendering_utils.ts
+++ b/frontend/javascripts/oxalis/view/rendering_utils.ts
@@ -1,7 +1,7 @@
 import * as THREE from "three";
 import { saveAs } from "file-saver";
 import Store from "oxalis/store";
-import { OrthoView } from "oxalis/constants";
+import { ARBITRARY_CAM_DISTANCE, OrthoView } from "oxalis/constants";
 import constants, {
   ArbitraryViewport,
   OrthoViewColors,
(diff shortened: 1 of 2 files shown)


frontend/javascripts/oxalis/view/rendering_utils.ts
```diff
@@ -59,25 +59,28 @@ export function renderToTexture(
...
+        // achieved by the data plane which is not rendered during node picking
+        camera.far = ARBITRARY_CAM_DISTANCE;
       } else {
-        // The near value is already set in the camera (done in the CameraController).
         // The far value has to be set, since in normal rendering the far clipping is
         // achieved by offsetting the plane instead of setting the far property.
         camera.far = state.userConfiguration.clippingDistance;
       }
       camera.updateProjectionMatrix();
+      return camera;
```
daniel-wer, line 80: This is the fix for the actual regression. The camera was cloned and modified (in this function that was introduced in a recent PR), but then later in the original function, the old camera was used again.

normanrz (approved):

frontend/javascripts/oxalis/view/rendering_utils.ts
philippotto: argh, this explains a lot of the weirdness I experienced with the clipping distances :see_no_evil: thanks for fixing this!

PR #7736, opened 2024-04-04
Add text in case no data is shown in timetracking overview

dieknolle3333:
### Steps to test:
- Go to time tracking overview and select a range/filters without any tracked times.
- Make sure an appropriate text is displayed, rather than "no data"

------


hotzenklotz (approved):
Thank you! 🚢

PR #7737, opened 2024-04-08
Improve error message for largest segment id that exceeds dtype range

fm3:
avoiding bare error message


hotzenklotz (approved):
THank you

PR #7739, opened 2024-04-08
Adapt screenshot tests to new meshfile

fm3:
The new meshfile on the dev instance is computed for the oversegmentation, so I now request a different segment id so that we can see a mesh in the screenshot (green now instead of yellow).

I am not certain why the other two meshes have also changed. They are ad-hoc rendered, and the default is to use mag 4-4-2 there, which is visible in the new screenshot. Not sure how why old screenshot seems to have run marchingCubes on mag1.


philippotto:
> I am not certain why the other two meshes have also changed. They are ad-hoc rendered, and the default is to use mag 4-4-2 there, which is visible in the new screenshot. Not sure how why old screenshot seems to have run marchingCubes on mag1.

My initial assumption was that all three meshes are incorrectly loaded from the precomputed meshfile. However, this hypothesis didn't turn out to be true. I assume it's not worth it to investigate deeper..

philippotto (approved):
Thank you for fixing this!

PR #7741, opened 2024-04-09
Don't let toast elapse in background in some browsers

### LINKED ISSUE #4719: Ensure toast timeout doesn't elapse when tab is in background
NML Upload for large NMLs can take several minutes. Users leave the tab and when they return, the error toast is gone and they don’t get to know what was wrong with their NML or the server.

Maybe we should do this only for large NMLs? Any opinions?

knollengewaechs:
- https://timeouttabnotinbackground.webknossos.xyz/

### Steps to test:
- Use Chrome or sth. Chromium based
- Do anything that takes a bit of time and where toasts are shown in the process, such as:
      - upload a larger dataset and switch the tab
      - wait until the upload is done for a while
      - go back to the WK tab and make sure both toasts are shown at least for a very short time (there should be at least two toasts after switching back)
- create scenarios where error toasts and warning toasts are shown. Make sure that their content is printed to the console, except if the toast content is a more complex react node (containing a link, for example).

### Issues:
- fixes #4719

------

### DIFF
--- a/CHANGELOG.unreleased.md
+++ b/CHANGELOG.unreleased.md
@@ -13,6 +13,7 @@ For upgrade instructions, please check the [migration guide](MIGRATIONS.released
 ### Added
 - Creating and deleting edges is now possible with ctrl+(alt/shift)+leftclick in orthogonal, flight and oblique mode. Also, the flight and oblique modes allow selecting nodes with leftclick, creating new trees with 'c' and deleting the active node with 'del'. [#7678](https://github.com/scalableminds/webknossos/pull/7678)
 - Added Typescript defintions for @scalableminds/prop-types package. [#7744](https://github.com/scalableminds/webknossos/pull/7744)
+- Time Tracking now also works when editing other users’ shared annotations, and when editing proofreading annotations (a.k.a. editable mappings). [#7749](https://github.com/scalableminds/webknossos/pull/7749)
 
 ### Changed
 - Improved task list to sort tasks by project date, add option to expand all tasks at once and improve styling. [#7709](https://github.com/scalableminds/webknossos/pull/7709)
--- a/app/controllers/AnnotationController.scala
+++ b/app/controllers/AnnotationController.scala
@@ -106,7 +106,7 @@ class AnnotationController @Inject()(
         _ <- Fox.runOptional(request.identity) { user =>
           if (typedTyp == AnnotationType.Task || typedTyp == AnnotationType.Explorational) {
             timeSpanService
-              .logUserInteraction(Instant(timestamp), user, annotation) // log time when a user starts working
+              .logUserInteractionIfTheyArePotentialContributor(Instant(timestamp), user, annotation) // log time when a user starts working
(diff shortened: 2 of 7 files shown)


frontend/javascripts/oxalis/model_initialization.ts
```diff
@@ -700,11 +700,11 @@ async function applyLayerState(stateByLayer: UrlStateByLayer) {
       // The name of the layer could have changed if a volume tracing was created from a viewed annotation
       effectiveLayerName = getLayerByName(dataset, layerName, true).name;
     } catch (e) {
-      console.error(e);
```
dieknolle3333, line 703: In cases like this I didnt omit the separate log because Error not only has a message, but also a type, and I didnt want that to be omitted. Let me know if you think I should change that :)

frontend/javascripts/libs/toast.tsx
```diff
@@ -112,13 +112,29 @@ const Toast = {
     }
 
     notification[type](toastConfig);
+
+    // Make sure that toasts don't just dissapear while user has WK in background tab (e.g. while uploading large dataset).
```
philippotto, line 116:
```suggestion
    // Make sure that toasts don't just disappear while user has WK in background tab (e.g. while uploading large dataset).
```

frontend/javascripts/libs/toast.tsx
```diff
@@ -112,13 +112,29 @@ const Toast = {
...
+    // Most browsers pause requestAnimationFrame() if the current tab is not active, but Firefox does not seem to do that.
+    if (!sticky && key != null) {
+      requestAnimationFrame(async () => {
+        // ensure tab is active
+        const timeoutAfterTabSwitch = timeout >= 1000 ? 1000 : 0;
+        await sleep(timeout - timeoutAfterTabSwitch);
```
philippotto, line 122: how about:

```suggestion
        const timeoutAfterTabSwitch = timeout / 2; // todo: maybe another var name would make more sense now
        await sleep(timeoutAfterTabSwitch);
```

this effectively splits the sleep into two equally long phases which can only start when the animation frame is fulfilled each. I find that conceptionally more intuitive.

frontend/javascripts/libs/toast.tsx
```diff
@@ -112,13 +112,29 @@ const Toast = {
     }
 
     notification[type](toastConfig);
+
+    // Make sure that toasts don't just dissapear while user has WK in background tab (e.g. while uploading large dataset).
+    // Most browsers pause requestAnimationFrame() if the current tab is not active, but Firefox does not seem to do that.
+    if (!sticky && key != null) {
+      requestAnimationFrame(async () => {
```
philippotto, line 119: You can use `await animationFrame();` to avoid the callback style. `animationFrame` is a helper in libs/utils.ts.

philippotto (commented):
nice :) I left two small suggestions.

philippotto (approved):
sweet :)

PR #7743, opened 2024-04-10
Remove left over ESLint artifacts

hotzenklotz:
With the recent switch to Biome for linting (#7665) we removed eslint. Some plugins and config files were left behin. This PR removes them.

Replaced circular dependency detection tool from [`madge`](https://github.com/pahen/madge) to [`dpdm`](https://github.com/acrazing/dpdm) since I ran into dependency problems. Newer versions of `madge` would detect 100+ circular imports, mostly due to type imports which were not reliably ignored. (see this [Madge issue](https://github.com/pahen/madge/issues/232)

### Steps to test:
- None. CI is enough

### Issues:
- Related to #7665

------


hotzenklotz:
> Great 👍 While you are at it, you could also remove a bunch of `// eslint-disable` comments that still linger around..

Ok, done. See commit:  https://github.com/scalableminds/webknossos/pull/7743/commits/8b79f0747d3f7752d9a959156fdee0017420fa03

PR #7747, opened 2024-04-10
Fix position of context menu in layer settings for layers with long names

dieknolle3333:
### Steps to test:
- Open annotation, create new layer with a very long name
- open the context menu for said layer

### Issues:
- fixes this: https://scm.slack.com/archives/C5AKLAV0B/p1712602844542619

------


philippotto (approved):
nice :+1:

PR #7748, opened 2024-04-11
Added type definitions for react-remarkable

hotzenklotz:
I added type definitions for the `react-remarkable` package. Unfortunately, there were not `@types/react-remarkable` definitions available to I had to write my own.

### Steps to test:
- Run type checker
- CI should be enough

------


frontend/javascripts/types/react-remarkable.d.ts
```diff
@@ -0,0 +1,19 @@
+
+declare module 'react-remarkable' {
+    import type React from "react";
+    
+    type ReactRemarkableProps = {
+        source: string,
+        container?: string,
```
MichaelBuessemeyer, line 7: React Remarkable also has the option to pass the markdown content as strings mixed with other components as children. Therefore, I would suggest adding this to the type so that future components also have the option to pass the markdown content as children and are not forced to use the source prop. IMO including this in the type definition helps us discover the option use the children props more easily without looking at the docs :).

Moreover, the `source` props is therefore optional. 

You can give this a quick check here in the code of the lib: https://github.com/acdlite/react-remarkable/blob/master/src/index.js

```suggestion
    type ReactRemarkableProps = {
        source?: string,
        container?: string,
        children?: React.ReactNode,
```

Note, that I am unsure whether it is necessary to define the `children` props as the defined Markdown class is defined as a react component. The definition below might already add typical props like `children`, `style` and `className`:
`  declare class Markdown extends React.Component<ReactRemarkableProps> {}`
But this needs a quick check :)

MichaelBuessemeyer (approved):
Thanks for adding the type and improving the typing :pray:

I only found one minor thing, regarding the defined type.
If you disagree with me there, feel free to merge this pr without the small suggestion :)

frontend/javascripts/types/react-remarkable.d.ts
hotzenklotz: `children` and `style` were not automatically part of the default props, hence I had to list the manually. Possibly one could declare the props as a Union of the standard React component props and the specific props but this was the quick and easy way.

MichaelBuessemeyer: thanks for applying the suggest. Let's mergy merge :rocket:

Issue #6861, opened 2023-02-20
Upgrade antd to v5


hotzenklotz:
## Detailed Description
`Antd` [v5 has been released](https://github.com/ant-design/ant-design/issues/38671). We should eventually upgrade to benefit from the lastest patches and new components. Let's track the progress in the issue.

Recommended upgrade steps are:
- [x] Upgrade to latest version of antd v4
  - [x] `4.24.8` #6865 
  - [ ] [4.24.15](https://4x.ant.design/changelog) #7505 
- [x] Solve all deprecation warnings in v4
  - [x] Fix `<Menu>` #6860
  - [x] Fix `<Dropdown>` #6898
  - [x] Fix `<TabPane>` #7469
- [x] Replace `momentjs`with an alternative like `dayjs` #6849 
- [ ] Migrate `Less` styling / Replace antd `Less` variables or use compatibility lib
- [ ] Remove library version resolution added in [4.24.15 PR](https://github.com/scalableminds/webknossos/pull/7505)

Full upgrade guide: https://ant.design/docs/react/migration-v5#legacy-browser-support

## Context

* [ ] Specific to long-running jobs (set `jobsEnabled=true` in `application.conf`)
* [ ] Specific to webknossos.org (set `isDemoInstance=true` in `application.conf`)

PR #7751, opened 2024-04-15
Fix Toast Theme Support and other minor improvements

MichaelBuessemeyer:
This PR modifies the toast lib to render the notifications in the user-set theme. Previously, the theme was always set to light and therefore the background of the toast messages in dark mode was white. The API of the toast lib doesn't change at all in this PR.

How it works: According to the [antd toast docs](https://ant.design/components/notification#faq) antd automatically creates a new rendering hierarchy for each spawned toast. For rendering the toast in the configured theme, a hook needs to be used that returns a context holder and API to open the toasts from. This is now automatically done by the Toast lib by mounting a single functional component (FC) using this hook as a direct child of the `GlobalThemeProvider`. Once this FC is rendered for the first time, the theme-aware notification API is available to the toast lib and can be used to render some toasts. Note, that the FC with its notification API should be treated as a singleton and only mounted once as else an old notification API might be lost with references to e.g. some older toast that are sticky and still being displayed.

- https://allowtoastoutsiderenderincontext.webknossos.xyz/

### Steps to test:
- Open the dev instance and switch to light theme. Then trigger some toasts via e.g. copying the position of a tracing with the pin-icon button next to the position view. The toasts should have a light background.
- Switch to dark mode. And trigger some toasts again. The toasts should now be in dark mode. Switching themes while a toast is visible should also adapt the theme of the toast while it is still visible.
- Open the copy position toast multiple time after another. This should always close the current toast before reopening the same toast.
- While a toast like the copy position toast is visible, switch to a different browser tab. Wait 10 sec and return to the wk tab. The toast should still be visible and close after half of the timeout of the toast (likely 6 -> half is 3 secs).

### Issues:
- No issue exists for this.
- Maybe related to #6861

------
- ~~[ ] Updated [changelog](../blob/master/CHANGELOG.unreleased.md#unreleased)~~
  - I'd say that this is not necessary


MichaelBuessemeyer (commented):
@philippotto Could you please review these changes as I'd say that this PR changes quite heavily how the toast lib works under the hood. The API does not change but each toast now opens a new rendering context to match the desired theme.

Sadly, these changes introduce cyclic dependencies. Please see below.

frontend/javascripts/oxalis/model/bucket_data_handling/bounding_box.ts
```diff
@@ -81,6 +81,15 @@ class BoundingBox {
...
 
+  extend(other: BoundingBox): BoundingBox {
+    const newMin = V3.min(this.min, other.min);
+    const newMax = V3.max(this.max, other.max);
+    return new BoundingBox({
+      min: newMin,
+      max: newMax,
+    });
+  }
+
```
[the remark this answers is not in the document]
philippotto: it can stay in my opinion :)

frontend/javascripts/libs/toast.tsx
```diff
@@ -18,6 +20,29 @@ export type ToastConfig = {
   onClose?: () => void;
 };
 
+export type NotificationAPI = ReturnType<typeof notification.useNotification>[0];
+type ToastMessageCallback = (api: NotificationAPI) => React.ReactNode;
+
+function ToastFunctionalComponentWrapper(props: {
+  type: ToastStyle;
+  message: (api: NotificationAPI) => React.ReactNode;
```
philippotto, line 28:
```suggestion
  message: ToastMessageCallback;
```

frontend/javascripts/libs/toast.tsx
```diff
@@ -141,5 +168,39 @@ const Toast = {
...
   },
+  message(
+    type: ToastStyle,
+    message: ToastMessageCallback | React.ReactNode = "Success :-)",
+    config: ToastConfig = {},
+    details?: string | undefined,
+  ) {
+    renderIndependently((destroy) => {
+      // Deferring destroy to give the Notification API a chance to close the toast via timeout
+      // before unmounting the parent component created with renderIndependently.
```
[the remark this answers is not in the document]
philippotto: please mention that this avoids an error in the code comment. I don't understand 100% why, though. Won't there still be two parties who try to destroy the toast? and the latter one will "lose" as it's already destroyed?

tools/check-cyclic-dependencies.js
```diff
@@ -1,57 +1,72 @@
...
+    "frontend/javascripts/oxalis/shaders/main_data_shaders.glsl.ts",
   ],
   [
-    'frontend/javascripts/oxalis/geometries/plane.ts',
-    'frontend/javascripts/oxalis/geometries/materials/plane_material_factory.ts',
-    'frontend/javascripts/oxalis/shaders/main_data_shaders.glsl.ts'
-  ]
+    "frontend/javascripts/libs/toast.tsx",
+    "frontend/javascripts/libs/render_independently.tsx ",
+    "frontend/javascripts/oxalis/throttled_store.ts ",
```
philippotto, line 55: where does render_independently import throttled store? I don't see this in the code :thinking:

tools/check-cyclic-dependencies.js
```diff
@@ -1,57 +1,72 @@
...
+    "frontend/javascripts/oxalis/store.ts ",
+    "frontend/javascripts/oxalis/model/reducers/annotation_reducer.ts ",
+    "frontend/javascripts/oxalis/model/accessors/view_mode_accessor.ts ",
+    "frontend/javascripts/oxalis/model/accessors/flycam_accessor.ts ",
```
philippotto, line 59: there are trailing spaces in these paths.

philippotto (changes requested):
thanks for fixing this :100: I left some small comments.

tools/check-cyclic-dependencies.js
MichaelBuessemeyer: Yeah it also took me a few seconds to find :eyes:.

The import is done in-time when needed in line 14: https://github.com/scalableminds/webknossos/blob/c47cd6500e653e9b36ec102f6b7bfee9500a45ad/frontend/javascripts/libs/render_independently.tsx#L9-L22

MichaelBuessemeyer (commented):
Thanks :pray: 

I applied your feedback and tried to clarify why the destroy call needs to be deferred.

tools/check-cyclic-dependencies.js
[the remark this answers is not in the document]
philippotto: ahh, but that's a dynamic import. these should be ignored imo. can you add `skipDynamicImports: true` [here](https://github.com/scalableminds/webknossos/blob/master/tools/check-cyclic-dependencies.js#L59)? the dpdm package doesn't document this api parameter well, but I found it [here](https://github.com/acrazing/dpdm/blob/42e9f17f9d108b0422479df80821fc406b6e5a67/src/bin/dpdm.ts#L175).

frontend/javascripts/libs/toast.tsx
```diff
@@ -155,5 +181,40 @@ const Toast = {
...
+      // Making sure onClose destroys the ToastFunctionalComponentWrapper instance.
+      config.onClose =
+        userOnClose != null
+          ? () => {
+              userOnClose();
+              deferredDestroy();
+            }
+          : () => {
+              deferredDestroy();
+            };
```
philippotto, line 205:
```suggestion
        config.onClose = () => {
          if (userOnClose != null) {
            userOnClose();
          }
          deferredDestroy();
        };
```

frontend/javascripts/libs/toast.tsx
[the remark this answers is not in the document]
philippotto: ah ok, now I understand :) antd tries to close the inner toast and `destroy` destroys the outer parent component. therefore, calling `destroy` within onClose is too early, because the toast is not torn down yet. somehow I assumed that both parties would try to destroy the same thing, but they refer to parent VS child component.
knowing that, the comment looks good :sweat_smile:

tools/check-cyclic-dependencies.js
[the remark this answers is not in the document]
MichaelBuessemeyer: Thanks for finding this option :pray:. Works like a charm <3

philippotto:
> Then trigger some toasts via e.g. copying the position of a tracing with the pin-icon button next to the position view. 

The theme looks correct now :+1: But the toast doesn't disappear on its own? The default delay is 6s. I debugged it briefly and notification.close is called, but it doesn't do anything? It works on wkorg, so it shouldn't be related to https://github.com/scalableminds/webknossos/pull/7741 which was merged this week.

frontend/javascripts/theme.tsx
```diff
@@ -82,9 +86,10 @@ export default function GlobalThemeProvider({ children }: { children?: React.Rea
           className={isDarkMode ? "dark-theme" : undefined}
           style={{
             background: "var(--ant-color-bg-base)",
-            height: "calc(100vh - var(--navbar-height))",
+            height: isMainProvider ? "calc(100vh - var(--navbar-height))" : "auto",
           }}
         >
+          {isMainProvider && <ToastContextMountRoot />}
```
MichaelBuessemeyer, line 92: Here the single global instance of `ToastContextMountRoot` is rendered

philippotto:
> Therefore, I came up with a whole new approach by always having only one permanent mounted component wrapping the context for a single notification api.

Great! :100: Could you update the PR description to reflect this? A short sentence would be enough (you could link to your newer comment).

> I fixed this by caching the first async close call for the toast and ignoring future async close calls

Would it simplify things when we would simply call `close(key)` before showing a new toast? if two toasts share the same key, it makes sense to dismiss the first one in my opinion. The current approach seems like it would ignore a second toast instead of extending its life span. Here's what I mean:

```
Current approach:
- open first toast (6s duration)
- after 5s, show second toast with same key (again 6s)
<--- the second toast will be ignored and after 6s in total, no toast will be shown, right?

My suggestion:
- open first toast (6s duration)
- after 5s, show second toast with same key (again 6s) <-- this closes the first toast and opens the second one
<-- after 5s, the toast will reopen. in total, the toast is shown for 11s instead of 6s
```

Does this make sense?

MichaelBuessemeyer:
> Great! 💯 Could you update the PR description to reflect this? A short sentence would be enough (you could link to your newer comment).

Sure :+1: I just edited the description mentioning that it might be important to only have one existing `ToastContextMountRoot`.

> Would it simplify things when we would simply call close(key) before showing a new toast?

Yes, a little imo. But the previous behaviour of the notification / toast API is the one I reproduced: In case a toast with some key already existed, the call was "ignored". The toast was not closed and its timeout also wasn't refreshed. 

I just investigated this further and it seems I am a little wrong: The timeout time and content of the toast is refreshed, but there is no close & open animation. See https://ant.design/components/notification#notification-demo-update for reference.

```
Current approach:
- open first toast (6s duration)
- after 5s, show second toast with same key (again 6s)
<--- the second toast will be ignored and after 6s in total, no toast will be shown, right?
```
Exactly, that how the current version operates. However, in case the content of the toast changes, it should also update in my version. But the timeout is not refreshed.

```
My suggestion:
- open first toast (6s duration)
- after 5s, show second toast with same key (again 6s) <-- this closes the first toast and opens the second one
<-- after 5s, the toast will reopen. in total, the toast is shown for 11s instead of 6s
```
> Does this make sense?

Sure should I implement it exactly this way or should I leave out the closing and reopening animation?

philippotto:
> Sure should I implement it exactly this way or should I leave out the closing and reopening animation?

if it's easy, yes :) if not, let's keep it as is. in the end, it's not super important.

MichaelBuessemeyer:
> if it's easy, yes :) if not, let's keep it as is. in the end, it's not super important.

Yep, it was doable :+1: and should work now. But during testing chrome showed some weird flickering animation while opening a toast. This also appeared on older versions of the branch but not on wk.org. I couldn't track this down to a cause without a larger time investment. This flickering did not appear in Firefox as well and was also suddenly gone in the current version of the code Thus I deemed this to be a chrome hiccup or so.

Could you please look out for such an animation bug during testing? I cannot reproduce it in the current version (testing locally) but I would still like to avoid weird-looking opening animations of toasts on wk.org :sweat_smile:.

Already thank you for testing in advance :pray: :parrot:

frontend/javascripts/libs/toast.tsx
```diff
@@ -142,23 +142,37 @@ const Toast = {
...
+      let cancelledTimeout = false;
       const timeoutToastManually = async () => {
         const splitTimeout = timeout / 2;
         await animationFrame(); // ensure tab is active
         await sleep(splitTimeout);
         await animationFrame();
+        if (cancelledTimeout) {
+          // If the toast has been closed early, don't close it again.
+          return;
+        }
```
philippotto, line 159: I think this should be moved below the second sleep. Otherwise, an old close call could close a new toast if the old one was closed between the first and second sleep :thinking:

MichaelBuessemeyer: Oh you are right 😓

philippotto (approved):
Excellent! Works well for me :+1: Thanks for taking the time to fix this!

Issue #7734, opened 2024-04-03
Adapt UI to show dataset-creating Jobs only to Privileged Users


fm3:
Dataset-creating jobs run a trigger API, which is forbidden for member-only users. This will be enforced  in the backend/worker starting with #6547.

Quoting from https://github.com/scalableminds/webknossos/pull/6547#issuecomment-2034813785

> I think we should make the changes to the UI to prevent members to run jobs that create datasets.

The relevant jobs are currently

 - convert_to_wkw
 - infer_neurons
 - infer_nuclei
 - materialize_volume_annotation

Their buttons should be hidden for users that are neither dataset manager nor team manager nor admin.

PR #7753, opened 2024-04-15
Disable UI starting dataset creating jobs for unprivileged users

MichaelBuessemeyer:
This PR disables the UI elements allowing the creation of a new dataset as an output of a job for normal unprivileged users.

- https://onlyallowprivilegdeuserdatasetcreationjobs.webknossos.xyz/
### Steps to test:
- On the dev instance log in as the sample2@scm.io user and try to access UI allowing you to start a job creating a dataset:
- `<host-ulr-part>/datasets/upload` should now show a 403 error page
  - For this, an improved icon might be preferable that follows wks drawing style
- In tracing view:
  - The "materialize merger mode annotation" button should not be rendered
  - The AI Analysis Button should not be rendered
  - The option "Merge this volume annotation with its fallback layer" of a segmentation layer with fallback data (e.g. l4_sample) should not be rendered

### Open question:
- Is there a UI element I missed while scanning the codebase?

### TODOs:
- [x] Confirm whether keeping the 403 page is ok
  - [x] Get a better image for the forbidden page

### Issues:
- fixes [#7734](https://github.com/scalableminds/webknossos/issues/7734)

------


frontend/javascripts/router.tsx
```diff
@@ -428,6 +428,7 @@ class ReactRouter extends React.Component<Props> {
               <SecuredRouteWithErrorBoundary
                 isAuthenticated={isAuthenticated}
                 path="/datasets/upload"
+                requiresAdminOrManagerRole
```
philippotto, line 431: I think, this prop can be added to a few other components:
```
path="/users/:userId/details"
path="/users"
path="/teams"
path="/timetracking"
path="/reports/projectProgress"
path="/reports/availableTasks"
path="/tasks"
path="/tasks/create"
path="/tasks/:taskId/edit"
path="/tasks/:taskId"
path="/projects"
path="/projects/create"
path="/projects/:projectId/tasks"
path="/projects/:projectId/edit"
path="/datasets/:organizationName/:datasetName/import"
path="/datasets/:organizationName/:datasetName/edit"
path="/taskTypes"
path="/taskTypes/create"
path="/taskTypes/:taskTypeId/edit"
path="/taskTypes/:taskTypeId/tasks"
path="/taskTypes/:taskTypeId/projects"
path="/scripts/create"
path="/scripts/:scriptId/edit"
path="/scripts"
```

philippotto (commented):
Beautiful :) See my one comment, though, as this should be added to lots of other routes, too, as far as I see.

frontend/javascripts/router.tsx
MichaelBuessemeyer: Thanks for listing all other components :+1: 

Initially, I though that this should be added in a separate PR but doing it already here is also fine to me and might also save some time.

I'll test each route as admin and non-admin later :)

MichaelBuessemeyer: > I'll test each route as admin and non-admin later :)

Done, worked out as expected :+1:. All pages/view listed above are available to sample@scm.io but not to sample2@scm.io

PR #7757, opened 2024-04-16
Fix reading CompressedSegmentation with uint32 data

fm3:
The old code padded every 32-bit output value with another 32 zeroes due to the to-long conversion.

As mentioned in https://github.com/scalableminds/webknossos/pull/6947 we didn’t have a uint32 test dataset yet when implementing this.

I also removed some superfluous parentheses.

### Steps to test:
- Import neuroglancer precomputed dataset with compressed uint32 segmentation, e.g. from https://scm.slack.com/archives/C5AKLAV0B/p1713253591614229
- Segmentation should look plausible

------


normanrz (approved):
nice!

PR #7763, opened 2024-04-18
Remove version field from package.json

philippotto:
The value was inaccurate and used for the front-end docs.

- https://___.webknossos.xyz

### Steps to test:
- CI

### Issues:
- fixes that the front-end-api docs mentioned an outdated version


philippotto:
@normanrz thanks for your commits! can you approve?

normanrz (approved):

PR #7764, opened 2024-04-18
Remove support for meshfiles version 2 and older

### LINKED ISSUE #7597: Drop support of old mesh files
The datastore mesh routes currently support two kinds of meshfiles (“v3” and un-versioned (old)).

We should create a small script that detects existing old meshfiles so that we can decide which we need to re-compute in the new format with voxelytics.

This script could reuse some of the code from https://github.com/scalableminds/webknossos/blob/master/tools/migrate-axis-bounds/migration.py but instead of the jsons, read meshfiles with pyhdf5 and check if the format matches.

After all relevant meshfiles have been recreated in the new format, we can drop the support for the old meshfiles from the webknossos datastore and the frontend.

Talk to me for example files of the old and new format

fm3:
Removes support for the old meshfile format

### Steps to test:
- Load some precomputed meshes, with and without mappings

### Issues:
- fixes #7597

------


normanrz (approved):
Backend looks good :+1:

philippotto (approved):
front-end looks great, too :1st_place_medal:

Issue #7688, opened 2024-03-14
Remove Support for On-Disk JSON mappings, convert existing


fm3:
It is tedious to support all kinds of features for both json and agglomerate-file based mappings.

On-Disk Json mappings aren’t really used anymore. We should find and convert the existing ones and then remove this feature.

Note that some of the front-end code dealing with json mappings will probably survive, as it is also used in merger-mode and proofreading.

fm3:
Quoting myself from #7767:

I had hoped to convert all existing json mapping files, but it turns out the new format isn’t actually suited for a 1:1 conversion (agglomerate files are designed to always yield continuous agglomerate ids, json mappings would use the a segment id of each group as agglomerate id). So that’s on hold until we change the agglomerate file format to support sparse agglomerate ids.

fm3:
We’ll revisit this once we decide about a new agglomerate file format.

PR #7767, opened 2024-04-22
Script to find json mappings

fm3:
Adding this script searching the binaryData folder for json mapping files. I had hoped to convert all of these files, but turns out the new format isn’t actually suited for a 1:1 conversion. So that’s on hold until we change the agglomerate file format to support sparse agglomerate ids.

Let’s still save this script in the repository.

### Steps to test:
- `python tools/mapping-format/find-json-mappings.py binaryData`
- (You may want to have a dataset with json mappings such as [ROI2017_wkw](https://www.notion.so/scalableminds/Test-Datasets-c0563be9c4a4499dae4e16d9b2497cfb?pvs=4#419474260d1941e3b4fb4ea58fe23bb1))

### Issues:
- contributes to #7688


tools/mapping-format/find-json-mappings.py
```diff
@@ -0,0 +1,79 @@
...
+            file=sys.stderr,
+        )
+
+
+def format_bytes(num):
+    for unit in ("", "K", "M", "G", "T"):
+        if abs(num) < 1000.0:
+            return f"{int(num)}{unit}"
+        num /= 1000.0
+    return f"{int(num)}P"
```
MichaelBuessemeyer, line 75: Whats the idea behind the `P` here? :see_no_evil:

MichaelBuessemeyer (approved):
LGTM & script works :)

tools/mapping-format/find-json-mappings.py
fm3: Indeed if this line is reached, we are at petabyte :D  the function doesn’t go higher than that.

PR #7768, opened 2024-04-22
Add upgrade motivation banner

### LINKED ISSUE #7745: Add banner "Upgrade version" for self-hosted when version is older than 6 months
## Detailed Description
Create a banner similar to the one in the mockup below. The banner should appear as soon as the self-hosted version of WEBKNOSSOS has not updated since 6 months. It should be possible to close it, but it should appear again with every new session, until a newer version is used (sessionstore). 
Text: You are using an old version of WEBKNOSSOS. Switch to webknossos.org for automatic updates and exclusive features! 
Button: Learn more (Links to [https://webknossos.org/self-hosted-upgrade](url))
banner top self hosted

## Context
When users of the self-hosted version of WEBKNOSSOS haven't upgraded since 6 months, this banner should appear, motivating them to switch to webknossos.org. The button "Learn more" will lead them to a landing page where the advantages of wk.org compared to the open-source version are listed.

dieknolle3333:
### Steps to test:
- mock an outdated WK version, e.g. by mocking getBuildInfo(), see comment in maintenance_banner.tsx
- see banner, make sure it looks good (e.g. across different themes)
- test link
- click it away 
- it shouldn't be shown again, even if you log out / close the browser. it should only be shown again after another 3 days
- it should be shown again in a private browser tab though.

### TODOs:
- [x] fix styling of button, find out how to do theming well
- [x] fix banner when browser window is narrow 
- [x] fix upcoming maintenance banner when user is logged out

### Issues:
- fixes #7745 

------


frontend/javascripts/maintenance_banner.tsx
```diff
@@ -144,3 +147,85 @@ export function MaintenanceBanner() {
...
+      const lastCommitDate = dayjs(commitDateWithoutWeekday, "MMM DD HH:mm:ss YYYY ZZ"); // todo two digit dates? test more once time tracking is merged
+      const needsUpdate = currentDate.diff(lastCommitDate, "month") >= 6;
+      return needsUpdate;
+    },
+    false,
+    [],
+  );
+
+  const getShouldBannerBeShown = () => {
+    if (!isVersionOutdated) return false;
```
dieknolle3333, line 184: (obviously this can be set to true during testing, dont forget to remove)

frontend/javascripts/maintenance_banner.tsx
```diff
@@ -144,3 +147,85 @@ export function MaintenanceBanner() {
...
+  const currentDate = dayjs();
+
+  const isVersionOutdated = useFetch(
+    async () => {
+      const buildInfo = await getBuildInfo();
+      const commitDateWithoutWeekday = buildInfo.webknossos.commitDate.replace(
+        /(Mon)|(Tue)|(Wed)|(Thu)|(Fri)|(Sat)|(Sun)\w*/,
+        "",
+      );
+      const lastCommitDate = dayjs(commitDateWithoutWeekday, "MMM DD HH:mm:ss YYYY ZZ"); // todo two digit dates? test more once time tracking is merged
```
dieknolle3333, line 175: todo

dieknolle3333: https://git-scm.com/docs/git-log
`--date=default is the default format, and is based on ctime(3) output. It shows a single line with three-letter day of the week, three-letter month, day-of-month, hour-minute-seconds in "HH:MM:SS" format, followed by 4-digit year, plus timezone information, unless the local time zone is used, e.g. Thu Jan 1 00:00:00 1970 +0000.`
maybe not

frontend/javascripts/maintenance_banner.tsx
```diff
@@ -144,3 +148,83 @@ export function MaintenanceBanner() {
...
+    left: 0,
+    height: constants.BANNER_HEIGHT,
+    textAlign: "center",
+    backgroundColor: blue,
+    color: white,
+    fontSize: "medium",
+    minWidth: "fit-content",
+  };
+  const customParseFormat = require("dayjs/plugin/customParseFormat");
+  dayjs.extend(customParseFormat);
```
philippotto, line 167: can you {move that code to} and {avoid the require call like it's done} [here](https://github.com/scalableminds/webknossos/blob/2eed0758dbcc0a8f00762c9f3a251a8c6286418d/frontend/javascripts/libs/format_utils.ts#L19)?

frontend/javascripts/maintenance_banner.tsx
```diff
@@ -17,12 +20,13 @@ import { MaintenanceInfo } from "types/api_flow_types";
 
 const INITIAL_DELAY = 5000;
 const INTERVAL_TO_FETCH_MAINTENANCES_MS = 60000; // 1min
+const UPGRADE_BANNER_LOCAL_STORAGE_KEY = "upgradeBannerWasClickedAwayy";
```
philippotto, line 23:
```suggestion
const UPGRADE_BANNER_DISMISSAL_TIMESTAMP_LOCAL_STORAGE_KEY = "upgradeBannerWasClickedAway";
```

frontend/javascripts/navbar.tsx
```diff
@@ -722,6 +724,7 @@ function AnonymousAvatar() {
         icon={<UserOutlined />}
         style={{
           marginLeft: 8,
+          marginTop: navbarHeight,
```
philippotto, line 727: does this fix an old problem or is it necessary for the new banner?

philippotto (changes requested):
Cool stuff :smiley: During testing, I noticed two things:

- I think the banner should only be shown when logged in. Showing this to external visitors doesn't make much sense, in my opinion.
- webknossos.org should be clickable probably (should open in a new tab)

frontend/javascripts/maintenance_banner.tsx
```diff
@@ -144,3 +148,83 @@ export function MaintenanceBanner() {
...
+    fontSize: "medium",
+    minWidth: "fit-content",
+  };
+  const customParseFormat = require("dayjs/plugin/customParseFormat");
+  dayjs.extend(customParseFormat);
+  const currentDate = dayjs();
+
+  const isVersionOutdated = useFetch(
+    async () => {
+      let buildInfo = await getBuildInfo();
```
philippotto, line 172: I'd add a `await Utils.sleep(5000)` before this line, so that WK can make more high-priority requests first.

frontend/javascripts/navbar.tsx
dieknolle3333: This picture was supposed to show that it was a problem before, and I think you can tell from the picture. Anyway, I think I will fix this bug aswell (the black space it supposed to show an upcoming maintenance)
image

dieknolle3333: there we go, this is how it was before
image

frontend/javascripts/maintenance_banner.tsx
```diff
@@ -144,3 +152,102 @@ export function MaintenanceBanner() {
...
+    minWidth: "fit-content",
+  };
+  const currentDate = dayjs();
+
+  const activeUser = useSelector((state: OxalisState) => state.activeUser);
+  const isVersionOutdated = useFetch(
+    async () => {
+      Utils.sleep(INITIAL_DELAY);
+      let buildInfo = await getBuildInfo();
+      buildInfo.webknossos.commitDate = "Wed Apr 24 22:40:38 2023 +0200"; // one year back for testing purposes, TODO remove
```
dieknolle3333, line 177: hello

frontend/javascripts/navbar.tsx
philippotto: > This picture was supposed to show that it was a problem before

> there we go, this is how it was before

I'm confused. Both pictures show the "before this PR" state?

dieknolle3333: exactly.

dieknolle3333: image
this is how it looks now

frontend/javascripts/navbar.tsx
```diff
@@ -701,6 +700,9 @@ function LoggedInAvatar({
 }
 
 function AnonymousAvatar() {
+  const navbarHeight = useSelector(
```
philippotto, line 703: rename to `bannerHeight` or something like this.

frontend/javascripts/maintenance_banner.tsx
```diff
@@ -1,10 +1,13 @@
 import {
```
philippotto, line 1: rename file to banners.

frontend/javascripts/banners.tsx
```diff
@@ -144,3 +152,101 @@ export function MaintenanceBanner() {
...
+    color: white,
+    fontSize: "medium",
+    minWidth: "fit-content",
+  };
+  const currentDate = dayjs();
+
+  const activeUser = useSelector((state: OxalisState) => state.activeUser);
+  const isVersionOutdated = useFetch(
+    async () => {
+      Utils.sleep(INITIAL_DELAY);
```
philippotto, line 175:
```suggestion
      await Utils.sleep(INITIAL_DELAY);
```

philippotto: also you can add a `if (!activeUser) return false` at the top of this hook (don't forget to add activeUser to the dependencies). this saves one request for not-logged-in-users.

philippotto (commented):
Testing went well mostly :tada: However, when switching the theme while the banner is shown, the navbar height is not set properly:

image

If it's easy to fix, please go ahead. Otherwise, it's not that bad since the case is rather rare.

Also, see my two newest comments :)

frontend/javascripts/banners.tsx
```diff
@@ -144,3 +152,101 @@ export function MaintenanceBanner() {
...
+    }
+  }, [shouldBannerBeShown]);
+
+  return shouldBannerBeShown ? (
+    <Alert
+      className="upgrade-banner"
+      message={
+        <Space size="middle">
+          <Space size="small">
+            You are using an old version of WEBKNOSSOS. Switch to
```
hotzenklotz, line 218:
```suggestion
            You are using an outdated version of WEBKNOSSOS. Switch to
```

dieknolle3333:
@philippotto thanks for the feedback! actually it was quite easy to ensure the right nav bar height while switching the theme. 

I also worked on the case that both a maintenance and an upgrade banner could be shown. I solved it so that only the maintenance banner is shown in that case. Once that is gone, the navbar height is broken (like in your last picture) though. I left it like that because the upgrade banner can be clicked away, because it is a rare case anyway and because fixing it would require letting maintenance and upgrade banner know about each other, which would need some refactoring. Let me know if you disagree and think that I should look into it further :)

philippotto (approved):
awesome :)

PR #7765, opened 2024-04-22
Update React to v17

hotzenklotz:
- Update [React to v17](https://legacy.reactjs.org/blog/2020/10/20/react-v17.html)
  - Update React, React-Dom, React-Test-Renderer
- Fix some dayjs types that used different dayjs imports
- Updated several antd / `rc-` libs
- Remove last remains ofthe `enzyme` React component testing framework

### TODOs:
- [x] Blocked by #7772  which solves a Typescript issue
- [x] Blocked by #7733 which removes `ReactGoogleCharts` package
- [x] Replace `react-remarkable` with a maintained markdown renderer, e.g. `react-markdown` #7813
- [x] Replace `react-sortable-tree` dependency with antd `<Tree/>` #7819

### Issues:
- Contributes to #5795

------
(Please delete unneeded items, merge only when none are left open)
- [x] Updated [changelog](../blob/master/CHANGELOG.unreleased.md#unreleased)
- [ ] Updated [migration guide](../blob/master/MIGRATIONS.unreleased.md#unreleased) if applicable
- [ ] Updated [documentation](../blob/master/docs) if applicable
- [ ] Adapted [wk-libs python client](https://github.com/scalableminds/webknossos-libs/tree/master/webknossos/webknossos/client) if relevant API parts change
- [x] Removed dev-only changes like prints and application.conf edits
- [x] Considered [common edge cases](../blob/master/.github/common_edge_cases.md)
- [ ] Needs datastore update after deployment

PR #7772, opened 2024-04-26
Refactor Antd Table columns

hotzenklotz:
In preparation for upgrading to React 17 (#7765), I ran into trouble with Typescript and infinite recursions ("type definition to deep") resulting in crashes. Most likely this is not caused by the React update per se but some other dependency, likely any from the `rc-*` packages used by `antd`. 

In particular, Typescript was unhappy with the type `RowRenderer` in any of the table `<Column>` elements:
```
<Column
   ...
  render: (_name: string, rowRenderer: RowRenderer) => rowRenderer.renderNameColumn()
/>
```

Instead of using `<Column />` children for the `FixedExpandableTable`, I refactored this components to use objects and props instead, e.g. `<Table columns={ ... } />`. Judging from the [antd documentation](https://ant.design/components/table#api) this seems to be the preferred way these days. (Similar to `<Dropdown />`, `<Menu />` etc).

### Steps to test:
- Make sure that Dashboard > Datasets table works
- Make sure that Admin > Tasks table works

### Issues:
- Contributes to #7765 

------


MichaelBuessemeyer (changes requested):
Thanks for fixing this :tada: 

> Judging from the [antd documentation](https://ant.design/components/table#api) this seems to be the preferred way these days. (Similar to <Dropdown />, <Menu /> etc).

Yes, you are correct as far as I remember :+1: 

Thanks for this PR. Please find my comments below.

One thing: The PR does not state why moving the column definition into an object instead of an react child fixes the Typescript hiccup. In case you know why, please add a short explanation in a comment in case someone later comes back to this PR and thus gets the proper reasoning / understanding for this change :pray:

Once the mini feedback is applied I'd say its all green :green_circle:

hotzenklotz:
> One thing: The PR does not state why moving the column definition into an object instead of an react child fixes the Typescript hiccup. In case you know why, please add a short explanation in a comment in case someone later comes back to this PR and thus gets the proper reasoning / understanding for this change 🙏

I don't know why React v17 + Typescript ran into an issue. I can only assume that TS tries to automagically infer types for `<Table />` children and that the React v17 internals changed somehow which leads to an infinite recursion and TS crashing. The solution in the PR applies more(manual) type hints for the column children leading to less dark type magic by TS.

MichaelBuessemeyer (approved):
Thanks for applying the suggestion.

Let's :shipit: :ship:

frontend/javascripts/admin/task/task_list_view.tsx
```diff
@@ -237,6 +236,176 @@ function TaskListView({ initialFieldValues }: Props) {
...
+      sorter: Utils.compareBy<APITask>((task) => task.created),
+      render: (created: APITask["created"]) => <FormattedDate timestamp={created} />,
+      defaultSortOrder: "descend",
+    },
+    {
+      title: "Action",
+      key: "actions",
+      width: 170,
+      fixed: "right",
+      render: (_, task: APITask) => (
```
MichaelBuessemeyer, line 355: Here you are shadowing the lodash variable. That's why calls within this rendering function using lodash do not work. See my highlighted example below. Just give this a different name and it should work:
```suggestion
      render: (_unused, task: APITask) => (
```

frontend/javascripts/admin/task/task_list_view.tsx
```diff
@@ -237,6 +236,176 @@ function TaskListView({ initialFieldValues }: Props) {
...
+          <div>
+            <a href={`/tasks/${task.id}/edit`} title="Edit Task">
+              ,
+              <EditOutlined className="icon-margin-right" />
+              Edit
+            </a>
+          </div>
+          {task.status.pending > 0 ? (
+            <div>
+              <LinkButton onClick={_.partial(assignTaskToUser, task)}>
```
MichaelBuessemeyer, line 377: Here is an example that causes the rendering to crash as the current code shadows the lodash variable

frontend/javascripts/admin/task/task_list_view.tsx
```diff
@@ -237,6 +236,176 @@ function TaskListView({ initialFieldValues }: Props) {
...
+                }}
+                title="Download all Finished Annotations"
+                icon={<DownloadOutlined className="icon-margin-right" />}
+              >
+                Download
+              </AsyncLink>
+            </div>
+          ) : null}
+          <div>
+            <LinkButton onClick={_.partial(deleteTask, task)}>
```
MichaelBuessemeyer, line 399: Here as well btw :)

frontend/javascripts/admin/task/task_list_view.tsx
```diff
@@ -336,174 +506,8 @@ function TaskListView({ initialFieldValues }: Props) {
...
-            sorter={Utils.compareBy<APITask>((task) => task.created)}
-            render={(created) => <FormattedDate timestamp={created} />}
-            defaultSortOrder={"descend"}
-          />
-          <Column
-            title="Action"
-            key="actions"
-            width={170}
-            fixed="right"
-            render={(__, task: APITask) => (
```
MichaelBuessemeyer, line 455: The previous code seems to have used two underscores to avoid the shadowing. Imo using  `_unused` is more explicit and easier to spot than using two underscores to avoid the name shadowing. Therefore, the `_unused`  approach is more explicit and therefore imo more preferable.

MichaelBuessemeyer (changes requested):
I found the bug: It was cased by name shadowing. Please see my comments below.

After this is fix, please ask for a quick re-testing that then it should be ready to merge (hopefully :see_no_evil:).

And sorry for the take backsies, but it was necessary

MichaelBuessemeyer (approved):
Now it works :tada:

PR #7776, opened 2024-04-29
Fix error message on localFolderWhitelist

normanrz:


fm3 (approved):
to be honest I think localFolderWhitelist is not a good name for this config setting. While we’re here, maybe we should clean that up. I think to avoid confusion, we should use “Folder” only in context of the virtual in-database folders used for the dashboard, and “Directory” or “Path” everywhere else, if possible. What do you think?

However, if you need this to produce a readable error message quickly, feel free to merge this as is first and we’ll do the cleanup renaming later.

normanrz:
Fair enough. We can change the name of the key. Is that actually a per-datastore config?

fm3:
> Is that actually a per-datastore config?

Yes, as it is assumed that datastores can each have their own filesystem. The wk side never reads this config value, only the datastore module.

fm3:
Looks like this is messed up more than I expected, the wk side also reads this list from the datastore config block, to assert this during exploring. I think that does not really make sense with a multi-datastore setup. I’ll have to have a closer look at this stuff.

fm3:
Follow-up issues https://github.com/scalableminds/webknossos/issues/7781 and https://github.com/scalableminds/webknossos/issues/7780

Should’t block this error message fix.

PR #7782, opened 2024-05-02
Minor Upgrades for Backend Dependencies

fm3:
### Steps to test:
- CI should be enough

------


project/Dependencies.scala
```diff
@@ -92,19 +92,19 @@ object Dependencies {
     // End-to-end test specs
     specs2 % Test,
     // Writing XML. import com.sun.xml.txw2
-    "org.glassfish.jaxb" % "txw2" % "4.0.2",
+    "org.glassfish.jaxb" % "txw2" % "4.0.5",
     // Makes txw2 write self-closing tags in xml (which we want). Not imported.
     "org.codehaus.woodstox" % "wstx-asl" % "4.0.6",
     // Json Web Tokens (used for OIDC Auth). import pdi.jwt
-    "com.github.jwt-scala" %% "jwt-play-json" % "9.2.0",
+    "com.github.jwt-scala" %% "jwt-play-json" % "10.0.0",
```
fm3, line 99: https://github.com/jwt-scala/jwt-scala/releases says no breaking changes, just the play 3 upgrade

test/backend/DataVaultTestSuite.scala
```diff
@@ -68,7 +68,7 @@ class DataVaultTestSuite extends PlaySpec {
         "returns failure" when {
           "requesting invalid range" in {
             val result = (vaultPath / dataKey)
-              .readBytes(Some(Range.Long(-10, 10, 1)))(globalExecutionContext)
+              .readBytes(Some(Range.Long(-5, -10, 1)))(globalExecutionContext)
```
fm3, line 71: Negative starts for byte ranges are now actually allowed in GCS. So we need to test an *even more invalid* range to check that a failure is indeed returned (this test is intended to check that no exceptions bubble up but instead the tryo turns them into Box.Failures).

MichaelBuessemeyer (approved):
LGTM. Thanks for bumping those dependencies :tada:

Issue #6577, opened 2022-10-21
Support voxel size in other units than nm


hotzenklotz:
## Detailed Description
ATM, voxel size is defined in nm. While that is good default for high-resolution EM datasets, it is very impractical for MRI/CT dataset, light microscopy, and low-resolution, large FOV EM scans. 

We should refactor the voxel size such that the actual unit does not matter as much as long as the ratio is correct, e.g., 150x150x300nm vs 150000x150000x300000nm (converted from um) == 15x15x30 <some unit>.

Maybe the unit could be selected from a dropdown in the UI and wK does all the conversion internally.

This is also relevant for any work for correlative, multi-scale datasets. Related to #6085

## Context

* [ ] Specific to long-running jobs (set `jobsEnabled=true` in `application.conf`)
* [ ] Specific to webKnossos.org (set `isDemoInstance=true` in `application.conf`)

fm3:
We would also have to decide if the internal float32 nm representation is still good enough. If so, this is a frontend-only issue. If we need very large voxel sizes, say, multiple meters (e.g. for geodata?), we would have to change the internal representation too (possibly storing the unit) to not lose precision.

philippotto:
> If we need very large voxel sizes, say, multiple meters (e.g. for geodata?), we would have to change the internal representation too (possibly storing the unit) to not lose precision.

I think, storing the unit in addition to the scale would be a good approach. The migration would be easy (a missing `unit` would default to `nm`). Most of the front-end code wouldn't need to be touched (as often only the relations are important for code correctness). Only the parts where the actual unit is important to the user would need adaption. This is also in line with @hotzenklotz suggestion:

> We should refactor the voxel size such that the actual unit does not matter as much as long as the ratio is correct

This would automatically fix/improve some things, such as:

- move speed would be in the appropriate order of magnitude
- clipping distances in the 3D viewport wouldn't have problems with large numbers for "large" physical distances (e.g., meters)

will-moore:
Hi - this issue would include the ability to read units from e.g. OME-NGFF.
When I import an image with microns: e.g. https://uk1s3.embassy.ebi.ac.uk/idr/zarr/v0.4/idr0062A/6001240.zarr
the pixel size values are read correctly but the units are ignored:
Public view at https://webknossos.org/datasets/18c44a2271fb102d/omeroinfo/view?token=8QbWDRXlCP9aWxxQUHzayQ

[...]

PR #7783, opened 2024-05-02
Support other units than nm for datasets

MichaelBuessemeyer:
- https://___.webknossos.xyz

### Steps to test:
- See "Features to test" below

### TODOs:
- [x] Never first calculate to nm and then into the desired unit to avoid potential precision loss for very large and very small units.
- [x] Check Arbitrary View support for this feature :see_no_evil: 
- [x] Write tests!
- [x] Write tests for uncommon units like imperial units or pc
- [x] Adjust datasource schema validated by the frontend to support all unit. File: frontend/javascripts/types/schemas/datasource.schema.ts
- [x] Move Parsec to uncommon units
- [x] Support `cm` as a format to be formatted to. 
  - [x] Now the tests need to be fixed :woozy_face: 
- [x] Import remote dataset should also support the units
  - [x] Backend
  - [x] Frontend (?)
- [x] How does this interact with meshes? Do meshfiles need unit metadata?
- [x] Bump backend API version, provide adapter (datasource json format has changed)
- [x] Write into mesh file docs that the mesh file scale now references the dataset unit. (sneaked it into https://github.com/scalableminds/voxelytics/pull/3491 )
- [x] PR [#7800](https://github.com/scalableminds/webknossos/pull/7800/files) introduced a lot of merge conflicts and they need to be resolved.
- [ ] Build auto conversion for huge scales into a better fitting unit.

### After this PR / Follow up
- [ ] webknossos-libs need to support units: https://github.com/scalableminds/webknossos-libs/issues/1076
- [ ] Follow-up: Worker job convert_to_wkw should pass this to the libs

### Open Questions:
- [x] Open question: When a length is given in the unit set for the datasource scale, how should this be called in the code? Suggestions:
  - [ ] `...InDatasourceUnit`
  - [ ] `...InDatasetUnit`
  - [x] `...InUnit`
- [x] Open question: What to do when an nml is loaded where the unit differs to the dataset?
  - [x] Gets ignored
- [x] Currently, `1.23456789cm²` is formatted to `0.001m²` in case a decimal precision of 4 is wanted. Is this how it should be or is this unwanted?
  - This is controlled by the parameter `preferShorterDecimals` of the `formatNumberToUnit` function. Here is an example
  ```
  1.23456789cm ->  1.23457 cm prefer shorter decimals: false
  1.23456789cm ->  0.00001 km prefer shorter decimals: true
  ``` 
  I would prefer to have `preferShorterDecimals` false as default as it does not cut off that much precision! -> Agreed upon.
- [x] `preferShorterDecimals = false` leads to `0.1nm³` getting formatted to `100000000.0 pm³`. This is very likely unwanted?!. I could refactor the formatting to always prefer the unit that has fewer digits. But this might require some additional string operations thus making the code potentially slower -> In such a case: measure and look at how frequently the code is called.

### Features to test:
- [x] Adhoc meshing
- [x] loading precomputed meshes
- [x] Configuring and saving a unit for a local dataset
- [x] Configuring and saving a unit for a remote dataset
- [x] Dataset Info tab (right menu); dataset scale & extent
- [x] Measure tree length
- [x] Find the shortest path between nodes
  - [x] Bug Fixed: The vx of "Path length to this Node" is wrong. In the following, the `mm` seems correct, but the vx is off.
    - The shortest path length between the nodes is 2.0 mm (175.14 vx).
    - The shortest path length between the nodes is 4.1 mm (103.25 vx).
- [x] 3d viewport interactions: rotation, movement, clicking on meshes/skeletons, context menu, ... 
- [x] Bounding Box tool & Tab
- [x] Skeleton Tool & Tab & Tree hierarchy view
- [x] Volume annotations
- [x] Saving & reloading of annotations
- [x] Line & Area Measurement Tool & its Tooltip
- [x] Moving through the dataset (moving the flycam)
- [x] nml serialization should include the unit (frontend)
- [x] nml serialization should include the unit (backend)
- [x] Proofreading mode
- [x] Download annotation modal
- [x] AI Jobs Modal (displaying bounding boxes in proper unit)
- [x] Context Menu
- [x] `api...getViewportData`
  - [x] tested via "Click Histogram feature" as this is the only code part using this API function.
- [x] volume interpolation (uses -> getBoundingBoxForViewport)
- [x] Segment Stats download

### Issues:
- fixes [#6577](https://github.com/scalableminds/webknossos/issues/6577)

------
- [ ] Updated [documentation](../blob/master/docs) if applicable -> Should we do this?
- [ ] Adapted [wk-libs python client](https://github.com/scalableminds/webknossos-libs/tree/master/webknossos/webknossos/client) if relevant API parts change → Follow-up https://github.com/scalableminds/webknossos-libs/issues/1076


frontend/javascripts/oxalis/controller/camera_controller.ts
```diff
@@ -72,20 +72,25 @@ class CameraController extends React.PureComponent<Props> {
...
     );
-    this.props.cameras[OrthoViews.TDView].far = diagonalDatasetExtent * 2;
+    const far = Math.max(8000000, diagonalDatasetExtent * 2);
+
+    for (const cam of _.values(this.props.cameras)) {
+      cam.near = 0;
+      cam.far = far;
+    }
+    // This value is correct
+    this.props.cameras[OrthoViews.TDView].far = far;
```
MichaelBuessemeyer, line 93: Previously, the value calculated for the far plane was overwritten in lines 103-108 after the 3d viewport was properly initialized. I modified the code to the larger of both values is kept.

frontend/javascripts/oxalis/api/api_latest.ts
```diff
@@ -1162,18 +1160,17 @@ class TracingApi {
...
         const neighbourDistance =
-          distance + V3.scaledDist(nextNodePosition, neighbourPosition, datasetScale);
+          distance + V3.scaledDist(nextNodePosition, neighbourPosition, datasetScaleFactor);
 
         if (neighbourDistance < getDistance(neighbourNodeId)) {
           distanceMap[neighbourNodeId] = neighbourDistance;
           parentMap[neighbourNodeId] = source === nextNodeId ? source : target;
           const neighbourDistanceVx = V3.length(V3.sub(nextNodePosition, neighbourPosition));
-          distanceMapVx[neighbourNodeId] = neighbourDistanceVx;
+          distanceMapVx[neighbourNodeId] = neighbourDistanceVx + distanceMapVx[nextNodeId];
```
MichaelBuessemeyer, line 1169: Here, I fixed a longer term bug that came up during testing: Previously, the vx distance was only the distance between the last two nodes (in vx). The normal length unit was correct. The bug was due to not accumulating the previously calculated vx distance. 

The bug is also referenced / explained in the pr description.

frontend/javascripts/libs/format_utils.ts
```diff
@@ -217,13 +383,15 @@ export function formatLengthAsVx(lengthInVx: number, roundTo: number = 2): strin
...
-  formattingFunction: (arg0: number) => string,
+  unit: string,
+  formattingFunction: (length: number, unit: string) => string,
 ): string {
-  return `${formattingFunction(extent.width)}${ThinSpace}×${ThinSpace}${formattingFunction(
+  return `${formattingFunction(extent.width, unit)}${ThinSpace}×${ThinSpace}${formattingFunction(
     extent.height,
-  )}${ThinSpace}×${ThinSpace}${formattingFunction(extent.depth)}`;
+    unit,
+  )}${ThinSpace}×${ThinSpace}${formattingFunction(extent.depth, unit)}`;
```
philippotto, line 394: been like that before, but still: it probably would be a bit nicer to write

```
[formattingFunction(extent.width, unit), ...].join(`${ThinSpace}×${ThinSpace}`)
```

frontend/javascripts/libs/format_utils.ts
```diff
@@ -183,32 +315,66 @@ export function formatCountToDataAmountUnit(
...
   decimalPrecision: number = 1,
-): number {
-  const sortedFactors = getSortedFactors(unitMap);
-  let closestFactor = sortedFactors[0];
-  const minumumToRoundUpToOne = 0.95;
+): [number, string] {
+  const isLengthUnit = unit in UnitsMap;
+  let factorToNextSmallestCommonUnit = 1;
+  if (isLengthUnit) {
+    // In case of an length unit, ensure it is among the common length units that we support conversion for.
```
philippotto, line 349:
```suggestion
    // In case of a length unit, ensure it is among the common length units that we support conversion for.
```

frontend/javascripts/libs/format_utils.ts
```diff
@@ -183,32 +315,66 @@ export function formatCountToDataAmountUnit(
...
   preferShorterDecimals: boolean = false,
   decimalPrecision: number = 1,
-): number {
-  const sortedFactors = getSortedFactors(unitMap);
-  let closestFactor = sortedFactors[0];
-  const minumumToRoundUpToOne = 0.95;
+): [number, string] {
+  const isLengthUnit = unit in UnitsMap;
+  let factorToNextSmallestCommonUnit = 1;
+  if (isLengthUnit) {
```
philippotto, line 348: in what cases would this be false? for voxels and pixels?

frontend/javascripts/oxalis/view/scalebar.tsx
```diff
@@ -21,6 +21,14 @@ type StateProps = {
 };
 type Props = OwnProps & StateProps;
 
+function convertPixelsToUnit(
```
philippotto, line 24: move to formatting utils?

frontend/javascripts/oxalis/model/accessors/dataset_accessor.ts
```diff
@@ -321,11 +323,11 @@ export function getDatasetExtentAsString(
 
   if (inVoxel) {
     const extentInVoxel = getDatasetExtentInVoxel(dataset);
-    return `${formatExtentWithLength(extentInVoxel, (x) => `${x}`)} voxel`;
+    return `${formatExtentInUnitWithLength(extentInVoxel, "vx", (x, _) => `${x}`)} voxel`;
```
philippotto, line 326: I think, the interface of formatExtentInUnitWithLength should be as before (the unit should not be passed separately). The unit can simply be accessed within the arrow function. In the current case, the unit is not needed (and formatExtentInUnitWithLength also only needs the unit to pass it to the lambda). Therefore, the code here wouldn't change at all, I think.
Maybe consider renaming the function.

frontend/javascripts/oxalis/model/accessors/dataset_accessor.ts
```diff
@@ -321,11 +323,11 @@ export function getDatasetExtentAsString(
...
   }
 
-  const extent = getDatasetExtentInLength(dataset);
-  return formatExtentWithLength(extent, formatNumberToLength);
+  const extent = getDatasetExtentInUnit(dataset);
+  return formatExtentInUnitWithLength(extent, dataset.dataSource.scale.unit, formatNumberToLength);
```
philippotto, line 330: according to my suggestion from above, this would be:

```
  return formatExtentInUnitWithLength(extent, (length) => formatNumberToLength(length, dataset.dataSource.scale.unit));
```

philippotto (changes requested):
awesome! really like the changes :) I left a couple of comments, but nothing major. didn't test, though. first, a back-end review makes sense, I think.

one question: what happens for old datasets that have a huge scale (e.g., 9000000**3) in nm? will these be adapted to a more appropriate scale automatically? I think, this would be cool to fix imprecision problems. the conversion could be done when the front-end loads the scale information (either by back-end or front-end). what do you think?

fm3:
> one question: what happens for old datasets that have a huge scale (e.g., 9000000**3) in nm? will these be adapted to a more appropriate scale automatically? I think, this would be cool to fix imprecision problems. the conversion could be done when the front-end loads the scale information (either by back-end or front-end). what do you think?

Interesting idea! I think I would prefer to go through some datasets manually and adapt their jsons where problems are seen. This way we can ensure that no mismatches with respective mesh files are created, and we don’t introduce a magic conversion between what’s on disk and what’s used in the code. I would be afraid that this on-the-fly change could make future debugging harder. What do you think?

CHANGELOG.unreleased.md
```diff
@@ -19,5 +20,6 @@ For upgrade instructions, please check the [migration guide](MIGRATIONS.released
 ### Fixed
 
 ### Removed
+- HTTP API versions 1 and 2 are no longer supported. Current is 7.
```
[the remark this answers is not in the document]
fm3: As @frcroth pointed out, there is quite some debate about what makes a HTTP API qualify as RESTful, also compare https://stackoverflow.com/questions/28510876/is-my-api-a-restful-api – and while ours certainly clears some of those points, I understand why he says we shouldn’t be claiming to have a REST API, so I went for the softer term.  Do you think HTTP API is ambiguous?

webknossos-datastore/app/com/scalableminds/webknossos/datastore/models/datasource/InboxDataSource.scala
```diff
@@ -37,12 +37,12 @@ package object inbox {
...
   case class UnusableDataSource[+T <: DataLayerLike](id: DataSourceId,
                                                      status: String,
-                                                     scale: Option[Vec3Double] = None,
+                                                     scale: Option[VoxelSize] = None,
                                                      existingDataSourceProperties: Option[JsValue] = None)
       extends GenericInboxDataSource[T] {
     val toUsable: Option[GenericDataSource[T]] = None
 
-    val scaleOpt: Option[Vec3Double] = scale
+    val scaleOpt: Option[VoxelSize] = scale
```
MichaelBuessemeyer, line 45: here again: Maybe call ist `voxelSizeOpt` as its type is VoxleSize. But I do also understand that the property name on a dataset json sent to a client still stays `dataset.scale` and thus `scaleOpt` also makes kinda sense. But I would prefer to have it callend `voxelSizeOpt` so that just the publicWrites methods needs to care about calling the json property sent to the client `scale`.

Fee free to argue :)

webknossos-datastore/app/com/scalableminds/webknossos/datastore/models/DataRequests.scala
```diff
@@ -50,7 +50,7 @@ case class WebknossosAdHocMeshRequest(
     mag: Vec3Int,
     cubeSize: Vec3Int, // In target mag
     segmentId: Long,
-    scale: Vec3Double,
+    scale: Vec3Double, // voxel size factor, assumed to be in dataset’s unit
```
MichaelBuessemeyer, line 53: Then maybe call the member `voxeSizeFactorInUnit`?  That's at least the naming schema which I used in the frontend

webknossos-datastore/app/com/scalableminds/webknossos/datastore/models/LengthUnit.scala
```diff
@@ -0,0 +1,81 @@
...
+      case "Em" | "exameter"   => Some(Em)
+      case "Zm" | "zettameter" => Some(Zm)
+      case "Ym" | "yottameter" => Some(Ym)
+      case "Å" | "angstrom"    => Some(Å)
+      case "in" | "inch"       => Some(in)
+      case "ft" | "foot"       => Some(ft)
+      case "yd" | "yard"       => Some(yd)
+      case "mi" | "mile"       => Some(mi)
+      case "pc" | "parsec"     => Some(pc)
+      case _                   => None
```
MichaelBuessemeyer, line 38: Shouldn't it be `Some(None)` here, or is it implicitly converted to an Option because thats what the method expects as a return type and scala does implicit conversion magic?

webknossos-datastore/app/com/scalableminds/webknossos/datastore/services/AdHocMeshService.scala
```diff
@@ -28,7 +28,7 @@ case class AdHocMeshRequest(dataSource: Option[DataSource],
                             dataLayer: SegmentationLayer,
                             cuboid: Cuboid,
                             segmentId: Long,
-                            scale: Vec3Double,
+                            voxelSizeFactor: Vec3Double, // assumed to be in dataset’s unit
```
MichaelBuessemeyer, line 31: :tada:

webknossos-tracingstore/app/com/scalableminds/webknossos/tracingstore/controllers/VolumeTracingController.scala
```diff
@@ -144,15 +149,24 @@ class VolumeTracingController @Inject()(
                  tracingId: String,
                  volumeDataZipFormat: String,
                  version: Option[Long],
-                 voxelSize: Option[String]): Action[AnyContent] =
+                 voxelSize: Option[String],
```
MichaelBuessemeyer, line 152: So this should be called `voxelSizeFactor`, or am I wrong here?
Because in line 160 it is converted to a `voxelSIzeFactorParsedOpt`

webknossos-tracingstore/app/com/scalableminds/webknossos/tracingstore/tracings/editablemapping/EditableMappingService.scala
```diff
@@ -571,7 +571,7 @@ class EditableMappingService @Inject()(
         dataLayer = segmentationLayer,
         cuboid = request.cuboid(segmentationLayer),
         segmentId = request.segmentId,
-        scale = request.scale,
+        voxelSizeFactor = request.scale,
```
MichaelBuessemeyer, line 574: Is this field really called `request.scale` and sent this way by the frontend? Shouln't we change this to `voxelSizeFactor` instead (backend & frontend)? This would make the naming more consistent in my opinion

MichaelBuessemeyer (commented):
Thanks for implementing the backend part @fm3 :pray: 

2 Things besides my inline comments:
1. I found multiple variable names still called `scale` I would prefer to rename them `voxelSize` / `voxelSizeFactor` / `voxelSizeUnit` / `lengthUnit` or such. I understand that the json format exposed to clients needs to stay the same, but I think only the publicWrites methods should take care of calling the dataset property `scale` and not voxelSize` or so.
2. I did not really understand the changes to the files in the folder `webknossos-datastore/app/com/scalableminds/webknossos/datastore/explore/`. Can we please look over this together :pray: :)

conf/evolutions/114-voxel-size-unit.sql
```diff
@@ -0,0 +1,16 @@
+START TRANSACTION;
+
+do $$ begin ASSERT (select schemaVersion from webknossos.releaseInformation) = 113, 'Previous schema version mismatch'; end; $$ LANGUAGE plpgsql;
+
+DROP VIEW webknossos.datasets_;
+
+CREATE TYPE webknossos.LENGTH_UNIT AS ENUM ('ym', 'zm', 'am', 'fm', 'pm', 'nm', 'µm', 'mm', 'cm', 'dm', 'm', 'hm', 'km', 'Mm', 'Gm', 'Tm', 'Pm', 'Em', 'Zm', 'Ym', 'Å', 'in', 'ft', 'yd', 'mi', 'pc');
+
+ALTER TABLE webknossos.datasets RENAME COLUMN scale TO voxelSizeFactor;
+ALTER TABLE webknossos.datasets ADD COLUMN voxelSizeUnit webknossos.LENGTH_UNIT;
```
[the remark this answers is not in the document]
fm3: There shouldn’t be a need for such a migration, since the next inbox scan (every datastore does this once per minute) will report nm for all datasets that don’t have a voxel size unit in their datasource-properties.json. Then it should automatically be set for all usable datasets.

You are correct that this is nullable without a default value because unusable datasets are voxelsize-less, so a unit does not make sense for them either.

conf/webknossos.versioned.routes
```diff
@@ -4,21 +4,40 @@
...
 ->       /v5/                                                                 webknossos.latest.Routes
 
 
-# support changes to v5
+# v4: support changes to v7
+GET      /v4/datasets                                                         controllers.LegacyApiController.listDatasetsV6(isActive: Option[Boolean], isUnreported: Option[Boolean], organizationName: Option[String], onlyMyOrganization: Option[Boolean], uploaderId: Option[String], folderId: Option[String], includeSubfolders: Option[Boolean], searchQuery: Option[String], limit: Option[Int], compact: Option[Boolean])
+GET      /v4/datasets/:organizationName/:datasetName                          controllers.LegacyApiController.readDatasetV6(organizationName: String, datasetName: String, sharingToken: Option[String])
+
+# v4: support changes to v6
+GET      /v4/datasets/:organizationName/:datasetName/isValidNewName           controllers.LegacyApiController.assertValidNewNameV5(organizationName: String, datasetName: String)
```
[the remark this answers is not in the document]
fm3: Yes, this was missing. I went through this file to organize a bit and noticed that this was indeed missing. Every old API version needs to adapt everything that has changed since then.

webknossos-datastore/app/com/scalableminds/webknossos/datastore/explore/ExploreLocalLayerService.scala
```diff
@@ -120,7 +120,7 @@ class ExploreLocalLayerService @Inject()(dataVaultService: DataVaultService)
       layersWithVoxelSizes <- explorer.explore(vaultPath, None)
       (layers, voxelSize) <- adaptLayersAndVoxelSize(layersWithVoxelSizes, None)
       relativeLayers = makeLayersRelative(layers)
-      dataSource = new DataSourceWithMagLocators(dataSourceId, relativeLayers, voxelSize)
+      dataSource = new DataSourceWithMagLocators(dataSourceId, relativeLayers, voxelSize) // Unit
```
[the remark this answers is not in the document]
fm3: Nope, just a remnant, I’ll remove it

webknossos-datastore/app/com/scalableminds/webknossos/datastore/models/datasource/InboxDataSource.scala
fm3: I renamed it to voxelSizeOpt, I think that’s fair. I’ll go through more code and see which parts can be renamed easily without worrying about compatibility.

webknossos-datastore/app/com/scalableminds/webknossos/datastore/models/DataRequests.scala
fm3: Ok, I did that now in the backend. Could you adapt the frontend accordingly?

webknossos-datastore/app/com/scalableminds/webknossos/datastore/datareaders/zarr/NgffMetadata.scala
```diff
@@ -87,12 +60,16 @@ object NgffMultiscalesItem {
...
     val datasets = mags.map(
       mag =>
-        NgffDataset(path = mag.toMagLiteral(allowScalar = true),
-                    List(NgffCoordinateTransformation(
-                      scale = Some(List[Double](1.0) ++ (dataSourceScale * Vec3Double(mag)).toList)))))
+        NgffDataset(
+          path = mag.toMagLiteral(allowScalar = true),
+          List(NgffCoordinateTransformation(
+            scale = Some(List[Double](1.0) ++ (dataSourceVoxelSize.factor * Vec3Double(mag)).toList)))
+      )) // TODO unit of voxel size
```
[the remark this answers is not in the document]
fm3: Looks like it actually is still a todo. I’ll have a look :eyes:

fm3:
Thanks for your suggestions! I incorporated the easy ones already. I’ll have another look at maybe reporting the unit in the NGFF metadata for zarr streaming, and at renaming more instances of scale to voxelSize where possible without breaking compatibility.

Sure, let’s talk about the explore code in person!

fm3:
I also adapted the compose route to expect a full voxel size. Since I don’t know what’s in `linkedDatasets[1].dataSource.scale` in `04_configure_new_dataset.tsx` I’m not sure if that will already work. Could you have a look at the frontend there?

MichaelBuessemeyer:
### Bugs:
- [ ] Adhoc mesh computation does not work after changing the dataset unit.
  - 1. Open the settings for l4_sample and change the unit to cm.
  - 2. View the dataset, load precomputed meshes -> should work; load ad-hoc meshes -> crashes without any error message
 
image

fm3:
>  Adhoc mesh computation does not work

I’d expect that to be due to the renaming in the request json, compare https://github.com/scalableminds/webknossos/pull/7783#discussion_r1624432174

MichaelBuessemeyer:
> I also adapted the compose route to expect a full voxel size. Since I don’t know what’s in linkedDatasets[1].dataSource.scale in 04_configure_new_dataset.tsx I’m not sure if that will already work. Could you have a look at the frontend there? This is additionally to the ad hoc mesh request

Should work :)

MichaelBuessemeyer:
I also renamed `datasetScale` to `voxelSize` everywhere in the frontend, as voxelSize makes more sense and is also the term used in the backend and afaik by the wklibs as well

MichaelBuessemeyer (commented):
I just had another look at the backend code and could not locate anything bug/mistake / improvement :tada:


Besides that, here are my pending comments regarding the feedback from @philippotto

philippotto (approved):
awesome :+1: code looks good and it works great, too.

> - [ ] Build auto conversion for huge scales into a better fitting unit.

I think, it's not really necessary for this PR. we can advise users to do this manually if problems arise.

MichaelBuessemeyer:
Sorry to ping you again @philippotto. Could you please check the newest changes regarding using the long unit name internally instead of the short one in the frontend?

MichaelBuessemeyer:
Newest backend change LGTM :+1:

philippotto:
> Sorry to ping you again @philippotto. Could you please check the newest changes regarding using the long unit name internally instead of the short one in the frontend?

The changes look good :+1: However, I would suggest to rename `UnitShortMap` to `LongUnitToShortUnitMap` or `LongToShortUnitMap` to make clearer what it contains.

MichaelBuessemeyer:
@fm3 Mergy merge today?

PR #7733, opened 2024-04-03
Improve time tracking overview

dieknolle3333:
### URL of deployed dev instance (used for testing):
- https://timetrackingexpandtable.webknossos.xyz/

### Steps to test:
- Make sure to have some tracked times in explorative annotations and tasks
- log in as admin-user and go to "Time tracking overview" through navbar
- query time tracking entries
- expand details for users
- test download buttons 
- test as non-priviledged user (should only see yourself)
- test without login (view shouldn't be accessible) 

### TODOs:
- [x] see issue
- [x] think about csv export structure
- [x] fix types
- [x] adjust tests to new param: getTimeTrackingForUserSpans, getTimeTrackingForUser
- [x] fix that non-admin user cannot see themselves because "page doesnt exist"
- [x] fix that expendable button is in wrong position after adjusting filters
- [x] make sure new project line looks good with segmentation stats

### Issues:
- fixes #7708 

------
(Please delete unneeded items, merge only when none are left open)
- [x] Updated [changelog](../blob/master/CHANGELOG.unreleased.md#unreleased)
- [x] Removed dev-only changes like prints and application.conf edits
- [x] Considered [common edge cases](../blob/master/.github/common_edge_cases.md)

PR #7786, opened 2024-05-06
Show timetracking page to non-privileged users again

fm3:
Follow-up fix for the currently held-back https://github.com/scalableminds/webknossos/pull/7733


MichaelBuessemeyer (approved):
:tada:

PR #7787, opened 2024-05-07
Fix timespan summing for annotations due to annotationLayer join

fm3:
Joining in the annotation layers created duplicate entries for annotations with multiple layers (most of them). This would result in their timespans being counted twice for the SUM.

Extracting this into a WITH statement fixes that.

- https://timetrackingfixjoin.webknossos.xyz

### Steps to test:
- Create some annotations, trace some
- Timetracking should show sensible values (user sum should be sum of user annotations)

### Issues:
- fixes https://scm.slack.com/archives/C5AKLAV0B/p1715064386815799?thread_ts=1715010849.424359&cid=C5AKLAV0B

------


MichaelBuessemeyer (approved):
Thanks for fixing this :pray: 

The dev instance does not appear to have the bug present. 

I also tested locally and compared this branch with the master. By doing so I was able to confirm the bug on the master and that this PR / branch fixes the bug :rocket:

Issue #7788, opened 2024-05-07
TimeTracking overview Follow-up


fm3:
Wishes for the time-tracking overview (follow-up for #7733)
 - [x] search in dropdown by typing project name (looks like currently only the id works)
 - [x] left alignment for stats column, centered looks very noisy with widely differing numbers, see image below
 - [x] speed up query for annotations of user (and/or add spinner for expanding the view)
 - [x] Add option to filter by annotation state (active, finished/archived)
 - [x] Include annotation state in table and CSVs
 - [x] long project names are not fully visible (dropdown menu is very narrow)

PR #7789, opened 2024-05-07
Timetracking Overview Improvements

fm3:
Contributes to #7788 by fixing
 - left-aligned annotation stats in table
 - faster sql query for time summed by annotation (on a production database with 6M annotations: before: 7000ms, after: 100ms)
 - prepares backend to filter by annotation state
 - search in dropdown by typing project name instead of id

Not included for now:
 - frontend for filtering by annotation state
 - include annotation state in csv and displayed table

- https://timetrackingfilterbyannotationstate.webknossos.xyz

### Steps to test:
- Annotate some, open timetracking page, expand user entries, should be fast, look pretty even for annotations with many trees.

### Issues:
- contributes to #7788

------
webknossos/webknossos/client) if relevant API parts change

### DIFF
--- a/conf/webknossos.latest.routes
+++ b/conf/webknossos.latest.routes
@@ -250,9 +250,9 @@ GET           /termsOfService/acceptanceNeeded                                  
 
 # Time Tracking
 # Note, there is also /users/:id/loggedTime
-GET           /time/user/:userId/spans                                                          controllers.TimeController.timeSpansOfUser(userId: String, start: Long, end: Long, annotationTypes: String, projectIds: Option[String])
-GET           /time/user/:userId/summedByAnnotation                                             controllers.TimeController.timeSummedByAnnotationForUser(userId: String, start: Long, end: Long, annotationTypes: String, projectIds: Option[String])
-GET           /time/overview                                                                    controllers.TimeController.timeOverview(start: Long, end: Long, annotationTypes: String, teamIds: Option[String], projectIds: Option[String])
+GET           /time/user/:userId/spans                                                          controllers.TimeController.timeSpansOfUser(userId: String, start: Long, end: Long, annotationTypes: String, projectIds: Option[String], annotationStates: String)
+GET           /time/user/:userId/summedByAnnotation                                             controllers.TimeController.timeSummedByAnnotationForUser(userId: String, start: Long, end: Long, annotationTypes: String, projectIds: Option[String], annotationStates: String)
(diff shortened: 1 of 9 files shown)


frontend/javascripts/admin/admin_rest_api.ts
```diff
@@ -2016,6 +2016,7 @@ export async function getTimeTrackingForUserSummedPerAnnotation(
   if (annotationTypes != null) params.append("annotationTypes", annotationTypes);
   if (projectIds != null && projectIds.length > 0)
     params.append("projectIds", projectIds.join(","));
+  params.append("annotationStates", "Active,Finished");
```
dieknolle3333, line 2019: This is just a prerequisite to show the annotations state in the table and in the CSV, and to filter it, right?

fm3: Right, only for the filtering at this time. I wrote this before I learned that the priority is lower for now. I guess we can use it in the frontend in a later iteration, or remove it again if we decide we don’t want to go for this option :)

MichaelBuessemeyer (approved):
The backend code also looks good to me :+1: 

@fm3, did you take measurements on the speed up of the `summedByAnnotationForUser` query? If you still have them, could you please add them to the PR for documentation purposes? :pray:

conf/webknossos.latest.routes
```diff
@@ -250,9 +250,9 @@ GET           /termsOfService/acceptanceNeeded
...
-GET           /time/overview                                                                    controllers.TimeController.timeOverview(start: Long, end: Long, annotationTypes: String, teamIds: Option[String], projectIds: Option[String])
+GET           /time/user/:userId/spans                                                          controllers.TimeController.timeSpansOfUser(userId: String, start: Long, end: Long, annotationTypes: String, projectIds: Option[String], annotationStates: String)
+GET           /time/user/:userId/summedByAnnotation                                             controllers.TimeController.timeSummedByAnnotationForUser(userId: String, start: Long, end: Long, annotationTypes: String, projectIds: Option[String], annotationStates: String)
+GET           /time/overview                                                                    controllers.TimeController.timeOverview(start: Long, end: Long, annotationTypes: String, teamIds: Option[String], projectIds: Option[String], annotationStates: String)
```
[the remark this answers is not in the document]
fm3: I think the order isn’t relevant here for the compiler, as this is, so to speak, always called by name. If an optional parameter is omitted from a request query, it is auto-filled with None (different from Some(null). In fact we want to avoid null wherever possible, since the type-checker does not deal with it).

However, you are right; at least for style, I’ll move the new parameter so that the optional ones stay last.

PR #7791, opened 2024-05-08
Assert unique volume layer names when parsing NML

### LINKED ISSUE #7775: Raw SQL Error on NML Upload with duplicate layer names
When uploading an annotation zipfile that has duplicate volume layer names in the NML, a raw SQL error is thrown. We should add a scala-side assertion to ensure a readable error message.

[WARN] controllers.AnnotationIOController - Answering 400 at /api/annotations/upload for user xxx: {"messages":[{"error":"SQL Failure: ERROR: duplicate key value violates unique constraint \"annotation_layers_name__annotation_key\"\n  Detail: Key (name, _annotation)=(myLayerName, myAnnotationId) already exists."}]}
at models.annotation.AnnotationDAO.insertOne(Annotation.scala:589)

fm3:
- https://nmluniquelayernames.webknossos.xyz

### Steps to test:
- In an NML file, duplicate the <volume> block, set new ID, but use duplicate name
- Upload should fail with readable error message
- If the name is not duplicated, upload should work

### Issues:
- fixes #7775


MichaelBuessemeyer (approved):
Thanks for fixing this :tada:.

The code lgtm & testing worked :t-rex:

PR #7792, opened 2024-05-08
Remove the await in NmlParser/AnnotationUploadService

### LINKED ISSUE #7551: Remove await from AnnotationUploadService
With #7437 NMLParser/parse returns a Fox, which is handled with await. This await should be removed.

The Fox return value is necessitated by a request to the datastore to check if a segment index file exists (which will then set hasSegmentIndex in the annoation).  

Possible solutions:
- Propagate the Fox through the NMLResult types 
- Call parse with the result of the datastore request (may be difficult because the necessary properties may not be known at that point)

fm3:
Implementing the suggestion from here https://github.com/scalableminds/webknossos/issues/7551#issuecomment-1913085950 (move the determination of hasSegmentIndex to adaptPropertiesToFallbackLayer, which happens *after* the NML parsing)

- https://nmlparserremoveawait.webknossos.xyz

### Steps to test:
- [x] Download + Reupload volume annotation  without fallback layer, should work
- [x] Download + Reupload volume annotation with fallback layer, should work

### Issues:
- fixes #7551

------


MichaelBuessemeyer (approved):
Awesome. Thanks for the refactoring :+1:

PR #7794, opened 2024-05-08
Show wk user in vx workflow list

### LINKED ISSUE #7735: Show WK user name next to owner in voxelytics workflows list
There is already a _user reference in the db for voxelytics_runs.

However, the table only shows the username on the machine where voxelytics was run.

It would be nice if the actual user could also be shown.

(The API would also need to add a user field to `WorkflowListingRunEntry` to achieve this, hence the backend label)

dieknolle3333:
### Steps to test:
- Install voxelytics and follow instructions in worker readme (set directory and enable jobs)
- start a job such as computing a mesh file
- go to voxelytics overview and see your current wk user name in table
- try out filter of user column 

### TODOs:
- [x] host user only as backup for wk user
- [x] view workflows of multiple users -> change in pgadmin
- [x] implement user filter

### Issues:
- fixes #7735

------
- [x] :exclamation:  **Removed dev-only changes like prints and application.conf edits** :exclamation: 


conf/application.conf
```diff
@@ -150,8 +150,8 @@ features {
   taskReopenAllowedInSeconds = 30
   allowDeleteDatasets = true
   # to enable jobs for local development, use "yarn enable-jobs" to also activate it in the database
-  jobsEnabled = false
-  voxelyticsEnabled = false
```
dieknolle3333, line 154: as always ;) dont forget to remove!

frontend/javascripts/types/api_flow_types.ts
```diff
@@ -976,28 +985,18 @@ export type VoxelyticsWorkflowReport = {
...
-  | {
-      state:
-        | VoxelyticsRunState.COMPLETE
-        | VoxelyticsRunState.FAILED
-        | VoxelyticsRunState.CANCELLED
-        | VoxelyticsRunState.STALE;
-      beginTime: Date;
-      endTime: Date;
-    }
-) & {
```
dieknolle3333, line 994: I understand that I took away some complexity. But the type is only used in `workflow_view.tsx `and `api_flow_types.ts`, and for the workflow view there used to be a `as any as VoxelyticsWorkflowListingRun` cast which made this distinction here a little bit irrelevant if I understand correctly. 
Thus I removed the cast (see below) and widened  the type, so that it can be used without casting. I think that the types of `beginTime` and `endTime` (a few lines below) forbid the `VoxelyticsRunState.PENDING` and `VoxelyticsRunState.SKIPPED` values anyway. 
Maybe there is an even better solution, then I will be happy to learn about it!

frontend/javascripts/admin/voxelytics/workflow_list_view.tsx
```diff
@@ -96,18 +104,22 @@ export default function WorkflowListView() {
...
         taskCounts: workflow.taskCounts,
         children: workflow.runs.map((run) => ({
           workflowName: workflow.name,
           workflowHash: workflow.hash,
+          userDisplayName: getUserDisplayName(run),
           ...run,
         })),
       })),
-    [workflows],
-  ) as any as Array<RenderRunInfo>;
```
dieknolle3333, line 110: this is the cast I am talking about in the other comment

app/models/voxelytics/VoxelyticsService.scala
```diff
@@ -77,14 +77,16 @@ object ChunkCounts {
 
 case class WorkflowListingRunEntry(id: ObjectId,
                                    name: String,
-                                   username: String,
+                                   hostusername: String,
                                    hostname: String,
                                    voxelyticsVersion: String,
```
dieknolle3333, line 82: upon first sight it seems to me that the naming convention is mixed here (snake case vs. camel case vs. all small letters). Would this be an opportunity to unify the properties or is it on purpose?

fm3: Yes, feel free to camelCase this stuff!

Also, hostusername can probably be thrown out, as the frontend does not display it anymore?

dieknolle3333: alright, I'll do that! actually the user name on the host is displayed if the wk user couldnt be found, so it is still used. is it okay if I keep it in?

fm3: Ah now I see the left join. Ok fair, if we want to support the rare case of a deleted user. However, in this case the type in scala needs to be Option[String]. And please test locally once that the query still behaves correctly if the user is missing/deleted.

But yes, then let’s keep the hostusername in as a fallback.

frontend/javascripts/admin/voxelytics/workflow_list_view.tsx
```diff
@@ -86,7 +87,13 @@ export default function WorkflowListView() {
 
   usePolling(loadData, VX_POLLING_INTERVAL);
 
-  const renderRuns = useMemo(
+  const getUserDisplayName = (run: VoxelyticsWorkflowListingRun) => {
+    return run.userFirstName != null || run.userLastName != null
+      ? [run.userFirstName, run.userLastName].join(" ")
```
philippotto, line 92:
```suggestion
      ? [run.userFirstName, run.userLastName].join(" ").trim()
```

because `[null, "hi"].join(" ")` evaluates to `' hi'`.

frontend/javascripts/admin/voxelytics/workflow_list_view.tsx
```diff
@@ -86,7 +87,13 @@ export default function WorkflowListView() {
...
+      : run.hostusername;
```
philippotto, line 93: I don't see the property `hostusername` on `VoxelyticsWorkflowListingRun`?

frontend/javascripts/types/api_flow_types.ts
philippotto: > and for the workflow view there used to be an as any as Array<RenderRunInfo> cast which made this distinction here a little bit irrelevant if I understand correctly

The cast is still there, though? the old way was a bit more precise and could lead to case where you know it's a "complete" state and therefor endTime must not be null. in general, I would want to keep these distinctions unless they cause a major pain.

frontend/javascripts/admin/voxelytics/workflow_list_view.tsx
dieknolle3333: but it should be there :sweat_smile: as of now, in . 971 in api_flow_types.

philippotto: ok, apparently, I was blind :see_no_evil: by the way, why is it not camel case and `hostUserName`? (similar to `userFirstLastName`)

dieknolle3333: because hostname wasnt camel cased :sweat_smile: I will change it :)

frontend/javascripts/types/api_flow_types.ts
dieknolle3333: ah, you meant the cast in `parseRunInfo` in `workflow_list_view.tsx`, right? I removed that too. Now I only have type annotations left, or am I mistaking?
I was trying to keep the types as they were before but I am running into the same problems that probably lead to the cast...

app/models/voxelytics/VoxelyticsService.scala
dieknolle3333: local testing after deleting the user from the database went well :) 
image

dieknolle3333:
@fm3 I renamed the scala class members as we talked about in the other thread. This lead to a few changes, can you have another look? I tested it locally and everything seems to work like before.

frontend/javascripts/types/api_flow_types.ts
```diff
@@ -881,27 +882,15 @@ export type VoxelyticsArtifactConfig = {
...
+export type VoxelyticsRunInfo = {
   id: string;
   name: string;
   username: string;
   hostname: string;
   voxelyticsVersion: string;
+  state: VoxelyticsRunState;
+  beginTime: Date | null;
```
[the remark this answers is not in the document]
dieknolle3333: note to self: reverted

frontend/javascripts/admin/voxelytics/workflow_list_view.tsx
```diff
@@ -25,20 +25,24 @@ const persistence = new Persistence<Pick<{ searchQuery: string }, "searchQuery">
...
   return {
     ...workflowInfo,
-    runs: workflowInfo.runs
-      .map(parseRunInfo)
-      .sort((a, b) => b.beginTime.getTime() - a.beginTime.getTime()),
+    runs: workflowInfo.runs.map(parseRunInfo).sort((a, b) => {
+      if (a.beginTime != null && b.beginTime != null)
+        return b.beginTime.getTime() - a.beginTime.getTime();
+      else if (a.beginTime != null) return a.beginTime.getTime();
+      else if (b.beginTime != null) return b.beginTime.getTime();
```
philippotto, line 43: quoting from mdn on the compareFn:

> [compareFn] should return a number where:
> 
> A negative value indicates that a should come before b.
> A positive value indicates that a should come after b.
> Zero or NaN indicates that a and b are considered equal.

in line 42 and 43 you return a positive number which means that a will always come after b (regardless of which one is undefined) which doesn't really make sense. I'm not entirely sure what the context of this function is (does it sort different runs of a workflow?), but you probably want that the element with beginTime != null comes before the one with beginTime == null (or the opposite?):

```suggestion
      else if (a.beginTime != null) return 1;
      else if (b.beginTime != null) return -1;
```

or swap 1 and -1 (depending on the order you think makes sense)

philippotto (commented):
front-end looks good to me :tada: only left one remark about proper sorting. will leave the final approval to @fm3.

fm3 (approved):
Backend LGTM and re-testing looked good :) Feel free to merge after addressing Philipp’s latest comment

frontend/javascripts/admin/voxelytics/workflow_list_view.tsx
dieknolle3333: oh damn, I even read the docs but mixed it up anyways :sweat_smile: yes, it sorts the runs of a workflow once the user expanded them.

dieknolle3333: >  but you probably want that the element with beginTime != null comes before the one with beginTime == null
exactly

frontend/javascripts/admin/voxelytics/workflow_list_view.tsx
```diff
@@ -187,26 +200,26 @@ export default function WorkflowListView() {
...
             })),
             onFilter: (value: string | number | boolean, run: RenderRunInfo) =>
-              run.username.startsWith(String(value)),
+              run.userDisplayName?.startsWith(String(value)) || false,
             filterSearch: true,
           },
           {
-            title: "Hostname",
+            title: "Host",
             dataIndex: "hostname",
```
MichaelBuessemeyer, line 215: The prop `hostname` was renamed to `hostName` right? But not here? is this maybe something?

MichaelBuessemeyer (commented):
I only found one small thing that might have an effect on the correctness of these changes, but I am not sure :woozy_face:

PR #7798, opened 2024-05-13
Add codespell support (config, workflow to detect/not fix) and make it fix few typos

yarikoptic:
More about codespell: https://github.com/codespell-project/codespell .

I personally introduced it to dozens if not hundreds of projects already and so far only positive feedback.

CI workflow has 'permissions' set only to 'read' so also should be safe.

- [x] Signed CLA
- [ ] someone needs to run CI??? seems I am not authorized hence it makes it red


fm3 (commented):
Hi and thank you for this Pull Request and your interest in WEBKNOSSOS!

Codespell seems like a valuable addition to the development toolbox! Since many of our developers aren’t native English speakers, this will potentially catch a lot of small errors.

I added a couple of suggestions for the config file and some individual false positives that I’d like to revert.

Also, we decided that we do not want to run codespell on every push for the moment. We’ll have a look later at how to integrate it into our process in the best way.

Could you add these suggestions to the PR implementation? Then I’ll approve the CI run and we can hopefully merge this soon.

PS: Happy to see in your Github readme that you too enjoyed Survive Style 5+ ;-)

Issue #5087, opened 2021-01-26
Introduce Finished state for Explorative Annotations?


fm3:
When publishing annotations, you do not want to accidentally edit them again later.

The current Finished state is really "Archived". We should rename it to make this distinction. The new "Finished" state should make it read-only, but it should still be in the normal annotation list.

philippotto:
From our discussion:

- (potentially, rename "finished" state to "archived". however, maybe we can also keep it as is. especially, since "archived" wouldn't be a perfect fit for tasks.)
- add a "lockedByUser" (naming up for debate) property to the annotation object. only the owner should be able to mutate this.
- update actions for layers of an annotation will be rejected if lockedByUser is true
- the front-end should show locked annotations as "read-only" in the dashboard (look into how other read-only annotations are rendered)
- a tooltip should explain the reason
- opening an annotation that is read-only, should render as such with an explanation
- for the owner, the possibility to change read-only could be:
  - everywhere where "read-only" is rendered (e.g., in a popover)
  - for not locked annotations, the dropdown menu in the annotation view should allow to do the locking

Two questions I just had:
- Can an annotation layer be referenced by multiple annotations? If so: What happens if one annotation is locked, and the other annotation is mutated that references the same layer? Ideally, this should not be allowed.
- Changing the metadata of an annotation should still be allowed, even when the annotation is locked, right? Otherwise, one wouldn't be able to unlock it. Or should the locked property be the only exception for that? If all meta data can be edited, the UI should explain well that the locking only refers to the actual content of the annotation.

fm3:
> Can an annotation layer be referenced by multiple annotations? 

Please don’t do that :D  (With the current data model, you can’t, since it’s actually the annotation layer that references exactly one annotation)

> Changing the metadata of an annotation should still be allowed, even when the annotation is locked, right? Otherwise, one wouldn't be able to unlock it. Or should the locked property be the only exception for that? If all meta data can be edited, the UI should explain well that the locking only refers to the actual content of the annotation.

Good question, I have no strong preference here. Maybe the locked boolean should really be the only thing, that would certainly be easier to explain to users.

MichaelBuessemeyer:
> Good question, I have no strong preference here. Maybe the locked boolean should really be the only thing, that would certainly be easier to explain to users.

I think that is a good approach. In case some metadata needs to be edited, the owner can unlock it, change the metadata settings as wanted, and then lock it again

PR #7801, opened 2024-05-14
Add option to lock explorative annotations

MichaelBuessemeyer:
- https://___.webknossos.xyz

### Steps to test:
- Create an explorative annotation of your favourite dataset
- Go to the explorative annotations view. There should be a button in the actions column to lock the annotation. Click it. It should add the "locked" tag to the annotation.
- View the annotation by clicking the respective action. In the annotation view, you should not be able to annotate the annotation. A "banner" in the top right corner should inform you about the fact that the annotation is locked by you. A tooltip should give information by whom this annotation is locked. 
- Moreover, there should be a menu item to unlock the annotation in the menu header bar. Once clicked, the annotation is unlocked and reloaded.
- In the explorative annotations view, it should not be possible to archive a locked annotation.

- An annotation cannot be locked and unlocked by a user with whom the annotation is shared.

### TODOs:
- [x] Lock all edit possibilities in the Tracing view
- [x] Write migration
- [x] Fix the 
- [x] bug, that after unlocking an annotation in the annotation view, the locked tag was still present in the explorative annotations list.
- [x] Do not allow to archive locked mappings
- [x] In the frontend: Get rid of `allowUpdateAndIsNotLocked` as a locked annotation will always be set to `allowUpdate: false`

### Issues:
- fixes [#5087](https://github.com/scalableminds/webknossos/issues/5087)

------
  - likely yes


frontend/javascripts/oxalis/view/action-bar/tracing_actions_view.tsx
```diff
@@ -461,16 +475,11 @@ class TracingActionsView extends React.PureComponent<Props, State> {
...
-    const copyAnnotationText =
-      !restrictions.allowUpdate &&
-      activeUser != null &&
-      annotationOwner?.id === activeUser.id &&
-      othersMayEdit
-        ? "Duplicate"
-        : "Copy To My Account";
+    const isAnnotationOwner = activeUser && annotationOwner?.id === activeUser?.id;
+    const copyAnnotationText = isAnnotationOwner ? "Duplicate" : "Copy To My Account";
     const archiveButtonText = task ? "Finish and go to Dashboard" : "Archive";
```
MichaelBuessemeyer, line 483: The previous logic reads to me like this: In case updates are not allowed, I am the owner and others may not edit, I get the text "Copy To My Account" although I am the only person allowed to edit my own annotation?! Wouldn't "Duplicate" make much more sense here?

Therefore, I simplified the logic to whether the annotation is owned by the current user (-> "Duplicate") or not (-> "Copy to my account")

What do you think about my decision? Maybe I am missing something here :thinking:

frontend/javascripts/oxalis/view/components/setting_input_views.tsx
```diff
@@ -527,23 +535,17 @@ export class UserBoundingBoxInput extends React.PureComponent<UserBoundingBoxInp
                   }}
                   onPressEnter={this.handleNameChanged}
                   onBlur={this.handleNameChanged}
-                  disabled={!allowUpdate}
+                  disabled={disabled}
```
MichaelBuessemeyer, line 538: disabled sounded better to me as this is designed like a normal "form/info" input and doesn't need the semantics of the "allowUpdate" restriction setting. IMO the parent component should care about the semantic whether the input should be enabled / disabled :)

But feel free to argue

MIGRATIONS.unreleased.md
```diff
@@ -8,4 +8,7 @@ User-facing changes are documented in the [changelog](CHANGELOG.released.md).
 ## Unreleased
 [Commits](https://github.com/scalableminds/webknossos/compare/24.05.0...HEAD)
 
+- Edited the annotations table to add the option to locked explorative annotations. Migration 114 needs to be executed to update the database schema.
```
philippotto, line 11: I think, we don't mention the reason for a migration usually. cc @fm3

conf/messages
```diff
@@ -231,6 +231,11 @@ annotation.reopen.tooLate=The annotation cannot be reopened anymore, since it ha
 annotation.reopen.notAllowed=You are not allowed to reopen this annotation.
 annotation.reopen.notFinished=The requested annotation is not finished.
 annotation.reopen.failed=Failed to reopen the annotation.
+annotation.reopen.locked=This annotation is locked by the owner and therefore cannot be reopened.
+annotation.isLockedByUser.notAllowed=Only the owner of this annotation is allowed to change the locked state of an annotation.
```
philippotto, line 235: not even an admin can do so?

frontend/javascripts/dashboard/explorative_annotations_view.tsx
```diff
@@ -145,6 +148,24 @@ class ExplorativeAnnotationsView extends React.PureComponent<Props, State> {
...
+    const newTracings = tracings.map((currentTracing) => {
+      if (currentTracing.id !== tracing.id) {
+        return currentTracing;
+      } else {
+        updatedTracing = callback(currentTracing);
+        return updatedTracing;
+      }
+    });
+    this.setModeState({ tracings: newTracings }, this.state.shouldShowArchivedTracings);
+    return updatedTracing;
```
philippotto, line 166: I think, the function does not need to return anything (not used anywhere, I think). then, you can remove the local `updatedTracing` variable etc.

frontend/javascripts/dashboard/explorative_annotations_view.tsx
```diff
@@ -324,24 +378,7 @@ class ExplorativeAnnotationsView extends React.PureComponent<Props, State> {
...
-      }
-    });
-    this.setModeState(
-      {
-        tracings: newTracings,
-      },
-      this.state.shouldShowArchivedTracings,
-    );
+    this.updateTracingInLocalState(tracing, (t) => update(t, { name: { $set: name } }));
     editAnnotation(tracing.id, tracing.typ, {
```
philippotto, line 382: could you also catch the error case here in case the update didn't work? i.e., show an error toast and restore the name again. maybe it makes sense to switch the order, too (first update on server, then update locally; however I'm not sure if this causes a flicker or something similar).

frontend/javascripts/oxalis/view/left-border-tabs/mapping_settings_view.tsx
```diff
@@ -170,17 +185,18 @@ class MappingSettingsView extends React.Component<Props, State> {
...
-      this.props.mapping &&
-      this.props.hideUnmappedIds != null;
-    const isDisabled = this.props.isEditableMappingActive || this.props.isMappingLocked;
-    const disabledMessage = this.props.isEditableMappingActive
-      ? "The mapping has been edited through proofreading actions and can no longer be disabled or changed."
-      : this.props.mapping
-        ? "This mapping has been locked to this annotation, because the segmentation was modified while it was active. It can no longer be disabled or changed."
-        : "The segmentation was modified while no mapping was active. To ensure a consistent state, mappings can no longer be enabled.";
+      (shouldMappingBeEnabled || isMergerModeEnabled) && mapping && hideUnmappedIds != null;
+    // TODO: Ask whether changing the mapping should be disabled when the annotation is locked.
```
philippotto, line 191: i would say it should not be allowed, because the active mapping is stored in the annotation. changing it would require storing that update (and that is not allowed).

frontend/javascripts/navbar.tsx
```diff
@@ -859,6 +883,15 @@ function Navbar({
         />,
       );
     }
+    if (isLockedByUser) {
+      trailingNavItems.push(
+        <AnnotationLockedByOwnerTag
```
philippotto, line 888: maybe the new property should be called `lockedByOwner`? it's a bit confusing, that the code here has the logic:
```
!isLockedByUser => <LockedByUser />
isLockedByuser => <LockedByOwner />
```

frontend/javascripts/oxalis/view/action-bar/tracing_actions_view.tsx
```diff
@@ -449,6 +453,15 @@ class TracingActionsView extends React.PureComponent<Props, State> {
     });
   };
 
+  handleUnlockAnnotation = async () => {
+    const { annotationId, annotationType, annotationTags } = this.props;
+    await editLockedState(annotationId, annotationType, false);
+    await editAnnotation(annotationId, annotationType, {
+      tags: annotationTags.filter((tag) => tag !== LOCKED_TAG),
+    });
+    location.reload();
```
philippotto, line 462: maybe show a toast before reloading?

frontend/javascripts/oxalis/view/action-bar/tracing_actions_view.tsx
philippotto: yes, this is good! this has annoyed me for some time now :)

philippotto (commented):
nice, didn't test yet, but the code looks good :+1:

philippotto:
Testing went smooth with only two exceptions:

- I'm not sure whether we really want to re-use the tags in the dashboard for "locked". for skeleton, volume and hybrids, we also have these special tags, but I'd rather get rid of that instead of extending these. the reason is mostly that this blocks these names for user-provided tags and it can be hard to understand that these are "automatic". what do you think, @MichaelBuessemeyer @fm3 ? apart from that: when clicking on an "locked" tag, the tag appears in the search bar but not in the correct color.
- even though "archive" is disabled in the dashboard for a locked annotation, I can still click it and get an error toast. ideally, nothing should happen when clicking on an disabled button/link.

MIGRATIONS.unreleased.md
fm3: I agree, please have a look in Migrations.released.md to see how we listed the sql migrations.

conf/messages
fm3: I think this is what the current implementation does, yes. We should discuss it. Would you prefer if admins were allowed to lock/unlock other people’s annotations? I would vote for it. cc @normanrz

fm3:
Regarding the tags, I think we will want to clean those up soon (enable key-value pairs, etc). I agree that the existing “automatic” tags are confusing and we should get rid of them. However, this is at least not getting much worse by this PR, because we’ll have to deal with that anyway. But still, if we could find another way to display this locked state, I agree that it would be preferrable. Related: #7814

fm3 (commented):
Backend looking good :) I just added two comments

conf/messages
philippotto: > Would you prefer if admins were allowed to lock/unlock other people’s annotations?

I think, I'm against that. For the same reason that only the owner can mutate an annotation.. If we should decide to allow admins this, then, the interface should make it very clear that one is locking/unlocking somebody else's annotation.

fm3: Fine by me too. Restricting it to the owner certainly is easier for the moment

frontend/javascripts/dashboard/explorative_annotations_view.tsx
MichaelBuessemeyer: I decided to first update locally so the user does not get confused why the name not changed successful when having something of a connection bottleneck. In case of an error I shown an error toast and revert back to the previous name (so the UI should by in sync with the data saved by the backend)

conf/messages
normanrz: So, if a user leaves the team, an admin would need to copy the annotation to make it editable again?

MichaelBuessemeyer: > So, if a user leaves the team, an admin would need to copy the annotation to make it editable again?

In your described scenario -> yes

frontend/javascripts/dashboard/explorative_annotations_view.tsx
MichaelBuessemeyer: Nvm this contradicts how I implemented `setLockedState`. Thus, I first update the backend and if that worked out, update the view in the frontend

MichaelBuessemeyer:
Ok I removed the "locked" tag and instead added a hint to the id column in case the annotation is locked: 
image


This is similar to read-only tracings which are marked / highlighted in the same way.

> even though "archive" is disabled in the dashboard for a locked annotation, I can still click it and get an error toast. ideally, nothing should happen when clicking on a disabled button/link.

Good find: I adjusted the `AsyncLink` to do nothing in case it is disabled.

app/models/annotation/Annotation.scala
```diff
@@ -322,6 +322,43 @@ class AnnotationDAO @Inject()(sqlClient: SqlClient, annotationLayerDAO: Annotati
...
+    val ownerId = <<[ObjectId]
+    val ownerFirstName = <<[String]
+    val ownerLastName = <<[String]
+    val othersMayEdit = <<[Boolean]
+    val teamIds = parseArrayLiteral(<<[String]).map(ObjectId(_))
+    val teamNames = parseArrayLiteral(<<[String])
+    val teamOrganizationIds = parseArrayLiteral(<<[String]).map(ObjectId(_))
+    val modified = <<[Instant]
+    val tags = parseArrayLiteral(<<[String]).toSet
+    val state = AnnotationState.fromString(<<[String]).getOrElse(AnnotationState.Active)
```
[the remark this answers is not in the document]
MichaelBuessemeyer: i tried bug failed :see_no_evil::

I changed to 
```scala
 val state = <<[AnnotationState]
    val isLockedByOwner = <<[Boolean]
    val dataSetName = <<[String]
    val typ = <<[AnnotationType]
    val visibility = <<[AnnotationVisibility]
```
But scala could not find the matching `GetResult` "converters". Thus, I added them as implicit params:
```scala

  implicit def GetResultAnnotationCompactInfo(implicit
                                                e0: GetResult[AnnotationState],
                                                e1: GetResult[AnnotationType],
                                                e2: GetResult[AnnotationVisibility]
                                             ): GetResult[AnnotationCompactInfo] = GetResult { prs =>
```
But then the call `run(query.as[AnnotationCompactInfo])` claims it needs to define the implicit params ` GetResult[AnnotationState]`

=> I'd say no to the question / suggestion or we need to define the `GetResult`s somewhere. But I honestly don't know how to do that :see_no_evil:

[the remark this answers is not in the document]
fm3: OK, thanks for experimenting! Then let’s do this another time. Feel free to go via the string route as is :)

conf/messages
MichaelBuessemeyer: After we agreed upon this, this pr should (and its potential changes) this pr should be ready for the 2nd review round

frontend/javascripts/oxalis/model/reducers/reducer_helpers.ts
```diff
@@ -108,7 +108,7 @@ export function convertServerAnnotationToFrontendAnnotation(annotation: APIAnnot
     description,
     name,
     annotationType,
-    isLockedByUser,
+    isLockedByOwner: isLockedByOwner,
```
philippotto, line 111:
```suggestion
    isLockedByOwner,
```

philippotto (approved):
front-end looks good to me :+1: it would be fine with me to merge this with the current "only the owner may change the locking" behavior. we will see whether we need to change this in the future.

conf/messages
MichaelBuessemeyer: I just talked with @fm3 about this pending decision: -> Should we decided that admins can also unlock/lock an annotation, this can be done as a follow-up pr :)

MichaelBuessemeyer:
@philippotto Yesterday I got feedback from @valentin-pinkau that locking from within the annotation view would also be very useful / essential as vx produced links directly open an annotation and the one can just lock the annotation from there instead of first having to navigate to the annotations table, locate the annotation and lock it there. Thus, I added the following changes:
- The option to unlock an annotation from within the navbar dropdown menu was extended to also lock the active annotation in case the current user in the owner of the annotation.
- The option was moved down to be the last item in the dropdown list to make it less "present" as this is likely not the most important feature of the dropdown menu -> Thus, let's not put it into prime (first) position in the list.
- I added a short waiting time between the successful update to the locked annotation state and the reloading of the annotation view to give the UI some time to render the toast properly and give the user the potential time to read the toast to understand why wk is reloading.
- The toast is now a success toast, as the update to the "lockedState" was successful.
- I edited the changelog entry to hint at the new navbar dropdown menu item.

@philippotto Could you please give this a quick recheck :pray:?

philippotto:
Great, looks and feels good :+1: Just one thing:

image

Can you add something like this to this tooltip: "You can unlock the annotation in the annotation menu." (only for the owner obviously). Otherwise, it could be confusing and users might think they cannot unlock it.

PR #7810, opened 2024-05-16
Trim firstName, lastName in the register form

### LINKED ISSUE #7804: Why do so many users have spaces in their names?
I see a lot of users with spaces in their names (at last position, usually). Is there a common reason there? Where should we add a trim?

Currently 46 users

Example: https://scm.slack.com/archives/CMBMU5684/p1715772199662139

fm3:
My theory is that people started typing their full name in the register form’s »first name« field, and then removed their last name but not the space when they saw that there is a separate last name field. Let’s just trim the form field content and see if that helps for new users.

### Steps to test:
- set features.isWkOrgInstance=true in application.conf
- register new account, type spaces around your firstName
- should be trimmed in the resulting user name + organization name.

### Question
 - Should we repair existing users + organization display names? My assumption would be that it doesn’t hurt, as nothing should reference users directly by name. What do you think?

### Issues:
- fixes #7804

------


philippotto:
Thanks for doing this :pray: I pushed a change which adds the trimming in the other (non wk) registration form, too. My other attempt to use `normalize` didn't work well, since the user couldn't even type a space at the end of a name field..

> Should we repair existing users + organization display names? My assumption would be that it doesn’t hurt, as nothing should reference users directly by name. What do you think?

Yes, I think so, too :+1:

hotzenklotz (approved):
LGTM

PR #7813, opened 2024-05-21
Bump frontend deps

MichaelBuessemeyer:
This PR updates some outdated frontend dependencies and replaces the markdown library with react-markdown. Moreover, prop types, esbuild loader, protobuffjs, and onnx web runtime are updated.

In order to fix the frontend tests, the new markdown library needs to be loaded lazily. Thus, I created the `markdown_adapter.tsx` file.

- https://bumpfrontenddeps.webknossos.xyz/

### Steps to test:
- Check whether wk works on a rudimentary level.
- Check whether markdown is rendered correctly in: (Already tested by me but re-testing some of them might be helpful)
  - the publication gallery
  - the annotation description in the info tab an annotation
  - task list view of a user
  - new task info modal (shown upon opening a task)
  - In the annotation view: Hovering the burger menu on the left to an annotation title should show the description

### TODOs:
- [x] Test whether the current markdown rendering matches the previous settings `options={{
                    html: false,
                    breaks: true,
                    linkify: true,
                  }`

### Issues:
- No issues exist for this

------
- ~~[ ] Updated [changelog](../blob/master/CHANGELOG.unreleased.md#unreleased)~~ I'd say it is not needed as this is not user-facing.


package.json
```diff
@@ -41,6 +42,7 @@
     "dpdm": "^3.14.0",
     "enzyme": "^3.7.0",
     "enzyme-adapter-react-16": "^1.9.1",
+    "esbuild": "^0.16.10",
```
philippotto, line 45: can we bump this to 0.20? since we use `"esbuild-loader": "^4.1.0",`, it should be possible [apparently](https://github.com/privatenumber/esbuild-loader/releases).

webpack.config.js
```diff
@@ -83,18 +77,21 @@ module.exports = function (env = {}) {
           ],
         },
         {
-          test: /\.tsx?$/,
```
philippotto, line 86: why did you split tsx? into ts and tsx? the config looks the almost same or is there a difference because of the `loader` attribute?

webpack.config.js
```diff
@@ -166,6 +169,11 @@ module.exports = function (env = {}) {
     },
     optimization: {
       minimize: env.production,
+      minimizer: [
+        new EsbuildPlugin({
+          target: buildTarget, // Syntax to transpile to (see options below for possible values)
```
philippotto, line 174:
```suggestion
          target: buildTarget,
```

philippotto: did you sanity check the size of the js modules in the devtools (master vs this branch)?

frontend/javascripts/libs/lazy_loader.tsx
```diff
@@ -0,0 +1,14 @@
+import * as React from "react";
+
+export default function loadable<Props>(
+  loader: () => Promise<{ default: React.ComponentType<Props> }>,
+) {
+  const InternalComponent = React.lazy(loader) as any;
+  return function AsyncComponent(_props: Props) {
```
philippotto, line 7: why the `_` prefix? since the var is used, it can simply be `props`, right?

philippotto (changes requested):
awesome! the changes look very good. only left some smaller comments. will test once these are addressed.

package.json
MichaelBuessemeyer: Yeah, I initally also upgraded to the 0.20 version but during the debug session regarding the tests we tried a lot of ways to fix this and this included downgrading to the previously used esbuild version. Now I forget to test whether it is ok to bump esbuild up. 
Thanks for being so nitpicky :D => The 0.20 version seems to be working alright

philippotto (approved):
awesome :1st_place_medal:

PR #7816, opened 2024-05-22
Fix dragging annotation into a different orga’s dataset

### LINKED ISSUE #7681: Uploading Annotation into dataset should also include overwritingOrganizationName
Uploaded an NML file with no organization name in the NML file, into the dataset view mode

This sends an overwritingDatasetName with the request to steer the NML into the open dataset and not into the dataset that was specified in the NML.

This failed, since the organization id is looked up by only the dataset name (comment says `Disambiguation method for legacy URLs and NMLs: if the user has access to multiple datasets of the same name, use the oldest. This is reasonable, because the legacy URL/NML was likely created before this ambiguity became possible`).

This resulted in the wrong organization’s dataset of that name. Instead, the frontend should also send an overwritingOrganizationName when uploading into an existing dataset, and the backend should use that.

Corresponding libs issue to always write out an organization name: https://github.com/scalableminds/webknossos-libs/issues/1015

fm3:
When drag’n’dropping an annotation into a dataset in view-mode, the annotation is uploaded into that dataset, regardless of what dataset is reference din the NML tags.

This PR now makes this available if the organization name has changed as well.

- https://uploadannotationorgaspecific.webknossos.xyz

### Steps to test:
- Create annotation, download
- In the NML file, edit the organization name in the experiment tag to some non-existing orga, or remove it completely
- Uploading that annotation by dragging it into a dataset view mode should still work, and create an annotation for that dataset.

### Issues:
- fixes #7681

------


app/models/annotation/nml/NmlParser.scala
```diff
@@ -62,7 +63,7 @@ object NmlParser extends LazyLogging with ProtoGeometryImplicits with ColorGener
         _ <- TreeValidator.validateTrees(treesSplit, treeGroupsAfterSplit, branchPoints, comments)
         additionalAxisProtos <- parseAdditionalAxes(parameters \ "additionalAxes")
         datasetName = overwritingDatasetName.getOrElse(parseDatasetName(parameters \ "experiment"))
-        organizationName = if (overwritingDatasetName.isDefined) None
+        organizationName = if (overwritingDatasetName.isDefined) overwritingOrganizationName
         else parseOrganizationName(parameters \ "experiment")
```
MichaelBuessemeyer, line 67: Shouldn't the if-check check for `overwritingOrganizationName` being defined and not for `overwritingDatasetName`? Or is there some other place where it is ensured that in case `overwritingDatasetName` is defined, then `overwritingOrganizationName` is also defined?

And maybe the same syntax as in the line above can be used here as well, as imo the code is more readable this way. Or is there a semantic difference between my suggestion and your code?
```suggestion
        organizationName = overwritingOrganizationName.getOrElse(parseOrganizationName(parameters \ "experiment"))
```

MichaelBuessemeyer (approved):
Thanks for this improvement :pray: 

I only found one potential improvement of the changes. But as I am unsure whether my suggestion keeps the same functionality, feel free to ignore it and merge

MichaelBuessemeyer (changes requested):
Please wait before merging: I am running into CSP errors in Chrome (version 125.0.6422.60) & Ubuntu:
image

Let's tackle this in this PR as well, as the zip uploading in an annotation does not work on the master as well (with the chrome version)

app/models/annotation/nml/NmlParser.scala
fm3: Yes, there is in fact a difference. I want the overwritingOrganizationName, even if it is None, if the overwritingDatasetName is defined.

This case will probably not happen, because the frontend sends both or neither. But from the backend point of view, the organizationName from the NML becomes invalid once the datasetName is overwritten (this was already the case before this PR). So in that moment I want to yield the None, so that it will later be filled by the disambiguation logic via dataset name, and not the organizationName from the NML.

MichaelBuessemeyer (approved):
Found a CSP fix :tada:

PR #7818, opened 2024-05-22
Clean up BucketProviders + DatasetArray handles

### LINKED ISSUE #7107: De-duplicate code in DatasetArray-based BucketProvider classes
I feel like many BucketProvider classes share a lot of code. This should be made DRYer by introducing a trait. DatasetArrayBucketProvider?

### LINKED ISSUE #7566: Clean Up ShardHandleCache
After #7528 the wkw shards are no longer opened with memory mapping, so they don’t need to be closed when they are removed from the cache. We can probably use a normal ALFU cache there now. (Maybe LRUConcurrentCache can go entirely?)

related: #6969

fm3:
Reducing code dupliaction in the logic that loads buckets (32³ voxels) from on-disk image data.

The ShardHandleCache is no longer needed, because we do not open individual shard files with memory mapping anymore. It was a relic of the old WKW loading implementation removed in https://github.com/scalableminds/webknossos/pull/7528. The DatasetArrays can be used directly now.

Also, the individual BucketProviders could be unified into a single DatasetArrayBucketProvider (which is still different from the VolumeTracingBucketProvider and a few other tracing-related ones), with a match/case statement dealing with the differences.

Removes CubePosition class (was almost unused, also since we changed the WKW loading implementation)

I also fixed a minor bug, where the VaultPath cache was not cleared when reloading WKW layers.

- https://cleanupbucketproviders.webknossos.xyz

### Steps to test:
- [x] Load data from WKW,Zarr, should look ok
- [x] Test remote dataset (tested neuroglancer, zarr & zarr-ngff)
- [x] Load volume annotation data, should look ok,
- [x] Hit Reload in the left sidebar, data should reload.

### Issues:
- fixes #7566
- fixes #7107

------


webknossos-datastore/app/com/scalableminds/webknossos/datastore/services/DataSourceService.scala
```diff
@@ -251,6 +254,7 @@ class DataSourceService @Inject()(
           case layer: PrecomputedLayer => Some(layer.mags)
           case layer: ZarrLayer        => Some(layer.mags)
           case layer: Zarr3Layer       => Some(layer.mags)
+          case layer: WKWLayer         => Some(layer.mags)
```
MichaelBuessemeyer, line 257: Maybe this match case statement is not needed anymore. :thinking: 
Needs investigation :mag:

MichaelBuessemeyer (changes requested):

MichaelBuessemeyer (approved):
Awesome thanks for applying the feedback. Testing worked out perfectly :+1:

PR #7819, opened 2024-05-22
Refactor Skeletontree tab as Antd  <Tree /> component

### LINKED ISSUE #5241: Shift select a range in tree tab
## Detailed Description
Use shift for selection of ranges, like in a file explorer

## Context
Sometimes you need to select a lot of trees.

### LINKED ISSUE #5872: Use antd Tree in skeleton view
When the skeleton view was built we couldn't use the antd Tree component due to performance reasons, because it was not virtualized. However, with v4 the Tree component became virtualized and we could switch to it now.

Personally, I see the following benefits:
- More uniform design
- https://github.com/frontend-collective/react-sortable-tree appears to be effectively abandoned
- Drag-and-dropping is rather clunky currently (+ has problems in Firefox) and appears to work smoother in antd

hotzenklotz:
In an effort to use the same antd `<Tree />` tree component for the skeletons, comments, and segments, I refactored the skeleton tab to use this component.

- https://treetree.webknossos.xyz/

### Steps to test:
- Do something outside the skeleton view so that it must update (select different tree, node, etc)
- Rename skeletons
- Drag and drop skeletons
- Try context menu stuff
  - color changes
  - expand/collapse everything
- Try really large skeleton
- Try single select
- Try multi select with CTRL/meta or SHIFT

### TODOs:
- [x] Drag & Drop
- [x] Auto-Scrolling
- [x] Styling
- [x] Multi-Select
- [x] Search
- [x] Expand All / Expand Subgroup
- [x] Fix `---` label on hover
- [x] Fix repeated SHIFT selection
- [x] Hide drag handler icon
- [x] Fix crash for "delete tree"
- [x] Empty groups should not show checkbox
- [x] Fix drag and drop moving too many trees
- [ ] (Maybe) fix search behavior

### Issues:
- Contributes to #7765 
- Fixes #5872 
- Fixes #5241 

------


MichaelBuessemeyer:
Here is my pre-review testing feedback:
The tree items are too spacy:
This is the current master version which does not line break the tree names and thus is horizontally scrollable but more dense
image
And this is your version:
image
I do prefer the dense (master) version


Rearranging trees within a group does no longer work on your branch.

Empty tree groups are opened by default on the master when: 
1. creating a new group and 
3. moving some trees into them. 

On your branch one needs to expand the group as step 3.

On the master, the level intend seems to be smaller (due to the missing drag handle). Is it possible to make this a little narrower to make the view more compact?



These are the thing I noticed during testing. So far it looks very good beside this few points :rocket:

MichaelBuessemeyer (commented):
Please just re-request a review once its time :pray:  :)

frontend/javascripts/oxalis/model/accessors/skeletontracing_accessor.ts
```diff
@@ -35,6 +35,7 @@ export function getSkeletonTracing(tracing: Tracing): Maybe<SkeletonTracing> {
 
   return Maybe.Nothing();
 }
+
```
hotzenklotz, line 38: Added some white space for readability.

frontend/javascripts/oxalis/shaders/utils.glsl.ts
```diff
@@ -113,10 +113,12 @@ export function jsColormapJet(x: number): Vector3 {
 
   return [r, g, b];
 }
-export const hslaToCSS = (hsla: Vector4) => {
-  const [h, s, l, a] = hsla;
-  return `hsla(${360 * h}, ${100 * s}%, ${100 * l}%, ${a})`;
+
+export const rgbaToCSS = (rgba: Vector4) => {
```
hotzenklotz, line 117: We save color values as RGBA, not HSLA. There is no reason to convert between the color models just for some CSS.

frontend/javascripts/oxalis/view/right-border-tabs/tree_hierarchy_view.tsx
```diff
@@ -61,255 +64,242 @@ import { formatNumberToLength, formatLengthAsVx } from "libs/format_utils";
...
-    },
-    groupTree: [],
-    prevProps: null,
-    searchFocusOffset: 0,
-    activeTreeDropdownId: null,
-    activeGroupDropdownId: null,
-  };
-
-  static getDerivedStateFromProps(nextProps: Props, prevState: State) {
-    if (prevState.prevProps == null || didTreeDataChange(prevState.prevProps, nextProps)) {
```
hotzenklotz, line 131: `didTreeDataChange` change has been replaces with `React.memo(<TreeHierarchyView/>)` at the very bottom of this file.

frontend/javascripts/oxalis/view/right-border-tabs/tree_hierarchy_view.tsx
```diff
@@ -516,354 +482,275 @@ class TreeHierarchyView extends React.PureComponent<Props, State> {
 
     // Make sure the displayed name is not empty
     const displayableName = name.trim() || "<Unnamed Group>";
+    return (
```
hotzenklotz, line 485: The `<Dropdown>` + menu moved here from old method `generateNodeProps()`

frontend/javascripts/oxalis/view/right-border-tabs/tree_hierarchy_view_helpers.ts
```diff
@@ -97,31 +89,25 @@ export function insertTreesAndTransform(
       treeNode.children, // Groups that don't contain any trees should not influence the state of their parents
       (groupOrTree) => groupOrTree.isChecked || !groupOrTree.containsTrees,
     );
-    treeNode.isIndeterminate = treeNode.isChecked
```
hotzenklotz, line 100: The `indeterminate` state is not no longer directly controlled by use. Antd takes care of that.

MichaelBuessemeyer (changes requested):
Overall everything looks very good :tada:. Thanks for implementing this.

I only found minor things. -> Please see my comments.

I additionally noticed, that when hovering a group or node a tooltip with the content `---` is shown. I was unable to identify what leads to this behaviour. Could you please investigate this and remove / hide this strange tooltip?

frontend/javascripts/oxalis/view/right-border-tabs/tree_hierarchy_view.tsx
```diff
@@ -448,17 +416,16 @@ class TreeHierarchyView extends React.PureComponent<Props, State> {
...
           disabled: isEditingDisabled,
-          onClick: () => this.deleteGroup(id),
+          onClick: () => deleteGroup(id),
           icon: <DeleteOutlined />,
           label: "Delete group",
         },
         hasSubgroup
           ? {
               key: "collapseSubgroups",
-              disabled: !hasExpandedSubgroup,
```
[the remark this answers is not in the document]
hotzenklotz: Checking for subgroups on every render call is expensive. There is no real benefit from disabling these options. Having the ability to expand/collapse all groups is always a good feature regardless if does anything on selection.

frontend/javascripts/oxalis/view/right-border-tabs/tree_hierarchy_view.tsx
```diff
@@ -61,255 +64,242 @@ import { formatNumberToLength, formatLengthAsVx } from "libs/format_utils";
...
+      (node) => node.key,
+    );
 
-    this.setState((prevState) => ({
-      groupTree: collapseAllGroups(prevState.groupTree),
-      expandedGroupIds: newExpandedGroupIds,
-    }));
-  };
+    if (expanded) {
+      setExpandedNodeKeys([...expandedNodeKeys, selectedGroupIdKey, ...subGroupKeys]);
```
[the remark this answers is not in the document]
hotzenklotz: Done

philippotto:
Really looking forward to see this merged! I tested this PR for a bit and found some things:

- [ ] the scroll bar is not visible in dark mode (because the background color of the scroll bar is too dark?). in light mode, it works fine
- [ ] when selecting a tree, it is always scrolled to be at the top of the view (maybe both points can be solved with `scrollTo({..., align: "auto"})`?
  - [ ] when manually selecting an item from the list, this is very annoying, because the tree item was perfectly visible, anyway.
  - [ ] when selecting a tree/node in the data viewport, this auto-scroll behavior also always enforces the item to be at the top of the view (if the list is long enough). however, when the item was visible anyway, it's not necessary to scroll at all. 

won't fix?
- if a tree name is very long, you cannot scroll horizontally. apparently, this is a limitation of antd. so probably a won't fix. the full name can be seen via the tooltip or the rename-field.

philippotto:
Hmm, I found another thing:

- [ ] shift select often doesn't work. could it be that it performs the range by sorting alphabetically? however, the visible sorting might be by creation time. in the screenshot, the blue item was active and I shift-clicked on the gray item. nothing happens in that case

image

MichaelBuessemeyer:
Please wait until the backend changes of this pr have been reviewed (they haven't been reviewed yet. See: https://github.com/scalableminds/webknossos/pull/7939#issuecomment-2269370736)

@normanrz Could you take over looking over these changes? There shouldn't be many changes.

MichaelBuessemeyer (approved):

package.json
```diff
@@ -206,7 +205,6 @@
     "react-redux": "^7.2.0",
     "react-router-dom": "^5.1.2",
     "react-sortable-hoc": "^2.0.0",
-    "react-sortable-tree": "^2.8.0",
```
daniel-wer, line 209: The day has finally come :tada:

PR #7820, opened 2024-05-23
Align sections job

### LINKED ISSUE #3536: Persist organization name to NMLs?
If you have read access to multiple datasets of the same name (across organizations, e.g. public datasets), uploading NMLs is ambiguous. As of #3533, we use the disambiguation method which returns the older dataset. This is not always going to be correct, though! So I guess the orga name should also be written to NML

Tobias314:
### Steps to test:
- Checkout the [corresponding vocxelytics PR](https://github.com/scalableminds/voxelytics/pull/3542) and start the worker. Afterwards, open a dataset in you local Webknossos and select alignment from the AI tools

### Issues:
- fixes [#3536](https://github.com/scalableminds/voxelytics/issues/3536)

------
- [X] Considered [common edge cases](https://github.com/scalableminds/webknossos/blob/master/.github/common_edge_cases.md)


Tobias314:
> Where does the align_example.jpg come from? Did you double-check that we may publish this data?

The pictures can also be found on our website ([https://scalableminds.com/alignment-service](https://scalableminds.com/alignment-service)) so that should not be a problem. 

> Do the left and right side really point to the same data? (The blood vessel on the right is so bright, I would have expected to see something of that in the non-aligned version too).

It is not the same picture on the left and on the right but rather one dataset with the left side showing the left section of the dataset in the unaligned state and the right side showing the right section of the dataset in the aligned state.

fm3:
> The pictures can also be found on our website so that should not be a problem.

Ok, yes these were checked :)

> It is not the same picture on the left and on the right but rather one dataset with the left side showing the left section of the dataset in the unaligned state and the right side showing the right section of the dataset in the aligned state.

Fair enough. I would have expected a before→after image, of the same crop, but it’s not very important. Feel free to leave it as is for the moment, maybe we’ll insert a more refined image later.

frontend/javascripts/oxalis/view/action-bar/starting_job_modals.tsx
```diff
@@ -324,7 +327,7 @@ export function OutputSegmentationLayerNameFormItem({
 
 export function StartAIJobModal({ aIJobModalState }: StartAIJobModalProps) {
   const onClose = () => Store.dispatch(setAIJobModalStateAction("invisible"));
-  const isSuperUser = Store.getState().activeUser?.isSuperUser || false;
+  const isSuperUser = useSelector((state: OxalisState) => state.activeUser?.isSuperUser || false);
```
MichaelBuessemeyer, line 330: Awesome, thanks :tada:

frontend/javascripts/oxalis/view/action-bar/starting_job_modals.tsx
```diff
@@ -446,12 +450,33 @@ function RunAiModelTab({ aIJobModalState }: { aIJobModalState: string }) {
                 </Card>
               </Radio.Button>
             </Tooltip>
+            <Tooltip title="Coming soon">
+              <Radio.Button
+                className="aIJobSelection"
+                checked={aIJobModalState === "align_sections"}
+                disabled={!Store.getState().activeUser?.isSuperUser}
```
MichaelBuessemeyer, line 457: Here is still a access to the `isSuperUser` field of the active user, although it was already selected above. You can just use the selected `isSuperUser` variable.
```suggestion
                disabled={!isSuperUser}
```

frontend/javascripts/oxalis/view/action-bar/starting_job_modals.tsx
```diff
@@ -446,12 +450,33 @@ function RunAiModelTab({ aIJobModalState }: { aIJobModalState: string }) {
...
+                onClick={() => Store.dispatch(setAIJobModalStateAction("align_sections"))}
```
MichaelBuessemeyer, line 458: And here is still a`Store.dispatch`
```suggestion
                onClick={() => dispatch(setAIJobModalStateAction("align_sections"))}
```

frontend/javascripts/oxalis/view/action-bar/starting_job_modals.tsx
```diff
@@ -446,12 +450,33 @@ function RunAiModelTab({ aIJobModalState }: { aIJobModalState: string }) {
...
+              >
+                <Card bordered={false}>
+                  <Space direction="vertical" size="small">
+                    <Row className="ai-job-title">Align Sections</Row>
+                    <Row>
+                      <img
+                        src={`/assets/images/${jobNameToImagePath.align_sections}`}
+                        alt={"Mitochondria detection example"}
```
MichaelBuessemeyer, line 466: This should be renamed to something like
```suggestion
                        alt={"Example of improved alignment of slices"}
```

MichaelBuessemeyer (changes requested):
Thanks for applying my suggestions :t-rex: 

I found a few other things, which I did not notice in the first review round. Please also fix them. Then the code should be fine :)

MichaelBuessemeyer (approved):
The frontend code looks good to me :tada:

PR #7822, opened 2024-05-23
Persist mappingName and mappingIsLocked in NML

fm3:
- https://mappingnml.webknossos.xyz

### Steps to test:
- Create annotation on dataset with agglomerate mapping, brush some (mapping will be locked)
- download + reupload
- Mapping of fallback layer should still be selected and locked
- Also test download+reupload with mapping but without volume brushing (mapping should be set but not locked)
- Also test without mapping (download+reupload should still run through without error)
- Beware of unrelated bug https://github.com/scalableminds/webknossos/issues/7823 – if your test dataset has a segment index file, you might have to remove that before you can test this PR

### Issues:
- fixes #7822

------


MichaelBuessemeyer:
The code lgtm :tada: 

> Create annotation on dataset with agglomerate mapping, brush some (mapping will be locked)
download + reupload
Mapping of fallback layer should still be selected and locked

This worked, but now I have a green dataset bounding box :thinking:
image

> Also test without mapping (download+reupload should still run through without error)

I created a new annotation, added a skeleton and manipulated the volume without a mapping enabled. This locks the annotation to the state, that no mapping is selected. But when I download this annotation and reupload it, the fact that the state of "no mapping is selected" is locked is no longer present and I can simply activate a mapping. I would have thought that this would need to be persisted as well & also applied when reuploading.

fm3:
> This worked, but now I have a green dataset bounding box 🤔 => I created an https://github.com/scalableminds/webknossos/issues/7831 for this.

Thanks! I commented in that issue, I tend to think it’s unrelated to the mappings, but we should investigate that.

> manipulated the volume without a mapping enabled. This locks the annotation to the state, that no mapping is selected

Good catch! I hadn’t considered this case. Added it now!

MichaelBuessemeyer (approved):
Tröt :tada: 

Thanks for fixing this :pray:

PR #7837, opened 2024-05-28
Better error message in case of duplicate edge during min-cut

fm3:
Before, the »null« exception would just bubble up. Now, we get a proper error chain including a message about the root cause.


valentin-pinkau (approved):
Thank you for improving the error message! LGTM!

PR #7838, opened 2024-05-28
Fix NML download with skipVolumeData

fm3:
The writeComment for the `skipVolumeData` hint closed the current tag. That means that if we want to add more attributes to that tag, we have to do that first.

Regression from #7822


valentin-pinkau (approved):
Merci!

PR #7847, opened 2024-05-31
Fix lock explorative annotations

MichaelBuessemeyer:
- https://___.webknossos.xyz

### Steps to test:
- as Non admin user create an annotation
- log in as admin sample user. In the explorative annotations table the annotation of the non admin user should appear without the option to lock this annotation
- as the non admin user, lock the annotation
- the annotation should also appear locked in the admin's annotation table
- no weird newlines between action items should be visible

### Issues:
- No issue exists for this. Follow up of #7801

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -77,3 +77,6 @@ coverage-ts
 temp-webknossos-schema**
 conf/application.conf-e
 .env
+.bloop
+.metals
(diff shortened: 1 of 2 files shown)


hotzenklotz (approved):
Works like a charm. Thank you.

PR #7849, opened 2024-06-03
Follow up fix for lock annotations

MichaelBuessemeyer:
Currently, it is not possible to modify user bounding boxes in the boudning box tab due to an inverted logic bug.
This pr fixes this and ensures all pending annotation updates are sent to the server updating an annotations locked state from within the annotations view.

- https://___.webknossos.xyz

### Steps to test:
- Open a new annotation on some dataset
- The boudning box tab should not be disabled
- Fast create some bounding boxes and then lock the annotation via the menu in the nav bar. This should happen before the annotation view reached a saved state.
- The locking should first push all pending updates to the server (the sandclock icon should turn into a checkmark) and then update the locked state, show a toast and reload the webpage
- The annotation state should now be "locked"

### Issues:
- No issue exists for this


fm3 (approved):
Code LGTM and test went well :) :shipit:

PR #7850, opened 2024-06-03
Reenable viewing mappings in viewmode after lock annotations pr

MichaelBuessemeyer:
The lock annotations pr accidentally included a change disabling the ability to view mappings in view only mode of a dataset.
This PR fixes this issue and renames a variable

### Steps to test:
- View a dataset with a mapping in view only mode. It should be possible to select and view a mapping in this view only mode.

### Issues:
- fixes No issue exists for this


fm3 (approved):
Works for me :)

PR #7854, opened 2024-06-03
Show wk user in vx workflow list and enable cache busting for bundle chunks

### LINKED ISSUE #7735: Show WK user name next to owner in voxelytics workflows list

dieknolle3333:
New attempt to merge old PR #7794 after deployment didnt show the right changes. We found out that the JS bundles are chunked, and the chunk that holds the changes of this PR did not include any cache busting mechanism. 

Changes beside old PR:
- include commit hash in name of webpack entry points (because scala doesnt know content hash)
    - introduce new library, that does not contain any dependencies itself
- include content hash in name of chunked bundles to introduce cache busting 

- https://backupshowwkuserinvxworkflowlist.webknossos.xyz/workflows (not latest changes of branch, because it contains mock data. Includes cache busting fix.)

### Steps to test:
- go to workflow view and see wk users

### ToDos
- [x] explore whether double request can be avoided

### Issues:
- fixes #7735

------


yarn.lock
```diff
@@ -10882,7 +10887,16 @@ string-convert@^0.2.0:
...
+"string-width-cjs@npm:string-width@^4.2.0":
+  version "4.2.3"
+  resolved "https://registry.yarnpkg.com/string-width/-/string-width-4.2.3.tgz#269c7117d27b05ad2e536830a8ec895ef9c6d010"
+  integrity sha512-wKyQRQpjJ0sIp62ErSZdGsjMJWsap5oRNihHhu6G7JVO/9jIB6UyevL+tXuOqrng8j/cxKTWyWUwvSTriiZz/g==
+  dependencies:
+    emoji-regex "^8.0.0"
+    is-fullwidth-code-point "^3.0.0"
+    strip-ansi "^6.0.1"
+
+string-width@^4.0.0, string-width@^4.2.2, string-width@^4.2.3:
```
dieknolle3333, line 10899: we think this might be due to changes that happened during `yarn add` and that this shouldnt be a problem.

frontend/javascripts/admin/voxelytics/workflow_list_view.tsx
```diff
@@ -84,9 +89,18 @@ export default function WorkflowListView() {
     }
   }
 
-  usePolling(loadData, VX_POLLING_INTERVAL);
+  usePolling(async () => {
```
dieknolle3333, line 92: @fm3 I tried to address your comment about requesting the workflows twice while trying not to mess around in the code too much. I thought I could delete the `loadData` call where the search query is persisted, but then a dataless page is rendered. So I think this might be an acceptable solution for now, but I am keen to hear what you say :)

hotzenklotz (commented):
webpack changes look good to me.

hotzenklotz (approved):
LGTM

fm3 (commented):
Awesome, thanks for investigating the cache problems! Weird of the browsers that ctrl+shift+r cache clear apparently did not work for the sub-bundles…

I created a follow-up issue to properly clean up the pollingcode https://github.com/scalableminds/webknossos/issues/7862

Could you remove the mock json data, and expand the changelog to mention the cache fix? Then this should be good to go :)

PR #7856, opened 2024-06-04
Fix uploading zipped tiff (should set needsConversion=true)

### LINKED ISSUE #7855: uploading a zip with tiff files does not set needsConversion=true
I believe the frontend unpacks the zip and should inspect it. My zipfile only contained tif files so a conversion job should have been started.

Instead, the finishUpload job sends needsConversion=false, and then fails in the backend (and with a bad error message „head of empty list“).

The backend should yield a better error message, and the frontend should auto-detect that this needs a conversion job.

fm3:
Fixes two bugs in dataset upload
 - The frontend failed to set needsConversion=true in case of zipped data (e.g. a zip with tiffs), because the return statement exited not only the for loop but the whole function
 - The backend detected N5 data too eagerly (return value of `containsMatchingFile` was not used)

### Steps to test:
- `yarn enable-jobs` (no actual worker needed though)
- In dataset upload view, upload a zip with tiff files
- After the upload, a conversion job should be started
- Upload bare tiffs, upload wkw, these other cases should still behave as expected

### Issues:
- fixes #7855

------

### DIFF
--- a/frontend/javascripts/admin/dataset/dataset_upload_view.tsx
+++ b/frontend/javascripts/admin/dataset/dataset_upload_view.tsx
@@ -513,7 +513,7 @@ class DatasetUploadView extends React.Component<PropsWithFormAndRouter, State> {
         }
         // We return here since not more than 1 zip archive is supported anyway. This is guarded
         // against via form validation.
-        return;
+        break;
       }
     }
 
--- a/frontend/javascripts/admin/dataset/dataset_upload_view.tsx
+++ b/frontend/javascripts/admin/dataset/dataset_upload_view.tsx
@@ -511,8 +511,8 @@ class DatasetUploadView extends React.Component<PropsWithFormAndRouter, State> {
             zipFile: [],
           });
         }
-        // We return here since not more than 1 zip archive is supported anyway. This is guarded
-        // against via form validation.
+        // The loop breaks here in case of zip because at most one zip archive is supported anyway.
+        // Form validation takes care of that assertion.
         break;
       }
     }
--- a/webknossos-datastore/app/com/scalableminds/webknossos/datastore/services/uploading/UploadService.scala
+++ b/webknossos-datastore/app/com/scalableminds/webknossos/datastore/services/uploading/UploadService.scala
@@ -290,7 +290,8 @@ class UploadService @Inject()(dataSourceRepository: DataSourceRepository,
             } yield explored)
           .toList)
       combinedLayers = exploreLocalLayerService.makeLayerNamesUnique(dataSources.flatMap(_.dataLayers))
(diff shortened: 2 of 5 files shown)


frontend/javascripts/admin/dataset/dataset_upload_view.tsx
```diff
@@ -513,7 +513,7 @@ class DatasetUploadView extends React.Component<PropsWithFormAndRouter, State> {
         }
         // We return here since not more than 1 zip archive is supported anyway. This is guarded
```
hotzenklotz, line 514:
```suggestion
        // We break here since not more than 1 zip archive is supported anyway. This is guarded
```

hotzenklotz: Maybe rephrase entire comment

fm3: good point, done :)

webknossos-datastore/app/com/scalableminds/webknossos/datastore/services/uploading/UploadService.scala
```diff
@@ -290,7 +290,8 @@ class UploadService @Inject()(dataSourceRepository: DataSourceRepository,
             } yield explored)
           .toList)
       combinedLayers = exploreLocalLayerService.makeLayerNamesUnique(dataSources.flatMap(_.dataLayers))
-      dataSource = GenericDataSource[DataLayer](dataSourceId, combinedLayers, dataSources.head.scale)
+      firstExploredDatasource <- dataSources.headOption.toFox
+      dataSource = GenericDataSource[DataLayer](dataSourceId, combinedLayers, firstExploredDatasource.scale)
```
MichaelBuessemeyer, line 294: What is the advantage of this change?
To me the code semantically reads the same. Just that the testing for whether `dataSources` has atleast a one item is more explicit. Or am I misunderstanding something here?

MichaelBuessemeyer (approved):
Works :tada:

I tested all tiff formats included in: https://www.notion.so/scalableminds/Test-Datasets-c0563be9c4a4499dae4e16d9b2497cfb?pvs=4#7db7f26b5e494352a8153ab073487ab3 as well as wkw.
 

Just left one question :)

webknossos-datastore/app/com/scalableminds/webknossos/datastore/services/uploading/UploadService.scala
fm3: `head` on an empty list raises an exception, which bypasses some of our error handling. headOption.toFox creates a Fox.failure that will yield nicer error messages in our context.

PR #7752, opened 2024-04-15
Add Infer Mitochondria Job

fm3:
Adds infer_mitochondria worker job (for now for superusers only)

Corresponding PR in voxelytics: https://github.com/scalableminds/voxelytics/pull/3504

### Steps to test:
- Set up worker with vx branch `infer-mitochondria-workflow`, in wk, run `yarn enable-jobs`, open dataset/annotation in wk, create a bbox with a mito, hit “AI Analysis”, run mito inferral job, result should look ok

------
- [x] (no changelog for now, since this is still behind the superUser condition)
- [x] Updated [migration guide](../blob/master/MIGRATIONS.unreleased.md#unreleased) if applicable
- [x] Removed dev-only changes like prints and application.conf edits
- [x] Considered [common edge cases](../blob/master/.github/common_edge_cases.md)

PR #7857, opened 2024-06-04
Format infer_mitochondria job in job list view

fm3:
The job list view should be able to handle infer_mitochondria jobs being present. This new job command was introduced in #7752 .

Before this change, the list view crashed with an assertion error.

Note: the worker job in the current voxelytics master does not yet return the new dataset name, so for old jobs, the view button will not be available. However, I’ll change that in the worker in the coming days so new jobs will then have the view button (already tested that locally).

### Steps to test:
- I tested locally that the infer_mitos entry looks good in the job list view:
image

------


dieknolle3333 (approved):
Looks good to me, and local testing went well to! 

Side note: When I checked out master, the view didn't crash for me, it only showed a very basic description, but maybe there's a difference between running and pending jobs or something. In any way this is very useful :smiley:

Issue #7412, opened 2023-10-25
Extend dataset view configuration to include ID mapping


hotzenklotz:
## Detailed Description
Whenever we run an AI segmentation through the worker, the resulting dataset will be viewed without any mapping, i.e. the full oversegmentation including the chunking borders. This not a great user experience for new users who do not know how to manually select and other ID mapping.

I propose that WK configures the selected ID mapping to a default value, .e.g. 65, and adds it to the view configuration.

## Context

* [ ] Specific to long-running jobs (set `jobsEnabled=true` in `application.conf`)
* [ ] Specific to webknossos.org (set `isDemoInstance=true` in `application.conf`)

PR #7858, opened 2024-06-04
Mapping in default view config

MichaelBuessemeyer:
- https://___.webknossos.xyz

### Steps to test:
- Set default mapping from view mode! 
  - Open a dataset with at least one segmentation layer with a mapping. Enable that mapping and save the config as default.
  - Open the dataset settings. Navigate to the View Configuration Tab and check whether your settings were set as default appropriately in the layer settings.
- Set default mappings from dataset config view
  - Open the dataset settings and open the View Configuration Tab. 
  - In the layer settings configure a mapping by defining `"mapping: {"name": <name>, "type": <"JSON"|"HDF5">}` as a new property to a layer. In case the json is valid but the name & type combination cannot be found, this should be reported to the user and submitting the config should not work.
  - After successfully submitting, reload / reopen the View Configuration. The updated settings should be shown.
- As a new user (e.g. sample2@scm.io),  view the just updated dataset. The configured mapping should be active by default.
- Make the dataset public and open it in a private tab (somewhere where you are not logged in). The mapping should be active by default.

- Confirm the UI bottom margin fix in the layer settings view by viewing a dataset with multiple segmentation layers as an admin. Then disable the last segmentation layer. The button to save the config as the dataset default should have a proper distance to the last segmentation layer settings in any scenario.   

### TODOs:
- [x] Ensure loading when having an annotation 
- [x] consider moving the activeMappingByLayer to DatasetConfiguration.layer
- [x] ~~only show mapping selection for segmentation layers with mappings~~ No longer possible due to TODO directly above
- [x] open new issue for not initially rendering a mapping under when switching layer in an annotation  -> #7859
- [x] Allow saving "no mapping active" from view mode as default view configuration (currently the mapping is still kept) 
- [x] in settings view cache fetched available mappings & agglomerates
- [x] rename from `defaultMapping` to `mapping`

### Issues:
- fixes [#7412](https://github.com/scalableminds/webknossos/issues/7412)

------


frontend/javascripts/dashboard/dataset/dataset_settings_viewconfig_tab.tsx
```diff
@@ -14,24 +14,89 @@ import {
...
 
 const FormItem = Form.Item;
 
-export default function DatasetSettingsViewConfigTab() {
+export default function DatasetSettingsViewConfigTab(props: {
+  datasetId: APIDatasetId;
+  dataStoreURL: string | undefined;
+}) {
+  const { datasetId, dataStoreURL } = props;
+  const [availableMappingsPerLayerCache, setAvailableMappingsPerLayer] = useState(
```
hotzenklotz, line 41: `useState<T>` accepts a type hint.

```suggestion
  const [availableMappingsPerLayerCache, setAvailableMappingsPerLayer] = useState<Record<string, [string[], string[]]>>(
```

frontend/javascripts/dashboard/dataset/dataset_settings_viewconfig_tab.tsx
```diff
@@ -14,24 +14,89 @@ import {
...
+  );
+
+  // biome-ignore lint/correctness/useExhaustiveDependencies: Update is not necessary when availableMappingsPerLayer[layerName] changes.
+  const validateDefaultMappings = useMemo(
+    () => async (configStr: string, dataStoreURL: string, datasetId: APIDatasetId) => {
+      let config = {} as DatasetConfiguration["layers"];
+      try {
+        config = JSON.parse(configStr);
+      } catch (e: any) {
+        return Promise.reject(new Error("Invalid JSON format for : " + e.message));
```
hotzenklotz, line 52: This error would crash the whole WK app / UI page until the next React error boundary. Is this really necessary? Can we just apply some default instead and log the error to the console/a toast

frontend/javascripts/dashboard/dataset/dataset_settings_viewconfig_tab.tsx
```diff
@@ -14,24 +14,89 @@ import {
...
+        }
+        let jsonAndAgglomerateMappings = [[], []] as [string[], string[]];
+        try {
+          jsonAndAgglomerateMappings = await Promise.all([
+            getMappingsForDatasetLayer(dataStoreURL, datasetId, layerName),
+            getAgglomeratesForDatasetLayer(dataStoreURL, datasetId, layerName),
+          ]);
+        } catch (e: any) {
+          console.error(e);
+          throw new Error(messages["mapping.loading_failed"](layerName));
```
hotzenklotz, line 70: Same here.

frontend/javascripts/dashboard/dataset/dataset_settings_viewconfig_tab.tsx
```diff
@@ -14,24 +14,89 @@ import {
...
+            mappingType === "HDF5"
+              ? agglomeratesForLayer.some((agglomerate) => agglomerate === mappingName)
+              : mappingsForLayer.some((mapping) => mapping === mappingName);
+          return doesMappingExist
+            ? null
+            : `The default mapping "${mappingName}" of type "${mappingType}" does not exist for layer ${layerName}.`;
+        })
+        .filter((error) => error != null);
+      if (errors.length > 0) {
+        throw new Error("The following default mappings are invalid: " + errors.join("\n"));
```
hotzenklotz, line 94: Same here. Is this error even displayed somewhere?

frontend/javascripts/dashboard/dataset/dataset_settings_viewconfig_tab.tsx
```diff
@@ -50,23 +115,45 @@ export default function DatasetSettingsViewConfigTab() {
       dataIndex: "comment",
     },
   ];
-  const comments: Partial<Record<keyof DatasetLayerConfiguration, string>> = {
-    alpha: "20 for segmentation layer",
-    min: "Only for color layers",
-    max: "Only for color layers",
-    intensityRange: "Only for color layers",
+  const comments: Partial<Record<keyof DatasetLayerConfiguration, [string, string | null]>> = {
+    alpha: ["20 for segmentation layer", null],
```
hotzenklotz, line 119: Why are using arrays to hold the descriptions? Would it not be easier to read to use an object with names keys, e.g. `tooltip`, `shortDescription`?

frontend/javascripts/dashboard/dataset/dataset_settings_viewconfig_tab.tsx
```diff
@@ -50,23 +115,45 @@ export default function DatasetSettingsViewConfigTab() {
...
   };
   const layerViewConfigurationEntries = _.map(
     { ...getDefaultLayerViewConfiguration(), min: 0, max: 255, intensityRange: [0, 255] },
     (defaultValue: any, key: string) => {
       // @ts-ignore Typescript doesn't infer that key will be of type keyof DatasetLayerConfiguration
       const layerViewConfigurationKey: keyof DatasetLayerConfiguration = key;
       const name = layerViewConfigurations[layerViewConfigurationKey];
+      const commentParts = comments[layerViewConfigurationKey];
+      const comment =
+        commentParts != null && commentParts[1] != null ? (
```
hotzenklotz, line 145: Using named keys would make reading this easier. If you add a description for `alpha` then there is no need to check them for `null`.

frontend/javascripts/oxalis/view/left-border-tabs/layer_settings_tab.tsx
```diff
@@ -1338,14 +1342,42 @@ class DatasetSettings extends React.PureComponent<DatasetSettingsProps, State> {
...
+            ) {
+              updatedLayers[layerName] = {
+                ...updatedLayers[layerName],
+                mapping: {
+                  name: mappingInfo.mappingName,
+                  type: mappingInfo.mappingType,
+                },
+              };
+            } else {
+              updatedLayers[layerName] = {
```
hotzenklotz, line 1370: This is the default Store value anyway, right? No need to reassign it.

hotzenklotz (approved):
Overall the feature works as expected. 

Please see some of my code comments.

frontend/javascripts/dashboard/dataset/dataset_settings_viewconfig_tab.tsx
MichaelBuessemeyer: Yes, like the other errors / rejected promisses, this message is shown below the layer config input in the view configurations

MichaelBuessemeyer: image
(Note that I removed the "default" from the error message as the property name of the setting no longer contains default in its name. In a prior version of this branch the property was called `defaultMapping` instead of just `mapping`)

frontend/javascripts/dashboard/dataset/dataset_settings_viewconfig_tab.tsx
MichaelBuessemeyer: Sure, also fine. Might make thinks also easier to understand :tada:

frontend/javascripts/oxalis/view/left-border-tabs/layer_settings_tab.tsx
MichaelBuessemeyer: > This is the default Store value anyway, right?

I'd say no: I modifed the type of the DatasetConfiguration so that layers may have the mapping setting (which is null by default). But in case an admin sets a mapping for layer A, goes into viewmode, disables the mapping for layer A and than saves these changes, `datasetConfiguration.layers` should still contain a mapping for layer A as it was previously served this way by the backend.

=> I just tested this scenario with the debugger and was able to confirm that this case is necessery. Else one cannot deactivate a mapping previously activated by useing the "Save View Configuration as Default" feature.

MichaelBuessemeyer (commented):
Thanks for you review :pray:

Please find my comment below / attached to yours. 

Do you think this is mergable in the newest commit / with the newest changes?

hotzenklotz:
Let's go 🚀

PR #7869, opened 2024-06-06
In wkorg proxy, forward mimetype instead of static text/html

fm3:
We now also proxy js and css files etc, so the mimetype can’t always be text/html


valentin-pinkau (approved):
🚀

PR #7872, opened 2024-06-10
Don't show agglomerate zoom warning in case 3d viewport is maximized

### LINKED ISSUE #7865: suppress »agglomerate mapping not possible while zoomed out« toast when only 3d viewport is maximized
When showing mapped data, this toast warns that zooming out leads to the mapping no longer being rendered.

However, it should not appear when maximizing the 3d viewport (no segmentation data is shown then anyway – other than meshes – so this does not apply).

MichaelBuessemeyer:
This PR suppresses the `Segmentation data which is mapped using an agglomerate file cannot be rendered in this magnification. Please zoom in further.` warning in case the user only has the 3d viewport maximized. This bug was noticed on a mobile device but can easily be reproduced on a normal screen.

- https://___.webknossos.xyz

### Steps to test:
- Open a dataset with an agglomerate mapping and enable it.
- Zoom out until the described warning appears.
- Miximze the 3d viewport, the warning should be closed automatically and no longer appear.
- Un-maximiize the 3d viewport again, zoom out and the warning should appear again

### Additional Changes:
@hotzenklotz I also changed the fix regarding the cached webpack chunk files: https://scm.slack.com/archives/C5AKLAV0B/p1717429338824009
The fix included putting the latest git commit hash into the main chunk file's name. But this makes development quite tedious: When commit the hash changes, webpack takes notice of this and produces a different `main.<different_hash_then_preivous>.js` but scala does not notice this change and still references the old `main.<old_hash>.js` file. Which means in the current dev setup one needs to restart the backend server after each commit in order to load the newest js / CSS changes. Thus, I rolled back the part that included the hash in the file's name and reverted back to using a nocache query param. The chunk files which originally produced the problem of not getting cache busted, do still include their content hash in their name. Thus, the fix should still work and with my changes the local development should be back to its original comfort without having to restart the backend after each commit.  

### Issues:
- fixes #7865

------

### DIFF
--- a/frontend/javascripts/oxalis/model/sagas/annotation_saga.tsx
+++ b/frontend/javascripts/oxalis/model/sagas/annotation_saga.tsx
@@ -125,7 +125,7 @@ function shouldDisplaySegmentationData(): boolean {
 
   const segmentationLayerName = segmentationLayer.name;
   const isSegmentationLayerDisabled =
-    Store.getState().datasetConfiguration.layers[segmentationLayerName].isDisabled;
+    state.datasetConfiguration.layers[segmentationLayerName].isDisabled;
   if (isSegmentationLayerDisabled) {
     return false;
   }
--- a/frontend/javascripts/oxalis/model/sagas/annotation_saga.tsx
+++ b/frontend/javascripts/oxalis/model/sagas/annotation_saga.tsx
@@ -31,7 +31,7 @@ import {
   cancelled,
 } from "typed-redux-saga";
 import { select } from "oxalis/model/sagas/effect-generators";
-import { getMappingInfo } from "oxalis/model/accessors/dataset_accessor";
+import { getMappingInfo, is2dDataset } from "oxalis/model/accessors/dataset_accessor";
 import { getActiveMagIndexForLayer } from "oxalis/model/accessors/flycam_accessor";
 import { Model } from "oxalis/singletons";
 import Store from "oxalis/store";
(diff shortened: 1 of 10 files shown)


frontend/javascripts/oxalis/model/sagas/annotation_saga.tsx
```diff
@@ -122,7 +126,16 @@ function shouldDisplaySegmentationData(): boolean {
   const segmentationLayerName = segmentationLayer.name;
   const isSegmentationLayerDisabled =
     Store.getState().datasetConfiguration.layers[segmentationLayerName].isDisabled;
```
hotzenklotz, line 128:
```suggestion
    state.datasetConfiguration.layers[segmentationLayerName].isDisabled;
```

frontend/javascripts/oxalis/view/layouting/flex_layout_helper.ts
```diff
@@ -68,3 +69,20 @@ export function getPositionStatusOf(tabSetNode: TabSetNode): NodePositionStatus
...
+// searching for the td viewport component in the json config.
+export function is3dViewportMaximized(modelConfig: ModelConfig) {
+  return modelConfig.layout.children.some((child) => {
+    return (
+      child.type === "row" &&
+      child.children.some((rowChild) => {
+        return (
+          rowChild.type === "tabset" &&
+          rowChild.maximized &&
+          rowChild.children?.[0]?.id === "TDView"
```
hotzenklotz, line 83: 🙈

hotzenklotz (approved):
LGTM

frontend/javascripts/oxalis/model/sagas/annotation_saga.tsx
MichaelBuessemeyer: Thanks for finding this :pray:

Issue #7830, opened 2024-05-24
Rethink how segment names are assigned when merging two segments


Albane-dlv:
We should rethink how segment names are assigned when merging two segments. 
- e.g., if one segment has a user-specified name, always pick that (does this make sense?). 
- or: always ask the user? 
- or: show a toast which allows to recover the other name, but does not force the user. 
- or: always concatenate the names.

PR #7877, opened 2024-06-12
Combine custom names of merged agglomerates & transfer custom name to split-off segments

MichaelBuessemeyer:
This PR improves the way custom names of segments / agglomerates are handled during proofreading. Previously, the potential custom name of the target segment was dropped in a merge action and a split-off segment always "lost" its previous custom name. Now, the custom names are merged in case of a merge and split-off segments inherit the custom name of the original agglomerate.

- https://___.webknossos.xyz

### Steps to test:
- Open a dataset with a segmentation layer with a hdf5 mapping.
- load 2 segments & give each a custom name
- merge these segments. The newly merged segment should have both custom names separated by a comma.
- Merge two segments out of which only one has a custom name. The merged segment should have retained the custom name but not the default one.
- Split of a segment from an agglomerate with a custom name. The split-off segment should have the same name as the original agglomerate.   
- Split a segment from all its neighbors. All new split off neighbors should have the custom name of the original segment in case the original segment had a custom name.

### TODOs:
- [x] Check whether there are other proofreading actions that should be considered.

### Issues:
- fixes [#7830](https://github.com/scalableminds/webknossos/issues/7830)

------


CHANGELOG.unreleased.md
```diff
@@ -11,6 +11,7 @@ For upgrade instructions, please check the [migration guide](MIGRATIONS.released
 [Commits](https://github.com/scalableminds/webknossos/compare/24.06.0...HEAD)
 
 ### Added
+- Added that proofreading actions no longer drop custom names. A merge action now combines the potenial existing custom names of both segments and a split-action transfers the custom name to the split-off segment. [#7877](https://github.com/scalableminds/webknossos/pull/7877)
```
```suggestion
- Added that proofreading merge actions reuse custom names of segments. A merge action now combines the potenial existing custom names of both segments and a split-action copies the custom name to the split-off segment. [#7877](https://github.com/scalableminds/webknossos/pull/7877)
```

philippotto (approved):
code lgtm :+1: didn't test because I assume that you tested the mentioned proofreading operations well.

MichaelBuessemeyer:
just did another quick test on the dev instance to avoid a big accidental woopsie. Testing worked :+1:

PR #7879, opened 2024-06-13
Start Job Modal Improvements

fm3:
`_corrected` doesn’t make sense for jobs that don’t process a layer, but rather create a new one.

Also, the model dropdown select was too narrow so that the full name of the model could not be read.


MichaelBuessemeyer (approved):
Work purrfectly :cat: :cat2:

PR #7883, opened 2024-06-14
Validate animation job bounding box

hotzenklotz:
This PR adds another validation for the animation modal. It checks that the user selected bounding box size is larger then 0.

### Steps to test:
- Enable jobs
- Create an annotation with contain a BB with depth, height or width set to 0
- Launch animation modal
- Select your BB
- There should be a warning preventing you from starting the animation jon

### Issues:
- fixes https://scm.slack.com/archives/CMBMU5684/p1718354190366309

------

### DIFF
--- a/frontend/javascripts/oxalis/view/action-bar/create_animation_modal.tsx
+++ b/frontend/javascripts/oxalis/view/action-bar/create_animation_modal.tsx
@@ -50,14 +50,14 @@ const MAX_MESHES_PER_ANIMATION = 40; // arbitrary limit to not overload the serv
 
 function selectMagForTextureCreation(
   colorLayer: APIDataLayer,
-  boundingBox: BoundingBoxType,
+  boundingBox: BoundingBox,
 ): [Vector3, number] {
   // Utility method to determine the best mag in relation to the dataset size to create the textures in the worker job
   // We aim to create textures with a rough length/height of 2000px (aka target_video_frame_size)
   const colorLayerBB = new BoundingBox(
     computeBoundingBoxFromBoundingBoxObject(colorLayer.boundingBox),
   );
-  const bb = new BoundingBox(boundingBox).intersectedWith(colorLayerBB);
+  const bb = boundingBox.intersectedWith(colorLayerBB);
 
   const longestSide = Math.max(...bb.getSize());
   const dimensionLongestSide = bb.getSize().indexOf(longestSide);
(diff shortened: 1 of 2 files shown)


frontend/javascripts/oxalis/view/action-bar/create_animation_modal.tsx
```diff
@@ -167,7 +168,14 @@ function CreateAnimationModal(props: Props) {
         `You selected too many meshes for the animation. Please keep the number of meshes below ${MAX_MESHES_PER_ANIMATION} to create an animation.`,
       );
 
-    const validationStatus = hasEnoughMags && isDtypeSupported && isDataset3D && !isTooManyMeshes;
+    const isBoundingBoxEmpty = boundingBox.getVolume() === 0;
```
MichaelBuessemeyer, line 171: But here the BBox is not intersected with the color layer.

-> a color layer of bbox (0,0,0,10,10,10) and a bbox of (10,10,10,10,10,10) would also lead to zero volume. But this is not properly checked here.

MichaelBuessemeyer: Or does such a scenario not result in an error?
I would at least say, that this would be an unwanted scenario, no part of the color layer would be rendered

MichaelBuessemeyer (changes requested):
Your fix works :+1:

But please consider my suggestion: I idenfied a scenario in which case nothing of the color layer would be rendered but no warning would be shown.

I'd say to also fix this scenario before merging. I would pull the interect opertion out of the 'selectMagForTextureCreation`  method and already pass the intersected bbox to the funciton

frontend/javascripts/oxalis/view/action-bar/create_animation_modal.tsx
```diff
@@ -50,14 +50,14 @@ const MAX_MESHES_PER_ANIMATION = 40; // arbitrary limit to not overload the serv
...
-  boundingBox: BoundingBoxType,
+  boundingBox: BoundingBox,
 ): [Vector3, number] {
   // Utility method to determine the best mag in relation to the dataset size to create the textures in the worker job
   // We aim to create textures with a rough length/height of 2000px (aka target_video_frame_size)
   const colorLayerBB = new BoundingBox(
     computeBoundingBoxFromBoundingBoxObject(colorLayer.boundingBox),
   );
-  const bb = new BoundingBox(boundingBox).intersectedWith(colorLayerBB);
+  const bb = boundingBox.intersectedWith(colorLayerBB);
```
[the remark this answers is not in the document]
hotzenklotz: yes

MichaelBuessemeyer (approved):
splendid :shipit:

PR #7884, opened 2024-06-17
default state wording to indicate loading

### LINKED ISSUE #7863: Rename default dataset & organization name of default store state to something like "loading"
## Detailed Description

Currently, a new dataset viewing tab shows "Test Dataset | Connectomics Department" as the default tab title until the dataset is properly stored in the frontend store. This may be confusing and should be replaced with something indicating that the annotation view is still initializing ... :)

fm3:
Changes the default state dataset to values that shouldn’t confuse the user if they show up in the tab title briefly.

fixes #7863

However, I can’t reproduce this error. The function getTabTitle in tracing_layout_view looks like this:
```javascript

  getTabTitle = () => {
    const getDescriptors = () => {
      switch (this.state.status) {
        case "loading":
          return ["Loading"];

        case "failedLoading":
          return ["Error"];

        default:
          return [this.props.displayName, this.props.organization];
      }
    };

    const titleArray: Array<string> = [...getDescriptors(), "WEBKNOSSOS"];
    return titleArray.filter((elem) => elem).join(" | ");
  };
```

For my tests, this seems to work, I never get the defaultState values. Are there other states than loading and failedLoading where the props are not properly set yet? Do we still have the original report for this, do we know how to reproduce?


MichaelBuessemeyer (approved):
LGTM & works :dancers: 

> Do we still have the original report for this, do we know how to reproduce?

Sure: In the annotations table right click a view link and open in another tab. Keep the tab in the background. Without your fix it should switch from "Loading" to "Test Dataset" to the actual dataset name. With your fix it only shows "Loading" before switching to the actual dataset name.

fm3:
Thanks, turns out this only happens in Chrome, not Firefox (at least for me). Apparently, Chrome holds back more requests for tabs opened in the background. Ok, I’ll merge this then :) Thanks for taking care of the snapshot test!

Issue #5968, opened 2022-01-19
User-editable dataset and folder metadata


normanrz:
## Detailed Description
Add user-editable metada for datasets and folders. 
* The metadata should be in the form of `Dict<str, str | int | Array<str>>`.
* The existing `dataset.details` can be reused for this.
* Dataset tags should move into this metadata structure as `"tag": Array<str>` key-value pair.

This needs to be implemented in backend and frontend. We'll probably need a migration to migrate the tags. In the frontend there should be a key-value editor for datasets (in dataset settings) and folders (in the edit modal).

The API should also be exposed to be used in the wklibs to read/write metadata.

PR #7886, opened 2024-06-17
Folder And Dataset Metadata

MichaelBuessemeyer:
- https://folderdatasetproperties.webknossos.xyz/

### Steps to test:
- Open the dashboard & create another folder.
- Try adding metadata to the folder. Try out all three kinds of entries. Test whether autosave works
- Try editing metadata for a dataset
- Switch between datasets, the metadata should stay consistent
- Switch between folders, the metadata should stay consistent
- And also switch between folders and datasets and also look for any errors / inconsistencies / lost metadata entries.

### TODOs:
- [x] Clean up git history 
- [x] Allow to search for metadata props -> Other issue https://github.com/scalableminds/webknossos/issues/8034
- [x] Add props & UI and everything for folders 
- [x] Update UI to the newest design proposal
- [x] Clean up code
- [x] Test dataset update route without supplying the metadata field 
  - [x] Works. Tested with firefox's edit and resend functionality. 

### Open Question
- [x] What about supporting dataset tags? Should they be removed? Should one still be able to update the dataset tags via the API? Should they be removed from the schema? Should they be transferred into metadata entries?
- [x] How should tags be transferred into metadata?
  - As a single string array entry with key "tags"
  - For each on entry with numbered keys like "tagX" and x is the index of the tag in the tag array
  - other ideas 

### Issues:
- fixes [#5968](https://github.com/scalableminds/webknossos/issues/5968)

------


MichaelBuessemeyer:
For the record:

During development / testing, I noticed an error with the antd Select component used in the new MetadataTable. Looks like this is this one of the related issues: https://github.com/ant-design/ant-design/issues/46348. I just gave it a quick look. The error might be gone after a proper antd & react update :)

image

hotzenklotz:
Things I noticed during testing:
- [ ] The type `a,b` always has an empty tag as a default
  Screenshot 2024-07-17 at 11 23 30

- [ ] No matter which dataset I selected from my dashboard they all had the same values 
  Screenshot 2024-07-17 at 11 27 29


Some UI/UX feedback:
- [ ] The colon between "property" and "value" is not centered. Better yet, we can remove it. 
- [ ] The properties are always displayed as text boxes even when one does not want to edit them. Text boxes are not great for readability. Maybe make the inputs `borderless` when they are not actively being edited (https://ant.design/components/input#components-input-demo-variant)
- [ ] Similar to the comment above, there is very little space in general. Maybe the "type" dropdown" and the "delete" button don't always have to be there. They are only relevant when working/editing an entry. Similar to other UI elements in WK every entry can enter an edit mode where all of that stuff is displayed. When not in editing mode, then the properties are just displayed as text. To avoid a pencil icon, one could trigger the edit mode when clicking on the text similiar to the antd `<Typography.Text>` component (https://ant.design/components/typography#components-typography-demo-editable). 
- [ ] Have a look a the antd `<Space.Compact>` component for clean compact input fields (https://ant.design/components/space#space-demo-compact). e.g. 
  Screenshot 2024-07-17 at 11 57 18
- [ ] The "types" field is only relevant for the values. All keys are always strings anyway. Therefore the "type" dropdown should can be right next to the "value" input and need not be the first entry in the table. (see comment directly above)
- [ ] I am not a big fan of the "plus" button to add new entry. Especially the empty state looks weird with just the table header and button. Maybe always show and empty to row, ready to edit?
- [ ] I wonder if the "types" options could be better explained through icons. Especially the multiple values/tags thing is not obvious. Antd has several options that might make sense, e.g. `<TagsOutlined />` 
  Screenshot 2024-07-17 at 11 38 27 Screenshot 2024-07-17 at 11 40 52

MichaelBuessemeyer:
>  The type a,b always has an empty tag as a default

Thanks, I remove fix that

> No matter which dataset I selected from my dashboard they all had the same values

Afaik, these are default added values for a new dataset. I can change that in the backend. You can change them individually for each dataset and the should be shown and saved individually.

>The colon between "property" and "value" is not centered. Better yet, we can remove it.

The colon was explicitly requested to be added. I centered it now.

> The properties are always displayed as text boxes even when one does not want to edit them. Text boxes are not great for readability. Maybe make the inputs borderless when they are not actively being edited (https://ant.design/components/input#components-input-demo-variant)

The inputs were already borderless. But I now also made the background transparent when the entry is not being edited.

>  Similar to the comment above, there is very little space in general. Maybe the "type" dropdown" and the "delete" button don't always have to be there. They are only relevant when working/editing an entry. Similar to other UI elements in WK every entry can enter an edit mode where all of that stuff is displayed. When not in editing mode, then the properties are just displayed as text. To avoid a pencil icon, one could trigger the edit mode when clicking on the text similiar to the antd <Typography.Text> component (https://ant.design/components/typography#components-typography-demo-editable).
> Have a look a the antd <Space.Compact> component for clean compact input fields (https://ant.design/components/space#space-demo-compact). e.g.

Thanks something to be discussed (I'd say as I got a different design proposal).

> The "types" field is only relevant for the values. All keys are always strings anyway. Therefore the "type" dropdown should can be right next to the "value" input and need not be the first entry in the table. (see comment directly above)

The whole type dropdown was removed to save some space in the table. Users can now select a type when creating a new entry and that type will no longer be changable after creation.

> I am not a big fan of the "plus" button to add new entry. Especially the empty state looks weird with just the table header and button. Maybe always show and empty to row, ready to edit?


Initially there always was an empty row. But it was decided against this. See https://scm.slack.com/archives/C5AKLAV0B/p1719923640762759 and https://scm.slack.com/archives/C5AKLAV0B/p1718896223081789

> I wonder if the "types" options could be better explained through icons. Especially the multiple values/tags thing is not obvious. Antd has several options that might make sense, e.g. <TagsOutlined />

Thanks, I did know about these icons. I added them to the newest version.

app/models/dataset/DatasetService.scala
```diff
@@ -77,8 +77,15 @@ class DatasetService @Inject()(organizationDAO: OrganizationDAO,
   ): Fox[Dataset] = {
     implicit val ctx: DBAccessContext = GlobalAccessContext
     val newId = ObjectId.generate
-    val details =
-      Json.obj("species" -> "species name", "brainRegion" -> "brain region", "acquisition" -> "acquisition method")
+    val metadata =
+      if (publication.isDefined)
+        Json.arr(
+          Json.obj("type" -> "string", "key" -> "species", "value" -> "species name", "index" -> 0),
```
fm3, line 83: Why is the "index" needed? If the data model is an array, the ordering should already be fixed, right?

CHANGELOG.unreleased.md
```diff
@@ -12,8 +12,9 @@ For upgrade instructions, please check the [migration guide](MIGRATIONS.released
 
 ### Added
 - Added the option to move a bounding box via dragging while pressing ctrl / meta. [#7892](https://github.com/scalableminds/webknossos/pull/7892)
+- Added the option to add metadata entries to dataasets and folders. The metadata can be viewed and edited in the dashboard in the right details tab.[#7886](https://github.com/scalableminds/webknossos/pull/7886)
```
fm3, line 15: Maybe rename the PR to also use the word “Metadata” instead of Properties? (Or update the code to also use Properties) I’d like to introduce only one new term, not two :sweat_smile:

app/models/folder/Folder.scala
```diff
@@ -133,10 +137,16 @@ class FolderDAO @Inject()(sqlClient: SqlClient)(implicit ec: ExecutionContext)
   protected def isDeletedColumn(x: Folders): Rep[Boolean] = x.isdeleted
 
   protected def parse(r: FoldersRow): Fox[Folder] =
-    Fox.successful(Folder(ObjectId(r._Id), r.name))
+    for {
+      metadata <- JsonHelper.parseAndValidateJson[JsArray](r.metadata.getOrElse("[]")).toFox
```
fm3, line 141: this might already be worth a one-liner helper function `private def parseMetadata(literal: Option[String]): JsArray)` :)

fm3 (commented):
Backend looking good already. Let’s decide on a terminology (metadata vs properties).

Also, since the dataset json changes in the API (details field is replaced by metadata field with different structure), we need to take special care. This PR adapts the frontend. The python libs don’t use the details field as far as I can see. @normanrz My guess is that the publications rendering code from the about page of webknossos needs to be adapted. Do you think that would be straightforward? Also do you know if external collaborators rely on the details field in the dataset json? If so, we would need to bump the API version.

normanrz:
> My guess is that the publications rendering code from the about page of webknossos needs to be adapted. Do you think that would be straightforward?

Yes, I can take of this once this PR is deployed.

app/models/dataset/DatasetService.scala
MichaelBuessemeyer: > I could remove the index and always re-add it in the frontend 🤔

Its now implemented that way

MichaelBuessemeyer:
@philippotto I prepared the code ready for your feedback and testing :D.

Moreover, I just removed an unnecessary dependency of a `useEffect` hook introduced by me in another pr which was just merged today.

MichaelBuessemeyer:
@markbader Metadata should now be updateable via the normal dataset update route

philippotto (commented):
haven't looked at the code yet, but testing results are in :)

image

- [x] the table head ("Property") and the validation warning are not aligned with the rest of the table entries. strictly speaking, the input box is also a bit too wide, but since the text within that box is aligned, it feels good.
- [ ] the trash icon is incorrectly positioned when a validation warning is there
- [ ] auto-save also happens when a field is focussed. after auto-saving that focus is not there anymore. it's probably easiest to not auto-save as long as a field is focussed.

frontend/javascripts/dashboard/advanced_dataset/dataset_table.tsx
```diff
@@ -387,7 +366,7 @@ class FolderRenderer {
 class DatasetTable extends React.PureComponent<Props, State> {
   state: State = {
     sortedInfo: {
-      columnKey: useLruRank ? "" : "created",
+      columnKey: useLruRank ? undefined : "created",
```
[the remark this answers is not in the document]
philippotto: hmm, good catch. it's definitely a bit fishy.

the logic should be as follows:
- if the table is sorted by a certain column (columnKey is properly defined), let antd sort the table accordingly (I think, this happens automatically)
- if not:
  - if no search is active: sort by lastUsedByUser
  - if a search is active: sort by dice coefficient

can you doublecheck that it works now correctly?

> although this should be done by the backend nowadays, or am I wrong?

no, I think, the frontend should do the sorting.

frontend/javascripts/dashboard/dataset/queries.tsx
```diff
@@ -324,6 +324,7 @@ export function useUpdateFolderMutation() {
               ? {
                   ...updatedFolder,
                   parent: oldFolder.parent,
+                  metadata: oldFolder.metadata,
```
philippotto, line 327: I wonder whether these two lines should be above `...updatedFolder` so that they don't have precedence. however, moving a folder and updating the metadata seem to work currently :thinking:

frontend/javascripts/dashboard/folders/metadata_table.tsx
```diff
@@ -0,0 +1,414 @@
...
+  context: DatasetCollectionContextValue,
+  { dontUpdateState }: EffectCancelSignal = { dontUpdateState: false }, // equals whether the component is unmounted
+  setIsSaving: (isSaving: boolean) => void,
+  setHasUnsavedChanges: (hasUnsavedChanges: boolean) => void,
+) => {
+  const hasAnyErrors = metadata.some((entry) => entry.error != null);
+  if (hasAnyErrors) {
+    return;
+  }
+  !dontUpdateState && setIsSaving(true);
```
philippotto, line 64: normal if would be more idiomatic

frontend/javascripts/dashboard/folders/metadata_table.tsx
```diff
@@ -0,0 +1,414 @@
...
+        allowedTeams: folder.allowedTeams?.map((t) => t.id) || [],
+        metadata: metadataWithoutIndexAndError,
+      });
+    }
+    if (!_.isEqual(serverResponse.metadata, metadataWithoutIndexAndError)) {
+      Toast.error(
+        `Failed to save metadata changes for ${datasetOrFolderString} ${datasetOrFolderToUpdate.name}.`,
+      );
+    } else {
+      !dontUpdateState && setHasUnsavedChanges(false);
```
philippotto, line 91: same here

frontend/javascripts/dashboard/folders/metadata_table.tsx
```diff
@@ -0,0 +1,414 @@
...
+    }
+  } catch (error) {
+    Toast.error(
+      `Failed to save metadata changes for ${datasetOrFolderString} ${datasetOrFolderToUpdate.name}. Please look in the console for more details.`,
+    );
+    console.error(error);
+  } finally {
+    !dontUpdateState && setIsSaving(false);
```
philippotto, line 99: same here

frontend/javascripts/dashboard/folders/metadata_table.tsx
```diff
@@ -0,0 +1,414 @@
...
+};
+
+const saveCurrentMetadataDebounced = _.debounce(saveCurrentMetadata, 3000);
+
+const getKeyInputId = (record: APIMetadataWithIndexAndError) =>
+  `metadata-key-input-id-${record.index}`;
+
+type APIMetadataWithIndexAndError = APIMetadata & { index: number; error?: string | null };
+type IndexedMetadataEntries = APIMetadataWithIndexAndError[];
+type EffectCancelSignal = { dontUpdateState: boolean };
```
philippotto, line 110: I'd move the types to the top of the module

frontend/javascripts/dashboard/folders/metadata_table.tsx
```diff
@@ -0,0 +1,414 @@
...
+  const previousDatasetOrFolder: APIDataset | Folder | undefined =
+    usePreviousValue(datasetOrFolder);
+  const [metadata, setMetadata] = useState<IndexedMetadataEntries>(
+    datasetOrFolder?.metadata?.map((entry, index) => ({ ...entry, index, error: null })) || [],
+  );
+  const [focusedRow, setFocusedRow] = useState<number | null>(null);
+  const [isSaving, setIsSaving] = useState<boolean>(false);
+  const [hasUnsavedChanges, setHasUnsavedChanges] = useState<boolean>(false);
+
+  const isFirstRenderingCycle = previousDatasetOrFolder == null;
```
philippotto, line 128: can be moved into effect. I'd probably even inline it.

frontend/javascripts/dashboard/folders/metadata_table.tsx
```diff
@@ -0,0 +1,414 @@
...
+        previousDatasetOrFolder,
+        metadata,
+        context,
+        effectCancelSignal,
+        setIsSaving,
+        setHasUnsavedChanges,
+      );
+    }
+    setMetadata(
+      datasetOrFolder.metadata?.map((entry, index) => ({ ...entry, index, error: null })) || [],
```
philippotto, line 153: maybe extract into helper since it is needed multiple times?

frontend/javascripts/dashboard/folders/metadata_table.tsx
```diff
@@ -0,0 +1,414 @@
...
+  );
+  const [focusedRow, setFocusedRow] = useState<number | null>(null);
+  const [isSaving, setIsSaving] = useState<boolean>(false);
+  const [hasUnsavedChanges, setHasUnsavedChanges] = useState<boolean>(false);
+
+  const isFirstRenderingCycle = previousDatasetOrFolder == null;
+
+  // Always update local state when folder / dataset changes. Prior to that send pending updates.
+  // biome-ignore lint/correctness/useExhaustiveDependencies: Only execute this hook when underlying datasetOrFolder changes.
+  useEffect(() => {
```
philippotto, line 132: not sure whether I understand it correctly: this metadata table is not remounted when activating another dataset/folder item in the list, right? therefore, you check whether the item changed in this effect. if yes, you flush the pending updates. is this correct? I think, it would be cleaner to unmount and remount the component when the dataset/folder changed. that way, the component's state is perfectly isolated (right now; I'd be a bit afraid that the metadata slips into the other dataset/folder in edge cases). the "use previous" code here wouldn't be necessary, either, I think.
you could achieve this easily by passing a `key` like this `<MetadataTable key={"${item.name}-${item.type}"} />`.
or was there a reason for not doing this?

frontend/javascripts/dashboard/folders/metadata_table.tsx
```diff
@@ -0,0 +1,414 @@
...
+        </span>
+      );
+  }
+}
+
+const saveCurrentMetadata = async (
+  datasetOrFolderToUpdate: APIDataset | Folder,
+  metadata: IndexedMetadataEntries,
+  context: DatasetCollectionContextValue,
+  { dontUpdateState }: EffectCancelSignal = { dontUpdateState: false }, // equals whether the component is unmounted
```
philippotto, line 56: I'd rename dontUpdateState to `isComponentMounted` and add the comment that this is necessary to avoid the set-functions if it's not.
another idea: you could even wrap the set functions in the other component and adapt them so that they don't do anything if its not mounted. I think that would be cool because the wrapping would only happen once. currently, you need to define and update this signal once per useEffect definition.

frontend/javascripts/dashboard/folders/metadata_table.tsx
```diff
@@ -0,0 +1,414 @@
...
+    }
+    return () => {
+      effectCancelSignal.dontUpdateState = true;
+    };
+  }, [metadata]);
+
+  const updateMetadataKey = (index: number, newPropName: string) => {
+    setMetadata((prev: IndexedMetadataEntries) => {
+      let error = null;
+      const entry = prev.find((prop) => prop.index === index);
```
philippotto, line 182: why isn't this equivalent to `prev[index]`?

frontend/javascripts/dashboard/folders/metadata_table.tsx
```diff
@@ -0,0 +1,414 @@
...
+      if (!entry) {
+        return prev;
+      }
+      const maybeAlreadyExistingEntry = prev.find((prop) => prop.key === newPropName);
+      if (maybeAlreadyExistingEntry) {
+        error = `Property ${newPropName} already exists.`;
+      }
+      if (newPropName === "") {
```
philippotto, line 190: since both set the error I'd turn this into an if-else.

frontend/javascripts/dashboard/folders/metadata_table.tsx
```diff
@@ -0,0 +1,414 @@
...
+      const detailsWithoutEditedEntry = prev.filter((prop) => prop.index !== index);
+      setHasUnsavedChanges(true);
+      return [
+        ...detailsWithoutEditedEntry,
+        {
+          ...entry,
+          key: newPropName,
+          error,
+        },
+      ];
```
philippotto, line 202: is there a reason for moving the item to the back? intuitively, I would have mapped the items and only update the one with the requested index (even though I'm still not sure I get the relevance of the index; isn't it redundant?).

frontend/javascripts/dashboard/folders/metadata_table.tsx
```diff
@@ -0,0 +1,414 @@
...
+  const updateMetadataValue = (index: number, newValue: number | string | string[]) => {
+    setMetadata((prev) => {
+      const entry = prev.find((prop) => prop.index === index);
+      if (!entry) {
+        return prev;
+      }
+      const updatedEntry = { ...entry, value: newValue };
+      const detailsWithoutEditedEntry = prev.filter((prop) => prop.index !== index);
+      setHasUnsavedChanges(true);
+      return [...detailsWithoutEditedEntry, updatedEntry];
```
philippotto, line 215: same here

frontend/javascripts/dashboard/folders/metadata_table.tsx
```diff
@@ -0,0 +1,414 @@
...
+    const isFocused = record.index === focusedRow;
+    const sharedProps = {
+      className: isFocused ? undefined : "transparent-input",
+      onFocus: () => setFocusedRow(record.index),
+      onBlur: () => setFocusedRow(null),
+      placeholder: "Value",
+      size: "small" as InputNumberProps<number>["size"],
+      disabled: isSaving,
+    };
+    switch (record.type) {
```
philippotto, line 295: the switch block could be a good candidate for an own functional component.

frontend/javascripts/dashboard/folders/metadata_table.tsx
```diff
@@ -0,0 +1,414 @@
...
+                placement="bottom"
+                trigger={["click"]}
+                autoFocus
+              >
+                <Button icon={<PlusOutlined style={{ marginLeft: -2 }} />}>
+                  Add First Metadata Entry
+                </Button>
+              </Dropdown>
+            </span>
+          </div>
```
philippotto, line 409: own component? :)

philippotto (changes requested):
I reviewed the code now. Cool stuff :) I left a couple of nitpicks. on a higher level I have two things (I commented both with more detail).

1. I don't really get the index property of the metadata items. wouldn't it work without it?
2. I'd enforce a remount when the user clicks on another folder/dataset, unless there is a strong reason against it.

it's probably best to discuss this in person before writing long walls of text :) :call_me_hand:

MichaelBuessemeyer (commented):
Thanks for all the feedback @philippotto 

Everything should be incorporated now. Please have another look :)

frontend/javascripts/dashboard/folders/details_sidebar.tsx
```diff
@@ -282,7 +287,7 @@ function FolderDetails({
           <div style={{ marginBottom: 4 }}>
             <FolderTeamTags folder={folder} />
           </div>
-          <MetadataTable datasetOrFolder={folder} />
+          <MetadataTable datasetOrFolder={folder} key={`${folder.id}#folder`} />
```
philippotto, line 290: please add a comment that the key is crucial to enforce remounts when the id changes.

frontend/javascripts/dashboard/folders/details_sidebar.tsx
```diff
@@ -177,7 +177,12 @@ function DatasetDetails({ selectedDataset }: { selectedDataset: APIDatasetCompac
             </Tag>
           )}
         </div>
-        {fullDataset && <MetadataTable datasetOrFolder={fullDataset} />}
+        {fullDataset && (
+          <MetadataTable
+            datasetOrFolder={fullDataset}
+            key={`${fullDataset.dataSource.id.name}#dataset`}
```
philippotto, line 183: see below

frontend/javascripts/dashboard/folders/metadata_table.tsx
```diff
@@ -49,22 +52,106 @@ function getMetadataTypeLabel(type: APIMetadata["type"]) {
...
+      return null;
+  }
+};
+
 const saveCurrentMetadata = async (
   datasetOrFolderToUpdate: APIDataset | Folder,
   metadata: IndexedMetadataEntries,
   context: DatasetCollectionContextValue,
-  { dontUpdateState }: EffectCancelSignal = { dontUpdateState: false }, // equals whether the component is unmounted
+  // When the component is no longer mounted, state updates should not be performed.
```
philippotto, line 145: the comment can now be removed thanks to the wrapped setIsSaving, I think

frontend/javascripts/dashboard/folders/metadata_table.tsx
```diff
@@ -116,136 +220,129 @@ export default function MetadataTable({
...
+          _.noop, // No state updates on unmounted component.
+          _.noop, // No state updates on unmounted component.
+        );
+      }
+    },
+    [metadataRef, hasUnsavedChangesRef],
+  );
 
-  // Sent automatic debounced updates to the server when metadata changes.
+  // Sent automatic async debounced updates to the server when metadata changed and there are no focused rows.
```
philippotto, line 251:
```suggestion
  // Send automatic async debounced updates to the server when metadata changed and there are no focused rows.
```

frontend/javascripts/dashboard/folders/details_sidebar.tsx
MichaelBuessemeyer: I also added a comment to the component itself

frontend/javascripts/dashboard/folders/metadata_table.tsx
MichaelBuessemeyer: yeah right, thanks for paying attention :pray:

MichaelBuessemeyer (commented):
Thanks for the fast review :pray: . I implemented your feedback :+1:

philippotto (commented):
front-end looks good :+1: one visual nitpick:

image

the trashcans should have a margin-right or something like that. there's very little room right now.

also, the PR has two open questions which need to be addressed. I don't have a strong opinion about them. maybe ask for opinions in slack?

MichaelBuessemeyer:
Before merging the migration & revision needs to be tested. I already tested this for me locally by:
- check out master
- run `yarn refresh-schema` 
- open wk dashboard
- Add some tags for at least one dataset and remember them
- Open the datasets table of the Postgres db manually and add some "species", "acquisition", and "brainRegion" to the details like done for publications
- Then checkout this branch
- Run the migration on the db -> this should succeed
- open the wk dataset dashboard and take a look at the metadata of the datasets that previously had tags. They should now be transferred to a tags entry
- use the dbtool to run a schema diff. This should run successfully without showing any schema diffs
- run the revision SQL script
- checkout master
- run schema diff -> should also show no schema diff
- Check the dataset dashboard: The previously added tags should appear there

philippotto:
the front-end looks good to me :+1: leaving final approval to @fm3. he probably should have a look at the migration code, too. also, does the python lib need to be adapted to this pr?

MichaelBuessemeyer:
>he probably should have a look at the migration code, too. 

Yeah definitely.

Moreover: as the tags should be deprecated should they be remove from the schema and API objects returned by the API or should they stay but the frontend cannot update and view them?

fm3 (changes requested):
Backend looking good already :) I added a couple of comments.
I also tested a bit more and it’s looking pretty good :)

I did not yet review the migrations. I think we need to decide about the NULLability first, and also about what to do with the tags.

If I understand correctly, the dataset tags are now turned into the new metadata. If this is true, the tags should be removed from the SQL schema and the Dataset case class.

However, the dataset json still needs to serve the tags for the moment, until the libs client is adapted. I think it would be good if the old tags could be served there. Meaning that the json serialization should convert the relevant metadata to tags and serve them, alongside the new metadata attribute. In the same way, updating the tags from the libs should also stay possible, so an adapter for that is needed as well.

Also, could you create a libs issue to adapt the libs to the new metadata format and deprecate the tags there too?

Also note that https://github.com/scalableminds/webknossos/pull/8006 was merged first so that the schema version in this PR needs to be increased to avoid conflicts.

tools/postgres/schema.sql
```diff
@@ -123,14 +123,14 @@ CREATE TABLE webknossos.datasets(
   sharingToken CHAR(256),
   logoUrl VARCHAR(2048),
   sortingKey TIMESTAMPTZ NOT NULL,
-  details JSONB,
+  metadata JSONB DEFAULT '[]',
```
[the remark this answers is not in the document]
MichaelBuessemeyer: :thinking: My thought here was that it might be easier to work with at least an empty array instead of null. I added the null / None support to the code to ensure easier compatibility. This includes update calls not necessarily having to pass metadata (at least I hope that's the case, as I wanted to implement it that way).

But I can easily remove the default value from the column :) Or even go further and make metadata non-optional if you think this is preferable.

MichaelBuessemeyer:
Ok what to do with tags :)

I just talked with @normanrz and the conclusion is to **keep tags** as some are using them as well as the pretty important filtering feature.

If we later want to merge the tags into the metadata, we can still do that. In case dataset/folder already has a metadata field named "tags" already exists, we just come up with something :).

frontend/javascripts/dashboard/folders/details_sidebar.tsx
```diff
@@ -172,15 +177,15 @@ function DatasetDetails({ selectedDataset }: { selectedDataset: APIDatasetCompac
             </Tag>
           )}
         </div>
+        {fullDataset && (
```
MichaelBuessemeyer, line 180: Open question? Is it ok to move the tags into the "spinner area"? Previously they weren'T but, the metadata should be within the "spinner area" as it depends on `fullDataset` and IMO the newly added UI element should appear after the already known tags.

philippotto: sure, why not :+1:

MichaelBuessemeyer:
testing went well. Tags should be restored in the frontend :)

tools/postgres/schema.sql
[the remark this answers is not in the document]
fm3: I don’t really have a preference for default value versus nullable. But I don’t much like having both. Please pick what’s easier implementation-wise :) My guess is that removing the default value (and so have the default be NULL/None) is easier.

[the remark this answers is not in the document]
MichaelBuessemeyer: Not is should be non nullable and with `DEFAULT '[]` for dataset and folders :).

The update route for datasets parses the metadata field as optional and used the already existing metadata in the backend query in case the user / client did not provide any metadata (untested) :see_no_evil:

fm3 (commented):
Backend LGTM

philippotto (approved):
just tested the dev instance again and everything looks good :+1: I didn't test the migration. ifi that is tested well, feel free to merge and deploy :shipit:

conf/evolutions/reversions/119-add-metadata-to-folders-and-datasets.sql
```diff
@@ -0,0 +1,68 @@
...
+-- Folder part
+ALTER TABLE webknossos.folders DROP COLUMN metadata;
+
+-- Dataset part
+ALTER TABLE webknossos.datasets ADD COLUMN details JSONB;
+ALTER TABLE webknossos.datasets ADD CONSTRAINT detailsIsJsonObject CHECK(jsonb_typeof(details) = 'object');
+
+-- Add existing info on species of metadata to details
+UPDATE webknossos.datasets
+SET details = jsonb_set(COALESCE(details, '{}'), '{species}', (
```
MichaelBuessemeyer, line 17: The newest commit changed 
```
SET details = jsonb_set(details, '{species}', (
```
to
```
SET details = jsonb_set(COALESCE(details, '{}'), '{species}', (
```

This was necessary, as the default value of the newly re-added details column was null and thus the `jsonb_set` would also return null. This should work now :)

MichaelBuessemeyer (commented):
@fm3 I just pushed a commit to fix the reversion. Could you please give me a final thumbs up?

I just tested the migration and reversion. both should work

MichaelBuessemeyer:
And @fm3 no need to test the migration again. I tested it and it should work :tm: :see_no_evil:

PR #7890, opened 2024-06-19
Improve case handling of segment group child mesh visibility options

MichaelBuessemeyer:
Prior to this PR, the decision whether to show the "Hide Meshes" or "Show Meshes" option was not case complete. This PR adds support for all cases that might occur:
- A group has some segments that are invisible, some that are visible -> show both options.
- A group has only invisible segments -> only show "Show Meshes"
- A group has only visible segments -> only show "Hide Meshes"
- A group has no loaded meshes -> Show none of the options

Moreover, this PR fixes that the property saving whether a group had visible meshes now also considers its own groups' meshes (recursive checking). Furthermore, this calculation is done for all groups and not only for the groups of the children of the root group (and the root group itself).

- https://___.webknossos.xyz

### Steps to test:
- Open an annotation with a segmentation
- Load some segments and group them in nested levels via groups. Kinda like this:
  image
- Load all their meshes.
- Now the "Show Meshes" action should not be shown for the groups.
- Hide some meshes in a nested group. On a parent -> parent group the option to "Show Meshes" should be available (hide as well).
- Test whether the options do what they are supposed to do and whether they are properly working recursively and do not only affect their immediate children
- Play a little more around. The options should be visible whenever at least one child segment is loaded but (not) visible.

### Issues:
- No issue exists for this. See https://scm.slack.com/archives/C02H5T8Q08P/p1718784540270399

------

### DIFF
--- a/frontend/javascripts/oxalis/view/right-border-tabs/segments_tab/segments_view.tsx
+++ b/frontend/javascripts/oxalis/view/right-border-tabs/segments_tab/segments_view.tsx
@@ -610,10 +610,8 @@ class SegmentsView extends React.Component<Props, State> {
           const segmentMesh = meshes[segment.id];
           // Only regard loaded, but invisible meshes
           if (segmentMesh != null) {
-            visibilityEntry.areSomeSegmentsVisible =
-              visibilityEntry.areSomeSegmentsVisible || segmentMesh.isVisible;
-            visibilityEntry.areSomeSegmentsInvisible =
-              visibilityEntry.areSomeSegmentsInvisible || !segmentMesh.isVisible;
+            visibilityEntry.areSomeSegmentsVisible ||= segmentMesh.isVisible;
+            visibilityEntry.areSomeSegmentsInvisible ||= !segmentMesh.isVisible;
           }
         });
         newVisibleMap[group.groupId] = visibilityEntry;
--- a/frontend/javascripts/oxalis/view/right-border-tabs/segments_tab/segments_view.tsx
+++ b/frontend/javascripts/oxalis/view/right-border-tabs/segments_tab/segments_view.tsx
@@ -121,6 +121,7 @@ import { SegmentStatisticsModal } from "./segment_statistics_modal";
 import { APIJobType, type AdditionalCoordinate } from "types/api_flow_types";
 import { DataNode } from "antd/lib/tree";
 import { ensureSegmentIndexIsLoadedAction } from "oxalis/model/actions/dataset_actions";
+import { ValueOf } from "types/globals";
 
 const { confirm } = Modal;
 const { Option } = Select;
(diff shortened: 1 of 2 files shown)


frontend/javascripts/oxalis/view/right-border-tabs/segments_tab/segments_view.tsx
```diff
@@ -582,28 +585,44 @@ class SegmentsView extends React.Component<Props, State> {
...
-        const isSomeSegmentLoadedAndInvisible = segmentsOfGroup.some((segment) => {
+        segmentsOfGroup.forEach((segment) => {
           const segmentMesh = meshes[segment.id];
           // Only regard loaded, but invisible meshes
-          return segmentMesh != null && !meshes[segment.id].isVisible;
+          if (segmentMesh != null) {
+            visibilityEntry.areSomeSegmentsVisible =
+              visibilityEntry.areSomeSegmentsVisible || segmentMesh.isVisible;
+            visibilityEntry.areSomeSegmentsInvisible =
+              visibilityEntry.areSomeSegmentsInvisible || !segmentMesh.isVisible;
```
philippotto, line 616:
```suggestion
            visibilityEntry.areSomeSegmentsVisible ||= segmentMesh.isVisible;
            visibilityEntry.areSomeSegmentsInvisible ||= !segmentMesh.isVisible;
```

frontend/javascripts/oxalis/view/right-border-tabs/segments_tab/segments_view.tsx
```diff
@@ -582,28 +585,44 @@ class SegmentsView extends React.Component<Props, State> {
...
+          ),
+          areSomeSegmentsInvisible: group.children.some(
+            (childGroup) => newVisibleMap[childGroup.groupId].areSomeSegmentsInvisible,
+          ),
+        };
+
         const segmentsOfGroup = groupToSegmentsMap[group.groupId];
         if (segmentsOfGroup == null) return;
-        const isSomeSegmentLoadedAndInvisible = segmentsOfGroup.some((segment) => {
+        segmentsOfGroup.forEach((segment) => {
```
philippotto, line 609: couldn't you still use `some` to avoid inspecting all segments of a group once both bools are true? within the arrow function you could return `return visibilityEntry.areSomeSegmentsVisible && visibilityEntry.areSomeSegmentsInvisible` if I'm not mistaken.

philippotto (approved):
thank you for fixing this! works great! just left two small suggestions.

frontend/javascripts/oxalis/view/right-border-tabs/segments_tab/segments_view.tsx
MichaelBuessemeyer: Oh right, sounds smart :tada:. On it :+1:

philippotto:
@MichaelBuessemeyer Can this be merged? :)

PR #7893, opened 2024-06-20
Better error message when trying to upload annotation to missing dataset

fm3:
When uploading an NML to a dataset that has no content (e.g. empty folder in binaryData dir) or is otherwise broken, the user now gets a readable error message rather than the generic Couldn’t find the requested resource.

Also, we save a few database requests by only constructing the dataSource once.

### Steps to test:
- Create empty dataset folder in binaryData dir, let wk scan the dir.
- Upload an annotation with `<experiment name="missing" />`
- Should get a readable error message
- Uploading to an existing working dataset should still work too.


MichaelBuessemeyer (approved):
Thanks for the improvements. :tada: 

Here are my testing notes:
- I created an empty dataset called `Empty`
- When I use ` <experiment name="missing" organization="sample_organization" description="" wkUrl="http://localhost:9000"/>`, I get: `Could not access dataset missing. Does your team have access?` as an error message. This is not equal to what `Messages("dataset.notImported", dataset.name)` should produce, as messages.conf states `dataset.notImported=Dataset {0} is not imported`
- When I use `<experiment name="missing"/>` I also get `Could not access dataset missing. Does your team have access?` as an error message. 
- When using the name of the actual empty dataset `<experiment name="Empty" organization="sample_organization" description="" wkUrl="http://localhost:9000"/>`, I get the intended error message: `Dataset Empty is not imported` 

Is this how it is intended?

Nevertheless: The changes look good and seem to work :checkered_flag:

fm3:
Yes, I’d say that’s the intended behavior :) Thanks!

Issue #7670, opened 2024-03-06
[AI models] AI model training and management


normanrz:
We want to add model training and management capabilities to WEBKNOSSOS. In the future, users shall be able to train their own models from annotations, run segmentation models on new datasets, and manage their models.
![Untitled-2024-03-06-1334(2)](https://github.com/scalableminds/webknossos/assets/335438/4eb84497-3a16-4eca-aded-cbdc7cf54cf8)

- Frontend
  - Add train model modal #7672 
  - Expand the inference modal #7673
  - Add a table view for models #7674
  - Automatically pick bboxes for labelling #7718
  - Multi-annotation training #8005 
- Backend
  - Add a model entity and the CRUD routes #7675
  - Link jobs with vx runs https://github.com/scalableminds/voxelytics/issues/3146
- Worker
  - Expand the inference job
  - Add a model training job

In the first iteration, these features should only be available to super admins. The yaml presets should not be exposed in the open-source code.

See also: https://www.notion.so/scalableminds/Design-Doc-Model-Training-Inference-Management-in-WK-28e4934134b640eaac826f86ae135b17

PR #7902, opened 2024-06-25
Allow supplying custom workflow yaml for infer_with_model job

fm3:
- https://___.webknossos.xyz

### Steps to test:
- set up local worker, with branch from https://github.com/scalableminds/voxelytics/pull/3605
- yarn enable-jobs
- train a model
- infer with the model, both with and without custom yaml
- example custom yaml: https://github.com/scalableminds/voxelytics/pull/3605#issuecomment-2250234257

### TODOs:
- [x] Backend: add this param to the route, to job_args
- [x] Worker:
   - [x] use the passed yaml https://github.com/scalableminds/voxelytics/pull/3605
   - [x] decide about the `--until` call, [compare discussion](https://scm.slack.com/archives/C03B6160ATY/p1720085016993699?thread_ts=1720083853.328949&cid=C03B6160ATY)
   - [x] fill out template part of passed yaml
- [x] Frontend: 
  - [x] Show textarea for yaml, append it to route params
  - [x] Prefill textarea

### Issues:
 - contributes to https://github.com/scalableminds/webknossos/issues/7670

------


philippotto:
probably read for review now, right?

frontend/javascripts/oxalis/view/action-bar/default-workflow.ts
```diff
@@ -0,0 +1,24 @@
+export default `predict:
```
philippotto, line 1: re file name: how about `default-predict-workflow-template.ts`. the folder location is not optimal, either, but I'd love to restructure a lot of files, anyway. maybe in another pr...

fm3:
> probably read for review now, right?

The worker side isn’t ready, I’ll do some end-to-end testing next. I think that more checkboxes may arise (e.g. we will likely drop the option to select the output layer name)

util/src/main/scala/com/scalableminds/util/requestlogging/RequestLogging.scala
```diff
@@ -31,9 +31,10 @@ trait AbstractRequestLogging extends LazyLogging {
   def logTime(notifier: String => Unit, durationThreshold: FiniteDuration = 30 seconds)(
       block: => Future[Result])(implicit request: Request[_], ec: ExecutionContext): Future[Result] = {
     def logTimeFormatted(executionTime: FiniteDuration, request: Request[_], result: Result): Unit = {
-      val debugString = s"Request ${request.method} ${request.uri} took ${BigDecimal(executionTime.toMillis / 1000)
-        .setScale(2, BigDecimal.RoundingMode.HALF_UP)} seconds and was${if (result.header.status != 200) " not "
-      else " "}successful"
+      val debugString =
+        s"Request ${request.method} ${request.uri} took ${BigDecimal(executionTime.toMillis.toDouble / 1000)
```
fm3, line 35: slipped in here: the decimals were lost due to integer division

frontend/javascripts/admin/api/jobs.ts
```diff
@@ -185,13 +185,11 @@ export function startNeuronInferralJob(
   datasetName: string,
   layerName: string,
   bbox: Vector6,
-  outputSegmentationLayerName: string,
```
philippotto, line 188: was there an active decision to not support providing a custom output name?

frontend/javascripts/oxalis/view/action-bar/starting_job_modals.tsx
```diff
@@ -55,10 +50,11 @@ import { isBoundingBoxExportable } from "./download_modal_view";
 import features from "features";
 import { setAIJobModalStateAction } from "oxalis/model/actions/ui_actions";
 import { InfoCircleOutlined } from "@ant-design/icons";
-import { TrainAiModelTab } from "../jobs/train_ai_model";
+import { TrainAiModelTab, CollapsibleWorkflowYamlEditor } from "../jobs/train_ai_model";
 import { LayerSelectionFormItem } from "components/layer_selection";
 import { useGuardedFetch } from "libs/react_helpers";
 import _ from "lodash";
+import DefaultPredictWorkflow from "./default-workflow";
```
philippotto, line 57: since this is a constant, I'd name it `DEFAULT_PREDICT_WORKFLOW`.

frontend/javascripts/admin/api/jobs.ts
fm3: Kinda. We made the decision to use the vx publish task in the inference workflows. Unfortunately, it does not currently support custom output layer names, and also does not link in any of the input dataset’s layers, except the color layer the inference ran on. So a standard name for the output layer also doesn’t really make sense.

If in the future we want to add these functionalities to the publish task, we can decide if we want the user to supply this name again. However, all uses we saw used the suggested standard name, so maybe it’s not worth the extra complexity for the user anyway.

fm3:
@MichaelBuessemeyer could you please take a look at Philpp’s frontend comments?

frontend/javascripts/admin/api/jobs.ts
philippotto: > Unfortunately, it does not currently support custom output layer names, and also does not link in any of the input dataset’s layers, except the color layer the inference ran on

hm, ok, it feels like a step back, but I understand that it doesn't make sense now due to the publish task. let's see when it will become a pain point again. luckily, we can compose datasets in the UI afterwards now.

philippotto:
testing went ok :+1:

frontend/javascripts/oxalis/view/action-bar/default-workflow.ts
MichaelBuessemeyer: > the folder location is not optimal, either,

Yeah, agreed. Maybe let's tackle this at the next hackathon or so?

MichaelBuessemeyer (commented):
> @MichaelBuessemeyer could you please take a look at Philpp’s frontend comments?

Thanks for the ping. Done :)

philippotto (approved):
there's a conflict to resolve, but other than that, everything seems good :+1:

PR #7903, opened 2024-06-26
Fix not being able to save dataset settings 

MichaelBuessemeyer:
Likely due to the frontend changes of #7697, the dataset settings got messed up. The reason is, that the current state of the component does not have the state field `savedDataSourceOnServer`, but it is referenced in a diffiing operation of the datasource schema. See

https://github.com/scalableminds/webknossos/blob/c50a81860b5e9882b24677e5a750b10f63c5d6a8/frontend/javascripts/dashboard/dataset/dataset_settings_view.tsx#L277-L282

The diffing is used to decide whether the frontend should send an update of the datasource schema to the backend. But in the current version of the master, the diff check always fails as the state field `savedDataSourceOnServer` is always undefined (it does not exist in the state). My fix re-added this field to the state (by renaming the unused `dataSource` to  `savedDataSourceOnServer`)  and adds a default empty object to the diff, to avoid such an error in any case. Moreover, the type of `savedDataSourceOnServer` says it can be undefined / null thus the additional optional empty object is needed for the diff. 

- https://___.webknossos.xyz

### Steps to test:
- Check whether you can do changes in the dataset settings of one dataset on the master (e.g.: https://master.webknossos.xyz/datasets/sample_organization/ROI2017_wkw/edit) this should not work.
- Do the same on the dev instance for this branch. This should now work. 
- Open the dev tools, in `dataset_settings_view.tsx` in `submit` in line 365 set a break point. Do not edit the datasource settings and hit submit. The debugging should show, that now update is set to the server. Alternatively, just record the sent requests to the server. Only when the datasource settings were modified, an update should be sent to the server.

### Issues:
- Not issue exists for this

------

### DIFF
--- a/CHANGELOG.unreleased.md
+++ b/CHANGELOG.unreleased.md
@@ -30,6 +30,7 @@ For upgrade instructions, please check the [migration guide](MIGRATIONS.released
 
 ### Fixed
 - Fixed a bug where the warning to zoom in to see the agglomerate mapping was shown to the user even when the 3D viewport was maximized and no volume data was shown. [#7865](https://github.com/scalableminds/webknossos/issues/7865) 
+- Fixed a bug that prevented saving new dataset settings. [#7903](https://github.com/scalableminds/webknossos/pull/7903)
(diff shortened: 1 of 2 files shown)


philippotto (approved):
Code looks good and I didn't found problems during testing.

PR #7905, opened 2024-06-27
Fix segments tab not rendering when a segment group does not have at least on direct child segment

MichaelBuessemeyer:
This PR fixes an error that crashes the segments tab in case the segments tab has a group with no direct segments as children. 

During the development of #7890 I missed the path where a segment group has no children. Please find my inline comment below.

- https://___.webknossos.xyz

### Steps to test:
- Set up the following: 
image
- This should now work, and not result in a rendering error. On the master it should crash with a `TypeError: Cannot read properties of undefined (reading 'areSomeSegmentsVisible')` error.

### Issues:
- fixes a bug introduced with pr #7890

### DIFF
--- a/frontend/javascripts/oxalis/view/right-border-tabs/segments_tab/segments_view.tsx
+++ b/frontend/javascripts/oxalis/view/right-border-tabs/segments_tab/segments_view.tsx
@@ -605,19 +605,20 @@ class SegmentsView extends React.Component<Props, State> {
         };
 
         const segmentsOfGroup = groupToSegmentsMap[group.groupId];
-        if (segmentsOfGroup == null) return;
-        segmentsOfGroup.some((segment) => {
-          const segmentMesh = meshes[segment.id];
-          // Only regard loaded, but invisible meshes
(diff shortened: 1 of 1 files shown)


philippotto (approved):
good find!

frontend/javascripts/oxalis/view/right-border-tabs/segments_tab/segments_view.tsx
```diff
@@ -605,19 +605,20 @@ class SegmentsView extends React.Component<Props, State> {
         };
 
         const segmentsOfGroup = groupToSegmentsMap[group.groupId];
-        if (segmentsOfGroup == null) return;
```
MichaelBuessemeyer, line 608: This is the root of all evil. Due to the early return, a group with no segments as direct children will not have its `visibilityEntry` entry  added to `newVisibleMap`. Thus the assumption that all child groups of a group already have an entry in `newVisibleMap` does not hold in lines 599-604 leading to the bug.

PR #7876, opened 2024-06-11
Improve mobile support of views

MichaelBuessemeyer:
This PR improves the responsiveness of the login, registration

### URL of deployed dev instance (used for testing):
- https://___.webknossos.xyz

### Steps to test:
- Use your browser dev tools to emulate a mobile device
- The login, registration & forgot password view should all look ok and should not overflow the screen width
- This also goes for the login view hidden in the navbar while viewing a public annotation / dataset and not being logged in.
  - Create an annotation, make it public, copy the sharing link.
  - Log out and open the sharing link on the mobile device browser tab
  - In the navbar hit the user icon -> the auth popup should look ok :)
- Open the dataset dashboard in the mobile browser tab. Compared to before, the dataset table should at least be a little visible and when working with a vertical mobile device, the table should now be kinda usable compared to before.

### TODOs:
- [ ] Any other ideas of views to make more responsive?

### Issues:
- No issue exists for this

------
(Please delete unneeded items, merge only when none are left open)
- [x] Updated [changelog](../blob/master/CHANGELOG.unreleased.md#unreleased)

PR #7909, opened 2024-06-28
(Re-)Center login forms [not urgent]

MichaelBuessemeyer:
Due to the recent PR #7876 making wk a little more accessible for mobile devices, the login forms were not centered anymore on devices with a high screen width. This PR aims to fix this.

- https://___.webknossos.xyz

### Steps to test:
- Open the login view of the dev instance. Zoom the page out. The login form should always be horizontally centered. Compare it to the current master version where the login form is not always centered.
- Create an annotation of a non-public dataset. Then, share this annotation as public, copy the link and log out from the dev instance. Open the annotation. The annotation should not be opened as the underlying dataset it not public instead the other version of the login form should be rendered. This login form should be center horizontally as well regardless of the zooming :) 

### Issues:
- No issue exists for this

------

### DIFF
--- a/frontend/javascripts/admin/auth/login_view.tsx
+++ b/frontend/javascripts/admin/auth/login_view.tsx
@@ -30,7 +30,7 @@ function LoginView({ history, redirect }: Props) {
   return (
     <Row justify="center" align="middle" className="login-view">
       <Col xs={22} sm={20} md={16} lg={12} xl={8}>
-        <Card className="login-content">
+        <Card className="login-content" style={{ margin: "0 auto" }}>
           <h3>Login</h3>
           <LoginForm layout="horizontal" onLoggedIn={onLoggedIn} />
         </Card>
--- a/frontend/javascripts/components/brain_spinner.tsx
+++ b/frontend/javascripts/components/brain_spinner.tsx
@@ -107,8 +107,10 @@ export function CoverWithLogin({ onLoggedIn }: { onLoggedIn: () => void }) {
         align="middle"
       >
         <Col xs={22} sm={20} md={16} lg={12} xl={8}>
-          <h3>Try logging in to view the dataset.</h3>
-          <LoginForm layout="horizontal" onLoggedIn={onLoggedIn} />
+          <span style={{ margin: "0 auto", display: "table" }}>
+            <h3>Try logging in to view the dataset.</h3>
+            <LoginForm layout="horizontal" onLoggedIn={onLoggedIn} />
+          </span>
         </Col>
       </Row>
     </div>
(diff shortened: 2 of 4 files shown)


frontend/javascripts/admin/auth/login_view.tsx
```diff
@@ -30,7 +30,7 @@ function LoginView({ history, redirect }: Props) {
   return (
     <Row justify="center" align="middle" className="login-view">
       <Col xs={22} sm={20} md={16} lg={12} xl={8}>
-        <Card className="login-content">
+        <Card className="login-content" style={{ margin: "0 auto" }}>
```
hotzenklotz, line 33: Die `row` in Zeile 31 ist `centered` doch schon den content. Ich versteh das nicht?

MichaelBuessemeyer: The row centers its children, thats correct. Thus, the column is centered. But the Card (child of the column)  does not fill the columns full width in case of large screens. In that scenario the Card just starts at the left edge of the column and is therefore not centered on large screens. Thist margin auto fix, mitigates this and the result is a centered login view.

frontend/javascripts/components/brain_spinner.tsx
```diff
@@ -107,8 +107,10 @@ export function CoverWithLogin({ onLoggedIn }: { onLoggedIn: () => void }) {
         align="middle"
       >
         <Col xs={22} sm={20} md={16} lg={12} xl={8}>
-          <h3>Try logging in to view the dataset.</h3>
-          <LoginForm layout="horizontal" onLoggedIn={onLoggedIn} />
+          <span style={{ margin: "0 auto", display: "table" }}>
```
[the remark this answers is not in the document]
MichaelBuessemeyer: > :see_no_evil: 

Just my thought :laughing:. But this was the only CSS fix I could come up with in a reasonable amount of time. Somehow  this centers the the whole login form :man_shrugging:

frontend/javascripts/admin/auth/login_view.tsx
hotzenklotz: Right. In that case I would add the `margin: auto` style to the `login-content` CSS class/Less file directly. That way "Sign up" and "Reset Password" also profit from theses changes.

frontend/javascripts/components/brain_spinner.tsx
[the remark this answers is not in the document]
hotzenklotz: I propose to style this view identical to the regular login view. That way we can 1) reuse styling classes, 2) use the same nice design/background, and 3( avoid the weird `table` stuff. 

See my commit ff14f5ebbd90b71f7293180ed49dad52be6ce8ac for a proposal.

MichaelBuessemeyer (commented):
Awesome @hotzenklotz. Thanks for pushing the changes. I just tested them and they are working splendid. Could you please approve this pr to get this merged?

hotzenklotz (approved):

Issue #7827, opened 2024-05-24
Focus in segment list for segments in a collapsed group


Albane-dlv:
The "focus in segment list" command accessible through the context menu should also work when the segment is in a collapsed group.

PR #7911, opened 2024-07-01
Add option to expand and collapse all children of segment group

### LINKED ISSUE #7829: Collapse subgroups under any higher order group in the segment list
Similarly to the tree list, it should be possible to collapse and expand all children of a group.

dieknolle3333:
### Dev instance
https://collapseallsegmentgroupchildren.webknossos.xyz/

### Steps to test:
- go to segments tab in any annotation
- click some segments and create some groups, groups should be nested for at least three levels
- right click any group to access group context menu
- if the segment group has any subgroups, there should be one or two new items: "expand/collapse all subgroups".
- if the group is expanded and some subgroups are expanded and some are collapsed, both entries should be shown
- if the group itself is expanded and all subgroups are collapsed, you should only be able to expand them. the opposite is true if all subgroups are expanded.
- if the group itself is collapsed, you should only be able to expand all subgroups. the group itself will be expanded aswell.
~~- # 7827: Focus in segment list should also work for segments in a collapsed group~~

### TODOs:

- [x] icon
- [x] switch label
- [x] default: all expanded
- [x] recursively expand subgroups

### Issues:
- fixes #7829 

------

### DIFF
--- a/frontend/javascripts/oxalis/view/right-border-tabs/segments_tab/segments_view.tsx
+++ b/frontend/javascripts/oxalis/view/right-border-tabs/segments_tab/segments_view.tsx
@@ -548,7 +548,7 @@ class SegmentsView extends React.Component<Props, State> {
               ),
             );
           },
-          onCancel() {},
+          onCancel() { },
         });
       } else {
         // If only one segment is selected, select group without warning (and vice-versa) even though ctrl is pressed.
@@ -849,9 +849,9 @@ class SegmentsView extends React.Component<Props, State> {
 
       const maybeMappingName =
         !isEditableMapping &&
-        mappingInfo.mappingStatus !== MappingStatusEnum.DISABLED &&
-        mappingInfo.mappingType === "HDF5" &&
-        mappingInfo.mappingName != null
+          mappingInfo.mappingStatus !== MappingStatusEnum.DISABLED &&
+          mappingInfo.mappingType === "HDF5" &&
+          mappingInfo.mappingName != null
           ? mappingInfo.mappingName
           : undefined;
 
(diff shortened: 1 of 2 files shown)


dieknolle3333:
I decided to work on #7827 here because I need the changes that are made in this PR and I think it's only a small change to solve this issue.

frontend/javascripts/oxalis/view/right-border-tabs/segments_tab/segments_view.tsx
```diff
@@ -1842,6 +1885,55 @@ class SegmentsView extends React.Component<Props, State> {
     );
   }
 
+  getExpandSubgroupsItem(groupId: number) {
+    const children = this.getSubGroupsAsTreeNodes(groupId);
+    const areAllChildrenExpanded =
+      children.filter((childNode) => this.state.expandedGroupKeys?.includes(childNode)).length ===
+      children.length;
```
philippotto, line 1892:
```suggestion
    const areAllChildrenExpanded =
      children.every((childNode) => this.state.expandedGroupKeys?.includes(childNode));
```

frontend/javascripts/oxalis/view/right-border-tabs/segments_tab/segments_view.tsx
```diff
@@ -1842,6 +1885,55 @@ class SegmentsView extends React.Component<Props, State> {
     );
   }
 
+  getExpandSubgroupsItem(groupId: number) {
+    const children = this.getSubGroupsAsTreeNodes(groupId);
+    const areAllChildrenExpanded =
```
philippotto, line 1890: there are several spots where you use `include()` which has a linear performance overhead. combined with calling it in a collection, this becomes O(n²). initially, I thought, `expandedGroupKeys` would be small usually, but I think it can become quite large if users prefer keeping their groups open. to circumvent this, I'd define a `const expandedKeySet = new Set(this.state.expandedGroupKeys)` and then you can do a look up via `expandedKeySet.has(key)`.

frontend/javascripts/oxalis/view/right-border-tabs/segments_tab/segments_view.tsx
```diff
@@ -1842,6 +1885,55 @@ class SegmentsView extends React.Component<Props, State> {
...
+    );
+    if (areAllChildrenExpanded && isGroupItselfExpanded) {
+      return null;
+    }
+    const groupsToExpand = children;
+    // It doesn't make sense to expand subgroups if the group itself is collapsed, so also expand the group.
+    if (!isGroupItselfExpanded) groupsToExpand.push(this.getKeyForGroupId(groupId));
+    return {
+      key: "expandAll",
+      disabled: !this.props.allowUpdate,
```
philippotto, line 1904: can I expand/collapse a group itself when the annotation is read-only? I think, I would allow this. the save_saga should not save anything, when it's not allowed. so, we don't do any damage (it would be worse to allow mutating actual data and then it is not saved without warning).

philippotto (changes requested):
cool, looks good overall! just see my few comments.

frontend/javascripts/oxalis/view/right-border-tabs/segments_tab/segments_view.tsx
dieknolle3333: oh I didnt know that this is what this prop means. makes total sense, I agree that this should be possible.

frontend/javascripts/oxalis/view/right-border-tabs/segments_tab/segments_view.tsx
dieknolle3333: nice, I think I can actually tell that its faster now!

philippotto (approved):
nice, works well :+1:

Issue #7714, opened 2024-03-25
Recursive explore for listable DataVaults


normanrz:
Some storage backends, e.g. S3, GCS, local file system, support directory listing. On these storage backends, it would be great if we could run the dataset exploration recursively to capture nested or multiple layers into one dataset.

For example, this dataset is addressed by `s3://janelia-cosem-datasets/jrc_mus-nacc-4/jrc_mus-nacc-4.zarr/` but the actual multi-scale image is actually at `s3://janelia-cosem-datasets/jrc_mus-nacc-4/jrc_mus-nacc-4.zarr/recon-2/em/fibsem-int16/`. That requires a bit of guess work on the user side that we should be able to avoid. See here https://open.quiltdata.com/b/janelia-cosem-datasets/tree/jrc_mus-nacc-4/jrc_mus-nacc-4.zarr/

PR #7912, opened 2024-07-02
Recursive exploration of remote datasets

MichaelBuessemeyer:
This PR adds recursive exploration to the already existing remote dataset exploration. This is supported for the local file system, GCS and S3.

- https://___.webknossos.xyz

### Steps to test:
- Test GCS via `gs://neuroglancer-fafb-data/fafb_v14/`. This should result in a successfully explored dataset.
- Test S3 via `s3://janelia-cosem-datasets/jrc_mus-nacc-4/jrc_mus-nacc-4.zarr/`. This should result in a successfully explored dataset.
- Test locally: 
  - Create a new local folders e.g. `<wk-root>/binaryData2/some_dir/more_dir`, `<wk-root>/binaryData2/other_dir/more_dir` 
  - Add `<wk-root>/binaryData2/` to the whilelist in the `application.conf` in line 197.
  - Enter  `file://`<wk-root>/binaryData2/` into the add remote dataset form. The request should fail and only include a short error message, not leaking any information about the underlying folder structure of the server.
 
  - Add a new dataset (not wkw, as wkw exploration is not implemented) e.g. `l4_sample_zarr3_sharded` to  `<wk-root>/binaryData2/other_dir/more_dir`
  - Enter  `file://`<wk-root>/binaryData2/` into the add remote dataset form. The request should successfully find the dataset.
 

### TODOs:
- [x] Currently, the backend leaks the directory structure of the whitelisted directories allowed by the whitelisting feature in case the exploration fails. This should [not be exposed](https://portswigger.net/kb/issues/00600100_directory-listing). Moreover, @normanrz argued that the information is not useful to users. 
   -> I'd add a warning to the docs about the whitelisting feature to only include datasets in the subdirectories and not any kind of sensitive information like ssh key and such. Moreover, the debug log should not be exposed to the end users. At least in case a local file system is used. In case wk crawls remote cloud storages, the person using wk already has the necessary credentials (if necessary) to list the cloud storage. In that case wk does not leak any information the user would already have.
- [x] Should the mutable report be included in the client answer even for non local datasets? -> It is kept for now and maybe removed in the future.

### Issues:
- fixes [#7714](https://github.com/scalableminds/webknossos/issues/7714)

------


project/Dependencies.scala
```diff
@@ -56,7 +56,7 @@ object Dependencies {
     // MultiArray (ndarray) handles. import ucar
     "edu.ucar" % "cdm-core" % "5.4.2",
     // Amazon S3 cloud storage client. import com.amazonaws
-    "com.amazonaws" % "aws-java-sdk-s3" % "1.12.584",
+    "com.amazonaws" % "aws-java-sdk-s3" % "1.12.584", // TODO Update?!!
```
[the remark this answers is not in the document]
fm3: I wrote https://github.com/scalableminds/webknossos/issues/7913 Sounds good (reads like it may improve async api?) but should not block this PR

package.json
```diff
@@ -76,6 +76,10 @@
   "scripts": {
     "start": "node tools/proxy/proxy.js",
     "build": "node --max-old-space-size=4096 node_modules/.bin/webpack --env production",
+    "build-backend": "yarn build-wk-backend && yarn build-wk-datastore && yarn build-wk-tracingstore",
+    "build-wk-backend": "sbt -no-colors -DfailOnWarning compile stage",
+    "build-wk-datastore": "sbt -no-colors -DfailOnWarning \"project webknossosDatastore\" copyMessages compile stage",
+    "build-wk-tracingstore": "sbt -no-colors -DfailOnWarning \"project webknossosTracingstore\" copyMessages compile stage",
```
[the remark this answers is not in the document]
fm3: Could you add a comment to the package.json what these are for? Might confuse new developers, suggesting that this is necessary to build wk to then start it

webknossos-datastore/app/com/scalableminds/webknossos/datastore/datavault/DataVault.scala
```diff
@@ -5,6 +5,9 @@ import com.scalableminds.util.tools.Fox
 import scala.concurrent.ExecutionContext
 
 trait DataVault {
+  val MAX_EXPLORED_ITEMS_PER_LEVEL = 10
```
fm3, line 8: I’d say this is not a property of DataVault, but of the exploring code. Maybe move this to the abstract explorer or ExploreRemoteLayerService or something like that? DataVault should not have to know that it is used to explore datasets.

webknossos-datastore/app/com/scalableminds/webknossos/datastore/datavault/DataVault.scala
```diff
@@ -5,6 +5,9 @@ import com.scalableminds.util.tools.Fox
...
   def readBytesAndEncoding(path: VaultPath, range: RangeSpecifier)(
       implicit ec: ExecutionContext): Fox[(Array[Byte], Encoding.Value)]
+
+  def listDirectory(path: VaultPath)(implicit ec: ExecutionContext): Fox[List[VaultPath]]
```
fm3, line 12: I’d add a parameter `limit: Option[Int]` here, rather than use the constant, see above.

webknossos-datastore/app/com/scalableminds/webknossos/datastore/datavault/FileSystemDataVault.scala
```diff
@@ -8,21 +8,28 @@ import org.apache.commons.lang3.builder.HashCodeBuilder
...
 import java.nio.file.{Files, Path, Paths}
+import java.util.stream.Collectors
 import scala.concurrent.ExecutionContext
+import scala.jdk.CollectionConverters._
 
 class FileSystemDataVault extends DataVault {
 
-  override def readBytesAndEncoding(path: VaultPath, range: RangeSpecifier)(
-      implicit ec: ExecutionContext): Fox[(Array[Byte], Encoding.Value)] = {
+  private def vaultPathToLocalPath(path: VaultPath)(implicit ec: ExecutionContext): Fox[Path] = {
```
fm3, line 17: Please move the private method below the public one that uses it. Either below the first usage, or further down. (That’s the typical reading order of the backend files. or at least it should be 😅)

webknossos-datastore/app/com/scalableminds/webknossos/datastore/datavault/GoogleCloudDataVault.scala
```diff
@@ -72,6 +73,18 @@ class GoogleCloudDataVault(uri: URI, credential: Option[GoogleServiceAccountCred
     } yield (bytes, encoding)
   }
 
+  override def listDirectory(path: VaultPath)(implicit ec: ExecutionContext): Fox[List[VaultPath]] = {
+    val objName = path.toUri.getPath.tail
+    val blobs = storage.list(bucket,
+                             Storage.BlobListOption.prefix(objName),
+                             Storage.BlobListOption.currentDirectory(),
+                             Storage.BlobListOption.pageSize(MAX_EXPLORED_ITEMS_PER_LEVEL))
+    val subDirectories = blobs.getValues.asScala.toList.filter(_.isDirectory)
```
fm3, line 82: The limit should probably be applied after this isDirectory filtering. Maybe use a default pageSize (1000?). This should not be super expensive. The exploring itself will be.

webknossos-datastore/app/com/scalableminds/webknossos/datastore/datavault/HttpsDataVault.scala
```diff
@@ -44,6 +44,9 @@ class HttpsDataVault(credential: Option[DataVaultCredential], ws: WSClient) exte
 
   }
 
+  override def listDirectory(path: VaultPath)(implicit ec: ExecutionContext): Fox[List[VaultPath]] =
+    Fox.successful(List.empty)
```
fm3, line 48:
```suggestion
    // HTTP file listing is not currently supported.
    Fox.successful(List.empty)
```

webknossos-datastore/app/com/scalableminds/webknossos/datastore/datavault/S3DataVault.scala
```diff
@@ -73,6 +75,26 @@ class S3DataVault(s3AccessKeyCredential: Option[S3AccessKeyCredential], uri: URI
     }
   }
 
+  private def performGetObjectSummariesRequest(bucketName: String, keyPrefix: String)(
```
fm3, line 78: The other method was called performRequest, and got a request as a parameter. I’d vote to call this one getObjectSummaries, since it constructs its own request, rather than “performing” a passed one. What do you think?

webknossos-datastore/app/com/scalableminds/webknossos/datastore/explore/ExploreRemoteLayerService.scala
```diff
@@ -83,20 +83,21 @@ class ExploreRemoteLayerService @Inject()(dataVaultService: DataVaultService,
...
       remotePath <- dataVaultService.getVaultPath(remoteSource) ?~> "dataVault.setup.failed"
-      layersWithVoxelSizes <- exploreRemoteLayersForRemotePath(
-        remotePath,
+      layersWithVoxelSizes <- recursivelyExploreRemoteLayerAtPaths(
+        List((remotePath, 0)),
         credentialId,
-        reportMutable,
         List(
-          new ZarrArrayExplorer(Vec3Int.ones),
+          // Explorers are ordered to prioritize the explorer reading meta information over raw Zarr, N5, ... data.
```
fm3, line 90: 👍

webknossos-datastore/app/com/scalableminds/webknossos/datastore/explore/ExploreRemoteLayerService.scala
```diff
@@ -106,26 +107,93 @@ class ExploreRemoteLayerService @Inject()(dataVaultService: DataVaultService,
         uri.getPath.startsWith(whitelistEntry))) ?~> s"Absolute path ${uri.getPath} in local file system is not in path whitelist. Consider adding it to datastore.localFolderWhitelist"
     } else Fox.successful(())
 
-  private def exploreRemoteLayersForRemotePath(remotePath: VaultPath,
-                                               credentialId: Option[String],
-                                               reportMutable: ListBuffer[String],
-                                               explorers: List[RemoteLayerExplorer])(
+  private val MAX_RECURSIVE_SEARCH_DEPTH = 8
+  private def handleExploreResult(explorationResult: Box[List[(DataLayerWithMagLocators, VoxelSize)]],
```
fm3, line 111: let’s use either exploreResult or explorationResult, but not both :)

webknossos-datastore/app/com/scalableminds/webknossos/datastore/explore/ExploreRemoteLayerService.scala
```diff
@@ -106,26 +107,93 @@ class ExploreRemoteLayerService @Inject()(dataVaultService: DataVaultService,
...
+    case Full(layersWithVoxelSizes) =>
+      reportMutable += s"Found ${layersWithVoxelSizes.length} ${explorer.name} layers at $path."
+      Fox.successful(layersWithVoxelSizes)
+    case f: Failure =>
+      reportMutable += s"Error when reading $path as ${explorer.name}: ${Fox.failureChainAsString(f)}"
+      Fox.empty
+    case Empty =>
+      reportMutable += s"Error when reading $path as ${explorer.name}: Empty"
+      Fox.empty
+
```
fm3, line 125:
```suggestion
```

webknossos-datastore/app/com/scalableminds/webknossos/datastore/explore/ExploreRemoteLayerService.scala
```diff
@@ -106,26 +107,93 @@ class ExploreRemoteLayerService @Inject()(dataVaultService: DataVaultService,
...
+                                            searchDepth: Int,
+                                            remainingPaths: List[(VaultPath, Int)],
+                                            credentialId: Option[String],
+                                            explorers: List[RemoteLayerExplorer],
+                                            reportMutable: ListBuffer[String])(implicit ec: ExecutionContext) =
+    explorationResultOfPath match {
+      case Full(layersWithVoxelSizes) =>
+        val sameParentRemainingPaths = remainingPaths.filter(_._1.parent == path.parent)
+        // Try to look for additional layers in sibling directories.
+        // TODO: Problem! This runs in parallel with other sub dirs
```
fm3, line 139: TODO?

webknossos-datastore/app/com/scalableminds/webknossos/datastore/explore/ExploreRemoteLayerService.scala
```diff
@@ -106,26 +107,93 @@ class ExploreRemoteLayerService @Inject()(dataVaultService: DataVaultService,
...
-            Fox.successful(layersWithVoxelSizes)
-          case f: Failure =>
-            reportMutable += s"Error when reading $remotePath as ${currentExplorer.name}: ${Fox.failureChainAsString(f)}"
-            exploreRemoteLayersForRemotePath(remotePath, credentialId, reportMutable, remainingExplorers)
-          case Empty =>
-            reportMutable += s"Error when reading $remotePath as ${currentExplorer.name}: Empty"
-            exploreRemoteLayersForRemotePath(remotePath, credentialId, reportMutable, remainingExplorers)
-        }
+    remotePathsWithDepth match {
+      case Nil =>
```
fm3, line 166:
```suggestion
      case List.empty =>
```
(I think that is equivalent, but more readable. Do retest though…)

webknossos-datastore/app/com/scalableminds/webknossos/datastore/explore/ExploreRemoteLayerService.scala
```diff
@@ -106,26 +107,93 @@ class ExploreRemoteLayerService @Inject()(dataVaultService: DataVaultService,
...
+        Fox.empty
+      case (path, searchDepth) :: remainingPaths =>
+        if (searchDepth > MAX_RECURSIVE_SEARCH_DEPTH) return Fox.empty
```
fm3, line 169: please avoid `return`. Always use full if/else. Compare https://tpolecat.github.io/2014/05/09/return.html

webknossos-datastore/app/com/scalableminds/webknossos/datastore/explore/ExploreRemoteLayerService.scala
```diff
@@ -106,26 +107,93 @@ class ExploreRemoteLayerService @Inject()(dataVaultService: DataVaultService,
...
+                  handleExploreResult(_, explorer, path)
+                }
+                .toFox
+            }
+          })
+          .map(explorationResults => Fox.firstSuccess(explorationResults.map(_.toFox)))
+          .toFox
+          .flatten
+          .futureBox
+          .flatMap(
```
fm3, line 188: Phew, that’s a mouthful. Maybe extracting some of these into functions with annotated return types could help with understanding what’s going on here?

webknossos-datastore/app/com/scalableminds/webknossos/datastore/explore/ExploreRemoteLayerService.scala
```diff
@@ -106,26 +107,93 @@ class ExploreRemoteLayerService @Inject()(dataVaultService: DataVaultService,
...
-          case Empty =>
-            reportMutable += s"Error when reading $remotePath as ${currentExplorer.name}: Empty"
-            exploreRemoteLayersForRemotePath(remotePath, credentialId, reportMutable, remainingExplorers)
-        }
+    remotePathsWithDepth match {
+      case Nil =>
+        Fox.empty
+      case (path, searchDepth) :: remainingPaths =>
+        if (searchDepth > MAX_RECURSIVE_SEARCH_DEPTH) return Fox.empty
+        implicit val implReportMutable: ListBuffer[String] = reportMutable
```
fm3, line 170: I’m not sure if the more compact style gained here is worth making this implicit, creating more “magic” that is going on here 😬

webknossos-datastore/app/com/scalableminds/webknossos/datastore/explore/ExploreRemoteLayerService.scala
```diff
@@ -106,26 +107,93 @@ class ExploreRemoteLayerService @Inject()(dataVaultService: DataVaultService,
...
+    case f: Failure =>
+      reportMutable += s"Error when reading $path as ${explorer.name}: ${Fox.failureChainAsString(f)}"
+      Fox.empty
+    case Empty =>
+      reportMutable += s"Error when reading $path as ${explorer.name}: Empty"
+      Fox.empty
+
+  }
+
+  private def handleExplorationResultOfPath(explorationResultOfPath: Box[List[(DataLayerWithMagLocators, VoxelSize)]],
```
fm3, line 128: Could you add a return type annotation and a comment summarizing the logic of this function? i.e. continue with remainingPaths unless something good was already found?

fm3 (changes requested):
Cool stuff, thanks for looking into this!

To be honest, I had a bit of a hard time reading the new code in ExploreRemoteLayerService. This is complicated nested logic, and a lot is happening in very compact code. I think that a few extractions into well-named private methods could already help a lot.

Also, please have a look at my individual comments. Feel free to ask if anything is unclear :)

frontend/javascripts/admin/dataset/dataset_add_remote_view.tsx
```diff
@@ -394,12 +394,14 @@ function AddRemoteLayer({
     if (userInput.startsWith("https://") || userInput.startsWith("http://")) {
       setSelectedProtocol("https");
     } else if (userInput.startsWith("s3://")) {
-      setSelectedProtocol("s3");
+      setSelectedProtocol("s3"); // Unused
```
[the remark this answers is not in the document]
MichaelBuessemeyer: In case the `selectedProtocol` is set to "s3", nothing happens in the frontend code and this info is not even used for the backend call. So it is technically correct to call `setSelectedProtocol("s3");` here, but no line in this file somehow "reacts" to the protocol being s3.

webknossos-datastore/app/com/scalableminds/webknossos/datastore/datavault/FileSystemDataVault.scala
MichaelBuessemeyer: I hope that's correct now :see_no_evil:

webknossos-datastore/app/com/scalableminds/webknossos/datastore/explore/ExploreRemoteLayerService.scala
MichaelBuessemeyer: `List.empty` leads to the following compilation error

> [error] /home/michael/ScalableMinds/webknossos/webknossos-datastore/app/com/scalableminds/webknossos/datastore/explore/ExploreRemoteLayerService.scala:170:17: stable identifier required, but scala.`package`.List.empty found.
[error]       case List.empty =>
[error]                 ^

fm3: ok then nevermind :)

frontend/javascripts/admin/dataset/dataset_add_remote_view.tsx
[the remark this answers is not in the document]
fm3: Ok, if that is really intended, maybe you could expand the comment slightly to make it explain what you just explained to me :) unused could also mean is never called, or it may indicate an open todo.
Something like `Note: the view will not change in case s3 is selected`?
I thought the authentication options are shown in different formats depending on the selected protocol. Does that mean that when s3 is selected, the same options are shown that were shown before the change? Is that what we want?

webknossos-datastore/app/com/scalableminds/webknossos/datastore/explore/ExploreRemoteLayerService.scala
MichaelBuessemeyer: I introduced this search for additional layers before we moved the explorers of more elaborate dataset formats ahead (like we moved NgffExplorer ahead of the zarr explorer) to ensure finding all missing datalayers.

But maybe this is not needed anymore:

I see the following problems here: 
1. In case we have a more elaborate dataset format listing all layers in a config file like `.zarray` one could have the following folder / nested key structure:
- top level dir
  - dataset 1
    - .zarray
  - dataset 2
    - .zarray
In that case the code looking for a sibling layer is likely to find dataset 1 and dataset 2, concat them and treat them as one dataset. This seems like a bug to me, but I would need to create a test scenario to confirm this.

2. In case we have a simple dataset format without a config file and only layers:
- top level dir
  - dataset 1
    - layer 1
    - layer 2
    - layer 3
  - dataset 2
    - some layers

In this scenario, the code looking for sibling layers is likely useful, as it finds layer 1, layer 2, layer 3 of dataset 1 in one go.  But this also needs testing. @fm3 Do you maybe know how to test this?

Thus, I'd say the sibling exploration is useful in the second case but should be avoided in the first case. Maybe we can add a boolean instance val to the explorers that mark them whether they are for simple or advanced formats and only use the sibling exploration for simple dataset format explorers?

webknossos-datastore/app/com/scalableminds/webknossos/datastore/datavault/GoogleCloudDataVault.scala
MichaelBuessemeyer: I used the default page size limit here (removed the `Storage.BlobListOption.pageSize(MAX_EXPLORED_ITEMS_PER_LEVEL)`)  option

webknossos-datastore/app/com/scalableminds/webknossos/datastore/explore/ExploreRemoteLayerService.scala
MichaelBuessemeyer: Maybe the newest version is easier to read? Each method has something like max 8 lines as a method body

webknossos-datastore/app/com/scalableminds/webknossos/datastore/explore/ExploreRemoteLayerService.scala
fm3: Yes, an `exploreSiblings=true` bool seems reasonable. I’m afraid I don’t have a test dataset. Maybe you could create one in the local file system by modifying an existing one?

I would assume the siblings of the top-level dir (the one of the uri the user originally submitted) are not searched. Is this already so?

MichaelBuessemeyer: > I’m afraid I don’t have a test dataset. Maybe you could create one in the local file system by modifying an existing one?

Oh right. That seems like an easy way to test :)

> I would assume the siblings of the top-level dir (the one of the uri the user originally submitted) are not searched. Is this already so?

Yeah, this should be the case, but I can test this.

MichaelBuessemeyer:
There is a problem I noticed: Exploring `gs://iarpa_microns/minnie/minnie65/` takes very long as it contains a lot of datasets / subfolders. The current code supports at max depth of 8 and max 10 subfolders are explored in each level. This makes up to a limit of 10*10*10.*..*10 =10^8 explored subfolders -> This potentially takes up a lot of time :woozy_face:. Should I decrease both limits?

Another thing: The exploration seems to not yield any results on many folders that seem to sound like a valid dataset / layer. But I cannot test this, as I was unable to find a proper gcs explorer in a reasonable time frame :/

And I also have the feeling that the sibling exploration is not working as I / we want it to:
If a dataset is like this:
 - dataset
   - layer 1
     - mag 1
       - shard 1
         - actual data files
       - shard 2
         - actual data files
       - ...
     - mag 2
       - shard 1
         - actual data files
       - shard 2
         - actual data files
       - ...
   - layer 2   
     - mag 1
       - shard 1
         - actual data files
       - shard 2
         - actual data files
       - ...
     - mag 2
       - shard 1
         - actual data files
       - shard 2
         - actual data files
       - ...
       
then the current exploration would find dataset/layer 1/mag 1 as a layer and then would treat dataset/layer 1/mag 2 as a sibling and thus an additional layer although it is a different mag of the layer...
How should we handle this? Should the sibling check be something like: 
`val sameParentRemainingPaths = remainingPaths.filter(_._1.parent == path.parent.parent)`.

Moreover, I suspect the code to keep on running recursive search in the sibling directories. IMO we should restrict this to a max nesting level of 2, as the data of sibling layers should be at the same nesting level imo.

fm3:
Yes, 10^8 is too big as a maximum. How about 10^3? How long does the tested dataset take then?

Not sure I fully understood the problem with mixed siblings/mags, but yes, I guess we can restrict this more, with the assumption that the structure is not arbitrarily mixed.

MichaelBuessemeyer:
> How about 10^3

Done: Max 3 depth & max 10 items per level. But it seems a little restrictive in depth :/. Maybe something like 4 depth & 6 items per level (max 1296) or 5 depth and 5 items per level (max 3125) would be better :thinking:?

> How long does the tested dataset take then?

I tested twice. Once it took 30 sec, the other run it took 60 sec :man_shrugging:. Still quite long.

> Not sure I fully understood the problem with mixed siblings/mags, but yes, I guess we can restrict this more, with the assumption that the structure is not arbitrarily mixed

The "Sibling exploration" was now removed and an issue for this was created (see #7924). See also: https://scm.slack.com/archives/C5AKLAV0B/p1721129347974129 where it was decided to postpone this feature and move it to a separate issue.

fm3:
I think 60s is acceptable. I guess it will only take this long if there are actually this many folders, right? Maybe you could test a couple more from our example remote dataset table. I have no strong opinion on depth 3 or 4. 3125 tests seems like too much, though.
Did the sibling test removal help with runtime by any chance?

MichaelBuessemeyer:
> I think 60s is acceptable. I guess it will only take this long if there are actually this many folders, right?

Yeah

> Did the sibling test removal help with runtime by any chance?

Oh right, that might explain the difference between 30 sec and 60 sec :thinking: 

> Maybe you could test a couple more from our example remote dataset table. I have no strong opinion on depth 3 or 4. 3125 tests seems like too much, though.

I'll do some more :)

docs/datasets.md
```diff
@@ -57,7 +57,7 @@ With other converters, you may need to add the layers separately.
 1. From the *Datasets* tab in the user dashboard, click the *Add Dataset* button.
 2. Select the *Add Remote Dataset* tab
 3. For each layer, provide some metadata information:  
-    - a URL or domain/collection identifier to locate the dataset on the remote service (supported protocols are HTTPS, Amazon S3 and Google Cloud Storage).
+    - a URL or domain/collection identifier to locate the dataset on the remote service (supported protocols are HTTPS, Amazon S3 and Google Cloud Storage). In case the URL or domain/collection identifier do not directly point to a dataset, WEBKNOSSOS will try to locate any dataset in subfolders / sub-collection identifiers.
```
fm3, line 60: I think this is too technical, users will expect their dataset to be found even if the uri does not point to the array itself. I’d undo this particular change

test/backend/DataVaultTestSuite.scala
```diff
@@ -134,6 +134,8 @@ class DataVaultTestSuite extends PlaySpec {
       class MockDataVault extends DataVault {
         override def readBytesAndEncoding(path: VaultPath, range: RangeSpecifier)(
             implicit ec: ExecutionContext): Fox[(Array[Byte], Encoding.Value)] = ???
+
+        override def listDirectory(path: VaultPath,maxItems: Int)(implicit ec: ExecutionContext): Fox[List[VaultPath]] = ???
```
fm3, line 138:
```suggestion
        override def listDirectory(path: VaultPath, maxItems: Int)(implicit ec: ExecutionContext): Fox[List[VaultPath]] = ???
```
did the formater not lament this? :thinking:

webknossos-datastore/app/com/scalableminds/webknossos/datastore/datavault/FileSystemDataVault.scala
```diff
@@ -53,6 +50,29 @@ class FileSystemDataVault extends DataVault {
       }
     } else Fox.empty
 
+  override def listDirectory(path: VaultPath, maxItems: Int)(implicit ec: ExecutionContext): Fox[List[VaultPath]] =
+    vaultPathToLocalPath(path).map(localPath => {
+      if (!Files.isDirectory(localPath)) return Fox.successful(List.empty)
```
fm3, line 55: Please avoid `return` in scala, always use full if/else instead.

webknossos-datastore/app/com/scalableminds/webknossos/datastore/datavault/GoogleCloudDataVault.scala
```diff
@@ -72,6 +73,15 @@ class GoogleCloudDataVault(uri: URI, credential: Option[GoogleServiceAccountCred
     } yield (bytes, encoding)
   }
 
+  override def listDirectory(path: VaultPath, maxItems: Int)(implicit ec: ExecutionContext): Fox[List[VaultPath]] = {
+    val objName = path.toUri.getPath.tail
+    val blobs = storage.list(bucket, Storage.BlobListOption.prefix(objName), Storage.BlobListOption.currentDirectory())
+    val subDirectories = blobs.getValues.asScala.toList.filter(_.isDirectory).take(maxItems)
+    val paths = subDirectories.map(dirBlob =>
+      new VaultPath(new URI(s"${uri.getScheme}://$bucket/${dirBlob.getBlobId.getName}"), this))
```
fm3, line 81: can any of these synchronous functions throw exceptions? If so, please wrap them (or the whole block) in a tryo. Then, the Fox.successful can be removed (as tryo will create a Box, which is implicitly converted to Fox)

webknossos-datastore/app/com/scalableminds/webknossos/datastore/datavault/S3DataVault.scala
```diff
@@ -82,10 +84,38 @@ class S3DataVault(s3AccessKeyCredential: Option[S3AccessKeyCredential], uri: URI
...
+      (bytes, encodingString) <- performGetObjectRequest(request)
       encoding <- Encoding.fromRfc7231String(encodingString)
     } yield (bytes, encoding)
 
+  override def listDirectory(path: VaultPath, maxItems: Int)(implicit ec: ExecutionContext): Fox[List[VaultPath]] =
+    for {
+      prefixKey <- Fox.box2Fox(S3DataVault.objectKeyFromUri(path.toUri))
+      s3SubPrefixKeys <- getObjectSummaries(bucketName, prefixKey, maxItems)
+      vaultPaths <- Fox.successful(
+        s3SubPrefixKeys.map(key => new VaultPath(new URI(s"${uri.getScheme}://$bucketName/$key"), this)))
```
fm3, line 96: Fox.successful in combination with `<-` is meaningless, the arrow can be replaced by `=`, combined with omitting the `Fox.successful`.
However, I think a tryo may be a good choice here, since `new URI` can throw exceptions.

webknossos-datastore/app/com/scalableminds/webknossos/datastore/explore/ExploreRemoteLayerService.scala
```diff
@@ -106,26 +107,95 @@ class ExploreRemoteLayerService @Inject()(dataVaultService: DataVaultService,
...
+      Fox.successful(layersWithVoxelSizes)
+    case f: Failure =>
+      reportMutable += s"Error when reading $path as ${explorer.name}: ${Fox.failureChainAsString(f)}"
+      Fox.empty
+    case Empty =>
+      reportMutable += s"Error when reading $path as ${explorer.name}: Empty"
+      Fox.empty
+  }
+
+  // In case a data layers were found, explore sibling directories for more layers, else continue recursive exploration.
```
fm3, line 174: I thought we’re not doing siblings anymore?

webknossos-datastore/app/com/scalableminds/webknossos/datastore/explore/N5MultiscalesExplorer.scala
```diff
@@ -17,6 +17,8 @@ class N5MultiscalesExplorer(implicit val ec: ExecutionContext) extends RemoteLay
 
   override def name: String = "N5 Multiscales"
 
+  override def exploreSiblings: Boolean = false
```
fm3, line 20: I thought we’re not doing siblings anymore?

fm3 (commented):
Yes, this is more readable, thanks! Of course, the logic is still pretty complicated, and with everything being async, it will always be a bit hard to read, but the task _is_ fairly complicated, and I think this is certainly better than before :)

I added a couple more comments

webknossos-datastore/app/com/scalableminds/webknossos/datastore/explore/ExploreRemoteLayerService.scala
MichaelBuessemeyer: Sorry, I simply forgot to remove this comment. Originally it was above the method doing the sibling exploration

webknossos-datastore/app/com/scalableminds/webknossos/datastore/explore/N5MultiscalesExplorer.scala
MichaelBuessemeyer: Also forgot to remove this here

MichaelBuessemeyer (commented):
Thanks for your feedback. I hopefully covered everything :)

test/backend/DataVaultTestSuite.scala
fm3: I guess the tests aren’t covered by the formatter config. That’s not a big deal imo :shrug:

webknossos-datastore/app/com/scalableminds/webknossos/datastore/datavault/S3DataVault.scala
```diff
@@ -82,10 +84,38 @@ class S3DataVault(s3AccessKeyCredential: Option[S3AccessKeyCredential], uri: URI
...
+      Fox.successful(s3SubPrefixes)
+    } catch {
+      case e: AmazonServiceException =>
+        e.getStatusCode match {
+          case 404 => Fox.empty
+          case _   => Fox.failure(e.getMessage)
+        }
+      case e: Exception => Fox.failure(e.getMessage)
+    }
+
```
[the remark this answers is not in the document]
fm3: tryo is a shortcut, it will always produce a Box.Failure in case of any exception. In this code, however, we want to create different results based on a parameter of the exception (Fox.empty for status code 404), so we need the full try/catch to implement that custom logic.

fm3 (approved):
Good Stuff :) I’d say this is good to go

Could you update the PR description before merging? (The TODOs are solved, right?)

PR #7922, opened 2024-07-15
Minor upgrades for backend dependencies, scala, sbt, run codespell

fm3:
CI should be enough


util/src/main/scala/com/scalableminds/util/cache/LRUConcurrentCache.scala
```diff
@@ -7,7 +7,7 @@ trait LRUConcurrentCache[K, V] {
 
   private val cache = new java.util.LinkedHashMap[K, V]() {
     override def removeEldestEntry(eldest: java.util.Map.Entry[K, V]): Boolean =
-      if (size > maxEntries) {
+      if (this.size > maxEntries) {
```
fm3, line 10: Got a compiler warning here after scala upgrade, arguing that `size` was ambiguous

PR #7923, opened 2024-07-16
Forbid editing other orga’s public annotations

### LINKED ISSUE #7914: allowUpdate should be false for annotations of other orgas, even if isPublic and editableByOthers are true
This rare combination leads to allowUpdate=true in the annotation restrictions. This should not be the case.

fm3:
### Issues:
- fixes #7914

### Steps to test:
- create two orgas
- create an annotation for a public dataset, set it to public and to others-may-edit
- another user from the same organization should be able to edit it (respecting mutex)
- another user from a *different* organization should be able to view it but never edit it

------

### DIFF
--- a/CHANGELOG.unreleased.md
+++ b/CHANGELOG.unreleased.md
@@ -16,6 +16,7 @@ For upgrade instructions, please check the [migration guide](MIGRATIONS.released
 
 ### Changed
 - The warning about a mismatch between the scale of a pre-computed mesh and the dataset scale's factor now also considers all supported mags of the active segmentation layer. This reduces the false posive rate regarding this warning. [#7921](https://github.com/scalableminds/webknossos/pull/7921/)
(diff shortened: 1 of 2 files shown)


MichaelBuessemeyer:
Works, but the navbar shows a tag like the annotation is locked by a user (which was not the case while I tested). That might be confusing.

I'll give this a quick check whether I can hide this 
image

MichaelBuessemeyer:
> I'll give this a quick check whether I can hide this

Ok this actually rather difficult to do. The frontend needs to know whether the current user belongs to the same orga as the annotation (owner). But currently that's not the case. And I do not want to remove this info to the user as when the user is know it is actually useful.  

Cf.
https://github.com/scalableminds/webknossos/blob/9f2eb107862a74d2c319790015d072af61ff4d57/frontend/javascripts/navbar.tsx#L889-L897

Let me check tomorrow, whether maybe `blockedByUser == null` can be used to detect this case and avoid the "Locked by unknown user" this way.

MichaelBuessemeyer (approved):
LGTM, My previous mentioned improvement is deferred to a new issue

PR #7928, opened 2024-07-17
Save expanded/collapsed state of segment groups

### LINKED ISSUE #7825: Reloading the page should retain collapsed-state of (segment) groups
Reloading the page should retain collapsed-state of (segment) groups. Also meshes should re-appear or at least be listed.

dieknolle3333:
### Steps to test:
- Open an annotation and go to segments tab. Create a group hierarchy (around three levels min) with some segments in each group.
- Expand and collapse some groups, remember state.
- Change to another tab and back. The state (groups expanded/collapsed) should be the same.
- Reload the page. The state should still be the same 
- Test that old behavior is still working. Expanding and collapsing via arrow icon, expansion/collapse of all subgroups etc.

### TODOs:
- [x] fix bug that expand children on root doesnt work after collapsing all subgroups

### Issues:
- fixes #7825 

------

### DIFF
--- a/frontend/javascripts/oxalis/view/right-border-tabs/segments_tab/segments_view.tsx
+++ b/frontend/javascripts/oxalis/view/right-border-tabs/segments_tab/segments_view.tsx
@@ -367,6 +367,25 @@ function renderEmptyMeshFileSelect() {
   );
 }
 
+const getExpandedKeys = (treeData: TreeGroup[]) => {
+  return treeData.reduce((expandedKeysAcc: string[], node) => {
+    if (node.isExpanded || node.isExpanded == null) {
+      expandedKeysAcc.push(getKeyForGroupId(node.groupId));
+    }
+    if (node.children.length > 0) expandedKeysAcc.push(...getExpandedKeys(node.children));
+    return expandedKeysAcc;
+  }, []);
+};
+
+const getExpandedKeysWithRoot = (segmentGroups: TreeGroup[], isRootGroupExpanded: boolean) => {
+  const expandedGroups = getExpandedKeys(segmentGroups);
+  return isRootGroupExpanded
+    ? [getKeyForGroupId(MISSING_GROUP_ID), ...expandedGroups]
+    : expandedGroups;
+};
+
+const getKeyForGroupId = (groupId: number) => `group-${groupId}`;
+
 function constructTreeData(
   groups: { name: string; groupId: number; children: SegmentGroup[] }[],
   groupToSegmentsMap: Record<number, Segment[]>,
@@ -445,35 +464,11 @@ class SegmentsView extends React.Component<Props, State> {
     Store.dispatch(ensureSegmentIndexIsLoadedAction(this.props.visibleSegmentationLayer?.name));
 
     this.setState({
-      expandedGroupKeys: this.getExpandedKeysWithRoot(),
+      expandedGroupKeys: getExpandedKeysWithRoot(this.props.segmentGroups, this.state.isRootGroupExpanded),
     });
   }
 
-  getExpandedKeys = (treeData: TreeGroup[]) => {
-    return treeData.reduce((expandedKeysAcc: string[], node) => {
-      if (node.isExpanded || node.isExpanded == null) {
-        expandedKeysAcc.push(this.getKeyForGroupId(node.groupId));
-      }
-      if (node.children.length > 0) expandedKeysAcc.push(...this.getExpandedKeys(node.children));
-      return expandedKeysAcc;
-    }, []);
-  };
-
-  getExpandedKeysWithRoot = () => {
-    const expandedGroups = this.getExpandedKeys(this.props.segmentGroups);
-    return this.state.isRootGroupExpanded
-      ? [this.getKeyForGroupId(MISSING_GROUP_ID), ...expandedGroups]
-      : expandedGroups;
-  };
-
   componentDidUpdate(prevProps: Props) {
-    if (this.props.segmentGroups[0] !== prevProps.segmentGroups[0]) {
-      // TODO_c find faster way to determine if update is necessary!
-      this.setState({
-        expandedGroupKeys: this.getExpandedKeysWithRoot(),
-      });
-    }
-
     if (prevProps.visibleSegmentationLayer !== this.props.visibleSegmentationLayer) {
       Store.dispatch(
         maybeFetchMeshFilesAction(this.props.visibleSegmentationLayer, this.props.dataset, false),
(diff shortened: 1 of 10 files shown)


frontend/javascripts/oxalis/view/right-border-tabs/segments_tab/segments_view.tsx
```diff
@@ -1842,6 +1954,55 @@ class SegmentsView extends React.Component<Props, State> {
...
+      children.filter((childNode) => this.state.expandedGroupKeys?.includes(childNode)).length ===
+      children.length;
+    const isGroupItselfExpanded = this.state.expandedGroupKeys?.includes(
+      this.getKeyForGroupId(groupId),
+    );
+    if (areAllChildrenExpanded && isGroupItselfExpanded) {
+      return null;
+    }
+    const allExpandedGroups = children.concat(this.state.expandedGroupKeys ?? []);
+    // still not good TODO_c
```
dieknolle3333, line 1969: @dieknolle3333 remove comment

fm3 (commented):
Backend LGTM :) Testing worked nicely, backwards compatibility looks OK. As mentioned in [Slack](https://scm.slack.com/archives/C5AKLAV0B/p1721316115054539?thread_ts=1721305479.455829&cid=C5AKLAV0B), the root group gets expanded again after reload, raising the question if it needs to be collapsible in the first place, but I’d say that’s not particularly important.

Please add a changelog entry :)

dieknolle3333:
I am not 100% sure, but I feel like expanding and collapsing got slower. I compared it to webknossos.org and I am still not super sure. But I wanted to mention it. Maybe there is a faster way to get the current expanded/collapsed group state from the segmentGroups?

frontend/javascripts/oxalis/view/right-border-tabs/segments_tab/segments_view.tsx
```diff
@@ -461,6 +505,67 @@ class SegmentsView extends React.Component<Props, State> {
     }
   }
 
+  onExpandTree = (expandedKeys: Key[]) => {
+    this.expandGroups(expandedKeys);
+  };
```
philippotto, line 510: can't this method be removed? `expandGroups` should be identical.

frontend/javascripts/oxalis/view/right-border-tabs/segments_tab/segments_view.tsx
```diff
@@ -461,6 +505,67 @@ class SegmentsView extends React.Component<Props, State> {
...
+
+  getSubGroupsAsTreeNodes = (groupId: number) => {
```
philippotto, line 512: why `asTreeNodes`? doesn't this method return the keys?

frontend/javascripts/oxalis/view/right-border-tabs/segments_tab/segments_view.tsx
```diff
@@ -461,6 +505,67 @@ class SegmentsView extends React.Component<Props, State> {
...
+          ...group,
+          isExpanded: true,
+        };
+      } else {
+        return {
+          ...group,
+          isExpanded: false,
+          // Close all groups that are not in the expanded list so this method
+          // can be called for every update, e.g. when a group is collapsed.
+        };
```
philippotto, line 547: this way, you don't create new objects for all groups where the isExpanded property does not change:

```suggestion
      const isExpanded = expandedKeySet.has(getKeyForGroupId(group.groupId));
      if (group.isExpanded === isExpanded) {
        return group;
      } else {
        return {
          ...group,
          isExpanded,
        };
      }
```

frontend/javascripts/oxalis/view/right-border-tabs/segments_tab/segments_view.tsx
```diff
@@ -461,6 +505,67 @@ class SegmentsView extends React.Component<Props, State> {
...
+      }
+    });
+    Store.dispatch(setSegmentGroupsAction(newGroups, this.props.visibleSegmentationLayer?.name));
+  };
+
+  collapseGroups = (groupsToCollapse: Key[]) => {
```
philippotto, line 553: why doesn't this method handle the root group like the `expandGroups` method does? is it not necessary in this case?

philippotto: also: the two methods differ in their semantics of the parameters. for `collapseGroups` I understand it so that the referenced groups will be collapsed (and other collapsed groups will stay as before). for `expandGroups`, it seems like only the referenced groups should be expanded (and all the others should not be expanded). can this be unified?

frontend/javascripts/oxalis/view/right-border-tabs/segments_tab/segments_view.tsx
```diff
@@ -461,6 +505,67 @@ class SegmentsView extends React.Component<Props, State> {
...
+    const newGroups = mapGroups(this.props.segmentGroups, (group) => {
+      if (groupsToCollapseSet.has(getKeyForGroupId(group.groupId))) {
+        return {
+          ...group,
+          isExpanded: false,
+        };
+      } else {
+        return group;
+      }
+    });
```
philippotto, line 565: potentially adapt this like suggested for the sibling method. however, this would change the behavior (see my other comment).

frontend/javascripts/oxalis/view/right-border-tabs/segments_tab/segments_view.tsx
```diff
@@ -1842,6 +1953,49 @@ class SegmentsView extends React.Component<Props, State> {
...
 
+  getExpandSubgroupsItem(groupId: number) {
+    const children: Key[] = this.getSubGroupsAsTreeNodes(groupId);
+    const expandedGroupsSet = new Set(this.state.expandedGroupKeys);
+    const areAllChildrenExpanded = children.every((childNode) => expandedGroupsSet.has(childNode));
+    const isGroupItselfExpanded = expandedGroupsSet.has(getKeyForGroupId(groupId));
+    if (areAllChildrenExpanded && isGroupItselfExpanded) {
+      return null;
+    }
+    const allExpandedGroups = children.concat(this.state.expandedGroupKeys ?? []);
```
philippotto, line 1964: why `?? []`? isn't `this.state.expandedGroupKeys` always defined?

also this variable is only needed in the `onClick` handler. it can be moved there.

frontend/javascripts/oxalis/view/right-border-tabs/segments_tab/segments_view.tsx
```diff
@@ -1842,6 +1953,49 @@ class SegmentsView extends React.Component<Props, State> {
...
+      icon: <ExpandAltOutlined />,
+      label: "Expand all subgroups",
+    };
+  }
+
+  getCollapseSubgroupsItem(groupId: number) {
+    const children = this.getSubGroupsAsTreeNodes(groupId);
+    const expandedKeySet = new Set(this.state.expandedGroupKeys);
+    const areAllChildrenCollapsed =
+      children.filter((childNode) => expandedKeySet.has(childNode)).length === 0;
```
philippotto, line 1982:
```suggestion
    const areAllChildrenCollapsed =
      children.every((childNode) => !expandedKeySet.has(childNode));
```

`every` and `some` are faster on average because they can abort the iteration early as soon as the predicate cannot change anymore.

frontend/javascripts/oxalis/view/right-border-tabs/segments_tab/segments_view.tsx
```diff
@@ -1842,6 +1953,49 @@ class SegmentsView extends React.Component<Props, State> {
...
+    const isGroupItselfCollapsed = !expandedKeySet.has(getKeyForGroupId(groupId));
+    if (areAllChildrenCollapsed || isGroupItselfCollapsed) {
+      return null;
+    }
+    return {
+      key: "collapseAll",
+      disabled: !this.props.allowUpdate,
```
philippotto, line 1989: I wouldn't disable collapse and expand UI actions, because users might be interested in inspecting the hierarchy. it shouldn't be saved anyway and there will no be data loss when it's not saved.

philippotto (changes requested):
cool :) I left some feedback.

frontend/javascripts/oxalis/view/right-border-tabs/segments_tab/segments_view.tsx
dieknolle3333: you're right, the methods differ. this is because `expandGroups` is called by the Tree component _and_ the `expandSubgroups` menu entry. The tree component also calls this if groups are collapsed, by removing the collapsed groups trees from the list. Thus `isExpanded` has to be set to false for these groups. Other than that, the root group is handled differently because on `expand all subgroups` the parent group is expanded if it was collapsed before. This is different for expanded groups, their state doesn't change if their children are collapsed.
I could use three methods: one for the Tree component and two for "collapse/expand subgroups". And then comment what I said above. Do you think that this might be a good solution? @philippotto

dieknolle3333:
>Otherwise, all the changes of the first PR will also be shown here in the diff, too.

I see. So I set it to the base branch although I will merge it into master eventually. And then I can probably set it to master once the base branch is merged and I have rebased the commits of this PR, right?

frontend/javascripts/oxalis/view/right-border-tabs/segments_tab/segments_view.tsx
dieknolle3333: yes! this is a remnant of the old PR

philippotto:
> I see. So I set it to the base branch although I will merge it into master eventually. And then I can probably set it to master once the base branch is merged and I have rebased the commits of this PR, right?

Exactly! Even better, GitHub will automatically set the base to "master" once the first PR is merged.

frontend/javascripts/oxalis/view/right-border-tabs/segments_tab/segments_view.tsx
philippotto: ah, I understand. two methods are fine then. I'd the `expandGroups` method to `setExpandedGroups`. That way, it should be clearer that the semantics is different to `collapseGroups`.

> Other than that, the root group is handled differently because on expand all subgroups the parent group is expanded if it was collapsed before. This is different for expanded groups, their state doesn't change if their children are collapsed.

A comment for that would be good, yes :+1:

dieknolle3333:
@philippotto because of the different base branch I rebased my changes on master. The numbers of the diff are slightly different (used to be +170, -12). I hope that is somewhat expected. It still works well, so I hope this all worked.

philippotto (approved):
Great :tada: Performance feels still a bit laggy, but this is unrelated to this PR. Let's merge it and iterate later :)

PR #7929, opened 2024-07-17
Include docker-compose steps in hosting for localized database backups

aaronkanzer:
Provides docker-compose steps that allow an end user to easily access and run commands for data integrity and reliability against a local fossil-db and postgres instances

Cc @kabilar 

N/A

### Steps to test:
- Create a deployment of WebKNOSSOS via instructions on https://docs.webknossos.org/webknossos/installation.html
- Once all containers are healthly on your local EC2, run `docker-compose run --rm fossil-db-backup`
- Confirm backup is present in `/persistent/fossildb/backup/private` on EC2 instance
- Run `docker-compose run --rm fossil-db-restore` to observe successful restoration
(additionally, you can destroy the local `fossildb` instance and see if you can re-populate via `restore`)

### TODOs:
- [x] Include similar docker-compose scripts for `pg_dump` equivalent 

### Issues:
- fixes #

------

### DIFF
--- a/tools/hosting/docker-compose.yml
+++ b/tools/hosting/docker-compose.yml
@@ -135,6 +135,15 @@ services:
     networks:
       - webknossos_default
 
+  # Postgres backup
+  postgres-backup:
+    image: postgres
+    command: [ "/bin/bash", "-c", "PGPASSWORD=postgres pg_dump -Fc -h postgres -U postgres webknossos | gzip > /backups/backup_$(date +%Y-%m-%d_%H-%M).sql.gz" ]
+    volumes:
+      - ./backups:/backups
+    depends_on:
+      - postgres
+
 # Explicitly declare networks for fossil-db-backup connectivity
 networks:
   webknossos_default:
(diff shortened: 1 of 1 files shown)


normanrz (commented):
Thanks for your PR. I agree that it would be useful to include the backup commands in the docker-compose.yaml. I wonder if we should add a `backup.sh` to also zip up the fossildb backup with a timestamped filename?

I don't think the explicit networks are required and would prefer to keep them out.

In order to merge PRs from contributors, we need the CLA check to pass. For that all commits need to be authored with matching email info. Sorry for the inconvenience.

philippotto:
Thanks again for your contribution! I'm closing this PR due to inactivity, but please feel free to re-open when you have a chance to revisit it. Note that we cannot take over this PR due to the license issue @normanrz mentioned above:

> In order to merge PRs from contributors, we need the CLA check to pass. For that all commits need to be authored with matching email info. Sorry for the inconvenience.

Issue #7606, opened 2024-02-05
Select User Bounding Boxes from Viewports


fm3:
It would be nice to jump from the bounding box in the viewport (data + 3d) to the list in the bboxes tab

PR #7935, opened 2024-07-19
Show clicked Bounding Box in  BB-Tab

MichaelBuessemeyer:
- https://allowhavingselectedboundingbox.webknossos.xyz

### Steps to test:
- Create an annotation of your favourite dataset.
- Use the bbox tool to create a lot of bounding boxes.
- Whenever a new bounding box is created, the bounding box tab should scroll to the bottom and highlight this bounding box.
- Create a lot of bounding boxes to get into a scrolling state in the bbox tab
- When a bounding box is clicked while the bbox tool is active, the bbox should also be highlighted.
- Same goes for resizing

### TODOs:
- [x] Scroll selected bbox into view. Maybe https://github.com/ant-design/ant-design/issues/6570 helps -> :-1: 

### Open Questions
- [x] Should the context menu support to highlight a bounding box? -> Not necessary now.

### Issues:
- fixes [#7606](https://github.com/scalableminds/webknossos/issues/7606)

------


daniel-wer (approved):
Nicely done, works very well :+1: 

One more thing that came to my mind, because the view is not the fastest having 20+ bounding boxes (also hasn't been before!): Now that an antd table is used one could try to set the `virtual` prop and check whether that improves performance. There's no need to do that in this PR, but if you'd like you can try.

frontend/javascripts/oxalis/view/components/setting_input_views.tsx
```diff
@@ -378,6 +378,7 @@ type UserBoundingBoxInputProps = {
   disabled: boolean;
   isLockedByOwner: boolean;
   isOwner: boolean;
+  id?: string | undefined;
```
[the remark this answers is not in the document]
MichaelBuessemeyer: Hmmm, my thought was to not enforce an id for each bounding box input element created and thus made it optional. I wanted to avoid in case the component is used elsewhere in the code, it must also need a unique id. 
-> But in the current version of the code base `UserBoundingBoxInput` is only used in the bbox tab in the table (where the id is needed).
=> I removed the optional from `id`. In case `UserBoundingBoxInput` is used elsewhere and an id is not necessary, it can be made optional again :)

frontend/stylesheets/trace_view/_right_menu.less
```diff
@@ -217,3 +217,22 @@
     height: 24px;
   }
 }
+
+.bounding-box-table {
```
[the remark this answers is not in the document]
MichaelBuessemeyer: I moved the + button into the table footer

MichaelBuessemeyer:
Sorry @daniel-wer, but I did some code adjustments that need your review:
- I moved the +-button into the footer of the table.
- No longer rendering an empty table in view mode; Instead, nothing is rendered.
- The `id` prop was removed from the bounding box inputs as a ref to the table supports scrolling the table to a specific element. I did not find this earlier in the docs (only found it while testing virtualizing the table). => The new highlighting does not have a scrolling animation anymore. Afaik this also goes for the other tabs, so I thought this would be fine.

I also tested making the table virtualized but stopped doing so, as it sadly requires setting a specified scroll width and height. In order to get the correct values of width and height I would have needed something like hardcoded calculation of the tab's current width and height. This is due to the tab being resizable. As I wanted to avoid this rather ugly code, I undid my changes and kept the table in a not virtualized state.

So this remains a future TODO.


Could you please recheck the newest version of this PR @daniel-wer

MichaelBuessemeyer:
And @daniel-wer what do you think about this:

>Should the context menu support to highlight a bounding box?

frontend/javascripts/oxalis/view/right-border-tabs/bounding_box_tab.tsx
```diff
@@ -102,22 +103,12 @@ export default function BoundingBoxTab() {
     APIJobType.EXPORT_TIFF,
   );
 
-  const bboxInputId = (id: number) => `bounding-box-input-${id}`;
-
   // biome-ignore lint/correctness/useExhaustiveDependencies: Always try to scroll the active bounding box into view.
```
daniel-wer, line 106: I'd expect this is no longer relevant?

daniel-wer (approved):
LGTM :+1: 

Thank you for testing the virtualization! I second your opinion that it's best to look at that as a followup.

> Should the context menu support to highlight a bounding box?

I don't think that is necessary as simply clicking on the bounding box will highlight it in the list.

frontend/javascripts/oxalis/view/right-border-tabs/bounding_box_tab.tsx
MichaelBuessemeyer: Yeah, this was needed to be able to select the active bounding box via a getElementById and then calling scrollIntoView. But luckly I found out that antd has its own method supporting this feature which does not need an id attached to the selected table entry. The relevant effect it directly below this comment and `bboxTableRef.current.scrollTo({ key: activeBoundingBoxId });` does all the magic needed to select and highlight the bbox item in the table

daniel-wer: Thank you for explaining but I was referring to the `biome-ignore` comment :sweat_smile: In any case, not a big deal :)

MichaelBuessemeyer: Oh :see_no_evil:, well according to biome the `bboxTableRef.current` dependency is unnecessary :man_shrugging:.

Sorry for that. I'll remove this in another pr :)

MichaelBuessemeyer: Please see https://github.com/scalableminds/webknossos/pull/7886#issuecomment-2255735830

PR #7453, opened 2023-11-22
Support remote datasets from s3 compliant data sources

frcroth:
### URL of deployed dev instance (used for testing):
- https://___.webknossos.xyz

### Steps to test:
- [Set up a minio server](https://min.io/docs/minio/linux/index.html)
- Create a bucket in that server and upload a dataset to it (do this via the web console)
- Create an access key on the server
- Explore this dataset with URL: "s3://\<minio-server-ip>:\<minio-server-port>/\<bucket-name>/\<root-key-of-the-dataset>"
- View dataset
- Also explore S3 datasets to ensure they still work

<details>
<summary> Alternative repl testing </summary>

```scala
import java.net.URI;
import com.scalableminds.webknossos.datastore.datavault.S3DataVault;
implicit val ec: scala.concurrent.ExecutionContext = scala.concurrent.ExecutionContext.global;
import com.scalableminds.webknossos.datastore.datavault.VaultPath;
val key = com.scalableminds.webknossos.datastore.storage.S3AccessKeyCredential(name="someKey", accessKeyId="FhXRPMiC9KWMBh9WXF54", secretAccessKey="9eVWQYZckmshjIMsebOBkztMqfHpC5t88bFWOMhc", user="", organization="");
val uri =  new URI("s3://10.191.1.63:8000/somezarr/e75_zarr_local");
val minioVault = new S3DataVault(Some(key), new URI("s3://10.191.1.63:8000/somezarr/e75_zarr_local"));
val vp = new VaultPath(uri, minioVault);
(vp / "color" / "1" / "credentials.json").readBytes().get("")
```

</details>

### TODOs:
- [ ] ...

### Issues:
- fixes #7432

------
(Please delete unneeded items, merge only when none are left open)
- [x] Updated [changelog](../blob/master/CHANGELOG.unreleased.md#unreleased)
- [ ] Considered [common edge cases](../blob/master/.github/common_edge_cases.md)
- [x] Needs datastore update after deployment

PR #7936, opened 2024-07-22
Upgrade to s3 client v2

### LINKED ISSUE #7913: Upgrade to AWS Client lib v2, use new async API?
We use AWS Client in the S3DataVault. A v2 of the lib has been released. We should have a look at wether we can benefit from integrating it. Since we only use very few calls to this lib, the effort should be manageable.

https://github.com/aws/aws-sdk-java-v2/#using-the-sdk

fm3:
The new s3 java client lib offers an async api :tada: 

- https://newsclient.webknossos.xyz

### Steps to test:
- Open an s3 remote dataset, should load as expected (hopefully now with fewer blocked threads)
- I also added a couple new unit tests
- I also repeated the steps-to-test from https://github.com/scalableminds/webknossos/pull/7453

### TODOs:
- [x] test range request
- [x] test directory listing
- [x] test error handling for directory listing (should be Box.Empty)
- [x] Region
- [x] test with custom endpoint
- [x] Encoding

### Issues:
- fixes  #7913

------


MichaelBuessemeyer (approved):
Thanks for this awesome pr and adding tests @fm3 :tada:

Codewise: I only found that you are "explicitly using" `net.liftweb.common.` instead importing the package. Could you please tell me why that is? 

Testing worked on the first check: 
But it seems that the dataset under `s3://janelia-cosem-datasets/jrc_mus-nacc-4/jrc_mus-nacc-4.zarr/` is not displayed correctly in webknossos. I also checked this on the master and its the same case there. Maybe the dataset was somehow changed? :thinking: Do you maybe know more about this?

Here is the link to the dataset on this branch's dev instance: https://newsclient.webknossos.xyz/datasets/sample_organization/test-two#2541,2561,381,0,1.3
And here on the master dev instance: https://master.webknossos.xyz/datasets/sample_organization/Test-Import-Int16-S3#2541,2561,381,0,1.3

But as this pr uses a different dataset in the tests, I'd say it is good to go?! (But please clarify why you used `net.liftweb.common.` within the code :)

fm3:
https://github.com/scalableminds/webknossos/issues/7960

fm3:
> But it seems that the dataset under s3://janelia-cosem-datasets/jrc_mus-nacc-4/jrc_mus-nacc-4.zarr/ is not displayed correctly in webknossos.

Yep, that’s unrelated. The data did not change afaik, but we never supported reading zstd-compressed zarr2 data. I wrote a separate issue: https://github.com/scalableminds/webknossos/issues/7960

> Is there a naming collision that I did not discover?

That’s right. The `transformWith` method to access exceptions within futures works with `scala.util.Try`, one flavor of which is `scala.util.Failure`, creating a naming collision with `net.liftweb.common.failure`. To make this more explicit, I now opted for renaming imports rather than using the full qualifiers in code. Do you agree that this is more discoverable?

MichaelBuessemeyer (approved):
Thanks for your latest commit. IMO this is a great improvement :tada:

aaronkanzer:
@fm3 @MichaelBuessemeyer thanks for these updates @normanrz -- we were just testing with a similar update f for our fork (https://github.com/lincbrain/webknossos/pull/10/commits/24d2a097d688c9fcc3625be3dade24ca14db7f28) and were about to push upstream 😂 -- quite the coincidence!

One follow-up question here -- but since there is now async behavior for the `S3DataVault` does this mean that threading configuration can change slightly -- [as defined here](https://github.com/scalableminds/webknossos/blob/ca5f2aa6a0c5174b2f79ec1a9b8ff17a5a37c582/conf/application.conf#L66-L76) -- we are continuing to tinker with these values to see if it makes the requests even faster -- thanks in advance for the consideration.

Cc @kabilar @ayendiki

fm3:
@aaronkanzer Oh, I hope you didn’t spend too much effort on something that we now also did? If you are missing features / upgrades like that, feel free to open issues in this repository, we will then discuss if they also make sense for us. There is always a chance that we want to go in similar directions.

You are correct, the optimal threading configuration may be affected by this change, especially if you know that you have mostly s3-hosted datasets. Then this change means fewer blocked threads, meaning a lower total number of threads may be more efficient. The values in the default configuration are very rough guesses, and the optmial values will depend on your workload. Did you already make some measurements on this and did you find something new? We’d be interested :)

Since in our case we usually have a very diverse set of access paths, with gcs and local file being at least as common as s3. These are not yet properly async, so I do not think we will change the default values at this time.

Hope this helps!

PR #7939, opened 2024-07-23
Persist skeleton group expansion state

dieknolle3333:
https://moveskeletonexpansiontostore.webknossos.xyz/
### Steps to test:
- Go to skeletons tab, make sure to have some trees, ordered in some groups
- Expand/collapse groups and subgroups, make sure everything works well
- reload site and check that the same groups are expanded/collapsed, except root group

### TODOs:
- [x] frontend
  - [x] Add `isExpanded` to nml
      - [x] read
      - [x] write
      - [x] fix tests
- [x] backend
  - [x] Improve NML Support
- [x] fix bug where empty groups cant be dragged

### Issues:
- fixes #

------


dieknolle3333:
@MichaelBuessemeyer would you be willing to make the same changes in the backend as in the segment group PR? the plan is the same here, to be able to store whether a segment group is collapsed or expanded.

MichaelBuessemeyer:
Sure :)

dieknolle3333:
@philippotto the backend part is missing so testing will hopefully just yield the same behaviour as before. But I'm requesting review already for the first iteration so that most of the review is already done :) (funny: GitHub automatically set the right base branch.)

frontend/javascripts/oxalis/view/right-border-tabs/tree_hierarchy_view_helpers.ts
```diff
@@ -90,6 +90,7 @@ export function insertTreesAndTransform(
       timestamp: 0,
       children: _.orderBy(treeNodeChildren, ["name"], ["asc"]),
       disableCheckbox: treeNodeChildren.length === 0,
+      expanded: group.isExpanded == null ? true : group.isExpanded,
```
philippotto, line 93:
```suggestion
      expanded: group.isExpanded == null || group.isExpanded,
```

usually, if `true` or `false` appear in a logical expression, they can be removed :)

frontend/javascripts/oxalis/view/right-border-tabs/tree_hierarchy_view.tsx
```diff
@@ -157,8 +151,15 @@ function TreeHierarchyView(props: Props) {
       ).map((node) => node.key);
       expandedKeys = _.without(expandedKeys, ...subGroupKeys);
     }
-
-    setExpandedNodeKeys(expandedKeys);
+    const expandedKeySet = new Set(expandedKeys);
```
philippotto, line 154: in line 152, there is a _.without which is a bit expensive since it iterates over the entire array. since you convert the array into a Set anyway, you can do the removal on the set and afterwards convert to an array if needed.

frontend/javascripts/oxalis/view/right-border-tabs/tree_hierarchy_view.tsx
```diff
@@ -157,8 +151,15 @@ function TreeHierarchyView(props: Props) {
...
+    const newGroups = mapGroups(props.treeGroups, (group) => {
+      if (expandedKeySet.has(getNodeKey(GroupTypeEnum.GROUP, group.groupId))) {
+        return { ...group, isExpanded: true };
+      } else {
+        return { ...group, isExpanded: false };
+      }
+    });
```
philippotto, line 161: you can avoid creating new instances if the expanded property did not change. I think, I suggested the same for the segment groups PR. please do the same change there in case it slipped through.

```suggestion
    const newGroups = mapGroups(props.treeGroups, (group) => {
      const shouldBeExpanded = expandedKeySet.has(getNodeKey(GroupTypeEnum.GROUP, group.groupId))
      if (shouldBeExpanded !== group.isExpanded) {
        return { ...group, isExpanded: shouldBeExpanded };
      } else {
        return group;
      }
    });
```

frontend/javascripts/oxalis/view/right-border-tabs/tree_hierarchy_view.tsx
```diff
@@ -238,17 +239,24 @@ function TreeHierarchyView(props: Props) {
...
-    );
-
-    if (expanded) {
-      setExpandedNodeKeys(_.uniq([...expandedNodeKeys, ...subGroupKeys]));
-    } else {
-      setExpandedNodeKeys(_.without(expandedNodeKeys, selectedGroupIdKey, ...subGroupKeys));
+  function setExpansionOfAllSubgroupsTo(parentGroup: TreeNode, expanded: boolean) {
+    if (parentGroup.id === MISSING_GROUP_ID) {
+      const newGroups = mapGroups(props.treeGroups, (group) => {
+        return { ...group, isExpanded: expanded };
```
philippotto, line 245: same here (the code will be longer but it's always good to avoid object creations)

philippotto (commented):
first round of feedback :) will do testing in another iteration.

dieknolle3333:
faster than lightning :zap:

frontend/javascripts/oxalis/view/right-border-tabs/tree_hierarchy_view.tsx
dieknolle3333: ah, damn. yeah, it did. I may need to fix it somewhere else, right? because that code isnt in the base branch. as I understood, master wont be merged into this branch here and I am not sure whether this branch or its base will be merged first.

philippotto: yeah, maybe simply open a new pr :)

frontend/javascripts/oxalis/view/right-border-tabs/tree_hierarchy_view.tsx
```diff
@@ -238,17 +240,31 @@ function TreeHierarchyView(props: Props) {
...
+    // Do expand the group if the subgroups are expanded though.
+    if (expanded === false) subGroupsMap.delete(parentGroup.id);
+    const newGroups = mapGroups(props.treeGroups, (group) => {
+      if (subGroupsMap.has(group.groupId) && expanded !== group.isExpanded) {
+        return { ...group, isExpanded: expanded };
+      } else {
+        return group;
+      }
+    });
+    console.log(newGroups);
```
dieknolle3333, line 266: brrrr

frontend/javascripts/oxalis/view/right-border-tabs/tree_hierarchy_view.tsx
```diff
@@ -302,7 +317,14 @@ function TreeHierarchyView(props: Props) {
     }
 
     // in either case expand the parent group
-    setExpandedNodeKeys([...expandedNodeKeys, getNodeKey(GroupTypeEnum.GROUP, parentGroupId)]);
+    const newGroups = mapGroups(props.treeGroups, (group) => {
+      if (group.groupId === parentGroupId) {
```
dieknolle3333, line 321: and !group.expanded

frontend/javascripts/oxalis/view/right-border-tabs/tree_hierarchy_view.tsx
dieknolle3333: (for reference https://github.com/scalableminds/webknossos/pull/7947)

MichaelBuessemeyer:
The basics should work. However, I still need to check the nml support of the new "isExpanded" field

philippotto:
> However, I still need to check the nml support of the new "isExpanded" field

If the NML format is extended with an `isExpanded` property, the frontend parser and writer also need to be adapted. Wklibs, too. I wonder whether that's necessary (on the other hand, it's probably relatively easy to do)... cc @fm3

fm3:
It would not be complex to add that, I think. However, I’m also not sure if losing the expansion state during download + reupload would maybe be fair?

MichaelBuessemeyer:
> However, I’m also not sure if losing the expansion state during download + reupload would maybe be fair?

Is that a frequent use case?

> If the NML format is extended with an isExpanded property, the frontend parser and writer also need to be adapted. Wklibs, too. I wonder whether that's necessary (on the other hand, it's probably relatively easy to do)... cc @fm3

I wanted to make this field optional anyway. Thus, reading and writing skeleton nmls without the field should still work, correct?

philippotto:
On second thought, my opinion is: Let's support writing and reading it in WK (it's easy to achieve and for complex annotations, it can be annoying to lose the expansion states). When reading, a missing `isExpanded` property should default to true. When writing, `isExpanded` should not be emitted, if it's true.
wklibs will probably ignore the property for now in the older versions (which is fine). however, adding support there should also be straightforward. cc @markbader 

Feel free to veto!

MichaelBuessemeyer:
@dieknolle3333 will you adapt the frontend nml support to support the new `isExpanded` field?

frontend/javascripts/oxalis/model/helpers/nml_helpers.ts
```diff
@@ -502,8 +502,11 @@ function serializeTreeGroups(treeGroups: Array<TreeGroup>, trees: Array<Tree>):
       serializeTagWithChildren(
         "group",
         {
-          id: treeGroup.groupId,
-          name: treeGroup.name,
+          ...{
+            id: treeGroup.groupId,
+            name: treeGroup.name,
+          },
```
philippotto, line 508: the original code should also work, unless I'm mistaken?

```suggestion
          id: treeGroup.groupId,
          name: treeGroup.name,
```

frontend/javascripts/oxalis/model/helpers/nml_helpers.ts
```diff
@@ -502,8 +502,11 @@ function serializeTreeGroups(treeGroups: Array<TreeGroup>, trees: Array<Tree>):
...
+          ...(treeGroup.isExpanded === false && { isExpanded: treeGroup.isExpanded }),
```
philippotto, line 509: didn't know that `false` was spreadable anyway. I think, it's cleaner with a ternary:

```suggestion
          ...(treeGroup.isExpanded ? {} : { isExpanded: treeGroup.isExpanded }),
```

frontend/javascripts/oxalis/view/right-border-tabs/tree_hierarchy_view.tsx
```diff
@@ -238,17 +240,30 @@ function TreeHierarchyView(props: Props) {
...
+          return { ...group, isExpanded: expanded };
+        }
+        return group;
+      });
+      setUpdateTreeGroups(newGroups);
+      return;
     }
+    const subGroups = getGroupByIdWithSubgroups(props.treeGroups, parentGroup.id);
+    const subGroupsMap = new Set(subGroups);
+    // If the subgroups will should be collapsed, not collapse the group itself.
```
philippotto, line 256:
```suggestion
    // If the subgroups should be collapsed, do not collapse the group itself.
```

frontend/javascripts/test/snapshots/public-test/test-bundle/test/libs/nml.spec.js.md
```diff
@@ -57,7 +57,7 @@ Generated by [AVA](https://avajs.dev).
       </comments>␊
       <groups>␊
         <group id="1" name="Axon 1">␊
-          <group id="3" name="Blah" />␊
+          <group id="3" name="Blah" isExpanded="false" />␊
```
philippotto, line 60: I thought, isExpanded false won't be serialized?

philippotto:
code looks mostly good to me :+1: however: during testing I couldn't move an empty group into a child group of the root. I tried dragging group 4 into group 1, but nothing happened. on https://treetree.webknossos.xyz/ this works, so it seems to be related to this PR? can you reproduce this?

image

frontend/javascripts/oxalis/model/helpers/nml_helpers.ts
dieknolle3333: no youre right, much better!

dieknolle3333:
>can you reproduce this?

oh, yes. I will look into it!

frontend/javascripts/oxalis/view/right-border-tabs/tree_hierarchy_view.tsx
```diff
@@ -297,12 +313,19 @@ function TreeHierarchyView(props: Props) {
...
     } else {
       // A group was dragged - update the groupTree
-      const newTreeGroups = moveGroupsHelper(props.treeGroups, draggedNode.id, parentGroupId);
-      setUpdateTreeGroups(newTreeGroups);
+      updatedTreeGroups = moveGroupsHelper(props.treeGroups, draggedNode.id, parentGroupId);
     }
 
     // in either case expand the parent group
-    setExpandedNodeKeys([...expandedNodeKeys, getNodeKey(GroupTypeEnum.GROUP, parentGroupId)]);
+    // TODO_c look here
```
dieknolle3333, line 320: remove me

MichaelBuessemeyer:
Please wait before merging. I just noticed that the backed review ist still pending :)

dieknolle3333:
@philippotto super weird: I was able to reproduce the bug on https://moveskeletonexpansiontostore.webknossos.xyz/ but cannot reproduce it locally. Even before the small change I made. It doesnt seem to be browser related, but locally I was able to drag and drop empty groups with the current changes (also at commit 5e4c06db79b40679821d9e2eed9b9e470c037e8b, where I addressed the second round of review).

philippotto:
hm, weird indeed! however, your "fix commit" makes sense to me and now dnd seems to work reliably. so, from my side, this should be good to go :)

dieknolle3333:
oh its works for you now? because it wasnt working for me on the dev instance. but I can confirm that now its working?! whatever, maybe a weird browser caching thing :man_shrugging: what matters is that its working now :) :+1:

philippotto:
> because it wasnt working for me on the dev instance like 10 minutes ago.

maybe you had an old tab open? after running `/install_dev` it can take one or two minutes until the new files are served. if you mistakenly assumed, that the new stuff was already served and kept that tab open, this might explain it. cache problems are plausible, too, of course.

let's test again tomorrow and then merge this branch into tree-tree. on monday, we can hopefully merge tree-tree.

dieknolle3333:
> if you mistakenly assumed, that the new stuff was already served and kept that tab open, this might explain it. 

yeah, you're right, I might have been to fast. It felt longer than 1-2 minutes that I was switching between dev and local instance, and I felt like I was refreshing quite often, but maybe this was the cause after all. Thanks for the explanation, I will look out for this in the future.

philippotto:
> yeah, you're right, I might have been to fast. It felt longer than 1-2 minutes that I was switching between dev and local instance, and I felt like I was refreshing quite often, but maybe this was the cause after all. Thanks for the explanation, I will look out for this in the future.

by the way, you can always check the version number that is running here:

image

MichaelBuessemeyer:
> Please wait before merging. I just noticed that the backed review is still pending :)

I think the backend review was still missing :see_no_evil:

fm3 (commented):
Backend LGTM

philippotto:
Oops, I missed your comment between the other discussion.

PR #7942, opened 2024-07-24
Fix sandbox annotation api route

fm3:
The sandbox annotations tried to look up their owner during json serialization, but the dummyObjectId owner can’t be found. This must be handled while determining the annotation restrictions.

Regression from #7923 

I tested that http://localhost:9000/api/datasets/sample_organization/test-agglomerate-file/sandbox/skeleton returns a usable json again


philippotto (approved):
lgtm :+1:

PR #7946, opened 2024-07-24
Fix error reporting for unhandled rejections

philippotto:
At some point I "fixed" that the reporting of unhandled rejections failed because the message property couldn't be written. I did so by shallowy copying the object with an object spread. However, this copy wasn't an "Error" instance anymore (but a plain object). I couldn't reproduce the original problem with setting the `message` which is why I re-added it. I put it into a try-catch to be extra safe. Having the unchanged message should not be a deal breaker, anyway.

- https://fixerrorreportingforunhandledrejections.webknossos.xyz/dashboard/datasets/Datasets-66a1249301000001006bc00c

### Steps to test:
- I executed `new Promise((res, rej) => rej(new Error("blaaa")))` in the devtools on the dev instance and on master
- Here is a screenshot of airbrake how it looks with the fixed reporting:
image

### DIFF
--- a/.circleci/not-on-master.sh
+++ b/.circleci/not-on-master.sh
@@ -1,8 +1,8 @@
 #!/usr/bin/env bash
 set -Eeuo pipefail
 
-if [ "${CIRCLE_BRANCH}" == "master" ]; then
+# if [ "${CIRCLE_BRANCH}" == "master" ]; then
   echo "Skipping this step on master..."
-else
-  exec "$@"
-fi
+# else
+#   exec "$@"
+# fi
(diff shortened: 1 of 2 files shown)


.circleci/not-on-master.sh
```diff
@@ -1,8 +1,8 @@
 #!/usr/bin/env bash
 set -Eeuo pipefail
 
-if [ "${CIRCLE_BRANCH}" == "master" ]; then
+# if [ "${CIRCLE_BRANCH}" == "master" ]; then
```
philippotto, line 4: revert

MichaelBuessemeyer (approved):
Looks good to me. I did not test myself as your screenshots seemed enough for me.

Please don't forget to revert the changes made to `not-on-master.sh`

PR #7947, opened 2024-07-24
Avoid object creation when segment groups are collapsed/expanded

dieknolle3333:
### Steps to test:
- Go to segments tab and collapse and expand segment groups 
- Expand/Collapse all sub groups

### Issues:
- follow-up for #7928


dieknolle3333:
@philippotto please let me know if you think that this needs a changelog entry. :sweat_smile:

philippotto (approved):
great, thank you! no changelog entry needed :)

PR #7964, opened 2024-07-31
support reading zstd-compressed zarr2 datasets

### LINKED ISSUE #7894: Support ZStandard Compression for Zarr (version 2)
## Detailed Description

The dataset under `s3://janelia-cosem-datasets/jrc_mus-nacc-4/jrc_mus-nacc-4.zarr/` uses ZStandard and is in Zarr version 2. Currently, wk does not support ZStandard compression for this Zarr version, but it would be very nice to have :)

fm3:
I didn’t find details about the zarr2 specification for the zstd compressor, so I’m just taking what I saw in the example dataset (just "level"). That means that for datasets with non-integer level or a different feature like checksums it may still not work. Let’s work on that when we notice problems.

- https://zarrzstd.webknossos.xyz

### Steps to test:
- Add a zstd-compressed remote dataset, e.g. `s3://janelia-cosem-datasets/jrc_mus-nacc-4/jrc_mus-nacc-4.zarr/`
- Should load correctly (note: the contrast of that dataset is very low compared to the int16 data range)

### Issues:
- fixes #7894 

------


MichaelBuessemeyer (approved):
Aesome, works :tada:

PR #7966, opened 2024-07-31
[WIP] Faster CI

philippotto:
- https://___.webknossos.xyz

### Steps to test:
- abc

### TODOs:
- [ ] ...

### Issues:
- fixes #

------

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -166,8 +166,8 @@ jobs:
       - run:
           name: Run webknossos smoke test
           command: |
-            for i in {1..10}; do # retry
-              sleep 10
+            for i in {1..20}; do # retry
+              sleep 5
               curl --fail -v http://localhost:9000/api/health && s=0 && break || s=$?
             done
             (exit $s)
@@ -182,8 +182,8 @@ jobs:
       - run:
           name: Run webknossos-datastore smoke test
           command: |
-            for i in {1..10}; do # retry
-              sleep 10
+            for i in {1..20}; do # retry
+              sleep 5
               curl --fail -v http://localhost:9090/data/health && s=0 && break || s=$?
             done
             (exit $s)
@@ -198,8 +198,8 @@ jobs:
       - run:
           name: Run webknossos-tracingstore smoke test
           command: |
-            for i in {1..10}; do # retry
-              sleep 10
+            for i in {1..20}; do # retry
+              sleep 5
               curl --fail -v http://localhost:9050/tracings/health && s=0 && break || s=$?
             done
             (exit $s)
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,7 +2,7 @@ version: 2
 jobs:
   build_test_deploy:
     machine:
-      image: ubuntu-2004:202111-02
+      image: cimg/base:2021.04
       docker_layer_caching: true
     resource_class: large
     environment:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -11,8 +11,7 @@ jobs:
       USER_GID: 1000
       TZ: Europe/Berlin
     steps:
-      steps:
-        - run: git clone --depth 5 "$CIRCLE_REPOSITORY_URL" --branch "$CIRCLE_BRANCH"
+      - run: git clone --depth 5 "$CIRCLE_REPOSITORY_URL" --branch "$CIRCLE_BRANCH"
       - run:
           name: "Custom environment variables"
           command: |
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,7 +2,7 @@ version: 2
 jobs:
   build_test_deploy:
     machine:
-      image: cimg/base:2021.04
+      image: ubuntu-2004:2024.05.1
       docker_layer_caching: true
     resource_class: large
     environment:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -160,9 +160,10 @@ jobs:
           command: |
             .circleci/not-on-master.sh docker-compose run base yarn typecheck
       - run:
-          name: Start webknossos
+          name: Start webknossos, datastore and tracingstore
           background: true
-          command: docker-compose up webknossos
+          command: docker-compose up webknossos & docker-compose up webknossos-datastore & docker-compose up webknossos-tracingstore
+
       - run:
           name: Run webknossos smoke test
           command: |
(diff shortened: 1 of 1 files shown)


daniel-wer:
@philippotto Please find some of my observations and ideas from my initial look at the CI pipeline for webknossos:

- [ ] As you have already observed, the "Spin up environment" step takes the longest out of all steps (~3-4m). Looking into it, the time is spent downloading (`Downloading: downloaded 13.01GiB, expanded to 14.54GiB, ratio 0.90, after 4m11.724810166s.`). This is likely due to the docker layer caching (DLC) that is turned on. The caching should be worthwhile but one could still test to delete the docker layer cache and compare times without it (note that the DLC push at the end also takes 1-2m!). Also one could check whether the cache size inflated over time and whether a reset might shrink it back down again.
- [ ] It's a little hard to find information on that, but I think open source projects can make use of concurrently running jobs (the free plan includes 30 concurrent jobs). A [fan out](https://circleci.com/docs/concurrency/#fan-out-fan-in-workflow-example) where the build steps are executed first and the frontend tests, backend linting, and end-to-end tests are then run in parallel should speed things up.
- [ ] For completeness sake (you have already prototyped this on another branch): The yarn cache restore and install takes rather long (50s + 30s). Using pnpm and/or installing without cache will/might improve the times.
- [ ] We should look into the other longer running steps
  - [ ] Can we speed up the frontend tests, for example by executing less tests serially? I have not found the option to print test times, but there's likely a way to identify the longest running tests.
  - [ ] The backend linting spends quite a bit of time compiling stuff, although there is a build step before. This might be expected.
  - [ ] The end-to-end tests are all ran serially. Maybe some groups can be executed in parallel?
  
I'm happy to discuss these in a call and/or contribute as part of this PR :)

daniel-wer:
The DLC can be deleted at https://app.circleci.com/settings/project/github/scalableminds/webknossos/docker-layer-caching

philippotto:
> [ ] As you have already observed, the "Spin up environment" step takes the longest out of all steps (~3-4m). Looking into it, the time is spent downloading (`Downloading: downloaded 13.01GiB, expanded to 14.54GiB, ratio 0.90, after 4m11.724810166s.`). This is likely due to the docker layer caching (DLC) that is turned on. The caching should be worthwhile but one could still test to delete the docker layer cache and compare times without it (note that the DLC push at the end also takes 1-2m!). Also one could check whether the cache size inflated over time and whether a reset might shrink it back down again.

very good point :+1: I wasn't aware of that. I just disabled DLC and the spin up went down from 3.5 min to 0 s. Building the docker images went up from 27s to 1.5 min (so, total 4 min vs 1.5 min). the overall impact of DLC is negative, I'd say :sweat_smile: 

> note that the DLC push at the end also takes 1-2m!

that isn't blocking though (the CI in github is already marked as green before dlc teardown finishes). so, this is (unfortunately?) not relevant for us, I think.

> The DLC can be deleted at https://app.circleci.com/settings/project/github/scalableminds/webknossos/docker-layer-caching

I tried this out and ran the pipeline twice (the first run for filling up the cache again). the second run was very fast. Spin up took 17s (downloaded approx 2 GB) and building the docker images took 29s. So, in total only 46s. However, subsequent runs got way slower again. Apparently, the layer cache bloats up very quickly so that 7 GB are downloaded. See the overview here where I re-ran 31167e1 multiple times: https://app.circleci.com/pipelines/github/scalableminds/webknossos?branch=faster-ci
therefore, I think that we should disable DLC as I don't see a way to avoid it getting bloated that quickly (by the way, I just read this: "Once you have enabled DLC for your project, any layers that are saved will be available for three days between job runs. Any cached layers that are not reused after three days will be deleted."; so, the cache retention was very short to begin with but for us this doesn't help much).

philippotto:
> It's a little hard to find information on that, but I think open source projects can make use of concurrently running jobs (the free plan includes 30 concurrent jobs). A [fan out](https://circleci.com/docs/concurrency/#fan-out-fan-in-workflow-example) where the build steps are executed first and the frontend tests, backend linting, and end-to-end tests are then run in parallel should speed things up.

Yes, I think that would be great :+1: if you have time for that, feel free to explore. I don't think I will get around to it for now :S

philippotto:
@daniel-wer I made some other improvements to the CI which I think can already be merged. See the PR description for details :)

daniel-wer (approved):
Awesome, great time savings!

I'm a little confused why, although DLC is disabled and you deleted the cache, the "Build webknossos-dev docker image" step still uses cached results for all build steps. Could it be that during your testing (or from other branches) the cache was filled again and is now used anyways? :thinking: I just want to avoid that in a few weeks once the change propagated everywhere we suddenly notice that the docker build takes way more time all of a sudden.
However, I just noticed that the other steps that build docker images don't use any cache, so it's unlikely this is a big issue.

philippotto:
> I'm a little confused why, although DLC is disabled and you deleted the cache, the "Build webknossos-dev docker image" step still uses cached results for all build steps. Could it be that during your testing (or from other branches) the cache was filled again and is now used anyways? 🤔 I just want to avoid that in a few weeks once the change propagated everywhere we suddenly notice that the docker build takes way more time all of a sudden.

Interesting! I just looked into this and the dev image is explicitly instructed to use a cache (see [here](https://github.com/search?q=+owner%3Ascalableminds+repo%3Ascalableminds%2Fwebknossos%20DEV_CACHE&type=code)). So, this is independent of DLC and should be reliable in the future.

I wonder whether we could use the same trick for the other three docker images that we are building in the CI. A short overview:

| Image    | Time | Cache? | Size |
| -------- | ------- | ------- | ------- |
| webknossos-dev   | 16s    | yes | ~890 MB |
| webknossos | 35s | no | ~760 MB |
| webknossos-datastore | 13s | no |  ~470 MB |
| webknossos-tracingstore | 12s | no |  ~470 MB |

@normanrz Do you have thoughts on that? Especially, for the main dockerfile, I think it would be useful to use the cache so that the installation of the dependencies can be skipped. Or was there a deliberate decision against using the same cache mechanism for these dockerfiles?

philippotto:
@daniel-wer I experimented a bit further but now I cleaned up and only left two additional improvements after your review (see the last two bullet points in the PR description) :) Please have one final look :pray:

I also noticed that the backend compilation is not really deterministic which is bad for the docker layer caching (not the one circleci does but the "native" docker caching). however, I didn't really find a good solution yet how this can be avoided. a fight for another time :grinning:

daniel-wer:
Thank you for the continued investigation and improvements :tada: Looking forward to have this merged!

PR #7967, opened 2024-07-31
Allow alignment job to take matches from skeleton into account

Tobias314:
### Description:
- Moves alignment job to separate tab of "AI Tools"
- Adds option to use manual matches from a skeleton for alignment

### Steps to test:
- Create some skeleton trees for some sections of a dataset and then start an alignment job via the "AI Tools"

### TODOs:

### Issues:
- fixes [voxelytics issue 3616](https://github.com/scalableminds/voxelytics/issues/3616)


frontend/javascripts/oxalis/view/action-bar/starting_job_modals.tsx
```diff
@@ -503,8 +491,78 @@ function RunAiModelTab({ aIJobModalState }: { aIJobModalState: string }) {
...
+function UseSkeletonFormItem() {
+  const tracing = useSelector((state: OxalisState) => state.tracing);
+  return (
+    <div>
+      <Form.Item
+        name="useAnnotation"
+        label={
+          <Space>
+            <div style={{}}>
+              Manual Matches
```
MichaelBuessemeyer, line 531: A forced space is needed here:
```suggestion
              Manual Matches{" "}
```
Otherwise there is no space between the text and the icon which doesn't look nice :)

frontend/javascripts/oxalis/view/action-bar/starting_job_modals.tsx
```diff
@@ -503,8 +491,78 @@ function RunAiModelTab({ aIJobModalState }: { aIJobModalState: string }) {
...
+                    "No annotation was found. Please create an annotation first.",
+                  );
+                }
+                if (tracing.skeleton == null) {
+                  return Promise.reject("No skeleton was found. Please create a skeleton first.");
+                }
+              }
+              return Promise.resolve();
+            },
+          },
```
MichaelBuessemeyer, line 554: Just a suggestion: Maybe also include a check for whether the skeleton tracing has at lease one tree? (Is not empty)

frontend/javascripts/oxalis/view/action-bar/starting_job_modals.tsx
```diff
@@ -503,8 +491,78 @@ function RunAiModelTab({ aIJobModalState }: { aIJobModalState: string }) {
...
+            </Space>
+          </Card>
+        </Radio.Button>
+      </Space>
+      <AlignSectionsForm />
+    </div>
+  );
+}
+
+function UseSkeletonFormItem() {
```
MichaelBuessemeyer, line 522: I'd prefer if we could avoid letting a component start with "Use" as this usally implies that the function is a hook and not a whole component. 

Maybe something like "ShouldUseTreesFormItem" (tree should be the more up to date name compoared to skeleton), "ShouldAlignAccordingToTreesFormItem" or something else in that direction :) The name not starting with "use" would already be helpful :pray:

MichaelBuessemeyer (changes requested):
Thanks for this PR @Tobias314,

I only found minor things. Well done :) The only must change form my side would be the renaming of the skeleton form item. Please see my comment for that.

Additionally I want to double check something with @fm3 :)

Edit: And maybe also the additional check for an empty skeleton tracing. I just managed to submit the job with the check mark set at the new form item, but with an empty skeleton tracing. I'd welcome a error in such a case :)

conf/webknossos.latest.routes
```diff
@@ -265,7 +265,7 @@ POST          /jobs/run/exportTiff/:organizationName/:datasetName
...
 POST          /jobs/run/inferNeurons/:organizationName/:datasetName                             controllers.JobController.runInferNeuronsJob(organizationName: String, datasetName: String, layerName: String, bbox: String, outputSegmentationLayerName: String, newDatasetName: String)
 POST          /jobs/run/inferMitochondria/:organizationName/:datasetName                        controllers.JobController.runInferMitochondriaJob(organizationName: String, datasetName: String, layerName: String, bbox: String, outputSegmentationLayerName: String, newDatasetName: String)
-POST          /jobs/run/alignSections/:organizationName/:datasetName                            controllers.JobController.runAlignSectionsJob(organizationName: String, datasetName: String, layerName: String, newDatasetName: String)
+POST          /jobs/run/alignSections/:organizationName/:datasetName                            controllers.JobController.runAlignSectionsJob(organizationName: String, datasetName: String)
```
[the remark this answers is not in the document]
fm3: It is not a problem to change these routes, even with breaking changes.
You are right that we need to generally be careful with breaking changes in the api, but there exist very different cases.
The case here is that these routes are ONLY called by our own javascript frontend. This means that caller and server are deployed together automatically. The only inconsistency that can arise is if users keep the tab open and try do to requests with their outdated loaded javascript code. This case is currently considered acceptable.

For other routes, however, there are multiple potential callers, some are called by the [python libs client](https://github.com/scalableminds/webknossos-libs/blob/master/webknossos/webknossos/client/api_client/wk_api_client.py), some are called directly by voxelytics (I think currently that’s `POST /triggers/reload/:organizationName/:datasetName` and the two `fullMesh.stl` routes). Some are also called by other people’s code (zarr data endpoints, and a small number of other routes). This is a bit messy currently, and I’d like to boil this down to (1) the wk js frontend and (2) the wk python client, but we’re not there yet.

However, `/jobs/run/` is safe. The only thing to consider there is that already-submitted jobs have their full argument set in the database, so if any jobs are still pending when the new code and new worker are deployed, they may fail. But this, too, is currently a rare case and should still be acceptable.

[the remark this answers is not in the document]
fm3: On your second question, url params vs request body is not really a big issue for me (there is no super standardized way of doing this in wk in general). But I see that Tobias now changed it to url params and yes, I guess that’s the most consistent way with the other jobs/run routes. Note that there is still a now-unused line `alignSectionsJobOptions = request.body` that should be removed.

app/controllers/JobController.scala
```diff
@@ -302,8 +303,10 @@ class JobController @Inject()(
           dataset <- datasetDAO.findOneByNameAndOrganization(datasetName, organization._id) ?~> Messages(
             "dataset.notFound",
             datasetName) ~> NOT_FOUND
+          alignSectionsJobOptions = request.body
```
MichaelBuessemeyer, line 306: remove me :)
```suggestion
```

conf/webknossos.latest.routes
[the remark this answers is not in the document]
MichaelBuessemeyer: @Tobias314 
> Note that there is still a now-unused line alignSectionsJobOptions = request.body that should be removed.

frontend/javascripts/oxalis/view/action-bar/starting_job_modals.tsx
```diff
@@ -503,8 +491,84 @@ function RunAiModelTab({ aIJobModalState }: { aIJobModalState: string }) {
...
+                    "No skeleton edges were found. Please create a skeleton with at least one edge first.",
+                  );
+                }
+              }
+              return Promise.resolve();
+            },
+          },
+        ]}
+      >
+        <Checkbox> Use manual matches from skeleton. </Checkbox>
```
MichaelBuessemeyer, line 563: I think we decided to call every tree/skeleton text in the UI consistently tree, correct @philippotto? 

Sorry, for the confusion @Tobias314. Back in the early WK days we the trees were called skeletons and not trees. Nowadays we try to kepp the naming consistent and use tree at least in all UI related things. Due to code-wise historical reasons we cannot full switch from naming skeletons to trees. That's why one might encounter the name skeleton still in the code. Afaik it is still the dominant word for naming variable.


```suggestion
        <Checkbox> Use manual matches from trees. </Checkbox>
```

frontend/javascripts/oxalis/view/action-bar/starting_job_modals.tsx
```diff
@@ -651,7 +718,8 @@ function StartJobForm(props: StartJobFormProps) {
         onChangeSelectedBoundingBox={(bBoxId) => form.setFieldsValue({ boundingBoxId: bBoxId })}
         value={form.getFieldValue("boundingBoxId")}
       />
-      <div style={{ textAlign: "center" }}>
+      {isSkeletonSelectable && <ShouldUseTreesFormItem />}
+      <div style={{ textAlign: "left" }}>
```
MichaelBuessemeyer, line 722: Why did you change the text alignment?

MichaelBuessemeyer (approved):
Sorry, @Tobias314 I found another few nitpicky suggestions about variable naming. Afterwards this PR should be good to go :)
Please apply them before merging

MichaelBuessemeyer:
And could you please increase the width of the buttons presented in the "run a model" view? Due to one item less, there is some free and sad looking white space.

This is how it looks on the master where there is no whitespace:
image

And this is your branch with white space on the right side where another fourth button should be.
image

frontend/javascripts/oxalis/view/action-bar/starting_job_modals.tsx
Tobias314: The button is also on the left for the "Run a model" tab so I guess left is consistent. I will keep it on the left now but I do not really care.

PR #7974, opened 2024-08-05
Virtualize Bounding Box List

### LINKED ISSUE #7938: Virtualize the bounding box tab content
## Detailed Description

The Bounding Box tab tends to be slow with many bounding boxes. Therefore, lets virtualize it. 

For reference see comment https://github.com/scalableminds/webknossos/pull/7935#issuecomment-2245194225.
Maybe using a virtualized list & managing the highlighting and scrollIntoView ourselves might be a valid solution to this.

dieknolle3333:
- https://virtualizebb.webknossos.xyz/
- e.g. some bounding boxes are in this annotation: https://virtualizebb.webknossos.xyz/links/hDU6gJX0zhrR8jRS

### Steps to test:
- Open an annotation and go to bounding box tab.
- Make sure you have at least 25 bounding boxes. The list should scroll quite some content.
- UI wise, nothing should be different than before, test that. It should look and behave as before (layout, scrolling, etc)
- Check that only the visible bounding boxes (plus some for padding) are loaded, e.g. by inspecting the html. 

### Issues:
- fixes #7938 

------


philippotto:
The code looks good and the UI too :+1: however, there is one thing: due to the "add bbox button", there is a native scrollbar (in addition to the virtualized but invisible one):

image

I think, you need to adapt the layout so that the div, that contains the table and the plus-button, is exactly as tall as there is available space. I see two options for that:
a) use a flexbox for that div (not sure how well this behaves with the autosizer). ah, I just see that the button is in a footer row. so, this option probably doesn't work.
b) when the auto-sizer provides a height of Y pixels, subtract the 38 pixels (the height of the button) from that height when passing it along to the table.

since the plus-button is part of the table, I'm not sure why this is happening in the first place. maybe the culprit is somewhere else. can you reproduce this?

dieknolle3333:
>since the plus-button is part of the table, I'm not sure why this is happening in the first place. maybe the culprit is somewhere else. can you reproduce this?

Oh yes, I can reproduce this. I didnt notice that this was an additional scroll bar. But yeah, its weird that this happens! 
Anyway, its fixed with your second option! So I'll just leave it at that I think. I didnt see any comment in the docs or any option to include the footer height

philippotto (approved):
works well :tada:

PR #7975, opened 2024-08-05
Don't display "Fill segment" option if editable mapping is active

### LINKED ISSUE #7925: Fill Segment can be triggered while an editable mapping is active
## Context

I was proofreading an annotation, so an editable mapping was active.

## Expected Behavior

The "Fill Segment (flood-fill region)" option should not be available in the context menu and if triggered somehow anyways it should not do anything.

## Current Behavior

The "Fill Segment (flood-fill region)" option is available and does something, leading to an error during the next save. The only option then is for the user to reload the page.

## Steps to Reproduce the bug

1. Create an annotation for a dataset with an HDF5 mapping.
2. Perform a proofreading action, so an editable mapping is created.
3. Open the context menu and select "Fill Segment (flood-fill region)"
4. Press save or wait until the automatic save is triggered.

## Your Environment for bug

- Browser name and version: Chrome 126
- Operating System and version: Linux
- Version of WEBKNOSSOS (Release or Commit): 28690

dieknolle3333:
- https://disablefillsegmentwithactiveeditablemapping.webknossos.xyz/
- e.g. DS ROI2017_wkw_sparse_segmentation (I think it only has JSON mappings, not sure where a HDF5 mapping can be found to reproduce the bug)

### How to test
- create editable mapping and make sure that flood filling is not available in the context menu any more
- advanced: change code so that it is available in context menu again, and click is while editable mapping is active. there should be a toast and a console.error.
- in an annotation with a segmentation layer, turn it invisible and turn on a skeleton layer. now only skeleton options should be available in the context menu in the viewport, matching the disabled volume actions in the action bar.

### TODOs:
- [x] disable flood filling if there is no active segmentation layer
- [x] see if disabledVolumeInfo can be used even more
- [x] DRY nonSkeletonActions

### Issues:
- fixes #7925 

------


philippotto (commented):
great :) only see my one comment.

philippotto (approved):

PR #7979, opened 2024-08-06
Register all segments in a bounding box

### LINKED ISSUE #7906: Create Segments for all ids in a bounding box
## Detailed Description
The bounding box tab could get a feature that registers segments for all ids that are within that bounding box.

## Implementation
See code snippet in the context link. We need to be careful with large bounding boxes, because this could become very slow.

## Context
https://scm.slack.com/archives/C5AKLAV0B/p1719482479270149

dieknolle3333:
- https://registerallsegmentsinbb.webknossos.xyz/

### Steps to test:
- Go to bounding box tab 
- inspect new overflow menu
- make sure everything works as before
- try out new action: "register all segments for bounding box"
- test segment and data limits

## TODOs
- [x] add context menu
- [x] add new entry with functionality
- [x] include limits on segments and data
- [x] add constant
- [x] add padding to BB entries
- [x] fix spacing
- [x] add registered segments to group
- [x] test that all the old disabling of context menu entries is still working
- [x] fix segment 0 bug 
- [x] disable if allowupdate is false

### Issues:
- fixes #7906 

------


CHANGELOG.unreleased.md
```diff
@@ -26,6 +26,7 @@ For upgrade instructions, please check the [migration guide](MIGRATIONS.released
 - Upgraded s3 client for improved performance when loading remote datasets. [#7936](https://github.com/scalableminds/webknossos/pull/7936)
 - To improve performance, only the visible bounding boxes are rendered in the bounding box tab (so-called virtualization). [#7974](https://github.com/scalableminds/webknossos/pull/7974)
 - Added support for reading zstd-compressed zarr2 datasets [#7964](https://github.com/scalableminds/webknossos/pull/7964)
+- Added a feature to register all segments for a given bounding box at once with an action in the bounding boxe's context menu. [#7979](https://github.com/scalableminds/webknossos/pull/7979)
```
philippotto, line 29: I think, it's `box's` but "of ..." feels more natural.

```suggestion
- Added a feature to register all segments for a given bounding box at once via the context menu of the bounding box. [#7979](https://github.com/scalableminds/webknossos/pull/7979)
```

frontend/javascripts/oxalis/constants.ts
```diff
@@ -337,6 +337,8 @@ const Constants = {
   SCALEBAR_HEIGHT: 22,
   SCALEBAR_OFFSET: 10,
   OBJECT_ID_STRING_LENGTH: 24,
+  REGISTER_SEGMENTS_BB_MAX_VOLUME_VX: 512 * 512 * 512,
+  REGISTER_SEGMENTS_BB_MAX_NO_SEGMENTS: 2000,
```
philippotto, line 341:
```suggestion
  REGISTER_SEGMENTS_BB_MAX_SEGMENT_COUNT: 2000,
```

"NO" feels like the opposite of yes :smile:

frontend/javascripts/oxalis/view/right-border-tabs/tree_hierarchy_view_helpers.ts
```diff
@@ -287,3 +292,68 @@ export function findParentGroupNode(nodes: TreeNode[], parentGroupId: number): T
   });
   return foundParentNode;
 }
+
+export const registerSegmentsForBoundingBox = async (value: Vector6, bbName: string) => {
```
philippotto, line 296: I think, this code should be moved into the `TracingApi` in `api_latest.ts`.

philippotto: the Vector6 interface is not very user-friendly in my opinion (because it's not clear whether the second vec3 is the extent or the max of the bounding box). why not two separate variables? either min and max or min and extent?

frontend/javascripts/oxalis/view/right-border-tabs/bounding_box_tab.tsx
```diff
@@ -129,6 +130,7 @@ export default function BoundingBoxTab() {
           onBoundingChange={_.partial(handleBoundingBoxBoundingChange, bb.id)}
           onDelete={_.partial(deleteBoundingBox, bb.id)}
           onExport={isExportEnabled ? _.partial(setSelectedBoundingBoxForExport, bb) : () => {}}
+          onRegisterSegmentsForBB={registerSegmentsForBoundingBox}
```
philippotto, line 133: couldn't the UserBoundingBoxInput component simply import the function itself?

frontend/javascripts/oxalis/view/components/setting_input_views.tsx
```diff
@@ -477,33 +493,87 @@ export class UserBoundingBoxInput extends React.PureComponent<UserBoundingBoxInp
...
       isOwner,
     );
+
+    const getContextMenu = () => {
+      const items: MenuProps["items"] = [
+        {
+          key: "registerSegments",
+          label: (
+            <>
+              <ScanOutlined style={marginRightStyle} /> {""}
```
philippotto, line 528: I think you can put the icon into an `icon` property (next to key and label). same for the others

frontend/javascripts/oxalis/view/components/setting_input_views.tsx
```diff
@@ -477,33 +493,87 @@ export class UserBoundingBoxInput extends React.PureComponent<UserBoundingBoxInp
...
+              Register all segments in this bounding box
+              <Tooltip title="Moves/registers all segments within this bbox into a new group">
```
philippotto, line 530:
```suggestion
              <Tooltip title="Moves/registers all segments within this bounding box into a new segment group">
```

I'd avoid the term bbox when it's possible. the tab title is an exception because the space is scarce there.

philippotto (changes requested):
already quite cool :)

philippotto (changes requested):
getting there :)

frontend/javascripts/oxalis/view/components/setting_input_views.tsx
dieknolle3333: is it okay to name the newly created segment group with the registered segments `Segments for BBox <name>`, e.g. `Segments for BBox Bounding box 2`? it wouldnt feel good to have `Segments for Bounding Box Bounding box 2`, and I think this would occur often as the default name for bboxes is `Bounding box <no>`

philippotto: good question! how about `Segments for <name>`? I think, this should suffice in most cases. usually, the users use something like "bounding box" or "region" etc. and if not, they probably picked a very good name, anyway.

frontend/javascripts/oxalis/api/api_latest.ts
```diff
@@ -636,23 +643,22 @@ class TracingApi {
       for (let y = min[1]; y < max[1]; y++) {
         for (let x = min[0]; x < max[0]; x++) {
           const id = data[idx];
-          if (!segmentIdToPosition.has(id)) {
+          if (!segmentIdToPosition.has(id) && id !== 0) {
```
philippotto, line 646:
```suggestion
          if (id !== 0 && !segmentIdToPosition.has(id)) {
```

it's better to do the faster quick first for a better early out :)

frontend/javascripts/oxalis/view/components/setting_input_views.tsx
dieknolle3333: okay! :)

philippotto (approved):
:+1:

Issue #7931, opened 2024-07-18
Make resumable dataset uploads robust against browser restarts


fm3:
https://www.npmjs.com/package/resumablejs allows a testChunks option to restart an upload even with a fresh browser. Needs the server to implement a test route, answering if a chunk is already present on the server or not.

PR #7981, opened 2024-08-07
Allow to continue upload after browser restart for same user

MichaelBuessemeyer:
- https://morerobustresumableupload.webknossos.xyz/

### Steps to test:
- Open the upload view on the dev instance: https://morerobustresumableupload.webknossos.xyz/datasets/upload?to=66b60af101000001003deac5
- Open the dev tools. In the debugger open file `dataset_upload_view.tsx` and set a breakpoint in line 425 (before `resumableUpload.upload();` is called upon the files added action.
- Keep the dev tools open, upload some small dataset (to have less traffic) 
  - Enter dataset name, allowed teams, target folder and drag & drop a dataset zip
- Hit the upload button. The upload should be reserved and then the debugger should kick in an interrupt starting the upload process. 
- Close your browser and do not skip the breakpoint.
- Reopen the browser and reopen the upload view.
- Open the dev tools to track the network traffic
- The view should now show the dataset name you just tried to upload but manually interrupted before starting the upload (stopping during the upload should also work. Click the continue upload
- The UI should be autofilled with the previous values and disabled.
- Drag & drop the same file as before and hit upload
- The upload should happily continue and succeed :)
- Now test the network traffic. There should be **get** requests to the `/datasets` route testing whether a chunk is already present. They should be returning 204 if no chunk was previously uploaded. If you tweaked the testing such that some chunks were already uploaded, some of these requests should answer 200 (signaling the chunk is present) (see below)
- Test the uploaded dataset. It should be viewable as usual. 

##
- Alternatively, for the debugger interruption, choose a larger dataset to upload and interrupt the upload in time.

### TODOs:
- [x] Test whether this works with multi-file upload (e.g. multiple tiffs at once)

### Issues:
- fixes [#7931](https://github.com/scalableminds/webknossos/issues/7931)

------


MichaelBuessemeyer:
@philippotto could you please review the frontend and @normanrz could you please review the backend code?

package.json
```diff
@@ -76,7 +76,7 @@
     "start": "node tools/proxy/proxy.js",
     "build": "node --max-old-space-size=4096 node_modules/.bin/webpack --env production",
     "@comment build-backend": "Only check for errors in the backend code like done by the CI. This command is not needed to run WEBKNOSSOS",
-    "build-backend": "yarn build-wk-backend && yarn build-wk-datastore && yarn build-wk-tracingstore",
+    "build-backend": "yarn build-wk-backend && yarn build-wk-datastore && yarn build-wk-tracingstore && rm webknossos-tracingstore/conf/messages webknossos-datastore/conf/messages",
```
philippotto, line 79: why are these removed?

frontend/javascripts/admin/admin_rest_api.ts
```diff
@@ -1307,6 +1307,35 @@ export function reserveDatasetUpload(
   );
 }
 
+export type OngoingUpload = {
+  uploadId: string;
+  dataSourceId: { name: string; organizationName: string };
```
philippotto, line 1312: usually we call this datasetId in the frontend (see `APIDatasetId`). can we use that term here too?

frontend/javascripts/admin/dataset/dataset_upload_view.tsx
```diff
@@ -212,6 +228,19 @@ class DatasetUploadView extends React.Component<PropsWithFormAndRouter, State> {
...
 
+  updateOngoingUploads = async () => {
+    const currentFormRef = this.formRef.current;
+    const datastoreUrl = currentFormRef?.getFieldValue("datastoreUrl");
+    const activeOrga = this.props.activeUser?.organization;
+    if (!datastoreUrl || !activeOrga) {
+      return;
+    }
+    const ongoingUploads = await getOngoingUploads(datastoreUrl, activeOrga);
+    // Sort by created
```
philippotto, line 239:
```suggestion
    // Sort so that most-recent uploads come first
```

frontend/javascripts/admin/dataset/dataset_upload_view.tsx
```diff
@@ -91,9 +99,12 @@ type State = {
   isRetrying: boolean;
   uploadProgress: number;
   selectedTeams: APITeam | Array<APITeam>;
-  uploadId: string;
+  possibleTeams: Array<APITeam>;
+  uploadId: string | null;
```
philippotto, line 103: it is still initialized to an empty string. can be changed to null now probably.

frontend/javascripts/admin/dataset/dataset_upload_view.tsx
```diff
@@ -681,9 +724,72 @@ class DatasetUploadView extends React.Component<PropsWithFormAndRouter, State> {
             />
           ) : null}
 
+          {this.state.ongoingUploads.length > 0 && (
+            <Alert
+              message={
+                <>
+                  Ongoing unfinished Dataset Uploads{" "}
```
philippotto, line 731: I think I would omit the word "ongoing" because it could be interpreted as if the upload was still progressing. However, this is likely not the case.

frontend/javascripts/admin/dataset/dataset_upload_view.tsx
```diff
@@ -681,9 +724,72 @@ class DatasetUploadView extends React.Component<PropsWithFormAndRouter, State> {
...
+                              initialTeams: this.state.possibleTeams.filter((team) =>
+                                ongoingUpload.allowedTeams.includes(team.id),
+                              ),
+                            });
+                            this.setState({
+                              uploadId: ongoingUpload.uploadId,
+                              continuingOldUpload: true,
+                            });
+                            Toast.info(
+                              "To continue the selected Upload please make sure to select the same file(s) as before. Else unexpected behavior may occur.",
```
philippotto, line 769:
```suggestion
                              "To continue the selected upload please make sure to select the same file(s) as before. Otherwise, the upload may be corrupted.",
```

philippotto (changes requested):
cool stuff!! didn't test yet but already left some comments :)

frontend/javascripts/admin/dataset/dataset_upload_view.tsx
MichaelBuessemeyer: :thinking:, down in the code I do explicitly set the `uploadId` to null to make it explicit that a new one should be generated after a successful or failed upload or in case no previous existing upload was selected from the list. Using `""` as a placeholder seems a little to implicit to me. I am aware that js evaluates `""` as false this is too implicit for me.

While writing this I came up with the following: I made the default value of uploadId undefined as a new uploadId should be generated when none is defined. This includes the initial value and the value set after an upload.  

Does that make sense? In short: I prefer to make the state of "don't use an already ongoing upload" more explicit via an undefined than an empty string :)

But feel free to argue

package.json
MichaelBuessemeyer: These message files are generated by the script as the tracingstore and datastore are built. When they are not deleted and one tries to run the script again the compilation of the two components somehow fails. After running into this multiple times (me having to manually delete these files) as well as taking care of not pushing these files, I simply wanted the script to also take care of cleaning up the auto-generated message files :)

MichaelBuessemeyer (commented):
Thanks for your feedback philipp :) 

I applied / commented everything :)

frontend/javascripts/admin/dataset/dataset_upload_view.tsx
philippotto: I meant [line 189](https://github.com/scalableminds/webknossos/pull/7981/files/7730af5293635cccbf5b2f61411bbd64e2d3bbcf#diff-1ad4a75b67f7779e859eb39c32a1e3bd5d4c220c5146b4b1089fbe36014d8d42R189) where `uploadId` is set to `""`. that should probably be changed to `null`.

> Using "" as a placeholder seems a little to implicit to me.

yes, i fully agree.

MichaelBuessemeyer: Oh sorry big misread from my side. Concentration kinda slipped I guess. I read it like: "Please change all `uploadId: null` assignments to `uploadId: ""`  :see_no_evil:

philippotto: it's good that you didn't agree to that :sweat_smile:

frontend/javascripts/admin/admin_rest_api.ts
```diff
@@ -1327,11 +1327,11 @@ export function getOngoingUploads(
       {
         host: datastoreHost,
       },
-    )) as Array<OngoingUpload & { dataSourceId: OldDataSourceIdFormat }>;
+    )) as Array<OngoingUpload & { datasetId: OldDatasetIdFormat }>;
     // Rename "team" to "organization" as this is the actual used current naming.
-    return ongoingUploads.map(({ dataSourceId: { name, team }, ...rest }) => ({
+    return ongoingUploads.map(({ datasetId: { name, team }, ...rest }) => ({
```
philippotto, line 1332: just tested this, and I think this needs to be changed back to dataSourceId here because the back-end uses that. otherwise, I get:

```
TypeError: Cannot read properties of undefined (reading 'name')
```

MichaelBuessemeyer: Oh right, thanks for spotting. Should be better now hopefully. will retest on monday

MichaelBuessemeyer: tested and worked :)

app/models/dataset/DatasetService.scala
```diff
@@ -69,6 +71,18 @@ class DatasetService @Inject()(organizationDAO: OrganizationDAO,
...
+  def getAllNotYetUploadedDatasetOfUser(userId: ObjectId, organizationId: ObjectId)(
+      implicit ctx: DBAccessContext): Fox[List[DatasetCompactInfo]] =
+    datasetDAO.findAllCompactWithSearch(
+      uploaderIdOpt = Some(userId),
+      organizationIdOpt = Some(organizationId),
+      isActiveOpt = Some(false),
+      includeSubfolders = true,
+      statusOpt = Some(notYetUploadedStatus),
+      // Only list pending upload since the two last weeks.
+      createdSinceOpt = Some(Instant.now - Duration(14, TimeUnit.DAYS))
```
fm3, line 83: If you import scala.concurrent.duration._ you can write
```suggestion
      createdSinceOpt = Some(Instant.now - 14 days))
```

app/models/dataset/DatasetService.scala
```diff
@@ -82,14 +96,14 @@ class DatasetService @Inject()(organizationDAO: OrganizationDAO,
     val dataSourceHash = if (dataSource.isUsable) Some(dataSource.hashCode()) else None
     for {
       organization <- organizationDAO.findOneByName(owningOrganization)
-      orbanizationRootFolder <- folderDAO.findOne(organization._rootFolder)
+      organizationRootFolder <- folderDAO.findOne(organization._rootFolder)
```
fm3, line 99: oops :sweat_smile:

webknossos-datastore/app/com/scalableminds/webknossos/datastore/models/OngoingUpload.scala
```diff
@@ -0,0 +1,16 @@
+package com.scalableminds.webknossos.datastore.models
+
+import com.scalableminds.util.time.Instant
+import com.scalableminds.webknossos.datastore.models.datasource.DataSourceId
+import play.api.libs.json.{Format, Json}
+
+// Defines the real-world size in a length unit for a mag1-voxel.
```
fm3, line 7: reads like a copy-paste error?

webknossos-datastore/app/com/scalableminds/webknossos/datastore/controllers/DataSourceController.scala
```diff
@@ -177,6 +189,22 @@ class DataSourceController @Inject()(
...
+      for {
+        dataSourceId <- uploadService.getDataSourceIdByUploadId(
+          uploadService.extractDatasetUploadId(resumableIdentifier)) ?~> "dataset.upload.validation.failed"
+        result <- accessTokenService.validateAccess(UserAccessRequest.writeDataSource(dataSourceId),
+                                                    urlOrHeaderToken(token, request)) {
+          for {
+            isKnownUpload <- uploadService.isKnownUploadByFileId(resumableIdentifier)
+            _ <- bool2Fox(isKnownUpload) ?~> "dataset.upload.validation.failed"
+            isPresent <- uploadService.isChunkPresent(resumableIdentifier, resumableChunkNumber)
+          } yield if (isPresent) Ok else NoContent
```
fm3, line 203: nice :sunglasses:

webknossos-datastore/app/com/scalableminds/webknossos/datastore/services/DSRemoteWebknossosClient.scala
```diff
@@ -67,6 +68,16 @@ class DSRemoteWebknossosClient @Inject()(
...
 
+  def getOngoingUploadsForUser(userTokenOpt: Option[String], organizationName: String): Fox[List[OngoingUpload]] =
+    for {
+      userToken <- option2Fox(userTokenOpt) ?~> "reserveUpload.noUserToken"
+      datasetIds <- rpc(s"$webknossosUri/api/datastores/$dataStoreName/getOngoingUploadsForUser")
+        .addQueryString("key" -> dataStoreKey)
+        .addQueryString("token" -> userToken)
+        .addQueryString("organizationName" -> organizationName)
+        .getWithJsonResponse[List[OngoingUpload]]
+    } yield datasetIds
```
fm3, line 79:
```suggestion
    } yield ongoingUploads
```
(is what is returned here, right? I’d rename the variable accordingly)

webknossos-datastore/app/com/scalableminds/webknossos/datastore/services/uploading/UploadService.scala
```diff
@@ -141,9 +147,47 @@ class UploadService @Inject()(dataSourceRepository: DataSourceRepository,
...
+      maybeOngoingUploads <- Fox.sequence(
+        ongoingUploadsWithoutIds.map(
+          upload =>
+            runningUploadMetadataStore
+              .find(Json.stringify(Json.toJson(upload.dataSourceId)))
+              .map(uploadIdOpt => uploadIdOpt.map(uploadId => upload.copy(uploadId = uploadId)))
+        ))
+      foundOngoingUploads = maybeOngoingUploads.filter(idBox => idBox.isDefined).flatMap(idBox => idBox.value).collect {
+        case Some(value) => value
+      }
```
fm3, line 161: This seems convoluted. what is the type of maybeOngoingUploads? I think flatten/flatMap may already do what you need?

webknossos-datastore/app/com/scalableminds/webknossos/datastore/storage/RedisTemporaryStore.scala
```diff
@@ -93,6 +93,11 @@ trait RedisTemporaryStore extends LazyLogging {
       r.sadd(id, value).getOrElse(0L) > 0
     }
 
+  def containedInSet(id: String, value: String): Fox[Boolean] =
```
fm3, line 96:
```suggestion
  def isContainedInSet(id: String, value: String): Fox[Boolean] =
```

webknossos-datastore/app/com/scalableminds/webknossos/datastore/services/uploading/UploadService.scala
```diff
@@ -133,6 +135,10 @@ class UploadService @Inject()(dataSourceRepository: DataSourceRepository,
         redisKeyForDataSourceId(reserveUploadInformation.uploadId),
         Json.stringify(Json.toJson(DataSourceId(reserveUploadInformation.name, reserveUploadInformation.organization)))
       )
+      _ <- runningUploadMetadataStore.insert(
+        Json.stringify(Json.toJson(DataSourceId(reserveUploadInformation.name, reserveUploadInformation.organization))),
+        reserveUploadInformation.uploadId
+      )
       _ <- runningUploadMetadataStore.insert(
         redisKeyForLinkedLayerIdentifier(reserveUploadInformation.uploadId),
```
fm3, line 143: Now we are inserting two different types of data into the same set. Maybe this could be unified? Otherwise, it may be better practice to have two stores and do the set existence test on each specific one?

fm3 (changes requested):
Good stuff! Works nicely, also for multi-file uploads. I added a couple of comments to the backend code.

Also during testing I had a couple of ideas to improve the UX:
 - the unfinished upload entry should be removed once I clicked on continue. hide banner entirely if that makes the list empty
 - can the voxel size also be pre-filled? (for in needsConversion case)
 - would it be simple to do some sanity checks before I can hit upload again? e.g. number of files to upload is equal to that of the unfinished upload?

Additionally, different wording is used in many places (I saw pending uploads, OngoingUpload, unfinished upload, in progress uploads). Please pick one and unify, both in code and UI :) I’d vote for unfinished upload, except where an explicit difference from that needs to be expressed.

webknossos-datastore/app/com/scalableminds/webknossos/datastore/services/uploading/UploadService.scala
MichaelBuessemeyer: I found a more consice version: 
```scala
foundOngoingUploads = maybeOngoingUploads.collect { case Full(Some(value)) => value }
```

The type of `maybeOngoingUploads` is `List[Box[Option[OngoingUpload]]]`

MichaelBuessemeyer (commented):
Thanks for the feedback :pray:. It should be applied now except for one comment which I do not understand and need clarification. I commented the comment in question :)

MichaelBuessemeyer:
And now everything should be named consistently. At least I tried :see_no_evil:. I'll do retesting tomorrow :)

MichaelBuessemeyer:
>  I'll do retesting tomorrow :)

Should work :)

fm3 (commented):
I replied to the thread about redis keys. Other than that, the backend looks good to me now :)

MichaelBuessemeyer:
And what about 


> > can the voxel size also be pre-filled? (for in needsConversion case)
> 
> This information is not present in the backend as it is only passed to the worker conversion job once the conversion job is started. Storing this in the backend would need some further implementation effort :)
> 
> > would it be simple to do some sanity checks before I can hit upload again? e.g. number of files to upload is equal to that of the unfinished upload?
> 
> I think that would be possible. I would need to retrieve this information from Redis and then append it to the OngoingUploads case class objects (which will be renamed soon). I might need some time but definitely doable. The question: Should this be included in the first version?

?

fm3:
Ah, sorry, I missed that. Ok then let’s skip the pre-filling for voxel size. I think the little sanity check would be good to have already in this PR.

fm3 (commented):
Nice, thanks for adding that! Worked for me! Backend LGTM, leaving final approval to frontend folks

MichaelBuessemeyer:
I tweaked the frontend check a little (compared to the version from Friday):
1. The frontend does no longer check for the exact order of files selected for the upload. It only ensures that files with the exact naming are present as from the initial upload try
2. The frontend now lists all files required for the continued upload in the error message. Previously, it was only guess work and thus didn't feel so good / easy to me when one file was missing and one had to guess which file might be missing for a multifile upload

frontend/javascripts/admin/dataset/dataset_upload_view.tsx
```diff
@@ -890,6 +1014,29 @@ class DatasetUploadView extends React.Component<PropsWithFormAndRouter, State> {
...
+                      return (
+                        unfinishedUploadToContinue.filePaths.length === filePaths.length &&
+                        _.difference(unfinishedUploadToContinue.filePaths, filePaths).length === 0
+                      );
+                    },
+                    "The selected files do not match the files of the unfinished upload. Please select the same files as before." +
+                      unfinishedUploadToContinue?.filePaths !=
+                      null
+                      ? `The file names are ${unfinishedUploadToContinue?.filePaths?.join(", ")}.`
+                      : "",
```
philippotto, line 1037: I think this is missing parentheses (the string is concated with the != null condition right now). I found this because I didn't see the proper error message during testing.

MichaelBuessemeyer: Hmm strange, afaik this worked perfectly fine for me. But you are certainly correct, that the parentheses are needed due to operator precedence :thinking:

Thanks for catching this.

MichaelBuessemeyer:
Should be good now :)

PR #7988, opened 2024-08-12
Allow expanding/collapsing tree groups in locked annotations

dieknolle3333:
- https://fixskeletongroupexpansionlockedannotations.webknossos.xyz/

### Steps to test:
- open locked annotation and expand/collapse skeleton groups
  - test "collapse/expand all subgroups" actions, at least with some nested groups
  - do the same in the segments tab
  - make sure neither expansion is actually saved
- in an editable annotation, make sure expansion/collapsing of segment and skeleton groups works like before
- make sure you can still save whether the groups are expanded or collapsed

### TODOs:
- [x] find out how to improve action. right now manipulated segment groups can be set like this, although the annotation is locked
- [x] improve naming
- [x] add the same logic for segments

### Issues:
- follow-up for #7939

------


frontend/javascripts/oxalis/view/right-border-tabs/segments_tab/segments_view.tsx
```diff
@@ -1520,18 +1499,19 @@ class SegmentsView extends React.Component<Props, State> {
...
+      return mappedIdsToKeys.concat(getNodeKey(GroupTypeEnum.GROUP, this.props.selectedIds.group));
     }
     return mappedIdsToKeys;
   };
 
   getSegmentOrGroupIdsForKeys = (segmentOrGroupKeys: Key[]) => {
     const selectedIds: { segments: number[]; group: number | null } = { segments: [], group: null };
+    const groupPrefix = "Group-";
     segmentOrGroupKeys.forEach((key) => {
       const keyAsString = String(key);
```
philippotto, line 1511: in which cases is key not a string? in general, I wonder why the `Key` type is used, because it seems like these should always be strings? :thinking_face:

philippotto (commented):
didn't test yet, but left some code comments :)

frontend/javascripts/oxalis/view/right-border-tabs/segments_tab/segments_view.tsx
dieknolle3333: It's because the Antd component is using `Key` and I wanted to avoid type conversion. E.g. the tree `onSelect` method needs to take the selected keys as `Key`s. So in the `onSelectTreeItem` method I would need to handle the case that the keys are numbers and convert them to strings. I felt like using `Key` as a type was nice and easy

philippotto: ok, I understand the idea :) the issue is that antd supports numbers and strings for keys, but we never use that in this context. using `Key` everywhere widens the type unnecessarily across various places. for example, if future-me looks into this action in a couple of months, I will definitely wonder why one can pass a number to the expand action (even though, that should never happen and won't really work). so, it's better to catch the "imprecise type" as early as possible and narrow it down.

> So in the onSelectTreeItem method I would need to handle the case that the keys are numbers and convert them to strings.

yes, that would be a good place to narrow the type. I wouldn't even convert from number to string, because this should never happen anyway. if it wasn't an array, I'd throw an exception. however, for an array, that assertion is too costly in my opinion. i'd simply cast from Key to string :) then, you also don't need to do the `String(key)` conversion in line 1511.

dieknolle3333:
TODO: fix that nested segment groups cant be selected

PR #7992, opened 2024-08-14
[WIP] Compose TPS

philippotto:
- https://___.webknossos.xyz

### Steps to test:
- abc

### TODOs:
- [ ] ...

### Issues:
- fixes #

------

### DIFF
--- a/frontend/javascripts/admin/dataset/composition_wizard/04_configure_new_dataset.tsx
+++ b/frontend/javascripts/admin/dataset/composition_wizard/04_configure_new_dataset.tsx
@@ -55,9 +55,13 @@ export function ConfigureNewDataset(props: WizardComponentProps) {
     const transformationArr =
       sourcePoints.length > 0 && targetPoints.length > 0
         ? [
+            // {
+            //   type: "affine" as const,
+            //   matrix: flatToNestedMatrix(estimateAffineMatrix4x4(sourcePoints, targetPoints)),
+            // },
             {
-              type: "affine" as const,
-              matrix: flatToNestedMatrix(estimateAffineMatrix4x4(sourcePoints, targetPoints)),
+              type: "thin_plate_spline" as const,
+              correspondences: { source: sourcePoints, target: targetPoints },
             },
           ]
         : [];
--- a/.circleci/not-on-master.sh
+++ b/.circleci/not-on-master.sh
@@ -1,8 +1,8 @@
 #!/usr/bin/env bash
 set -Eeuo pipefail
 
-if [ "${CIRCLE_BRANCH}" == "master" ]; then
+# if [ "${CIRCLE_BRANCH}" == "master" ]; then
   echo "Skipping this step on master..."
-else
-  exec "$@"
-fi
+# else
+#   exec "$@"
+# fi


frontend/javascripts/oxalis/model/accessors/dataset_accessor.ts
```diff
@@ -722,7 +722,7 @@ function _getOriginalTransformsForLayerOrNull(
   } else if (type === "thin_plate_spline") {
     const { source, target } = transformation.correspondences;
 
-    return createThinPlateSplineTransform(target, source, dataset.dataSource.scale.factor);
+    return createThinPlateSplineTransform(source, target, dataset.dataSource.scale.factor);
```
philippotto, line 725: this was an error before. since we don't have any tps in production, there shouldnt be any problems.

MichaelBuessemeyer (changes requested):
The code looks good to me but sadly I found something during testing locally:

When I used the following nmls I could not submit the dataset creation (using affine transformations) and the console showed the following error:
image
Here are the nmls I used:
[l4_first_data_types_second.zip](https://github.com/user-attachments/files/16662661/l4_first_data_types_second.zip)
The nml & dataset order was l4_sample first and then data_types.

Creating a tps dataset with the exact same setup works but showing the dataset crashes due to the same error

MichaelBuessemeyer:
Ok with the help of @valentin-pinkau I was able to identify that all nodes must not be present in the same z slice. This fixes the matrix calculation (the error above).

But I am still unable to see anything with the test setup described above. But that might be due to using two completely different datasets :see_no_evil:

MichaelBuessemeyer:
>But I am still unable to see anything with the test setup described above. But that might be due to using two completely different datasets 🙈

For the record: Turns out, that the bbox of the dataset is the union of both layers and thus the default view starts at the default center of the "too large" bounding box resulting in not showing any data initially. Using the find my data button solved the problem :)

philippotto:
@MichaelBuessemeyer please have a look at my latest commit :)

frontend/javascripts/admin/dataset/composition_wizard/04_configure_new_dataset.tsx
```diff
@@ -1,30 +1,51 @@
...
+import { Vector3 } from "oxalis/constants";
+import { WkDevFlags } from "oxalis/api/wk_dev";
 
 const FormItem = Form.Item;
 
+async function guardedWithErrorToast(fn: () => Promise<any>) {
+  try {
+    await fn();
+  } catch (error) {
+    Toast.error("An error unexpected occurred. Please check the console for details");
```
MichaelBuessemeyer, line 43:
```suggestion
    Toast.error("An unexpected error occurred. Please check the console for details");
```

frontend/javascripts/admin/dataset/composition_wizard/04_configure_new_dataset.tsx
```diff
@@ -1,30 +1,51 @@
...
+    console.error(error);
+    ErrorHandling.notify(error as Error);
+  }
+}
```
MichaelBuessemeyer, line 47: Isn't that a potential candidate for utils.ts?

frontend/javascripts/admin/dataset/composition_wizard/04_configure_new_dataset.tsx
```diff
@@ -97,6 +140,35 @@ export function ConfigureNewDataset(props: WizardComponentProps) {
...
+      if (tryAugmentation) {
+        layersWithTransforms = withTransforms(
+          layersWithoutTransforms,
+          augmentLandmarks(sourcePoints),
+          augmentLandmarks(targetPoints),
+        );
+      } else {
+        throw exception;
+      }
+    }
```
MichaelBuessemeyer, line 170: I am afraid, that always using z for "extrusion" won't cover all cases. In case I annotate all my point in the zy viewport without moving in x direction, the x axis value of all points should be equal. Thus, moving the points by z should fix the problem here. In case we want to catch all possible planes, we should calculate the otrhogol vector to the plane and move the nodes in that direction.


But I can also understand that this might be a bit too much for this pr and can be solved in a follow up pr :)

MichaelBuessemeyer (commented):
I only found minor things but also identified, that pushing a copy of all points in the z direction does not cover all cases. Feel free to do this in a follow up pr :)

Will retest now

MichaelBuessemeyer (approved):
Testing went well :)

Feel free to merge in case you want to do the only z as a direction for extrusion in another pr

philippotto:
Thanks for the review + test!

> I only found minor things but also identified, that pushing a copy of all points in the z direction does not cover all cases. Feel free to do this in a follow up pr :)

Yes, I think, this is fine, because the user is asked whether a constant translation in Z should be assumed. if this is not the case, they should decline and fix it themself.

PR #8003, opened 2024-08-15
Add warning if opacity of volume layer is zero

### LINKED ISSUE #7407: Add warning icon to layer if it's visible, but opacity is 0
Multiple times, it has bitten us that a layer is enabled, but it's opacity is set to 0. For example, if a volume layer is enabled, but opacity=0, the volume tools are disabled. It's hard to miss this which is why I'd suggest to put a yellow warning-icon next to the layer name if this is the case.

dieknolle3333:
### Steps to test:
- Open an annotation with a volume layer
- Using the slider, set the opacity of a volume layer to zero
- Then a yellow (/orange) warning icon should appear next to the opacity label
- It should disappear when the opacity is set to a value higher than 0

### TODOs:
- [x] layout
- [x] color
- [x] code structure
- [x] test

### Issues:
- fixes #7407 

------


philippotto (approved):
great :+1:

PR #8006, opened 2024-08-16
add links for skipped tasks in voxelytics workflow reports

### LINKED ISSUE #7080: Display existing artifacts for skipped tasks in voxelytics reports
## Detailed Description

Consider the case where artifact A is computed by workflow A. Then, there is a workflow B which reuses artifact A, but continues differently. Currently, the output paths, logs, and statistics for artifact A are only displayed in the voxelytics report for workflow A. It would be very helpful if it would also be displayed in the report for workflow B.

Apart from artifact reuse, this issue also occurs when iterating on a workflow. Every time a name is changed or another task is added to the workflow, a new voxelytics report is created losing all information about previously computed artifacts.

Displaying the logs and statistics for artifacts computed in another workflow could be challenging, because they exist per-run and there is a global run selection. One idea would be to only show the output paths for existing artifacts for skipped tasks and then link to the original workflow report where the logs and stats can be inspected.

normanrz:
This PR (finally) adds links for skipped tasks in the workflow reports to the workflows/runs where the artifacts were produced. The artifact information is also shown inline.
Screenshot 2024-08-16 at 21 30 36

### Steps to test:
- Get 2 workflows, 1 that has just the first few tasks and a second that appends more tasks
- Run the first workflow; then, run the second workflow skipping the existing tasks
- Look at the workflow report of the second workflow

### Issues:
- fixes #7080

------


app/models/voxelytics/VoxelyticsDAO.scala
```diff
@@ -134,10 +137,24 @@ class VoxelyticsDAO @Inject()(sqlClient: SqlClient)(implicit ec: ExecutionContex
...
           a.version,
           a.metadata,
-          t.name AS taskName
+          t.name AS taskName,
+          o.workflow_hash AS other_workflow_hash,
+          o._run AS other_run
         FROM webknossos.voxelytics_artifacts a
         JOIN latest_complete_tasks t ON t._id = a._task
-        """.as[(String, String, String, String, Long, Long, String, String, String)])
+        LEFT JOIN ( -- when the task is skipped, the artifact from the producing task run is joined for linking
```
fm3, line 145: please add the discussed distinct constraint to the schema, and a comment here explaining how the fanout effect is prevented (joining at most one item, due to the distinct constraint)

fm3 (commented):
Looking good already :)

fm3 (commented):
Works for me and backend looks good :)
Waiting for final approval from @daniel-wer 

Maybe a tooltip on the links to the “foreign workflow” would be nice, but not super important

daniel-wer (approved):
Spectacular :tada: This already made my week and is going to be very helpful in the future :)

Should I retest this somehow or do you think it's fine to test this in prod using the real-world reports?

How does this affect the file size shown next to a workflow in the workflow list? Same for the "Show Disk Usage of Artifacts" modal that can be displayed from within a report? I don't have a strong opinion on which way it should be counted, but I'd like to at least document the current behavior in this PR.

normanrz:
> How does this affect the file size shown next to a workflow in the workflow list? Same for the "Show Disk Usage of Artifacts" modal that can be displayed from within a report? I don't have a strong opinion on which way it should be counted, but I'd like to at least document the current behavior in this PR.

The storage display in the workflow list is not affected by this change. I think that makes sense. Only artifacts that are produced by that particular workflow are counted.

The "Show Disk Usage of Artifacts" would show include the foreign artifacts. Should I filter them out for consistency?

daniel-wer:
> The "Show Disk Usage of Artifacts" would show include the foreign artifacts. Should I filter them out for consistency?

I'd say it's ok as it is

PR #8007, opened 2024-08-16
Upgrade ant icons to 5.4

normanrz:
Upgrades ant icons to 5.4 matching the version of the main antd package.
Also, ant-icons now use inline svg instead of iconfonts, so we can remove our custom handling of the iconfont.

- [x] Changelog


package.json
```diff
@@ -1,17 +1,14 @@
 {
     "name": "webknossos",
     "license": "AGPL-3.0",
-    "engines":
```
philippotto, line 4: in another pr, I also had the problem that the package.json was entirely reformatted. later I had annoying merge conflicts in other prs. could you try to undo these changes?
the proper solution will be to update biome and then let it run on package.json, too. the current version will always skip package.json (see https://github.com/biomejs/biome/issues/1668).
updating requires lots of small fixes, though, as biome got smarter. so, this should be a different pr.

PR #8011, opened 2024-08-20
Include vx workflow link in slack notification for jobs that have one

fm3:
makes the slack notification for successful and failed jobs a bit more useful :)

contributes to #5236

I tested locally that the output looks good =) No need to retest as this change is very small


app/models/job/JobService.scala
```diff
@@ -72,11 +73,11 @@ class JobService @Inject()(wkConf: WkConf,
       organization <- organizationDAO.findOne(user._organization)(GlobalAccessContext)
       resultLink = jobAfterChange.resultLinkPublic(organization.name, wkConf.Http.uri)
       resultLinkSlack = jobAfterChange.resultLinkSlackFormatted(organization.name, wkConf.Http.uri)
+      workflowLink = jobAfterChange.workflowLinkSlackFormatted(wkConf.Http.uri)
       multiUser <- multiUserDAO.findOne(user._multiUser)(GlobalAccessContext)
       superUserLabel = if (multiUser.isSuperUser) " (for superuser)" else ""
       durationLabel = jobAfterChange.duration.map(d => s" after ${formatDuration(d)}").getOrElse("")
-      msg = s"Job ${jobBeforeChange._id} succeeded$durationLabel. Command ${jobBeforeChange.command}, organization: ${organization.displayName}.${resultLinkSlack
-        .getOrElse("")}"
+      msg = s"Job ${jobBeforeChange._id} succeeded$durationLabel. Command ${jobBeforeChange.command}, organization: ${organization.displayName}.$resultLinkSlack$workflowLink"
```
MichaelBuessemeyer, line 80: Is it always one or the other?  otherwise I would at least add a space inbetween `$resultLinkSlack$workflowLink`

MichaelBuessemeyer (approved):

MichaelBuessemeyer (approved):
Changes look good to me except for one small potential improvement :)

app/models/job/JobService.scala
fm3: Both include their own leading space if they are set. This way, no double spaces are included if both are none.

PR #8015, opened 2024-08-20
keep asset timestamps from filesystem when building jar

fm3:
As documented in https://github.com/playframework/playframework/issues/10572 this option can be set to no longer overwrite the asset modified timestamps (which play uses for Last-Modified http headers).

With this change, my guess is that the CI build time will be used (because git sets the file attributes on checkout).

I tried this locally with `sbt clean compile stage`, `docker-compose build --pull webknossos` and `./start_docker.sh` and it seems to work. Dev instance should be even easier

- https://assettimestamps.webknossos.xyz

### Steps to test:
- Request assets
- Last-Modified header should look reasonable.


philippotto (approved):
nice, works, thank you!

PR #8019, opened 2024-08-22
[WIP] Use pnpm

philippotto:
Installation of frontend modules would go down from ~80s to ~10s with this PR. However, the asset compilation in sbt still fails, because it cannot find pnpm (even though, it is in the PATH). Putting this PR into the backlog for now...

- https://___.webknossos.xyz

### Steps to test:
- abc

### TODOs:
- [ ] ...

### Issues:
- fixes #

------


philippotto:
install performance can be compared with https://github.com/scalableminds/webknossos/issues/8051 before fixing the scala problem in the ci.

PR #8022, opened 2024-08-22
Fix active tree's comments not getting scrolled into view when switching to comments tab

MichaelBuessemeyer:
Currently, the active tree's comment is not reliably highlighted when switching from skeleton to the comments tab due to some likely unexpected behavior: Once the comments tab is opened but then not visible, e.g. hidden behind the skeleton tab, the comments tab is still active but not antd is being rendered. Thus, the ref to this antd tree component is null which results in the component not scrolling to the active tree in the described case. To fix this behaviour I did some state lifting: The comments tab is now aware whether it is visible. When the component is visible again, the scroll to call is triggered and thus the active tree is shown :)

- https://___.webknossos.xyz

### Steps to test:
- Open an skeleton annotation
- create some skeletons with comments
- reduce the browser window height to decrease the amount of needed comments to get the comments tab into scrolling behaviour
- save and reload
- open skeleton tab, select a tree with at least one comment and switch to the comments tab. The tree's comments should be scrolled into view upon switching
- Optional: Compare with master behaviour

### Issues:
- fixes https://discuss.webknossos.org/t/comments-for-current-tree-are-not-listed-at-the-top/1855/1

------


philippotto (approved):
thank you for fixing this! works great :+1:

PR #8028, opened 2024-08-26
save some io when loading oversegmentation meshfiles

fm3:
- Instantiates the MeshFileService as singleton in the Datastore. This way, all requests will go to the same MeshFileCache. (Note that it caches only the hdf5 file handles, does not do that much)
 - When reading info for multiple segments, metadata that is constant for the whole meshfile is no longer read for each segment, but only once. I hope that this especially saves seek time on spinning disks, since this info is probably at the very beginning of the hdf5 file.

### Steps to test:
- Load some oversegmentation-based precomputed meshes with active mapping, should still work (e.g. test-agglomerate-file dataset)
- Hopefully a little faster on spinning disks (?)

---


MichaelBuessemeyer (approved):
Nice the code looks good :tada:. 
I only left one comment about the introduced tryo, so that it is added on purpose.

>Load some oversegmentation-based precomputed meshes with active mapping, should still work (e.g. test-agglomerate-file dataset)

Works :+1:

PR #8049, opened 2024-09-02
Make "Go to center of bounding box" available at one click

### LINKED ISSUE #8001: Make go to center of BBox a priority
## Detailed Description

Currently, jumping to the center of a bbox is only available via the menu in a bbox tab entry but it is a frequent use case to jump to the bbox. However, the color picker for the bounding boxes is not such a frequent action. Therefore, I suggest to move the color picker into the bbox menu and pull the "move to center" out into an easily accessible button where the color picker is currently located. 

## Context

dieknolle3333:
- https://centerbbsingleclick.webknossos.xyz/

### Steps to test:
- Open an annotation with some bounding boxes
- Click anywhere in a bounding box entry. This should center the bounding box within the view ports
- Editing the color, the name, the coordinates, clicking the context menu or any of its entries or switching the visibility shouldnt center the bbox
- the rest of the UI should be working as before

### TODOs:
- [x] editing the coordinates or the color shouldnt focus bbox

### Issues:
- fixes #8001

------


philippotto:
cool!! ideally, the cursor should also indicate that the entire row is clickable. this should work with css (`cursor: pointer`).

philippotto:
getting there :smile: unfortunately, the hover target area is not identical to the click target area. clicking close to the edge does not work (the cursor also rightfully indicates that):

image

maybe a margin or padding needs to be adapted or the click handler (+ css class) needs to be one level further up in the hierarchy?

dieknolle3333:
Very good point! I think I fixed it :)

philippotto (approved):
awesome! feels really slick :)

PR #8054, opened 2024-09-03
Improve Bounding Box Selection

### LINKED ISSUE #8004: Improve BBox Selection Mechanism
## Detailed Description
Currently, it is only possible to select a bounding box as active and highlight it in the bbox tab in the ortho views while the bbox tool is active. It would be nice to be able to select bbox also in move tool; just like tree nodes. Moreover, BBox selection in 3d viewport would also be very awesome.

From (high-pri + easy) to (low-pri + medium):

- [x] Right-click on a bbox should also have "Focus in BBox tab" entry
- [x] Allow to select a bbox in move tool via `ctrl + click` (just like in bbox tool)
- [ ] Allow to select a bbox in 3d viewport

## Context

Use cases:
- easier navigate bboxes
- No need to be afraid of accidentally moving or creating a bounding box bc. the selection works also in move tool and not only when bbox tool is active.

dieknolle3333:
- https://improvebboxselection.webknossos.xyz/

### Steps to test:
- Open an annotation with some bounding boxes
- Within a view port (except 3D), right-click on a bounding box edge. In the context menu, a new entry should appear, that lets you focus the BB in the BB tab (if open). Make sure this work. If the list entry wasnt visible due to scrolling, clicking this new context menu action should focus it
- Hovering above the edge of a bounding box should highlight it
- Within the move tool, left clicking the edge of a bounding box should focus it in the bounding box tab and make it the active bounding box
- This behaviour (on click) should not interfere with other tools. Although the highlighting of the edges is still active in other tools, similar to the highlighting of segments.

### TODOs:
- [x]  Allow to select a bbox in move tool via ctrl + click (just like in bbox tool)
    - [x] what about event.shiftKey || !useLegacyBindings
- [ ] ~~Allow to select a bbox in 3d viewport~~ is a separate issue
- [x] In any tool, the bbox borders are highlighted upon hovering. Why? Is this good? -> "If a tool does not define a specific mouse binding, the bindings of the MoveTool are used as a fallback."

### Issues:
- fixes #8004 

------


philippotto (approved):
awesome, works great :+1:

PR #8069, opened 2024-09-11
Add assertion for no duplicated keys in dataset and folder metadata in update routes

### LINKED ISSUE #8068: Validate for no duplicate metadata keys in folder & dataset metadata
## Context

Currently, there is only frontend validation for duplicate keys in metadata info of datasets and folders. The backend should also valide this upon update requests.

For more details on this see discussion https://github.com/scalableminds/webknossos-libs/pull/1163#discussion_r1727104418

Note: This is not a real bug and nothing severe but should be implemented.

MichaelBuessemeyer:
This PR adds an assertion to not have duplicate keys in metadata updates to datasets and folders.

- https://___.webknossos.xyz

### Steps to test:
- I suggest testing in firefox
- Open the dev instance, after it is loaded open the network tab
- Edit some metadata
- The network tab should show the update request. On this request use right click and select "edit and resent"
- Duplicate an already existing entry in the metadata of the dataset or folder and submit the request
- The request should result in a 400 bad request with the message: `Metadata keys must be unique.`

### Issues:
- fixes #8068

------


app/controllers/DatasetController.scala
```diff
@@ -69,6 +69,13 @@ object SegmentAnythingMaskParameters {
   implicit val jsonFormat: Format[SegmentAnythingMaskParameters] = Json.format[SegmentAnythingMaskParameters]
 }
 
+trait MetadataAssertions {
+  def assertNoDuplicateMetadataKeys(metadata: JsArray)(implicit ec: ExecutionContext, provider: MessagesProvider): Fox[Unit] = {
+    val keys = metadata.value.flatMap(_.as[JsObject] \\ "key").map(_.as[String]).toList
+    if (keys.size == keys.distinct.size) Fox.successful(()) else Fox.failure(Messages("dataset.metadata.duplicateKeys"))
+  }
+}
+
```
MichaelBuessemeyer, line 78: Maybe you know a better place where to put this in the code base. 

I figured that a trait is better to not have to implement `assertNoDuplicateMetadataKeys` in both controllers of dataset and folder

fm3 (approved):
LGTM

app/controllers/DatasetController.scala
MichaelBuessemeyer: Ok I put it into  `app/utils` into an own file :)

PR #8071, opened 2024-09-11
Allow training a model on multiple annotations

### LINKED ISSUE #8005: Start train_model with multiple annotations
## Detailed Description
Add a new modal in the "AI models" list to start a new training. The modal should have a selector (in the simplest case a number of text boxes for URLs or IDs) to select annotations to train with. The other options of the existing train model modal should also be included, if possible.

## Context
Training new models from annotations is already a breeze. However, we often times want to train with data from multiple annotations (e.g. to include data from multiple datasets).

philippotto:
- https://multiannotraining.webknossos.xyz
- I already created a few annotations (with differing configurations) and made the default user a super-admin
- details about the annotations:
	- https://multiannotraining.webknossos.xyz/annotations/66e2e04b010000df00b3f6fa
		- no-tracing segmentation layer: "segmentation"
	- https://multiannotraining.webknossos.xyz/annotations/66e2dfe6010000ab00b3f6f7
		- volume layer with fallback: "segmentation"
	- https://multiannotraining.webknossos.xyz/annotations/66e2e06d010000ad00b3f6ff
		- volume layer with fallback: "custom_segmentation"
	- https://multiannotraining.webknossos.xyz/annotations/66e2e08c010000ad00b3f702
		- volume layer with fallback: custom_segmentation
		- volume layer without fallback: custom_empty
			- finest mag is mag4
	- https://multiannotraining.webknossos.xyz/annotations/66e2e0d6010000ac00b3f706
		- no-tracing segmentation layer: "segmentation"
		- volume layer without fallback: custom_empty

### Steps to test:
- go to admin > ai models
- click on the "train new model" button
- paste this CSV
```
66e2e04b010000df00b3f6fa
https://multiannotraining.webknossos.xyz/annotations/66e2dfe6010000ab00b3f6f7
https://multiannotraining.webknossos.xyz/annotations/66e2e06d010000ad00b3f6ff
https://multiannotraining.webknossos.xyz/annotations/66e2e08c010000ad00b3f702
https://multiannotraining.webknossos.xyz/annotations/66e2e0d6010000ac00b3f706
```
- click "add"
- use the form
- submit it (it will fail because there is no worker configured on the dev instance). however, you can use the network tab to inspect the sent payload.

image
image
- also open an annotation and use the train-model dialog from there to ensure that it still works
- special care needs to be taken for the volume layer name, because we usually use the tracing id as the name and then have a human-readable name for the UI. however, the worker always wants the human readable one and never the tracing id.

### Issues:
- fixes #8005

------


frontend/javascripts/oxalis/default_state.ts
```diff
@@ -240,7 +240,7 @@ const defaultState: OxalisState = {
     showVersionRestore: false,
     showDownloadModal: false,
     showPythonClientModal: false,
-    aIJobModalState: "invisible",
+    aIJobModalState: "neuron_inferral",
```
philippotto, line 243:
```suggestion
    aIJobModalState: "invisible",
```

frontend/javascripts/oxalis/view/jobs/train_ai_model.tsx
```diff
@@ -67,106 +201,122 @@ export function TrainAiModelTab({ onClose }: { onClose: () => void }) {
...
 
-  const segmentationLayers = getSegmentationLayers(dataset);
-  const fixedSelectedLayer = segmentationLayers.length === 1 ? segmentationLayers[0] : null;
-
-  const userBoundingBoxes = useSelector((state: OxalisState) =>
-    getUserBoundingBoxesFromState(state),
-  );
   const bboxesVoxelCount = _.sum(
-    userBoundingBoxes.map((bbox) => new BoundingBox(bbox.boundingBox).getVolume()),
+    (userBoundingBoxes || []).map((bbox) => new BoundingBox(bbox.boundingBox).getVolume()),
```
daniel-wer, line 217: While I'm reading this I thought of a potential issue. There is an assertion that there is at least one bounding box (later in the file). Now that multiple annotations can be supplied, it should be checked that each of them includes at least one bounding box - otherwise it does not make sense to supply that annotation since voxelytics won't use it for training.

frontend/javascripts/oxalis/view/jobs/train_ai_model.tsx
```diff
@@ -67,106 +201,122 @@ export function TrainAiModelTab({ onClose }: { onClose: () => void }) {
...
+            .filter((layer) => layer.typ === "Volume")
+            .map((layer) => layer.name),
+        ]);
+        const segmentationLayers: Array<{ name: string }> = segmentationLayerNames.map(
+          (layerName) => ({
+            name: layerName,
+          }),
+        );
+
+        const fixedSelectedLayer = segmentationLayers.length === 1 ? segmentationLayers[0] : null;
```
daniel-wer, line 251: Can you rename this to indicate that this is the fixed segmentation layer. While you're at it, you could also unify it with the fixing of the color layer (line 275). Either fix both in the props or both here.

frontend/javascripts/oxalis/view/jobs/train_ai_model.tsx
```diff
@@ -67,106 +201,122 @@ export function TrainAiModelTab({ onClose }: { onClose: () => void }) {
...
+            <Col span={8}>
+              <LayerSelectionFormItem
+                name={["trainingAnnotations", idx, "layerName"]}
+                chooseSegmentationLayer
+                layers={segmentationLayers}
+                getReadableNameForLayer={(layer) => {
+                  return layer.name;
+                }}
+                fixedLayerName={fixedSelectedLayer?.name || undefined}
+                label="Groundtruth Layer"
```
daniel-wer, line 300:
```suggestion
                label="Ground Truth Layer"
```

frontend/javascripts/oxalis/view/jobs/train_ai_model.tsx
```diff
@@ -67,106 +201,122 @@ export function TrainAiModelTab({ onClose }: { onClose: () => void }) {
...
         setActive={setUseCustomWorkflow}
       />
 
-      <FormItem hasFeedback name="dummy" label="Training Data">
-        <div>
-          {userBoundingBoxes.length} bounding boxes ({formatVoxels(bboxesVoxelCount)})
-        </div>
-      </FormItem>
+      {userBoundingBoxes != null ? (
+        <FormItem hasFeedback name="dummy" label="Training Data">
```
daniel-wer, line 314: You did not introduce it, but do you know whether there's a good reason for that `name="dummy"` or should it be replaced by something more descriptive?

frontend/javascripts/oxalis/view/jobs/train_ai_model.tsx
```diff
@@ -240,3 +393,98 @@ function areBoundingBoxesValid(userBoundingBoxes: UserBoundingBox[]): {
...
+}: {
+  onAdd: (newItems: Array<AnnotationWithDataset<APIAnnotation>>) => void;
+}) {
+  const [value, setValue] = useState("");
+  const onClickAdd = async () => {
+    const newItems = [];
+
+    const lines = value.split("\n");
+    for (const line of lines) {
+      const [annotationUrlOrId] = line.trim().split(",");
```
daniel-wer, line 408: Isn't there a validation rule disallowing `,` in each line? Also see my feedback from testing regarding that.

frontend/javascripts/oxalis/view/jobs/train_ai_model.tsx
```diff
@@ -240,3 +393,98 @@ function areBoundingBoxesValid(userBoundingBoxes: UserBoundingBox[]): {
...
+      newItems.map(async (item) => {
+        const annotation = await getAnnotationInformation(item.annotationId);
+        const dataset = await getDataset({
+          owningOrganization: annotation.organization,
+          name: annotation.dataSetName,
+        });
+
+        const volumeTracings: ServerVolumeTracing[] = await Promise.all(
+          annotation.annotationLayers
+            .filter((layer) => layer.typ === "Skeleton")
```
daniel-wer, line 430: Is this right or should it be `!==`?

frontend/javascripts/oxalis/view/jobs/train_ai_model.tsx
```diff
@@ -240,3 +393,98 @@ function areBoundingBoxesValid(userBoundingBoxes: UserBoundingBox[]): {
...
+                      "Each line should only contain an annotation ID or URL (without # or ,)",
+                    ),
+                  );
+            },
+          }),
+        ]}
+      >
+        <TextArea
+          className="input-monospace"
+          placeholder="annotationUrlOrId[, colorLayerName, volumeLayerId]"
```
daniel-wer, line 474: See my feedback from testing

daniel-wer (commented):
Thank you for implementing this, I'm looking forward to use the feature! Very thorough implementation :+1: 

Notes from testing:

- [ ] Either the placeholder or the validation for the csv modal is wrong. The placeholder indicates that the color and segmentation layer names can be supplied, but the validation complains that only the url/id should be given.
  placeholder
- [ ] The csv and the training modal are closed when clicking onto the back drop. The second one (and arguably the first one as well) should not do that, because it makes it to easy to lose entered information.
- [ ] Whether the modal is triggered from within an annotation or from the models list, allows to select different layers (i.e., the fallback layer):
  within_annotation
  from_models_list

normanrz:
@MichaelBuessemeyer Could you please carry this over the finish line?

frontend/javascripts/oxalis/view/jobs/train_ai_model.tsx
MichaelBuessemeyer: > While I'm reading this I thought of a potential issue. There is an assertion that there is at least one bounding box (later in the file). Now that multiple annotations can be supplied, it should be checked that each of them includes at least one bounding box - otherwise it does not make sense to supply that annotation since voxelytics won't use it for training.

That is a very good catch. In the current version the ai training modal is completely unaware of whether bounding boxes exist. 
Therefore, I added  the boxes of an annotation to `AnnotationWithDataset` which is a parameter of the `TrainAiModelTab`. It is now called `AnnotationWithDatasetAndBBoxes`. The tracings matching the annotations are now loaded on depend as soon as the csv input is validated to pass the bounding boxes to the `TrainAiModelTab` component.

frontend/javascripts/oxalis/view/jobs/train_ai_model.tsx
MichaelBuessemeyer: I simply removed the option to directly enter `colorLayerName` or `volumeLayerId`. If you think this would be a nice improvement, I'd prefer to implement this as a follow-up. The current code version simply does not support the option to initially supply a color layer name and volume layer id.

MichaelBuessemeyer:
Ok here we go :)
I hope I implemented everything you mentioned or commented it. Could you please do some retesting & re-reviewing of the changes @daniel-wer :pray:?

>  Either the placeholder or the validation for the csv modal is wrong. The placeholder indicates that the color and segmentation layer names can be supplied, but the validation complains that only the url/id should be given.

I removed the placeholder as the current code does not support directly supplying the color and segmentation layer to use for an annotation. If this is a helpful feature I'd prefer to do this as a follow-up :)

> Whether the modal is triggered from within an annotation or from the models list, allows to select different layers (i.e., the fallback layer):

I don't understand what the issue is with the option to be able to select layers. An annotation might have multiple volume layers and thus the segmentation data that should be used is most of the time ambiguous. Same goes for color layers.

frontend/javascripts/oxalis/view/jobs/train_ai_model.tsx
MichaelBuessemeyer: The name field apparently be removed without any errors. :shrug:

frontend/javascripts/oxalis/view/jobs/train_ai_model.tsx
```diff
@@ -369,43 +375,80 @@ export function CollapsibleWorkflowYamlEditor({
   );
 }
 
-function areBoundingBoxesValid(userBoundingBoxes: UserBoundingBox[] | undefined): {
-  valid: boolean;
-  reason: string | null;
+function areAllAnnotationsInvalid<T extends HybridTracing | APIAnnotation>(
```
daniel-wer, line 378: The `All` seems to be incorrect and can probably simply be dropped

frontend/javascripts/oxalis/view/jobs/train_ai_model.tsx
```diff
@@ -425,19 +468,43 @@ function AnnotationsCsvInput({
...
+        const volumeServerTracings: ServerVolumeTracing[] = await Promise.all(
           annotation.annotationLayers
-            .filter((layer) => layer.typ === "Skeleton")
+            .filter((layer) => layer.typ === "Volume")
             .map(
               (layer) =>
                 getTracingForAnnotationType(annotation, layer) as Promise<ServerVolumeTracing>,
             ),
         );
+        // TODO: make bboxs a member again
```
daniel-wer, line 479: TODO

frontend/javascripts/oxalis/view/jobs/train_ai_model.tsx
```diff
@@ -425,19 +468,43 @@ function AnnotationsCsvInput({
...
+            const skeletonTracing = await getTracingForAnnotationType(annotation, skeletonLayer);
+            userBoundingBoxes = convertUserBoundingBoxesFromServerToFrontend(
+              skeletonTracing.userBoundingBoxes,
+            );
+          } else {
+            new Error(`Annotation ${annotation.id} has neither a volume nor a skeleton layer`);
+          }
+        }
+
+        console.log(volumeTracings); // TODOM remove
```
daniel-wer, line 498: TODO

frontend/javascripts/oxalis/view/jobs/train_ai_model.tsx
```diff
@@ -425,19 +468,43 @@ function AnnotationsCsvInput({
...
+          const skeletonLayer = annotation.annotationLayers.find(
+            (layer) => layer.typ === "Skeleton",
+          );
+          if (skeletonLayer) {
+            const skeletonTracing = await getTracingForAnnotationType(annotation, skeletonLayer);
+            userBoundingBoxes = convertUserBoundingBoxesFromServerToFrontend(
+              skeletonTracing.userBoundingBoxes,
+            );
+          } else {
+            new Error(`Annotation ${annotation.id} has neither a volume nor a skeleton layer`);
```
daniel-wer, line 494: Should this error be thrown? Otherwise, I think the line doesn't have any effect :thinking:

frontend/javascripts/oxalis/view/jobs/train_ai_model.tsx
```diff
@@ -425,19 +468,43 @@ function AnnotationsCsvInput({
...
+        if (!userBoundingBoxes) {
+          const skeletonLayer = annotation.annotationLayers.find(
+            (layer) => layer.typ === "Skeleton",
+          );
+          if (skeletonLayer) {
+            const skeletonTracing = await getTracingForAnnotationType(annotation, skeletonLayer);
+            userBoundingBoxes = convertUserBoundingBoxesFromServerToFrontend(
+              skeletonTracing.userBoundingBoxes,
+            );
+          } else {
```
daniel-wer, line 493: Could you add a comment here that user bounding boxes are always saved in all annotation layers, but if there is no volume annotation layer, the skeleton layer is checked. At least that's how I interpret it, but I was a bit confused at first :)

daniel-wer (commented):
Thank you for taking over this PR!

Please also see my previous comment regarding the testing feedback. I can no longer test this on the dev instance, because the AI modal cannot be opened from within an annotation (Philipp patched the code to open the modal automatically, but you already removed that). Please test this locally or let me know if you don't know what I mean :)

frontend/javascripts/oxalis/view/jobs/train_ai_model.tsx
MichaelBuessemeyer: ups, forgot to remove this. I already implemented it this way.

frontend/javascripts/oxalis/view/jobs/train_ai_model.tsx
MichaelBuessemeyer: thanks, sorry I sometimes tend to skip reading comments :see_no_evil:

frontend/javascripts/oxalis/view/jobs/train_ai_model.tsx
MichaelBuessemeyer: Yeah, I wanted to throw an error here in case somehow no tracing exists for an annotation (which should never be the case). By throwing an error I was hoping to show the user a useful error message. But as this should not happen per se, I did not make an effort to show a toast and log an error and so on and simply wanted to rely on the default error catching mechanism of wk.

frontend/javascripts/oxalis/view/jobs/train_ai_model.tsx
MichaelBuessemeyer: This should be done now :)

frontend/javascripts/oxalis/view/jobs/train_ai_model.tsx
daniel-wer: Yes, but I meant that this is missing the actual `throw`, doesn't it? :)

```suggestion
            throw new Error(`Annotation ${annotation.id} has neither a volume nor a skeleton layer`);
```

frontend/javascripts/oxalis/view/jobs/train_ai_model.tsx
```diff
@@ -1,63 +1,194 @@
...
 import { V3 } from "libs/mjs";
+import type { APIAnnotation, APIDataset, ServerVolumeTracing } from "types/api_flow_types";
+import type { Vector3 } from "oxalis/constants";
+import { serverVolumeToClientVolumeTracing } from "oxalis/model/reducers/volumetracing_reducer";
+import { convertUserBoundingBoxesFromServerToFrontend } from "oxalis/model/reducers/reducer_helpers";
 
 const { TextArea } = Input;
 const FormItem = Form.Item;
 
+export type AnnotationInfoForAIJob<GenericAnnotation> = {
```
MichaelBuessemeyer, line 46: Please note that I also added this comment. It hopefully explains why all these additional properties (`volumeTracings, userBoundingBoxes, volumeTracingResolutions`) are needed. Please tell me your opinion on whether you understand what I want to express and maybe on wording improvements :pray:

MichaelBuessemeyer:
> What I wanted to express is that the layer options offered to the user are different ones, depending on where the modal is opened. If you look at the screenshots, one time there are two options offered, and the other time there are three options, although I specified the same annotation. This should not be the case.


Oh good finding :tada: 

I think I fixed it now by simply filtering out volume data layers of the dataset that work as a fallback layer for one of the volume tracing layers.

The rest of your awesome feedback :foot: should also be applied. :dancers:

frontend/javascripts/oxalis/view/jobs/train_ai_model.tsx
```diff
@@ -43,6 +43,13 @@ import { convertUserBoundingBoxesFromServerToFrontend } from "oxalis/model/reduc
 const { TextArea } = Input;
 const FormItem = Form.Item;
 
+// This type is used for GenericAnnotation = HybridTracing | APIAnnotation as in case of multi annotation training,
+// only the APIAnnotations of the given annotations to train on are loaded from the backend.
+// Thus, the code needs to handle both HybridTracing | APIAnnotation where APIAnnotation is missing some information.
+// Therefore, volumeTracings with the matching volumeTracingResolutions are needed to get more details on each volume annotation layer and its resolutions.
+// The userBoundingBoxes are needed for checking for equal bounding box sizes. As training on fallback data is supported and an annotation is not required to have VolumeTracings,
+// it is necessary to save userBoundingBoxes separately and not load them from volumeTracings entries to support skeleton only annotations.
+// Note that a copy the userBoundingBoxes is included in each volume and skeleton tracing of an annotation. Thus, it doesn't matter from which the userBoundingBoxes are taken.
```
daniel-wer, line 52:
```suggestion
// Note that a copy of the userBoundingBoxes is included in each volume and skeleton tracing of an annotation. Thus, it doesn't matter from which the userBoundingBoxes are taken.
```

daniel-wer (approved):
Awesome, thank you for polishing this. I'm looking forward to using this :)

daniel-wer:
As a note, I've created https://github.com/scalableminds/webknossos/issues/8097 for two smaller followup issues that came to my mind, but I didn't want to further bloat and delay this PR :) After merging this, we should test it in production, extend the mentioned issue with any other shortcomings we notice, and then implement the followups :rocket:

PR #8077, opened 2024-09-13
Fix create-mesh-file button

philippotto:
- https://___.webknossos.xyz

### Steps to test:
- open a dataset or annotation
- in the segments view, click on the plus button -> a popover should open

image

### Issues:
- fixes a regression from #7958 

------


hotzenklotz (approved):
LGTM

PR #8084, opened 2024-09-18
When reloading layer, also close its meshfiles

### LINKED ISSUE #8074: “Reload layer” should also clear meshfile handle cache
The hdf5 file handles from the MeshFileService should also be cleared in the `/triggers/reload` route.

fm3:
(Analogous to the agglomerate files)

- https://reloadclosemeshfiles.webknossos.xyz

### Steps to test:
- Open a dataset with meshfile, load precomputed mesh, hit reload layer button
- backend logging should show that a meshfile was closed.

### Issues:
- fixes #8074

------


MichaelBuessemeyer (commented):
Hmmm :thinking: 

to me the code looks fine, and the buckets are reloaded (not the meshes):
image

But the logging shows no closed file handles :thinking:
```
backend: 2024-09-19 09:14:30,420 [info] com.scalableminds.webknossos.datastore.controllers.DataSourceController - Reloading layer '88d2db48-877d-478d-98c4-4b6803112dc4' of dataset sample_organization/l4_sample: closed 0 agglomerate file handles and 0 mesh file handles, removed 0 bucketProviders, 0 vault cache entries and 0 image chunk cache entries.
```

Do you have an idea @fm3?

fm3:
I think you are in an annotation. Apparently, the frontend still calls the _datastore_ route but with the tracing id as layer name. It should, if at all, call the route with the fallback layer name.

However, I’d say that is a different issue. I wrote https://github.com/scalableminds/webknossos/issues/8087 to track that.

To test this PR, I’d recommend using the dataset view mode rather than an annotation.

MichaelBuessemeyer:
> To test this PR, I’d recommend using the dataset view mode rather than an annotation.

Thanks for the help, now it looks good :tada: 
```
backend: 2024-09-19 13:18:28,064 [info] com.scalableminds.webknossos.datastore.controllers.DataSourceController - Reloading layer 'segmentation' of dataset sample_organization/l4_sample: closed 0 agglomerate file handles and 3 mesh file handles, removed 1 bucketProviders, 5 vault cache entries and 824 image chunk cache entries.
```

MichaelBuessemeyer (approved):

PR #8085, opened 2024-09-18
Remove superfluous _type field when writing zarr3 codec jsons

### LINKED ISSUE #7968: Remove superfluous `_type` field from zarr codec jsons
Now that we are starting with zarr3 streaming, we should really get rid of the auto-generated `_type` field in the json-serialized CodecConfiguration classes.

If we can’t get the json macros to skip them, we’ll have to postprocess the json.

image

fm3:
Sometimes play-json adds a "_type" field to the json-serialized case classes, when it thinks they can’t be distinguished otherwise. We need to remove it manually.

This PR does this for zarr3 codec jsons.

- https://codecjsonnotypefield.webknossos.xyz

### Steps to test:
- Visit http://localhost:9000/data/zarr3_experimental/sample_organization/l4_sample/color/1/zarr.json?token=secretSampleUserToken
- Should not include `_type` fields.

### Issues:
- fixes #7968

------


util/src/main/scala/com/scalableminds/util/tools/JsonHelper.scala
```diff
@@ -119,4 +119,21 @@ object JsonHelper extends BoxImplicits with LazyLogging {
       case _ =>
         None
     }
+
+  // Sometimes play-json adds a "_type" field to the json-serialized case classes,
+  // when it thinks they can’t be distinguished otherwise. We need to remove it manually.
+  def removeGeneratedTypeFieldFromJson(jsValue: JsValue): JsValue =
+    removeKey(jsValue, "_type")
+
+  private def removeKey(jsValue: JsValue, keyToRemove: String): JsValue =
```
MichaelBuessemeyer, line 128: the name does not suggest a recurive / deep removal of the key. I'd suggest to make this clear in the naming. e.g. `removeKeyRecursivly` or `removeKeyDeep` or something similar.

MichaelBuessemeyer (approved):
Awesome, works perfectly :tada: 

I only found a potential minor naming improvement. Besides that the code LGTM :)

PR #8091, opened 2024-09-23
Disable proofreading actions for mapped meshes

### LINKED ISSUE #7868: Triggering proofreading actions on precomputed meshes calculated for a mapping should be disabled
The user manually loaded some precomputed meshes, and then selected »merge with active« in the 3d viewport.

This should not be allowed for precomputed meshes that stem for a meshfile that was not computed for the unmapped data.

dieknolle3333:
- https://___.webknossos.xyz

### Steps to test:
(see this comment too: https://github.com/scalableminds/webknossos/issues/7868#issuecomment-2361782931)
- Select a meshfile that was created for a mapping in proofreading mode (e.g. [this one](https://www.notion.so/scalableminds/Test-Datasets-c0563be9c4a4499dae4e16d9b2497cfb?pvs=4#10ab51644c63805faa63c2bd1dec22c1) ) 
- Load some meshes
- Open the context menu in the 3D vieport on a mesh
- If the option "Merge with active segment" is disabled with the tooltip "The mesh wasn't loaded in proofreading mode. Please reload the mesh.", you may need to refresh the meshes
- Afterwards, if you open the context menu for a mesh in the 3D viewport, the three entries "Merge with active segment", "Split from active segment" and "Split from all neighboring segments" should be disabled with a tooltip that explains that the current mesh file was created for a mapping, but one for the oversegmentation is needed. 

### Issues:
- fixes #7868 

------

### DIFF
--- a/frontend/javascripts/oxalis/view/context_menu.tsx
+++ b/frontend/javascripts/oxalis/view/context_menu.tsx
@@ -394,6 +394,7 @@ function getMeshItems(
   maybeUnmappedSegmentId: number | null | undefined,
   visibleSegmentationLayer: APIDataLayer | null | undefined,
   voxelSizeFactor: Vector3,
+  meshFileMapping: string | null | undefined,
 ): MenuItemType[] {
   if (
     clickedMeshId == null ||
@@ -422,7 +423,8 @@ function getMeshItems(
             !isProofreadingActive ||
             activeSegmentMissing ||
             clickedMeshId === activeCellId ||
-            maybeUnmappedSegmentId == null,
+            maybeUnmappedSegmentId == null ||
+            meshFileMapping != null,
           onClick: () => {
             if (maybeUnmappedSegmentId == null) {
               // Should not happen due to the disabled property.
@@ -437,9 +439,11 @@ function getMeshItems(
                   ? "Cannot merge because the proofreading tool is not active."
                   : maybeUnmappedSegmentId == null
                     ? "The mesh wasn't loaded in proofreading mode. Please reload the mesh."
-                    : activeSegmentMissing
-                      ? "Select a segment first."
-                      : null
+                    : meshFileMapping != null
+                      ? "This mesh was created for a mapping. Please use a meshfile that is based on an oversegmentation."
+                      : activeSegmentMissing
+                        ? "Select a segment first."
+                        : null
               }
             >
               Merge with active segment
@@ -454,7 +458,8 @@ function getMeshItems(
             clickedMeshId !== activeCellId ||
             maybeUnmappedSegmentId == null ||
             activeUnmappedSegmentId == null ||
-            maybeUnmappedSegmentId === activeUnmappedSegmentId,
+            maybeUnmappedSegmentId === activeUnmappedSegmentId ||
+            meshFileMapping != null,
           onClick: () => {
             if (maybeUnmappedSegmentId == null) {
               // Should not happen due to the disabled property.
@@ -471,9 +476,11 @@ function getMeshItems(
                   ? "Cannot split because the proofreading tool is not active."
                   : maybeUnmappedSegmentId == null
                     ? "The mesh wasn't loaded in proofreading mode. Please reload the mesh."
-                    : activeSegmentMissing
-                      ? "Select a segment first."
-                      : null
+                    : meshFileMapping != null
+                      ? "This mesh was created for a mapping. Please use a meshfile that is based on an oversegmentation."
+                      : activeSegmentMissing
+                        ? "Select a segment first."
+                        : null
               }
             >
               Split from active segment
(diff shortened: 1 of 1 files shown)


frontend/javascripts/oxalis/view/context_menu.tsx
```diff
@@ -422,7 +423,8 @@ function getMeshItems(
             !isProofreadingActive ||
             activeSegmentMissing ||
             clickedMeshId === activeCellId ||
-            maybeUnmappedSegmentId == null,
+            maybeUnmappedSegmentId == null ||
+            meshFileMapping != null,
```
MichaelBuessemeyer, line 427: :thinking: There is quite some code duplication here in the boolean clause. While you are already editing this file, could you maybe please refactor the boolean expression together with the one in ll 456 ff.? E.g. call it `const proofreadingActionsDisabled = <shared_boolean_expression>` or so :thinking:

frontend/javascripts/oxalis/view/context_menu.tsx
```diff
@@ -437,9 +439,11 @@ function getMeshItems(
                   ? "Cannot merge because the proofreading tool is not active."
                   : maybeUnmappedSegmentId == null
                     ? "The mesh wasn't loaded in proofreading mode. Please reload the mesh."
-                    : activeSegmentMissing
-                      ? "Select a segment first."
-                      : null
+                    : meshFileMapping != null
+                      ? "This mesh was created for a mapping. Please use a meshfile that is based on an oversegmentation."
```
MichaelBuessemeyer, line 443: Maybe this is a more explicit / easier to understand message? In explicitly includes the word unmapped which might help users to understand that the meshfile should be computed on the "raw oversegmented" prediction data?
```suggestion
                      ? "This mesh was created for a mapping. Please use a meshfile that is based on an unmapped oversegmentation data."
```

@fm3 Maybe you have a better phrasing for this? Or do you think @dieknolle3333 message is already enough and mine it so detailed?

frontend/javascripts/oxalis/view/context_menu.tsx
```diff
@@ -394,6 +394,7 @@ function getMeshItems(
   maybeUnmappedSegmentId: number | null | undefined,
   visibleSegmentationLayer: APIDataLayer | null | undefined,
   voxelSizeFactor: Vector3,
+  meshFileMapping: string | null | undefined,
```
MichaelBuessemeyer, line 397: IMO `meshFileMappingName` would be more accurate here as a name as the parameter value itself is the mapping name: `currentMeshFile?.mappingName`

frontend/javascripts/oxalis/view/context_menu.tsx
MichaelBuessemeyer: And while I am at it: The conditional error messages are also pretty much duplicated (ll 439 ff. and ll 475 ff.) Could you maybe extract shared things here too? That would be very awesome :t-rex:

MichaelBuessemeyer (changes requested):
Nice, testing went well :tada: 

I only found a very small improvement of your changes. But while looking at the code I found some duplications. It would be very nice if you could refactor some of the duplication away. For more details please see my comments below.

Besides that, this pr should be ready to be :shipit:

frontend/javascripts/oxalis/view/context_menu.tsx
dieknolle3333: (I know I am not florian but I feel like this is a good idea so I added the more detailed phrase :) )

fm3: sounds good :)

frontend/javascripts/oxalis/view/context_menu.tsx
```diff
@@ -414,15 +415,29 @@ function getMeshItems(
...
+    return !isProofreadingActive
+      ? `Cannot ${actionVerb} because the proofreading tool is not active.`
+      : maybeUnmappedSegmentId == null
+        ? "The mesh wasn't loaded in proofreading mode. Please reload the mesh."
+        : meshFileMappingName != null
+          ? "This mesh was created for a mapping. Please use a meshfile that is based on unmapped oversegmentation data."
+          : actionNeedsActiveSegment && activeSegmentMissing
+            ? "Select a segment first."
+            : null;
+  };
```
MichaelBuessemeyer, line 428: Très bien! :tada:

frontend/javascripts/oxalis/view/context_menu.tsx
```diff
@@ -414,15 +415,29 @@ function getMeshItems(
...
+
+  const shouldAgglomerateSkeletonActionsBeDisabled =
+    !isProofreadingActive ||
+    activeSegmentMissing ||
+    maybeUnmappedSegmentId == null ||
+    meshFileMappingName != null;
```
MichaelBuessemeyer, line 434: :dancers:

MichaelBuessemeyer (approved):
Thanks for incorporating my feedback. 
Re testing still worked so lets :shipit:

PR #8095, opened 2024-09-24
Slider improvements

### LINKED ISSUE #7266: Slider UX improvements
The Antd sliders webknossos uses in many places offer rudimentary slider functionality. I am used to adobe lightroom’s sliders, which have some additional functionality, which are very helpful.

I would guess that wrapping the antd sliders to allow some of those would not be super difficult (but I may be entirely wrong)

 1. double click the slider knob to reset the value to the default
 2. mouse wheel on hover should change the slider value
 3. clicking on the slider (but not on the knob) already sets the value to that position, but if you keep the mouse pressed and move it, the knob no longer moves. It would be nice if you could move the slider in this way.

dieknolle3333:
- https://sliderimprovements.webknossos.xyz

### Steps to test:
- Almost every slider in webknossos should react to wheel actions, meaning that wheeling while hovering above it should change the value, if the slider is not disabled. Also, double-clicking the knob (or one of the knobs for range sliders) should reset it to its default value.
- If possible, test with mouse and touchpad to ensure both interactions feel alright (e.g. step/"speed" of the slider)
- This is true for range sliders and sliders with a single value
- A kind of slider for which it doesnt work are MagSliders
- A good place to start testing is the layer settings tab on the left

### TODOs:
- [x] test all sliders
     - (done except node radius)
- [x] pass through step size
- [x] for range: make range smaller or bigger
- [x] register click on knob?
- [x] make default values/initial values accessible for sliders
### Special candidates
- [x] MagSlider in download_modal_view doesnt wheel
- [x] Slider in quick_select_settings dont wheel well 
- [x] flight options: Mouse Rotation, Keyboard Rotation, Crosshair Size
- [x] layer settings: node radius not editable, how to do that?; particle size onWheel

### Issues:
- fixes #7266

------
- [ ] Updated [migration guide](../blob/master/MIGRATIONSwhom.unreleased.md#unreleased) if applicable
- [ ] Considered [common edge cases](../blob/master/.github/common_edge_cases.md) dont think these are applicable here because onChange method is not altered, and only click and wheel handling are altered


frontend/javascripts/oxalis/view/components/setting_input_views.tsx
```diff
@@ -75,6 +73,8 @@ export class NumberSliderSetting extends React.PureComponent<NumberSliderSetting
   isValueValid = (_value: number | null) =>
     _.isNumber(_value) && _value >= this.props.min && _value <= this.props.max;
 
+  range = this.props.max - this.props.min;
```
dieknolle3333, line 76: remove me

frontend/javascripts/dashboard/advanced_dataset/create_explorative_modal.tsx
```diff
@@ -89,6 +90,7 @@ export function RestrictResolutionSlider({
   resolutionIndices,
   setResolutionIndices,
 }: RestrictResolutionSliderProps) {
+  console.log("restrictresolutionslider");
```
dieknolle3333, line 93: remove me

frontend/javascripts/oxalis/view/action-bar/quick_select_settings.tsx
```diff
@@ -180,6 +181,7 @@ export function HeuristicQuickSelectControls() {
         step={0.25} // a granular step is important so that all 256 values can be effectively targeted
         onChange={onChangeThreshold}
         disabled={!isQuickSelectActive}
+        defaultValue={defaultState.userConfiguration.quickSelect.threshold}
```
dieknolle3333, line 184: /256 * 100?

frontend/javascripts/components/slider.tsx
```diff
@@ -0,0 +1,100 @@
...
+  } else if (range === true || typeof range === "object") {
+    if (!disableOnWheel) {
+      handleWheelEvent = (event) => {
+        const newMin = Math.round(
+          value[0] + getDiffPerSliderStep(event.deltaY, sliderRange, wheelFactor),
+        );
+        const newMax = Math.round(
+          value[1] - getDiffPerSliderStep(event.deltaY, sliderRange, wheelFactor),
+        );
+        if (newMin >= min && newMax <= max && min < max) onChange([newMin, newMax]);
```
MichaelBuessemeyer, line 83: I'd say this code behaves unexpected the following scenario:
If the range is `[0,100] ` and min is currently at 2 and max at 50. How I scroll to increase the range which would result in `newMin = -2` and `newMax = 54` (assuming `getDiffPerSliderStep(event.deltaY, sliderRange, wheelFactor)` returns 4).
But the ` if (newMin >= min && newMax <= max && min < max)` condition would ignore this update.

I would expect that in this scenario the range is updated to `min = 0, max = 54` -> clamping to the min and max boards.

Therefore I would suggest something more like:

```suggestion
        let newMin = Math.round(
          value[0] + getDiffPerSliderStep(event.deltaY, sliderRange, wheelFactor),
        );
        let newMax = Math.round(
          value[1] - getDiffPerSliderStep(event.deltaY, sliderRange, wheelFactor),
        );
        newMin = clamp(min, newMin, Math.min(newMax, max));
        newMax = clamp(newMin, newMax, max);
        onChange([newMin, newMax]);
```

frontend/javascripts/components/slider.tsx
```diff
@@ -0,0 +1,100 @@
...
+
+const getDiffPerSliderStep = (
+  sliderRange: number,
+  factor: number | undefined = DEFAULT_WHEEL_FACTOR,
+  step: number | undefined | null,
+) => {
+  let stepNotNull = step || DEFAULT_STEP;
+  let result = (factor || DEFAULT_WHEEL_FACTOR) * sliderRange;
+  if (result < stepNotNull) return stepNotNull;
+  return result;
```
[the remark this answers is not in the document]
dieknolle3333: Good point. I also tried that. The problem I see with the default parameter is that `step` (coming from the slider props) can be initialized as  `undefined`. And in that case, the default param would not be set, leading to a typing error when comparing `result < stepNotNull`.
You are right for the wheelFactor though!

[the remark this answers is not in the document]
dieknolle3333: `step` is a property from a slider that is passed through, thats why I named it that way. Your are right that the returned number of that function cant be smaller than `step`, but its not really a `minStep` because its not the minimum slider step. Does that make sense?

frontend/javascripts/components/slider.tsx
```diff
@@ -0,0 +1,100 @@
...
+const getWheelStepFromEvent = (
+  step: number | undefined | null,
+  deltaY: number,
+  wheelStep: number,
+) => {
+  // Make sure that result is a multiple of step
+  return (
+    (step || DEFAULT_STEP) *
+    Math.round((wheelStep * deltaY) / Math.abs(deltaY) / (step || DEFAULT_STEP))
+  );
```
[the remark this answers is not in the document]
dieknolle3333: from what I can tell, this doesnt work :/ see my point above

frontend/javascripts/components/slider.tsx
```diff
@@ -0,0 +1,100 @@
...
+  let handleDoubleClick: React.MouseEventHandler<HTMLDivElement> = () => {};
+  const wheelStep = getDiffPerSliderStep(sliderRange, wheelFactor, step);
+  // diffentiate between single value and range slider
+  if (range === false || range == null) {
+    if (!disableOnWheel) {
+      handleWheelEvent = (event) => {
+        const newValue = value - getWheelStepFromEvent(step, event.deltaY, wheelStep);
+        if (newValue < min) onChange(min);
+        else if (newValue > max) onChange(max);
+        else onChange(newValue);
```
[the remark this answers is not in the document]
dieknolle3333: nice!

frontend/javascripts/components/slider.tsx
```diff
@@ -0,0 +1,100 @@
...
+    }
+    // This code is duplicated because TypeScript doesn't understand that onChange
+    // always takes the type of defaultValue.
+    handleDoubleClick = (event) => {
+      if (event.target instanceof HTMLElement) {
+        if (event.target.className.includes("ant-slider-handle")) {
+          if (defaultValue != null) onChange(defaultValue);
+        }
+      }
+    };
```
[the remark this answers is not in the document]
dieknolle3333: I just very much wanted to avoid the ts-ignore, but if you think its okay then I will use it :)

frontend/javascripts/components/slider.tsx
```diff
@@ -0,0 +1,100 @@
...
+export function Slider(props: SliderProps) {
+  const {
+    min,
+    max,
+    onChange,
+    value,
+    range,
+    defaultValue,
+    wheelFactor,
+    disableOnWheel,
```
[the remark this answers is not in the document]
dieknolle3333: wasnt a type, but you're right! maybe `onWheelDisabled` is even more clear?

frontend/javascripts/components/slider.tsx
```diff
@@ -0,0 +1,94 @@
...
+        onChange(clampedNewValue);
+      };
+    }
+  } else if (range === true || typeof range === "object") {
+    if (!disableOnWheel) {
+      handleWheelEvent = (event) => {
+        const diff = getDiffPerSliderStep(event.deltaY, sliderRange, wheelFactor);
+        const newMin = Math.round(value[0] + diff);
+        const newMax = Math.round(value[1] - diff);
+        console.log("newMin", newMin, "newMax", newMax, diff);
```
dieknolle3333, line 83: remove me

frontend/javascripts/components/slider.tsx
dieknolle3333: I feel like this is more a design decision than the cause for the error :) The error was caused by me calling the wrong method :sweat_smile:

dieknolle3333: anyway I think this makes total sense and I adopted the code (essentially) and the behavior!

dieknolle3333: ... and hopefully fixed the weird behaviour :) thank you for the thorough testing!

dieknolle3333:
thank you for the first round of review! :pray: I commented above and incorporated a lot of the suggestions. I also seemingly fixed the histogram slider. Looking forward to the next round, but no pressure obviously! @MichaelBuessemeyer

MichaelBuessemeyer (approved):
Looks good now :tada: :dancers:. Thanks for the awesome slider improvements :fire: 

Before merging please consider my `ensuredStep` suggestion :)

frontend/javascripts/components/slider.tsx
MichaelBuessemeyer: jup its gone, cannot reproduce it anymore :rocket:

frontend/javascripts/components/slider.tsx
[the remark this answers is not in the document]
dieknolle3333: sounds like a very good idea!

PR #8096, opened 2024-09-24
bump sbt plugin versions+protobuf, run codespell

fm3:
Just some housekeeping.
CI should be enough


conf/evolutions/120-remove-old-organization-id.sql
```diff
@@ -144,7 +144,7 @@ ALTER TABLE webknossos.organizations DROP CONSTRAINT organizations_name_key;
 
 
 
--- Readd Foreign Key Constraints
+-- Rea-dd Foreign Key Constraints
```
daniel-wer, line 147:
```suggestion
-- Re-add Foreign Key Constraints
```

daniel-wer (approved):
:+1:

PR #8098, opened 2024-09-25
Fix dataset upload for absolute baseFolder paths

fm3:
The commonPrefix method, used by the finishUpload call, dropped the starting `/` of absolute paths, thus returning invalid paths.

Since the baseFolder was changed to absolute, this broke uploads of non-zipped datasets.

### Steps to test:
- in application.conf set datastore.baseFolder to an absolute path
- upload a non-zipped dataset via the UI
- should work

### Issues:
- fixes https://scm.slack.com/archives/C02H5T8Q08P/p1727248592960659

------

### DIFF
--- a/CHANGELOG.unreleased.md
+++ b/CHANGELOG.unreleased.md
@@ -16,6 +16,7 @@ For upgrade instructions, please check the [migration guide](MIGRATIONS.released
 ### Changed
 
 ### Fixed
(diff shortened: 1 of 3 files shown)


hotzenklotz:
Can you add a test for this to avoid these breaking changes in the future?

fm3:
Not sure what should be tested there exactly. I think a unit test for the commonPrefix method wouldn’t add much now, since we basically never touch that.

We can try to add an e2e test for dataset upload, but here, too, I’m not sure if we can cover all possible configurations without needing huge amounts of test time. I created an issue for that https://github.com/scalableminds/webknossos/issues/8099

Issue #7683, opened 2024-03-13
Sort imports in typescript files


philippotto:
See https://github.com/scalableminds/webknossos/pull/7655#issuecomment-1993851177.

PR #8102, opened 2024-09-25
Remove unused imports while linting

hotzenklotz:
For some reasons the linting rule for unused impprts was not active any more. Perhaps it was excluded from the default rule set by the `biome` team.
I enabled it again and ran the automatic fixing on all frontend files. 

For the most part it removed a ton of `import React` statements. With React 18 there is no need to import react in every file. Further the Typescript compiler will detect and transform JSX code automatically based on the `*.tsx` file exension. If we really want to keep those react import statemetnts there is an option for that: https://biomejs.dev/reference/configuration/#javascriptjsxruntime.

We may also want to enabled automatic sorting of imports at some point. See issue #7683

### Steps to test:
- Running all CI tests should be enough

### Issues:
- fixes https://scm.slack.com/archives/C5AKLAV0B/p1727085281308479

------


MichaelBuessemeyer (approved):
> If we really want to keep those react import statemetnts

I don't really seen an advantage in this. 

Code lgtm and works :tada:

Issue #7293, opened 2023-08-21
Decouple DataLayer information from JSON representation


frcroth:
Currently, datalayers are parsed and written via JSON macro, which is simple, but does not allow for great flexibility. E.g. when adding new properties, these have to options, so that existing datasource-properties.json can still be read.

Therfore, the JSON formatting should be done manually (cmp. https://stackoverflow.com/questions/45496635/play-framework-json-reads-from-empty-string-to-empty-list)

frcroth:
Related to #8384 if datasource-properties.json in the current form will be deprecated.

PR #8111, opened 2024-10-05
Rename Resolution to Mag

### LINKED ISSUE #7378: Rename Resolution to Mag (in UI and code where possible)
## Detailed Description
In code and UI, we should switch to the term Mag everywhere for consistency reason. Quoting @fm3, there should be strategic tooltips which explain the concept.

## Context
https://scm.slack.com/archives/C5AKLAV0B/p1695737597207699

dieknolle3333:
In this PR, the term "resolution" is replaced with "mag"/"magnification" in a lot of places.
- For the backend, see [this comment below](https://github.com/scalableminds/webknossos/pull/8111#issuecomment-2399313689). The DB schema was adjusted, too.
- In the frontend code, method names, their parameters, types, properties, error messages and console prints were adjusted. Function or local scoped variables were not adjusted, because this would make the changes way too large to review. As I am not familiar with the shader code, I didnt change `*.glsl.ts` files either. 
- In the UI, the term "resolution" is only being used in the context of animations after this PR. Other than that, "magnification" or sometimes "mag" is (or should be) used in toasts, tooltips, modals, sliders etc.

- https://magalltheway.webknossos.xyz/

### Steps to test:
- Test that everythings works as before by trying to test features that depend on mags. 
- This includes:
    - Opening an annotation, zooming in and out
    - Creating an annotation with restricted mags
    - Creating a volume layer with restricted mags
    - Testing all tools in the toolbar 
    - Create & Edit task type with magRestrictions, create task + open it, should only have the allowed mags
    - reload dataset list from disk

### TODOs:
- [x] backend
- [x] frontend
- [x] rename files, too
- [x] fix frontend tests
- [x] fix e2e tests
- [x] add some explanation to info tab tooltip

backup branch: https://github.com/scalableminds/webknossos/tree/backup-rename-resolution-to-mag

### Issues:
- contributes to #7378

------


fm3:
@dieknolle3333 I pushed a commit changing a lot of usages in the backend. This changes the protocol in a couple of ways (please adapt the frontend):

 - resolutionRestrictions → magRestrictions (everywhere, e.g. tasktype json, AnnotationLayerParameters. just search for all occurrences)
 - volumeTracing proto → resolutions → mags
 - skeletonTracing proto → node.resolution → node.mag

There are also a couple of spots where a change might be expected but cannot be easily done without breaking backwards compatibility with hard-to-migrate data. I did *not* change:

 - CreateNodeSkeletonUpdateAction.resolution
 - DataLayer.resolutions (this is a big one. we need to get rid of the datasource-properties.json eventually)
 - WKWDataLayer.wkwResolutions

I also did not yet change the postgres tables. I might add this later if I manage to find time. That should not have impacts on the frontend-backend protocol though. For me to test that I don’t break anything there, it would be nice if everything else worked again first.

Please ping me again if anything seems out of place. I might have missed some spots of course.

dieknolle3333:
thank you so much! :pray: I will get back to you once I am done with the frontend or if I have a question about the backend part.

dieknolle3333:
after trying to rename "resolution" in a lot of places, I came to the conclusion that I will have to narrow the scope. I will rename the most important types and methods, but not to it everywhere. This also means that I will probably rebase.

dieknolle3333:
@fm3 from my side, the first iteration for the frontend is done. I renamed most methods and most parameters and added some sentences to the UI.

fm3:
Did that now :) From a first glance, it looks like things still work fine :raised_hands: 

Could you elaborate a bit on how you narrowed the scope? What areas/things did you skip? I guess this is related to the things I didn’t want to change in the backend [as listed above](https://github.com/scalableminds/webknossos/pull/8111#issuecomment-2399313689)?

fm3:
I fixed the snapshot tests, added a changelog entry and slightly updated the PR description. Once you described the scope as I mentioned above, this should be ready for a first review round :)

dieknolle3333:
Nice! Thanks for taking care of the PR description and the changelog :raised_hands: otherwise I would have done that after getting back to work from my lecture :) Anyway, I added some details about the scope to the PR description, so I hope this makes the scope more clear.

fm3:
Nice, thanks! One more thing: I’d suggest to undo the changes in the docs. I think the usages of “resolution” there made sense in their respective context. Only exception: the “(also called mags, mipmaps […]” in `data_formats.md` is a good idea.

docs/neuroglancer_precomputed.md
```diff
@@ -28,7 +28,7 @@ WEBKNOSSOS expects the following file structure for Neuroglancer Precomputed dat
 ```
 my_dataset.precomputed             # One root folder per dataset
 ├─ info                            # Dataset [metadata in JSON format](https://github.com/google/neuroglancer/blob/master/src/neuroglancer/datasource/precomputed/volume.md#info-json-file-specification)
-├─ scale_1                         # One subdirectory with the same name as each scale/magnification "key" value specified in the info file. Each subdirectory contains a chunked representation of the data for a single resolution.
```
dieknolle3333, line 31: @fm3 I see your point that "resolution" made sense in the other places, so I reverted the changes. But here, I am not sure I understand. I think "resolution" can be used as a synonym for "mag" to explain the concept, and from there on, only "mag(nification)" should be used. The latter seems to be the case here. What am I missing?

fm3: that this is specifically about the “neuroglancer precomputed” file format, so it makes sense to follow their terminology. In fact, it looks like this sentence is a [direct quote from here](https://github.com/google/neuroglancer/blob/master/src/datasource/precomputed/volume.md). However, maybe both terms can be used to bridge the gap to webknossos. How about something like “contains a chunked representation of the data for a single resolution (meaning single magnification/mag for WEBKNOSSOS)”?

dieknolle3333: Oh, I see. Yes, good idea!

MichaelBuessemeyer:
Thanks for this pr you two :)

> I came to the conclusion that I will have to narrow the scope. I will rename the most important types and methods, but not do it everywhere

Sounds to me like not every occurrence that can be renamed is renamed in this pr (even cases that wouldn't break backwards compatibility). Do I understand this correctly? (Haven't looks at the code, will do now)

In that case what about keeping the original issue open and adjusting its description naming that some places are still missing and an open TODO?

frontend/javascripts/messages.tsx
```diff
@@ -368,9 +368,11 @@ instead. Only enable this option if you understand its effect. All layers will n
   "dataset.is_scratch":
     "This dataset location is marked as 'scratch' and meant for testing only. Please move this dataset to a permanent storage location and reimport it.",
   "dataset.resolution_mismatch":
```
MichaelBuessemeyer, line 370: Maybe also rename this one here -> optional

frontend/javascripts/messages.tsx
```diff
@@ -368,9 +368,11 @@ instead. Only enable this option if you understand its effect. All layers will n
...
-    "This dataset contains multiple layers which differ in their magnification. Please convert the layers to make their resolutions match. Otherwise, rendering errors cannot be avoided.",
+    "This dataset contains multiple layers which differ in their magnification. Please convert the layers to make their magnifications match. Otherwise, rendering errors cannot be avoided.",
   "dataset.z1_downsampling_hint":
     "The currently rendered quality is not optimal due to the available magnifications and the viewport arrangement. To improve the quality try to increase the size of the XY viewport (e.g. by maximizing it).",
+  "dataset.mag_explanation":
+    "Layers contain image data in one or multiple magnifications. The image data in full magnification is referred to as the finest mag, e.g. mag 1-1-1. Magnification 4-4-4 describes a downsampling factor of 4 in each dimension.",
```
MichaelBuessemeyer, line 375: I am unsure whether "full" in `full magnification` is the correct word here. :thinking: I'd vote for something like this
> The image data in best scanning magnification is referred to as the finest mag, e.g. mag 1-1-1.

I am open to discuss this :)
Because "best scanning magnification" doesn't sound ideal as well :/

MichaelBuessemeyer: And maybe extend the explaination a little:

```suggestion
    "Layers contain image data in one or multiple magnifications. The image data in full magnification is referred to as the finest mag, e.g. mag 1-1-1. Magnification 4-4-4 describes a downsampling factor of 4 in each dimension compared to magnification 1-1-1.",
```

MichaelBuessemeyer (changes requested):
Thanks a lot for this PR and taking care of more consistent naming.

Please find my remarks below.

During testing I sadly could not find any errors :upside_down_face:

app/models/dataset/Dataset.scala
```diff
@@ -664,28 +664,26 @@ class DatasetDAO @Inject()(sqlClient: SqlClient, datasetLayerDAO: DatasetLayerDA
...
+      rowsParsed <- Fox.combined(rows.map(parseRow)) ?~> "could not parse mag row"
     } yield rowsParsed
 
-  def updateResolutions(datasetId: ObjectId, dataLayersOpt: Option[List[DataLayer]]): Fox[Unit] = {
-    val clearQuery = q"DELETE FROM webknossos.dataset_resolutions WHERE _dataset = $datasetId".asUpdate
+  def updateMags(datasetId: ObjectId, dataLayersOpt: Option[List[DataLayer]]): Fox[Unit] = {
+    val clearQuery = q"DELETE FROM webknossos.dataset_mags WHERE _dataset = $datasetId".asUpdate
     val insertQueries = dataLayersOpt.getOrElse(List.empty).flatMap { layer: DataLayer =>
-      layer.resolutions.map { resolution: Vec3Int =>
+      layer.resolutions.map { mag: Vec3Int =>
```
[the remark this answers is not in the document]
fm3: The HTTP API is only part of the problem here. The bigger deal is the datasource-properties.json files on disk (and the code that creates and reads them, scala, python and matlab). #7293 describes an idea of how to get around this. However, I did not want to include that in this PR since it’s a more complicated change. It’s also made even more complicated because the MagLocators in e.g. ZarrDataLayers use the keyword "mags", so at least there it is already taken (that probably was a mistake…). I don’t know how to solve this elegantly. I hope we can get rid of datasource-properties.jsons eventually, but I don’t really have a plan for that either.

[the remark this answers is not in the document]
MichaelBuessemeyer: > The bigger deal is the datasource-properties.json files on disk

Ah damn it. Sorry, didn't see this 

> However, I did not want to include that in this PR since it’s a more complicated change.

Sure, makes a lot of sense.

dieknolle3333:
>Sounds to me like not every occurrence that can be renamed is renamed in this pr (even cases that wouldn't break backwards compatibility). Do I understand this correctly? (Haven't looks at the code, will do now)

>In that case what about keeping the original issue open and adjusting its description naming that some places are still missing and an open TODO?

I agree that this is a good idea for the shader code. The local variables feel like a wont-do to me, see [this slack conversation](https://scm.slack.com/archives/C5AKLAV0B/p1728997127519639). I am open to discuss this of course.

frontend/javascripts/messages.tsx
fm3: I think for this one it may be fine to use "resolution", since this is where we explain what we mean with the word magnification/mag. Same as in https://docs.webknossos.org/webknossos/terminology.html

frontend/javascripts/oxalis/model_initialization.ts
```diff
@@ -503,10 +503,10 @@ function getMergedDataLayersFromDatasetAndVolumeTracings(
 
     const fallbackLayer = fallbackLayerIndex > -1 ? originalLayers[fallbackLayerIndex] : null;
     const boundingBox = getDatasetBoundingBox(dataset).asServerBoundingBox();
-    const resolutions = tracing.resolutions || [];
+    const resolutions = tracing.mags || [];
     const tracingHasResolutionList = resolutions.length > 0;
-    // Legacy tracings don't have the `tracing.resolutions` property
-    // since they were created before WK started to maintain multiple resolution
+    // Legacy tracings don't have the `tracing.mags` property // TODO_c check
```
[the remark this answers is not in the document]
dieknolle3333: nvm :sweat_smile:

frontend/javascripts/router.tsx
```diff
@@ -647,13 +647,13 @@ class ReactRouter extends React.Component<Props> {
...
                       const { autoFallbackLayer, fallbackLayerName } = getParams;
-                      const resolutionRestrictions: APIResolutionRestrictions = {};
+                      const resolutionRestrictions: APIMagRestrictions = {};
 
                       if (getParams.minRes !== undefined) {
                         resolutionRestrictions.min = Number.parseInt(getParams.minRes);
 
                         if (!_.isNumber(resolutionRestrictions.min)) {
-                          throw new Error("Invalid minRes parameter");
+                          throw new Error("Invalid minRes parameter"); // TODO_c maybe change get params
```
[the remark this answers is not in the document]
dieknolle3333: nvm again

frontend/javascripts/messages.tsx
dieknolle3333: "Full resolution" is used in other places too, such as the terminology doc. Maybe "highest resolution" could be an alternative right here? "Best" could mean coarse too, right? Depending on the zoom, for example.

dieknolle3333: (I already committed the changes here that we agree on)

dieknolle3333: @MichaelBuessemeyer what do you think about "highest resolution"? or "full resolution"?

dieknolle3333 (commented):
@fm3 or @MichaelBuessemeyer: there were merge conflicts in the migrations because the worker name already had no. 121. So I gave the migration in this PR no. 122 and changed it in `MIGRATIONS.unreleased.md`, and renamed the files in conf/evolutions/122-resolution-to-mag.sql and /reversions/. I hope this was correct!
edit: nvm, the pipeline turned red

fm3:
Yes, thanks! I see you found the various pitfalls of schema versions :sweat_smile: I added a commit also adapting the sql reversion (It is not checked in the CI, but could become important if we need to roll back after deploying this)

frontend/javascripts/messages.tsx
MichaelBuessemeyer: I'd say both are fine. You can choose as you want. I might have a very small tendency towards highest, but 
> "Full resolution" is used in other places too, such as the terminology doc.

sounds like a good argument, So let's better keep that :tada:

frontend/javascripts/admin/tasktype/task_type_create_view.tsx
```diff
@@ -48,7 +48,7 @@ type FormValues = {
     mergerMode: boolean;
     preferredMode?: APIAllowedMode;
     allowedModes: APIAllowedMode[];
-    resolutionRestrictions: APIResolutionRestrictions;
+    magRestrictions: APIMagRestrictions;
```
MichaelBuessemeyer, line 51: Maybe this renaming here is the cause for the task type mag restrictions not working on this branch. Maybe the backend still expects the old name `resolutionRestrictions` ? @fm3?

MichaelBuessemeyer (changes requested):
Nice, thanks you two :)

I didn't do another "extensive" testing session as only local vars were changed.

But sadly I noticed that the mag restrictions setting of task types is now buggy on this branch. @dieknolle3333 could you please fix this before merging? I just identified a potential reason for this.

How to test / reproduce:
1.
- create a new task type with mag restrictions (e.g. min 1, max 2)
- edit this task type -> the mag restriction setting should be empty -> broken
2.
- create a task with this task type
- open such a task and try to zoom out of the valid range. On the master a warning toast is shown, but on this branch no toast is shown.

frontend/javascripts/admin/tasktype/task_type_create_view.tsx
fm3: hmm, no, resolutionRestrictions should no longer exist anywhere :thinking:  We’ll have a look.

frontend/javascripts/messages.tsx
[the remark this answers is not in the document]
dieknolle3333: no, they are not used. I also noticed this. I'll remove them

frontend/javascripts/router.tsx
```diff
@@ -647,21 +647,21 @@ class ReactRouter extends React.Component<Props> {
...
                       const { autoFallbackLayer, fallbackLayerName } = getParams;
-                      const resolutionRestrictions: APIResolutionRestrictions = {};
+                      const resolutionRestrictions: APIMagRestrictions = {};
 
                       if (getParams.minRes !== undefined) {
                         resolutionRestrictions.min = Number.parseInt(getParams.minRes);
 
                         if (!_.isNumber(resolutionRestrictions.min)) {
-                          throw new Error("Invalid minRes parameter");
+                          throw new Error("Invalid minMag parameter");
```
dieknolle3333, line 656: TODO: adjust get params too.

frontend/javascripts/admin/tasktype/task_type_create_view.tsx
dieknolle3333: the form item still had old names :(

dieknolle3333: Adapting the names seems to fix it. Editing the task type now shows the restricted mags. Zooming out in a task with the task type with restricted mags yields the error "Annotating data is restricted to a certain zoom range. (...)".

MichaelBuessemeyer (approved):
Awesome thanks :pray: 

The fix works and thanks for removing the unused messages

PR #8112, opened 2024-10-07
Allow admins to view and cancel other users jobs

### LINKED ISSUE #7498: List all jobs of orga to admin
admins should see and be able to cancel all jobs of their organization. A “user” column should be added in the frontend to the job list view.

frcroth:
- https://___.webknossos.xyz

### Steps to test:
- Create a job as a non admin user (Or create a job as an admin user and edit the owner in the DB)
- Be able to see it and cancel it
- Note that the owner is now displayed in the job list

### Notes

-  No changes were required to make canceling possible
- The SQL change is copied from the readAccessQ
- The frontend change (new column) is copied from the Project List

### Issues:
- fixes #7498

------
- [ ] ~~Updated [documentation](../blob/master/docs) if applicable~~ The jobs page was already old before (e.g. screenshot  shows old logo, different UI), but the text is unchanged

### DIFF
--- a/app/models/job/Job.scala
+++ b/app/models/job/Job.scala
@@ -141,7 +141,9 @@ class JobDAO @Inject()(sqlClient: SqlClient)(implicit ec: ExecutionContext)
      """
 
   private def listAccessQ(requestingUserId: ObjectId) =
-    q"""_owner = $requestingUserId"""
+    q"""_owner = $requestingUserId OR
+       ((SELECT u._organization FROM webknossos.users_ u WHERE u._id = _owner) IN (SELECT _organization FROM webknossos.users_ WHERE _id = $requestingUserId AND isAdmin))
+     """
 
   override def findAll(implicit ctx: DBAccessContext): Fox[List[Job]] =
     for {
--- a/CHANGELOG.unreleased.md
+++ b/CHANGELOG.unreleased.md
@@ -16,6 +16,7 @@ For upgrade instructions, please check the [migration guide](MIGRATIONS.released
 
 ### Changed
 - Some mesh-related actions were disabled in proofreading-mode when using meshfiles that were created for a mapping rather than an oversegmentation. [#8091](https://github.com/scalableminds/webknossos/pull/8091)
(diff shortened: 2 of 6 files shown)


fm3 (approved):
LGTM and works for me :) 

Let’s see how the performance will look in production (the lists get longer now, as all users’ jobs will be shown). It may become necessary to transition the list route to a complex query eventually.

app/models/job/Job.scala
```diff
@@ -141,7 +141,9 @@ class JobDAO @Inject()(sqlClient: SqlClient)(implicit ec: ExecutionContext)
      """
 
   private def listAccessQ(requestingUserId: ObjectId) =
-    q"""_owner = $requestingUserId"""
+    q"""_owner = $requestingUserId OR
+       ((SELECT u._organization FROM webknossos.users_ u WHERE u._id = _owner) IN (SELECT _organization FROM webknossos.users_ WHERE _id = $requestingUserId AND isAdmin))
```
MichaelBuessemeyer, line 145: Hey, I know that the review was already done by @fm3 

But to me the added SQL reads kinda strange. How I read it is that you 
> `(SELECT u._organization FROM webknossos.users_ u WHERE u._id = _owner)`

Select the organization id by reading it from the user table from he column where the user is the actual owner of the organization in question. And then 
> `(SELECT _organization FROM webknossos.users_ WHERE _id = $requestingUserId AND isAdmin)`

you check whether the current user (`requestingUserId`) is in the same organization and is an admin? 

To me that sounds like this can be simplyied into a single select statement and not a `((Select ...) in (Select...))` statement which reads much more complex.

I'd say that 
```
(SELECT true FROM webknossos.users_ u WHERE u._id = _owner AND u._id = $requestingUserId AND isAdmin)
```
should evaluate to the same request on a logical level.

But I also might be wrong here ... :thinking:

frcroth: To me it seems like your query seems to check that the requesting user is the owner, which is what should not be needed since the admin should also be able to view it.

MichaelBuessemeyer: Oh ah I see. The first subquery is needed to identify the orga id as only the owner id is know and the second checks whether the user is in the same orga and an admin :+1:

In that case it looks good :) => Nvm my mistake :see_no_evil:

PR #8113, opened 2024-10-07
Add e2e test adding metadata to skeleton tracing

### LINKED ISSUE #8089: e2e tests for segment + tree metadata
Follow-up for  https://github.com/scalableminds/webknossos/pull/7875: include some metadata in annotation e2e tests

frcroth:
Added an e2e test that contains an update action that updates metadata. Some notes:
- This could also be included into the "Send complex update actions and compare resulting tracing" test if we want to keep duplication / number of tests lower
- This does not include tests for adding metadata to volume annotations. It seems there arent any tests for updating volume tracings?

### Issues:
- fixes #8089


fm3 (commented):
Looking good, thanks!
Could you have a look at how long this new test is running? If it is slow, maybe it can be combined with another test that already creates trees? That way I would hope the CI time does not increase.

frcroth:
This new test does not seem to really affect CI time. For me locally, it takes about 0.2 s.
It can definitely be combined with an existing test, if that is desired.

fm3 (approved):
ok for 0.2s I’d say we can leave it as is for now :)

PR #8115, opened 2024-10-07
Add name to worker in database

### LINKED ISSUE #8018: Add name for workers, to show in slack notification
Add a name field in sql for the wk workers, show it in api/jobs/status and worker-lost/return slack notifications

frcroth:
There is no way to view or change this from the frontend, as it was not specified in the issue. Is this intended?
If yes, then this probably does not need to be in the changelog as not to confuse people.

### Issues:
- fixes #8018 

------


fm3 (approved):
Perfect :) Yes, no need to modify this field via API or frontend at this time. Setting it in the DB manually is enough.

PR #8123, opened 2024-10-14
Add button to select all trees and all segments that match a search

### LINKED ISSUE #4626: Select all trees that match a search
Recently, I used the tree search a lot. It would be nice if the search widget would also have a button for selecting all trees that match the query.

dieknolle3333:
- https://selectallmatchingtrees.webknossos.xyz/
skeleton tab: image
segments tab: 
image

### Steps to test:
- Go to segments and skeleton tab and search for something that yields multiple results
- click button "select all matches"
- all matching leaves should be selected and scrolled to
- clicking arrow up or down afterwards should still work
- the selected segments and skeletons should behave appropriately, so e.g. dragging and dropping them and multiple-segment-actions should work as usual
- try to search for a term that yields groups and leaves as a result. only multiple leaves can be selected
- if only groups are the result, the button should be disabled
- parent groups of results should be expanded after hitting the new "select all matches" button

- a little unrelated: if a search only yields one result, you can now use the arrow buttons to select it

### TODOs:
- [x] test for long lists
- [x] test after merging the fix for focussing
- [x] should not work for skeleton groups, segment groups
- [x] segments: mixed results: disable or select only leaves
- [x] focus segments
- [x] focus skeletons
- [x] you can search for skeleton groups but not segment groups? 
- [x] dont disable arrows if there is only one result (otherwise, navigation is only possible by hitting enter)

### Issues:
- fixes #4626 

------

## Release Notes

- **New Features**
	- Added metadata support in annotations for Trees and Segments.
	- Summary row in time tracking overview for aggregated times and annotations/tasks.
	- Enhanced slider functionality with value changes via wheeling and double-click resets.
	- Search for unnamed segments using full default names.
	- Improved loading speed for precomputed meshes.
	- New button in search popover for selecting all matching non-group results.
	- Unified terminology in the UI, replacing "Resolution" with "Magnification".
	- Introduced functionality for selecting all matching segments and trees.

- **Bug Fixes**
	- Resolved issues with dataset uploads, group expansions, incorrect shape returns, sorting errors, and annotation creation for public datasets.
	- Users without edit permissions can no longer delete sharing tokens via the API.

- **Improvements**
	- Admins can now view and cancel all jobs with job ownership displayed.
	- Enhanced segment and group selection logic for a better user experience.
	- Streamlined selection logic in the TreeHierarchyView for improved navigation.
	- Adjusted control flow in the AdvancedSearchPopover for better selection handling.
	- Improved interaction with segment tree through refined selection methods.


frontend/javascripts/oxalis/view/right-border-tabs/segments_tab/segments_view.tsx
```diff
@@ -1606,16 +1606,45 @@ class SegmentsView extends React.Component<Props, State> {
...
+      return match.id;
+    });
+    Store.dispatch(
+      setSelectedSegmentsOrGroupAction(
+        allMatchingSegmentIds,
+        null,
+        this.props.visibleSegmentationLayer.name,
+      ),
+    );
+    this.tree.current?.scrollTo({ key: allMatches[0].key });
```
MichaelBuessemeyer, line 1645: Could you please give this a small timeout like in lines 1614-1616? Sometimes the browser is kinda slow due to some workload resulting in the segment groups not being all expanded and thus `allMatches[0].key` might still not be shown in the list. In such a case the antd lib ignored the scrollTo call.
```suggestion
 setTimeout(() => {
    this.tree.current?.scrollTo({ key: allMatches[0].key });
     }, 50);
```

frontend/javascripts/oxalis/view/right-border-tabs/tree_hierarchy_view.tsx
```diff
@@ -254,11 +253,11 @@ function TreeHierarchyView(props: Props) {
   const checkedKeys = deepFlatFilter(UITreeData, (node) => node.isChecked).map((node) => node.key);
 
   // selectedKeys is mainly used for highlighting, i.e. blueish background color
-  const selectedKeys = props.selectedTreeIds.map((treeId) =>
-    getNodeKey(GroupTypeEnum.TREE, treeId),
-  );
+  let selectedKeys = props.selectedTreeIds.map((treeId) => getNodeKey(GroupTypeEnum.TREE, treeId));
+
+  if (props.activeGroupId) selectedKeys = [getNodeKey(GroupTypeEnum.GROUP, props.activeGroupId)];
```
MichaelBuessemeyer, line 258: Is there a reason why you remove the selected keys here in case there is an active group?

frontend/javascripts/oxalis/view/right-border-tabs/tree_hierarchy_view.tsx
```diff
@@ -254,11 +253,11 @@ function TreeHierarchyView(props: Props) {
...
 
-  if (props.activeGroupId) selectedKeys.push(getNodeKey(GroupTypeEnum.GROUP, props.activeGroupId));
+  treeRef.current?.scrollTo({ key: selectedKeys[0], align: "auto" });
```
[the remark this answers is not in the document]
MichaelBuessemeyer: The current state of the code can be used to trigger the unwanted behavior:

- create a lot of trees
- give two trees at the top of the list a specific name
- search for that name and select it via the select all
- scroll down so that those trees are outside the rendered part of the tree list
- open the context menu on one of the bottom trees. 
=> this will trigger an auto scroll to the top of the list which is unwanted and the context menu no longer hovers the skeleton on which the context menu was initially opened

MichaelBuessemeyer (changes requested):
Thanks for this pr :tada: Makes moving and organizing segments / trees likely much easier for our users :100:  

I noticed a few things / had questions regarding your changes. Please find them below.

Moreover, I kinda stumbled upon the X/Y indicator of the search popover which shows 0/Y even using the new selecting all matches. To me it looks like 0/Y say that there is no active selection although there is. But I also could not come up with a good visual indication that all matches are currently selected. In case you can come up with something, I am glad so see / hear your idea :D

hotzenklotz:
@coderabbitai full review

frontend/javascripts/oxalis/view/right-border-tabs/tree_hierarchy_view.tsx
dieknolle3333: yes! it is because when selecting multiple elements with ctrl, you cannot select skeletons and skeleton groups at the same time. next to that, there is no multi group selection. Its not like there is a distiction between active and selected group like for the segments, right?

frontend/javascripts/oxalis/view/right-border-tabs/tree_hierarchy_view.tsx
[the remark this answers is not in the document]
dieknolle3333: good point!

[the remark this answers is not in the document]
dieknolle3333: I think my changes solved this :)

dieknolle3333:
>To me it looks like 0/Y say that there is no active selection although there is. But I also could not come up with a good visual indication that all matches are currently selected. In case you can come up with something, I am glad so see / hear your idea :D

Hm, very good point! I feel like the easiest would be to display nothing at all? and change the color of the button. I built something that felt good, let me know what you think!
image

frontend/javascripts/oxalis/view/right-border-tabs/tree_hierarchy_view.tsx
dieknolle3333: There is even this modal which says "You have ${numberOfSelectedTrees} selected Trees. Do you really want to select this group? This will deselect all selected trees." 

I thought about adding this here to in this specific case, but it seems hard to realize without very much changing the current architecture:
- the logic that the selected key is a group is within tree_hierarchy_view (for the skeletons tab)
- we would need it in`AdvancedSearchPopover` though, because if the user selects "no" within the modal, we dont want to change the selection. As there is no history to the selection, this needs to happen in the search popover modal.
- the components dont know each other, they are both placed in the `SkeletonView`

Adding to this, the user pressed the arrow to select the next tree / treegroup, so its hopefully not surprising behaviour. So I personally dont see the need to restructure the code for this to work.
What do you think?

MichaelBuessemeyer:
> Hm, very good point! I feel like the easiest would be to display nothing at all? and change the color of the button. I built something that felt good, let me know what you think!

I think that's a good idea :tada:. Maybe this could be  improved a little more by not having an empty x/y but simply replacing the text with "all" or something like that? What do you think? To me the empty space seems kinds "out of place / unused / useless"

image

frontend/javascripts/oxalis/view/right-border-tabs/tree_hierarchy_view.tsx
MichaelBuessemeyer: Oh right, I forgot about this limitation. 
> There is even this modal which says "You have ${numberOfSelectedTrees} selected Trees. Do you really want to select this group? This will deselect all selected trees."

Thanks for the clarification :pray: 
So you kinda fixed the code a little here, correct? It behaves "more correctly" than before. I guess the code before was not broken because when a group was selected the code took care of deselecting all trees? Beside that, in regard to your clarification your changes are much nicer than the previous version :sparkles: 

> So I personally dont see the need to restructure the code for this to work.
What do you think?

Sure, fine by me. If we get the feedback that this should be included in the "search multi selection" then we could still develop this as a follow up :)

frontend/javascripts/oxalis/view/right-border-tabs/advanced_search_popover.tsx
```diff
@@ -7,6 +7,8 @@ import Shortcut from "libs/shortcut_component";
 import DomVisibilityObserver from "oxalis/view/components/dom_visibility_observer";
 import { mod } from "libs/utils";
 
+const PRIMARY_COLOR = "var(--ant-color-primary)";
```
MichaelBuessemeyer, line 10: Uh nice :tada:, this made me come up with the following possible refactoring: https://github.com/scalableminds/webknossos/issues/8150

frontend/javascripts/oxalis/view/right-border-tabs/advanced_search_popover.tsx
```diff
@@ -171,9 +181,16 @@ export default class AdvancedSearchPopover<
...
                         currentPosition: null,
+                        areAllMatchesSelected: false,
                       })
                     }
-                    addonAfter={`${currentPosition + 1}/${numberOfAvailableOptions}`}
+                    addonAfter={
+                      <div style={{ minWidth: 25 }}>
+                        {!areAllMatchesSelected
+                          ? `${currentPosition + 1}/${numberOfAvailableOptions}`
+                          : ""}
```
MichaelBuessemeyer, line 191: I'd add some placeholder here
```suggestion
                          : "all"}
```

frontend/javascripts/oxalis/view/right-border-tabs/advanced_search_popover.tsx
```diff
@@ -171,9 +181,16 @@ export default class AdvancedSearchPopover<
                       this.setState({
                         searchQuery: evt.target.value,
                         currentPosition: null,
+                        areAllMatchesSelected: false,
                       })
                     }
-                    addonAfter={`${currentPosition + 1}/${numberOfAvailableOptions}`}
+                    addonAfter={
+                      <div style={{ minWidth: 25 }}>
```
MichaelBuessemeyer, line 188: In the case of `areAllMatchesSelected` I would color this in the disabled color so that the placeholder is not that prominent

MichaelBuessemeyer (approved):
Nice, looks good. I'd only add a placeholder to the empty field next to the search input when the "all selection" is active (as mentioned above)

frontend/javascripts/oxalis/view/right-border-tabs/advanced_search_popover.tsx
dieknolle3333: I LOVE THE IDEA, THANK YOU :raised_hands:

frontend/javascripts/oxalis/view/right-border-tabs/tree_hierarchy_view.tsx
dieknolle3333: Yes, I think I removed inconsistent behaviour regarding the multiselection via ctrl vs the selection in the search. Previously if you had selected skeletons and selected a group via the search, nothing happened to the skeleton selection. I think this wasnt too bad but it doesnt make much sense either. you cannot do any actions with your heterogenous selection anyway, such as drag and drop or actions via a context menu
image

dieknolle3333:
found bug that if i select a group, the skeletons are not selected

dieknolle3333:
@MichaelBuessemeyer do you want to rereview my lastest changes?

MichaelBuessemeyer (approved):
Awesome, looks good :rocket:.

PR #8126, opened 2024-10-16
Async reading in FileSystemDataVault

### LINKED ISSUE #8010: Async IO in FileSystemDataVault
I’m sure scala/java has support for asynchronously reading (byte ranges of) local files. That should be used in the FileSystemDataVault. This should free more threads

frcroth:
### Steps to test:
- View datasets (local, zarr3 sharded as it range queries)

### Issues:
- fixes #8010

------

## Summary by CodeRabbit

- **New Features**
	- Enhanced slider functionality for easier value adjustments.
	- Improved search functionality for unnamed segments.
	- Increased loading speeds for precomputed meshes.
	- Added a button to select all matching non-group results in the search popover.
	- Unified terminology in the UI, replacing "Resolution" with "Magnification."
	- Superusers can now delete workflow reports.

- **Bug Fixes**
	- Resolved issues with dataset uploads, bbox export, and annotation saving.
	- Corrected scrolling functionality in the trees and segments tab.

- **Improvements**
	- Migrated nightly screenshot tests to GitHub Actions.
	- Thumbnails for datasets now utilize selected mapping from view configuration.
	- Enhanced asynchronous reading of image files for better performance.


fm3:
I did some testing here with thread count set to 1 and a large file.

  ```
    val vaultPath = new VaultPath(new URI(f"file:///largeFile"), FileSystemDataVault.create)
    for {
      before <- Instant.nowFox
      bytes <- vaultPath.readBytes(Some(Range.Long(0, 1147483646, 1)))
      _ = Instant.logSince(before, "reading bytes")
    } yield Ok
  ```

I requested this route several times in parallel. This way I could confirm that the thread is indeed not blocked in this async variant, and can reply to each request after the respective reading operation is complete. On master, it can only reply after all previous reading operations are complete (meaning the thread is blocked).

However, I also noticed that reading the file was consistently faster on master. On average I measured 634 ms on master versus 1160 ms here.

So it could be argued that using the synchronous variant (with more threads that consume some memory) is worth the faster loading, compared to this change (that could work with fewer threads). Do you have any opinions on this @frcroth @normanrz ? Of course the testing conditions were pretty specific (ssd on my laptop, warm OS cache, always the same file) and differ somewhat from realistic datastore load conditions.

normanrz:
Async usually helps with throughput when dealing with a lot of requests. It doesn't necessarily yield improvements with just a few requests. However, it shouldn't be much slower, either. So, these results are a bit surprising.

fm3 (commented):
Alright, I did some more measuring and it looks like my numbers from yesterday were a little unfair, since they used parallel reading in the async case, while on master, everything was sequential (due to single thread + blocking).
In my new measurements from today, both implementations come in at 600–650ms.
So I’d say we should go for the async variant :)

@frcroth how do you feel about the suggestions by our AI reviewer? I guess it might make sense to add some validations for too-large files / ranges. Not sure about the suggested try/catch, but it does seem plausible too. What do you think?

fm3 (approved):

PR #8127, opened 2024-10-16
Route to register aiModel

fm3:
To be called by a voxelytics PublishModelTask. Needs the token of a superuser.

We tested that this works during the review of https://github.com/scalableminds/voxelytics/pull/3763

### Issues:
- contributes to https://github.com/scalableminds/voxelytics/issues/3750

------

## Release Notes

- **New Features**
	- Introduced a new route for registering AI models via `POST /aiModels/register`.
	- Added support for registering AI models with validations for uniqueness and existence checks.
	- Expanded AI model categories to include `em_synapses`, `em_neuron_types`, and `em_cell_organelles`.

- **Bug Fixes**
	- Improved error handling in model retrieval methods to provide a more neutral response when an object is not found.

- **Schema Updates**
	- Updated the database schema to include new tables and types related to AI models and their training annotations.


fm3:
@valentin-pinkau I created the route `POST api/v8/aiModels/register`. It expects a json object body with

id: String, // must be a valid MongoDB ObjectId, compare https://pymongo.readthedocs.io/en/stable/api/bson/objectid.html
dataStoreName: String,
name: String,
comment: Option[String],
category: Option[String] // if present, must be one of em_neurons, em_nuclei, em_synapses, em_neuron_types, em_cell_organelles

It expects that the model is already present in the .aiModels directory, compare worker. The route does not currently assert that the model is there on disk and usable.

Do you think that will work for the PublishModelTask?

app/utils/sql/SQLDAO.scala
```diff
@@ -47,7 +47,7 @@ abstract class SQLDAO[C, R, X <: AbstractTable[R]] @Inject()(sqlClient: SqlClien
       case Some(r) =>
         parse(r) ?~> ("sql: could not parse database row for object" + id)
       case _ =>
-        Fox.failure("sql: could not find object " + id)
+        Fox.empty
```
fm3, line 50: We used Fox.empty already in many overwriting findOnes, but not this one in the superclass. I don’t think it will hurt, though, and this way I can distinguish between error and empty downstream.

app/controllers/AiModelController.scala
```diff
@@ -57,6 +57,16 @@ object UpdateAiModelParameters {
   implicit val jsonFormat: OFormat[UpdateAiModelParameters] = Json.format[UpdateAiModelParameters]
 }
 
+case class RegisterAiModelParameters(id: ObjectId, // must be a valid MongoDB ObjectId
```
frcroth, line 60: Can this be validated in registerAiModel?

fm3: It’s automatically validated by the json reads of ObjectId. I just added the comment as a hint for writing the client.

conf/evolutions/reversions/123-more-model-categories.sql
```diff
@@ -0,0 +1,9 @@
+START TRANSACTION;
+
+do $$ begin ASSERT (select schemaVersion from webknossos.releaseInformation) = 123, 'Previous schema version mismatch'; end; $$ LANGUAGE plpgsql;
+
+-- removing enum values is not supported in postgres, see https://www.postgresql.org/docs/current/datatype-enum.html#DATATYPE-ENUM-IMPLEMENTATION-DETAILS
```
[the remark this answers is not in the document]
fm3: Good Idea! I’ll use the isDeleted boolean for that

frcroth (approved):

PR #8129, opened 2024-10-16
Fix skeleton search

### LINKED ISSUE #8057: Skeleton search does not focus the search result
## Context / Current Behavior
In the tracing UI > Skeleton Tab, the skeleton search does not focus/active the search result. Parent groups/folders are also not expanded.

`Die Suchfunktion im Skeleton Tab findet zwar das Gesuchte, bringt einen aber nicht mehr zum Treffer. Besonders nervig, wenn sich der Tree in eingeklappten Ordnern befindet.`

Likely, a regression of #5872 

## Expected Behavior
The search result should be focues/set to active and any parent groups should be expanded.

- Browser name and version: all
- Operating System and version: all
- Version of WEBKNOSSOS (Release or Commit): 24.08.0 

MichaelBuessemeyer:
The skeleton search did not expand the group of the skeleton it selected. I changed the code so that all parent groups of a selected skeleton via the search are not automatically expanded.

- https://___.webknossos.xyz

### Steps to test:
- open https://fixskeletontabsearch.webknossos.xyz/annotations/6710b779010000af00c4b903#3584,3584,1024,0,2
- open the skeleton tab
- collapse all groups in case they aren't
- use the skeleton search with the term `search me` and go through all selection
- all necessary skeleton groups should auto-expand in case the tree selected via the search is within a collapsed group
- The selected trees should be highlighted as well

### Issues:
- fixes #8057

------


frontend/javascripts/oxalis/view/right-border-tabs/segments_tab/segments_view.tsx
```diff
@@ -420,7 +416,6 @@ class SegmentsView extends React.Component<Props, State> {
     groupToDelete: null,
     groupsSegmentsVisibilityStateMap: {},
     activeStatisticsModalGroupId: null,
-    expandedGroupKeys: [],
```
MichaelBuessemeyer, line 423: I made this an memoized function of the component. 

While working with this component I saw the following potential danger: The expanded groups were correctly derived from the props / store state, but having this as local state as well as in the store might lead to a state where the local state might be ahead of the store because some codeline accidentally uses `setState` instead of updating the store and from their receiving an updated state via the `getDerivedStateFromProps` method. 

Moreover, I thought that's how it was done in the Trees tab as well, but seems that I was wrong here. => Feel free to argue about this change and I'll change it back.

frontend/javascripts/oxalis/view/right-border-tabs/segments_tab/segments_view.tsx
```diff
@@ -486,6 +477,8 @@ class SegmentsView extends React.Component<Props, State> {
     }
   }
 
+  getExpandedGroupKeys = () => getExpandedKeysWithRoot(this.props.segmentGroups);
+
```
MichaelBuessemeyer, line 481: `getExpandedKeysWithRoot` is already memoized

frontend/javascripts/oxalis/view/right-border-tabs/tree_hierarchy_view.tsx
```diff
@@ -94,7 +94,7 @@ function TreeHierarchyView(props: Props) {
       // For some React rendering/timing reasons, the target element might not be rendered yet. That messes with calculating the offsets for scrolling. Hence delay this a bit
       setTimeout(() => {
         if (treeRef.current) treeRef.current.scrollTo({ key: activeTreeKey, align: "auto" });
-      }, 30);
+      }, 50);
```
MichaelBuessemeyer, line 97: I increased the time a little as it happened to me that under quite some load chrome wasn't fast enough to expand the groups before the scrollTo call was made and thus the scrollTo was ignored. The 50 ms might be more "slow browser tolerant"

hotzenklotz (approved):
Code LGTM. 

Skeletons work well. I had issues with segments, though.

PR #8134, opened 2024-10-18
Add Github action for nightly tests

hotzenklotz:
I converted the nightly screenshot tests from a CircleCI workflow to Github actions. See https://github.com/scalableminds/webknossos/actions/workflows/nightly.yml

I also switch from the `puppeteer` package to `puppeteer-core`, which will not download and maintain a Chrome browser on every test run. (See https://pptr.dev/guides/installation)

### Steps to test:
- Successful CI run should be enough: https://github.com/scalableminds/webknossos/actions/runs/11420258191/job/31775745570

###  Observations and Questions
- [ ] We use `puppetteer` to launch a headless Chrome and take the screenshots. I was under the impression that this launches a local, headless chrome. But instead it seams to run a Chrome instance on MacOS in BrowserStack exclusively. 
- [ ] We maintain a [docker image with puppeteer ](https://github.com/scalableminds/dockerfiles/tree/master/puppeteer)which might be obsolete after this PR and we should consider removing it.

### Issues:
- fixes #

------

### DIFF
--- a/frontend/javascripts/test/puppeteer/dataset_rendering_helpers.ts
+++ b/frontend/javascripts/test/puppeteer/dataset_rendering_helpers.ts
@@ -357,7 +357,7 @@ export function setupBeforeEachAndAfterEach() {
   test.beforeEach(async (t) => {
     t.context.browser = await puppeteer.launch({
       args: [
-        "--headless",
+        "--headless=new",
         "--hide-scrollbars",
         "--no-sandbox",
         "--disable-setuid-sandbox",
--- a/frontend/javascripts/test/puppeteer/dataset_rendering_helpers.ts
+++ b/frontend/javascripts/test/puppeteer/dataset_rendering_helpers.ts
@@ -362,7 +362,8 @@ export function setupBeforeEachAndAfterEach() {
         "--no-sandbox",
         "--disable-setuid-sandbox",
         "--disable-dev-shm-usage",
-        "--use-gl=swiftshader",
+        // "--use-gl=swiftshader",
+        // "--enable-gpu",
       ],
       dumpio: true,
     });
--- a/frontend/javascripts/test/puppeteer/dataset_rendering_helpers.ts
+++ b/frontend/javascripts/test/puppeteer/dataset_rendering_helpers.ts
@@ -2,17 +2,16 @@
 import urljoin from "url-join";
 // @ts-expect-error ts-migrate(7016) FIXME: Could not find a declaration file for module 'node... Remove this comment to see the full error message
 import fetch, { Headers, Request, Response, FetchError } from "node-fetch";
-import type { Browser } from "puppeteer";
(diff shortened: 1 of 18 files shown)


hotzenklotz:
@philippotto @daniel-wer  We use `puppetteer` to launch a headless Chrome and take the screenshots. I was under the impression that this launches a local, headless Chrome. But instead it seams to run a Chrome instance on MacOS in BrowserStack exclusively.
I was also not able to run `yarn test-screenshot` locally since it immediately asked for my BrowserStack credentials.

Looking at this code, it seems to me, that the local browser instance (`puppeteer.launch`) is immediately discarded and the overwritten by the BrowserStack connection (`puppeteer.connect`). 

Personally I don't have feelings either way. I don't need to run the screenshot test locally and would be fine with doing it entirely with BrowserStack to avoid any headaches with WebGL and headless Chrome in Github actions runners. What do you think?

frontend/javascripts/test/puppeteer/dataset_rendering_helpers.ts
```diff
@@ -357,12 +357,13 @@ export function setupBeforeEachAndAfterEach() {
...
         "--hide-scrollbars",
         "--no-sandbox",
         "--disable-setuid-sandbox",
         "--disable-dev-shm-usage",
-        "--use-gl=swiftshader",
+        // "--use-gl=swiftshader",
+        // "--enable-gpu",
       ],
       dumpio: true,
     });
```
MichaelBuessemeyer, line 369: AS this also ready like it is ignored code, I'd say we can comment it out for now and check whether the CI still works. If it does, I'd say we remove this completely

MichaelBuessemeyer (commented):
Awesome, thanks for this pr and fixing the CI.

Please see my one comment about removing potential unused code. Afterwards I'd say it should be ready to merge :)

frontend/javascripts/test/puppeteer/dataset_rendering_helpers.ts
hotzenklotz: I uncommented this code and added a comment explaining why. We should discuss this at some point with @philippotto and/or @daniel-wer . See my conversation/comment above.

hotzenklotz: Success

MichaelBuessemeyer (approved):
The test run is green. So I'd say we can merge this and have a working nightly. We can discuss this later with @daniel-wer @philippotto :)

daniel-wer:
I think at some point @philippotto switched the screenshot tests to run on browser stack since that was way more reliable and avoided crashes. I don't know much more about that, so let's wait until he is back.

philippotto:
> I think at some point @philippotto switched the screenshot tests to run on browser stack since that was way more reliable and avoided crashes.

Yes, I think, this is exactly it. The PR looks good :+1:

PR #8137, opened 2024-10-21
Zarr streaming e2e test

### LINKED ISSUE #7997: Add end to end tests for zarr streaming
## Detailed Description

Zarr streaming is a crucial feature for our pipelines. Thus, we should ensure that this feature is tested to avoid easily catchable mistakes breaking the streaming mechanism. Therefore, I suggest to write end to end tests testing all streaming routes, especially the ones that directly request an array

frcroth:
### TODOs:
- [x] Reset changes to tools/test.sh
- [x] Zarr 3 TEsts (zarr json and cube)

### Issues:
- fixes #7997 

## Release Notes

- **New Features**
  - Enhanced error messages for user authentication, organization management, and dataset handling for clearer user feedback.
  - Introduced new end-to-end tests for Zarr data formats, expanding dataset functionality coverage.

- **Bug Fixes**
  - Improved error handling messages related to dataset access and user management for better clarity.

- **Chores**
  - Updated `.gitignore` to track the `test-dataset.zip` file for important test data.
  - Modified test setup to ensure necessary datasets are prepared before running tests, enhancing robustness.
  - Adjusted volume mapping for the `e2e-tests` service in the configuration.
  - Enhanced CircleCI configuration to ensure the directory for test data is created with appropriate permissions before tests run.


frontend/javascripts/test/backend-snapshot-tests/datasets.e2e.ts
```diff
@@ -29,27 +37,28 @@ test.serial("getDatasets", async (t) => {
   writeTypeCheckingFile(datasets, "dataset", "APIDatasetCompact", {
     isArray: true,
   });
-  t.snapshot(datasets);
+  console.log(datasets);
```
fm3, line 40: I’m not super familiar with the e2e tests. Do they typically console.log stuff, in addition to writing the snapshots? If not, maybe remove this line again?

frontend/javascripts/test/e2e-setup.ts
```diff
@@ -70,6 +71,13 @@ global.fetch = function fetchWrapper(url, options) {
     newUrl = `http://localhost:9000${url}`;
   }
 
+  if (!options) {
+    options = {};
```
fm3, line 75: looks like this options thing was already used, e.g. in line 82. Why is this setup step needed now?

frontend/javascripts/test/e2e-setup.ts
```diff
@@ -78,6 +86,12 @@ global.fetch = function fetchWrapper(url, options) {
   return promise;
 };
 
+export async function scanDatasetsFromDisk() {
+  await fetch(`http://localhost:9000/data/triggers/checkInboxBlocking?token=${currToken}`, {
```
fm3, line 90: Could this use the `triggerDatasetCheck` function from `admin_rest_api`? I’d like to avoid having another client knowing about the http api, so that we don’t have to be extra careful when changing it.

.circleci/config.yml
```diff
@@ -118,6 +118,8 @@ jobs:
       - run:
           name: Run end-to-end tests
           command: |
+            mkdir -p "binaryData/Connectomics department"
+            chmod 777 "binaryData/Connectomics department"
```
fm3, line 122: below it looks like the new test actually uses Organization_X. Why is it necessary to create the Connectomics department directory here?

test/e2e/End2EndSpec.scala
```diff
@@ -43,4 +47,50 @@ class End2EndSpec(arguments: Arguments) extends Specification with GuiceFakeAppl
...
+    if (!dataDirectory.exists()) {
+      throw new Exception("Organization_X directory does not exist.")
+    }
+    val testDatasetZip = new File(testDatasetPath)
+    if (!testDatasetZip.exists()) {
+      throw new Exception("Test dataset zip file does not exist.")
+    }
+    // Stop if the test dataset is already unzipped
+    if (dataDirectory.listFiles().exists(_.getName == "test-dataset")) {
+      return
```
fm3, line 62: please avoid `return` in scala, always use full if/else instead. compare https://tpolecat.github.io/2014/05/09/return.html

test/e2e/End2EndSpec.scala
```diff
@@ -43,4 +47,50 @@ class End2EndSpec(arguments: Arguments) extends Specification with GuiceFakeAppl
...
+          fos.write(buffer, 0, len)
+          len = zis.read(buffer)
+        }
+        fos.close()
+
+      }
+      zipEntry = zis.getNextEntry
+    }
+
+    zis.closeEntry()
```
fm3, line 93: This seems quite involved. Is there a reason this does not use the ZipIO.unzipToFolder function that we already have?

fm3 (commented):
Nice stuff! I added a couple of questions

frontend/javascripts/test/backend-snapshot-tests/datasets.e2e.ts
frcroth: Yes, simply forgotten

test/e2e/End2EndSpec.scala
frcroth: There were no dependencies to any other local package in the test package and I thought that may have been intentional. I have no reason in particular.

.circleci/config.yml
fm3: Hm, I think that is a legacy artifact, from when things were named differently. Could you name it Organization_X there too and see if that breaks anything?

test/e2e/End2EndSpec.scala
fm3: I see, well I don’t know if that was intended originally, but to me using the zip utils here seems acceptable. We might even catch bugs there if this test fails…

frcroth: Changed it

fm3 (commented):
Very nice, glad to see that we now have some tests with actual on-disk image data. This will test a lot of other functionality too that the zarr streaming is built upon.

To keep the snapshot files short and readable, consider snapshotting not the whole 32³ bucket data but instead a slice or a checksum.

fm3 (approved):
:tada:

Issue #7309, opened 2023-09-05
URLs with expired sharing token prevent logged in users from editing


hotzenklotz:
## Context
Someone shared a WK shortened link to an annotation with me. The annotation URL contains an (expired) token. Since I was logged in, I was able to open the annotation nonetheless. In this particular case the annotation was "shared" and I wanted to edit it. WK wrongly assumed I can not make any changes since the token was expired.  I get a lot of warnings.

<img width="452" alt="image" src="https://github.com/scalableminds/webknossos/assets/1105056/0da718a9-ec48-45a7-916b-679b4b5680a9">

A likely fix could be that WK checks whether one has access to an annotation regardless if the URL has a token or not. 

## Expected Behavior
I should be able to edit any annotation that my user has access rights to regardless through which URL I opened it from.

## Current Behavior
A (team)-shared annotation open from a shortened WK link (e.g. https://webknossos.org/links/abc) including a expired token does not let me make any changes.

## Steps to Reproduce the bug
1. Create a new shared annotation
2. Share a link to annotation (WK will automagically include a token)
3. Expire the token (delete from DB?)
4. Log in, open the annotation through the shared URL
5. Start editing --> boom

- [ ] Cannot reproduce the bug anymore / needs deeper investigation.

## Your Environment for bug

- Browser name and version: Chrome 116
- Operating System and version: MacOS
- Version of WEBKNOSSOS (Release or Commit): 24599 (September 2023)
- [ ] Specific to long-running jobs (set `jobsEnabled=true` in `application.conf`)
- [ ] Specific to webknossos.org (set `isDemoInstance=true` in `application.conf`)

fm3:
The tracingstore auth always uses a token, cookies are not sent there. I think the frontend decides which token to send (probably the uri token has precedence).

I see a couple of options to fix this
 - (1) We could send multiple tokens to the backend (Needs changes in all tracingstore/datastore routes, though)
 - (2) The frontend could ping a separate route asking if a token is expired, and if so, discard it and send the user’s token instead
 - (3) The frontend could retry failing routes with the user’s own token

@philippotto do you have opinions on this? I guess 2 is the most viable option. There are certainly also other ways of solving this that I couldn’t think of just now

philippotto:
TLDR: A mix of 2 and 3 would probably work.

[...]

PR #8139, opened 2024-10-21
Automatically use user token when sharing token is not sufficient for a request

MichaelBuessemeyer:
This PR improves the request retry functionality by trying using the token from the URL only upon the first try. If this results in a 403 (permissions not sufficient), the new code retries with the personal token of the user and if that succeeds, the URL token is removed and the previous error toast regarding a potentially expired token is hidden.

- https://___.webknossos.xyz

### Steps to test:
- create an annotation
- make it public but turn off editing by others
- copy the annotation link & open it, it should include a sharing token
- do some annotation actions & try saving
- the first saving try should fail & show an error toast
- Then another try should be done, which should succeed, remove the token from the URL, close the error toast and show another info toast about the retry.

### Issues:
- fixes [#7309](https://github.com/scalableminds/webknossos/issues/7309)

------

## Release Notes

- **New Features**
  - Improved slider functionality and search capabilities for unnamed segments.
  - Enhanced loading speed for precomputed meshes.
  - Unified UI terminology for "Magnification."
  - Added support for remote OME-Zarr NGFF version 0.5 datasets.

- **Bug Fixes**
  - Resolved dataset upload failures with absolute paths.
  - Fixed issues with insufficient sharing tokens when saving annotations.
  - Corrected automatic group expansion in skeleton searches.
  - Addressed incorrect shape returns and sorting issues in zarr streaming.
  - Enabled annotation creation for public datasets of other organizations.

- **Improvements**
  - Enhanced error handling for user notifications and logging.
  - Adjusted error notification strategies for server requests.
  - Improved control flow for token retrieval based on availability.
  - Enhanced type safety in API request handling.


MichaelBuessemeyer:
Sadly, the new change that auto removes a token from the URL if the personal token is more privileged does introduce quite some new cyclic dependencies.
Therefore, the auto removal is discarded for now and can be re-added later.

MichaelBuessemeyer:
Here is the link about the follow up issue: https://github.com/scalableminds/webknossos/issues/8140

frontend/javascripts/admin/api/token.ts
```diff
@@ -33,22 +35,48 @@ export function getSharingTokenFromUrlParameters(): string | null | undefined {
...
 
       // If three new tokens did not fix the 403, abort, otherwise we'll get into an endless loop here
       if (tries < 3) {
-        return doWithToken(fn, tries + 1);
+        // If using the url sharing token failed, we try the user specific token instead.
+        const result = await doWithToken(fn, tries + 1, false);
+        // Upon successful retry with own token, discard the url token.
+        if (useURLTokenIfAvailable) {
+          shouldUseURLToken = false;
+          Toast.info(
```
hotzenklotz, line 74: I think this is a technical detail that we shouldn't bother users with. I would replace the `Toast` with a `console.log()` for debugging purposes but not actively show it in the UI.

MichaelBuessemeyer: So just hiding the error toast is enough you think? My thought was to somewhat inform the user that saving / or some other request worked on second try.

Therefore, maybe a toast with a message like "Request succeeded on upon retry" or something similar?

MichaelBuessemeyer: But yeah, I agree that the technical details are definitely too much in my toast message

MichaelBuessemeyer:
We had a little internal discussion about how to handle the error toasts.

The solution we came up with is to disable the toast that shows the backend error why using the auto retry mechasim of `doWithToken` when annotation updates are sent to the server. The `save_saga` itself shows an error toast about the saving not working and retrying once the three auto retries of `doWithToken` failed.

In order to not let the backend error go completely silent, I changed the request handling in such a way that if no error toast is shown, the error is at least logged to the console. (Showing an error toast itself already automatically logs the error to the console. Thus, whenever a request error occurs, the error is logged to the console, it is irrelevant what the request `showErrorToast` config setting says.

@daniel-wer Maybe you also have an opinion on this. I am open for suggestions how to maybe improve this behaviour

hotzenklotz:
@coderabbitai review

PR #8142, opened 2024-10-23
Activate HTTP API Version 8

fm3:
I forgot to bump that number in #7386 

This is needed to access the explicit /api/v8/ routes, e.g. `https://webknossos.org/api/v8/buildinfo` currently fails while v7 works. But webknossos.versioned.routes already provides v8.

This blocks https://github.com/scalableminds/webknossos-libs/pull/1185


MichaelBuessemeyer (approved):
Oh no, sorry I did not know about this :see_no_evil: 

The changes lgtm :tada:

fm3:
> Oh no, sorry I did not know about this

No worries, since we hadn’t started adapting the libs, nobody noticed until now :sweat_smile: Also this is pretty new and I’m still getting used to it myself. That’s why I also added it into the common-edge-cases.md thing

PR #8151, opened 2024-10-28
Fix datavault tests for non-existent s3 buckets

fm3:
Looks like someone created `s3://non-existing-bucket`. We shouldn’t rely on that really not existing. I now add a UUID to the request paths where we want to test stuff not existing.

I also changed that directory listing for a non-existing bucket is supposed to return Failure, not Empty. This was inconsistent between DataVaults.

I noticed that there are other combinations that we are not currently testing, and considered adding them here, but let’s instead get this in master asap, as it currently blocks the CI of all branches.

### Steps to test:
- CI should be enough

### DIFF
--- a/test/backend/DataVaultTestSuite.scala
+++ b/test/backend/DataVaultTestSuite.scala
@@ -18,6 +18,7 @@ import net.liftweb.common.{Box, Empty, EmptyBox, Failure, Full}
 import play.api.libs.json.JsString
 import play.api.test.WsTestClient
 
+import java.util.UUID
 import scala.collection.immutable.NumericRange
 import scala.concurrent.ExecutionContext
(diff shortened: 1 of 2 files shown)


frcroth (approved):

test/backend/DataVaultTestSuite.scala
```diff
@@ -141,8 +144,18 @@ class DataVaultTestSuite extends PlaySpec {
         }
 
         "return empty box" when {
-          "requesting a nox-existent object" in {
-            val uri = new URI("s3://non-existing-bucket/non-existing-object")
+          "requesting a nox-existent bucket" in {
```
frcroth, line 147:
```suggestion
          "requesting a non-existent bucket" in {
```

frcroth (approved):

PR #8156, opened 2024-10-30
Allow for the deletion of workflow reports

### LINKED ISSUE #7973: Allow to delete/archive voxelytics workflow reports
Occasionally, users want to delete voxelytics workflow reports to avoid sharing secret information.
We should have a look how such a functionality would fit in.

frcroth:
- https://___.webknossos.xyz

### Steps to test:
- Start a workflow
- Delete the workflow by clicking on the workflow and then on the three dots menu selecting Delete Workflow report.
- Notice it is no longer in the list
- (Note that there is nothing in the database)

### Questions
- It is possible to delete the workflow while it is running. Should this be disabled? This causes the workflow to fail, more or less as expected. It does not seem like anything breaks.
- Since this is only for super users, should the button in the frontend be only available for super users (disabled / not shown?)

### TODOs:
- [ ] ...

### Issues:
- fixes #7973 

------
	- I have on logger statement in the deletion route. I think that could remain?

## Release Notes

- **New Features**
	- Added the ability to add metadata to annotations for Trees and Segments.
	- Enhanced search functionality to find unnamed segments using their default names.
	- Introduced a new button in the search popover for selecting all matching non-group results.
	- Added a summary row in the time tracking overview.
	- Unified UI terminology by replacing "Resolution" with "Magnification."
	- Users can now delete workflow reports and workflows directly from the application.
	- Admins can view and cancel all jobs with job ownership displayed.

- **Bug Fixes**
	- Resolved issues with dataset uploads and bbox export menu.
	- Fixed automatic expansion of groups in skeleton searches and zarr streaming version corrections.
	- Ensured insufficient sharing tokens are discarded appropriately.

- **Chores**
	- Migrated nightly screenshot tests from CircleCI to GitHub Actions.

### DIFF
--- a/app/controllers/VoxelyticsController.scala
+++ b/app/controllers/VoxelyticsController.scala
@@ -3,6 +3,7 @@ package controllers
 import com.scalableminds.util.time.Instant
 import com.scalableminds.util.tools.{Fox, FoxImplicits}
 import models.organization.OrganizationDAO
+import models.user.UserService
 import models.voxelytics._
 import play.api.libs.json._
 import play.api.mvc._
@@ -19,6 +20,7 @@ class VoxelyticsController @Inject()(
     organizationDAO: OrganizationDAO,
     voxelyticsDAO: VoxelyticsDAO,
     voxelyticsService: VoxelyticsService,
+    userService: UserService,
     lokiClient: LokiClient,
     wkConf: WkConf,
     sil: Silhouette[WkEnv])(implicit ec: ExecutionContext, bodyParsers: PlayBodyParsers)
@@ -156,6 +158,17 @@ class VoxelyticsController @Inject()(
           )
         )
       } yield JsonOk(result)
+    }
+
+  def deleteWorkflow(workflowHash: String): Action[AnyContent] =
+    sil.SecuredAction.async { implicit request =>
+      for {
+        _ <- bool2Fox(wkConf.Features.voxelyticsEnabled) ?~> "voxelytics.disabled"
+        _ <- userService.assertIsSuperUser(request.identity)
+        _ <- voxelyticsDAO.findWorkflowByHash(workflowHash) ?~> "voxelytics.workflowNotFound" ~> NOT_FOUND
+        _ = logger.info(s"Deleting workflow with hash $workflowHash")
+        _ <- voxelyticsDAO.deleteWorkflow(workflowHash)
+      } yield Ok
     }
 
   def storeWorkflowEvents(workflowHash: String, runName: String): Action[List[WorkflowEvent]] =
--- a/app/controllers/VoxelyticsController.scala
+++ b/app/controllers/VoxelyticsController.scala
@@ -166,8 +166,8 @@ class VoxelyticsController @Inject()(
         _ <- bool2Fox(wkConf.Features.voxelyticsEnabled) ?~> "voxelytics.disabled"
         _ <- userService.assertIsSuperUser(request.identity)
         _ <- voxelyticsDAO.findWorkflowByHash(workflowHash) ?~> "voxelytics.workflowNotFound" ~> NOT_FOUND
-        _ = logger.info(s"Deleting workflow with hash $workflowHash")
-        _ <- voxelyticsDAO.deleteWorkflow(workflowHash)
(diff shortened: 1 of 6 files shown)


fm3:
> It is possible to delete the workflow while it is running. Should this be disabled? This causes the workflow to fail, more or less as expected. It does not seem like anything breaks.

I’d say that’s ok. Maybe include a warning in the deletion confirmation modal along the lines of “Note that if the workflow is still running, this may cause it to fail”. Can be a standard text, no need to figure out whether it is, in fact, still running.

> Since this is only for super users, should the button in the frontend be only available for super users (disabled / not shown?)

That’s right, it should not be shown for non-superusers. Maybe dieknolle3333 can give you some pointers as to how to achieve this.

app/controllers/VoxelyticsController.scala
```diff
@@ -158,6 +160,17 @@ class VoxelyticsController @Inject()(
       } yield JsonOk(result)
     }
 
+  def deleteWorkflow(workflowHash: String): Action[AnyContent] =
+    sil.SecuredAction.async { implicit request =>
+      for {
+        _ <- bool2Fox(wkConf.Features.voxelyticsEnabled) ?~> "voxelytics.disabled"
+        _ <- userService.assertIsSuperUser(request.identity)
+        _ <- voxelyticsDAO.findWorkflowByHash(workflowHash) ?~> "voxelytics.workflowNotFound" ~> NOT_FOUND
+        _ = logger.info(s"Deleting workflow with hash $workflowHash")
```
fm3, line 169:
```suggestion
        _ = logger.info(s"Deleting workflow with hash $workflowHash in organization ${request.identity._organization}")
```

app/models/voxelytics/VoxelyticsDAO.scala
```diff
@@ -1135,4 +1135,17 @@ class VoxelyticsDAO @Inject()(sqlClient: SqlClient)(implicit ec: ExecutionContex
         """.asUpdate)
     } yield ()
 
+  def deleteWorkflow(hash: String): Fox[Unit] =
```
fm3, line 1138:
```suggestion
  def deleteWorkflow(hash: String, organizationId: String): Fox[Unit] =
```

app/models/voxelytics/VoxelyticsDAO.scala
```diff
@@ -1135,4 +1135,17 @@ class VoxelyticsDAO @Inject()(sqlClient: SqlClient)(implicit ec: ExecutionContex
...
+    for {
+      _ <- run(q"""
+                  DELETE FROM webknossos.voxelytics_workflows
+                  WHERE hash = $hash;
```
fm3, line 1142:
```suggestion
                  WHERE hash = $hash;
                  AND _organization = $organizationId
```

app/models/voxelytics/VoxelyticsDAO.scala
```diff
@@ -1135,4 +1135,17 @@ class VoxelyticsDAO @Inject()(sqlClient: SqlClient)(implicit ec: ExecutionContex
...
+                  """.asUpdate)
+      _ <- run(q"""
+                  UPDATE webknossos.jobs
+                  SET _voxelytics_workflowhash = NULL
+                  WHERE _voxelytics_workflowhash = $hash;
```
fm3, line 1147:
```suggestion
                  SET _voxelytics_workflowHash = NULL
                  WHERE _voxelytics_workflowHash = $hash;
```

frontend/javascripts/admin/voxelytics/task_list_view.tsx
```diff
@@ -421,6 +425,26 @@ export default function TaskListView({
     }
   }
 
+  async function deleteWorkflowReport() {
+    await modal.confirm({
+      title: "Delete Workflow Report",
+      content:
+        "Are you sure you want to delete this workflow report?\nNote that if the workflow is still running, this may cause it to fail.",
```
fm3, line 432:
```suggestion
        "Are you sure you want to delete this workflow report? This can not be undone.\nNote that if the workflow is still running, this may cause it to fail.",
```

fm3 (changes requested):
Looking forward to this :tada:

fm3 (approved):
:tada: :shipit: (though please wait until at least tomorrow with deploying)

Issue #7334, opened 2023-09-18
Add "Rotate" button to dataset layers


philippotto:
## Detailed Description
The user should be able to rotate dataset layers on the fly. The rotation should probably be persisted. Implementation-wise the existing affine transformations can be used.
~One important aspect might be that skeleton layers don't know anything about transformations, yet. This means that the nodes would stay at their current position when rotating layers.~ transforms should work with nodes currently. however, one cannot modify the skeleton layer if it's rendered with transforms.

## Context
See https://scm.slack.com/archives/C5AKLAV0B/p1694764730277899

MichaelBuessemeyer:
@hotzenklotz What about this ThreeJs example?
https://threejs.org/examples/misc_controls_transform.html

We could use that in a smaller scale, show a cube with marked surfaces and only enable the rotation controls. If the controls are too quirky they can be replaced with three and sliders ranging from 0 to 360 and the three scene is only used for displaying what the will be rendered.

Moreover, I'd like to try to add some kind of camera to the scene so the user can see what to expect.

Another idea would be to have a small annotation view (3 rectangles) like the ortho viewports in which the current  "cube with marked surfaces" is rendered according to the current settings

Kinda like this: ![Image](https://github.com/user-attachments/assets/8e20705d-375f-4168-abf3-6a3293ba42ec)
(Note: I did not check whether that's a possible valid orientation)

hotzenklotz:
1. The ThreeJS example is really cool. The [code](https://github.com/rollup/three-jsnext/blob/master/examples/misc_controls_transform.html) for that is also fairly small (<200LOC) so that could be a viable option.
2. Regardless of the preview/threejs/etc, I would always have a synced number input so that a) the users can double check any changes made with actual numbers or b) input the numbers directly if they know them already.
3. I always images this feature/setting to be in a settings pages somewhere. So, I would not show/preview any actual data but just the generic "texture" with the colored viewport names XY, etc. 

I personally find the rotating, interactive cube preview (threejs) easier for my brain than to imagine the XZ plane in the XY viewport (like in you  mockup above). But I could imaging other people feeling differently about this.

MichaelBuessemeyer:
Additional idea form @valentin-pinkau: have a unrotated semi-transparent cube that shows the initial rotation for the proposed visualization.

Moreover, valentin also mentioned that a visualization may not be necessary. Instead the user would switch a few times between the settings and viewing the dataset until the wanted orientation is reached.

[...]

PR #8159, opened 2024-10-30
Allow permanent dataset layer rotation in dataset settings

MichaelBuessemeyer:
This PR adds new settings to the dataset settings for each layer. A interface (still in design process) enables the user to set a affine transformation matrix by letting the user define angles for each axis around which the dataset should be rotated.

- https://___.webknossos.xyz

### Steps to test:
- abc

### TODOs / Problems:
- [ ] Design a visualization of the result. For some first ideas look at the issue [#7334](https://github.com/scalableminds/webknossos/issues/7334)
- [ ] The rotation should be done around the center of the dataset. For this, a translation by the dataset center is needed
- [ ] The interface works by defining Euler angles. The problem is that Euler angles are not deterministic / there are multiple ways to express a certain rotation with different Euler angle values. See: https://stackoverflow.com/questions/32235395/does-euler-angles-quaternion-euler-angles-always-result-in-an-equivalent-r
  - This makes it hard to update the settings correctly ->  a 180 degree rotation around y is equivalent to 180 on x and 180 on z axis :/
  - A solution for this is needed as the interface currently is pretty quirky.
- [ ] 

### Issues:
- fixes [#7334](https://github.com/scalableminds/webknossos/issues/7334)

------

### DIFF
--- a/frontend/javascripts/dashboard/dataset/dataset_settings_data_tab.tsx
+++ b/frontend/javascripts/dashboard/dataset/dataset_settings_data_tab.tsx
@@ -22,6 +22,7 @@ import {
   FormItemWithInfo,
   RetryingErrorBoundary,
   jsonEditStyle,
+  AxisRotationSettingForLayer,
 } from "dashboard/dataset/helper_components";
 import { startFindLargestSegmentIdJob } from "admin/admin_rest_api";
 import { jsonStringify, parseMaybe } from "libs/utils";
@@ -391,8 +392,10 @@ function SimpleLayerForm({
               {
                 validator: syncValidator(
                   (value: string) =>
-                    dataLayers.filter((someLayer: APIDataLayer) => someLayer.name === value)
-                      .length <= 1,
+                    (dataLayers &&
+                      dataLayers.filter((someLayer: APIDataLayer) => someLayer.name === value)
+                        .length <= 1) ||
+                    dataLayers == null,
                   "Layer names must be unique.",
                 ),
               },
@@ -613,6 +616,12 @@ function SimpleLayerForm({
               )}
             </div>
           ) : null}
+          <Row gutter={32}>
+            <Col span={12}>
+              Permanent dataset rotation:
+              <AxisRotationSettingForLayer form={form} index={index} />
+            </Col>
+          </Row>
         </Col>
       </Row>
     </div>
--- a/frontend/javascripts/dashboard/dataset/dataset_settings_view.tsx
+++ b/frontend/javascripts/dashboard/dataset/dataset_settings_view.tsx
@@ -199,6 +199,7 @@ class DatasetSettingsView extends React.PureComponent<PropsWithFormAndRouter, St
         this.props.datasetId,
       );
       enforceValidatedDatasetViewConfiguration(datasetDefaultConfiguration, dataset, true);
+      // TODOM: extract x, y, z rotation from datasetDefaultConfiguration
       form.setFieldsValue({
         defaultConfiguration: datasetDefaultConfiguration,
         defaultConfigurationLayersJson: JSON.stringify(
--- a/frontend/javascripts/dashboard/dataset/dataset_settings_viewconfig_tab.tsx
+++ b/frontend/javascripts/dashboard/dataset/dataset_settings_viewconfig_tab.tsx
@@ -329,66 +329,6 @@ export default function DatasetSettingsViewConfigTab(props: {
           />
         </Col>
       </Row>
-      <Row gutter={32}>
-        <Col span={12}>
-          Permanent dataset rotation:
-          <Row gutter={24}>
-            <Col span={16}>
-              <FormItemWithInfo
-                name={["xAxisRotation"]}
-                label="X Axis Rotation"
-                info="Change the datasets rotation around the x-axis."
-                colon={false}
-              >
-                <Slider min={0} max={100} step={1} />
-              </FormItemWithInfo>
-            </Col>
-            <Col span={8} style={{ marginRight: -12 }}>
-              <FormItem
-                name={["defaultConfiguration", "segmentationPatternOpacity"]}
-                colon={false}
-                label=" "
-              >
-                <InputNumber min={0} max={100} step={1} precision={0} />
-              </FormItem>
-            </Col>
-          </Row>
-          <Row gutter={24}>
-            <Col span={16}>
-              <FormItemWithInfo
-                name={["yAxisRotation"]}
-                label="Y Axis Rotation"
-                info="Change the datasets rotation around the y-axis."
-                colon={false}
-              >
-                <Slider min={0} max={100} step={1} />
-              </FormItemWithInfo>
-            </Col>
-            <Col span={8} style={{ marginRight: -12 }}>
-              <FormItem name={["yAxisRotation"]} colon={false} label=" ">
-                <InputNumber min={0} max={100} step={1} precision={0} />
-              </FormItem>
-            </Col>
-          </Row>{" "}
-          <Row gutter={24}>
-            <Col span={16}>
-              <FormItemWithInfo
-                name={["zAxisRotation"]}
-                label="Z Axis Rotation"
-                info="Change the datasets rotation around the z-axis."
-                colon={false}
-              >
-                <Slider min={0} max={100} step={1} />
-              </FormItemWithInfo>
-            </Col>
-            <Col span={8} style={{ marginRight: -12 }}>
-              <FormItem name={["zAxisRotation"]} colon={false} label=" ">
-                <InputNumber min={0} max={100} step={1} precision={0} />
-              </FormItem>
-            </Col>
-          </Row>
-        </Col>
-      </Row>
     </div>
   );
 }
--- a/frontend/javascripts/dashboard/dataset/dataset_settings_viewconfig_tab.tsx
+++ b/frontend/javascripts/dashboard/dataset/dataset_settings_viewconfig_tab.tsx
@@ -30,7 +30,6 @@ import ColorLayerOrderingTable from "./color_layer_ordering_component";
 import type { APIDatasetId } from "types/api_flow_types";
 import { getAgglomeratesForDatasetLayer, getMappingsForDatasetLayer } from "admin/admin_rest_api";
 import { Slider } from "components/slider";
-
 const FormItem = Form.Item;
 
 export default function DatasetSettingsViewConfigTab(props: {
@@ -223,8 +222,8 @@ export default function DatasetSettingsViewConfigTab(props: {
         <Col span={6}>
           <FormItemWithInfo
             name={["defaultConfiguration", "rotation"]}
-            label="Rotation"
-            info="The default rotation that will be used in oblique and arbitrary view mode."
+            label="Rotation - Arbitrary View Modes"
+            info="The default rotation that will be used in oblique and flight view mode."
           >
             <Vector3Input />
           </FormItemWithInfo>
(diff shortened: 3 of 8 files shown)


MichaelBuessemeyer (commented):
#rotate

@philippotto This PR is ready for the first review round. To you have time to take on this task or should I ask daniel? Because I think you are already packed with the Unified Annotation Versioning but as you built the coordinateTransforms & multi modality feature you might be a little more suited for the review :thinking:

philippotto:
> @philippotto This PR is ready for the first review round. To you have time to take on this task or should I ask daniel?

Yes, I can review it :)

frontend/javascripts/oxalis/model/accessors/dataset_layer_rotation_accessor.ts
```diff
@@ -0,0 +1,414 @@
...
+  const sinOfAngle = matrix[sinusLocation[0]][sinusLocation[1]];
+  const cosOfAngle = matrix[cosineLocation[0]][cosineLocation[1]];
+  const rotation =
+    Math.abs(cosOfAngle) > 1e-6 // Avoid division by zero
+      ? Math.atan2(sinOfAngle, cosOfAngle)
+      : sinOfAngle > 0
+        ? Math.PI / 2
+        : -Math.PI / 2;
+  const rotationInDegrees = rotation * (180 / Math.PI);
+  // Round to multiple of 90 degrees and keep the result positive.
```
philippotto, line 87: can we encode the 90deg rounding behavior in the function name? I haven't read enough to understand when the function is called, but the function seems bit dangerous when used for the wrong scenarios (i.e., a matrix with arbitrary rotations).

frontend/javascripts/oxalis/model/accessors/dataset_layer_rotation_accessor.ts
```diff
@@ -0,0 +1,414 @@
...
+
+export function getTranslationToOrigin(bbox: BoundingBox): AffineTransformation {
+  const center = bbox.getCenter();
+  const translationMatrix = new THREE.Matrix4()
+    .makeTranslation(-center[0], -center[1], -center[2])
+    .transpose(); // Column-major to row-major
+  return { type: "affine", matrix: flatToNestedMatrix(translationMatrix.toArray()) };
+}
+
+export function getTranslationBackToOriginalPosition(bbox: BoundingBox): AffineTransformation {
```
philippotto, line 100: I'd use "fromCenterToOrigin" and "fromOriginToCenter" in the two functions above.

frontend/javascripts/oxalis/model/accessors/dataset_layer_rotation_accessor.ts
```diff
@@ -0,0 +1,414 @@
...
+  layer: APIDataLayer | APISkeletonLayer,
+  nativelyRenderedLayerName: string | null,
+): Transform | null {
+  if (isLayerWithoutTransformationConfigSupport(layer)) {
+    return getTransformsForLayerWithoutTransformationConfigOrNull(
+      dataset,
+      nativelyRenderedLayerName,
+    );
+  }
+  const layerTransforms = getOriginalTransformsForLayerOrNull(dataset, layer as APIDataLayer);
```
philippotto, line 196: can be moved below the following if-body.

frontend/javascripts/oxalis/model/accessors/dataset_layer_rotation_accessor.ts
```diff
@@ -0,0 +1,414 @@
...
+  layer: APIDataLayer | APISkeletonLayer,
+  nativelyRenderedLayerName: string | null,
+): Transform {
+  return (
+    getTransformsForLayerOrNull(dataset, layer, nativelyRenderedLayerName) || IdentityTransform
+  );
+}
+
+export function isIdentityTransform(transform: Transform) {
+  return transform.type === "affine" && _.isEqual(transform.affineMatrix, Identity4x4);
```
philippotto, line 234: we don't need to use Number.EPSILON when comparing affineMatrix to the identity matrix?

frontend/javascripts/oxalis/model/accessors/dataset_layer_rotation_accessor.ts
```diff
@@ -0,0 +1,414 @@
...
+}
+
+function _getTransformsForLayerWithoutTransformationConfigOrNull(
```
philippotto, line 237: at first, I interpreted this as "get transforms for a layer but ignore the transformation config". maybe rename to "_getTransformsForLayerThatDoesNotSupportTransformationConfigOrNull". it's a mouthful, but I'd rather use that to avoid confusion.

frontend/javascripts/oxalis/model/accessors/dataset_layer_rotation_accessor.ts
```diff
@@ -0,0 +1,414 @@
...
+    return false;
+  }
+  const threeMatrix = new THREE.Matrix4()
+    .fromArray(nestedToFlatMatrix(transformation.matrix))
+    .transpose();
+  threeMatrix.decompose(translation, quaternion, scale);
+  return (
+    translation.length() !== 0 &&
+    scale.equals(NON_SCALED_VECTOR) &&
+    quaternion.equals(new THREE.Quaternion())
```
philippotto, line 352: how about extracting new THREE.Quaternion() into a top-level constant?

frontend/javascripts/oxalis/model/accessors/dataset_layer_rotation_accessor.ts
```diff
@@ -0,0 +1,414 @@
...
+
+function isRotationOnly(transformation?: AffineTransformation) {
+  if (!transformation) {
+    return false;
+  }
+  const threeMatrix = new THREE.Matrix4()
+    .fromArray(nestedToFlatMatrix(transformation.matrix))
+    .transpose();
+  threeMatrix.decompose(translation, quaternion, scale);
+  return translation.length() === 0 && scale.equals(NON_SCALED_VECTOR);
```
philippotto, line 364: I think, an identity matrix would be categorized as rotationOnly, but not as translationOnly. this should be consistent.

frontend/javascripts/oxalis/model/accessors/dataset_layer_rotation_accessor.ts
```diff
@@ -0,0 +1,414 @@
...
+}
+
+/* This function checks if all layers have the same transformation settings that represent
+ * a translation to the dataset center and a rotation around each axis and a translation back.
+ * All together this makes 5 affine transformation matrices. */
+function _doAllLayersHaveTheSameRotation(dataLayers: Array<APIDataLayer>): boolean {
```
philippotto, line 370: maybe we can discuss this in person. I don't understand how this fits into the bigger picture yet.

philippotto (commented):
cool stuff! I already left some couple, but I'm not done with all files yet.

MichaelBuessemeyer:
@philippotto my newest changes should implement discussed changes and also fix the new position calculation after toggling which layer is rendered natively. Please have another look at my changes :pray:

philippotto (commented):
the code looks good :100: I'll test once the CI is green. currently, it fails due to cyclic dependencies :see_no_evil:

frontend/javascripts/oxalis/model/accessors/flycam_accessor.ts
```diff
@@ -253,6 +258,36 @@ function getMaximumZoomForAllMagsFromStore(state: OxalisState, layerName: string
...
+  const newSecondPosition = transformPointUnscaled(changeInAppliedTransformation)(secondPosition);
+
+  const scaleChange = _.mean(
+    // Only consider XY for now to determine the zoom change (by slicing from 0 to 2)
+    V3.abs(V3.divide3(V3.sub(newPosition, newSecondPosition), referenceOffset)).slice(0, 2),
+  );
+
+  return { newPosition, scaleChange };
+}
+
```
MichaelBuessemeyer, line 290: I moved the function leading to the newly introduced cyclic dependency here. Please feel free to argue with my decision.

The problem here imo is, that this function relies on camera and transformation information. Therefore, It has to be put into the only react component that uses this function right now, or into one of the accessors or into a completely new file. Putting the function into the `dataset_layer_transformation_accessor` created the cyclic dependency as the flycam already imports this module. Therefore I'd say putting it into `dataset_layer_transformation_accessor` is not a possible solution. 

If you'd prefer to have this function in the UI component( `layer_settings_tab.tsx`)  please just tell me. But this would make the already > 1.5k lines file even bigger :see_no_evil:

MichaelBuessemeyer (commented):
The Ci should hopefully be green now. I resolved the cyclic dependency at least.

Please see my comment regarding this

MichaelBuessemeyer:
From slack: https://scm.slack.com/archives/C5AKLAV0B/p1736434462796459
> i just tested the rotation feature a bit and i noticed that e.g. the hovering of segment ids does not work if the segmentation layer is rotated. this was probably already broken independently of your PR (but i didn't notice it because the previous feature was not used in this context). you would probably have to consider the rotation matrix when looking up from mouse position to voxel position

Should be fixed now with commit 3f1bdb9e03063274f06824126c4a0f313e7c9bf9

philippotto:
> From slack: https://scm.slack.com/archives/C5AKLAV0B/p1736434462796459
> 
> > i just tested the rotation feature a bit and i noticed that e.g. the hovering of segment ids does not work if the segmentation layer is rotated. this was probably already broken independently of your PR (but i didn't notice it because the previous feature was not used in this context). you would probably have to consider the rotation matrix when looking up from mouse position to voxel position
> 
> Should be fixed now with commit [3f1bdb9](https://github.com/scalableminds/webknossos/commit/3f1bdb9e03063274f06824126c4a0f313e7c9bf9)

Hmm, still doesn't work for me? I tested it here: https://allowpermadatasetrotation.webknossos.xyz/datasets/l4_sample-677e47bc010000b800249ffc/view#3584,3584,1024,0,1.3

MichaelBuessemeyer:
> Hmm, still doesn't work for me? I tested it here: https://allowpermadatasetrotation.webknossos.xyz/datasets/l4_sample-677e47bc010000b800249ffc/view#3584,3584,1024,0,1.3

Oh well. Thanks for finding this. I still had a little logic bug (I applied the segmentation layer's transform instead of the inverse of the transform). The newest commit [3c06db5](https://github.com/scalableminds/webknossos/pull/8159/commits/3c06db545e091fba67ab013da9b20082b21e9551) should fix this. :tm: 

Testing for me worked because I only had a 180 degree rotation for x configured and in that case applying the same rotation again instead of the inverse seems to result in the correct behaviour. As your test case was more complex and it worked for me locally with these settings, this should hopefully be fixed now :)

philippotto:
Great! Unfortunately, the context menu uses the wrong position apparently (e.g., load mesh or the segment id in the context menu footer is wrong). Can you have another look? :see_no_evil:

MichaelBuessemeyer:
While working on this I also noticed that bounding boxes are not rotated :see_no_evil:? Should this be included here as well or in a follow up pr?

MichaelBuessemeyer:
Same for meshes :see_no_evil:

philippotto:
Bummer! The layer bboxes are able to transform according to the active layer transform. user bounding boxes probably need to use the same transforms as the skeleton layer. meshes should use the same transforms as their segmentation layer.
however, I would suggest to do this later as a follow-up pr, as you have other stuff on our plate, too. can you open an issue? :)

MichaelBuessemeyer:
The issue is here: https://github.com/scalableminds/webknossos/issues/8340

And I found a but that toggling the skeleton layer crashed wk. I fixed this by disabling toggling skeleton layer / letting the user configure them as the active native layer. The mini discussion about this can be found here: https://scm.slack.com/archives/C083DBRCFC6/p1737378159151899

MichaelBuessemeyer:
Moreover, the newest commits should hopefully fix all code locations that try to get the segmentation information based on a global position. 

Please give this another PR testing try and check my newest changes :pray: 

I am sorry, that this PR takes so many iterations.

philippotto (approved):
Great! I found some other stuff but commented that in the follow-up issue, because it's not caused by this PR. let's merge and iterate from there :)

PR #8170, opened 2024-11-04
Some Timetracking view improvements

MichaelBuessemeyer:
- https://timetrackingfollowup.webknossos.xyz

### Steps to test:
- 
### TODOs:
- [x] Add option to filter by annotation state (active, finished/archived)
- [x] Include annotation state in table and CSVs
- [x] long project names are not fully visible (dropdown menu is very narrow)

### Issues:
- fixes [#7788](https://github.com/scalableminds/webknossos/issues/7788)

------

## Release Notes

- **New Features**
  - Introduced `AnnotationStateFilterEnum` for filtering time tracking data based on annotation states (ALL, ACTIVE, FINISHED_OR_ARCHIVED).
  - Enhanced time tracking API functions to support filtering by annotation state.

- **Improvements**
  - Updated dropdown components to allow users to select annotation states for better data filtering.
  - Modified CSV export functionality to include annotation state in time tracking reports.
  - Enhanced data retrieval logic to incorporate annotation states, improving the granularity of time tracking data.

- **Bug Fixes**
  - Improved data retrieval logic to ensure accurate representation of annotation states in time tracking views.


MichaelBuessemeyer:
@coderabbitai full review

frontend/javascripts/admin/statistic/project_and_annotation_type_dropdown.tsx
```diff
@@ -108,6 +134,7 @@ function ProjectAndAnnotationTypeDropdown({
       options={filterOptions}
       optionFilterProp="label"
       value={selectedFilters}
+      popupMatchSelectWidth={false}
```
MichaelBuessemeyer, line 137: This prop enables fully showing long project names as stated in the issue. 

But the antd docs to this prop state:
> Determine whether the popup menu and the select input are the same width. Default set min-width same as input. Will ignore when value less than select width. false will disable virtual scroll.

Where the last sentence is important IMO -> virtual scrolling is now disabled by this change. This means that for an organization with many projects, that the list might be slow.

Do you think this is acceptable?

frontend/javascripts/admin/admin_rest_api.ts
```diff
@@ -1760,13 +1774,19 @@ export async function getTimeEntries(
...
   const params = new URLSearchParams({
     start: startMs.toString(),
     end: endMs.toString(),
     annotationTypes: selectedTypes,
   });
+  if (selectedState !== AnnotationStateFilterEnum.ALL) {
+    params.append("annotationStates", selectedState);
+  } else {
+    params.append("annotationStates", "Active,Finished");
+  }
```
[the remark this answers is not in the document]
MichaelBuessemeyer: nice, thanks for saving me here :pray:

MichaelBuessemeyer:
@daniel-wer Do you have time to review this?

daniel-wer (commented):
Well done, I tried to test this to the best of my knowledge, but maybe @fm3 wants to do a quick test as well? :)

fm3 (commented):
Works for me 🎉

frontend/javascripts/admin/statistic/project_and_annotation_type_dropdown.tsx
MichaelBuessemeyer: > An alternative could be to set a reasonable width (somewhat larger than before) and then show the full project name on hover. No real preference on my side.

That's definitely also doable :)

I'll switch to this behavior :+1:

frontend/javascripts/admin/admin_rest_api.ts
```diff
@@ -1729,7 +1733,11 @@ export async function getTimeTrackingForUserSummedPerAnnotation(
   if (annotationTypes != null) params.append("annotationTypes", annotationTypes);
   if (projectIds != null && projectIds.length > 0)
     params.append("projectIds", projectIds.join(","));
-  params.append("annotationStates", "Active,Finished");
+  if (annotationState !== "All") {
```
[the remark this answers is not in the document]
MichaelBuessemeyer: Yeah, I also tried this one, but this means that a non type / full import of `AnnotationStateFilterEnum` is needed, which creates a new cyclic dependency :see_no_evil:.

The only think I could do it move the definition of `AnnotationStateFilterEnum` to some other file like `constants.ts`. What do you think?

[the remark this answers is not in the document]
daniel-wer: Yes, I'd move it to constants then and use it here. Strictly speaking constants shouldn't contain any types, but that ship has sailed :sweat_smile: This can be cleaned up later, together with the other types in there.

MichaelBuessemeyer:
> An alternative could be to set a reasonable width (somewhat larger than before) and then show the full project name on hover.

Yeah, I also thought about this and implemented it now this way :). There is already a title for each option in the filter select dropdown. I only adjusted the width of the dropdown to be at 800 px. If you think that's too much just tell me.

This is how it looks with a long name (the is also a title available for each dropdown menu item, but its not included in the screenshot):
image
And here with no long project name (the width of the dropdown is still at 800px due to the virtualization feature.
image

frontend/javascripts/admin/statistic/project_and_annotation_type_dropdown.tsx
```diff
@@ -134,7 +124,7 @@ function ProjectAndAnnotationTypeDropdown({
       options={filterOptions}
       optionFilterProp="label"
       value={selectedFilters}
-      popupMatchSelectWidth={false}
+      popupMatchSelectWidth={800}
```
daniel-wer, line 127:
```suggestion
      popupMatchSelectWidth={400}
```

daniel-wer (approved):
Nice! I'd reduce the width, but feel free to ignore.

frontend/javascripts/admin/statistic/project_and_annotation_type_dropdown.tsx
MichaelBuessemeyer: sure sounds good :)

PR #8173, opened 2024-11-06
Change long unit name in frontend from ångström to angstrom

fm3:
The backend uses "angstrom", which is consistent with ngff, compare https://ngff.openmicroscopy.org/latest/#axes-md – so the frontend needs to send the same.

- https://angstrom.webknossos.xyz

### Steps to test:
- Attempt to upload dataset with Å as unit, should work

### Issues:
- fixes https://scm.slack.com/archives/CMBMU5684/p1730907197948829

------


MichaelBuessemeyer:
@coderabbitai pause

valentin-pinkau (approved):
🤓

PR #8176, opened 2024-11-07
Fix performance bottleneck when deleting a lot of trees at once

daniel-wer:
Also fix a bug when importing an NML with groups when only groups but no trees exist in an annotation.
Also fix a bug where trying to delete a non-existing node (via the API, for example) would delete the whole active tree (at least that's what it looked like to me when reading the code).

- https://___.webknossos.xyz

### Steps to test:
- Create an annotation with more than 10,000 trees and delete them all at once. This should take less than a second and not multiple minutes.
- Also test other tree deletion related functionality:
  - Tree multi-select and deletion
  - Group deletion with and without children
  - Single tree deletion
- Import an NML with groups. Delete all but one group and import the NML again. On master, the newly imported group gets the same ID as the one that was not deleted before, leading to all sorts of issues. If only one of the groups gets the blueish selected background, the fix works.

See this gif for deleting 50,000 trees at once:
batched_tree_deletion

### Issues:
- fixes performance bottleneck

------

## Release Notes

- **New Features**
  - Enhanced efficiency for deleting multiple trees through batch actions.
  - Improved user interface interactions by ensuring context menus close after delete actions.

- **Bug Fixes**
  - Resolved a performance bottleneck when deleting numerous trees simultaneously.
  - Fixed issues related to importing NML files with groups and deleting non-existing nodes via the API.

- **Performance Improvements**
  - Asynchronous image file reading for better performance.
  - Improved error handling for clearer context in error messages.

These updates significantly enhance user experience and system performance.

### DIFF
--- a/CHANGELOG.released.md
+++ b/CHANGELOG.released.md
@@ -7,6 +7,56 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Calendar Versioning](http://calver.org/) `0Y.0M.MICRO`.
 For upgrade instructions, please check the [migration guide](MIGRATIONS.released.md).
 
+## [24.11.1](https://github.com/scalableminds/webknossos/releases/tag/24.11.1) - 2024-11-13
+[Commits](https://github.com/scalableminds/webknossos/compare/24.10.0...24.11.1)
+
+### Highlights
+- It is now possible to add metadata in annotations to Trees and Segments. [#7875](https://github.com/scalableminds/webknossos/pull/7875)
+- Added a button to the search popover in the skeleton and segment tab to select all matching non-group results. [#8123](https://github.com/scalableminds/webknossos/pull/8123)
+
+### Added
+- It is now possible to add metadata in annotations to Trees and Segments. [#7875](https://github.com/scalableminds/webknossos/pull/7875)
+- Added a summary row to the time tracking overview, where times and annotations/tasks are summed. [#8092](https://github.com/scalableminds/webknossos/pull/8092)
+- Most sliders have been improved: Wheeling above a slider now changes its value and double-clicking its knob resets it to its default value. [#8095](https://github.com/scalableminds/webknossos/pull/8095)
+- It is now possible to search for unnamed segments with the full default name instead of only their id. [#8133](https://github.com/scalableminds/webknossos/pull/8133)
+- Increased loading speed for precomputed meshes. [#8110](https://github.com/scalableminds/webknossos/pull/8110)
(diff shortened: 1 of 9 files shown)


hotzenklotz (approved):
Code LGTM. Haven't tested yet.

MichaelBuessemeyer (changes requested):
Thanks a lot for all these awesome fixes :tada:

> Import an NML with groups. Delete all but one group and import the NML again. On master, the newly imported group gets the same ID as the one that was not deleted before, leading to all sorts of issues. If only one of the groups gets the blueish selected background, the fix works.

Nice ;fix works :tada:

One unrelated thing I noticed is that under linux (ubuntu + chrome) the tree context menu does not close after clicking on delete or shuffle tree color. I would have expected for the context menu to close at least after the delete action was clicked, because the tree for with the context menu was opened does not exist anymore. Thus, the context menu points to a deleted tree and imo should be closed directly when hitting the deletec menu item.
Do you think you can also take a look at this in this pr or do you want to do this as a follow-up?

Moreover:
> Also fix a bug where trying to delete a non-existing node (via the API, for example) would delete the whole active tree (at least that's what it looked like to me when reading the code).

While testing this with 
[bug_test.nml.txt](https://github.com/user-attachments/files/17676736/bug_test.nml.txt)
(and having tree with id 2 (the blue one) active), I get the following error in the console:
image

Here is my mini test API script:  
```javascript
window.webknossos.apiReady(3).then(api => {
    const activeTreeId = api.tracing.getActiveTreeId();
    api.tracing.deleteNode(100, activeTreeId);
  });
```
Could you maybe give this a little try to avoid the `UnhandledRejection: "Error: Get empty"`?

At least, I hope that me trying to reproduce what you meant with was correct
> Also fix a bug where trying to delete a non-existing node (via the API, for example) would delete the whole active tree (at least that's what it looked like to me when reading the code).

MichaelBuessemeyer:
Oh an the rest of the deletion fixes worked perfectly. Thanks a lot :fire: :racing_car:

daniel-wer:
Thanks for the review @MichaelBuessemeyer!

> One unrelated thing I noticed is that under linux (ubuntu + chrome) the tree context menu does not close after clicking on delete or shuffle tree color.

Fixed :+1: 

> Could you maybe give this a little try to avoid the UnhandledRejection: "Error: Get empty"?

Thank you for testing this! Looks like the "bug" I fixed couldn't actually be triggered through the api. Still, I think the code is more explicit and correct now, so I'd leave it as-is. I adapted the code to avoid the thrown error in case of a non-existing node id. This is in line with the other accessor functions and in the spirit of the used Maybe pattern I'd argue.

MichaelBuessemeyer (approved):
Awesome, thanks for fixing the nitpicks :pray:

PR #8181, opened 2024-11-11
Improve error messages for starting jobs on other orgs datasets

### LINKED ISSUE #8165: Misleading error message when starting jobs for datasets of other organizations
Starting processing jobs for datasets of other organizations is not currently allowed, even if the datasets are public or viewed by a valid sharing token.

While fully allowing this for some jobs/situations is a topic for another day (also compare discussion at https://github.com/scalableminds/webknossos/issues/7360 ), what can be improved now is the error message.

Some jobs already have a readable error message, like “Currently nuclei inferral is only allowed for datasets of your own organization.”. However, some, like render_animation, just show “Organization b2275d664e4c2a96 could not be found”. This should be unified, e.g. by passing GlobalAccessContext to `organizationDAO.findOne`, and then checking `request.identity._organization == organization._id` with a suitable message.

The same should be done for runTraining and runInference in AiModelController. Looks like the frontend does not currently send the organizationId there at all, which leads to unusable 404 errors if tried on another orga’s dataset

frcroth:
- https://___.webknossos.xyz

### Steps to test:
- Make a dataset public
- Have a user of a different org access the dataset (user can be dataset manager and admin)
- Try AI Analysis
	- Not sure ow to test infer Mitochondria since this is marked as "Not yet available"
- Try running AI animation: This does not result in any error, but also does not show a popup interface. Not sure why (#8189)

~~As I understand the runInference Job, it is not necessary there because the organization is taken from the user's identity anyway.~~

### Issues:
- fixes #8165

------

## Release Notes

- **New Features**
  - Transitioned to asynchronous reading of image files on the datastore filesystem for improved efficiency.
  - Added `organizationId` to job parameters, enhancing context in job processing.

- **Bug Fixes**
  - Improved error messages related to job initiation on datasets from other organizations, providing clearer feedback.
  - Fixed performance issues when deleting multiple trees and bugs associated with importing NML files and deleting non-existing nodes.

- **Improvements**
  - Enhanced error handling and validation logic across job-related actions, ensuring users receive accurate and actionable information.
  - Updated error messages for various job permissions, clarifying restrictions based on dataset ownership.


conf/messages
```diff
@@ -329,10 +330,11 @@ job.globalizeFloodfill.notAllowed.organization = Globalizing floodfills is only
 job.materializeVolumeAnnotation.notAllowed.organization = Materializing volume annotations is only allowed for datasets of your own organization.
 job.noWorkerForDatastoreAndJob = No webknossos-worker supporting the requested job is available for the selected datastore.
 job.emailNotifactionsDisabled = Email notifications are not enabled for this job type.
-job.renderAnimation.notAllowed.organization = "Rendering animations is only allowed for datasets of your own organization."
-job.alignSections.notAllowed.organization = "Aligning sections is only allowed for datasets of your own organization."
-job.alignSections.notAllowed.onlySuperUsers = "For now, aligning sections is only allowed for super users."
-job.additionalCoordinates.invalid = "The passed additional coordinates are invalid."
+job.renderAnimation.notAllowed.organization = Rendering animations is only allowed for datasets of your own organization.
+job.alignSections.notAllowed.organization = Aligning sections is only allowed for datasets of your own organization.
+job.alignSections.notAllowed.onlySuperUsers = For now, aligning sections is only allowed for super users.
```
fm3, line 335: I think this one can go, we lifted that restriction

fm3 (commented):
Looking good!

> As I understand the runInference Job, it is not necessary there because the organization is taken from the user's identity anyway.

Yes, not sure why, to be honest. I think it would be better to change the RunInferenceParameters to include the organizationId. Could you maybe do this here? Not sure how simple adapting the frontend is, but I would hope it can be done the same way as for runTraining?

Also, could you write an issue detailing that the renderAnimation modal does not show up, as you mentioned in person? I think that there, too, we should have either an explanation or the menu item shouldn’t be there in the first place.

fm3 (approved):

PR #8184, opened 2024-11-11
Add dataset upload test (without using frontend)

### LINKED ISSUE #8099: e2e tests for dataset uploads
dataset uploads are a critical feature. We should discuss how we can add end-to-end tests for it, and which cases to cover.

frcroth:
### Steps to test:
- CI

### Issues:
- fixes #8099 

## Summary by CodeRabbit

- **New Features**
	- Introduced a new test case for dataset upload functionality, enhancing the testing suite.
	- Improved the import structure for handling HTTP requests with `FormData` and `File`.

- **Bug Fixes**
	- Enhanced URL validation logic in the `fetchWrapper` function to ensure correct protocol handling.

- **Refactor**
	- Modified logic in the `ensureTestDataset` method to delete and recreate the data directory, improving test setup reliability.


fm3 (commented):
LGTM, but I’m not very familiar with typescript or the existing e2e tests. @daniel-wer could you maybe also have a quick look?

daniel-wer (commented):
Sorry for the delayed review. Great addition to our tests, thank you! :pray: 

Why did you choose to implement this without any frontend api code (for example, the `reserveDatasetUpload` function from the `admin_rest_api`)? I can imagine the resumable-js lib to be hard to work with in the tests, but was that the reason?
If it was by choice, do you think it would be worthwhile and possible (with reasonable effort) to add a similar test in a followup issue, testing the frontend functionality as well?

frcroth:
I have tried to test the frontend. However it is just not easy to do. The resumable JS library uses Browser APIs e.g. File and XMLHttpRequest that are not available in the node test environment. I tried to mock them e.g. using JSDom, but it failed at sending the Multipart Formdata correctly (compare https://github.com/jsdom/jsdom/issues/2300). So maybe it would be possible to do it but that way was tried and failed.
I also looked into pupetter, but that did not seem easy either and would probably slow down the test suite significantly.

daniel-wer (approved):
Thanks for elaborating, that's understandable and good to have documented as part of this PR :+1:

PR #8186, opened 2024-11-12
Fix: Use forwardRef for scrollable virtualized tree

MichaelBuessemeyer:
The newly deployed scrollable fix added a wrapper for the antd tree implementation. The wrapper is functional component which needs to be wrapped in a `forwardRef` as it received a ref which should be forwarded to the antd tree component. The `forwardRef` was missing and is now added in this pr. 

- https://___.webknossos.xyz

### Steps to test:
- On master in local dev open a tracing with some skeletons and let the skeleton tab render by viewing it. The console should show an error something like: 
```
Warning: Function components cannot be given refs. Attempts to access this ref will fail. Did you mean to use React.forwardRef()?

Check the render method of `AutoSizer`. Error Component Stack
...
```
- Check out this branch and do the same. The error should disappear / not be printed.

### Issues:
- fixes https://scm.slack.com/archives/C02H5T8Q08P/p1731403785293789
- follow up fix for #8162

------

### DIFF
--- a/frontend/javascripts/oxalis/view/right-border-tabs/scrollable_virtualized_tree.tsx
+++ b/frontend/javascripts/oxalis/view/right-border-tabs/scrollable_virtualized_tree.tsx
@@ -1,7 +1,7 @@
 import { Tree as AntdTree, type TreeProps } from "antd";
 import type { BasicDataNode } from "antd/es/tree";
 import { throttle } from "lodash";
-import { useCallback, useRef } from "react";
+import { forwardRef, useCallback, useRef } from "react";
 import type RcTree from "rc-tree";
 
 const MIN_SCROLL_SPEED = 30;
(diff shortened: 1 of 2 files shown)


frontend/javascripts/oxalis/view/right-border-tabs/segments_tab/segments_view.tsx
```diff
@@ -1904,7 +1904,7 @@ class SegmentsView extends React.Component<Props, State> {
                                 overflow: "hidden",
                               }}
                             >
-                              <ScrollableVirtualizedTree<SegmentHierarchyNode>
+                              <ScrollableVirtualizedTree
```
MichaelBuessemeyer, line 1907: Can now be removed as the type casting is results in a "better" / "more understandable" typing to ts -> Explicitly telling the generic type is not needed anymore.

frontend/javascripts/oxalis/view/right-border-tabs/scrollable_virtualized_tree.tsx
```diff
@@ -56,9 +58,15 @@ function ScrollableVirtualizedTree<T extends BasicDataNode>(
...
     </div>
   );
 }
 
+const ScrollableVirtualizedTree = forwardRef(ScrollableVirtualizedTreeInner) as <
+  T extends BasicDataNode,
+>(
+  props: TreeProps<T> & { ref?: React.Ref<RcTree> },
+) => ReturnType<typeof ScrollableVirtualizedTreeInner>;
+
```
MichaelBuessemeyer, line 71: This way of using `forwardRef` was suggested by AI overlord as `forwardRef` does not seem to support generic typing

frontend/javascripts/oxalis/view/right-border-tabs/scrollable_virtualized_tree.tsx
```diff
@@ -56,9 +58,15 @@ function ScrollableVirtualizedTree<T extends BasicDataNode>(
 
   return (
     <div ref={wrapperRef}>
-      <AntdTree {...props} onDragOver={onDragOver} />
+      <AntdTree {...props} onDragOver={onDragOver} ref={ref} />
```
MichaelBuessemeyer, line 61: ref is now a separate arg and thus needs to be passed explicitly :man_shrugging:

daniel-wer (approved):
Thanks for fixing, LGTM :+1:

PR #8185, opened 2024-11-11
Send right parameter to createNodeSkeletonAction

dieknolle3333:
This PR undoes the changes to the `MutableNode` type, renaming `mag` to `resolution` again. The change in the opposite direction caused inconsistencies with the backend, the the slack thread below.
`mag` seems to work in every place except for the CreateNodeSkeletonAction, so its a bit sad this change is undone. This PR just tries to prevent that information is lost, maybe someone can find a better solution in the future.

### URL of deployed dev instance (used for testing):
- https://___.webknossos.xyz

### Steps to test:
- Create and save skeleton nodes. 

### TODOs:
- [ ] ...

### Issues:
- fixes https://scm.slack.com/archives/C5AKLAV0B/p1731340339911819

------
(Please delete unneeded items, merge only when none are left open)
- [x] Updated [changelog](../blob/master/CHANGELOG.unreleased.md#unreleased)
- [ ] Updated [migration guide](../blob/master/MIGRATIONS.unreleased.md#unreleased) if applicable
- [ ] Updated [documentation](../blob/master/docs) if applicable
- [ ] Adapted [wk-libs python client](https://github.com/scalableminds/webknossos-libs/tree/master/webknossos/webknossos/client) if relevant API parts change
- [ ] Removed dev-only changes like prints and application.conf edits
- [ ] Considered [common edge cases](../blob/master/.github/common_edge_cases.md)
- [ ] Needs datastore update after deployment

## Release Notes

- **New Features**
  - Added metadata support for annotations on Trees and Segments.
  - Introduced a summary row in the time tracking overview.
  - Enhanced slider functionality for value changes and reset options.
  - Improved search capabilities for unnamed segments.
  - Added a button to select all matching non-group results in the search popover.

- **Bug Fixes**
  - Resolved dataset upload issues and bbox export menu problems.
  - Fixed scrolling issues in the trees and segments tab.
  - Addressed permission-related issues with sharing tokens and layout persistence bugs.

- **Improvements**
  - Increased loading speeds for precomputed meshes.
  - Admins can now view and cancel all jobs, with job ownership displayed.
  - Enhanced error handling and messaging in proofreading actions.

PR #8187, opened 2024-11-12
Fix expected node type from server by renaming resolution to mag

MichaelBuessemeyer:
Fixes changes introduced by [#8111](https://github.com/scalableminds/webknossos/pull/8111). The server sends nodes with the field called mag and not resolution. This pr changes the frontend type to expect the name as now called mag.

- https://fixservernodetype.webknossos.xyz

### Steps to test:
- Create a tracing and add at least one tree with nodes.
- hard reload the tracing and download the nml. The `inMag` field should not be empty.
- Check on master: the `inMag` field should be empty

### Issues:
- fixes bug introduced by #8185

------

## Release Notes

- **New Features**
	- Enhanced search capabilities for unnamed segments.
	- Improved loading speed for precomputed meshes.
	- Unified terminology in the UI from "Resolution" to "Magnification" (mag).
	- Added a button in the search popover for selecting all matching non-group results.

- **Bug Fixes**
	- Resolved issues with dataset uploads, annotation saving, and scrolling in the trees and segments tab.
	- Fixed layout persistence bugs and ensured S3-compliant object storages are accessible via HTTPS.

- **Chores**
	- Migrated nightly screenshot tests from CircleCI to GitHub Actions.


frontend/javascripts/types/api_flow_types.ts
```diff
@@ -756,7 +756,7 @@ export type ServerNode = {
   rotation: Point3;
   bitDepth: number;
   viewport: number;
-  resolution: number;
+  mag: number;
```
MichaelBuessemeyer, line 759: Here is the root of the problem. The backend sends nodes with the field called mag and not resolution. See: https://github.com/scalableminds/webknossos/blob/8dec57826e253f3b1c2d17308e7d75d41fea26b1/webknossos-datastore/proto/SkeletonTracing.proto#L8-L19
By changing the type, TS should show all locations to change and the resulting code should be fine.

daniel-wer (approved):
Thank you for fixing this, works as described :+1:

PR #8192, opened 2024-11-13
Update default security.txt for 2025

frcroth:
Current info is stale.

<https://www.rfc-editor.org/rfc/rfc9116#name-expires>

Should be updated regularly


fm3 (approved):

frcroth:
> Should we create a yearly slack reminder?

Good idea, did this now.

PR #4439, opened 2020-02-12
Better onboarding on demo instance

fm3:
Lots of improvements for new users on webknossos.org:
- Signup on frontpage
- 1 organization per new user
- Public Datasets hidden in list (only on demo)
- Show featured publications in dashboard
- New user banner with information
- Better welcome email
- Improved info tab in annotation view
- Renamed tracing -> annotation
- Improved docs
- Refactored DatasetCache from Dashboard in own component with context

------
- [x] Updated [changelog](../blob/master/CHANGELOG.md#unreleased)
- ~[ ] Updated [migration guide](../blob/master/MIGRATIONS.md#unreleased) if applicable~
- [x] Updated [documentation](../blob/master/docs) if applicable
- ~[ ] Adapted [wk-connect](https://github.com/scalableminds/webknossos-connect) if datastore API changes~
- ~[ ] Needs datastore update after deployment~
- [x] Ready for review

PR #8193, opened 2024-11-13
Accept terms of Service at signup

### LINKED ISSUE #8149: Integrate Terms-Of-Service acceptance into signup form
Currently after creating an account you get another view asking you to accept the TOS. This could be unified for a more streamlined new user experience.

Needs an adapted form in the frontend and an additional optional boolean field “didAcceptTermsOfService” in the createOrganizationWithAdmin route.

This is only required if `webKnossos.termsOfService.enabled` is true in application.conf (Use /api/termsOfService )

We should also have a look at how the frontend determines if this is necessary or not, the protocol might have to change here too, since this info is now needed before the organization exists.

frcroth:
- https://___.webknossos.xyz

Accept terms of service already at sign up / organization creation.

Backend is done. Frontend needs to call /api/termsOfService route and add a checkbox / iframe to the account creation page. Then set "acceptedTermsOfService" in the submitted form to /api/auth/createOrganizationWithAdmin  to the TOS VERSION. The Terms of Service acceptance afterward can be removed.

### Steps to test:
- Register as a new user
- Try not to accept the Terms of Service, this should not be possible
- Switch `isWkOrgInstance` in application.conf, it should behave the same regarding the ToS
- Disable the ToS in application.conf, there should no longer be a ToS checkbox
- After having signed up, increase the ToS version in application.conf. The user should now see the ToS modal and have to accept the ToS

### TODOs:
- [x] Frontend
     - [x] wkorg
     - [x] generic
     - [x] styling
- [x] Remove dev-only changes to application.conf

### Issues:
- fixes #8149 

------


frcroth:
Backend should be done. @dieknolle3333 do you want to take a look at this for the frontend?

dieknolle3333:
sounds good!

dieknolle3333:
I feel like adding an iframe increases the probability that users actually read the ToS before accepting them, but then the privacy statement should also be an iframe. And two iframes either make the page quite long or are quite small, so users would have to scroll a lot to read them. Thats why I chose to add a link for now. If anybody disagrees with this decision, please let me know!
(pic below shows how it looks with the terms of service in an iframe, for illustrations purposes)
image

dieknolle3333:
this is how I built the frontend for now
image

dieknolle3333:
I will have a look

dieknolle3333:
@frcroth I implemented the org id generation similar to the org name generation now, and this id is sent to ` /api/auth/createOrganizationWithAdmin`. I also added the ToS modal again thanks to Toms [remark on slack](https://scm.slack.com/archives/C5AKLAV0B/p1731923682699469). I did not yet test the latter, will do that later today. I did notice that the registration of users works now, thanks to your latest changes! :tada:

MichaelBuessemeyer:
> @dieknolle3333 Thank you so much! One thing I noticed is that the frontend sends an "organization" value used as the org id. However, it is a random string, not the slug-name of the organization as I expected (from the changes from https://github.com/scalableminds/webknossos/pull/7386). So, if I create an account as "Monika Musterfrau" I would expect the organization value to be either empty (it will be handled by the backend then) or "monika-musterfrau-lab", not "3c98af371407d694"

@fm3 I need a little double-check here. I would have expected the frontend to not generate the orga id but letting the backend do this :thinking:.
Because I would imagine that the new implementation:
```javascript
function generateOrganizationId(firstName: string, lastName: string) {
  return `${firstName.toLowerCase()}-${lastName.toLowerCase()}-lab`;
}
```
can lead to naming collisions in case of using the same name when creating a orga. :thinking:.

frontend/javascripts/admin/auth/registration_form_wkorg.tsx
```diff
@@ -15,34 +17,34 @@ type Props = {
...
-  let output = "";
-
-  for (let i = 0; i < 8; i++) {
-    output += Math.floor(Math.random() * 255)
-      .toString(16)
-      .padStart(2, "0");
-  }
+function generateOrganizationId(firstName: string, lastName: string) {
+  return `${firstName.toLowerCase()}-${lastName.toLowerCase()}-lab`;
+}
```
MichaelBuessemeyer, line 22: I think this might create naming collisions. I'd like to double check this with @fm3

MichaelBuessemeyer: And this way of defining the orga id might lead to orga ids with whitespaces in it :thinking:

frontend/javascripts/admin/auth/registration_form_generic.tsx
```diff
@@ -274,28 +278,57 @@ function RegistrationFormGeneric(props: Props) {
...
+            <Checkbox>
+              I agree to storage and processing of my personal data as described in the{" "}
+              <a target="_blank" href="/privacy" rel="noopener noreferrer">
+                privacy statement
+              </a>
+              .
+            </Checkbox>
+          </FormItem>
+        )}
+        {terms != null && !terms.enabled ? null : (
```
MichaelBuessemeyer, line 304: Could you please invert the logic here? To me it does not read the intutive way I would have expected :see_no_evil:. Also I think the check box would be rendered when `terms == null` according my understanding of the current code

```suggestion
        {terms && terms.enabled ? (
            < check box component .../>
        ) : null }

```

Edit: Ah I see. Still rendering the checkbox even when `terms == null` is intentional :)

frontend/javascripts/admin/auth/registration_form_wkorg.tsx
```diff
@@ -155,32 +157,57 @@ function RegistrationFormWKOrg(props: Props) {
...
-          </a>
-          .
-        </Checkbox>
-      </FormItem>
-      <FormItem
-        style={{
-          marginBottom: 10,
-        }}
-      >
+        {terms != null && !terms.enabled ? null : (
```
MichaelBuessemeyer, line 182: I'd also prefer the inverted logic here

MichaelBuessemeyer (commented):
Thanks for this PR both of you :tada:

This is a code only review from my side. I'll test tomorrow :)

frontend/javascripts/admin/auth/registration_form_wkorg.tsx
fm3: Why change this in this PR at all? As I understand, the random-character string is supposed to be the typical organization id. The organization name on the other hand is supposed to be created from the user name. 

This was introduced in https://github.com/scalableminds/webknossos/pull/4439 (though organizationId was still called organizationName then, and what is now the name was then the displayName).

I agree that changing it to firstname-lastname-lab may lead to unwanted collisions.

MichaelBuessemeyer: To me it is also kinda strange, why the frontend generates this id string and not the backend :thinking:. Maybe this could be changed in the future?

> but it is indeed intended that new ones created by this form should have the random string.

@dieknolle3333 Could you please undo this change here? In case we agree to change the id generation, I'd do this in a separate PR. But I'd say that this is kinda unlikely because an id being a random hex string sounds fine to me. But feel free to open up a discussion on this :) (@frcroth)

frcroth: It is undone. I interpreted the PR and the issue so that the id should be human readable now.

MichaelBuessemeyer: Oh that indeed looks like a misunderstanding :see_no_evil: 

Thanks a lot for undoing the change so fast :) :pray:

MichaelBuessemeyer:
Oh and during testing I happend to notice, that the wkorg sign up form only has a single password input :thinking:?! The second one to prevent accidental mistyping is missing. Was this removed intentionally?
A small investigation showed that this has been the case for a long time. cf. pr https://github.com/scalableminds/webknossos/pull/5350/files where this was already the case.

But IMO there should be a second pwd field to prevent mistyping the password.

=> I opened an issue for this, so we can tackle this in another PR :D
https://github.com/scalableminds/webknossos/issues/8215

app/controllers/AuthenticationController.scala
```diff
@@ -621,6 +621,8 @@ class AuthenticationController @Inject()(
                     dataStoreToken <- bearerTokenAuthenticatorService.createAndInitDataStoreTokenForUser(user)
                     _ <- organizationService
                       .createOrganizationDirectory(organization._id, dataStoreToken) ?~> "organization.folderCreation.failed"
+                    _ <- Fox.runOptional(signUpData.acceptedTermsOfService)(version =>
+                      acceptTermsOfServiceForUser(user, version))
```
MichaelBuessemeyer, line 625: @frcroth I would prefer if the sign up request gets rejected in case `acceptedTermsOfService` is not set to true or empty but the application conf says that TOS is enabled.

What do you think?

MichaelBuessemeyer (changes requested):
Thanks you tow for this pr.

Testing went well :tada:.
Please take a look at my open comments. They should be rather small. Afterwards this should be ready to be merged :D

frontend/javascripts/admin/auth/registration_form_wkorg.tsx
fm3: > To me it is also kinda strange, why the frontend generates this id string and not the backend 🤔. Maybe this could be changed in the future?

Wrote #8216 to track that and future discussion of what the IDs should look like.

> I interpreted the PR and the issue so that the id should be human readable now.

Fair, but it was more of the opposite, really. Before #7386 there was a mongoId *and* this random string as organizationName, *and* the actual name (then called displayName), with spaces and all. In the PR, we dropped the mongoId and declared this random string the id, because it already had to be unique and was used as a key in many places.

Thanks to you all for clearing this up here :)

app/controllers/AuthenticationController.scala
```diff
@@ -639,8 +639,12 @@ class AuthenticationController @Inject()(
       )
   }
 
-  private def acceptTermsOfServiceForUser(user: User, termsOfServiceVersion: Int)(implicit m: MessagesProvider) =
-    organizationService.acceptTermsOfService(user._organization, termsOfServiceVersion)(DBAccessContext(Some(user)), m)
+  private def acceptTermsOfServiceForUser(user: User, termsOfServiceVersion: Option[Int])(
+      implicit m: MessagesProvider): Fox[Unit] =
+    for {
+      acceptedVersion <- Fox.option2Fox(termsOfServiceVersion)
```
MichaelBuessemeyer, line 645: While being nitpicky :see_no_evil: 
Could you please add an error message like

MichaelBuessemeyer, line 645: While being nitpicky 🙈
Could you please add an error message like

```suggestion
      acceptedVersion <- Fox.option2Fox(termsOfServiceVersion) ?~> "Terms of service must be accepted."
```

MichaelBuessemeyer: Else some kind of generic 400 error would be returned where an explanation would be useful IMO

frcroth: I can do this of course, no problem. Note that the user would not see 400 anyway because the frontend does not allow not accepting TOS.

frontend/javascripts/admin/auth/registration_form_generic.tsx
dieknolle3333: exactly. in case the ToS are still being fetched for whatever reason, I dont want it to seem that there are none

frontend/javascripts/admin/auth/registration_form_wkorg.tsx
dieknolle3333: youre so right, I had difficulty understanding it myself right now :sweat_smile:

conf/application.conf
```diff
@@ -116,8 +116,8 @@ webKnossos {
     Please add the information of the operator to comply with GDPR.
   """
   termsOfService {
-    enabled = false
-    # The URL will be embedded into an iFrame
+    enabled = true
```
dieknolle3333, line 119: just to be safe, is this change intentional? @frcroth

dieknolle3333:
from my perspective, this PR can be rereviewed :) I left the application.conf edits for testing purposes.

conf/application.conf
```diff
@@ -117,7 +117,7 @@ webKnossos {
   """
   termsOfService {
     enabled = false
```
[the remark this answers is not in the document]
frcroth: No, terms of service need to be generally disabled. In specific deployed instances it can be enabled.

frontend/javascripts/admin/auth/registration_form_wkorg.tsx
```diff
@@ -155,32 +159,57 @@ function RegistrationFormWKOrg(props: Props) {
...
+                  terms of service
+                </a>
+              )}
+              .
+            </Checkbox>
+          </FormItem>
+        ) : null}
+      </div>
+
+      <FormItem>
```
MichaelBuessemeyer, line 212: Hmmm :thinking: 

This looks like quite some code duplication. Could you please reduce it in this PR before merging? 
I'd say a mini wrapping component looks ideal here:
```javascript
function TOSFromItem({terms}: props){
{terms == null || terms.enabled ? (
          <FormItem
            name="tos_check"
            valuePropName="checked"
            rules={[
              {
                validator: (_, value) =>
                  value
                    ? Promise.resolve()
                    : Promise.reject(new Error(messages["auth.tos_check_required"])),
              },
            ]}
          >
            <Checkbox disabled={terms == null}>
              I agree to the{" "}
              {terms == null ? (
                "terms of service"
              ) : (
                <a target="_blank" href={terms.url} rel="noopener noreferrer">
                  terms of service
                </a>
              )}
              .
            </Checkbox>
          </FormItem>
        ) : null}
```

What do you think?

dieknolle3333: I totally agree. I just felt like both registration forms needed a refactoring because there is quite some code duplication, so I didnt even bother to make it DRY. but you are right, I shouldnt add more duplication.

dieknolle3333: I extracted it :) will resolve this conversation for now, let me know if you think that anything is missing.

PR #8201, opened 2024-11-16
Remove Google Analytics

### LINKED ISSUE #7388: Remove Google Analytics
I think we could remove Google Analytics within the Webknossos app. It doesn't track the information we need. So, there is no need to feed it more data.

fm3:
### Steps to test:
- CI should be enough

### Issues:
- fixes #7388

------

## Release Notes

- **New Features**
	- Enhanced asynchronous handling of image file reading for improved efficiency.
	- Improved handling of remote URIs from Neuroglancer by ignoring format prefixes.

- **Bug Fixes**
	- Resolved performance bottleneck during the deletion of multiple trees.
	- Fixed issue with importing NML files that lacked trees.
	- Corrected critical bug that deleted the entire active tree when attempting to delete a non-existing node via the API.
	- Improved clarity of error messages for starting jobs on datasets from other organizations.
	- Removed bounding box size restriction for inferral jobs for super users.

- **Chores**
	- Removed integration of Google Analytics across the application, including tracking functionality in various components.


frontend/javascripts/navbar.tsx
```diff
@@ -739,13 +738,9 @@ function AnonymousAvatar() {
   );
 }
 
-async function getAndTrackVersion(dontTrack: boolean = false) {
+async function getVersion() {
   const buildInfo = await getBuildInfo();
-  const { version } = buildInfo.webknossos;
-  if (dontTrack) {
-    trackVersion(version);
-  }
```
daniel-wer, line 747: :thinking: :upside_down_face:

fm3: Huh, I didn’t even notice this when removing these lines :sweat_smile:

PR #8205, opened 2024-11-18
Remove explicit returns

### LINKED ISSUE #6356: Get rid of return statements in scala

frcroth:
- https://___.webknossos.xyz

I did not find a linting option in scalac to find return statements (-Xlint:nonlocal-return seems to be something different). https://docs.scala-lang.org/overviews/compiler-options/index.html

### Steps to test:
- Generate ad-hoc meshes
- Use the findData route
- For the cum sum thing I am not sure how to test it.

### TODOs:
- [ ] ...

### Issues:
- fixes #6356

------

## Summary by CodeRabbit

- **New Features**
	- Enhanced the logic for checking segment IDs and finding data offsets, improving performance and readability.
	- Improved JSON parsing logic to handle incorrect order of data elements more effectively.
	- Updated data retrieval functions to return a magnification vector instead of a resolution vector.

- **Bug Fixes**
	- Streamlined error handling and logging for the JSON parsing process to ensure accurate timing and error reporting.

- **Refactor**
	- Simplified methods for better readability and maintainability across multiple services.
	- Renamed variables to reflect updated data structures in the data retrieval process.


fm3 (changes requested):
Very nice! Cumsum parsing and AdHoc meshing LGTM and appear to work :) Unfortunately, the findData route now always fails for me. I tested with the color layers of several datasets including l4_sample. Did it work for you? If so, maybe we can debug this together on my laptop :shrug: 

The CI rule is not that important IMO. Then let’s take care of this when doing reviews :)

PR #8218, opened 2024-11-20
Add key in volumetracing to store if volume bucket data has changed

frcroth:
- https://___.webknossos.xyz

### Steps to test:
- abc

### TODOs:
- [ ] ...

### Issues:
- fixes #

------

### DIFF
--- a/app/models/annotation/AnnotationService.scala
+++ b/app/models/annotation/AnnotationService.scala
@@ -177,7 +177,8 @@ class AnnotationService @Inject()(
         mappingName = mappingName,
         mags = magsRestricted.map(vec3IntToProto),
         hasSegmentIndex = Some(fallbackLayer.isEmpty || fallbackLayerHasSegmentIndex),
-        additionalAxes = AdditionalAxis.toProto(additionalAxes)
+        additionalAxes = AdditionalAxis.toProto(additionalAxes),
+        volumeBucketDataHasChanged = Some(false)
       )
   }
 
--- a/frontend/javascripts/dashboard/explorative_annotations_view.tsx
+++ b/frontend/javascripts/dashboard/explorative_annotations_view.tsx
@@ -699,7 +699,7 @@ class ExplorativeAnnotationsView extends React.PureComponent<Props, State> {
               </div>
               <div className="flex-container">
                 <div className="flex-item" style={{ flexGrow: 0 }}>
-                  {teamTags.length > 0 ? <TeamOutlined className="icon-margin-right"/> : null}
+                  {teamTags.length > 0 ? <TeamOutlined className="icon-margin-right" /> : null}
                 </div>
                 <div className="flex-item">{teamTags}</div>
               </div>
--- a/webknossos-datastore/proto/VolumeTracing.proto
+++ b/webknossos-datastore/proto/VolumeTracing.proto
@@ -48,6 +48,7 @@ message VolumeTracing {
   repeated AdditionalCoordinateProto editPositionAdditionalCoordinates = 21;
   repeated AdditionalAxisProto additionalAxes = 22; // Additional axes for which this tracing is defined
   optional bool mappingIsLocked = 23; // user may not select another mapping (e.g. because they have already mutated buckets)
+  optional bool volumeBucketDataHasChanged = 24; // The volume bucket data has been edited at least once
 }
 
 message SegmentGroup {
--- a/webknossos-tracingstore/app/com/scalableminds/webknossos/tracingstore/tracings/volume/VolumeTracingService.scala
+++ b/webknossos-tracingstore/app/com/scalableminds/webknossos/tracingstore/tracings/volume/VolumeTracingService.scala
@@ -210,7 +210,7 @@ class VolumeTracingService @Inject()(
           ) ?~> "failed to update segment index"
         } yield ()
       }
-    } yield volumeTracing
+    } yield volumeTracing.copy(volumeBucketDataHasChanged = Some(true))
 
   override def editableMappingTracingId(tracing: VolumeTracing, tracingId: String): Option[String] =
     if (tracing.getHasEditableMapping) Some(tracingId) else None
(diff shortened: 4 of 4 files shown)


frontend/javascripts/oxalis/view/context_menu.tsx
```diff
@@ -1453,6 +1457,7 @@ function ContextMenuInner(propsWithInputRef: Props) {
   const isLoadingVolumeAndBB = [isLoadingMessage, isLoadingMessage];
   const [segmentVolumeLabel, boundingBoxInfoLabel] = useFetch(
     async () => {
+      const { tracing, flycam } = Store.getState();
```
MichaelBuessemeyer, line 1460: nice change :tada:
(saving rerenders)

frontend/javascripts/oxalis/model/sagas/saga_helpers.ts
```diff
@@ -113,4 +113,36 @@ export function* ensureMaybeActiveMappingIsLocked(
...
+  const { isMappingLockedIfNeeded } = yield* call(ensureMaybeActiveMappingIsLocked, volumeTracing);
+  if (!isMappingLockedIfNeeded) {
+    return false;
+  }
+
+  // Mark that bucket data has changed
+  yield* put({
+    type: "SET_VOLUME_BUCKET_DATA_HAS_CHANGED",
+    tracingId: volumeTracing.tracingId,
+  });
```
MichaelBuessemeyer, line 144: I'd use the constructor for the action here:
Moreover, the action states, that the member of the action should be the `layerName` and not the `tracingId` -> this is kinda confusing. Does this even work?
Something like this might be more correct
```suggestion
  yield* put(setVolumeBucketDataHasChangedAction(volumeTracing.tracingId));
```

frontend/javascripts/oxalis/model/reducers/volumetracing_reducer.ts
```diff
@@ -377,6 +377,12 @@ function VolumeTracingReducer(
       return expandSegmentParents(state, action);
     }
 
+    case "SET_VOLUME_BUCKET_DATA_HAS_CHANGED": {
+      return updateVolumeTracing(state, action.layerName, {
+        volumeBucketDataHasChanged: true,
+      });
```
MichaelBuessemeyer, line 383: `updateVolumeTracing` requires a `volumeTracingId` as a second argument

```suggestion
      return updateVolumeTracing(state, action.volumeTracingId, {
        volumeBucketDataHasChanged: true,
      });
```

frontend/javascripts/oxalis/view/right-border-tabs/segments_tab/load_mesh_menu_item.tsx
```diff
@@ -0,0 +1,34 @@
...
+import FastTooltip from "components/fast_tooltip";
+import type { VolumeTracing } from "oxalis/store";
+import type { APIMeshFile } from "types/api_flow_types";
+
+type Props = {
+  currentMeshFile: APIMeshFile | null | undefined;
+  volumeTracing: VolumeTracing | null | undefined;
+};
+
+export function LoadMeshMenuItemLabel({ currentMeshFile, volumeTracing }: Props) {
```
MichaelBuessemeyer, line 11: The name does not match with the file name. Maybe the file name should be adjusted a little

frontend/javascripts/oxalis/model/reducers/volumetracing_reducer.ts
MichaelBuessemeyer: I gave this a debugging double check: The `volumeBucketDataHasChanged` field is never updated by the frontend. The warning only appears due to other reasons. See my comment in the LoadMeshMenuItemLabel

frontend/javascripts/oxalis/view/right-border-tabs/segments_tab/load_mesh_menu_item.tsx
```diff
@@ -0,0 +1,34 @@
...
+  const showWarning =
+    volumeTracing?.volumeBucketDataHasChanged ??
+    (!volumeTracing?.hasEditableMapping && volumeTracing?.mappingIsLocked);
```
MichaelBuessemeyer, line 14: As written in another comment `volumeTracing?.volumeBucketDataHasChanged` is never set to true by the frontend.

This warning also appears when the tracing has no editable mapping (`volumeTracing?.hasEditableMapping === undefined`) but the user did some annotation `volumeTracing?.mappingIsLocked === true`
As `!undefined` is evaluated to `true` the resulting logic is always true for some basic volume annotation which was edited but does not have an editable mapping. Therefore, the logic is currently working from a user perspective, but the code is not doing what it is intended to do (at least thats how I read it.)

But imo it should be fixed by using the correct field name in the `setVolumeBucketDataHasChangedAction`

Or to avoid such unintended conversions, maybe use `volumeTracing?.hasEditableMapping === false`. I think this would have shown the error right away

MichaelBuessemeyer (changes requested):
Hej you two, thanks for this PR :pray: 

I found one thing:
The frontend constructs its own action object at place instead of using the action constructor, which messes up the typing and due to different naming of actions params the `setVolumeBucketDataHasChangedAction` does not work anymore. You can find more details in my comments

The backend LGTM :)

frontend/javascripts/oxalis/model/sagas/saga_helpers.ts
philippotto: > I'd use the constructor for the action here

whoops, this is what I get for using windsurf to code this :see_no_evil:

> Moreover, the action states, that the member of the action should be the layerName and not the tracingId -> this is kinda confusing. Does this even work?

for volume tracings, `layerName` is always identical to the `tracingId`. it's a bit confusing and I'd love to change this, but the current implementation is in line with the rest. the reason for this is that there are some actions that both work on volumetracing layers as well as segmentation layer (these don't have tracing ids). so, the layer name is the common denominator.

MichaelBuessemeyer: But this action is only designed to be called on volumetracings, isn't it?  Because in that case `tracingId` would also be correct / more precise

MichaelBuessemeyer (approved):
Looks good and working for me :tada:

frontend/javascripts/oxalis/model/sagas/saga_helpers.ts
philippotto: ok, you are right :) for this specific case, tracingId also works without problems. I changed it now!

PR #8221, opened 2024-11-20
Evaluate segmentation in infer neurons task

cdfhalle:
- adds the functionality to evaluate a neuron segmentation using the ai modal.
- In the neuron segmentation tab of the modal the user (a scalable minds employee) can now additionally open the evaluation settings form and type in parameters for the evaluate-split-merger task in voxelytics. Whenever the evaluation settings are used the evaluation should happen subsequently to the inference and the error annotation should be uploaded to the resulting dataset.

### Steps to test:
- pull [3702-evaluate-segmentation-in-infer-neurons-task](https://github.com/scalableminds/voxelytics/tree/3702-evaluate-segmentation-in-infer-neurons-task) from voxelytics
- in the locally hosted webknossos instance open the ai modal, select Neuron Segmentation and click 

### TODOs:
- [ ] ...

### Issues:
- contributes to https://github.com/scalableminds/voxelytics/issues/3702

------


CHANGELOG.unreleased.md
```diff
@@ -11,6 +11,7 @@ For upgrade instructions, please check the [migration guide](MIGRATIONS.released
 [Commits](https://github.com/scalableminds/webknossos/compare/24.11.1...HEAD)
 
 ### Added
+- It is now possible to start a split-merger evaluation when starting a neuron inferrence. [#8221](https://github.com/scalableminds/webknossos/pull/8221)
```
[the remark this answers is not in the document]
MichaelBuessemeyer: I'd say there is no need to such a details explanation. But the message has a little typo :)

```suggestion
- It is now possible to start a split-merger evaluation when starting a neuron inference. [#8221](https://github.com/scalableminds/webknossos/pull/8221)
```

frontend/javascripts/oxalis/view/action-bar/starting_job_modals.tsx
```diff
@@ -742,6 +841,9 @@ export function NeuronSegmentationForm() {
           </Space>
         </>
       }
+      jobSpecificInputFields={
+        <CollapsibleEvaluationSettings isActive={useEvaluation} setActive={setUseEvaluation} />
+      }
```
MichaelBuessemeyer, line 846: I am just guessing here: The split merger evaluation requires an annotation to work, doesn't it? And more specific a skeleton annotation is required, isn't it?

If that's so please only set the `jobSpecificInputFields` to `CollapsibleEvaluationSettings` if the user currently has an opened annotation with a skeleton annotation.
You can retrieve whether there is a skeleton tracing via: 
```javascript
  const hasSkeletonAnnotation = useSelector((state: OxalisState) => state.tracing.skeleton != null);
```

frontend/javascripts/oxalis/view/action-bar/starting_job_modals.tsx
```diff
@@ -523,11 +527,84 @@ function ShouldUseTreesFormItem() {
...
+      onChange={() => setActive(!isActive)}
+      expandIcon={() => <Checkbox checked={isActive} />}
+      items={[
+        {
+          key: "evaluation",
+          label: "Evaluation Settings",
+          children: (
+            <Row>
+              <Col style={{ width: "100%" }}>
+                <Form.Item
```
MichaelBuessemeyer, line 546: This from item with the check box looks a little quirky in vertical mode. Could you please make it horizontal?
```javascript
<Form.Item
  ...
  layout="horizontal"
  />
```

MichaelBuessemeyer (changes requested):
Great Work. Thanks a lot for your PR @cdfhalle 
Especially how you integrated the new UI :+1: 

Testing went well :muscle: but I would like to have a few minor points addressed. The main points are:
1. Please rename almost all variable with evaluation in their name to include that the evaluation is a split-merger evaluation. E.g. `doEvaluation` -> `doSplitMergerEvaluation`. This makes it easier to understand what evaluation is meant, as maybe there might be other evaluations to other jobs or even potentially to the same job.
2. Please ensure in case an evaluation should be done, that the backend validates that the required parameters are given (not None) / sent by the client and not that they are empty options (None).
- The rest is minor stuff I'd say :)

frontend/javascripts/oxalis/view/action-bar/starting_job_modals.tsx
```diff
@@ -523,11 +527,84 @@ function ShouldUseTreesFormItem() {
...
+                  label="Use sparse ground truth tracing"
+                  name={["evaluationSettings", "useSparseTracing"]}
+                  valuePropName="checked"
+                  initialValue={false}
+                  tooltip="The evaluation mode can either be `dense`
+    in case all processes in the volume are annotated in the ground-truth.
+    If not, use the `sparse` mode."
```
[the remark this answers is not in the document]
cdfhalle: They are directly taken from the voxelytics task that does the evaluation, so they should be somewhat right.

frontend/javascripts/oxalis/view/action-bar/starting_job_modals.tsx
MichaelBuessemeyer: Oh I am sorry, I mislooked regarding since what version this property is available. I think I read `4.18` but it is `5.18`. I updated antd a little to support this prop, please pull again and run `yarn install`. This should update the version of antd so that the form items support the `layout` prop. 
Maybe your IDE needs a restart after the update because the TS background process did not register the upgrade and still thinks the code is wrong. At least that's what happened in my case.

MichaelBuessemeyer: And you can find the docs about the `layout` prop here: https://ant.design/components/form#formitem. In the version column it clearly says `5.18` (not  `4.18`)

frontend/javascripts/admin/api/jobs.ts
```diff
@@ -179,12 +179,37 @@ export function startNeuronInferralJob(
   layerName: string,
   bbox: Vector6,
   newDatasetName: string,
+  doSplitMergerEvaluation: boolean,
+  annotationId?: string,
+  useSparseTracing?: boolean,
+  evalMaxEdgeLength?: string,
+  evalSparseTubeThresholdNm?: string,
+  evalMinMergerPathLengthNm?: string,
```
MichaelBuessemeyer, line 187: The typing is incorrect here: the parameters are actually parsed as numbers here. Please see my other comments regarding improving the typing
```suggestion
  evalMaxEdgeLength?: number,
  evalSparseTubeThresholdNm?: number,
  evalMinMergerPathLengthNm?: number,
```

frontend/javascripts/oxalis/view/action-bar/starting_job_modals.tsx
```diff
@@ -523,11 +527,85 @@ function ShouldUseTreesFormItem() {
   );
 }
```
MichaelBuessemeyer, line 529: Please add a type for the settings created / configured by this component here in the lines above the component. This can be used to improve the typing of the data received from the form component. the correct type should be something like:

```suggestion
type SplitMergerEvaluationSettings = {
    useSparseTracing?: boolean;
    maxEdgeLength?: number;
    sparseTubeThresholdInNm?: number;
    minimumMergerPathLengthInNm?: number;
}
```
The settings are numbers because you are correctly using number inputs and not simple string inputs.

frontend/javascripts/oxalis/view/action-bar/starting_job_modals.tsx
```diff
@@ -523,11 +527,85 @@ function ShouldUseTreesFormItem() {
...
+                  <Checkbox style={{ width: "100%" }} />
+                </Form.Item>
+                <Form.Item
+                  label="Max edge length in nm"
+                  name={["splitMergerEvaluationSettings", "maxEdgeLength"]}
+                  tooltip="Ground truth tracings can be densified so that
+    nodes are at most max_edge_length nm apart.
+    However, this can also introduce wrong nodes in curved processes."
+                >
+                  <InputNumber style={{ width: "100%" }} placeholder="None" />
```
MichaelBuessemeyer, line 565: Here you use a `InputNumber` which leads to `maxEdgeLength` being of type number :) -> Thus the suggestion to improve the typing to avoid potential errors in the future

frontend/javascripts/oxalis/view/action-bar/starting_job_modals.tsx
```diff
@@ -710,13 +791,40 @@ export function NeuronSegmentationForm() {
       title="AI Neuron Segmentation"
       suggestedDatasetSuffix="with_reconstructed_neurons"
       isBoundingBoxConfigurable
-      jobApiCall={async ({ newDatasetName, selectedLayer: colorLayer, selectedBoundingBox }) => {
-        if (!selectedBoundingBox) {
+      jobApiCall={async (
+        { newDatasetName, selectedLayer: colorLayer, selectedBoundingBox, annotationId },
+        form: FormInstance<any>,
+      ) => {
+        const splitMergerEvaluationSettings = form.getFieldValue("splitMergerEvaluationSettings");
```
MichaelBuessemeyer, line 798: This should fix the typing of `splitMergerEvaluationSettings` from `any` to the desired type to improve the type checking
```suggestion
        const splitMergerEvaluationSettings = form.getFieldValue("splitMergerEvaluationSettings") as SplitMergerEvaluationSettings;
```

MichaelBuessemeyer (changes requested):
Thanks for applying the feedback :pray:

- There is still one open TODO from my last review
- I identified, that the typing of the split merger settings was incorrect and added some comments on how to improve them
- And the job is currently crashing because the param `ogranization_name` was not renamed to `organization_id`. Pulling and merging with the master in the vx repo should fix the error

frontend/javascripts/oxalis/view/action-bar/starting_job_modals.tsx
```diff
@@ -523,11 +527,92 @@ function ShouldUseTreesFormItem() {
   );
 }
 
+type SplitMergerEvaluationSettings = {
+  useSparseTracing?: boolean;
+  maxEdgeLength?: number;
+  sparseTubeThresholdInNm?: number;
+  minimumMergerPathLengthInNm?: number;
+};
```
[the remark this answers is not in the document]
MichaelBuessemeyer: These settings are actually optional. so please ignore this

frontend/javascripts/oxalis/view/action-bar/starting_job_modals.tsx
```diff
@@ -710,13 +798,42 @@ export function NeuronSegmentationForm() {
...
+        { newDatasetName, selectedLayer: colorLayer, selectedBoundingBox, annotationId },
+        form: FormInstance<any>,
+      ) => {
+        const splitMergerEvaluationSettings = form.getFieldValue(
+          "splitMergerEvaluationSettings",
+        ) as SplitMergerEvaluationSettings;
+        if (
+          !selectedBoundingBox ||
+          (doSplitMergerEvaluation && splitMergerEvaluationSettings == null)
+        ) {
```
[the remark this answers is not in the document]
MichaelBuessemeyer: Nope

MichaelBuessemeyer (commented):
Thanks for incorporating my feedback. I left two other suggestions which I missed in the last review.

After that, I'd give this another quick test run and then it should be ready to be merged :tada:

MichaelBuessemeyer (approved):
Awesome :tada. Thanks a lot for implementing all the feedback.

Should be ready to be merged :dancers: 


Moreover, while reviewing I noticed a misaligned HTML element (the drag handle). I just pushed a commit to your branch so that this is fixed when your pr is merge. This change is not important and also not worthy an own pr. so I just went ahead and added it to the branch I had checkout out (which was yours) :). Hope that's fine :v: 

Here is the visual bug:
image
And this is how it looks like when it is fixed:
image
-> This burger menu icon is now centered vertically :)

app/models/job/Job.scala
```diff
@@ -66,6 +67,10 @@ case class Job(
           }.getOrElse(datasetName.map(name => s"datasets/$organizationId/$name/view"))
         case JobCommand.export_tiff | JobCommand.render_animation =>
           Some(s"/api/jobs/${this._id}/export")
+        case JobCommand.infer_neurons if this.argAsBooleanOpt("do_evaluation").getOrElse(false) =>
+          returnValue.map { resultAnnotationLink =>
+            resultAnnotationLink
+          }
```
MichaelBuessemeyer, line 73: While trying to open the created evaluation annotation I just noticed a little bug:

As a result I got the following link in the jobs page: 
`http://localhost:9000/datasets/sample_organization//annotations/Explorational/678a1d535b0100aa21602e08/view`
This link includes a double `/` which needs to be removed. Likely for consistency sake it should be removed int the worker job as else this logic needs to be implemented in the backend here or in the frontend. Fixing this in the frontend or backend might lead to an "unexpected" special case for infer neuron jobs that have an evaluation. Fixing this in the worker has the advantage that the "special case is directly handled where it is created" IMO. Feel free to disagree

MichaelBuessemeyer: Oh wait, I misread the link. The `/datasets/sample_organization/` part needs to be removed I'll give this a check where to best implement this. 
My quick guess would be fixing this in the frontend

MichaelBuessemeyer: > My quick guess would be fixing this in the frontend

Mööööppp! Wrong :red_circle: 

The fix is to replace `this.argAsBooleanOpt("do_evaluation")` with `this.argAsBooleanOpt("do_split_merger_evaluation")`. This line simply was forgotten during the raming of the  `do_evaluation` variable. This should fix all the things :tm:

app/models/job/Job.scala
```diff
@@ -66,6 +67,10 @@ case class Job(
           }.getOrElse(datasetName.map(name => s"datasets/$organizationId/$name/view"))
         case JobCommand.export_tiff | JobCommand.render_animation =>
           Some(s"/api/jobs/${this._id}/export")
+        case JobCommand.infer_neurons if this.argAsBooleanOpt("do_evaluation").getOrElse(false) =>
```
MichaelBuessemeyer, line 70:
```suggestion
        case JobCommand.infer_neurons if this.argAsBooleanOpt("do_split_merger_evaluation").getOrElse(false) =>
```

MichaelBuessemeyer (approved):
Whit this small fix (please see my comments)  the pr should be ready to go. Please only merge together the corresponding worker pr :)

PR #8225, opened 2024-11-21
Fix changing color of active segment

### LINKED ISSUE #8223: Changing a segment's color modifies the color of the previously changed segment
## Context
As an example to reproduce the error, I copied the `Featured Publication` below to my trial account on webknossos.org so that I could edit the segment colors.

image

## Expected Behavior
I would expect that the `Change Segment Color` component allows for editing the selected segment's color.

## Current Behavior
When I edit the color of more that 1 segment, selecting `Change Segment Color` modifies the color of the previously changed segment.

## Steps to Reproduce the bug
See screen recording below.  This behavior also occurs if I press `Save` after editing each segment color.

https://github.com/user-attachments/assets/a7d1bbe5-76e0-4676-95c3-75197f9fcfeb

## Your Environment for bug
- Browser name and version: Safari 18.0.1
- Operating System and version: MacOS 15.0.1
- Version of WEBKNOSSOS (Release or Commit): Version currently deployed at webknossos.org

cc @aaronkanzer @jssyq

MichaelBuessemeyer:
When changing the color of multiple segments one after another, the color change was applied to the previous segment. This was due to the color-changing menu item being the only menu item that is a component that is cached by react. Therefore, the color-changing menu item always referred to the segment on which the context menu was previously rendered. By including the segment id in the color-changing menu item's key, an update for the color-changing menu item is enforced each time the segment id changes. Thus the bug should be vanished :sparkle: 

- https://___.webknossos.xyz

### Steps to test:
- try on master / wk org:
- Open a hybrid tracing on any dataset with an already existing volume layer
- select two segments to add them to the segments list 
- in the segment list change the color of the first segment via the context menu in the segments list. Should work
- change the color of the other segment ->  should update the color of the previous segment :woozy_face: 
- switch to this branch / dev instance
- do the exact same as before
- when changing the color for the second segment, this should update the segments color (not the previous one) -> bug should be gone

### Issues:
- fixes #8223

------

## Summary by CodeRabbit

- **New Features**
	- Users can now paste remote URIs from Neuroglancer without removing format prefixes.
	- Asynchronous reading of image files from the datastore for improved performance.
	- Enhanced clarity in error messages for starting jobs on external datasets.
	- Super users can now utilize larger bounding box sizes for inferral jobs.

- **Bug Fixes**
	- Resolved performance issues when deleting multiple trees.
	- Fixed color change inconsistency in segment selection.
	- Addressed NML file import issues and a critical deletion bug in the API.

- **Chores**
	- Removed Google Analytics integration from the project.

### DIFF
--- a/CHANGELOG.unreleased.md
+++ b/CHANGELOG.unreleased.md
@@ -20,6 +20,7 @@ For upgrade instructions, please check the [migration guide](MIGRATIONS.released
 
 ### Fixed
 - Fix performance bottleneck when deleting a lot of trees at once. [#8176](https://github.com/scalableminds/webknossos/pull/8176)
+- Fix a bug where changing the color of a segment via the menu in the segments tab would update the segment color of the previous segment, on which the context menu was opened. [#8225](https://github.com/scalableminds/webknossos/pull/8225)
 - Fix a bug when importing an NML with groups when only groups but no trees exist in an annotation. [#8176](https://github.com/scalableminds/webknossos/pull/8176)
 - Fix a bug where trying to delete a non-existing node (via the API, for example) would delete the whole active tree. [#8176](https://github.com/scalableminds/webknossos/pull/8176)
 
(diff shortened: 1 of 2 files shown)


philippotto (approved):
:rocket:

PR #8226, opened 2024-11-21
Fix docstring for registerSegment

philippotto:
Fixes the docstring for `registerSegment` in the JS API module.


MichaelBuessemeyer (approved):
looking good, thanks for updating the docs (and finding the issue here) :pray:

PR #8231, opened 2024-11-26
Misc layout improvements in dataset upload

philippotto:
The toasts were redundant because the popover showed the same information. I also fixed the layout in the upload modals (as well as one in the "this dataset does not exist" screen).

Best viewed with "Hide whitespace".

- https://___.webknossos.xyz

### Steps to test:
I checked the following scenarios (ignore the theme change):

before: 
image

after: 
image

before:
image

after:
image

### Issues:
- context: https://scm.slack.com/archives/C02A9H8CU/p1732611414662869?thread_ts=1732551186.574079&cid=C02A9H8CU

------


hotzenklotz (commented):
Code Changes LGTM. Will do some testing now.

hotzenklotz:
When testing I noticed two issues:

1. The upload progress bar always starts at 100% before "quickly" jumping to 0%. I assume the some default values is either wrong or missing entirely.
2. After the upload succeeded a spinner "Processing uploaded files" is shown. The spinner, however, is overlaid over the progress bar and hardly visible. See screenshot.

Screenshot 2024-11-28 at 13 35 28
Screenshot 2024-11-28 at 13 32 49

philippotto:
@hotzenklotz Thanks for catching these things! I could reproduce the progress=100 thingie when uploading another dataset after a first DS was uploaded. For the very first DS, it already worked for me. I pushed a fix for the non-first DS.

For your other point, I changed the layout so that the two phases look like this now:

image
image
(it's an animated spinner)

hotzenklotz (approved):
🎉

Issue #8172, opened 2024-11-06
Dataset settings need multiple submits with same form input to succeed


MichaelBuessemeyer:
The frontend form validation of the dataset settings seem to be a little faulty.

## Context

When I try to edit a dataset via the advanced datasource view and then try to submit the changes, I get an error that some fields have errors, although all values are valid. Only once I try to submit again, the frontend validation succeeds and then sends the data to ther server.

## Expected Behavior

The first try of submitting the changes should succeed.

## Current Behavior

The first try fails. I gave this a quick investigation.
The console shows an error that 
```
validator: syncValidator(
                  (value: string) =>
                    dataLayers.filter((someLayer: APIDataLayer) => someLayer.name === value)
                      .length <= 1,
                  "Layer names must be unique.",
                ),
```
yields an undefined error in the first try to submit it, as the useWatch value is outdated and undefined. Switching to 
```
validator: syncValidator(
                  (value: string) =>
                    form
                      .getFieldValue(["dataSource", "dataLayers"])
                      .filter((someLayer: APIDataLayer) => someLayer.name === value).length <= 1,
                  "Layer names must be unique.",
                ),
```
resolves this error, but the validation error still exists. With `outOfDate: true` being set in the error thrown by the validation. I did not have a quick idea regarding this, thus I created this issue.

## Steps to Reproduce the bug

1. Open the dataset settings for one ds
2. select the advanced tab
3. For one color layer increase the bbox by one voxel in each dim.
4. try to save -> the error should occur
5. try to save again -> this time it should succeed

## Your Environment for bug
Nothing special. Noticed why working on #8075 but seems like this was not introduced in this pr.

PR #8241, opened 2024-11-27
Allow dollar in layer name

MichaelBuessemeyer:
- https://allowdollarinlayernames.webknossos.xyz

### Steps to test:
- Open the dataset settings of any DS
- change one layer & the dataset name to include a $. Submitting should work (but likely running the DS in case a layer is renamed)
- Renaming a layer can be done via the advanced tab

### TODOs:
- [x] also allow `$` in `checkLayerNameForInvalidCharacters`
- [x] disallow `$` for datasetNames
- [x] test migration & reversion

### Issues:
- fixes https://scm.slack.com/archives/C5AKLAV0B/p1732724785230369
- fixes https://github.com/scalableminds/webknossos/issues/8172

------


fm3:
Can we allow it for layers but not for datasets?

daniel-wer (changes requested):
The functionality you describe works as intended, but I noticed two things during testing:

- [x] When trying to rename a volume tracing layer from within an annotation, $ is still not allowed (`This layer name includes the disallowed character "$". Please remove it to set the layer name.`)
- [ ] When changing the layer name in the advanced datasource settings, the first click always triggers a warning toast, indicating that one of the fields is not valid (`Please check that all form fields are valid.`), but no field is highlighted. This happens regardless of whether a $ character is included or not and only on the first submission. Clicking submit a second time succeeds. I briefly tried to debug this, but didn't find the cause. This issue probably existed before, so can also be solved in a separate PR.

MichaelBuessemeyer:
Thanks for the review @daniel-wer :pray: 

>  When changing the layer name in the advanced datasource settings, the first click always triggers a warning toast, indicating that one of the fields is not valid (Please check that all form fields are valid.), but no field is highlighted. This happens regardless of whether a $ character is included or not and only on the first submission. Clicking submit a second time succeeds. I briefly tried to debug this, but didn't find the cause. This issue probably existed before, so can also be solved in a separate PR.

Is a known bug in the frontend code: See: https://github.com/scalableminds/webknossos/issues/8172

> When trying to rename a volume tracing layer from within an annotation, $ is still not allowed (This layer name includes the disallowed character "$". Please remove it to set the layer name.)

Yeah, the frontend resticts this. While trying to add `$` as an allowed character I stumbled upon the variable which is the regex enforcing which characters are allowed for annotation layer names in the frontend. The regex is called `uriSafeCharactersRegex` in `add_volume_layer_modal.tsx`. This makes me think that maybe allowing unsave characters might lead to some problems with links or so. Do you maybe know more about this daniel or @fm3? You (@fm3) are the last person changing the regex in this PR: https://github.com/scalableminds/webknossos/pull/6326/files#diff-51a391c68ba6e4cf8ca5ba79e13d87714d938a303418e9ec9d325d4c202858ac. In the PR & linked issue I cannot find, why the late name needs to be URL save. Usually special characters should be URL encodable, shouldn't they?

Moreover, just adjusting the fronend is not enough. The backend also has a check in place in the DB. Therefore, a migration would also be necessary here :woozy_face: (quite the amount of work for such a small change)

MichaelBuessemeyer:
> Maybe it can be done in two steps? First the chanage for dataset layers as implemented here, and then as a follow-up also for annotation layer names?

Just doing one pr is fine for me. It's was just a little unexpected to me that this small change requires a migration.

Thanks for the hint, I'll do the zarr streaming testing :D

MichaelBuessemeyer:
> If it works, then allowing the dollar in frontend and backend should be fine.

Works with color layer, segmentation layer, volume layer with fallback layer, and without fallback layer. All with `$` in their names. => Should be ok

conf/evolutions/reversions/125-allow-dollar-in-layer-names.sql
```diff
@@ -0,0 +1,11 @@
+START TRANSACTION;
+
+-- This reversion might take a while because it needs to search in all annotation layer names for '$' and replace it with ''
+do $$ begin ASSERT (select schemaVersion from webknossos.releaseInformation) = 125, 'Previous schema version mismatch'; end; $$ LANGUAGE plpgsql;
+
+UPDATE webknossos.annotation_layers SET name = regexp_replace(name, '\$', '', 'g') WHERE name ~* '\$';
+
+ALTER TABLE webknossos.annotation_layers DROP CONSTRAINT IF EXISTS annotation_layers_name_check;
+ALTER TABLE webknossos.annotation_layers ADD CONSTRAINT annotation_layers_name_check  CHECK (name ~* '^[A-Za-z0-9\-_\.]+$');
+
```
daniel-wer, line 10: Does the schemaVersion need to be set to 124 again, here?

frontend/javascripts/oxalis/view/left-border-tabs/modals/add_volume_layer_modal.tsx
```diff
@@ -48,7 +48,7 @@ export function checkLayerNameForInvalidCharacters(readableLayerName: string): V
       message: messages["tracing.volume_layer_name_starts_with_dot"],
     };
   }
-  const uriSafeCharactersRegex = /[0-9a-zA-Z-._]+/g;
+  const uriSafeCharactersRegex = /[0-9a-zA-Z-._$]+/g;
```
daniel-wer, line 51: If I remember correctly you found that these characters don't necessarily have to be url safe, right? If so, maybe rename to `validLayerNameCharactersRegex` or so instead.

daniel-wer (commented):
Frontend code LGTM

conf/evolutions/reversions/125-allow-dollar-in-layer-names.sql
```diff
@@ -0,0 +1,11 @@
+START TRANSACTION;
+
+-- This reversion might take a while because it needs to search in all annotation layer names for '$' and replace it with ''
+do $$ begin ASSERT (select schemaVersion from webknossos.releaseInformation) = 125, 'Previous schema version mismatch'; end; $$ LANGUAGE plpgsql;
+
+UPDATE webknossos.annotation_layers SET name = regexp_replace(name, '\$', '', 'g') WHERE name ~* '\$';
```
MichaelBuessemeyer, line 6: @fm3 There is a problem here with the reversion. Removing `$` from existing annotation layer names makes them very likely invalid / not findable in the fossil db.
Therefore, I'd say that this PR is not really revertable? Otherwise one needs to fix the data in the fossilDB. Or at least the schema cannot be rolled back, because rolling back the `annotation_layers_name_check` would require no annotation layer names with `$` in their name :thinking: 

At least that's a effect I observed while checking the reversion. An annotation with an annotation layer without fallback that included `$` was suddenly empty (if I remember correctly that I did some annotations in that layer).

fm3: the layer names are not used for lookup in the fossildb, only the tracingid is used for that. I think renaming annotation layers should be fairly harmless.

On the other hand, if the fallback segmentation name changes, that may lead to issues.

Not sure what you saw there :thinking: Could you try again if you can reproduce this? After a page refresh the data should be ok.

frontend/javascripts/dashboard/dataset/dataset_settings_view.tsx
```diff
@@ -339,12 +339,18 @@ class DatasetSettingsView extends React.PureComponent<PropsWithFormAndRouter, St
...
+      // and defer the validation as it is done asynchronously by antd or so.
+      setTimeout(
+        () =>
+          form
+            .validateFields()
+            .then((formValues) => this.submit(formValues))
+            .catch((errorInfo) => this.handleValidationFailed(errorInfo)),
+        0,
+      );
+    };
```
MichaelBuessemeyer, line 353: I managed to fix https://github.com/scalableminds/webknossos/issues/8172. The reason for the error is (as correctly described in the issue) first: The syncValidator in `dataset_settings_data_tab.tsx` needs to use the up-to-date value of the datasetFields and the form field validation failed due to an `outOfDate` error signaled by antd. The reason for this seems to be, that the validation was done on outdated values. 

After some trial and error I came up with the fix of simply deferring the whole validation as the debugger showed that the form validation was done by antd in some kind of asynchronous context or so. Therefore, by forcefully making the validation and submitting also asynchronous, it should be reliable executed after the form validation leading to up to date validated value in the `validateFields` call avoiding the bug described in #8172.

I hope this makes kinda sense. I did not dive deeper into why this resolved the bug, but it seems to work reliably, and I do not want to put more time into it.

conf/evolutions/reversions/125-allow-dollar-in-layer-names.sql
MichaelBuessemeyer: Ok I gave this another investigation. I could not reproduce the error I remembered. Maybe I messed up somewhere.
The only error I could produce was with a exactly that:
> On the other hand, if the fallback segmentation name changes, that may lead to issues.

See: 
image
Before executing the reversion the name of the segmentation fallback layer & the volume annotation layer was `segmentation$dollar`. After the reversion the annotation layer name was `segmentationdollar` but the fallback layers name stayed at `segmentation$dollar`. Then I manually renamed the fallback segmentation layer to `segmentationdollar` in the datasource and folder structure. => viewing the ds works but the volume segmentation layer is now detached from the fallback layer (as shown in the screenshot). 
But I'd say that this is an expected problem due to the renaming of the underlying segmentation layer? 

In short: I think I was just wrong here, claiming that the reversion might cause problems. Sorry :see_no_evil:

fm3: That’s good news then :) Yes, that sounds plausible

fm3 (commented):
Backend LGTM, leaving final approval to frontend team

frontend/javascripts/dashboard/dataset/dataset_settings_view.tsx
daniel-wer: I found [this SO post](https://stackoverflow.com/questions/62731874/form-validatefields-fails-inside-onvalueschange-in-ant-design-forms) referencing [this antd issue](https://github.com/ant-design/ant-design/issues/26747) which are not about this exact use case but pretty similar. Since a maintainer suggests to use `setTimeout` it's probably fine as a workaround for now.

daniel-wer (approved):
LGTM, thank you! :+1:

PR #8244, opened 2024-11-28
Disallow empty annotation layer names and names starting with a '.'

MichaelBuessemeyer:
This pr adds frontend validation to annotation layer names and tries to put the same restrictions to the annotation layer names as to "normal" dataset layer names.
Sadly, it is not easily possible to reuse the rules set for dataset layer names, as these are formulated as antd form validation rules and the annotation layer name setting uses an `EditableTextLabel`. The `EditableTextLabel` has its own format for validation rules which is not directly compatible with the antd variation -> Therefore I duplicated the two missing rules:
- not starting with a `.`
- at least one character long  

- https://___.webknossos.xyz

### Steps to test:
- Open a hybrid annotation
- try renaming the segmentation layer to an empty string -> frontend validation should fail
- try renaming the segmentation layer to something starting with a `.` -> frontend validation should fail

### Issues:
- fixes https://scm.slack.com/archives/CMBMU5684/p1732748157886569

------


frontend/javascripts/oxalis/view/left-border-tabs/layer_settings_tab.tsx
```diff
@@ -748,6 +749,7 @@ class DatasetSettings extends React.PureComponent<DatasetSettingsProps, State> {
                     });
                   }}
                   rules={[
+                    layerNameRules[0] as { min: number }, // Ensuring minimum length
                     {
                       validator: (newReadableLayerName) =>
                         validateReadableLayerName(
```
philippotto, line 755: I think the additional rules should be implemented within `validateReadableLayerName` (or maybe `checkLayerNameForInvalidCharacters` for the `.`).

philippotto (changes requested):
thanks for doing this! see my one suggestion.

frontend/javascripts/oxalis/view/left-border-tabs/layer_settings_tab.tsx
MichaelBuessemeyer: sure thanks for the suggestion. Should be implemented now :)

philippotto (approved):

PR #8245, opened 2024-11-28
Sync dataset name with datasource in add remote view

MichaelBuessemeyer:
- https://___.webknossos.xyz

### Steps to test:
- try adding a remote dataset
- enter a DS name
- submitting should work
- when checking the advanced settings tab before submitting `id.name` should be in sync with the dataset name input field of the simple settings view

### Issues:
- fixes https://scm.slack.com/archives/C5AKLAV0B/p1732788765872929

------

### DIFF
--- a/frontend/javascripts/admin/dataset/dataset_add_remote_view.tsx
+++ b/frontend/javascripts/admin/dataset/dataset_add_remote_view.tsx
@@ -252,13 +252,13 @@ function DatasetAddRemoteView(props: Props) {
     form.setFieldsValue({ dataSourceJson });
     // Since this function sets the JSON string, we have to update the
     // data which is rendered by the "simple" page.
-    syncDataSourceFields(form, "simple");
+    syncDataSourceFields(form, "simple", true);
     form.validateFields();
   };
 
   async function handleStoreDataset() {
     // Sync simple with advanced and get newest datasourceJson
-    syncDataSourceFields(form, dataSourceEditMode === "simple" ? "advanced" : "simple");
+    syncDataSourceFields(form, dataSourceEditMode === "simple" ? "advanced" : "simple", true);
     try {
       await form.validateFields();
     } catch (_e) {
@@ -515,7 +515,7 @@ function AddRemoteLayer({
     }
 
     // Sync simple with advanced and get newest datasourceJson
-    syncDataSourceFields(form, dataSourceEditMode === "simple" ? "advanced" : "simple");
+    syncDataSourceFields(form, dataSourceEditMode === "simple" ? "advanced" : "simple", true);
     const datasourceConfigStr = form.getFieldValue("dataSourceJson");
     const datastoreToUse = uploadableDatastores.find(
       (datastore) => form.getFieldValue("datastoreUrl") === datastore.url,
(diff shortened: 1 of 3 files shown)


frontend/javascripts/dashboard/dataset/dataset_settings_data_tab.tsx
```diff
@@ -33,12 +33,16 @@ import { type APIDataLayer, type APIDataset, APIJobType } from "types/api_flow_t
...
+import type { ArbitraryObject } from "types/globals";
 
 const FormItem = Form.Item;
 
 export const syncDataSourceFields = (
   form: FormInstance,
   syncTargetTabKey: "simple" | "advanced",
+  // Syncing the dataset name is optional as this is needed for the add remote view, but not for the edit view.
+  // In the edit view, the datasource.id fields should never be changed and the backend will automatically ignore all changes to the id field.
+  syncDatasetName = false,
```
MichaelBuessemeyer, line 45: It there maybe a better place to put this comment?

philippotto: I think, it's alright.

philippotto (changes requested):
code looks alright, but testing didn't work for me. in my case I did **not** switch to the advanced tab. I think in that case the synchronization is not triggered? at least, I got this error: `toast.tsx:215 The datasource config could not be stored. Error: The dataset name must not be empty.`. only after switching to the tab, the import worked.

frontend/javascripts/admin/dataset/dataset_add_remote_view.tsx
```diff
@@ -252,13 +252,13 @@ function DatasetAddRemoteView(props: Props) {
...
     // data which is rendered by the "simple" page.
-    syncDataSourceFields(form, "simple");
+    syncDataSourceFields(form, "simple", true);
     form.validateFields();
   };
 
   async function handleStoreDataset() {
     // Sync simple with advanced and get newest datasourceJson
-    syncDataSourceFields(form, dataSourceEditMode === "simple" ? "advanced" : "simple");
+    syncDataSourceFields(form, dataSourceEditMode === "simple" ? "advanced" : "simple", true);
```
MichaelBuessemeyer, line 261: I missed a few occurrences of `syncDataSourceFields` :see_no_evil:, this one here being the most important.

frontend/javascripts/dashboard/dataset/dataset_settings_data_tab.tsx
```diff
@@ -47,12 +51,25 @@ export const syncDataSourceFields = (
   if (syncTargetTabKey === "advanced") {
     // Copy from simple to advanced: update json
     const dataSourceFromSimpleTab = form.getFieldValue("dataSource");
+    if (syncDatasetName && dataSourceFromSimpleTab) {
```
MichaelBuessemeyer, line 54: The `&& dataSourceFromSimpleTab` was added as `syncDataSourceFields` is called before the first exploration request is sent to the backend. In this case `dataSourceFromSimpleTab==undefined` and thus I guarded this assignment below as else the code would silently crash.

MichaelBuessemeyer (commented):
Thanks for noticing :pray: . Should be better now

PR #8246, opened 2024-11-28
Add debug logging for dataset list route

MichaelBuessemeyer:
Currently the route `https://webknossos.org/api/datasets?isActive=false` is broken and the bug is not reproducible in a dev environment. Thus, this pr adds logging to that route to give further hints about the causes of the bug. 

### Steps to test:
- should not be necessary; but in case you want to: Spin up a local serve, request `http://localhost:9000/api/datasets?isActive=false` => the server console should show more extensive logging


fm3 (approved):
LGTM. We’ll need to remove the logging again after we solve this of course :)

PR #8247, opened 2024-11-28
Add further debug logging for ds listing

MichaelBuessemeyer:
The method `listGrouped` needs to be investigated further.

> Answering 400 at /api/datasets?isActive=false for user \<id\>: {"messages":[{"error":"Grouping datasets failed"},{"chain":"[Server Time 2024-11-28 13:54]  <~ Failed to retrieve list of data sets."}]}

Follow up to #8246

### Steps to test:
- same as in #8246


app/controllers/DatasetController.scala
```diff
@@ -224,28 +224,42 @@ class DatasetController @Inject()(userService: UserService,
       implicit ctx: DBAccessContext,
       m: MessagesProvider): Fox[List[JsObject]] =
     for {
+      _ <- Fox.successful(())
+      _ = logger.info(s"datasets: $datasets, requestingUser: $requestingUser")
```
fm3, line 228:
```suggestion
      _ = logger.info(s"datasets: $datasets, requestingUser: ${requestingUser._id}")
```

fm3 (approved):
let’s avoid printing entire user objects, other than that LGTM

app/controllers/DatasetController.scala
MichaelBuessemeyer: > let’s avoid printing entire user objects,

only user objects?

fm3: Yes, should be ok. This is only temporary, this isn’t super sensitive data and the logs are of course not public.

PR #8251, opened 2024-11-29
Re-Add debug logging for ds listing

MichaelBuessemeyer:
Seems like #8249 did not fully fix the issue. I still cannot reproduce the error locally (even when working a very similar scenario to what caused the first error).

- https://___.webknossos.xyz


fm3 (approved):
This time, let’s leave this in until after the bugfix is tested in production :sweat_smile: 

Also, some of the `?~>`error messages may actually be nice to keep in afterwards

PR #8256, opened 2024-12-02
Remove debug logging for ds listing

MichaelBuessemeyer:
Simply removes the logging introduced in scalableminds/webknossos#8251

Some of the improved error messages are kept.


app/models/dataset/DataStore.scala
```diff
@@ -114,7 +114,7 @@ class DataStoreDAO @Inject()(sqlClient: SqlClient)(implicit ec: ExecutionContext
     for {
       accessQuery <- readAccessQuery
       r <- run(q"SELECT $columns FROM $existingCollectionName WHERE name = $name AND $accessQuery".as[DatastoresRow])
-      parsed <- parseFirst(r, name)
+      parsed <- parseFirst(r, name) ?~> "dataStore.notFound"
```
fm3, line 117: I wouldn’t add notFound to parseFirst. I think parseFirst has its own error message, like `s"Parsing failed for row in $collectionName queried by $queryLabel"`
```suggestion
      parsed <- parseFirst(r, name)
```

PR #8258, opened 2024-12-03
In NML upload with overwritingDatasetId, do not require valid orga field in the NML

### LINKED ISSUE #8254: When overwritingDatasetId is set in annotation upload, the NML shouldn’t need an existing orga tag
compare https://scm.slack.com/archives/C5AKLAV0B/p1733147657098769?thread_ts=1733137643.117969&cid=C5AKLAV0B

fm3:
The orga id should be taken from the already looked-up dataset.

- https://skiporgalookupifoverwritten.webknossos.xyz

### Steps to test:
- Download an NML, delete its organization attribute in the experiment
- Upload the NML via drag’n’drop into another dataset that was opened in view mode
- Should work

### Issues:
- fixes #8254


MichaelBuessemeyer (approved):
Thanks for this PR :pray: 

Works great.

But while checking the code I noticed that `NmlParsedParameters` `organizationId`  is still `String` and not an option, although it is now optional. Maybe this should be changed :thinking:. It currently works because `parseOrganizationId` in `NmlParser.scala` returns a `""` as default. But keep / do it the way you prefer :)

PR #8075, opened 2024-09-12
Allow renaming datasets & dataset with duplicate names

MichaelBuessemeyer:
### Further Notes:
- Quite some of the line changes are the result of moving the ObjectId class to the utils package so that all wk backend servers have access to this class.

### URL of deployed dev instance (used for testing):
- https://allowdatasetrenaming.webknossos.xyz

### Steps to test:
- Give two datasets the same name and check whether annotations and so on works
- Test whether the task system still works with duplicate dataset names
- check dataset upload
  - dataset upload
  - add remote
  - compose
- check moving datasets between folders (single & multiple at once)
- check whether the renaming works
- when there are duplicate names, the correct dataset should open upon view it (not another one with the same name)
- Test worker jobs (tested by @fm3)
- Test legacy routes
- And much much more :see_no_evil: 

### TODOs:
- [x] Add evolution and reversion
  - [x] testing needed
- [x] Test uploading: 
  - [x] Report upload fails
- [x] Adjust worker to newest job arguments as the dataset name can no longer be used to uniquely identify a dataset
- [x] rename `organization_name` in worker to `organization_id`. see #8038
- [x] Dataset Name settings field has an unwanted spinner (see upload view)
- [x] Check the job list
- [x] Properly implement legacy searching for datasets when old URI param is used 
- [x] Adjust legacy API routes to return dataset in old format
   - It is just an additional field. Thus, I would say it should be fine. 
- [x] datasets appear to be duplicated in the db 
  - Maybe these are created by jobs with an output dataset 
- [x] Fix dataset insert
- [x] Skeleton & VolumeTracings address a dataset via its name
  - Not really used only during task / annotation creation
  - Use heuristic upon upload and temporary patch the Tracing case classes to carry the datasetId during the creation process once the dataset is identified once.
  - Task creation works
  - [x] Needs testing
    - fix annotation upload 
  - [x] needs to support old nmls
- [x] Put datasetId into newly created nmls
- [x] In the backend `LinkedLayerIdentifier` still uses the datasetName as an identifier
  - used in wklibs, maybe just interpret the name as a path and work with this. in case it cannot be found the user needs to update wklibs. Add comment for this!
- ~~[ ] the dataset `C555_tps_demo` has quite some bucket loading errors. Unsure why some buckets do not work~~ The dataset seems to be broken. Could reproduce this on other branches
- [x] Notion-style URLs are missing (i.e. <name>-<id>, but only the id part is actually used)
- [x] Maybe remove `DatasetURIParser`
[...]

PR #8263, opened 2024-12-05
Fix legacy support for outdated displayName field of datasets

MichaelBuessemeyer:
Fixes a compatibility bug introduced in #8075. `wk.Dataset.open_remote("...").display_name = "fancy name"` did not result in any changes due to two bugs:
1. displayName was no longer part of the ds objects returned by the API, even in legacy routes.
2. When updating a dataset, the name given by the update request took precedence. This discarded all updates to the ds name done via `displayName` as "old" clients like wklibs always send the current name of the ds and the updated `displayName`. Therefore, the `displayName` should take precedence. Clients using new API version should no longer send the `displayName prop and thus renaming with them should also still work.

- https://___.webknossos.xyz

### Steps to test:
- Test whether renaming a dataset via the web UI works
- with wklibs run the following and check if the console shows the success of the renaming:
```python
    with webknossos_context(token="<token>", url="http://localhost:9000"):
        ds = Dataset.open_remote("l4_sample_small")
        print("current ds name", ds.display_name)
        ds.display_name = "blue"
        print("name after renaming", ds.display_name)
```
- run that script on another branch again. Each print should have a name `None` as the `displayName` is not included in the json format of the dataset. This should be fixed in this pr.

### Issues:
- fixes https://scm.slack.com/archives/C5AKLAV0B/p1733346943571389

------

### DIFF
--- a/app/controllers/DatasetController.scala
+++ b/app/controllers/DatasetController.scala
@@ -339,7 +339,7 @@ class DatasetController @Inject()(userService: UserService,
     sil.SecuredAction.async(parse.json) { implicit request =>
       withJsonBodyUsing(datasetPublicReads) {
         case (description, datasetName, legacyDatasetDisplayName, sortingKey, isPublic, tags, metadata, folderId) => {
-          val name = if (datasetName.isDefined) datasetName else legacyDatasetDisplayName
+          val name = if (legacyDatasetDisplayName.isDefined) legacyDatasetDisplayName else datasetName
           for {
             datasetIdValidated <- ObjectId.fromString(datasetId)
             dataset <- datasetDAO.findOne(datasetIdValidated) ?~> notFoundMessage(datasetIdValidated.toString) ~> NOT_FOUND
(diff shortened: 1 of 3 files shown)


app/controllers/DatasetController.scala
```diff
@@ -339,7 +339,7 @@ class DatasetController @Inject()(userService: UserService,
     sil.SecuredAction.async(parse.json) { implicit request =>
       withJsonBodyUsing(datasetPublicReads) {
         case (description, datasetName, legacyDatasetDisplayName, sortingKey, isPublic, tags, metadata, folderId) => {
-          val name = if (datasetName.isDefined) datasetName else legacyDatasetDisplayName
+          val name = if (legacyDatasetDisplayName.isDefined) legacyDatasetDisplayName else datasetName
           for {
             datasetIdValidated <- ObjectId.fromString(datasetId)
             dataset <- datasetDAO.findOne(datasetIdValidated) ?~> notFoundMessage(datasetIdValidated.toString) ~> NOT_FOUND
```
MichaelBuessemeyer, line 345: This fixes the precedence issue described in the pr description

fm3 (approved):

Issue #8097, opened 2024-09-24
Followup for multi-annotation trainings


daniel-wer:
## Detailed Description
#8071 allows starting a job to train a model on multiple annotations, selected by id or uri. Follow-up wishes came up:

- [x] When training a model from within wk, currently all bounding boxes of all annotations need to have the same size. This restricts which training data can be combined. Slight variations in size are usually no big deal and one even sometimes decides to "waste" some training data, because that's better than not using the data at all. Also, this becomes even more relevant with the next point. This will also require changes in the worker which should use the smallest bounding box of all to guide the training_sample_size computation (and related values). https://github.com/scalableminds/voxelytics/pull/3796 and https://github.com/scalableminds/webknossos/pull/8222
- [x] It should be possible to select a resolution for each annotation which is the resolution that will be trained on. Many a time this is not the finest resolution (which is chosen currently). Only mags that exist both in the image data layer and the ground truth layer should be selectable. This is a frontend only issue.

## Context

* [x] Specific to long-running jobs (set `jobsEnabled=true` in `application.conf`)
* [ ] Specific to webknossos.org (set `isDemoInstance=true` in `application.conf`)

daniel-wer:
@dieknolle3333 I updated the issue description to indicate that the first of the two TODOs was implemented already. I did not know this issue existed and was assigned to you. I hope you have not started working on that yet.

For the second TODO, @MatthisCl is implementing the worker side in https://github.com/scalableminds/voxelytics/pull/3833 and it would be great if you could make the mag selectable in the frontend as described in the issue :) To do so, please have a look at what I implemented in https://github.com/scalableminds/webknossos/pull/8222. The warnings about the minimum bounding box size and that dimensions should be multiples will need to take the mag into account if it is different from 1 (as an example, if mag 2 is selected bounding boxes should be at least 20 vx [in mag 1] which corresponds to 10 vx in mag 2). Let me know if you have any questions about that.

dieknolle3333:
@daniel-wer thank you for this update and the additional information. I didn't start working on this issue yet, but I am planning to take it on next :)

dieknolle3333:
>It should be possible to select a resolution for each annotation which is the resolution that will be trained on.

Does this only apply to trainings on multiple annotations?

daniel-wer:
> Does this only apply to trainings on multiple annotations?

It should also be possible when training on a single annotation (or from within an annotation).

PR #8266, opened 2024-12-09
Choose mag when training models on multiple annotations

knollengewaechs:
- https://___.webknossos.xyz

### Steps to test:
- Enable jobs
- Open an annotation and go to AI Analysis > Train a model
- Select a ground truth and a color layer
- Select a mag. Check that all available mags are present in both layers
- Start a training / check network tab that the right mag was sent in the request
- Do the same thing, but for multiple annotations. Do that by navigating to Administration > AI models > Train new model.

- ### TODOs:
- [x] update mags after any of the input layers change 
    - [x] deselect mag
    - [x] move value to upper component
- [x] find union of data and ground truth layer
    - [x] datastructure: save map annotation id -> datalayermags, groundtruthlayermags
- [x] check that it works from within annotation and with multiple annotations
- [x] fix spacing (required vs not required fields in form, see pic)
- [x] address first review
- [ ] regarding testing: does this need to be tested after the request was sent? if yes, how?
- [x] when opening the modal, there is a ground truth layer option for the active segmentation layer, and for the fallback layer. the first one's name is simply the tracing id, which makes it harder to understand what is happening. 
image

### Issues:
- contributes to #8097

------


conf/application.conf
```diff
@@ -151,8 +151,8 @@ features {
   taskReopenAllowedInSeconds = 30
   allowDeleteDatasets = true
   # to enable jobs for local development, use "yarn enable-jobs" to also activate it in the database
-  jobsEnabled = false
-  voxelyticsEnabled = false
+  jobsEnabled = true
+  voxelyticsEnabled = true
```
knollengewaechs, line 155: remove me before merging!!!

frontend/javascripts/oxalis/view/jobs/train_ai_model.tsx
```diff
@@ -311,19 +393,52 @@ export function TrainAiModelTab<GenericAnnotation extends APIAnnotation | Hybrid
...
               <LayerSelectionFormItem
                 name={["trainingAnnotations", idx, "layerName"]}
                 chooseSegmentationLayer
                 layers={segmentationLayers}
                 getReadableNameForLayer={(layer) => {
+                  console.log(layer);
                   return layer.name;
+                  //TODO_c fix that fallback layers are shown at least with the correct name?
+                  // eg. name (fallbacklayer)
+                  // remember to check-in with P
```
knollengewaechs, line 418: also see PR description. to me this feels like a separate issue

frontend/javascripts/components/mag_selection.tsx
```diff
@@ -0,0 +1,73 @@
...
+        // using the index of the mag *in the mag list* as value internally
+        // NB: this is different from the mag index
+        value == null || magInfo == null
+          ? null
+          : clamp(
+              0,
+              allMags.findIndex((v) => V3.equals(v, value)),
+              allMags.length - 1,
+            )
+      }
```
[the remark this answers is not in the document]
knollengewaechs: the value will be found because only present values can be selected

frontend/javascripts/oxalis/view/jobs/train_ai_model.tsx
```diff
@@ -311,19 +393,52 @@ export function TrainAiModelTab<GenericAnnotation extends APIAnnotation | Hybrid
...
                   fixedLayerName={fixedSelectedColorLayer?.name || undefined}
                   style={{ width: "100%" }}
+                  onChange={() => {
+                    setIntersectingMags(
+                      annotationId,
+                      dataset,
+                      form.getFieldValue(["trainingAnnotations", idx, "layerName"]),
+                      form.getFieldValue(["trainingAnnotations", idx, "imageDataLayer"]),
+                    );
+                    form.setFieldValue(["trainingAnnotations", idx, "mag"], undefined);
```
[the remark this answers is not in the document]
knollengewaechs: https://ant.design/components/form says `Use setFieldValue instead if you want to only config single value in Form.List` for `setFieldsValue` so I think its fine the way it is

frontend/javascripts/oxalis/view/jobs/train_ai_model.tsx
```diff
@@ -311,19 +388,50 @@ export function TrainAiModelTab<GenericAnnotation extends APIAnnotation | Hybrid
...
                 fixedLayerName={fixedSelectedSegmentationLayer?.name || undefined}
                 label="Ground Truth Layer"
+                onChange={() => {
+                  setIntersectingMags(
+                    annotationId,
+                    dataset,
+                    form.getFieldValue(["trainingAnnotations", idx, "layerName"]),
+                    form.getFieldValue(["trainingAnnotations", idx, "imageDataLayer"]),
+                  );
+                  form.setFieldValue(["trainingAnnotations", idx, "mag"], undefined);
```
philippotto, line 422: seems to be a copy of the other onchange handler. I'd extract it into an "onChangeLayer" handler.

frontend/javascripts/oxalis/view/jobs/train_ai_model.tsx
```diff
@@ -150,21 +158,78 @@ export function TrainAiModelFromAnnotationTab({ onClose }: { onClose: () => void
...
+    );
+    if (mags == null) {
+      return [{ annotationId, magInfo: new MagInfo(intersectingMags) }];
+    } else {
+      return mags.map((mag) =>
+        mag.annotationId === annotationId
+          ? { annotationId, magInfo: new MagInfo(intersectingMags) }
+          : mag,
+      );
+    }
```
philippotto, line 218: tbh, I don't understand the reasoning behind this code.

philippotto (commented):
sorry, I started reviewing the PR a few days ago and then apparently something else grabbed my attention. I finished the review now and left a few comments.

It's cool that the mag selection seems to work :+1: however, I find the code a bit confusing. I think my main gripe with it is that `mags` can be null and that it is written only on certain events. In my mind, all information should be available at all times and mags could always be computed on demand (memoized if necessary). I understand that antd forms can be annoying, so maybe it doesn't work how I imagine it, but please let's have a call to discuss this :)

knollengewaechs:
I think I addressed all the feedback for now :crossed_fingers:  From my side an open question that is left is how to handle volume annotation layer names (=tracing ids from what I understand) in the selection. Maybe something like this? 
image

frontend/javascripts/oxalis/view/jobs/train_ai_model.tsx
```diff
@@ -250,8 +251,8 @@ export function TrainAiModelTab<GenericAnnotation extends APIAnnotation | Hybrid
       await ensureSavedState();
     }
 
-    await runTraining({
-      trainingAnnotations: await getTrainingAnnotations(values),
```
philippotto: was this on purpose? you re-added the first await again [here](https://github.com/scalableminds/webknossos/pull/8266/commits/63258289b8889f5628300b93c5a3c93d1bd43a0f), but the second one not? I think, both awaits are important?

philippotto (approved):

PR #8268, opened 2024-12-09
Fix already upload of zarr datasets with already existing `datasource-properties.json` file

MichaelBuessemeyer:
This PR fixes the upload of zarr datasets which already have a `datasource-properties.json` file present. The bug originates from (I'd say) wrong prioritization during finishing a dataset upload. First, it was checked whether the uploaded DS is in zarr format and only if that failed, the backend checked if there is a `datasource-properties.json` present (meaning the dataset is already ready to be used). This PR moves the checking for a `datasource-properties.json` first position. Thus, a full dataset with an already existing `datasource-properties.json` is just assumed as already explored an no additional effored it tried to generate a `datasource-properties.json` file for the dataset. 
Previously, the backend would detect such a dataset as zarr and that it needed a `datasource-properties.json` whose contents needs to be guessed. And this is currently not supported by the zarr after upload exploration code. Thus, the uploads failed.

I hope this is the correct way of fixing this. An alternative (which I tried first before locating the wrong ordering of determining the ds format) was to fix the zarr upload exploration. But this created strange datasets -> all mags were interpreted as layers. Which lead to a wrong `datasource-properties.json` and moreover: !! @fm3 The server crashed with a heap overflow when trying to view the dataset. IMO this should never happen in case some `datasource-properties.json` is misconfigured.

- https://___.webknossos.xyz

### Steps to test:
- prepare a single layered zarr dataset with a `datasource-properties.json`
- zip it
- and upload it via the UI
- this should succeed. As well as viewing the data

### Issues:
- fixes https://scm.slack.com/archives/C02H5T8Q08P/p1733738810435789

------

### DIFF
--- a/CHANGELOG.unreleased.md
+++ b/CHANGELOG.unreleased.md
@@ -21,6 +21,7 @@ For upgrade instructions, please check the [migration guide](MIGRATIONS.released
 
 ### Fixed
 - Fixed that listing datasets with the `api/datasets` route without compression failed due to missing permissions regarding public datasets. [#8249](https://github.com/scalableminds/webknossos/pull/8249)
+- Fixed a bug that uploading a zarr dataset with an already existing `datasource-properties.json` file failed. [#8268](https://github.com/scalableminds/webknossos/pull/8268)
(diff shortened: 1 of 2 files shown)


fm3 (approved):
Good find!

Heap overflow doesn’t sound good. Could you create an issue with steps to reproduce (maybe link to such a json as a slack link?)

PR #8271, opened 2024-12-10
Data loading: save mem + cpu by re-using allocated fill-value chunks

### LINKED ISSUE #8269: Backend goes OOM when viewing special datasets
## Context

When viewing this dataset, the backend goes out of memory pretty fast without the user being able to see any data.

## Expected Behavior

The backend should not go out of memory

## Current Behavior

One of the many oom errors:
```
backend: java.lang.OutOfMemoryError: Java heap space
backend: 2024-12-10 10:27:38,249 [error] org.apache.pekko.actor.ActorSystemImpl - Uncaught error from thread [application-pekko.actor.default-dispatcher-36]: Java heap space, shutting down JVM since 'pekko.jvm-exit-on-fatal-error' is enabled for ActorSystem[application]
backend: java.lang.OutOfMemoryError: Java heap space
backend:        at ucar.ma2.ArrayLong.<init>(ArrayLong.java:89)
backend:        at ucar.ma2.ArrayLong$D4.<init>(ArrayLong.java:404)
backend:        at ucar.ma2.ArrayLong$D4.<init>(ArrayLong.java:394)
backend:        at ucar.ma2.ArrayLong.factory(ArrayLong.java:46)
backend:        at ucar.ma2.Array.factory(Array.java:103)
backend:        at ucar.ma2.Array.factory(Array.java:61)
backend:        at com.scalableminds.webknossos.datastore.datareaders.MultiArrayUtils$.createFilledArray(MultiArrayUtils.scala:32)
backend:        at com.scalableminds.webknossos.datastore.datareaders.ChunkTyper.createFromFillValue(ChunkTyper.scala:29)
backend:        at com.scalableminds.webknossos.datastore.datareaders.ChunkReader.$anonfun$read$5(ChunkReader.scala:33)
backend:        at com.scalableminds.webknossos.datastore.datareaders.ChunkReader$$Lambda$11110/0x0000000842de1840.apply(Unknown Source)
backend: Uncaught error from thread [application-pekko.actor.default-dispatcher-76]: Java heap space, shutting down JVM since 'pekko.jvm-exit-on-fatal-error' is enabled for ActorSystem[application]
``` 

## Steps to Reproduce the bug

1. Add the linked dataset to some orga's folder
2. rescan the DS in the dashboard and then view the DS :boom:

fm3:
Dataset arrays can define a fill value. When chunks don’t exist in a location, a virtual chunk filled with that fill value is created and used.

This PR introduces a single-entry cache for this fill value chunk (per array/mag), so that we don’t have to create thousands of them in datasets where a lot of chunks are missing. Most datasets have a uniform chunk shape (and even those that don’t usually only have a few deviating chunks at the border), so this single-entry cache allows re-using the same allocated fill-value chunk many times.

Before this PR, a test dataset with a flat chunk shape, a large bounding box, and very sparse existing chunks would lead to so many all-black all-identical chunks being allocated after hitting “find data” that the GC was overwhelmed. Also, the allocation and GC take time so that the route took multiple minutes. Both problems are gone now.

“find data” is still not super fast in this case, but that’s somewhat expected due to the almost 4000 sampled positions and the bad chunk shape that makes many partial array copys necessary. It’s definitely better than before and uses way less memory.

### Steps to test:
- Open test dataset from https://scm.slack.com/archives/C02H5T8Q08P/p1733833450557419?thread_ts=1733738810.435789&cid=C02H5T8Q08P
- Hit “Jump to data” in the layer menu. It should fail (as there is no data in the sampled positions), but it should not crash the server or take more than two minutes.

### Issues:
- fixes #8269

------

### DIFF
--- a/CHANGELOG.unreleased.md
+++ b/CHANGELOG.unreleased.md
@@ -12,6 +12,7 @@ For upgrade instructions, please check the [migration guide](MIGRATIONS.released
 
 ### Added
 - Added the total volume of a dataset to a tooltip in the dataset info tab. [#8229](https://github.com/scalableminds/webknossos/pull/8229)
+- Optimized performance of data loading with “fill value“ chunks. [#8271](https://github.com/scalableminds/webknossos/pull/8271)
 
 ### Changed
 - Renamed "resolution" to "magnification" in more places within the codebase, including local variables. [#8168](https://github.com/scalableminds/webknossos/pull/8168)
--- a/webknossos-datastore/app/com/scalableminds/webknossos/datastore/datareaders/ChunkReader.scala
+++ b/webknossos-datastore/app/com/scalableminds/webknossos/datastore/datareaders/ChunkReader.scala
@@ -26,11 +26,11 @@ class ChunkReader(header: DatasetHeader) {
         case Full(chunkBytes) if useSkipTypingShortcut =>
           shortcutChunkTyper.wrapAndType(chunkBytes, chunkShape).toFox ?~> "chunk.shortcutWrapAndType.failed"
         case Empty if useSkipTypingShortcut =>
-          shortcutChunkTyper.createFromFillValue(chunkShape).toFox ?~> "chunk.shortcutCreateFromFillValue.failed"
+          shortcutChunkTyper.createFromFillValueCached(chunkShape).toFox ?~> "chunk.shortcutCreateFromFillValue.failed"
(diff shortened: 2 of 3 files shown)


MichaelBuessemeyer (approved):
Thanks a lot for the effort to investigate this @fm3 :pray:  

Your changes lgtm and testing also worked well / as described (a failing find my data request)

PR #8273, opened 2024-12-12
Only show locked banner if current user is in organization of the annotation

### LINKED ISSUE #7934: Do not show locked banner if backend sends restrictions.allowUpdate=false
## Context

When an annotation with `othersMayEdit` turned on is shared publically and another user from a different organization opens the annotation a locked banner is shown. This might be confusing as the annotation is not locked but the user is not allowed to edit this annotation due to being from another organization (see #7923). 

## Expected Behavior

No logged banner should be rendered.
## Current Behavior

A logged banner is rendered. See https://github.com/scalableminds/webknossos/pull/7923#issuecomment-2233712513.

## Steps to Reproduce the bug

- Same as https://github.com/scalableminds/webknossos/pull/7923#issue-2411083446

## Your Environment for bug

- Version of WEBKNOSSOS (Release or Commit):

MichaelBuessemeyer:
And do not show it when the annotation is from another organization

- https://___.webknossos.xyz

### Steps to test:
- create another organization in a new context browser tab (e.g. private window) and stay logged in. 
- As a logged-in user of the sample_orga create a new annotation and make it public & `othersMayEdit` in the sharing settings.
- Copy the link into the private browser tab / open it with the other user. There shouldn't be a locked banner.
- confirm bug: Checkout master and reload the link as the user from the 2nd orga. A locked banner "locked by anonymous user" 

### Issues:
- fixes #7934

------


frontend/javascripts/oxalis/model/reducers/reducer_helpers.ts
```diff
@@ -95,6 +95,7 @@ export function convertServerAnnotationToFrontendAnnotation(annotation: APIAnnot
     tracingStore,
     owner,
     contributors,
+    organization,
```
MichaelBuessemeyer, line 98: apparently the `organization` prop was already sent by the backend but the frontend discarded this.
-> I fixed this here.

Side note: But there might also be other fields the frontend drops in this function which might be useful to still have :thinking:.

philippotto: > But there might also be other fields the frontend drops in this function which might be useful to still have 🤔.

if the frontend doesn't use these properties, there's no need to add them, I think.

philippotto (approved):
thank you for the fix!

PR #8277, opened 2024-12-13
Fix nml upload with only ds name in nml by using user's orga as fallback

MichaelBuessemeyer:
This PR fixes uploading an nml with only setting a dataset name in the nml and not passing the optional `datasetId` to the route. Wklibs currently behaves like this. And therefore this should be fixed.

- https://___.webknossos.xyz

### Steps to test:
-Just drop this nml in the dashboard. This sends the upload without a `datasetId`. The upload should succeed
[test_annotation.zip](https://github.com/user-attachments/files/18129451/test_annotation.zip)

### Issues:
- No issue or so exists for this.

------

### DIFF
--- a/app/models/annotation/AnnotationUploadService.scala
+++ b/app/models/annotation/AnnotationUploadService.scala
@@ -27,6 +27,7 @@ case class UploadedVolumeLayer(tracing: VolumeTracing, dataZipLocation: String, 
 
 case class SharedParsingParameters(useZipName: Boolean,
                                    overwritingDatasetId: Option[String] = None,
+                                   userOrganizationId: String,
                                    isTaskUpload: Boolean = false)
 
 class AnnotationUploadService @Inject()(tempFileService: TempFileService, nmlParser: NmlParser) extends LazyLogging {
@@ -53,8 +54,7 @@ class AnnotationUploadService @Inject()(tempFileService: TempFileService, nmlPar
       nmlParser.parse(
         name,
         inputStream,
-        sharedParsingParameters.overwritingDatasetId,
-        sharedParsingParameters.isTaskUpload,
+        sharedParsingParameters,
         basePath
       )
     parserOutput.futureBox.map {
--- a/app/models/annotation/nml/NmlParser.scala
+++ b/app/models/annotation/nml/NmlParser.scala
@@ -24,7 +24,7 @@ import com.scalableminds.webknossos.tracingstore.tracings.ColorGenerator
 import com.scalableminds.webknossos.tracingstore.tracings.skeleton.updating.TreeType
 import com.scalableminds.webknossos.tracingstore.tracings.skeleton.{MultiComponentTreeSplitter, TreeValidator}
 import com.typesafe.scalalogging.LazyLogging
-import models.annotation.UploadedVolumeLayer
+import models.annotation.{SharedParsingParameters, UploadedVolumeLayer}
 import models.dataset.DatasetDAO
 import net.liftweb.common.Box._
 import net.liftweb.common.{Box, Empty, Failure, Full}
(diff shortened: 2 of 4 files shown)


app/models/annotation/AnnotationUploadService.scala
```diff
@@ -53,8 +54,7 @@ class AnnotationUploadService @Inject()(tempFileService: TempFileService, nmlPar
       nmlParser.parse(
         name,
         inputStream,
-        sharedParsingParameters.overwritingDatasetId,
-        sharedParsingParameters.isTaskUpload,
```
MichaelBuessemeyer, line 57: all members of sharedParsingParameters are passed to `nmlParser.parse` anyway. So just pass the whole `sharedParsingParameters` object

app/models/annotation/nml/NmlParser.scala
```diff
@@ -379,8 +380,10 @@ class NmlParser @Inject()(datasetDAO: DatasetDAO) extends LazyLogging with Proto
   private def parseWkUrl(nodes: NodeSeq): Option[String] =
     nodes.headOption.map(node => getSingleAttribute(node, "wkUrl"))
 
-  private def parseOrganizationId(nodes: NodeSeq): String =
-    nodes.headOption.map(node => getSingleAttribute(node, "organization")).getOrElse("")
+  private def parseOrganizationId(nodes: NodeSeq, fallbackOrganization: String): String =
+    nodes.headOption
+      .map(node => getSingleAttributeOpt(node, "organization").getOrElse(fallbackOrganization))
+      .getOrElse(fallbackOrganization)
```
MichaelBuessemeyer, line 386: in both cases 
- the "experiment" tag does not exist 
- in case the experiment tag has no "organization" field 
the fallback organization (user orga) is used.

app/models/annotation/nml/NmlParser.scala
```diff
@@ -379,8 +380,10 @@ class NmlParser @Inject()(datasetDAO: DatasetDAO) extends LazyLogging with Proto
   private def parseWkUrl(nodes: NodeSeq): Option[String] =
     nodes.headOption.map(node => getSingleAttribute(node, "wkUrl"))
 
-  private def parseOrganizationId(nodes: NodeSeq): String =
-    nodes.headOption.map(node => getSingleAttribute(node, "organization")).getOrElse("")
+  private def parseOrganizationId(nodes: NodeSeq, fallbackOrganization: String): String =
+    nodes.headOption
+      .map(node => getSingleAttributeOpt(node, "organization").getOrElse(fallbackOrganization))
```
normanrz, line 385:
```suggestion
      .flatMap(node => getSingleAttributeOpt(node, "organization"))
```
would a flstMap work here?

normanrz (approved):

app/models/annotation/nml/NmlParser.scala
MichaelBuessemeyer: Yeah, good suggestion. Should work :tada:

test/backend/NMLUnitTestSuite.scala
```diff
@@ -54,8 +55,10 @@ class NMLUnitTestSuite @Inject()(nmlParser: NmlParser) extends PlaySpec {
         .parse(
           "",
           new ByteArrayInputStream(array),
-          overwritingDatasetId = None,
-          isTaskUpload = true,
+          SharedParsingParameters(useZipName = false,
+                                  overwritingDatasetId = None,
+                                  userOrganizationId = "testOrganization",
+                                  isTaskUpload = true),
```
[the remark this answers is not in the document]
MichaelBuessemeyer: @fm3 Writing such a test is rather difficult as the nmls for the tests are generated with `NmlWriter().toNmlStream` which requires an organization id. This would mean to rewrite `toNmlStream` which is not preferable as one should always supply the `toNmlStream` with an organization. But writing a copy of `toNmlSteam` that supports an optional organizationId is also not preferable :/

=> I skipped this for now but it should be useful for the future to avoid this regression bug introduced by https://github.com/scalableminds/webknossos/pull/8075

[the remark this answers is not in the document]
MichaelBuessemeyer: what are your thoughts?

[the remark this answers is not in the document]
fm3: I’d say this is part of the broader problem of testing the libs and webknossos together. That’s a topic we’ll definitely discuss next year.

PR #8285, opened 2024-12-16
Use ObjectId directly as parameter in routes

### LINKED ISSUE #8158: Can ObjectId be used as GET-Parameter directly?
Many routes implementations in our backend start with `ObjectId.fromString(aPassedIdString)`.
Maybe an implicit conversion can be done by play to directly support ObjectId as a GET parameter type?

If the returned error looks different then, we should check how the frontend displays it.

Should probably be tackled after #8075 to minimize merge conflicts

Since ObjectId is moved to util or datastore there anyway, we should also use it everywhere in datastore/tracingstore where objectIds are handled (currently they are only passed around as strings there in some places)

frcroth:
- https://___.webknossos.xyz

### Steps to test:
- I think this is mostly covered by the compiler, however I tested a few routes manually (though not all)

This does not cover all cases. I tried to be exhaustive with annotationId and tracingIds but it just didn't end and did not push the progress here (a lot in tracing store, update actions...) . That would be a separate issue, to use ObjectId consistently wherever it is semantically correct, but this PR is already big enough.

### Issues:
- fixes #8158 

------
- [ ] ~~Updated [changelog](../blob/master/CHANGELOG.unreleased.md#unreleased)~~


frcroth:
There seems to be a problem as seen in e2e tests

fm3 (approved):
love it 🎉

PR #8286, opened 2024-12-16
Implement option to export proofreading as segmentation

cdfhalle:
### Steps to test:
- In a freshly created Annotation (with agglomerate file mapping) use the proofreading tool and merge the volume annotation with its fallback layer (by clicking onto the three dots next to the desired segmentation layer)
- In the resulting Dataset the changes should be applied
- Other ways of materializing volume annotations should still work as before and could also be tested

[corresponding voxelytics pr](https://github.com/scalableminds/voxelytics/pull/3859)

------


CHANGELOG.unreleased.md
```diff
@@ -13,6 +13,7 @@ For upgrade instructions, please check the [migration guide](MIGRATIONS.released
 ### Added
 - Added the total volume of a dataset to a tooltip in the dataset info tab. [#8229](https://github.com/scalableminds/webknossos/pull/8229)
 - Optimized performance of data loading with “fill value“ chunks. [#8271](https://github.com/scalableminds/webknossos/pull/8271)
+- Added the option to export proofreading as segmentation [#8286](https://github.com/scalableminds/webknossos/pull/8286)
```
```suggestion
- Added the option to export a segmentation that was corrected with the proofreading tool to a new segmentation. [#8286](https://github.com/scalableminds/webknossos/pull/8286)
```

frontend/javascripts/oxalis/view/action-bar/starting_job_modals.tsx
```diff
@@ -925,6 +926,8 @@ export function MaterializeVolumeAnnotationModal({
         output dataset and the output segmentation layer.
       </p>
     );
+  } else {
+    includesProofreading = tracing.volumes.some((v) => v.hasEditableMapping === true);
```
philippotto, line 930: If my annotation contains two volume layers and only one has an editable mapping, that property will be true. However, i I want to export the layer that has no editable mapping, `true` will still be used in this UI and sent to the backend, which seems wrong?

frontend/javascripts/oxalis/view/action-bar/starting_job_modals.tsx
```diff
@@ -954,8 +957,13 @@ export function MaterializeVolumeAnnotationModal({
         jobName={"materialize_volume_annotation"}
         suggestedDatasetSuffix="with_merged_segmentation"
         chooseSegmentationLayer
+        isBoundingBoxConfigurable={includesProofreading}
```
philippotto, line 960: only proofread layers support configuring the bbox? would it be easy to always support it when exporting volume annotations? no prio though.

frontend/javascripts/admin/api/jobs.ts
```diff
@@ -208,6 +208,8 @@ function startSegmentationAnnotationDependentJob(
   annotationId: string,
   annotationType: APIAnnotationType,
   mergeSegments?: boolean,
+  includesProofreading?: boolean,
+  selectedBoundingBox?: Vector6,
```
philippotto, line 212: "selected" sounds as if this variable would refer to a bounding box (ID) selected in the UI. I'd rename it to simply `boundingBox`.

frontend/javascripts/admin/api/jobs.ts
```diff
@@ -222,6 +224,12 @@ function startSegmentationAnnotationDependentJob(
   if (mergeSegments != null) {
     requestURL.searchParams.append("mergeSegments", mergeSegments.toString());
   }
+  if (includesProofreading != null) {
+    requestURL.searchParams.append("includesProofreading", includesProofreading.toString());
+  }
+  if (selectedBoundingBox) {
+    requestURL.searchParams.append("selectedBoundingBox", selectedBoundingBox.toString());
```
philippotto, line 231:
```suggestion
    requestURL.searchParams.append("selectedBoundingBox", selectedBoundingBox.join(","));
```

should be identical, but in your variant one has to know how JS serializes number arrays. I find it better to be explicit about it.

fm3 (commented):
I took the liberty to add a commit for the backend to make the linter happy, and remove a no-op line (no bbox was parsed there).

Backend LGTM now :)

One thing: in the backend code, we usually speak of “editable mapping” rather than “proofreading”. @philippotto can you comment how this is handled in the frontend? Maybe we can unify the naming for the new job parameters as well.

fm3:
Alternatively, it might make sense to call the parameter use_zarr_streaming. and set this one to true if the frontend has an editable mapping for this layer. This way we don’t have to adapt the worker if the logic will change eventually. Long-term, we might be able to switch more than just proofreading annotations to the zarr streaming :thinking:

philippotto:
> Alternatively, it might make sense to call the parameter use_zarr_streaming. and set this one to true if the frontend has an editable mapping for this layer. This way we don’t have to adapt the worker if the logic will change eventually. Long-term, we might be able to switch more than just proofreading annotations to the zarr streaming 🤔

Yes, sounds good, too.

MichaelBuessemeyer:
@cdfhalle @philippotto Does anyone of you know why I am assigned to this PR?

philippotto:
> @cdfhalle @philippotto Does anyone of you know why I am assigned to this PR?

I don't. Maybe @cdfhalle wanted to pass this PR on to you?

frontend/javascripts/oxalis/view/action-bar/starting_job_modals.tsx
cdfhalle: The question then would be what the consequences of this bounding boxes should be. We are currently using Zarr Streaming (where one can specify a bounding box that should be streamed) for merging the annotations only when the annotation includes proofreading. In other cases the worker currently just downloads the whole annotation because that has proven to be faster than streaming it. I could not find an easy option to specify a bouding box in this case. However It would be easily possible to use the streaming approach whenever a bounding box was specified. This could then be faster or slower depending on the relationship of the dataset size and the size of the selected bounding boxes.

philippotto: > In other cases the worker currently just downloads the whole annotation because that has proven to be faster than streaming it.

this could stay as is, I think. after downloading, one could only read the bbox from the annotation. however, as I said, it doesn't have a high priority. so, if it's not a quick thing to do, nevermind :)

philippotto (commented):
frontend lgtm :+1: leaving final approval to @fm3.

fm3 (approved):

PR #8289, opened 2024-12-18
Fix screenshot tests

MichaelBuessemeyer:
- https://___.webknossos.xyz

### Issues:
- related thread: https://scm.slack.com/archives/C5AKLAV0B/p1734365760342809


frontend/javascripts/test/puppeteer/dataset_rendering.screenshot.ts
```diff
@@ -114,7 +114,7 @@ test.before("Retrieve dataset ids", async () => {
       3,
       async () => {
         const options = getDefaultRequestOptions(URL);
-        const url = `${URL}/api/datasets/disambiguate/sample_organization/${datasetName}/toId`;
+        const url = `${URL}api/datasets/disambiguate/sample_organization/${datasetName}/toId`;
```
philippotto, line 117: the rest of the screenshot tests use `urljoin` to join url parts. that way, double slashes can also be avoided. I'd rather do this then relying on a trailing slash.

MichaelBuessemeyer: Oh I overlooked that. Thanks for the suggestion. It should be applied now. 

I also did a quick test that requesting the URLs still works

philippotto (approved):
awesome, thank you!!

PR #8292, opened 2024-12-18
Refactor backend: Use “folder” only for dashboard folders, otherwise “directory”

### LINKED ISSUE #7780: Be strict about Folder (only for in-db dashboard folders) vs Directory (for actual filesystem)
There are a couple of spots where we use the terms interchangeably. Not a huge issue, but I think it would be nice to use the words strictly as in the issue title, in order to immediately know what to look for. 

This also concerns the config options datastore.baseFolder and datastore.localFolderWhitelist

fm3:
Unified terminology: we decided a while ago that “folder” should refer only to the dashboard dataset folders that are implemented in postgres, while “directory” should refer to the actual filesystem. This PR adapts the backend config and code to follow this terminology.

### Steps to test
 - Load datasets, create annotations
 - Upload a dataset
 - Test zarr streaming (e.g. self-streaming via “add remote dataset”)

### Issues:
- fixes #7780 

------


frcroth (approved):
LGTM

Issue #1207, opened 2016-05-18
Create node button creates node at wrong position


tmbo:
Clicking the create node button moves the viewport to (x+1, y+1,z) and sets the point there. Point should be set at the current location.
![bildschirmfoto 2016-05-18 um 15 46 03](https://cloud.githubusercontent.com/assets/1098412/15360866/a80b6c62-1d0f-11e6-9413-191af4621ee1.png)

<a href="https://timer.scm.io/repos/537237ce1a00001a003c4a14/issues/1207/create?referer=github" target="_blank">Log Time</a>

PR #8293, opened 2024-12-18
Remove unused legacy routes

MichaelBuessemeyer:
This PR removes some legacy routes accidentally implemented for #8075's backward compatibility. 
1. The `PUT /v8/tasks/:id controllers.LegacyApiController.updateTaskV8(id: String)` route was never used by wklibs.
2. The `dataSet` field was renamed to `dataset_name` in task object which may be attached to annotations, but wklibs ignores these potential existing tasks of an annotation. Therefore the other two legacy adaptation routes can be removed

- https://___.webknossos.xyz

### Steps to test:
- abc

### TODOs:
- [ ] ...

### Issues:
- None really. Noticed while working on #1207

------


fm3 (approved):
:tada:

PR #8296, opened 2024-12-20
Remove isScratch banner

philippotto:
Hasn't been used in 5 years.


fm3 (approved):
LGTM and works for me too :) Thanks!

Could you add a changelog entry?

frontend/javascripts/oxalis/view/layouting/tracing_layout_view.tsx
```diff
@@ -401,7 +374,6 @@ function mapStateToProps(state: OxalisState) {
     isUpdateTracingAllowed: state.tracing.restrictions.allowUpdate,
     showVersionRestore: state.uiInformation.showVersionRestore,
     storedLayouts: state.uiInformation.storedLayouts,
-    isDatasetOnScratchVolume: state.dataset.dataStore.isScratch,
```
MichaelBuessemeyer, line 404: Maybe also remove the property from the frontend as this is the onliest place where it is actively used?

philippotto: then, we also have to remove it from the backend's output. @fm3 can we simply do that? or is there a point against that? compatibility with wklibs?

fm3: I removed it from the json after checking that the libs [don’t use the field](https://github.com/scalableminds/webknossos-libs/blob/006083fcdc0c4b701d8bc02a8a9037d40b06fdd0/webknossos/webknossos/client/api_client/models.py#L31) (fortunately since https://github.com/scalableminds/webknossos-libs/pull/948 the libs client only expects the fields that are actually used :tada: )

PR #8297, opened 2024-12-23
Omit mag slider in TIFF export if only one mag is available

### LINKED ISSUE #8002: Disallow selecting mag in tiff export modal in case there is only one mag
## Context

When exporting a layer as tiff one can select the mag to export. However, when there is only one mag, a draggable slider is still rendered but the user cannot slide it. Therefore I suggest to hide this slider setting in case the current selected layer only has one mag.

## Expected Behavior

IT should either be possible to interact with the slider when it is rendered or when its not don't render it.

## Current Behavior

The slider is also rendered when a layer is selected that only has a single mag.

## Steps to Reproduce the bug

1. Create a dataset with a layer with only one mag
2. Opent the download modal and go to the tiff export tab
3. Try to mode the mag slider. This should not work

knollengewaechs:
image

- https://tiffexporthidesliderforonlyonemag.webknossos.xyz/

### Steps to test:
- Open an annotation with a layer that has more than one mag and a layer with only one mag
- Go to Menu > Download and open the TIFF download
- If you select a layer with one mag, there should not be a mag slider and only the current mag should be displayed
- If you select a layer with more than one mag, the mag slider should still be displayed
- Next to that, only annotations with actual additional coordinates should show a section about the chosen additional coords 

### TODOs:
- [x] ...

### Issues:
- fixes #8002

------


frontend/javascripts/oxalis/view/action-bar/download_modal_view.tsx
```diff
@@ -670,23 +672,27 @@ function _DownloadModalView({
...
+            </Row>
+          )}
           <Text
             style={{
               margin: "0 6px 12px",
               display: "block",
             }}
           >
+            {onlyOneMagAvailable && mag.join("-")}
+            <br />
```
MichaelBuessemeyer, line 695: This added `<br>` also adds a newline after the slides which leads to additional vertical spacing which is not needed imo.
E.g. using 
```suggestion
            {onlyOneMagAvailable && <div>{mag.join("-")}</div>}

```
would avoid this.

MichaelBuessemeyer (approved):
Awesome, thanks for solving this. I only found a very minor nitpick.

Should be ready to be shipped :shipit:

frontend/javascripts/oxalis/view/action-bar/download_modal_view.tsx
knollengewaechs: nice, thanks for catching this!

PR #8298, opened 2024-12-30
Call triggers/reload with fallback layer name

### LINKED ISSUE #8087: When reloading volume layer, frontend calls datastore reload route with tracingid
In an annotation with a volume annotation layer, the reload button will call the datastore triggers/reload route, but with the tracingId as layer name. The datastore does not know the tracingId.

Instead, the frontend could call the route with the volume layer’s fallback layer name (if it exists).

knollengewaechs:
- https://___.webknossos.xyz

### Steps to test:
- Open an annotation with a segmentation layer and brush anything
- In the "layers" tab on the left, select a color layer and hit "reload data from server". you should see how the buckets are reloaded.
- Do the same for the segmentation layer you just brushed in. you should see how the buckets are reloaded. in the network tab, there should be no requests with status 400 (before there was an unsuccessful request to `/histograms`)
- In your local test instance, change the name on disk of the current segmentation layer.
- Reload the layer. The segmentation should now be black. Before, old cached data would have been reloaded. [see this conversation in the original issue](https://github.com/scalableminds/webknossos/issues/8087#issuecomment-2566277991)

### TODOs:
- [x] test for segmentation layers without fallback layers

### Issues:
- fixes #8087 
- fixes that `/histogram` route was called for non-color layers from "reload data from server" button

------


frontend/javascripts/oxalis/view/left-border-tabs/layer_settings_tab.tsx
```diff
@@ -994,9 +1003,7 @@ class DatasetSettings extends React.PureComponent<DatasetSettingsProps, State> {
               marginLeft: 10,
             }}
           >
-            {isHistogramSupported(elementClass) && layerName != null && isColorLayer
```
knollengewaechs, line 997: ??

CHANGELOG.unreleased.md
```diff
@@ -36,6 +36,7 @@ For upgrade instructions, please check the [migration guide](MIGRATIONS.released
 - Fixed some layout issues in the upload view. [#8231](https://github.com/scalableminds/webknossos/pull/8231)
 - Fixed `FATAL: role "postgres" does not exist` error message in Docker compose. [#8240](https://github.com/scalableminds/webknossos/pull/8240)
 - Fixed the Zarr 3 implementation not accepting BytesCodec without "configuration" key. [#8282](https://github.com/scalableminds/webknossos/pull/8282)
+- Fixed that reloading a layer's data from the server was sending the tracing ID instead of the fallback layer name to the datastore. [#8298](https://github.com/scalableminds/webknossos/pull/8298)
```
philippotto, line 39:
```suggestion
- Fixed that reloading the data of a volume annotation layer did not work properly. [#8298](https://github.com/scalableminds/webknossos/pull/8298)
```

this is more more precise regarding the "when" and less precise about the "why" which is better for a changelog imo.

frontend/javascripts/oxalis/view/left-border-tabs/layer_settings_tab.tsx
```diff
@@ -976,6 +988,8 @@ class DatasetSettings extends React.PureComponent<DatasetSettingsProps, State> {
       );
 
     const defaultLayerViewConfig = getDefaultLayerViewConfiguration();
+    const isHistogramAvailable =
+      isHistogramSupported(elementClass) && layerName != null && isColorLayer;
```
philippotto, line 992: I know, it's merely moved, but still: is layerName not always not-null? I think that part can be removed from the condition.

philippotto (commented):
great, I only have two nitpicks :)

philippotto (approved):
:tada:

PR #8303, opened 2025-01-06
WIP faster mapped mesh loading

fm3:
- https://___.webknossos.xyz

### Steps to test:
- abc

### TODOs:
- [ ] use existing alfucache?
- [ ] cache buckets too?
- [ ] neuroglancer array seems too large to sensibly cache. Do other tricks like request sorting work?
- [ ] force sequential chunk requests?
- [ ] do the hpc file systems behave like a single spinning disk? what about block size? how about parallel access?

### Issues:
- fixes #

------


fm3:
Did not really achieve the goal here. The performance improvements are small, and to make this production-ready would still be some work.

I think we need to store oversegmentation meshfiles differently, or have some service precompute meshes for agglomerates.

PR #8311, opened 2025-01-08
Support OME-NGFF zarr dataset translation

### LINKED ISSUE #6600: OME NGFF coordinateTransformations translate
## Context

Tried importing OME NGFF sample at https://uk1s3.embassy.ebi.ac.uk/idr/zarr/v0.4/idr0101A/13457537.zarr
This has a 'translation' in the datasets coordinateTransformations:

```
"datasets": [
                {
                    "coordinateTransformations": [
                        {
                            "scale": [
                                1.0,
                                1.0,
                                0.4,
                                0.108335,
                                0.108335
                            ],
                            "type": "scale"
                        },
                        {
                            "translation": [
                                0,
                                0,
                                1.2,
                                52.109135,
                                60.88427
                            ],
                            "type": "translation"
                        }
                    ],
                    "path": "0"
                },
```

## Expected Behavior
The image should at least be viewable. Ideally, the `translation` should also be applied, to offset the image with respect to other images in the same coordinate system

## Current Behavior
Fails to import. It seems that each `coordinateTransformations` item is expected to have a `scale` attribute:

```
Trying to explore https://uk1s3.embassy.ebi.ac.uk/idr/zarr/v0.4/idr0101A/13457537.zarr as OME NGFF Zarr v0.4...
Error when reading https://uk1s3.embassy.ebi.ac.uk/idr/zarr/v0.4/idr0101A/13457537.zarr as OME NGFF Zarr v0.4: Failed to read OME NGFF header at https://uk1s3.embassy.ebi.ac.uk/idr/zarr/v0.4/idr0101A/13457537.zarr/.zattrs <~ Failed to validate json against data schema <~ Validating Json Failed: {"obj.multiscales[0].datasets[0].coordinateTransformations[1].scale":[{"msg":["error.path.missing"],"args":[]}],"obj.multiscales[0].datasets[1].coordinateTransformations[1].scale":[{"msg":["error.path.missing"],"args":[]}],"obj.multiscales[0].datasets[2].coordinateTransformations[1].scale":[{"msg":["error.path.missing"],"args":[]}]}
```

## Steps to Reproduce the bug
Try to import the image above

## Your Environment for bug

- Chrome
- Version of WebKnossos (Release or Commit): 20071

frcroth:
- https://___.webknossos.xyz

### Steps to test:
- Explore https://uk1s3.embassy.ebi.ac.uk/idr/zarr/v0.4/idr0101A/13457537.zarr/.zattrs **AND** https://uk1s3.embassy.ebi.ac.uk/idr/zarr/v0.4/idr0101A/13457227.zarr/.zattrs in one dataset (Use Add Layer)
- Set z=10, t=17
- Notice that the smaller layer (13457537) is displayed on top of one of the thingis

### TODOs:
- [ ] Before merging: Reenable label layers

### Issues:
- fixes #6600

------


webknossos-datastore/app/com/scalableminds/webknossos/datastore/datareaders/zarr/NgffMetadata.scala
```diff
@@ -36,8 +36,11 @@ object NgffMetadata {
       mag =>
         NgffDataset(
           path = mag.toMagLiteral(allowScalar = true),
-          List(NgffCoordinateTransformation(
-            scale = Some(List[Double](1.0) ++ (dataSourceVoxelSize.factor * Vec3Double(mag)).toList)))
+          List(
+            NgffCoordinateTransformation(
+              scale = Some(List[Double](1.0) ++ (dataSourceVoxelSize.factor * Vec3Double(mag)).toList),
+              translation = Some(List[Double](0.0, 0.0, 0.0))
```
fm3, line 42: Why add a zeroed translation by default? Is this in any way better than leaving out the translation?

webknossos-datastore/app/com/scalableminds/webknossos/datastore/explore/NgffExplorationUtils.scala
```diff
@@ -259,6 +261,42 @@ trait NgffExplorationUtils extends FoxImplicits {
       layerAndVoxelSizeTuples = layers.map((_, VoxelSize(voxelSizeFactor, unifiedAxisUnit)))
     } yield layerAndVoxelSizeTuples
 
+  protected def getTranslation(multiscale: NgffMultiscalesItem): Option[List[CoordinateTransformation]] = {
+    val is2d = !multiscale.axes.exists(_.name == "z")
+    val baseTranslation = if (is2d) List(1.0, 1.0) else List(1.0, 1.0, 1.0)
+    if (!multiscale.datasets.head.coordinateTransformations.exists(_.`type` == "translation")) {
```
fm3, line 267: This will throw exceptions if multiscale.datasets is empty. Could we use headOption, with the whole function returning None if that is the case?

webknossos-datastore/app/com/scalableminds/webknossos/datastore/explore/NgffV0_4Explorer.scala
```diff
@@ -34,7 +34,7 @@ class NgffV0_4Explorer(implicit val ec: ExecutionContext) extends RemoteLayerExp
           } yield layers
         })
       layers: List[(DataLayerWithMagLocators, VoxelSize)] = layerLists.flatten
-    } yield layers ++ labelLayers
+    } yield layers //++ labelLayers
```
fm3, line 37: just a reminder to re-add this before merging

webknossos-datastore/app/com/scalableminds/webknossos/datastore/explore/ExploreLayerUtils.scala
```diff
@@ -50,8 +50,8 @@ trait ExploreLayerUtils extends FoxImplicits {
                                                    preferredVoxelSize: Option[VoxelSize],
                                                    voxelSize: VoxelSize): List[DataLayerWithMagLocators] =
     layers.map(l => {
-      val coordinateTransformations = coordinateTransformationForVoxelSize(voxelSize, preferredVoxelSize)
-      l.mapped(coordinateTransformations = coordinateTransformations)
+      val generatedCoordinateTransformation = coordinateTransformationForVoxelSize(voxelSize, preferredVoxelSize)
+      l.mapped(coordinateTransformations = generatedCoordinateTransformation.orElse(l.coordinateTransformations))
```
fm3, line 54: I’m not sure about this orElse. Could we not have a generatedCoordinateTransformation _and_ a relevant `l.coordinateTransformations` so that we would have to combine this somehow? Or is this case not relevant?

fm3 (commented):
Looking pretty good already :) I added a couple of questions

webknossos-datastore/app/com/scalableminds/webknossos/datastore/datareaders/zarr/NgffMetadata.scala
frcroth: Actually None is better because the type by default is scale, so this would add translation to a scale which does not make a lot of sense

webknossos-datastore/app/com/scalableminds/webknossos/datastore/explore/NgffExplorationUtils.scala
frcroth: Ok, is there a reasonable case where datasets would be empty?

fm3: Not really, I’d just prefer the scala-esque handling of the empty case over exceptions.

webknossos-datastore/app/com/scalableminds/webknossos/datastore/explore/ExploreLayerUtils.scala
fm3: Lack of test datasets is of course not great :/ I don’t know of any either. Maybe it’s fair to just write a comment that this logic might not cover all cases.
However, maybe we can think about different hypothetical cases. What’s the current semantics here? the coordinateTransformationForVoxelSize is only a scale, right? When is it set? Because it looks like in that case the coordinate transformations from the metadata are not used at all? Maybe it would make sense to use the scale from the voxel size but the translation from the layer metadata? These could just be put into the matrix at the right spot, right?
I haven’t worked on this in a while so my current understanding is very rudimentary.

frcroth: We could add the translation, since the coordinateTransformations after. I am just not sure if it would work then because it's not clear to me intuitively if the translation e.g. would need to be scaled before by something.

fm3: Ok then maybe let’s do this and write in a comment about this uncertainty. So later, when something looks off with a dataset with translation, we directly have a place to start looking.

fm3 (approved):
:tada: (don’t forget to re-add label exploring)

PR #8312, opened 2025-01-08
Allow to specify mask during inference when using a custom workflow

daniel-wer:
See https://github.com/scalableminds/voxelytics/pull/3872 for context. (Needs to be deployed together)

Also the template was currently incomplete due to https://github.com/scalableminds/voxelytics/pull/3800 which I fixed by adding the `datastore_url` to the template (/cc @MatthisCl).

- https://___.webknossos.xyz

### Steps to test:
- I tested this
- Can be tested using a local worker (voxelytics branch `use-mask-from-annotation`) by adding a mask config to the template (under the `datasource_config` key) when doing neuron inferral using a custom workflow (create an annotation with some brush strokes first and copy the annotation link):
```
        mask:
          url: <annotation_url>
          segmentation_name: Volume
          mag: [1, 1, 1]
```

------


MatthisCl (approved):
Nice, i did an inference run with a mask on my local webknossos and it worked, lgtm :+1:

Only thing i was a bit confused about was `dataset: # your dataset here` in the template, shouldn't that always be `dataset: predict:prediction` on default?

daniel-wer:
Thank you!

> Only thing i was a bit confused about was dataset: # your dataset here in the template, shouldn't that always be dataset: predict:prediction on default?

The idea is that most users would add their own tasks in between the predict and publish, for example to perform a segmentation with custom parameters or so. That's why this field is left to be filled out by the user :)

PR #8322, opened 2025-01-13
Use user token when requesting from HttpsDataVault on same DS

### LINKED ISSUE #8255: Pass user token when self-streaming from Webknossos
## Detailed Description
I have a wk dataset that has a layer with mags that have a path set, which is itself a zarr link to the same webknossos-datastore.
The layer is not public and wk fails to fetch the data. Could it analyze the zarr link and understand that it points to the data store itself (self-streaming) and in this case use the user token it received to also request the zarr data from itself? This way, self-streaming could succeed without explicit credentials in the json.

Here is a link to the problematic dataset: 
https://scm.slack.com/archives/C5AKLAV0B/p1733150028983289

frcroth:
- https://___.webknossos.xyz

### Steps to test:
- Explore a dataset with the remote URI being the zarr streaming URI of a layer on the same datastore

~~This branches of unified annotation versioning, so we wait until #7917 is merged and can cherrypick / rebase then~~

So basically any method now has tokenContext if it does something with requesting data in one of its called methods (which is a lot of methods)

### Issues:
- fixes #8255 

------


webknossos-datastore/app/com/scalableminds/webknossos/datastore/datareaders/DatasetArray.scala
```diff
@@ -185,19 +188,22 @@ class DatasetArray(vaultPath: VaultPath,
...
+      chunkIndex: Array[Int])(implicit ec: ExecutionContext, tc: TokenContext): Fox[(VaultPath, NumericRange[Long])] =
+    ??? // Defined in subclass
 
   private def chunkContentsCacheKey(chunkIndex: Array[Int]): String =
     s"${dataSourceId}__${layerName}__${vaultPath}__chunk_${chunkIndex.mkString(",")}"
 
   private def getSourceChunkDataWithCache(chunkIndex: Array[Int], useSkipTypingShortcut: Boolean = false)(
-      implicit ec: ExecutionContext): Fox[MultiArray] =
+      implicit ec: ExecutionContext,
+      tc: TokenContext): Fox[MultiArray] =
```
fm3, line 200:
```suggestion
      tc: TokenContext): Fox[MultiArray] =
      // Note: we omit the tokenContext from the cacheKey because (a) Failures aren’t cached anyway and (b) the dataset access is checked again in the controllers. Omitting it here prevents wasteful data duplicates in the cache.
```

webknossos-datastore/app/com/scalableminds/webknossos/datastore/controllers/BinaryDataController.scala
```diff
@@ -94,7 +95,7 @@ class BinaryDataController @Inject()(
       // If true, use lossy compression by sending only half-bytes of the data
       halfByte: Boolean,
       mappingName: Option[String]
-  ): Action[AnyContent] = Action.async { implicit request =>
+  ): Action[AnyContent] = Action.async { implicit r =>
```
fm3, line 98: Why the renamings here? I think we use `implicit request` everywhere else too.

test/backend/DataVaultTestSuite.scala
```diff
@@ -36,10 +38,11 @@ class DataVaultTestSuite extends PlaySpec {
         "return correct response" in {
           WsTestClient.withClient { ws =>
             val uri = new URI("http://storage.googleapis.com/")
-            val vaultPath = new VaultPath(uri, HttpsDataVault.create(RemoteSourceDescriptor(uri, None), ws))
+            val vaultPath =
+              new VaultPath(uri, HttpsDataVault.create(RemoteSourceDescriptor(uri, None), ws, "example.com"))
```
fm3, line 42:
```suggestion
              new VaultPath(uri, HttpsDataVault.create(RemoteSourceDescriptor(uri, None), ws, dataStoreHost="example.com"))
```
Let’s make this a named parameter here, because I think it is not obvious from "example.com" what this is for.

webknossos-datastore/app/com/scalableminds/webknossos/datastore/datavault/HttpsDataVault.scala
```diff
@@ -133,6 +147,6 @@ class HttpsDataVault(credential: Option[DataVaultCredential], ws: WSClient) exte
 }
 
 object HttpsDataVault {
-  def create(remoteSourceDescriptor: RemoteSourceDescriptor, ws: WSClient): HttpsDataVault =
-    new HttpsDataVault(remoteSourceDescriptor.credential, ws)
+  def create(remoteSourceDescriptor: RemoteSourceDescriptor, ws: WSClient, dataStoreHost: String): HttpsDataVault =
```
fm3, line 150: please add a comment here explaining what this is for. It’s a pretty strong deviation from before, where the DataVaults and VaultPaths didn’t have to know anything about what a datastore is. I understand why this is still the best place, but we should explain to future readers of this code.
Also maybe explain that dataStoreHost is supposed to be not a remote host, but the “self” host. Maybe you can find better words for that than me…

webknossos-tracingstore/app/com/scalableminds/webknossos/tracingstore/tracings/editablemapping/EditableMappingLayer.scala
```diff
@@ -28,7 +28,8 @@ import scala.concurrent.ExecutionContext
 
 class EditableMappingBucketProvider(layer: EditableMappingLayer) extends BucketProvider with ProtoGeometryImplicits {
 
-  override def load(readInstruction: DataReadInstruction)(implicit ec: ExecutionContext): Fox[Array[Byte]] = {
+  override def load(readInstruction: DataReadInstruction)(implicit ec: ExecutionContext,
+                                                          tc: TokenContext): Fox[Array[Byte]] = {
```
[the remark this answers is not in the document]
fm3: @frcroth The Rabbit is actually on to something here. If a tc is now passed to `load` implicitly, maybe the EditableMappingLayer doesn’t need its tokenContext field anymore?

fm3 (commented):
Works great! Thanks for biting your way through all those call stacks…
I added only small suggestions

webknossos-datastore/app/com/scalableminds/webknossos/datastore/controllers/BinaryDataController.scala
fm3: I see. yes let’s do that

test/backend/DataVaultTestSuite.scala
frcroth: I opted for naming the variable, which I also think fulfills the purpose.

fm3 (approved):
:tada:

PR #8324, opened 2025-01-14
When jobsEnabled=False, send empty jobsSupportedByAvailableWorkers list

### LINKED ISSUE #8306: dataStore.jobsSupportedByAvailableWorkers should be empty when jobsEnabled=false
The database content should in this case not be sent to the frontend, but an empty list sent instead.

fm3:
Making the situation for frontend job checks simpler.

### Steps to test:
- `yarn enable-jobs`
- check that dataset json includes jobsSupportedByAvailableWorkers list
- in application.conf, set features.jobsEnabled=false
- now the list sent to the frontend should be empty

### Issues:
- fixes #8306


MichaelBuessemeyer (approved):
Looks good and works even greater :construction_worker:

Issue #6581, opened 2022-10-24
Complete dtype feature support


fm3:
This table should track the current state of wk dtype support.

|   dtype    | color layers  |       segmentation layers   | histogram      |  thumbnail   |
|------------|------------|------------|------------|------------|
|   uint8    |       ✓     |   ✓                       |     ✓        |     ✓      |
|   uint16   |       ✓     |   ✓                       |     ✓        |     ✓      |
|   uint24  rgb  |       ✓   |   does not apply        |     ✓        |     ✓      |
|   uint32   |       ✗     |   ✓                       |     ✓        |     ✓     |
|   uint64   |       ✗     |   (✓)  [(til 2⁵³−1)](https://github.com/scalableminds/webknossos/issues/6921)          |     ✓        |     ✓ (segmentation only)      |
|   |  |  |  |
|    int8    |       ✗     |        ✗                   |       ✗        |      ✗      |
|    int16   |       ✗     |        ✗                   |       ✗        |      ✗      |
|    int32   |       ✗     |        ✗                   |       ✗        |      ✗      |
|    int64   |       ✗     |        ✗                   |       ✗        |      ✗      |
|   |  |  |  |
|    float   |       ✓     |        ✗                   |        ✓       |      ✗     |
|    double  |       ✗     |        ✗                   |        ✗       |      ✗      |

normanrz:
Can we support this?

|   dtype    | color layers  |       segmentation layers   | histogram      |  thumbnail   |
|------------|------------|------------|------------|------------|
|   uint8    |       ✓     |   ✓                       |     ✓        |     ✓      |
|   uint16   |       ✓     |   ✓                       |     ✓        |     ✓      |
|   uint24  rgb  |       ✓   |   does not apply        |     ✓        |     ✓      |
|   uint32   |       ☑️     |   ✓                       |     ✓        |     ✓     |
|   uint64   |       ✗     |   ✓  (til 2⁵³−1)          |     ✓        |     ✓ (segmentation only)      |
|   |  |  |  |
|    int8    |       ☑️     |       ☑️                   |     ☑️       | ☑️    |
|    int16   |    ☑️    |  ☑️          |    ☑️       |  ☑️      |
|    int32   |    ☑️    |      ☑️               |  ☑️       |    ☑️     |
|    int64   |       ✗     |      ☑️                   |       ✗        |      ✗      |
|   |  |  |  |
|    float   |       ✓     |        ✗                   |        ✓       |   ☑️    |
[...]

PR #8325, opened 2025-01-14
Support rendering more dtypes

philippotto:
This PR adds rendering support for uint32 color layers, int8, int16, int32 for color + segmentation, and int64 for segmentation layers.

Histogram and thumbnail support were only done partially. The rest can happen in another PR.

- https://___.webknossos.xyz

### Steps to test:
- see 15 new datasets (dtype_test_*) on dev instance
- monkey test them (also use histogram view, brushing and editable mappings)
- no need to test all combinations, because there are automated screenshot tests for that

### TODOs:
|   dtype    | color layers  |       segmentation layers   | histogram      |  thumbnail   |
|------------|---------------|------------|------------|------------|
|   uint8    |       unchanged     |   unchanged                       |     unchanged        |     unchanged      |
|   uint16   |       unchanged     |   unchanged                       |     unchanged        |     unchanged      |
|   uint24  rgb  |       unchanged   |   does not apply        |     unchanged        |     unchanged      |
|   uint32   |       ☑️     |   unchanged                       |     unchanged        |     unchanged     |
|   uint64   |            |   unchanged          |     unchanged        |     unchanged (segmentation only)      |
|   |  |  |  |
|    int8    |       ☑️    |       ☑️                   |     ☑️       |  ☑️    |
|    int16   |    ☑️    |  ☑️          |    ☑️       |   ☑️      |
|    int32   |    ☑️    |      ☑️               |  ☑️       |     ☑️     |
|    int64   |            |      ☑️                   |               |            |
|   |  |  |  |
|    float   |       unchanged     |                           |        unchanged       |    ☑️    |
|    double  |            |                           |               |            |

### Issues:
- contributes to #6581

------


frontend/javascripts/oxalis/model/bucket_data_handling/bucket.ts
```diff
@@ -593,12 +611,11 @@ export class DataBucket {
 
     switch (this.state) {
       case BucketStateEnum.REQUESTED: {
-        // Clone the data for the unmergedBucketDataLoaded event,
-        // as the following merge operation is done in-place.
-        const dataClone = new TypedArrayClass(data);
```
philippotto, line 598: this is unrelated to the actual PR. the cloning of data was always done even though it's only necessary before doing a merge.

tools/test.sh
```diff
@@ -109,7 +109,8 @@ then
 elif [ $cmd == "test-screenshot" ]
 then
   ensureUpToDateTests
-  export NODE_PATH="$testBundlePath" && ava $(find "$testBundlePath" -name "*.screenshot.js") "$@"
+  # todop: revert
```
philippotto, line 112: revert

frontend/javascripts/test/puppeteer/dataset_rendering_helpers.ts
```diff
@@ -12,9 +12,19 @@ import type { RequestOptions } from "libs/request";
...
 import { sleep } from "libs/utils";
+import type { APIAnnotation } from "types/api_flow_types";
+import type Semaphore from "semaphore-promise";
 
 export const { WK_AUTH_TOKEN } = process.env;
 
+const PAGE_WIDTH = 1920;
+const PAGE_HEIGHT = 1080;
+
+// todop: change to false before merging
```
philippotto, line 23: todo

frontend/javascripts/test/puppeteer/dtype_datasets_rendering.screenshot.ts
```diff
@@ -0,0 +1,322 @@
...
+        const url = urljoin(URL, path);
+        const response = await fetch(url, options);
+        const { id } = await response.json();
+        datasetNameToId[datasetName] = id;
+        return true;
+      },
+      () => {},
+    );
+  }
+});
```
MichaelBuessemeyer, line 206: This is a complete duplication from the other `screenshot.ts` file. Could you please deduplicate this?

frontend/javascripts/test/puppeteer/dtype_datasets_rendering.screenshot.ts
```diff
@@ -0,0 +1,322 @@
...
+test.serial("Dataset IDs were retrieved successfully", (t) => {
+  for (const datasetName of datasetNames) {
+    t.truthy(datasetNameToId[datasetName], `Dataset ID not found for "${datasetName}"`);
+  }
+});
```
MichaelBuessemeyer, line 211: This included

frontend/javascripts/test/shaders/shader_syntax.spec.ts
```diff
@@ -22,8 +22,24 @@ test("Shader syntax: Ortho Mode", (t: ExecutionContext<any>) => {
     globalLayerCount: 2,
```
MichaelBuessemeyer, line 22: All these tests have the same / similar `textureLayerInfos` maybe having different settings in some tests might be nice. 

Although this is not a must, as this should also be covered by the e2e tests

webknossos-datastore/app/com/scalableminds/webknossos/datastore/services/DataConverter.scala
```diff
@@ -47,14 +47,16 @@ trait DataConverter extends FoxImplicits {
     dstArray
   }
 
-  def toUnsigned(data: Array[_ >: Byte with Short with Int with Long with Float])
-    : Array[_ >: UByte with UShort with UInt with ULong with Float] =
+  def toMaybeUnsigned(
```
MichaelBuessemeyer, line 50: @fm3 regarding naming here: In the backend the word `opt` is usually used, correct? But in this case, no real Option is involved. So is it ok to keep this name? Do you have better suggestions?

MichaelBuessemeyer (commented):
Thanks for this PR :100: :sunrise: :rainbow: 

I left some comments and questions. Next up is testing :D :test_tube:

MichaelBuessemeyer:
Findings during testing:

- This dataset has some white dots in it. Is this intentional?
https://moredtypes.webknossos.xyz/annotations/67b7004d0100002402a8a460#512,256,16,0,0.36
image

- Same for int 8 dataset: https://moredtypes.webknossos.xyz/datasets/dtype_test_int8_color-67b5a630010000c500ce1e79/view#512,256,16,0,1.182

- Cannot create annotation for DS: https://moredtypes.webknossos.xyz/datasets/dtype_test_int16_segmentation-67b5a630010000c500ce1e83/view#512,256,16,0,0.283 -> backend error

- Same for https://moredtypes.webknossos.xyz/datasets/dtype_test_int32_segmentation-67b5a630010000c500ce1eb0/view#512,256,16,0,1.3

- And https://moredtypes.webknossos.xyz/datasets/dtype_test_int64_segmentation-67b5a630010000c500ce1eb5/view#512,256,16,0,1.43

- And https://moredtypes.webknossos.xyz/datasets/dtype_test_int8_segmentation-67b5a630010000c500ce1eb6/view#512,256,16,0,1.3

=> All intX datasets are affected. I'll take care of that.


The rest seemed to work :raccoon: :heavy_check_mark:

webknossos-datastore/app/com/scalableminds/webknossos/datastore/image/ImageCreator.scala
```diff
@@ -121,28 +123,67 @@ object ImageCreator extends LazyLogging {
...
           }
+          grayValues(idx / bytesPerElement) = grayNormalized
+          elementClass match {
+            case ElementClass.uint24 => // assume uint24 rgb color data
+              (0xFF << 24) | ((b(idx) & 0xFF) << 16) | ((b(idx + 1) & 0xFF) << 8) | ((b(idx + 2) & 0xFF) << 0)
+            case _ =>
+              (0xFF << 24) | (colorRed(grayNormalized) << 16) | (colorGreen(grayNormalized) << 8) | (colorBlue(
+                grayNormalized) << 0)
+          }
         }
```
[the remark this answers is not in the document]
MichaelBuessemeyer: > Confirm that either the potential loss is acceptable for your use case or add documentation/checks to restrict/handle such extreme values

The potentail loss is ok as to should be represented as a 4byte argb value anyway and as it is gray scale (except for uint24) one color only has a byte. Thus, precision loss is ensured but doesn't matter as this is anyway only used to generate an image

webknossos-datastore/app/com/scalableminds/webknossos/datastore/models/datasource/DataLayer.scala
```diff
@@ -62,6 +62,36 @@ object ElementClass extends ExtendedEnumeration {
...
+    case ElementClass.uint32 => (0.0, math.pow(2, 32) - 1)
+    case ElementClass.uint64 => (0.0, math.pow(2, 64) - 1)
+    case ElementClass.float  => (-3.40282347e+38, 3.40282347e+38)
+    case ElementClass.double => (-1.7976931348623157e+308, 1.7976931348623157e+308)
+    case ElementClass.int8   => (-math.pow(2, 7), math.pow(2, 7) - 1)
+    case ElementClass.int16  => (-math.pow(2, 15), math.pow(2, 15) - 1)
+    case ElementClass.int32  => (-math.pow(2, 31), math.pow(2, 31) - 1)
+    case ElementClass.int64  => (-math.pow(2, 63), math.pow(2, 63) - 1)
+    case _                   => (0.0, 255.0)
+  }
```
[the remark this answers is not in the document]
MichaelBuessemeyer: the frontend also uses this range: `(-math.pow(2, 63), math.pow(2, 63) - 1)`

The potential precision loss in the frontend is currently accepted as unsupported value range for such layers

MichaelBuessemeyer:
@fm3 Could you please check my frontend changes? I adjusted the backend to render thumbnails for all data types

MichaelBuessemeyer:
> Findings during testing:
> 
> * This dataset has some white dots in it. Is this intentional?
>   https://moredtypes.webknossos.xyz/annotations/67b7004d0100002402a8a460#512,256,16,0,0.36
>   image
> * Same for int 8 dataset: https://moredtypes.webknossos.xyz/datasets/dtype_test_int8_color-67b5a630010000c500ce1e79/view#512,256,16,0,1.182
> * Cannot create annotation for DS: https://moredtypes.webknossos.xyz/datasets/dtype_test_int16_segmentation-67b5a630010000c500ce1e83/view#512,256,16,0,0.283 -> backend error
> * Same for https://moredtypes.webknossos.xyz/datasets/dtype_test_int32_segmentation-67b5a630010000c500ce1eb0/view#512,256,16,0,1.3
> * And https://moredtypes.webknossos.xyz/datasets/dtype_test_int64_segmentation-67b5a630010000c500ce1eb5/view#512,256,16,0,1.43
> * And https://moredtypes.webknossos.xyz/datasets/dtype_test_int8_segmentation-67b5a630010000c500ce1eb6/view#512,256,16,0,1.3
> 
> => All intX datasets are affected. I'll take care of that.
> 
> The rest seemed to work 🦝 ✔️


@philippotto Regarding the error when creating an annotation for all signed data types: I fixed this as well as added support for the thumnail rendering

webknossos-datastore/app/com/scalableminds/webknossos/datastore/services/DataConverter.scala
fm3: how about `toUnsignedIfNeeded`? I think we have “ifNeeded” already in a couple of places.

fm3 (commented):
Backend LGTM

philippotto:
@MichaelBuessemeyer thanks for your review! please have another look. the screenshot tests also succeeded now. also, could you update the checkmarks in the pr description for the thumbnails column. I assume the remaining boxes can be checked off now?

philippotto:
Sorry, I forgot to react to your testing results. Thanks for the thorough testing and also fixing of the "create annotation" bugs :pray: 


> * This dataset has some white dots in it. Is this intentional?

Yes, this is intentional. The rectangles that contain the min or max value of the current dtype range are marked with a white (or black) dot in the middle. I briefly mentioned in this at my offsite talk and it's also stated here in the [description in notion](https://www.notion.so/scalableminds/Test-Datasets-c0563be9c4a4499dae4e16d9b2497cfb?pvs=4#190b51644c638012be49d4e5f57c55e6). I probably could have mentioned this in the pr description, too, sorry about that!

frontend/javascripts/test/shaders/shader_syntax.spec.ts
```diff
@@ -48,9 +64,33 @@ test("Shader syntax: Ortho Mode + Segmentation - Mapping", (t: ExecutionContext<
...
-      ["color_layer_1"]: { packingDegree: 4.0, dataTextureCount: 1 },
-      ["color_layer_2"]: { packingDegree: 4.0, dataTextureCount: 1 },
-      ["segmentationLayer"]: { packingDegree: 1.0, dataTextureCount: 4 },
+      ["color_layer_1"]: {
+        isColor: true,
+        packingDegree: 4.0,
+        dataTextureCount: 1,
+        isSigned: false,
+        glslPrefix: "",
+        unsanitizedName: "color_layer_1",
```
MichaelBuessemeyer, line 73: `unsanitizedName` name is everywhere the same in this file and does not match the key of this dict / the actual layer name. Although as it currently does not yield any error I might in the future and therefore I would prefer to have it correct here :pray:

MichaelBuessemeyer (approved):
Thanks for incorporating all the feedback as well as taking care of the CI @philippotto :pray: 

Some more minor feedback:
- You added the screenshot test file `annotation_ROI2017_wkw_null.png` where I think the `_null` at the end of the name is unnecessary imo / not needed / may hint at a bug in the code path where a variable is not properly defined



This is all I found :D. I did not retest as I could not find any significant shader changes. The small once (including the fix you made) should be covered by the screenshot tests,  correct? Or do you think I should do another testing round (safety first)?


Not leaving a green checkmark as you seem to be currently still working on this pr

MichaelBuessemeyer (changes requested):
Well, I msisclicked, sorry :see_no_evil:

philippotto:
@MichaelBuessemeyer I just noticed the following: when creating an annotation for a int32 segmentation, the resulting volume layer has uint32 as the dtype. this probably is the same for all signed cases. could you have a look at fixing this?

frontend/javascripts/oxalis/model/helpers/shader_editor.ts
```diff
@@ -14,51 +14,193 @@ export default {
...
+  firstLineNumber: 1,
+  overwrite: false,
+  newLineMode: "auto",
+  useWorker: true,
+  useSoftTabs: true,
+  navigateWithinSoftTabs: false,
+  tabSize: 4,
+  wrap: "off",
+  indentedSoftWrap: true,
+  mode: "c_cpp",
```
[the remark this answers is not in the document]
philippotto: sorry, I replied here but the comments were pending and not submitted (same for a few other threads).

MichaelBuessemeyer:
> @MichaelBuessemeyer I just noticed the following: when creating an annotation for a int32 segmentation, the resulting volume layer has uint32 as the dtype. this probably is the same for all signed cases. could you have a look at fixing this?

Oh good catch! It should be fixed now. 
Reason for the Bug: The proto type of the type / element class did not yet support the new dtypes.

MichaelBuessemeyer:
@fm3 A few backend changes were necessary. Could you please review [c794c1e](https://github.com/scalableminds/webknossos/pull/8325/commits/c794c1ee810535a51f52a6c0652bfbdc1bc9c27a) and [9986d92](https://github.com/scalableminds/webknossos/pull/8325/commits/9986d92b6b9a04a8feccd5cb673aa31f946c3306)?

frontend/javascripts/oxalis/shaders/segmentation.glsl.ts
```diff
@@ -38,6 +42,44 @@ export const convertCellIdToRGB: ShaderModule = {
...
+        highp uint combinedHigh = ~high_uint + uint(combinedLow == 0u ? 1u : 0u); // Add carry if low overflows
+
+        // Output absolute value as two uint32
+        absLow = combinedLow;
+        absHigh = combinedHigh;
+      } else {
+        // The number is already positive
+        absLow = uint(low);
+        absHigh = uint(high);
+      }
```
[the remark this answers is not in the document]
MichaelBuessemeyer: Awesome :tada:

webknossos-datastore/proto/VolumeTracing.proto
```diff
@@ -23,6 +23,10 @@ message VolumeTracing {
     uint24 = 3;
     uint32 = 4;
     uint64 = 8;
+    int8 = 11;
+    int16 = 12;
+    int32 = 14;
+    int64 = 18;
```
philippotto, line 29: It took me a bit to see the pattern. Maybe add a comment that the id is `byteCount + (unsigned ? 0 : 10)`.

webknossos-datastore/app/com/scalableminds/webknossos/datastore/models/datasource/DataLayer.scala
```diff
@@ -110,14 +110,30 @@ object ElementClass extends ExtendedEnumeration {
...
+    case 2  => Some(ElementClass.uint16)
+    case 3  => Some(ElementClass.uint24)
+    case 4  => Some(ElementClass.uint32)
+    case 8  => Some(ElementClass.uint64)
+    case 11 => Some(ElementClass.int8)
+    case 12 => Some(ElementClass.int16)
+    case 14 => Some(ElementClass.int32)
+    case 18 => Some(ElementClass.int64)
+    case _  => None
   }
```
philippotto, line 137: now these ids are defined three times in total (twice here and once in the proto object). since you chose `byteCount + (unsigned ? 0 : 10)` as the id pattern, that could be used here?
alternatively, maybe one can define a bijective map here once which allows the look ups in both directions?

MichaelBuessemeyer: > since you chose byteCount + (unsigned ? 0 : 10) as the id pattern, that could be used here?

Not sure how this would be implemented. Seems to me like too much of an overhead and would complicate things.

I now made it 
```scala
private val protoFieldNumberToElementClass: Map[Int, ElementClass.Value] = Map(
    1 -> ElementClass.uint8,
    2 -> ElementClass.uint16,
    3 -> ElementClass.uint24,
    4 -> ElementClass.uint32,
    8 -> ElementClass.uint64,
    11 -> ElementClass.int8,
    12 -> ElementClass.int16,
    14 -> ElementClass.int32,
    18 -> ElementClass.int64
  ) 


  def volumeTracingElementClassProtoFieldNumber(elementClass: ElementClass.Value): Int ={
    protoFieldNumberToElementClass.find(_._2 == elementClass).get._1
  }

  def fromVolumeTracingElementClassProtoFieldNumber(protoEnumFieldNumber: Int): Option[ElementClass.Value] = {
    protoFieldNumberToElementClass.get(protoEnumFieldNumber)
  }
```

What do you think @fm3?
The lookup now likely has a small runtime overhead (but very likely negligible)

fm3: I think if we’re no longer using bytesPerElement for the proto ElementClass conversion, we should just disregard the enum’s integer values completely, and just write a translation table using the named Enum values. `object ElementClass` should have fromProto and toProto functions.

See for example the TreeType enum.

fm3:
I checked the latest backend commits, and they LGTM apart from [my suggestion to refactor proto enum conversion](https://github.com/scalableminds/webknossos/pull/8325#discussion_r1987042601)

webknossos-datastore/app/com/scalableminds/webknossos/datastore/models/datasource/DataLayer.scala
MichaelBuessemeyer: thanks for the suggestion I pushed the required changes, but just noticed the following:
I am unsure about the `float` and `double` support. The `ElementClass` supports `float` and `double` but `ElementClassProto` does not.

But I gave this a double check: Currently the only places where the these serialization & deserialization of `ElementClassProto` are used, is in context of volume layer / fallback layers. As these do not support floating values anyway, it should be fine right now. But in the future one might use the new `fromProto` and `toProto` for color layer :thinking: This would then turn into trouble.

MichaelBuessemeyer:
Okidoky: Here are my notes about the changes I did to support merging signed volume annotations:
In the backend we have an attraction around different data types that can be used for segmentation data. On master this is called `UnsignedInteger` but now as we want to support also signed integers, this needs to be changed. So here is what I did:
- Changed `UnsignedInteger` to `SegmentInteger`-> name is open to debate. I also had `VolumeInteger` in mind but I am not so happy with both.
- `toPositiveLong` was used to get the actual unsigned data of a segment voxel. This is now called `toLong` and in case of unsigned type it is auto converted to be positive.
- I added Int8, Int16, Int32 and Int64 as SegmentIntegers
- Adjusted `SegmentInteger` to support the new signed SegmentIntegers
- renamed `UnsignedIntegerArray` to `SegmentIntegerArray`
- Adjusted `SegmentIntegerArray` to support the new signed SegmentIntegers
- renamed variables from `unsignedInt(eger(Array))` to ``segmentInt(eger(Array))`

=> Merging annotations with signed volume layer should work again now (at least according to my testing). Although, merging int32 seems to take forever :/. Don't know why that's the case.

@fm3 could you please have another look at my changes? Especially at [aec67d6](https://github.com/scalableminds/webknossos/pull/8325/commits/aec67d68269cb44c063c8a07f53616f0b9f21f52)?

MichaelBuessemeyer:
Moreover: I noticed that the segment picker tool is not working for negative segment ids. Could you please take another look when you have time @philippotto? Hopefully a very small change :) :small_airplane:

MichaelBuessemeyer:
Backend Bug:
- While trying to merge two int32 annotations the backend printed
  ```
  backend: 2025-03-11 17:05:04,028 [warn] com.scalableminds.webknossos.tracingstore.controllers.TSAnnotationController - 
  Answering 400 at /tracings/annotation/mergedFromIds?token=WIMD1FcRFc_5h27SVCCfVg&toTemporaryStore=false&newAnnotationId=67d05bf9720100186358672e: {"messages":[{"error":"key not found: UInt32(-102261123)"}]}
```

=> Needs investigation about what actually happening here and why there is such an error.

**Comment by MichaelBuessemeyer** [comment]
I just encoutered another frontend bug :see_no_evil: :crying_cat_face: 
image

Here is what I did:
- Open a hybrid annotation on the int64 dataset
- Pick some cell
- try to brush -> crashes.

I am a little confused by the error as the same error occurred for segment id `142971416741890` but in the browser console `BigInt(142971416741890)` works completely fine :thinking:

**Comment by philippotto** [comment]
> I just encoutered another frontend bug 🙈 😿 image
> 
> Here is what I did:
> 
> * Open a hybrid annotation on the int64 dataset
> * Pick some cell
> * try to brush -> crashes.
> 
> I am a little confused by the error as the same error occurred for segment id `142971416741890` but in the browser console `BigInt(142971416741890)` works completely fine 🤔

thanks for the report! luckily, it was easy to fix :)

**Comment by MichaelBuessemeyer** [comment]
I just noticed that `encodeAsByte` has been adjusted in this PR. => Our SAM server needs to be adjusted to the new types.
- [ ] Add SAM support for new data types

**Comment by philippotto** [comment]
> Our SAM server needs to be adjusted to the new types.

will you do that? :see_no_evil: or who should we delegate this to?

**Comment by MichaelBuessemeyer** [comment]
Nvm: Our SAM server already has all dtypes implemented. See [here](https://github.com/scalableminds/torch-serve-sam/blob/717d7174daa37f03684bd89c96f51c53bb6c4579/handlers/sam2_handler.py#L9). The encoding is actually already just like in this PR. => So nothing TODO :dancers:

**Comment by MichaelBuessemeyer** [comment]
@fm3 If you have time, could you please give [aa22c3b](https://github.com/scalableminds/webknossos/pull/8325/commits/aa22c3b60b55d6fe28b384639fa5d2f5005b2c80) a quick look? I fixed the CI simply by adding the missing required cases and throwing an error in case of the unrecognized case.

File: webknossos-datastore/app/com/scalableminds/webknossos/datastore/models/datasource/DataLayer.scala [hunk 15]
```diff
@@ -115,12 +115,14 @@ object ElementClass extends ExtendedEnumeration {
...
       case ElementClassProto.uint8  => uint8
       case ElementClassProto.uint16 => uint16
+      case ElementClassProto.uint24 => uint24
       case ElementClassProto.uint32 => uint32
       case ElementClassProto.uint64 => uint64
       case ElementClassProto.int8   => int8
       case ElementClassProto.int16  => int16
       case ElementClassProto.int32  => int32
       case ElementClassProto.int64  => int64
+      case _                        => throw new RuntimeException(s"Cannot convert ElementClassProto $elementClassProto to ElementClass")
```
fm3, line 125: please state the protobuf unrecognized field here directly, instead of the _.

Also why is there no problem for the other direction? Shouldn’t we catch float and double?

MichaelBuessemeyer: > Also why is there no problem for the other direction? Shouldn’t we catch float and double?

I honestly don't know. Maybe Scala is smart enough to infer that this method is never called with a type that is not in the list :thinking:?!

MichaelBuessemeyer: I'll give this a quick check :D

MichaelBuessemeyer: hmm ... I added
```scala
val abc = ElementClass.toProto(ElementClass.float)
println(abc)
```
at the start of a route and noticed that no compilation error occurred. => So I'd say it is saver to make `toProto` return a Box and adjust the code accordingly.

MichaelBuessemeyer: Should be better now. `toProto` now returns a box and I adjusted the cases where it was used. Moreover I removed the `elementClassToProto` implicit.

It should now be saver to use IMO

fm3 (approved):
Backend LGTM, retest looked good. 

Could not test segment stats since there is no signed int segmentation dataset with a segment index file. Do you know if the segment index file format would support this, especially with negative IDs? Otherwise the job to compute one might fail for these segmentation layers. Same question for the compute meshes job.

I guess from now on we need to test all new segmentation/volume/mapping/proofreading/mesh features also with segmentations with negative ids… Admittedly, such datasets are rarely used by users and maybe we can fix some issues later as they occur.

philippotto:
> Could not test segment stats since there is no signed int segmentation dataset with a segment index file.

How can I trigger the index computation? Is it done on DS upload (if the worker is set up)? my suggestion would be to test this once it's deployed and then schedule a follow up if necessary.

> I guess from now on we need to test all new segmentation/volume/mapping/proofreading/mesh features also with segmentations with negative ids… Admittedly, such datasets are rarely used by users and maybe we can fix some issues later as they occur.

I think, we don't necessarily need to add this to our testing agenda. as you say, such DS should be very rare and the current PR ensures that at least basic functionality (like rendering, annotating and ad hoc meshing etc) works. if users run into problems, we can fix that ad hoc. prior to this PR, DS with negative ids wouldn't work either, so it's still an improvement.

fm3:
> How can I trigger the index computation?

I think it’s an item in the layer sidebar for segmentation layers. There is a function `getComputeSegmentIndexFileButton` in layer_settings_tab.tsx

> schedule a follow up if necessary

Ok, agreed!

MichaelBuessemeyer:
Wasn't the `ComputeSegmentIndexFile` job removed :thinking:?

If I recall correctly some months ago you deprecated one of the jobs @fm3

fm3:
> If I recall correctly some months ago you deprecated one of the jobs

That would probably be `globalize_floodfills` in #7173 ? I don’t know about any other removed job :thinking:

MichaelBuessemeyer (approved):
Latest frontend changes lgtm. Is retesting the fixes needed before merging or did you do that already @philippotto?

philippotto:
> Latest frontend changes lgtm. Is retesting the fixes needed before merging or did you do that already @philippotto?

I'll do that before merging. thanks for the approval!

frontend/javascripts/oxalis/controller/combinations/volume_handlers.ts
```diff
@@ -115,7 +115,7 @@ export function handlePickCellFromGlobalPosition(
 ) {
   const segmentId = getSegmentIdForPosition(globalPos);
 
-  if (segmentId > 0) {
+  if (segmentId !== 0) {
```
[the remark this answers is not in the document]
philippotto: not sure what the "potentiel issue" is here now :shrug:

PR #8341, opened 2025-01-20
Fix edge cases in NGFF dataset upload

frcroth:
- https://___.webknossos.xyz

### Steps to test:
- Upload the file here: https://scm.slack.com/archives/C02H5T8Q08P/p1737359864538299?thread_ts=1737117750.874179&cid=C02H5T8Q08P

### What this does
- If there is only one directory, search in that directory. This directly reverts the logic in https://github.com/scalableminds/webknossos/blob/d71e780a94d68ebdb87a44fab762da79a9a704f8/webknossos-datastore/app/com/scalableminds/webknossos/datastore/services/uploading/UploadService.scala#L587 . If we do not know why color and segmentation are excluded here, we can also remove this exclusion at that spot (would be cleaner)
- The test dataset has the multiscale name set as "/", which breaks routes. So the layer names are now also normalized. 

------


fm3 (approved):
LGTM

PR #8342, opened 2025-01-21
WIP: allow proofreading without agglomerate file (identity mapping)

### LINKED ISSUE #8327: Allow Proofreading without agglomerate file (start with oversegmentation)
When there is no agglomerate file, proofreading can currently not be used at all.

It could be supported such that every “agglomerate” starts with a single segment/supervoxel. No edges eixst. Proofreading tools would create new edges/graphs.

fm3:
- https://___.webknossos.xyz

### Steps to test:
- abc

### TODOs:
- [ ] Use None rather than emptystring for editable mappings without base mapping
  - [x] Double check this doesn’t break protobuf compatibility
  - [x] Scala code
  - [ ] Frontend
- [ ] Some Segments with Position 0,0,0 are written into the graph (visible when importing agglomerate skeleton). Why? Should have used updateAction positions instead.
  - [ ] Because the frontend does not send positions in updateActions anymore. Should be re-added.
- [ ] Support loading Agglomerate Skeleton for single segment (use click position?)
- [x] How to show this in UI? Always enable proofreading tools? Or Include explicit switching?
  - [ ] We decided to always show the proofreading tool, but add confirm modal if it is started on the base mapping even though hdf5 mappings exist.
- [x] Adapt makeMappingEditable
- [ ] Some requests can probably be skipped, as they now return identity in this case
- [x] Solve merge conflicts after UAV #7917 is merged (makeMappingEditable route is replaced by frontend-issued update action)

### Issues:
- fixes #8327

------


fm3:
The fact that we don’t have reliable positions for the (over)segments when proofreading without an agglomerate file makes this whole feature unreliable.

We decided to take a new approach of having a worker job that creates such an agglomerate file, compare https://github.com/scalableminds/webknossos/issues/8637

PR #8344, opened 2025-01-21
Redirect rather than proxy to about page

fm3:
We decided to no longer the about page to the same domain as wk, but use a redirect to a subdomain. This PR implements the wk side of this change.

### Steps to test:
- In application.conf, set `isWkOrgInstance.true`, set `aboutPageRedirect.prefix` to the target root address, and `aboutPagePredirect.routes` to a list of test url paths (e.g. `["/test.html"]`
- visiting `localhost:9000/test.html` should redirect to `targetHost/test.html`

------

### DIFF
--- a/app/controllers/AboutPageRedirectController.scala
+++ b/app/controllers/AboutPageRedirectController.scala
@@ -8,7 +8,6 @@ import com.scalableminds.util.mvc.CspHeaders
 import com.scalableminds.util.tools.Fox
 import models.user.{MultiUserDAO, Theme}
 import opengraph.OpenGraphService
-import play.api.libs.ws.WSClient
 import play.api.mvc.{Action, AnyContent}
 import play.filters.csp.CSPConfig
 import security.WkEnv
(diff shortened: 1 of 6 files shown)


app/controllers/AboutPageRedirectController.scala
```diff
@@ -17,18 +16,17 @@ import utils.WkConf
...
+                                            openGraphService: OpenGraphService)(implicit ec: ExecutionContext)
     extends Controller
     with CspHeaders {
 
-  def proxyPageOrMainView: Action[AnyContent] = sil.UserAwareAction.async { implicit request =>
-    if (matchesProxyPage(request)) {
-      ws.url(conf.Proxy.prefix + request.uri).get().map(resp => Ok(resp.bodyAsBytes.utf8String).as(resp.contentType))
+  def redirectToAboutPageOrSendMainView: Action[AnyContent] = sil.UserAwareAction.async { implicit request =>
+    if (matchesRedirectRoute(request)) {
+      Fox.successful(Redirect(conf.AboutPageRedirect.prefix + request.uri))
```
normanrz, line 29: Is this permanent redirect (301)?

normanrz (approved):

PR #8347, opened 2025-01-22
Assert that all migrations are mentioned in migration guide and that they have reversion counterparts

philippotto:
- https://___.webknossos.xyz

### Steps to test:
- see CI. without the adaptions to the migration guide and reversions folder, the script exited with 1.

### Issues:
- I noticed that the newest migration was not mentioned in the migration guide when I created the newest release.

### DIFF
--- a/conf/evolutions/reversions/080-job-add-cancelled.sql
+++ b/conf/evolutions/reversions/080-job-add-cancelled.sql
@@ -1,0 +1,1 @@
+-- unfortunately, ALTER TYPE ... ADD cannot not be reverted
--- a/tools/assert-complete-migrations.sh
+++ b/tools/assert-complete-migrations.sh
@@ -1,0 +1,46 @@
+#!/bin/bash
+set -euo pipefail
+
+# This script asserts that each migration file in conf/evolutions
+# - is mentioned in either MIGRATIONS.released.md or MIGRATIONS.unreleased.md (but not both)
+# - has a reversion sibling in conf/evolutions/reversions
+
+EVOLUTIONS_FOLDER="conf/evolutions"
+RELEASED_FILE="MIGRATIONS.released.md"
+UNRELEASED_FILE="MIGRATIONS.unreleased.md"
+
+CONTENT_RELEASED=$(cat "$RELEASED_FILE")
+CONTENT_UNRELEASED=$(cat "$UNRELEASED_FILE")
+
+exit_code=0
+
+for file in "$EVOLUTIONS_FOLDER"/*; do
+    filename=$(basename "$file")
+
+    # Skip the "reversions" directory and initial-schema
+    if [[ "$filename" == "reversions" || "$filename" == "000-initial-schema.sql" ]]; then
+        continue
+    fi
+
+    mentioned_in_x=$([[ "$CONTENT_RELEASED" == *"$filename"* ]] && echo 1 || echo 0)
+    mentioned_in_y=$([[ "$CONTENT_UNRELEASED" == *"$filename"* ]] && echo 1 || echo 0)
+
+    if (( mentioned_in_x == 1 && mentioned_in_y == 1 )); then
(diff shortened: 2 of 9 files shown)


conf/evolutions/reversions/080-job-add-cancelled.sql
```diff
@@ -0,0 +1 @@
+-- unfortunately, ALTER TYPE ... ADD cannot not be reverted
```
philippotto, line 1: not sure what the original idea was? there is simply no point of return after that migration? only from backups?

fm3: Well, the old scala code doesn’t care if the SQL enum has another possible value, as long as it doesn’t actually have to parse any rows with it, so we could have rolled back that commit in case of a bug, without changing the sql schema back.
The reversions are mostly there as a safeguard in case of bugs with the new code that show up immediately after deploying. So it’s not always worth going to the trouble of building a 100% accurate reversion, rather just enough to let the old code run again until we can repair the new.
That is also why the correctness of the reversions themselves is not asserted in the CI.

tools/assert-complete-migrations.sh
```diff
@@ -0,0 +1,46 @@
...
+    # Skip the "reversions" directory and initial-schema
+    if [[ "$filename" == "reversions" || "$filename" == "000-initial-schema.sql" ]]; then
+        continue
+    fi
+
+    mentioned_in_x=$([[ "$CONTENT_RELEASED" == *"$filename"* ]] && echo 1 || echo 0)
+    mentioned_in_y=$([[ "$CONTENT_UNRELEASED" == *"$filename"* ]] && echo 1 || echo 0)
+
+    if (( mentioned_in_x == 1 && mentioned_in_y == 1 )); then
+        echo "Conflict: '$filename' is listed in both MIGRATIONS.released.md and MIGRATIONS.unreleased.md."
```
fm3, line 29:
```suggestion
        echo "Conflict: '$filename' is listed in both $RELEASED_FILE and $UNRELEASED_FILE."
```
(or something like that? same below)

tools/assert-complete-migrations.sh
```diff
@@ -0,0 +1,46 @@
...
+for file in "$EVOLUTIONS_FOLDER"/*; do
+    filename=$(basename "$file")
+
+    # Skip the "reversions" directory and initial-schema
+    if [[ "$filename" == "reversions" || "$filename" == "000-initial-schema.sql" ]]; then
+        continue
+    fi
+
+    mentioned_in_x=$([[ "$CONTENT_RELEASED" == *"$filename"* ]] && echo 1 || echo 0)
+    mentioned_in_y=$([[ "$CONTENT_UNRELEASED" == *"$filename"* ]] && echo 1 || echo 0)
```
fm3, line 26:
```suggestion
    mentioned_in_released=$([[ "$CONTENT_RELEASED" == *"$filename"* ]] && echo 1 || echo 0)
    mentioned_in_unreleased=$([[ "$CONTENT_UNRELEASED" == *"$filename"* ]] && echo 1 || echo 0)
```
no strong opinion, but x and y looks somewhat arbitrary.

conf/evolutions/reversions/080-job-add-cancelled.sql
philippotto: Ok :+1:

> That is also why the correctness of the reversions themselves is not asserted in the CI.

does this mean you are against the CI check I added? it doesn't assert correctness but rather "completeness". like in the above case, one can always create an empty file with a comment for such special cases. in my opinion this would be a good practice, as it forces us to at least think about the reversion (and one cannot forget about it). but i have no strong opinion here. I can remove it again.

fm3 (approved):
Good idea!

LGTM but I’m not very familiar with bash and might not be the best reviewer for this.

conf/evolutions/reversions/080-job-add-cancelled.sql
fm3: Thanks for the comment! Fixed in baa129c

PR #8349, opened 2025-01-22
Log slow RPC requests

fm3:
- Log all slow (>2min) RPC requests (that don’t have `.silent`).
 - Clean up RPCRequest logging code
 - Unify duration formatting. I may have gone a little overboard there…
 - Use our Instant class in more places

- https://logslowrpcs.webknossos.xyz

### Steps to test:
- Insert Thread.sleep(2 * 60 * 1000) in the implementation of an RPC (e.g. UserTokenController.validateUserAccess)
- Call that RPC (e.g. by opening an annotation)
- Slow request should be logged in readable format

------


webknossos-datastore/app/com/scalableminds/webknossos/datastore/rpc/RPCRequest.scala
```diff
@@ -215,79 +232,53 @@ class RPCRequest(val id: Int, val url: String, wsClient: WSClient)(implicit ec:
...
-            s"Successful request (ID: $id). " +
-              s"Body: '<${responseBytes.length} raw bytes>'")
-        }
-        Fox.successful(responseBytes.toArray)
-      } else {
-        logger.error(
-          s"Failed to send WS request to $url (ID: $id). " +
-            s"RequestBody: '$requestBodyPreview'. Status ${response.status}. " +
-            s"ResponseBody: '${response.body.take(100)}'")
-        Fox.failure("Unsuccessful WS request to $url (ID: $id)")
```
MichaelBuessemeyer, line 231: Is there a reason why you removed this error handling here and below? It is because this case cannot happen anyway, because only successful requests fill end up in the flatMap method body? If that's true, why not keep the error message by some kind of  `flatMap(...).orElse(//handle error here)` manor. Don't know how well this can be applied here :thinking:

MichaelBuessemeyer (commented):
Nice :+1: Thanks for this PR

The code looks good. I only found 2 mini suggestions and left a question. Will test now :)

webknossos-datastore/app/com/scalableminds/webknossos/datastore/rpc/RPCRequest.scala
MichaelBuessemeyer: Oh nvm. Perform request does this. :+1:

MichaelBuessemeyer (approved):
Testing went well. Feel free to consider / disregard my suggestions :)

PR #8350, opened 2025-01-22
Resolve symbolic links to mags and store them in db

### LINKED ISSUE #6726: Find all datasets that share a layer
Many datasets share a layer (e.g. one color layer, many segmentation layers)
wk should detect this (by resolving symlinks or the paths in datasource-properties.json) and store this information in postgres so that queries can be possible against this info (e.g. list all datasets with this layer)

frcroth:
- https://___.webknossos.xyz

### Steps to test:
- Create ds with symlinks in mag, ds with symlinks in layer, remote datasets, ...
- Check DB or GET /datasets/:datasetId/layers/:layer/linkedMags

### Overview

In the datastore we now scan for mags and symlinks. Symlinks are resolved for layer directories and mag directories. The result is send to WK via a new route (subject to discussion).
In WK the database stores per mag:
- path: The path (stored as URI) that wk uses to read the data
- realPath: The path (stored as URI) where the data actually resides
- hasLocalData: Is the data in the realPath local to the dataset.

So that
- For normal datasets, realPath = path and hasLocalData = true
- For remote datasets, realPath = path and hasLocalData = false with the URIs also indicating the remote scheme
- For symlinked mags and layers, realPath != path and hasLocalData = false
- For locally explored datasets realPath = path (as there is no layer directory and WK directly uses the path property in the mag) and hasLocalData = false

Using hasLocalData and realPath we will be able to handle deletions. 

I added a route to get a list of mags that use a mags path, this is for testing purposes and can be deleted before merge.

### TODOs:
- [x] Test symlinked mags
- [x] Test mags with path as file:// 
- [x] Store in DB / Find a way to differentiate a ds with a mag that has a file path and the ds which has the data
- [x] Use toRealPath?
- [x] When removing directory from local directory whitelist an error occurs and no more paths are added to the db

### Issues:
- fixes #6726

------
	- no user facing changes right now so probably not needed


app/controllers/WKRemoteDataStoreController.scala
```diff
@@ -221,6 +221,22 @@ class WKRemoteDataStoreController @Inject()(
...
+      request.body.validate[List[DataSourcePathInfo]] match {
+        case JsSuccess(infos, _) =>
+          for {
+            _ <- datasetService.updateRealPaths(infos)(GlobalAccessContext)
+          } yield {
+            JsonOk
+          }
+        case e: JsError =>
+          logger.warn("Data store reported invalid json for data source paths.")
+          Fox.successful(JsonBadRequest(JsError.toJson(e)))
```
fm3, line 235: Is there a specific reason to validate explicitly and log errors here, rather than using a playBodyParser to build an `Action[List[DataSourcePathInfo]]`?

app/models/dataset/Dataset.scala
```diff
@@ -733,6 +762,53 @@ class DatasetMagsDAO @Inject()(sqlClient: SqlClient)(implicit ec: ExecutionConte
     replaceSequentiallyAsTransaction(clearQuery, insertQueries)
   }
 
+  def updateMagPathsForDataset(datasetId: ObjectId, magPaths: List[MagPathInfo]): Fox[Unit] =
+    for {
+      _ <- Fox.successful(())
+      updateQueries = magPaths.map(pathInfo => {
```
fm3, line 768:
```suggestion
      updateQueries = magPathInfos.map(magPathInfo => {
```
(Seems cleaner to me if the variable names match, and also match the type. Or, for the inner-most, use an obvious abbreviation like m. That way the reader doesn’t have to understand that these things are all the same)

app/models/dataset/Dataset.scala
```diff
@@ -733,6 +762,53 @@ class DatasetMagsDAO @Inject()(sqlClient: SqlClient)(implicit ec: ExecutionConte
...
+        val magLiteral = s"(${pathInfo.mag.x}, ${pathInfo.mag.y}, ${pathInfo.mag.z})"
+        q"""UPDATE webknossos.dataset_mags
+                 SET path = ${pathInfo.path}, realPath = ${pathInfo.realPath}, haslocaldata = ${pathInfo.hasLocalData}
+                 WHERE _dataset = $datasetId
+                  AND datalayername = ${pathInfo.layerName}
```
fm3, line 773: please follow capitalization from schema.sql here, dataLayerName and hasLocalData

app/models/dataset/Dataset.scala
```diff
@@ -733,6 +762,53 @@ class DatasetMagsDAO @Inject()(sqlClient: SqlClient)(implicit ec: ExecutionConte
...
+            FROM webknossos.dataset_mags
+            WHERE realPath = $realPath""".as[DatasetMagsRow])
+      mags <- Fox.serialCombined(rows.toList)(r => parseMag(r.mag))
+      magInfos = rows.toList.zip(mags).map {
+        case (row, mag) =>
+          DatasetMagInfo(ObjectId(row._Dataset),
+                         row.datalayername,
+                         mag,
+                         row.path.getOrElse("uninitialized"),
+                         row.realpath.getOrElse("uninitialized"),
```
fm3, line 807: Should we use None for this case? I feel like having a custom string literal in the json could cause headaches down the line. What do you think?

app/models/dataset/DatasetService.scala
```diff
@@ -338,6 +340,35 @@ class DatasetService @Inject()(organizationDAO: OrganizationDAO,
...
+  private def updateRealPath(pathInfo: DataSourcePathInfo)(implicit ctx: DBAccessContext): Fox[Unit] =
+    if (pathInfo.magPathInfos.isEmpty) {
+      Fox.successful(())
+    } else {
+      val dataset = datasetDAO.findOneByDataSourceId(pathInfo.dataSourceId).futureBox
+      dataset.flatMap {
+        case Full(dataset) => datasetMagsDAO.updateMagPathsForDataset(dataset._id, pathInfo.magPathInfos)
+        case Empty => // Dataset reported but ignored (non-existing/forbidden org)
+          Fox.successful(())
+        case e: EmptyBox =>
```
fm3, line 352: Huh, TIL about the EmptyBox trait. Looks like we should have used this in a lot of places where we match against Failure (but miss ParamFailure?). But I guess we instantiate so few ParamFailures that this hasn’t caused any real problems.

app/models/dataset/DatasetService.scala
```diff
@@ -338,6 +340,35 @@ class DatasetService @Inject()(organizationDAO: OrganizationDAO,
...
+          Fox.failure("dataset.notFound", e)
+      }
+    }
+
+  def updateRealPaths(pathInfos: List[DataSourcePathInfo])(implicit ctx: DBAccessContext): Fox[Unit] =
+    for {
+      _ <- Fox.serialCombined(pathInfos)(updateRealPath)
+    } yield ()
+
+  def getPathsForDatalayer(datasetId: ObjectId, layerName: String): Fox[List[(DatasetMagInfo, List[DatasetMagInfo])]] =
```
fm3, line 362: Please use capital L in DataLayer/dataLayer for consistency. There are a couple of usages.

webknossos-datastore/app/com/scalableminds/webknossos/datastore/services/DataSourceRepository.scala
```diff
@@ -46,6 +46,12 @@ class DataSourceRepository @Inject()(
       _ <- remoteWebknossosClient.reportDataSources(dataSources)
     } yield ()
 
+  def publishRealPaths(infos: List[DataSourcePathInfo]): Fox[Unit] =
+    for {
+      _ <- Fox.successful(())
+      _ <- remoteWebknossosClient.reportRealPaths(infos)
+    } yield ()
```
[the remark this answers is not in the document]
fm3: Also, if these infos don’t live in the DataSourceRepository, I would skip this forwarding function entirely and rather call `remoteWebknossosClient.reportRealPaths(infos)` directly from the `DataSourceService`.

webknossos-datastore/app/com/scalableminds/webknossos/datastore/services/DataSourceService.scala
```diff
@@ -86,6 +88,75 @@ class DataSourceService @Inject()(
...
   }
 
+  private def sendRealPaths(dataSources: List[InboxDataSource]) =
+    for {
+      _ <- Fox.successful(())
+      magPathBoxes = dataSources.map(ds => (ds, determineMagRealPathsForDataSource(ds)))
+      pathInfos = magPathBoxes.map {
+        case (ds, Full(magPaths)) => DataSourcePathInfo(ds.id, magPaths)
+        case (ds, Failure(msg, _, _)) =>
+          logger.error(s"Failed to determine real paths for mags of $ds: $msg")
```
fm3, line 98: This should use formatFailureChain from the Formatter trait. Have a look at the other usages.

webknossos-datastore/app/com/scalableminds/webknossos/datastore/services/DataSourceService.scala
```diff
@@ -86,6 +88,75 @@ class DataSourceService @Inject()(
...
+                          unresolvedPath.toUri.toString,
+                          realPath.toUri.toString,
+                          hasLocalData = isLocal)
+            }
+          }
+        }
+      }
+      case None => List()
+    }
+  }
```
fm3, line 141: This function has grown pretty complex, maybe some of its logic could be extracted to subfunctions? That could make it easier to read. A candidate might be the block handling one layer, and also the block handling one mag.

fm3 (commented):
Looking good! My first tests worked fine :) I think the new data structures make sense.

I’m not sure if we should maybe combine the two report requests into one, to avoid this uninitialized state for new datasets, but I guess having two is also fine for the moment.

I added a couple of code suggestions.

app/models/dataset/Dataset.scala
frcroth: I did this to avoid having options in the DatasetMagInfo, since this is only a theoretical case either way. The path and realPath are set when the dataset is reported, so it would only be empty between the two report functions that are separate at the moment.

app/controllers/WKRemoteDataStoreController.scala
fm3: I did it for updateAll in https://github.com/scalableminds/webknossos/pull/8402 – but I’d say it’s not super important to update this now. Both is valid. The older has the benefit of controller-side logging (but I don’t think I ever saw this)

app/models/dataset/Dataset.scala
fm3: I wouldn’t call it a theoretical case, since there is a real time difference between the two report functions. That means that this value will, at least for short durations, occur in the database and downstream functions will have to be able to handle it. If a function lists all datasets that use the same layer, anad does string comparison, this would return all that have "uninitialized", which could become a nonsensical list. I’d rather have the Option if we cannot prevent that this value is ever inserted.

app/controllers/WKRemoteDataStoreController.scala
frcroth: Created https://github.com/scalableminds/webknossos/issues/8406 for that.

fm3 (approved):
Nice! :sunglasses: Backend LGTM.
Please remove the getLinkedMags route before merging to avoid leaking non-public information.

PR #8351, opened 2025-01-23
fix: don't use internal sun.reflect exception

jrudolph:
Probably an accidental import. It's not compatible across JDKs.


fm3 (approved):
Thank you for your contribution! Yes, that import was not intended. @normanrz Do you know how to run the CI for this branch or merge it without?

Issue #8318, opened 2025-01-10
Implement credit system


philippotto:
See [Design Doc](https://www.notion.so/scalableminds/Design-Doc-Pricing-177b51644c638038ac7fd95a18fec7da#177b51644c63808db05cd57f1958269f).

PR #8352, opened 2025-01-24
Credit system for jobs

MichaelBuessemeyer:
This PR adds a credit based paying system to WEBKNOSSOS. Organizations can purchase / order credits which can be used for premium feature in WEBKNOSSOS.

- [ ] Please double check https://github.com/scalableminds/webknossos/pull/8352#discussion_r1928994522

- https://creditsystemforjobs.webknossos.xyz/

### Steps to test:
- I suggest testing locally as it makes checking automatic refunding and marking transactions of completed jobs easier
- Setup a local worker & start wk (checked out with this branch & check that the schema is up to date)
#### General Testing
- Go to the organizations page. It should now have a new card showing how many WEBKNOSSOS credits the organization has. At the beginning this should be 0.
- Now charge up some credits via 
  ```javascript
  b = await fetch("http://localhost:9000/api/creditTransactions/chargeUpCredits?organizationId=sample_organization&creditAmount=1&moneySpent=5", { "method": "POST" });
  ```
- This should add one credit to the organization. Confirm via a reload.
- Open up your favourite dataset (best: l4_sample due to its dimensions)
- Try starting a neuron inferral on this dataset. The credits should not be enough and modal should not let you submit the job.
- Open the dev tools network tab. Now record the network request to start a mito inferral. This should be allowed and succeed.
- Try to start another mito inferral. This should be denied by the frontend as the credits are not enough.
- Now switch to the network tab where the successful mito inferral start request was recorded. Submit it again (usually available via context menu on that request. Resubmitting should fail (as not enough credits should be available)!
- Switch to organization tab. The displayed credits should be correct.
- Charge up another WEBKNOSSOS credit with the same fetch call listed above. After a reload the credit amount should have increased by one.
- Switch back to your dataset. Now the organization should have enough to start a neuron inferral. Please do so. The request should succeed.

#### Testing failed & successful jobs
- Do not have the worker running.
- Maybe refresh the schema / cancel all pending jobs as l4_sample is quite large for local testing jobs 
- Open a smaller dataset & charge up some credits as needed with the fetch call listed above. !Note the amount of credits before the job was started!
- Start some kind of job that costs credits. e.g. alignment
- In the worker code. Go to the job's function you selected to start. In that function, before the real calculation starts throw an Exception to make all jobs of this kind crash.
- Check your organisation page. The credits for the job should be subtracted
- Start the worker and let the job fail. Then stop the worker again.
- Reload the organization page and check the credits. The credits from the failed job should be automatically refunded. 
- Check the jobs page. The job should be listed as failed.
- Now undo your worker patching (remove the `throw Exception` statement).
- Not the credit amount & open up the same dataset again and submit the previous job again. 
- The credits should be subtracted.
- Start the worker and let it do its job.
- Wait until the job is finished. The credit of the organization should not have changed.
- Open the database and view the credit_transaction table. The transaction corresponding to the successful job should be listed as `state = 'Completed'`. 

### TODOs:
- [x] Write steps to test
- [x] Define 0.0001 credits as a minimum and nothing lower can be saved.
- [x] ~~test postgres scheduler (pg_cron) to run automatically~~
  - [x] do with WEBKNOSSOS backend instead
- [x] TODO comments
- [x] After starting a job, the organization's balance saved in the frontend is not updated

### Issues:
- fixes [#8318](https://github.com/scalableminds/webknossos/issues/8318)

------


app/models/organization/CreditTransaction.scala
```diff
@@ -0,0 +1,180 @@
...
+
+  def getCreditBalance(organizationId: String)(implicit ctx: DBAccessContext): Fox[BigDecimal] =
+    for {
+      accessQuery <- readAccessQuery
+      r <- run(
+        q"SELECT COALESCE(SUM(credit_change), 0) FROM $existingCollectionName WHERE _organization = $organizationId AND $accessQuery"
+          .as[BigDecimal])
+      firstRow <- r.headOption
+    } yield firstRow
+
```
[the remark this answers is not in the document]
MichaelBuessemeyer: There is one: Before an update to the transaction table is done, a trigger is true to check that the resulting transaction does not cause a negative balance. As all single update statements are executed in a automatic wrapped transaction, which encloses the trigger checking function, it should not be possible to create a situation where the balance is negative. But feel free to disagree if you see a loop hole here.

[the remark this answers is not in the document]
MichaelBuessemeyer: Important parts from the postgres docs are: https://www.postgresql.org/docs/current/tutorial-transactions.html
> PostgreSQL actually treats every SQL statement as being executed within a transaction. If you do not issue a BEGIN command, then each individual statement has an implicit BEGIN and (if successful) COMMIT wrapped around it. A group of statements surrounded by BEGIN and COMMIT is sometimes called a transaction block.

app/models/organization/CreditTransactionService.scala
```diff
@@ -0,0 +1,77 @@
...
+import play.api.libs.json.{JsObject, Json}
+
+import javax.inject.Inject
+import scala.concurrent.ExecutionContext
+
+class CreditTransactionService @Inject()(creditTransactionDAO: CreditTransactionDAO,
+                                         organizationService: OrganizationService)(implicit ec: ExecutionContext)
+    extends FoxImplicits
+    with LazyLogging {
+
```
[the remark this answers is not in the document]
MichaelBuessemeyer: The only place that is critical here should be the refunding. The other methods only have a single statement is by default wrapped in a transaction. Moreover, if a thread reads an outdated transaction or outdated balance this should not be a problem here as the user should get the newest data after a refresh. Thus, it should be fine

app/models/organization/CreditTransactionService.scala
```diff
@@ -0,0 +1,77 @@
...
+      _ <- creditTransactionDAO.insertNewPendingTransaction(pendingCreditTransaction)
+      insertedTransaction <- creditTransactionDAO.findOne(pendingCreditTransaction._id)
+    } yield insertedTransaction
+
+  def doCreditTransaction(creditTransaction: CreditTransaction)(implicit ctx: DBAccessContext): Fox[Unit] =
+    for {
+      _ <- organizationService.ensureOrganizationHasPaidPlan(creditTransaction._organization)
+      _ <- creditTransactionDAO.insertTransaction(creditTransaction)
+    } yield ()
+
```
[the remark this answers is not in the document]
MichaelBuessemeyer: Nope, the `<-` operation waits for the future (Fox) to complete and if it fails, the whole function will return and `insertTransaction` will never be executed

app/models/organization/CreditTransaction.scala
```diff
@@ -0,0 +1,180 @@
...
+    for {
+      // TODO: check write access
+      _ <- run(
+        q"""UPDATE webknossos.organization_credit_transactions
+          SET state = ${CreditTransactionState.Completed}, updated_at = NOW()
+          WHERE _id = $transactionId
+          """.asUpdate
+      )
+    } yield ()
+
```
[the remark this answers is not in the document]
MichaelBuessemeyer: Nope, Postgres automatically wraps single statements in transaction. See my comment referencing the docs above

app/models/organization/CreditTransaction.scala
```diff
@@ -0,0 +1,180 @@
...
+      // insert refund transaction
+      _ <- run(q"""INSERT INTO webknossos.organization_credit_transactions
+          (_id, _organization, credit_change, comment, _paid_job, state)
+          VALUES
+          (${ObjectId.generate}, ${refundedTransaction._organization}, ${refundAmount.toString()}::DECIMAL,
+          $refundComment, ${refundedTransaction._paidJob}, ${CreditTransactionState.Completed})
+          """.asUpdate)
+      //_ <- run(q"COMMIT TRANSACTION".as[Unit])
+    } yield ()
+
```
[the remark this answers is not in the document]
MichaelBuessemeyer: should be done now. Please check again

app/controllers/CreditTransactionController.scala
```diff
@@ -0,0 +1,68 @@
...
+        BigDecimal(creditAmount),
+        Some(moneySpentInDecimal),
+        commentNoOptional,
+        None,
+        CreditTransactionState.Completed,
+        expirationDateOpt
+      )
+      _ <- creditTransactionService.doCreditTransaction(chargeUpTransaction)
+    } yield Ok
+  }
```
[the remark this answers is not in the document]
MichaelBuessemeyer: This is not currently relevant to our domain. Charge-ups happen rarely and in case a user reads outdated values, this isn't a problem to use as long as the data in the database is consistent.

tools/postgres/schema.sql
```diff
@@ -944,3 +967,128 @@ CREATE TRIGGER onDeleteAnnotationTrigger
...
+    END IF;
+    -- Allow the transaction
+    RETURN NEW;
+  END;
+$$ LANGUAGE plpgsql;
+
+
+CREATE TRIGGER enforce_balance_trigger
+BEFORE INSERT OR UPDATE ON webknossos.organization_credit_transactions
+FOR EACH ROW EXECUTE PROCEDURE webknossos.enforce_non_negative_balance();
```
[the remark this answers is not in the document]
MichaelBuessemeyer: credit change cannot be null as state by the schema and `AND state = 'Completed'` would not count refunded  transaction (and other like pending). But refunded transactions are still counted as deduced but a new transaction is created that refunds the same spent amount upon refunding. So to get an accurate balance of an organization all their transactions need to be used in the calculation

conf/evolutions/126-credit-transactions.sql
```diff
@@ -0,0 +1,160 @@
...
+
+                -- Update transaction state to 'REVOKED'
+                UPDATE webknossos.organization_credit_transactions
+                SET state = 'Revoked', updated_at = NOW()
+                WHERE id = free_credits_transaction.id;
+
+                -- Add the date to the revoked dates set
+                -- (In PostgreSQL, we don't need a set; we will use it for information in the comment)
+            END IF;
+        END LOOP;
```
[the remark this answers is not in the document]
MichaelBuessemeyer: I'd say this is not important / relevant. But this should be discussed in the review

MichaelBuessemeyer:
@fm3 I'd say this pr is ready for a first review round

app/models/organization/FreeCreditTransactionService.scala
```diff
@@ -0,0 +1,37 @@
...
+                                             val system: ActorSystem)(implicit val ec: ExecutionContext)
+    extends Controller
+    with FoxImplicits
+    with LazyLogging
+    with IntervalScheduler {
+
+  def revokeExpiredCredits(): Fox[Unit] = creditTransactionDAO.runRevokeExpiredCredits()
+  def handOutMonthlyFreeCredits(): Fox[Unit] = creditTransactionDAO.handOutMonthlyFreeCredits()
+
+  override protected def tickerInterval: FiniteDuration = 20 minute
```
fm3, line 25:
```suggestion
  override protected def tickerInterval: FiniteDuration = 20 minutes
```

app/models/job/JobService.scala
```diff
@@ -219,11 +220,30 @@ class JobService @Inject()(wkConf: WkConf,
...
       _ <- bool2Fox(boundingBoxInMag.volume <= wkConf.Features.exportTiffMaxVolumeMVx * 1024 * 1024) ?~> "job.volumeExceeded"
       _ <- bool2Fox(boundingBoxInMag.size.maxDim <= wkConf.Features.exportTiffMaxEdgeLengthVx) ?~> "job.edgeLengthExceeded"
     } yield ()
 
+  private def getJobCostsPerGVx(jobCommand: JobCommand): BigDecimal =
+    jobCommand match {
+      case JobCommand.infer_neurons      => wkConf.Features.neuronInferralCostsPerGVx
+      case JobCommand.infer_mitochondria => wkConf.Features.mitochondriaInferralCostsPerGVx
+      case JobCommand.align_sections     => wkConf.Features.alignmentCostsPerGVx
+      case _                             => throw new IllegalArgumentException(s"Unsupported job command $jobCommand")
```
fm3, line 233: Please return a Failure box instead of throwing the exception.

app/models/job/JobService.scala
```diff
@@ -219,11 +220,30 @@ class JobService @Inject()(wkConf: WkConf,
...
+    }
+
+  def calculateJobCosts(boundingBoxInTargetMag: BoundingBox, jobCommand: JobCommand): BigDecimal = {
+    val costsPerGVx = getJobCostsPerGVx(jobCommand)
+    val volumeInGVx = boundingBoxInTargetMag.volume / math.pow(10, 9)
+    val costs = BigDecimal(volumeInGVx) * costsPerGVx
+    if (costs < BigDecimal(0.001)) BigDecimal(0.001) else costs.setScale(4, RoundingMode.HALF_UP)
```
fm3, line 240: The number 0.001 should be extracted to a variable.

app/models/job/JobService.scala
```diff
@@ -219,11 +220,30 @@ class JobService @Inject()(wkConf: WkConf,
...
+    jobCommand match {
+      case JobCommand.infer_neurons      => wkConf.Features.neuronInferralCostsPerGVx
+      case JobCommand.infer_mitochondria => wkConf.Features.mitochondriaInferralCostsPerGVx
+      case JobCommand.align_sections     => wkConf.Features.alignmentCostsPerGVx
+      case _                             => throw new IllegalArgumentException(s"Unsupported job command $jobCommand")
+    }
+
+  def calculateJobCosts(boundingBoxInTargetMag: BoundingBox, jobCommand: JobCommand): BigDecimal = {
+    val costsPerGVx = getJobCostsPerGVx(jobCommand)
+    val volumeInGVx = boundingBoxInTargetMag.volume / math.pow(10, 9)
```
fm3, line 238: maybe avoid rounding here?

app/models/organization/CreditTransaction.scala
```diff
@@ -0,0 +1,364 @@
...
+        jobIdOpt,
+        state,
+        row.expirationDate.map(Instant.fromDate),
+        Instant.fromSql(row.createdAt),
+        Instant.fromSql(row.updatedAt),
+        row.isDeleted
+      )
+    }
+
+  implicit val getOrganizationCreditTransactions: GetResult[CreditTransaction] = GetResult { prs =>
```
fm3, line 72: is the generated function from Tables.scala not suitable?

app/models/organization/CreditTransaction.scala
```diff
@@ -0,0 +1,364 @@
...
+      state,
+      expiresAt,
+      createdAt,
+      updatedAt,
+      isDeleted
+    )
+  }
+
+  override protected def readAccessQ(requestingUserId: ObjectId): SqlToken =
+    q"""(_id IN (SELECT _organization FROM webknossos.users_ WHERE _multiUser = (SELECT _multiUser FROM webknossos.users_ WHERE _id = $requestingUserId)))
```
fm3, line 106: Who should be able to see the own organization’s credit balance? 
Admins, Team Managers, Members?
cc @normanrz

app/models/organization/CreditTransaction.scala
```diff
@@ -0,0 +1,364 @@
...
+    for {
+      accessQuery <- readAccessQuery
+      r <- run(
+        q"SELECT $columns FROM $existingCollectionName WHERE _paid_job = $jobId AND $accessQuery"
+          .as[OrganizationCreditTransactionsRow])
+      parsed <- parseFirst(r, jobId)
+    } yield parsed
+
+// TODO: Maybe remove this outer while loop here.
+  private def findAllOrganizationsWithExpiredCredits: Fox[List[String]] =
```
fm3, line 256:
```suggestion
  private def findAllOrganizationsWithExpiredCredits: Fox[List[ObjectId]] =
```

app/models/organization/CreditTransaction.scala
```diff
@@ -0,0 +1,364 @@
...
+  def findTransactionForJob(jobId: ObjectId)(implicit ctx: DBAccessContext): Fox[CreditTransaction] =
+    for {
+      accessQuery <- readAccessQuery
+      r <- run(
+        q"SELECT $columns FROM $existingCollectionName WHERE _paid_job = $jobId AND $accessQuery"
+          .as[OrganizationCreditTransactionsRow])
+      parsed <- parseFirst(r, jobId)
+    } yield parsed
+
+// TODO: Maybe remove this outer while loop here.
```
fm3, line 255: where?

app/models/organization/CreditTransactionService.scala
```diff
@@ -0,0 +1,92 @@
...
+import javax.inject.Inject
+import scala.concurrent.ExecutionContext
+
+class CreditTransactionService @Inject()(creditTransactionDAO: CreditTransactionDAO,
+                                         organizationService: OrganizationService)(implicit val ec: ExecutionContext)
+    extends FoxImplicits
+    with LazyLogging {
+
+  def hasEnoughCredits(organizationId: String, creditsToSpent: BigDecimal)(
```
fm3, line 17:
```suggestion
  def hasEnoughCredits(organizationId: String, creditsToSpend: BigDecimal)(
```

app/models/organization/CreditTransactionService.scala
```diff
@@ -0,0 +1,92 @@
...
+      implicit ctx: DBAccessContext): Fox[Boolean] =
+    creditTransactionDAO.getCreditBalance(organizationId).map(balance => balance >= creditsToSpent)
+
+  def getCreditBalance(organizationId: String)(implicit ctx: DBAccessContext): Fox[BigDecimal] =
+    creditTransactionDAO.getCreditBalance(organizationId)
```
fm3, line 22: I’m not a big fan of these single-line redirect functions. I’d rather inject the dao where getCreditBalance is needed.

app/models/organization/CreditTransactionService.scala
```diff
@@ -0,0 +1,92 @@
...
+                                                   None,
+                                                   comment,
+                                                   paidJob,
+                                                   CreditTransactionState.Pending,
+                                                   None)
+      _ <- creditTransactionDAO.insertNewPendingTransaction(pendingCreditTransaction)
+      insertedTransaction <- creditTransactionDAO.findOne(pendingCreditTransaction._id)
+    } yield insertedTransaction
+
+  def doCreditTransaction(creditTransaction: CreditTransaction)(implicit ctx: DBAccessContext): Fox[Unit] =
```
fm3, line 42: How about »add« or »insert« rather than »do«?

app/models/organization/CreditTransaction.scala
```diff
@@ -0,0 +1,364 @@
...
+        case (failedTransactions, organizationId) =>
+          run(revokeExpiredCreditsForOrganization(organizationId).transactionally).map(failedTransactions ++ _)
+      }
+      _ = if (failedTransactionsToRevoke.nonEmpty) {
+        val failedTransactions = failedTransactionsToRevoke.map(transaction =>
+          s"Failed to revoke credits for transaction ${transaction._id} for organization ${transaction._organization}.")
+        slackNotificationService.warn("Failed to revoke some expired credits for organizations",
+                                      s"${failedTransactions.mkString("\n")}")
+      }
+    } yield ()
```
fm3, line 359: I’m not sure I understand these Revoke functions. Maybe let’s talk about this in person.

app/controllers/CreditTransactionController.scala
```diff
@@ -0,0 +1,79 @@
...
+class CreditTransactionController @Inject()(organizationService: OrganizationService,
+                                            creditTransactionService: CreditTransactionService,
+                                            freeCreditTransactionService: FreeCreditTransactionService,
+                                            creditTransactionDAO: CreditTransactionDAO,
+                                            userService: UserService,
+                                            sil: Silhouette[WkEnv])(implicit ec: ExecutionContext)
+    extends Controller
+    with FoxImplicits {
+
+  def chargeUpCredits(organizationId: String,
```
fm3, line 32: I’m not familiar with the wording “charge up”. Where does this come from?

tools/postgres/schema.sql
```diff
@@ -349,6 +349,26 @@ CREATE TABLE webknossos.organization_usedStorage(
...
 
+-- Create the enum type for transaction states
+CREATE TYPE webknossos.credit_transaction_state AS ENUM ('Pending', 'Completed', 'Refunded', 'Revoked', 'PartiallyRevoked', 'Spent');
+
+-- Create the transactions table
+CREATE TABLE webknossos.organization_credit_transactions (
+    _id CHAR(24) PRIMARY KEY,
+    _organization VARCHAR(256) NOT NULL,
+    credit_change DECIMAL(14, 4) NOT NULL,
+    refundable_credit_change DECIMAL(14, 4) NOT NULL CHECK (refundable_credit_change >= 0), -- Ensure non-negative values
```
fm3, line 360: How did you decide that this precision 4 is what we need?

fm3 (changes requested):
Sweet! This already looks pretty good! I added a couple of comments. I’d also like to discuss the revoke logic in person, as I don’t really understand it on first sight, and I think it’s important to handle those cases correctly.

MichaelBuessemeyer (commented):
So here we go :)

Thanks a lot for all the valuable feedback. I think everything should be addressed now an after all these changes everything should still be working (hopefully). I already did some testing but did e.g. left out testing the revoking functionality.

conf/application.conf
```diff
@@ -151,8 +151,11 @@ features {
   taskReopenAllowedInSeconds = 30
   allowDeleteDatasets = true
   # to enable jobs for local development, use "yarn enable-jobs" to also activate it in the database
-  jobsEnabled = false
-  voxelyticsEnabled = false
+  jobsEnabled = true
+  voxelyticsEnabled = true
+  neuronInferralCostsPerGVx = 1
+  mitochondriaInferralCostsPerGVx = 0.5
+  alignmentCostsPerGVx = 0.5
```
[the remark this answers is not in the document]
MichaelBuessemeyer: Ok, the AI says you are also more correct here:
image

MichaelBuessemeyer:
TODO:
- [ ] Check whether revoking is running correctly in a transaction

MichaelBuessemeyer:
@normanrz / @fm3 or who else has an opinion:
How should I calculate what a Gigavoxel is? 
Is 
1. 1 GVx = 1000 * 1000 * 1000 or 
2. 1 GVx = 1024 × 1024 × 1024? 

Because initially, I was using 1. but the frontend already has a function to calculate & format the VX of a bbox and uses the 2. ways of calculation :thinking:

fm3:
I’d vote for decimal (this would be consistent with how we display amounts of bytes from the storage scan), but I also see the appeal of binary. I’d leave the final decision to @normanrz (maybe @philippotto has an opinion too?)

MichaelBuessemeyer:
Ok I gave the transaction handling some thorough testing:
1. I added a sytax error in the last query of the revoking transaction (the insert of the revoking transaction) and this made the transaction crash without any changes to the state of transaction which handed out the free credits (which should have happened but were rolled backen). Moreover, no revoking transaction was present. Once I corrected the intentional syntax error in the query,  everything worked perfectly.
2. I added extensive logging to postgres and here are the relevant log parts:
<details>
<summary>Postgres Logs</summary><blockquote>
The postgres logs. All parameterized queries are printed 3 times: 1. parsed, 2. provided with parameters, 3. executed.

```
2025-03-04 10:12:24.445 CET [1460013] postgres@webknossos LOG:  duration: 0.166 ms  parse S_6: SELECT DISTINCT _organization
	        FROM webknossos.credit_transactions
	        WHERE expiration_date <= NOW()
	        AND credit_state = $1
	        AND credit_change > 0
2025-03-04 10:12:24.446 CET [1460013] postgres@webknossos LOG:  duration: 0.202 ms  bind S_6: SELECT DISTINCT _organization
	        FROM webknossos.credit_transactions
	        WHERE expiration_date <= NOW()
	        AND credit_state = $1
	        AND credit_change > 0
2025-03-04 10:12:24.446 CET [1460013] postgres@webknossos DETAIL:  Parameters: $1 = 'Pending'
2025-03-04 10:12:24.446 CET [1460013] postgres@webknossos LOG:  execute S_6: SELECT DISTINCT _organization
	        FROM webknossos.credit_transactions
	        WHERE expiration_date <= NOW()
	        AND credit_state = $1
	        AND credit_change > 0
2025-03-04 10:12:24.446 CET [1460013] postgres@webknossos DETAIL:  Parameters: $1 = 'Pending'
2025-03-04 10:12:24.446 CET [1460013] postgres@webknossos LOG:  duration: 0.106 ms
2025-03-04 10:12:24.447 CET [1460013] postgres@webknossos LOG:  duration: 0.011 ms  bind S_4: BEGIN
2025-03-04 10:12:24.447 CET [1460013] postgres@webknossos LOG:  execute S_4: BEGIN
2025-03-04 10:12:24.447 CET [1460013] postgres@webknossos LOG:  duration: 0.016 ms
2025-03-04 10:12:24.447 CET [1460013] postgres@webknossos LOG:  duration: 0.353 ms  parse <unnamed>: SELECT *
	            FROM webknossos.credit_transactions
	            WHERE _organization = $1
	              AND expiration_date <= NOW()
	              AND credit_state = $2
	              AND credit_change > 0
	            ORDER BY created_at DESC
	         
2025-03-04 10:12:24.448 CET [1460013] postgres@webknossos LOG:  duration: 0.201 ms  bind <unnamed>: SELECT *
	            FROM webknossos.credit_transactions
	            WHERE _organization = $1
	              AND expiration_date <= NOW()
	              AND credit_state = $2
	              AND credit_change > 0
	            ORDER BY created_at DESC
	         
2025-03-04 10:12:24.448 CET [1460013] postgres@webknossos DETAIL:  Parameters: $1 = 'sample_organization', $2 = 'Pending'
2025-03-04 10:12:24.448 CET [1460013] postgres@webknossos LOG:  execute <unnamed>: SELECT *
	            FROM webknossos.credit_transactions
	            WHERE _organization = $1
	              AND expiration_date <= NOW()
	              AND credit_state = $2
	              AND credit_change > 0
	            ORDER BY created_at DESC
	         
2025-03-04 10:12:24.448 CET [1460013] postgres@webknossos DETAIL:  Parameters: $1 = 'sample_organization', $2 = 'Pending'
2025-03-04 10:12:24.448 CET [1460013] postgres@webknossos LOG:  duration: 0.127 ms
2025-03-04 10:12:24.449 CET [1460013] postgres@webknossos LOG:  duration: 0.374 ms  parse <unnamed>: 
	      SELECT COALESCE(SUM(credit_change), 0)
	      FROM webknossos.credit_transactions
	      WHERE _organization = $1
	        AND created_at >= $2::TIMESTAMPTZ
	        AND (
			      credit_change < 0
	          OR (credit_change > 0 AND _related_transaction IS NOT NULL AND credit_state = 'Refunding') -- Counts refunding transactions
	          OR (credit_change > 0 AND expiration_date <= NOW()) -- Counts also expired transactions
		      )
			
2025-03-04 10:12:24.449 CET [1460013] postgres@webknossos LOG:  duration: 0.206 ms  bind <unnamed>: 
	      SELECT COALESCE(SUM(credit_change), 0)
	      FROM webknossos.credit_transactions
	      WHERE _organization = $1
	        AND created_at >= $2::TIMESTAMPTZ
	        AND (
			      credit_change < 0
	          OR (credit_change > 0 AND _related_transaction IS NOT NULL AND credit_state = 'Refunding') -- Counts refunding transactions
	          OR (credit_change > 0 AND expiration_date <= NOW()) -- Counts also expired transactions
		      )
			
2025-03-04 10:12:24.449 CET [1460013] postgres@webknossos DETAIL:  Parameters: $1 = 'sample_organization', $2 = '2025-03-04 09:56:22.613+01'
2025-03-04 10:12:24.449 CET [1460013] postgres@webknossos LOG:  execute <unnamed>: 
	      SELECT COALESCE(SUM(credit_change), 0)
	      FROM webknossos.credit_transactions
	      WHERE _organization = $1
	        AND created_at >= $2::TIMESTAMPTZ
	        AND (
			      credit_change < 0
	          OR (credit_change > 0 AND _related_transaction IS NOT NULL AND credit_state = 'Refunding') -- Counts refunding transactions
	          OR (credit_change > 0 AND expiration_date <= NOW()) -- Counts also expired transactions
		      )
			
2025-03-04 10:12:24.449 CET [1460013] postgres@webknossos DETAIL:  Parameters: $1 = 'sample_organization', $2 = '2025-03-04 09:56:22.613+01'
2025-03-04 10:12:24.450 CET [1460013] postgres@webknossos LOG:  duration: 0.044 ms
2025-03-04 10:12:24.450 CET [1460013] postgres@webknossos LOG:  duration: 0.076 ms  parse <unnamed>: 
	            UPDATE webknossos.credit_transactions
	            SET credit_state = $1, updated_at = NOW()
	            WHERE _id = $2
	          
2025-03-04 10:12:24.450 CET [1460013] postgres@webknossos LOG:  duration: 0.101 ms  bind <unnamed>: 
	            UPDATE webknossos.credit_transactions
	            SET credit_state = $1, updated_at = NOW()
	            WHERE _id = $2
	          
2025-03-04 10:12:24.450 CET [1460013] postgres@webknossos DETAIL:  Parameters: $1 = 'Revoked', $2 = '67c6c03668852613c2000003'
2025-03-04 10:12:24.450 CET [1460013] postgres@webknossos LOG:  execute <unnamed>: 
	            UPDATE webknossos.credit_transactions
	            SET credit_state = $1, updated_at = NOW()
	            WHERE _id = $2
	          
2025-03-04 10:12:24.450 CET [1460013] postgres@webknossos DETAIL:  Parameters: $1 = 'Revoked', $2 = '67c6c03668852613c2000003'
2025-03-04 10:12:24.451 CET [1460013] postgres@webknossos LOG:  duration: 0.742 ms
2025-03-04 10:12:24.452 CET [1460013] postgres@webknossos LOG:  duration: 0.171 ms  parse <unnamed>: INSERT INTO webknossos.credit_transactions
	          (_id, _organization, credit_change, comment, _paid_job,
	          transaction_state, credit_state, expiration_date, created_at, updated_at, is_deleted)
	          VALUES
	          ($1, $2, $3::DECIMAL,
	          $4, $5, $6, $7,
	          $8, $9::TIMESTAMPTZ, $10::TIMESTAMPTZ, $11)
	          
2025-03-04 10:12:24.453 CET [1460013] postgres@webknossos LOG:  duration: 0.169 ms  bind <unnamed>: INSERT INTO webknossos.credit_transactions
	          (_id, _organization, credit_change, comment, _paid_job,
	          transaction_state, credit_state, expiration_date, created_at, updated_at, is_deleted)
	          VALUES
	          ($1, $2, $3::DECIMAL,
	          $4, $5, $6, $7,
	          $8, $9::TIMESTAMPTZ, $10::TIMESTAMPTZ, $11)
	          
2025-03-04 10:12:24.453 CET [1460013] postgres@webknossos DETAIL:  Parameters: $1 = '67c6c3f8af251527b7e1c111', $2 = 'sample_organization', $3 = '-2.0000', $4 = 'Revoked expired credits for transaction 67c6c03668852613c2000003', $5 = NULL, $6 = 'Complete', $7 = 'Revoking', $8 = NULL, $9 = '2025-03-04 10:12:24.45+01', $10 = '2025-03-04 10:12:24.45+01', $11 = 'f'
2025-03-04 10:12:24.453 CET [1460013] postgres@webknossos LOG:  execute <unnamed>: INSERT INTO webknossos.credit_transactions
	          (_id, _organization, credit_change, comment, _paid_job,
	          transaction_state, credit_state, expiration_date, created_at, updated_at, is_deleted)
	          VALUES
	          ($1, $2, $3::DECIMAL,
	          $4, $5, $6, $7,
	          $8, $9::TIMESTAMPTZ, $10::TIMESTAMPTZ, $11)
	          
2025-03-04 10:12:24.453 CET [1460013] postgres@webknossos DETAIL:  Parameters: $1 = '67c6c3f8af251527b7e1c111', $2 = 'sample_organization', $3 = '-2.0000', $4 = 'Revoked expired credits for transaction 67c6c03668852613c2000003', $5 = NULL, $6 = 'Complete', $7 = 'Revoking', $8 = NULL, $9 = '2025-03-04 10:12:24.45+01', $10 = '2025-03-04 10:12:24.45+01', $11 = 'f'
2025-03-04 10:12:24.453 CET [1460013] postgres@webknossos LOG:  duration: 0.750 ms
2025-03-04 10:12:24.454 CET [1460013] postgres@webknossos LOG:  duration: 0.016 ms  bind S_1: COMMIT
2025-03-04 10:12:24.454 CET [1460013] postgres@webknossos LOG:  execute S_1: COMMIT
2025-03-04 10:12:24.456 CET [1460013] postgres@webknossos LOG:  duration: 1.400 ms
2025-03-04 10:12:24.457 CET [1460013] postgres@webknossos LOG:  duration: 0.077 ms  parse S_7: SELECT webknossos.hand_out_monthly_free_credits($1::DECIMAL)
2025-03-04 10:12:24.457 CET [1460013] postgres@webknossos LOG:  duration: 0.063 ms  bind S_7: SELECT webknossos.hand_out_monthly_free_credits($1::DECIMAL)
2025-03-04 10:12:24.457 CET [1460013] postgres@webknossos DETAIL:  Parameters: $1 = '2'
2025-03-04 10:12:24.457 CET [1460013] postgres@webknossos LOG:  execute S_7: SELECT webknossos.hand_out_monthly_free_credits($1::DECIMAL)
2025-03-04 10:12:24.457 CET [1460013] postgres@webknossos DETAIL:  Parameters: $1 = '2'
2025-03-04 10:12:24.460 CET [1460013] postgres@webknossos LOG:  duration: 2.430 ms
```

</details>

So IMO it should be pretty damn save to assume that the revoking of transactions of a single organization runs in a transaction which is rolled back upon errors

conf/evolutions/127-credit-transactions.sql
```diff
@@ -0,0 +1,113 @@
...
+    comment TEXT NOT NULL,
+    -- The state of the transaction.
+    transaction_state webknossos.credit_transaction_state NOT NULL,
+    -- The state of the credits of this transaction.
+    credit_state webknossos.credit_transaction_state NOT NULL,
+    expiration_date TIMESTAMPTZ,
+    created_at TIMESTAMPTZ NOT NULL DEFAULT NOW(),
+    updated_at TIMESTAMPTZ NOT NULL DEFAULT NOW(),
+    is_deleted BOOLEAN NOT NULL DEFAULT FALSE
+);
```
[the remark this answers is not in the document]
MichaelBuessemeyer: jup nice, thanks for the find :t-rex:

app/models/job/JobService.scala
```diff
@@ -219,11 +220,30 @@ class JobService @Inject()(wkConf: WkConf,
...
+    val costs = BigDecimal(volumeInGVx) * costsPerGVx
+    if (costs < BigDecimal(0.001)) BigDecimal(0.001) else costs.setScale(4, RoundingMode.HALF_UP)
+  }
```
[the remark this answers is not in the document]
fm3: I’d say it should be a named constant in the code (maybe with a comment explaining that it must match the precision in the db). Should not be part of the config.

tools/postgres/schema.sql
fm3: In the linked discussion I always read millicredits as the integer unit. Wouldn’t that be Decimal(14,3)? Is the current plan to render four decimal places to the user? Feels like a bit much to me, tbh. (But maybe that’s just me, as I’m used to dollars and euros, which tend to be rounded to centieuros, resulting in just two decimal places)

app/controllers/CreditTransactionController.scala
fm3: How about add credit?

tools/postgres/schema.sql
```diff
@@ -349,6 +349,26 @@ CREATE TABLE webknossos.organization_usedStorage(
   PRIMARY KEY(_organization, _dataStore, _dataset, layerName, magOrDirectoryName)
 );
 
+-- Create the enum type for transaction states
+CREATE TYPE webknossos.credit_transaction_state AS ENUM ('Pending', 'Completed', 'Refunded', 'Revoked', 'PartiallyRevoked', 'Spent');
```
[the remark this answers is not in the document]
fm3: Please add this full list of comments to the sql schema enum definition (and add a comment in code saying see schema.sql for details about the enum states).

tools/postgres/schema.sql
```diff
@@ -944,3 +969,79 @@ CREATE TRIGGER onDeleteAnnotationTrigger
 AFTER DELETE ON webknossos.annotations
 FOR EACH ROW EXECUTE PROCEDURE webknossos.onDeleteAnnotation();
 
+CREATE FUNCTION webknossos.enforce_non_negative_balance() RETURNS TRIGGER AS $$
+  BEGIN
+    -- Assert that the new balance is non-negative
+    ASSERT (SELECT COALESCE(SUM(credit_change), 0) + COALESCE(NEW.credit_change, 0)
+            FROM webknossos.organization_credit_transactions
+            WHERE _organization = NEW._organization AND _id != NEW._id) >= 0, 'Transaction would result in a negative credit balance for organization %', NEW._organization;
+    -- Allow the transaction
```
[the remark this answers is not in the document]
fm3: I see, thanks for the explanation! Maybe the comment could be expanded to make this more obvious to future readers. Maybe “Assertion passed, transaction can go ahead”?

app/models/organization/CreditTransaction.scala
fm3: Ok fair enough. I think in other spots we use the corresponding autoGenerated Row reader and then a parse method that converts this to our own case class. But that might not work here, I’m not super familiar with the DBIO handling context.

app/models/organization/CreditTransaction.scala
fm3: Ok let’s defer this to the next iteration then.

app/controllers/CreditTransactionController.scala
```diff
@@ -0,0 +1,82 @@
...
+    for {
+      _ <- userService.assertIsSuperUser(request.identity) ?~> "Only super users can charge up credits"
+      moneySpentInDecimal <- tryo(BigDecimal(moneySpent)) ?~> s"moneySpent $moneySpent is not a valid decimal"
+      _ <- bool2Fox(moneySpentInDecimal > 0) ?~> "moneySpent must be a positive number"
+      _ <- bool2Fox(creditAmount > 0) ?~> "creditAmount must be a positive number"
+      commentNoOptional = comment.getOrElse(s"Charge up for $creditAmount credits for $moneySpentInDecimal $currency.")
+      _ <- organizationService.assertOrganizationHasPaidPlan(organizationId)
+      expirationDateOpt <- Fox.runOptional(expiresAt)(Instant.fromString)
+      _ <- Fox
+        .runOptional(expirationDateOpt)(expirationDate => bool2Fox(Instant.now < expirationDate)) ?~> "Expiration date must be in the future"
```
fm3, line 48:
```suggestion
        .runOptional(expirationDateOpt)(expirationDate => bool2Fox(!expirationDate.isPast)) ?~> "Expiration date must be in the future"
```

app/controllers/JobController.scala
```diff
@@ -85,6 +88,7 @@ class JobController @Inject()(
     for {
       _ <- bool2Fox(wkconf.Features.jobsEnabled) ?~> "job.disabled"
       jobs <- jobDAO.findAll
+      // TODO: Consider adding paid credits to job public writes.
```
fm3, line 91: yes, probably a good idea

app/controllers/JobController.scala
```diff
@@ -495,4 +518,20 @@ class JobController @Inject()(
       } yield Redirect(uri, Map(("token", Seq(userAuthToken.id))))
     }
 
+  def getJobCosts(command: String, boundingBoxInMag: String): Action[AnyContent] =
```
fm3, line 521: this one still uses costs in plural

fm3 (commented):
Getting there :) I added a few more comments

app/models/organization/FreeCreditTransactionService.scala
```diff
@@ -0,0 +1,40 @@
...
+
+class FreeCreditTransactionService @Inject()(creditTransactionDAO: CreditTransactionDAO,
+                                             val lifecycle: ApplicationLifecycle,
+                                             val system: ActorSystem)(implicit val ec: ExecutionContext)
+    extends Controller
+    with FoxImplicits
+    with LazyLogging
+    with IntervalScheduler {
+
+  override protected def actorSystem: ActorSystem = system
```
fm3, line 22: This line can be removed if the injected `val system: ActorSystem` is renamed to `val actorSystem: ActorSystem`.

frontend/javascripts/admin/api/jobs.ts
```diff
@@ -74,6 +74,20 @@ export async function cancelJob(jobId: string): Promise<APIJob> {
...
+  hasEnoughCredits: boolean;
+  organizationCredits: string;
+};
+
+export async function getJobCost(command: string, boundingBoxInMag: Vector6): Promise<JobCostInfo> {
+  const params = new URLSearchParams({
+    command,
+    boundingBoxInMag: boundingBoxInMag.join(","),
+  });
+  return await Request.receiveJSON(`/api/jobs/getCosts?${params}`);
```
philippotto, line 88: the function uses "cost" and the route "costs"

frontend/javascripts/admin/api/jobs.ts
```diff
@@ -74,6 +74,20 @@ export async function cancelJob(jobId: string): Promise<APIJob> {
   });
 }
 
+export type JobCostInfo = {
+  costInCredits: string;
```
philippotto, line 78: how is the cost encoded in the string? can you add a comment?

philippotto (changes requested):
I left a first round of review for the frontend. Overall it looks good :+1: didn't test, though.
regarding how the credits are stored, I don't feel like the prior discussion in the design doc was really taken into account.
originally I suggested that floats should be enough when the UI rounds carefully and that decimals cannot be used in JS without large effort. counter-suggestions were to use decimals or milli-credits with trade-offs and details being discussed.
now, decimals are used in the DB which requires string encoding in JSON. these strings are partly kept as strings in JS and sometimes parsed as float (in the store, I think). additionally, rounding is done but not without the required care.
I feel like this is unnecessary complex without being any better than using floats everywhere.

MichaelBuessemeyer (commented):
Here are my comments regarding fm3's review. Changes should be pushed now

MichaelBuessemeyer:
> I feel like this is unnecessary complex without being any better than using floats everywhere.

You mean float in the frontend or in the backend & database as well. I talked this through with valentin-pinkau and he mentioned that precision here is key with setting a maximal minimum for all transactions (0.0001 credits). Moreover, cost calculation should be consistent and that's why having a single point for such a calculation is important. This is done by the backend and the frontend requests the cost of a job accordingly.
Therefore, I'd say that the frontend does not need to do any calculation with the string. Meaning even rounding shouldn't be necessary.

> these strings are partly kept as strings in JS and sometimes parsed as float (in the store, I think). 

Yeah, you are right. I forgot to migrate the type of `APIOrganization[creditBalance]` to string. As the idea is to do all calculation in the backend to ensure correctness, the frontend IMO only needs the credit balance or cost information to display the data but no calculation should be necessary anywhere.

What do you think about this approach? This avoid all the necessary care that we would have to take when working with float.

frontend/javascripts/oxalis/view/action-bar/starting_job_modals.tsx
```diff
@@ -527,6 +546,108 @@ function ShouldUseTreesFormItem() {
   );
 }
 
+function JobCostInformation({
+  jobCostInfo,
+  jobCreditCostPerGVx,
```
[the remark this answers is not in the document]
MichaelBuessemeyer: I should now be consistent except for the variables in the application.conf / features defining the cost per GVx. Should these also be adjusted?

frontend/javascripts/oxalis/view/action-bar/starting_job_modals.tsx
```diff
@@ -611,29 +762,51 @@ function StartJobForm(props: StartJobFormProps) {
...
+      boundingBoxForJob
+        ? await getJobCost(jobName, computeArrayFromBoundingBox(boundingBoxForJob.boundingBox))
+        : undefined,
+    undefined,
+    [boundingBoxForJob, jobName],
+  );
+
+  useEffect(() => {
+    const orgaCreditsAsNumber = jobCostInfo
+      ? Number.parseFloat(jobCostInfo.organizationCredits)
```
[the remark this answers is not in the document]
MichaelBuessemeyer: Yeah the parsing was needed as it the type in the store was still number and not string which allows the arbitrary precision. I changed the type in the store to string to keep the precision.

Thanks for the handup here.

frontend/javascripts/admin/organization/organization_cards.tsx
```diff
@@ -252,46 +254,67 @@ export function PlanDashboardCard({
...
+            </Row>
+            <Row justify="center">Current Plan</Row>
+          </Card>
+        </Col>
+        <Col span={12}>
+          <Card actions={buyMoreCreditsAction}>
+            <Row justify="center" align="middle" style={{ minHeight: 160 }}>
+              <h3>
+                {organization.creditBalance != null
+                  ? roundTo(organization.creditBalance, 4)
```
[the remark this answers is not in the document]
MichaelBuessemeyer: Yeah, I forgot to change this when I switched to sending only cost info in form of string to the frontend as it does not need to do any calculation.

Should be better now as `organization.creditBalance` is of type string.

frontend/javascripts/oxalis/view/action-bar/starting_job_modals.tsx
```diff
@@ -677,6 +850,11 @@ function StartJobForm(props: StartJobFormProps) {
       if (!apiJob) {
         return;
       }
+      if (jobCreditCostPerGVx != null && activeUser?.organization) {
+        // As the job did cost credits, refetch the organization to have a correct credit balance.
+        const updatedOrganization = await getOrganization(activeUser?.organization);
```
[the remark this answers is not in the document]
MichaelBuessemeyer: > what happens when this route fails? then, the balance will be wrong and the user could try to trigger another job (even though the credits are not enough). but then the job submission would simply fail.

All should be yes but before starting the new job, the frontend requests its cost which then should show, that the organization has not enough credits.

> so, probably not an issue?

Yeah, I'd also say that. Adding error handling here and then upon error showing a toast that the organization credits could not be refresh, might lead to confusion instead of helping :thinking:

Maybe wrap this in a try & catch and to nothing upon error :thinking:?

MichaelBuessemeyer (commented):
Thanks a lot for all the feedback :pray: 

Now, I also incorporated the frontend feedback / commented on threads.
But please wait before fully re-reviewing as I still need to test and would defer the testing till next week in favour of other prs.

MichaelBuessemeyer:
Ok I gave this another testing session. I left out the auto-revoking mechanism as I think I changed nothing there and will retest anyway before the final merge.

So this should be ready for another review round :dancers:

philippotto:
> Yeah, you are right. I forgot to migrate the type of `APIOrganization[creditBalance]` to string. As the idea is to do all calculation in the backend to ensure correctness, the frontend IMO only needs the credit balance or cost information to display the data but no calculation should be necessary anywhere.
> 
> What do you think about this approach? This avoid all the necessary care that we would have to take when working with float.

Ok, if the frontend only handles strings and doesn't do any arithmetics with that, I understand that approach now and I'm ok with that :)

frontend/javascripts/admin/api/jobs.ts
philippotto: e.g., "The cost is encoded as a string decimal for precision reasons. The front-end should not do any arithmetics with this."

frontend/javascripts/admin/organization/upgrade_plan_modal.tsx
```diff
@@ -308,16 +309,17 @@ const CREDIT_COST_USD = 5.75;
 
 function OrderWebknossosCreditsModal({ destroy }: { destroy: () => void }) {
   const userInputRef = useRef<HTMLInputElement | null>(null);
-  const [costsAsString, setCostsAsString] = useState<string>("5€/5.75$");
-  const recalculateCosts = (newCosts: number | null) => {
+  const [creditCostAsString, setCreditCostsAsString] = useState<string>("5€/5.75$");
+  const [creditAmount, setCreditAmount] = useState<number | null>(1);
+  useEffect(() => {
     // TODOM: Remove magic numbers. E.g. put them into application configuration.
```
philippotto, line 315: todo

frontend/javascripts/admin/organization/upgrade_plan_modal.tsx
```diff
@@ -341,7 +343,7 @@ function OrderWebknossosCreditsModal({ destroy }: { destroy: () => void }) {
       <div className="drawing-upgrade-users">
         <p style={{ marginRight: "30%" }}>
           You can buy new WEBKNOSSOS credits to pay for premium jobs and services. Each credit costs
-          5€ or 5.75$.
+          {formatCurrency(5, "€")} or {formatCurrency(5.75, "$")}.
```
philippotto, line 346: also needs extraction probably

app/models/organization/CreditTransactionService.scala
```diff
@@ -68,6 +69,12 @@ class CreditTransactionService @Inject()(creditTransactionDAO: CreditTransaction
       implicit ctx: DBAccessContext): Fox[Unit] =
     creditTransactionDAO.addJobIdToTransaction(creditTransaction, jobId)
 
+  def findTransactionOfJob(jobId: ObjectId)(implicit ctx: DBAccessContext): Fox[Option[CreditTransaction]] =
+    creditTransactionDAO.findTransactionForJob(jobId).futureBox.flatMap {
+      case Full(transaction) => Fox.successful(Some(transaction))
+      case _                 => Fox.successful(None)
+    }
```
fm3, line 76: if you replace flatMap by map, you can leave out the Fox.successful.
I feel like we have a box2Option conversion somewhere that does pretty much this. Maybe you could go on a search, or implement it if it is missing. Maybe box even has a toOption?

fm3 (commented):
Apart from the previously unresolved github conversations, the backend looks good to me :)

https://github.com/scalableminds/webknossos/pull/8352#discussion_r1982026881
https://github.com/scalableminds/webknossos/pull/8352#discussion_r1980837715
https://github.com/scalableminds/webknossos/pull/8352#discussion_r1980842276
https://github.com/scalableminds/webknossos/pull/8352#discussion_r1980891254

frontend/javascripts/admin/organization/upgrade_plan_modal.tsx
MichaelBuessemeyer: ok should be done now

frontend/javascripts/admin/organization/upgrade_plan_modal.tsx
```diff
@@ -296,9 +297,82 @@ function UpgradePricingPlanModal({
...
+    const totalCostInEuro = newCosts * CREDIT_COST_EUR;
+    const totalCostInDollar = newCosts * CREDIT_COST_USD;
+    setCostsAsString(`${totalCostInEuro}€/${totalCostInDollar}$`);
+  };
+
+  const handleOrderCredits = async () => {
+    if (userInputRef.current) {
+      const requestedUsers = Number.parseInt(userInputRef.current.value);
+      await sendOrderCreditsEmail(requestedUsers);
+      Toast.success(messages["organization.credit_request_sent"]);
```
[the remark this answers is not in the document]
MichaelBuessemeyer: A quiet error is thrown. I added some error handling -> showing a toast in case of an error.

frontend/javascripts/oxalis/view/action-bar/starting_job_modals.tsx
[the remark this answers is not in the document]
MichaelBuessemeyer: I added a try and catch with a toast informing about the error but including that this should impact the user. What do you think?

MichaelBuessemeyer (commented):
Ok newest frontend feedback should be implemented

frontend/javascripts/admin/organization/upgrade_plan_modal.tsx
MichaelBuessemeyer: I am a little unsure what you meant here with extraction. Using the credits from the features?

tools/postgres/schema.sql
MichaelBuessemeyer: Yeah, I corrected it now to `Decimal(14,3)`. Same for the `CreditTransaction.scala`

MichaelBuessemeyer (commented):
Thanks for reminding me of the unresolved thread. I now worked an the following thread (which are the once I found)

  - https://github.com/scalableminds/webknossos/pull/8352#discussion_r1967253231
  - https://github.com/scalableminds/webknossos/pull/8352#discussion_r1980873992
  - https://github.com/scalableminds/webknossos/pull/8352#discussion_r1967275325
  - https://github.com/scalableminds/webknossos/pull/8352#discussion_r1967312486
  - https://github.com/scalableminds/webknossos/pull/8352#discussion_r1967322453
  - https://github.com/scalableminds/webknossos/pull/8352#discussion_r1967350617
  - https://github.com/scalableminds/webknossos/pull/8352#discussion_r1967352997
  - https://github.com/scalableminds/webknossos/pull/8352#discussion_r1967371321
  - https://github.com/scalableminds/webknossos/pull/8352#discussion_r1967372421
  - https://github.com/scalableminds/webknossos/pull/8352#discussion_r1981020887
  - https://github.com/scalableminds/webknossos/pull/8352#discussion_r2001683767

Is there any thread missing here? Could you please check the latest changes @fm3 :pray:

app/controllers/JobController.scala
```diff
@@ -283,9 +292,9 @@ class JobController @Inject()(jobDAO: JobDAO,
           _ <- datasetService.assertValidDatasetName(newDatasetName)
           _ <- datasetService.assertValidLayerNameLax(layerName)
           multiUser <- multiUserDAO.findOne(request.identity._multiUser)
-          _ <- bool2Fox(multiUser.isSuperUser) ?~> "job.inferMitochondria.notAllowed.onlySuperUsers"
-          _ <- Fox.runIf(!multiUser.isSuperUser)(jobService.assertBoundingBoxLimits(bbox, None))
           command = JobCommand.infer_mitochondria
+          parsedBoundingBox <- BoundingBox.fromLiteral(bbox).toFox
+          _ <- Fox.runIf(!multiUser.isSuperUser)(jobService.assertBoundingBoxLimits(bbox, None))
```
MichaelBuessemeyer, line 297: Should this limit be kept or is the credit limit enough? IMO the credit limit should be enough, correct?

frontend/javascripts/admin/organization/upgrade_plan_modal.tsx
```diff
@@ -296,9 +299,90 @@ function UpgradePricingPlanModal({
...
+          />
+        </div>
+        Total resulting cost: {creditCostAsString}
+        <>
+          <Divider style={{ marginTop: 40 }} />
+          <p style={{ color: "#aaa", fontSize: 12 }}>
+            Ordering WEBKNOSSOS credits for your organization will send an email to the WEBKNOSSOS
+            sales team. They will top up your account with the requested amount of credits and send
+            you an invoice with the required cost. We typically respond within one business day. See
+            our <a href="https://webknossos.org/faq">FAQ</a> for more information.
```
normanrz, line 374:
```suggestion
            Ordering WEBKNOSSOS credits for your organization will send an email to the WEBKNOSSOS
            sales team. We typically respond within one business day to discuss payment options and purchasing requirements. See
            our <a href="https://webknossos.org/faq">FAQ</a> for more information.
```

app/controllers/JobController.scala
MichaelBuessemeyer: should be kept as as credits won't be used to enforce max size for now

MichaelBuessemeyer:
As discussed in https://scm.slack.com/archives/C5AKLAV0B/p1742390562205399 we will first put the credit system into a testing phase. During this phase jobs will not cost credits but their costs will be shown to the user. After some testing time we want to end the testing phase making the jobs cost actual credit. I wrote issue https://github.com/scalableminds/webknossos/issues/8458 to track this

app/models/organization/CreditTransaction.scala
```diff
@@ -0,0 +1,362 @@
...
+      creditChange,
+      comment,
+      transactionState,
+      creditState,
+      expiresAt,
+      createdAt,
+      updatedAt,
+      isDeleted
+    )
+  }
```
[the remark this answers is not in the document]
MichaelBuessemeyer: Oh, very good catch!!!

conf/evolutions/129-credit-transactions.sql
```diff
@@ -0,0 +1,113 @@
...
+CREATE FUNCTION webknossos.enforce_non_negative_balance() RETURNS TRIGGER AS $$
+  BEGIN
+    -- Assert that the new balance is non-negative
+    ASSERT (SELECT COALESCE(SUM(credit_delta), 0) + COALESCE(NEW.credit_delta, 0)
+            FROM webknossos.credit_transactions
+            WHERE _organization = NEW._organization AND _id != NEW._id) >= 0, 'Transaction would result in a negative credit balance for organization %', NEW._organization;
+    -- Assertion passed, transaction can go ahead
+    RETURN NEW;
+  END;
+$$ LANGUAGE plpgsql;
```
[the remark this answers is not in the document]
MichaelBuessemeyer: @fm3 Do you think I should switch to raising an exception due to the reasons above? I chose Assert here as we are using this in the evolution scripts. But the arguments above sound very reasonable

philippotto:
> As in the current version jobs do not spent credits (see [#8352 (comment)](https://github.com/scalableminds/webknossos/pull/8352#issuecomment-2736992444)) should I also disable / hide the modal to order new credits?

yes, makes sense to me :+1:

frontend/javascripts/admin/organization/upgrade_plan_modal.tsx
philippotto: I still see the "TODOM" comment in the pr. maybe something is weird with my diff view, though?

frontend/javascripts/features.ts
```diff
@@ -1,13 +1,14 @@
...
 
 export async function load() {
   features = await getFeatureToggles();
   return features;
 }
 export function __setFeatures(_features: Record<string, any>) {
-  // Only use this function for tests.
-  features = _features;
+  // Only use this function for tests. Unsafe typecase is intended here to have improved type safety.
+  features = _features as APIFeatureToggles;
```
philippotto, line 11: :+1:

frontend/javascripts/oxalis/view/action-bar/starting_job_modals.tsx
```diff
@@ -611,9 +613,10 @@ function JobCreditCostInformation({
                   if (jobCreditCostPerGVx != null && currentBoundingBoxVolume == null) {
                     return Promise.reject("This is a requires a selected bounding box.");
                   } else if (jobCreditCostPerGVx != null && !orgaHasEnoughCredits) {
+                    /* TODO: Uncomment after credit system testing phase.
```
philippotto, line 616: reminder for this todo (todom?)

frontend/javascripts/oxalis/view/action-bar/starting_job_modals.tsx
```diff
@@ -857,8 +860,15 @@ function StartJobForm(props: StartJobFormProps) {
...
       if (jobCreditCostPerGVx != null && activeUser?.organization) {
         // As the job did cost credits, refetch the organization to have a correct credit balance.
-        const updatedOrganization = await getOrganization(activeUser?.organization);
-        dispatch(setActiveOrganizationAction(updatedOrganization));
+        try {
+          const updatedOrganization = await getOrganization(activeUser?.organization);
+          dispatch(setActiveOrganizationAction(updatedOrganization));
+        } catch (error) {
+          Toast.error(
+            "Failed to fetch updated amount of credits of your organization. See the console for more details. This usually shouldn't impact you.",
```
philippotto, line 868: this is too technical in my opinion. how about "There was an error while reloading the available credits. Consider reloading the page."

philippotto (commented):
the latest FE changes look good :) only left some small nits.

frontend/javascripts/oxalis/view/action-bar/starting_job_modals.tsx
MichaelBuessemeyer: No this should be kept IMO to make it easier to make the credit system acutally requiring payment easier. Please see https://github.com/scalableminds/webknossos/pull/8352#issuecomment-2742866271 and the linked comment & slack thread (linked in th linked comment)

MichaelBuessemeyer: I you want to avoid this uncommented code, I could also delete it and re-add it in the pr fully enabling the credit system

frontend/javascripts/oxalis/view/action-bar/starting_job_modals.tsx
MichaelBuessemeyer: Sure, sounds good. Thanks for the suggestion :pray:

frontend/javascripts/admin/organization/upgrade_plan_modal.tsx
MichaelBuessemeyer: Oh right, I forgot to delete this comment. Thanks for mentioning :+1:

MichaelBuessemeyer (commented):
Thanks for the input. Should be done no except for the commented out lines. Please give my comment a quick look :eye:

frontend/javascripts/oxalis/view/action-bar/starting_job_modals.tsx
philippotto: ah, I see, I thought the review process is the testing phase, but this means something different. then, nevermind :)

conf/evolutions/129-credit-transactions.sql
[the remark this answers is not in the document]
fm3: yes sounds good.

fm3 (approved):
Backend LGTM. I’d say @normanrz should probably give this a testing round and check if all the wording makes sense and if the behavior now matches what we currently need.

frontend/javascripts/admin/organization/upgrade_plan_modal.tsx
```diff
@@ -296,9 +299,89 @@ function UpgradePricingPlanModal({
...
+            Ordering WEBKNOSSOS credits for your organization will send an email to the WEBKNOSSOS
+            sales team. We typically respond within one business day to discuss payment options and
+            purchasing requirements. See our <a href="https://webknossos.org/faq">FAQ</a> for more
+            information.
+          </p>
+        </>
+      </div>
+    </Modal>
+  );
+}
```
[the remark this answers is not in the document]
MichaelBuessemeyer: Thanks for the hint :+1:

PR #8354, opened 2025-01-24
Add privacy note to dataset upload

### LINKED ISSUE #8317: Add note that uploaded data is private
## Detailed Description
I had conversations with a number of users that were assuming that if you upload data to Webknossos, it is public immediately. That is, of course, not true so we should educate users about that. Maybe we should put a note in the upload screen? Maybe also somewhere on the website. I think the docs are already clear about this.

knollengewaechs:
- https://adddataprivacynote.webknossos.xyz/

See feedback to first draft here: https://scm.slack.com/archives/C5AKLAV0B/p1737750240433629

### Steps to test:
- In the dashboard, click "Add dataset" and view the dataset upload view
- Check that is looks nice :)
- Also check the "Add remote dataset" tab and the "Compose from existing datasets" tab
    - the first one also contains the new note, the latter does not, but it should still be aligned nicely

### TODOs:
- [ ] ...

### Issues:
- fixes #8317

------


frontend/javascripts/admin/dataset/dataset_components.tsx
```diff
@@ -12,10 +12,12 @@ export function CardContainer({
   children,
   withoutCard,
   title,
+  subtitle,
 }: {
   children: React.ReactNode;
   withoutCard?: boolean;
   title: string;
+  subtitle?: JSX.Element;
```
MichaelBuessemeyer, line 20: I would prefer to use `React.ReactNode` over `JSX.Element` as `React.ReactNode` is more generic: From the react docs (according what the vscode hint showed me)
> Represents all of the things React can render.
> Where ReactElement only represents JSX, ReactNode represents everything that can be rendered.


```suggestion
  subtitle?: React.ReactNode;
```

MichaelBuessemeyer (approved):
Very nice :tada: 

Thanks a lot for adding this info to the upload / ... views. In only found a minor nitpick.

Moreover, I think it would be nice to have the info also in the Card of the `Compose a dataset from existing dataset layers` view. So that it is present in all three tabs and not only in two of them. Because now as the info is missing in the third tab, it feels like when I compose the dataset, the default private rule does not apply. But afaik it does. So I think it would be nice to have it added there as well :)

Besides this everything is :top: 

Moreover, I think no re-review is needed so here is a :green_circle:

knollengewaechs:
> So that it is present in all three tabs and not only in two of them. Because now as the info is missing in the third tab, it feels like when I compose the dataset, the default private rule does not apply. But afaik it does. So I think it would be nice to have it added there as well :)

good call! I omitted is there because if I understood correctly, no data is imported during DS composition. but youre right, it might make users insecure if the privacy rules apply here.

PR #8355, opened 2025-01-27
Readd fixed version of "Allow permanent dataset layer rotation in dataset settings""

MichaelBuessemeyer:
Reverts scalableminds/webknossos#8353
And re adds: https://github.com/scalableminds/webknossos/pull/8159

Related Slack canvas: https://scm.slack.com/docs/T02A8MN9K/F083DBRCFC6

 --> This PR adds new settings to the dataset settings for each layer. An interface (still in design process) enables the user to set a affine transformation matrix by letting the user define angles for each axis around which the dataset should be rotated.

- https://___.webknossos.xyz

### Steps to test:
- Basic testing:
  - View your favourite DS 
  - open the DS settings in another tab. Use the rotation slider to configure some rotation. e.g. z 90 degrees and save
  - reload the view DS tab. The DS should now be fully rotated by 90 degree around the z axis -> xy viewport should be rotated by 90 degrees
  - Play around with toggeling the rotation off and on in the layer setting in the view mode
- Testing annotation support
  - create a hybrid annotation of a rotated DS
  -  add an empty volume layer
  - turn of the transforms
  - toggling the transforms should not work on the skeleton layer and the volume layer without fallback. These do not have an underlying layer that might have transforms configured. Therefore, `nativelyRenderedLayerName` cannot be set to one of these layers. Moreover, the transforms cannot be toggled on on such layers as they do not have transformations.
  - toggle the transforms on other layer and have fun
  - toggle off the transforms, do volume annotations on each layer and create some skeleton nodes
  - toggle on the transforms again. The nodes and volume annotation should rotate accordingly. This also applies to the volume layer without fallback data.
  - This should hopefully cover all wanted annotation scenarios. If you can think up more, please also test them :pray: 
- In general: When toggling transformations, the camera position should update accordingly.
- Check conflicting `nativelyRenderedLayerName` settings.
  - Open a hybrid annotation on a dataset with a rotation setting that has at least one segmentation layer.
  - On the editable segmentation layer (volume layer with fallback) use the "transform button" to render the volume layer natively
  - In another tab open another such hybrid tracing. The DatasetViewConfigs should now have `nativelyRenderedLayerName` set to the tracingId of the volume layer of the first annotation. This does not exist in the newly opened other annotation. => The annotation should still open up (but with all layers transformed), as the `nativelyRenderedLayerName` setting should be automatically disregarded as the layer does not exist in this annotation.
- !!! important !!!
  - Please also test whether a DS with multi modality still works as expected :pray: 

### TODOs / Problems:
- ~~[ ] Design a visualization of the result. For some first ideas look at the issue [#7334](https://github.com/scalableminds/webknossos/issues/7334)~~ --> deferred / not planned
- [x] The rotation should be done around the center of the dataset. For this, a translation by the dataset center is needed
- [x] The interface works by defining Euler angles. The problem is that Euler angles are not deterministic / there are multiple ways to express a certain rotation with different Euler angle values. See: https://stackoverflow.com/questions/32235395/does-euler-angles-quaternion-euler-angles-always-result-in-an-equivalent-r
  - This makes it hard to update the settings correctly ->  a 180 degree rotation around y is equivalent to 180 on x and 180 on z axis :/
  - A solution for this is needed as the interface currently is pretty quirky.
- [x] Reset `nativelyRenderedLayerName` in model init if this layer does not exist. (Might happen when switching annotations)

### Problems:
- [x] Layer rotates out of the dataset bbox 

### Issues:
- fixes [#7334](https://github.com/scalableminds/webknossos/issues/7334)

------


frontend/javascripts/dashboard/dataset/dataset_rotation_form_item.tsx
```diff
@@ -151,9 +151,12 @@ export type DatasetRotation = {
...
 }: AxisRotationSettingForDatasetProps) => {
-  const dataLayers: APIDataLayer[] = form?.getFieldValue(["dataSource", "dataLayers"]);
-  const isRotationOnly = useMemo(() => doAllLayersHaveTheSameRotation(dataLayers), [dataLayers]);
-
+  // form -> dataSource -> dataLayers can be undefined in case of the add remote dataset form which is initially empty.
+  const dataLayers: APIDataLayer[] | undefined = form?.getFieldValue(["dataSource", "dataLayers"]);
+  const isRotationOnly = useMemo(
+    () => (dataLayers ? doAllLayersHaveTheSameRotation(dataLayers) : false),
+    [dataLayers],
+  );
```
MichaelBuessemeyer, line 159: This should be the acutal fix

frontend/javascripts/oxalis/model/accessors/dataset_layer_transformation_accessor.ts
```diff
@@ -402,7 +402,11 @@ function hasValidTransformationPattern(transformations: CoordinateTransformation
 }
 
 function _doAllLayersHaveTheSameRotation(dataLayers: Array<APIDataLayer>): boolean {
-  const firstDataLayerTransformations = dataLayers[0]?.coordinateTransformations;
+  if (dataLayers.length === 0) {
+    // The dataset does not have any layers. Therefore no layers can be rotated.
+    return false;
+  }
+  const firstDataLayerTransformations = dataLayers[0].coordinateTransformations;
```
MichaelBuessemeyer, line 409: And here I noticed while thinking about this bug, that the previous code would return true for `_doAllLayersHaveTheSameRotation` when it was called with an empty array of data layers. This seemed wrong to me on a conceptual level as one cannot rotate something that has no data. Feel free to disagree :)

MichaelBuessemeyer (commented):
This PR should readd and fix the dataset rotation feature.
Please only check commit 3e6a7804645374685218fac7c69496644f127657. This should include the bug fix

philippotto (approved):
great, thank you!

PR #8356, opened 2025-01-27
Avoid onboarding screen if /api/organizationsIsEmpty fails

### LINKED ISSUE #8056: Avoid Onboarding screen when /api/organizationsIsEmpty fails
In the past, we had at least two occurrences where a WK production system showed the onboarding system, because it considered the DB to be empty. No data was lost, but nevertheless the screen is scary.

We should do the following two things:
- [ ] avoid that the onboarding screen appears in such scenarios
- [x] double-check that the onboarding screen cannot delete data when somebody uses it to create a new organization

Some notes on when this happened:
- we updated a wk instance and used a docker tag that wasn't pushed. kubernetes didn't find the image and kept the old one deployed.
- the db migration was executed (--> the DB schema was too new for the wk version)
- the onboarding screen appeared instead of the dashboard

knollengewaechs:
- https://___.webknossos.xyz

### Steps to test:
- abc

### TODOs:
- [ ] ...

### Issues:
- fixes #8056 

------

### DIFF
--- a/frontend/javascripts/main.tsx
+++ b/frontend/javascripts/main.tsx
@@ -2,7 +2,7 @@ import { message } from "antd";
 import window, { document } from "libs/window";
 import rootSaga from "oxalis/model/sagas/root_saga";
 import UnthrottledStore, { startSagas } from "oxalis/store";
-import { createRoot } from "react-dom/client";
+import { createRoot, ErrorInfo } from "react-dom/client";
 import { Provider } from "react-redux";
 
 import { createSyncStoragePersister } from "@tanstack/query-sync-storage-persister";
@@ -13,7 +13,7 @@ import ErrorBoundary from "components/error_boundary";
 import { RootForFastTooltips } from "components/fast_tooltip";
 import { load as loadFeatureToggles } from "features";
 import checkBrowserFeatures from "libs/browser_feature_check";
-import ErrorHandling from "libs/error_handling";
+import ErrorHandling, { handleGenericError } from "libs/error_handling";
 import UserLocalStorage from "libs/user_local_storage";
 import { compress, decompress } from "lz-string";
 import { setupApi } from "oxalis/api/internal_api";
@@ -29,6 +29,7 @@ import Router from "router";
 
 import "../stylesheets/main.less";
 import GlobalThemeProvider, { getThemeFromUser } from "theme";
+import { escalateErrorAction } from "oxalis/model/actions/actions";
 
 // Suppress warning emitted by Olvy because it tries to eagerly initialize
 window.OlvyConfig = null;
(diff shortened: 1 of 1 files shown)


knollengewaechs:
@philippotto So this is my initial suggestion. I am a little insecure whether I tested it enough and all the way, because I made changes in such a meaningful component. I am looking forward to hear what you think!

frontend/javascripts/main.tsx
```diff
@@ -97,10 +93,24 @@ document.addEventListener("DOMContentLoaded", async () => {
   });
   message.config({ top: 30 });
   checkBrowserFeatures();
-  await Promise.all([loadFeatureToggles(), loadActiveUser(), loadHasOrganizations()]);
-  await Promise.all([loadOrganization()]);
   const containerElement = document.getElementById("main-container");
 
+  try {
+    await Promise.all([loadFeatureToggles(), loadActiveUser(), loadHasOrganizations()]);
+    await Promise.all([loadOrganization()]);
```
philippotto, line 100: there are now 4 function calls in this new try block. for `loadHasOrganizations`, you already did the testing. but what about the others? I'm not sure what their failure would cause on master (for example, what happens if `loadOrganization` fails). maybe the new error makes sense, but maybe it's too strict :thinking: could you test these scenarios, too?

knollengewaechs: you are right that I did not test this enough!
If `loadOrganizations` fails, the same failure screen is rendered as when `loadHasOrganizations` fails. This is probably too strict. 
An obvious solution would be to move all other api calls outside the try and catch block. the disadvantage would be that one more single `await` is introduced where it is not needed (within the try-catch). But thats probably the trade-off that has to be made, I dont have a simple solution where is not the case :sweat_smile:

knollengewaechs: I noticed that if `loadFeatureToggles` yields an error, a blank screen is rendered, so I included it in the try-catch.

philippotto: > Oh, I understand that this depends on the test scenario! During login, I got an `Action Not Found` toast and I thought that was better than the error screen. But if the user is logged in and suddenly the request fails, the blank screen is rendered.

could it be that you tested the "open wk & log in" scenario and I tested the "open wk while being logged in"? I also simply added `throw Error()` to the `loadOrganization` function (not sure whether that makes a difference).

knollengewaechs: >could it be that you tested the "open wk & log in" scenario and I tested the "open wk while being logged in"?

yes, I think thats right :sweat_smile:

frontend/javascripts/main.tsx
```diff
@@ -97,10 +93,32 @@ document.addEventListener("DOMContentLoaded", async () => {
...
+      // with hasOrganizations==true.
+      loadHasOrganizations(),
+    ]);
+    await loadOrganization();
+  } catch (e) {
+    console.error("Failed to load WEBKNOSSOS due to the following error", e);
+    if (containerElement) {
+      const react_root = createRoot(containerElement);
+      react_root.render(
+        <p style={{ margin: "20px" }}>
```
philippotto, line 114:
```suggestion
        <p style={{ margin: 20, marginTop: -20}}>
```

philippotto (commented):
great, testing went well :+1: please see my one style suggestion.

philippotto (approved):

PR #8359, opened 2025-01-28
Fix merging trees

MichaelBuessemeyer:
The merging trees code accidentally dropped all trees of the other type (which are not merged). The newest version of the code makes sure to not drop the trees of the other (not active) type. Moreover, I improved the test such that both types (default & agglomerate) are not tested in the reducer tests. I added an additional that tests that trees of different types are not merged together.

- https://___.webknossos.xyz

### Steps to test:
- Create an annotation on a dataset with an agglomerate mapping
- create some trees
- activate the agglomerate mapping
- load some agglomerate skeletons 
- merge some agglomerate skeletons (e.g. via context menu)
- the initially created trees should not disappear

### TODOs:
- [x] Test whether the changes to the tests would have caught this problem.
  - Yes they would, I tested

### Issues:
- fixes bugreport: https://scm.slack.com/archives/C02H5T8Q08P/p1738074705411519

------


MichaelBuessemeyer:
already requesting review

but I don't have anymore time to add the changelog entry and check whether the tests actually improved / would have shown this bug.

philippotto (approved):
thanks for fixing this!

PR #8362, opened 2025-01-29
Fix serializing DeleteSegmentDataUpdateAction

fm3:
Needs to be serialized now in order to write it to FossilDB

Regression from #7917


daniel-wer (approved):
Thanks for the quick fix! Is it possible to check the case list for exhaustiveness to prevent this kind of error?

fm3:
I’ll have a look. I think this might be possible. I wrote https://github.com/scalableminds/webknossos/issues/8363 to track it

PR #8364, opened 2025-01-29
Fix segment registering in coarser mags for non-mag-aligned bounding boxes

daniel-wer:
When registering segments for a bounding box in a mag other than the finest one and if the bounding box is not mag-aligned it was not taken into account that the `getDataForBoundingBox` might grow the bounding box to be mag-aligned. In that case more data is returned and the indexing may change, resulting in incorrect segment positions.

- https://___.webknossos.xyz

### Steps to test:
- Create an annotation for l4_v2_sample or another dataset which has mag 16-16-8, choose an empty volume layer and restrict the volume annotation to mag 16-16-8
- Create a bounding box with topleft and size 1119, 1111, 1103, 640, 653, 903 (this was chosen to be "maximally mag-unaligned" to provoke the bounding box growing)
- Make a single brush stroke at 1680, 1632, 1920
- Register Segments for the bounding box
- Switch to the segments tab and click on the segment - wk should jump to the segment (on master wk jumps somewhere else)

------


frontend/javascripts/oxalis/api/api_latest.ts
```diff
@@ -702,12 +704,15 @@ class TracingApi {
       existingMagIndex,
     );
     const [dx, dy, dz] = currentMag;
-
+    // getDataForBoundingBox grows the bounding box to be mag-aligned which can change the dimensions
+    const boundingBoxInMag1MagAligned = boundingBoxInMag1.alignWithMag(currentMag, "grow");
```
knollengewaechs, line 708: Just for my understanding, where can I find the code in which `getDataForBoundingBox` might be growing the bbox?

daniel-wer: It's quite hidden because the code is complex, which is unfortunate. `getDataForBoundingBox` requests all buckets that intersect with the mag 1 bounding box (`getBucketAddressesInCuboid`) so the gathered data usually is larger than the mag 1 bounding box, because the bounding box is not aligned to 32³ buckets. Then, `cutOutCuboid` is called which converts the mag 1 bounding box to the request mag by using the `scaleGlobalPositionWithMagnification` method. This method has a `ceil` argument which is not set for top left (so that is floored) and set for bottom right. This possibly grows the bounding box (when thinking about it in mag 1 coordinates, which is what is done to determine the segment positions).
Does that make sense?

knollengewaechs: Okay, that makes sense! I understand that both in `cutOutCuboid`->`scaleGlobalPositionWithMagnification` the topLeft is floored and the bottomRight is ceiled, which is the same as what is happening in `alignWithMag` when it's called with the argument `"grow"`.

knollengewaechs (approved):
Thank you for fixing this! :pray: 
Testing went well and the code makes sense!

PR #8377, opened 2025-02-05
Add button to set failed job to pending

### LINKED ISSUE #8370: Add button to processing jobs page to resume failed jobs for superusers.
## Detailed Description
When a processing job fails, a button for superusers to resume them after the issue has been fixed would be quite handy.

## Context
* [x] Specific to long-running jobs (set `jobsEnabled=true` in `application.conf`)
* [x] Specific to webknossos.org (set `isDemoInstance=true` in `application.conf`)

frcroth:
- https://___.webknossos.xyz

### Steps to test:
- Enable jobs
- Set a job to "FAILURE"
- Resume the job. The issue did not specify what should happen to the job so I assumed "PENDING". Should it be "RUNNING" instead?

### TODOs:
- [x] @ frontend (@knollengewaechs ?) Can you please configure the button to only show for super users? I don't know how to do that.

### Issues:
- fixes #8370 

------


daniel-wer:
Thanks for implementing! When a superuser retries/resumes a task by another user, ideally the user wouldn't get the "Job Failed" email if the job fails another time, because that is irritating and unnecessary email noise for him/her. Once the job is successful, the respective mail should go through. Could the two of you coordinate on how to best implement this as part of this PR or a followup? /cc @normanrz

MichaelBuessemeyer:
> Thanks for implementing! When a superuser retries/resumes a task by another user, ideally the user wouldn't get the "Job Failed" email if the job fails another time, because that is irritating and unnecessary email noise for him/her. Once the job is successful, the respective mail should go through. Could the two of you coordinate on how to best implement this as part of this PR or a followup?


Oh that's a good idea :banana: :dancers:.
First thoughts: Add a retry count to the jobs table and only send failed mail in case the retry count is 0 upon failure. Save some bool like `sentFailedMessage` to the job. And also maybe other ideas. But this sounds definitely doable and not so complicated. => I'd vote for implementing this in this pr :).

But maybe I am also overseeing things. 

Moreover: I think I like the approach saving the retry count more, as it provides more information for e.g. later analysis on how many attempts were needed instead of saving a simple boolean which provides less information. :thinking:

daniel-wer:
These are good ideas, but I think the differentiating factor should rather be whether another user resumed the workflow run. From a users perspective, when resuming my own workflow, I'd like to be notified of failure and success. However, if a super user resumed a run, the run owner should only be notified of success.

app/controllers/JobController.scala
```diff
@@ -112,6 +112,16 @@ class JobController @Inject()(
     } yield Ok(js)
   }
 
+  def resume(id: ObjectId): Action[AnyContent] = sil.SecuredAction.async { implicit request =>
+    for {
+      _ <- bool2Fox(wkconf.Features.jobsEnabled) ?~> "job.disabled"
+      _ <- userService.assertIsSuperUser(request.identity) ?~> "notAllowed" ~> FORBIDDEN
+      job <- jobDAO.findOne(id)
+      _ <- jobDAO.updateManualState(id, JobState.PENDING)
```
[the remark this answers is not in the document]
frcroth: yes, I think the problem was that the field manual state was used, which only has an effect on the displayed state.

frcroth:
@daniel-wer @MichaelBuessemeyer I added the column "resumedBySuperUser" to the jobs table, which is checked when sending a failure email.

conf/evolutions/126-job-resumed-by-super-user.sql
```diff
@@ -0,0 +1,12 @@
+START TRANSACTION;
+
+do $$ begin ASSERT (select schemaVersion from webknossos.releaseInformation) = 125, 'Previous schema version mismatch'; end; $$ LANGUAGE plpgsql;
+
+DROP VIEW webknossos.jobs_;
+
+ALTER TABLE webknossos.jobs ADD column resumedBySuperUser boolean NOT NULL DEFAULT false;
```
MichaelBuessemeyer, line 7: I know, very nitpicky but `boolean` should be in caps here as well, correct?

```suggestion
ALTER TABLE webknossos.jobs ADD column resumedBySuperUser BOOLEAN NOT NULL DEFAULT false;
```

conf/application.conf
```diff
@@ -151,8 +151,8 @@ features {
   taskReopenAllowedInSeconds = 30
   allowDeleteDatasets = true
   # to enable jobs for local development, use "yarn enable-jobs" to also activate it in the database
-  jobsEnabled = false
-  voxelyticsEnabled = false
+  jobsEnabled = true
+  voxelyticsEnabled = true
```
MichaelBuessemeyer, line 155: undo me -> should also make the ci green

```suggestion
  jobsEnabled = false
  voxelyticsEnabled = false
```

frcroth: oops

frontend/javascripts/admin/job/job_list_view.tsx
```diff
@@ -293,11 +297,23 @@ function JobListView() {
...
+    } else if (job.state === "FAILURE" && isCurrentUserSuperUser) {
+      return (
+        <AsyncLink
+          href="#"
+          onClick={async () => {
+            resumeJob(job.id).then(() => fetchData());
+          }}
+          icon={<PlayCircleOutlined className="icon-margin-right" />}
+        >
+          Resume
```
MichaelBuessemeyer, line 314: maybe restart/retry is a better word here :thinking:?


```suggestion
          Retry
```

MichaelBuessemeyer (approved):
Looks good and works :tada: 

Thanks for implementing this :firecracker: 

Maybe consider my final potential improvement. IMO the button restarts / retries a run and not resumes it :thinking:

frontend/javascripts/admin/job/job_list_view.tsx
frcroth: I see your reasoning, I am just following the specification. @valentin-pinkau @daniel-wer  Opinion on this?

daniel-wer: It's a bit unclear, because it's actually both. Resume, because the tasks that already finished are skipped and retry, because the failed task is retried. Still, `Retry` is probably more accurate. Since it's only visible for super user that's fine I'd say. Otherwise one could add a tooltip, noting `Restarts the workflow from the failed task, skipping and reusing artifacts from preceding tasks that were already successful.`

MichaelBuessemeyer: > Restarts the workflow from the failed task, skipping and reusing artifacts from preceding tasks that were already successful.

That sounds like a valuable addon regardless of this being only available to super users. As we might make this available to non super users in the future, we do not need to remember to add this tooltip. => I'd vote to add it here as this shouldn't be difficult and also should need a re-review thus not delaying merging this :)

frontend/javascripts/admin/job/job_list_view.tsx
```diff
@@ -293,11 +297,24 @@ function JobListView() {
...
         </AsyncLink>
       );
+    } else if (job.state === "FAILURE" && isCurrentUserSuperUser) {
+      return (
+        <AsyncLink
+          href="#"
+          title="Restarts the workflow from the failed task, skipping and reusing artifacts from preceding tasks that were already successful."
+          onClick={async () => {
+            resumeJob(job.id).then(() => fetchData());
+          }}
```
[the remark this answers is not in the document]
MichaelBuessemeyer: Yeah the AI is correct here, I'd say. Some kind of feedback would be useful. Maybe add an error toast upon failure and a success toast upon success?

Maybe something like this:
```javascript
 onClick={async () => {
  try{
    await retryJob(job.id)
    await fetchData();
    Toast.success("Job is being retried");
    } catch (e) {
      console.error("Could not retry job", e);
      Toast.error("Failed to start retrying the job");
    }
  }}
```

MichaelBuessemeyer (commented):
Still works, nice :tada:

Please seem my two comments (one is attached to an already existing thread).

After that I should be good to go :runner:

MichaelBuessemeyer:
Awesome looks great :tada: 

Did you do retesting a little or should I do this quickly? The changes are so minor that one person giving it a quick test should be enough IMO :)

daniel-wer (approved):
It would be great if we could get this merged. I just tested the functionality and it worked nicely. Is there anything else holding this back?

MichaelBuessemeyer (approved):
Yeah I think as well that this would be very helpful now.

I'd give it a go :goat:

PR #8380, opened 2025-02-05
include error chain when logging bucket loading errors

### LINKED ISSUE #8379: Unify formatChain methods
We have redundant implementations of formatChain in our scala backend code (for error logging / readable error messages from Box/Fox objects).

These should be unified, possibly live in a trait in util.

fm3:
Log the whole error chain when bucket loading fails.

Motivated by https://scm.slack.com/archives/C02H5T8Q08P/p1738766627989389?thread_ts=1738766454.436339&cid=C02H5T8Q08P

Also fixes #8379


conf/application.conf
```diff
@@ -161,7 +161,7 @@ features {
   exportTiffMaxVolumeMVx = 1024
   exportTiffMaxEdgeLengthVx = 8192
   defaultToLegacyBindings = false
-  editableMappingsEnabled = false
+  editableMappingsEnabled = true
```
fm3, line 164: unrelated to this pr. I talked to @normanrz and we agreed to include this here. will be disabled selectively in docker.

PR #8382, opened 2025-02-05
Fix flood fill tool for coarse mags

### LINKED ISSUE #8366: Flood fill bounding box size maximum should take into account available mags
## Context

I tried to use the restricted flood fill tool in 3D for a segmentation layer which is restricted to mag 16, in mag 16, for a bounding box of size 1280³ vx.

## Expected Behavior

The flood fill should work, since the finest available mag is 16 and the flood fill tool is used in mag 16. The bounding box size in mag 16 is 512,000 vx, so way below the maximum.

## Current Behavior

There is an error toast: "The bounding box that encloses the clicked position is too large. Shrink its size so that it does not contain more than 8847360 voxels."

## Steps to Reproduce the bug

1. Create an annotation with the volume layer restricted to mag 16.
2. Create a bounding box of size 1280³ vx and perform a brush stroke inside.
3. Zoom out to mag 16.
4. Activate the flood fill tool and try to use it to relabel the brushed segment.

daniel-wer:
The 2D flood fill tool did not work properly when flood filling in a mag other than the finest one if the "third-dimension" coordinate was not divisible by the respective mag. This resulted in early-out flood fills only filling a single or a few voxels.
Also, this commit adjusts the flood fill maximum bounding size limits if the segmentation layer's finest mag is not 1. This allows to use the flood fill tool in coarser mags if the segmentation layer is properly mag-restricted.

- https://___.webknossos.xyz

### Steps to test:
- To test the fix for 2D flood fill for coarse mags
  - Create an annotation with the volume layer restricted to mag 16.
  - Zoom out to mag 16.
  - Activate the flood fill tool in 2D and relabel a segment (on master this only works if the z-position is divisible by the z-component of mag 16, otherwise a single or only a few voxels are labeled)
- To test the adjusted bbox size limit:
  - Create an annotation with the volume layer restricted to mag 16.
  - Create a bounding box of size 1280³ vx and perform a brush stroke inside.
  - Zoom out to mag 16.
  - Activate the flood fill tool in 3D, activate the bbox restriction, and try to use it to relabel the brushed segment.

### Issues:
- fixes #8366 

------

### DIFF
--- a/frontend/javascripts/oxalis/model/bucket_data_handling/data_cube.ts
+++ b/frontend/javascripts/oxalis/model/bucket_data_handling/data_cube.ts
@@ -696,22 +696,27 @@ class DataCube {
               currentGlobalBucketPosition,
               V3.scale3(adjustedNeighbourVoxelXyz, currentMag),
             );
+            // When flood filling in a coarser mag, a voxel in the coarse mag is more than one voxel in mag 1
+            const voxelBoundingBoxInMag1 = new BoundingBox({
+              min: currentGlobalPosition,
+              max: V3.add(currentGlobalPosition, currentMag),
+            });
 
             if (bucketData[neighbourVoxelIndex] === sourceSegmentId) {
-              if (floodfillBoundingBox.containsPoint(currentGlobalPosition)) {
+              if (floodfillBoundingBox.intersectedWith(voxelBoundingBoxInMag1).getVolume() > 0) {
                 bucketData[neighbourVoxelIndex] = segmentId;
                 markUvwInSliceAsLabeled(neighbourVoxelUvw);
                 neighbourVoxelStackUvw.pushVoxel(neighbourVoxelUvw);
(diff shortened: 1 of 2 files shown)


philippotto (approved):
awesome! testing went well :+1: thanks for looking into this. I only left :nail_care: feedback.

daniel-wer:
I stumbled upon an unused import and wondered why biome doesn't fail. Turns out the rule was not explicitly enabled and seems to be turned off by default :shrug: Was automatically fixable, though :)

philippotto:
> I stumbled upon an unused import and wondered why biome doesn't fail. Turns out the rule was not explicitly enabled and seems to be turned off by default 🤷 Was automatically fixable, though :)

yes, that rule got lost in a pr which enabled import sorting. i already readded it in my dtype pr. hopefully that one gets merged soon :see_no_evil:

Issue #7962, opened 2024-07-31
Enforce Storage Quota already before upload


fm3:
We should determine the size of the to-uploaded files already in the frontend and check if they exceed the quota.

reserveUpload should send either a total file size or even a file-size per file map. the reserve call should be denied if the sum is too large.

also, the actual uploadChunk requests should be denied if they try to write more than reserved.

normanrz:
We should also enforce that an upload does not contain more bytes than were reserved.

PR #8385, opened 2025-02-06
Enforce storage quota before upload

MichaelBuessemeyer:
Currently, the storage quota enforcement does not consider the size of the current upload. With these changes the upload sends it upload size so the check can be done including the size of the dataset that should be uploaded. The backend auto rejects in case the total sum overflows the allowed storage quota. Additionally, the backend automatically rejects chucks which would overflow the reserved amount of bytes for the upload.

- https://___.webknossos.xyz

### Steps to test:
- I suggest testing locally
- Just upload a dataset via the frontend. Everything should be working nicely.
- Open the frontend file `dataset_upload_view.tsx` and edit line 319 to force it to be less bytes than the dataset you now want to upload. Then save and reload the frontend.
- Try uploading the dataset, during upload the request should fail as the frontend tries to upload more bytes than originally reserved with the request
- Undo the `dataset_upload_view.tsx` file changes
- Open the database and restrict the allowed storage amount of the default orga such that the next dataset you want to upload should exceed the limit.
  -  in `application.conf` set 
  ```
  fetchUsedStorage {
    rescanInterval = 30 seconds # do not scan organizations whose last scan is more recent than this
    tickerInterval = 10 seconds # scan some organizations at each tick
    scansPerTick = 10 # scan x organizations at each tick
  }
  ```
  - and `reportUsedStorage.enabled = true`
  - And set organization plan to `Team` or so (not sure whether that necessary)
- Now try to upload the dataset. The request should also fail.

### TODOs:
- [x] rename `totalFileSize` to `totalFileSizeInBytes`
- [x] Clean up redis keys upon successful upload
- [ ] When the reserve upload is rejected because the new dataset would exceed the orgas storage quota, the error sent by the backend to the frontend is nested and not so clear: 
image
Do you have an improvement idea @fm3?
- [x] Upon finishing upload check again before extracting the potential existing archive whether the uploaded storage exceeds the reserved amount. 
- [x] When the upload is rejected, delete the uploaded data to save disk space

### Issues:
- fixes [#7962](https://github.com/scalableminds/webknossos/issues/7962)

------


MichaelBuessemeyer:
@fm3 Would you please take a look at this PR? The changes in the frontend are very minimal right now. So it should be ok if no frontend dev takes a look at it imo.

webknossos-datastore/app/com/scalableminds/webknossos/datastore/services/uploading/UploadService.scala
```diff
@@ -239,6 +253,13 @@ class UploadService @Inject()(dataSourceRepository: DataSourceRepository,
       dataSourceId <- getDataSourceIdByUploadId(uploadId)
       (filePath, uploadDir) <- getFilePathAndDirOfUploadId(uploadFileId)
       isFileKnown <- runningUploadMetadataStore.contains(redisKeyForFileChunkCount(uploadId, filePath))
+      totalFileSize <- runningUploadMetadataStore.find(redisKeyForTotalFileSize(uploadId))
+      _ <- Fox.runOptional(totalFileSize) { maxFileSize =>
+        runningUploadMetadataStore
+          .find(redisKeyForCurrentUploadedTotalFileSize(uploadId))
+          .flatMap(alreadyUploadedAmountOfBytes =>
+            bool2Fox(alreadyUploadedAmountOfBytes.getOrElse("0").toLong + chunkSize <= maxFileSize.toLong))
+      } ?~> "dataset.upload.moreBytesThanReserved"
```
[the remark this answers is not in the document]
MichaelBuessemeyer: Yeah that might actually be a thing. Will take a look at it

webknossos-datastore/app/com/scalableminds/webknossos/datastore/services/uploading/UploadService.scala
```diff
@@ -163,6 +168,17 @@ class UploadService @Inject()(dataSourceRepository: DataSourceRepository,
...
+      _ <- Fox.runOptional(reserveUploadInfo.totalFileSizeInBytes) { fileSize =>
+        Fox.combined(
+          List(
+            runningUploadMetadataStore.insert(redisKeyForTotalFileSizeInBytes(reserveUploadInfo.uploadId),
+                                              String.valueOf(fileSize)),
+            runningUploadMetadataStore
+              .insert(redisKeyForCurrentUploadedTotalFileSizeInBytes(reserveUploadInfo.uploadId), "0")
+          ))
+      }
+      _ <- runningUploadMetadataStore.insert(redisKeyForCurrentUploadedTotalFileSizeInBytes(reserveUploadInfo.uploadId),
```
fm3, line 180: didn’t you just insert something at this key in the line above, inside of the runOptional?

webknossos-datastore/app/com/scalableminds/webknossos/datastore/services/uploading/UploadService.scala
```diff
@@ -296,6 +319,14 @@ class UploadService @Inject()(dataSourceRepository: DataSourceRepository,
       datasetNeedsConversion = uploadInformation.needsConversion.getOrElse(false)
       uploadDir = uploadDirectory(dataSourceId.organizationId, uploadId)
       unpackToDir = dataSourceDirFor(dataSourceId, datasetNeedsConversion)
+      totalFileSizeInBytesOpt <- runningUploadMetadataStore.find(redisKeyForTotalFileSizeInBytes(uploadId))
+      _ <- Fox.runOptional(totalFileSizeInBytesOpt) { maxFileSize =>
+        tryo(FileUtils.sizeOfDirectoryAsBigInteger(uploadDir.toFile).longValue).map(actualFileSize =>
```
fm3, line 324: did you double-check that this returns the same value as the frontend-side check? (both for archives and multi-file uploads)?

fm3 (commented):
Looking pretty good! First test worked for me.

Concerning the hidden error message: Do you think the frontend could do a check first, before the reserve request is even sent? (Maybe even before the upload button is hit?) In theory, the frontend should be able to determine if the selected files would be too much, right? Since it knows both limit and existing amount in the organization?

Did you also test uploading with the libs? If not i could probably also do that.

MichaelBuessemeyer (commented):
> In theory, the frontend should be able to determine if the selected files would be too much, right? Since it knows both limit and existing amount in the organization?

Sure, thats a good suggestion. I missed that. It should now be added to the changes of this pr.
image

> Did you also test uploading with the libs? If not i could probably also do that.

Now yes: also a very good suggestion :D
=> For testing I used `v0.16.9` and it worked without any problems.

webknossos-datastore/app/com/scalableminds/webknossos/datastore/services/uploading/UploadService.scala
fm3: How big are our chunks? Could this become slow if it runs for every chunk?

fm3 (commented):
Looking good! What happens exactly, when an uploadChunk request fails? Will the frontend retry it or abort? Should we add a cleanup functionality on the server in case of abort?

MichaelBuessemeyer:
> What happens exactly, when an uploadChunk request fails?

Then the upload should be cancelled if one of the following listed error codes is returned. Else it should retry infinitely. Here are all options set by the frontend to the [resumable upload js lib](https://www.npmjs.com/package/resumablejs#configuration):
https://github.com/scalableminds/webknossos/blob/e9f47318a752e3cd349e7881ed416614b6a85476/frontend/javascripts/admin/admin_rest_api.ts#L1124-L1135

> How big are our chunks? Could this become slow if it runs for every chunk?

So chunk size is something between 10MB and 20MB (according to docs sometimes up to two-fold of the chunk size may be used). Imo that should be fine.

> Will the frontend retry it or abort?

That depends on the error. See the `permanentErrors` setting.

> Should we add a cleanup functionality on the server in case of abort?

I think `cleanUpUploadedDataset(uploadDir, uploadId)` should already do the clean up. I changed the `handleUploadChunk` method so that it automatically calls `cleanUpUploadedDataset(uploadDir, uploadId)` in case the current chunk exceeds the upload limit.

Do you think I should give the retying / auto clean up some testing?

MichaelBuessemeyer:
AI says:

> ---

`66-75`: **Consider reducing code duplication with the `insert` method**

The `insertLong` method is almost identical to the existing `insert` method on lines 55-64, differing only in the parameter type.

You could either:

1. Refactor both methods to use a generic type parameter with a type class approach, or
2. Implement `insertLong` in terms of `insert` by converting the Long to String:

```diff
def insertLong(id: String, value: Long, expirationOpt: Option[FiniteDuration] = None): Fox[Unit] =
-  withExceptionHandler {
-    expirationOpt
-      .map(
-        expiration => r.setex(id, expiration.toSeconds, value)
-      )
-      .getOrElse(
-        r.set(id, value)
-      )
-  }
+  insert(id, value.toString, expirationOpt)
```

This would simplify maintenance and reduce the risk of inconsistencies between the methods.

@fm3 do you think its worth it? I wanted to make sure that long is actually passed to the setex / set call.

MichaelBuessemeyer (commented):
Ok I checked the auto clean up and it worked
Moreover, the auto retry also works (I rejected 10% of the chunks at random and the upload still succeeded. 

Moreover, I added some Long support functions to the redis client, removed all pending keys from redis on failure and success and also changed the route that fetches unfinished uploads to not return uploads that have no redis data (-> failed / cleaned up uploads).

> Also, since the frontend changes are growing, maybe it would be a good idea to ask another frontend dev to have a look.

@daniel-wer Do you maybe have few minutes to give this a quick look? Should be rather straight forward: The reserve upload route now has an additional parameter to reserve the space needed for the upload and also a new validation checking that the file that should be uploaded does not exceed the currently know left over storage space of the organization

webknossos-datastore/app/com/scalableminds/webknossos/datastore/services/uploading/UploadService.scala
```diff
@@ -185,21 +199,23 @@ class UploadService @Inject()(dataSourceRepository: DataSourceRepository,
   def addUploadIdsToUnfinishedUploads(
       unfinishedUploadsWithoutIds: List[UnfinishedUpload]): Fox[List[UnfinishedUpload]] =
     for {
-      maybeUnfinishedUploads: List[Box[UnfinishedUpload]] <- Fox.sequence(
+      maybeUnfinishedUploads: List[Box[Option[UnfinishedUpload]]] <- Fox.sequence(
```
fm3, line 202: what is the maybe encoding here? Could the listed uploads actually be finished?

webknossos-datastore/app/com/scalableminds/webknossos/datastore/services/uploading/UploadService.scala
```diff
@@ -185,21 +199,23 @@ class UploadService @Inject()(dataSourceRepository: DataSourceRepository,
...
             for {
               uploadIdOpt <- runningUploadMetadataStore.find(redisKeyForUploadId(unfinishedUpload.dataSourceId))
-              maybeUpdateUpload = uploadIdOpt
-                .map(uploadId => unfinishedUpload.copy(uploadId = uploadId))
-                .getOrElse(unfinishedUpload)
-              filePathsStringOpt <- runningUploadMetadataStore.find(redisKeyForFilePaths(maybeUpdateUpload.uploadId))
-              filePathsOpt <- filePathsStringOpt.map(JsonHelper.parseAndValidateJson[List[String]])
-              uploadUpdated <- filePathsOpt.map(filePaths => maybeUpdateUpload.copy(filePaths = Some(filePaths)))
-            } yield uploadUpdated
+              updateUploadOpt = uploadIdOpt.map(uploadId => unfinishedUpload.copy(uploadId = uploadId))
```
fm3, line 207: should that be update*d*?

webknossos-datastore/app/com/scalableminds/webknossos/datastore/services/uploading/UploadService.scala
```diff
@@ -185,21 +199,23 @@ class UploadService @Inject()(dataSourceRepository: DataSourceRepository,
...
+              updateUploadOpt <- Fox.runOptional(updateUploadOpt)(updatedUpload =>
```
fm3, line 208: please don’t reuse variable names in the same scope

webknossos-datastore/app/com/scalableminds/webknossos/datastore/services/uploading/UploadService.scala
```diff
@@ -185,21 +199,23 @@ class UploadService @Inject()(dataSourceRepository: DataSourceRepository,
...
+                  filePathsStringOpt <- runningUploadMetadataStore.find(redisKeyForFilePaths(updatedUpload.uploadId))
+                  filePathsOpt <- filePathsStringOpt.map(JsonHelper.parseAndValidateJson[List[String]])
+                  uploadUpdatedWithFilePaths <- filePathsOpt.map(filePaths =>
+                    updatedUpload.copy(filePaths = Some(filePaths)))
+                } yield uploadUpdatedWithFilePaths)
+            } yield updateUploadOpt
           }
         ))
-      foundUnfinishedUploads = maybeUnfinishedUploads.flatten
+      foundUnfinishedUploads = maybeUnfinishedUploads.flatten.flatten
```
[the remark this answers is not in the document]
fm3: I’d say it’s ok. The function is a little hard to read, so could you please add a comment on what it’s for, and that it does indeed filter the uploads no longer fully present in redis?

fm3:
> So chunk size is something between 10MB and 20MB (according to docs sometimes up to two-fold of the chunk size may be used). Imo that should be fine.

Ok, great!

> @fm3 do you think [converting the Long to string on insert] is worth it? I wanted to make sure that long is actually passed to the setex / set call.

I believe redis will convert everything to string anyway (otherwise, why would we have to parse on read?) So yes, I think we could reuse the existing insert.

Latest changes look good, good call on removing the outdated entries.

frontend/javascripts/admin/dataset/dataset_upload_view.tsx
```diff
@@ -331,6 +338,7 @@ class DatasetUploadView extends React.Component<PropsWithFormAndRouter, State> {
       ? unfinishedUploadToContinue.uploadId
       : `${dayjs(Date.now()).format("YYYY-MM-DD_HH-mm")}__${newDatasetName}__${getRandomString()}`;
     const filePaths = formValues.zipFile.map((file) => file.path || "");
+    const totalFileSizeInBytes = getFileSize(formValues.zipFile);
```
daniel-wer, line 341: Just to make sure, this file size measurement is not used for the quota computation, but is used by the server for something else, right? Otherwise, the unzipped file size should be measured.

frontend/javascripts/admin/dataset/dataset_upload_view.tsx
```diff
@@ -995,6 +1005,19 @@ class DatasetUploadView extends React.Component<PropsWithFormAndRouter, State> {
                     return archives.length === 0 || archives.length === files.length;
                   }, "Archives cannot be mixed with other files."),
                 },
+                {
+                  validator: syncValidator(
+                    (files: FileWithPath[]) => {
+                      const fileSize = getFileSize(files);
+                      return getLeftOverStorageBytes(this.props.organization) > fileSize;
```
daniel-wer, line 1012: Nitpicky, but I'd say this is more correct :)

```suggestion
                      return getLeftOverStorageBytes(this.props.organization) >= fileSize;
```

daniel-wer (commented):
Frontend code LGTM, did not test :+1:

webknossos-datastore/app/com/scalableminds/webknossos/datastore/services/uploading/UploadService.scala
MichaelBuessemeyer: During working on this PR and checking whether the auto clean-up upon rejection (due to reserved storage overflow / too much storage after upload and in the finish upload route) I noticed that the rejected uploads were listed in the the fronted as continuable. But this is undesired in this case. This had multiple reasons:
1. Not all redis entries for the upload were clean up upon rejection
2. The postgres entry for the not finished uploaded dataset still existed

So to tackle this I tired to solve both problems by, 1. delete all redis keys, 2. delete the dataset from postgres and 3. to only send unfinished uploads which are present in the redis db.

Do you think 3rd is a little too much?

fm3: No, I think that makes sense. The ones not in redis can’t be properly continued anyway if I’m not mistaken

frontend/javascripts/admin/dataset/dataset_upload_view.tsx
MichaelBuessemeyer: The file size is used to early reject far too large uploads. I discussed this with @fm3. In case the compressed version already exceeds the left over storage space, the upload is rejected. Else if it exceeds after the decompression on the server side, we see this as ok. Otherwise, the frontend would need to decompress the files itself before upload making the upload workflow a little less streamlined and maybe even needing pretty long time as it is a decompression done in the browser.

MichaelBuessemeyer:
TODO
- [ ] Test whether unfinished uploads without redis entries are not sent to the client as unfinished datasets uploads

MichaelBuessemeyer:
Sooo I think this can go for another iteration :recycle:

webknossos-datastore/app/com/scalableminds/webknossos/datastore/services/uploading/UploadService.scala
```diff
@@ -185,21 +199,24 @@ class UploadService @Inject()(dataSourceRepository: DataSourceRepository,
   def addUploadIdsToUnfinishedUploads(
       unfinishedUploadsWithoutIds: List[UnfinishedUpload]): Fox[List[UnfinishedUpload]] =
     for {
-      maybeUnfinishedUploads: List[Box[UnfinishedUpload]] <- Fox.sequence(
+      maybeUnfinishedUploads: List[Box[Option[UnfinishedUpload]]] <- Fox.sequence(
+        // Filter out unfinished uploads that do not have entries in the redis store and adds the uploadId and filePaths to those that have an entry.
```
fm3, line 203:
```suggestion
        // Previously rejected uploads may still appear in this list, but don’t have entries in redis. We can use that to filter them out here, since we don’t want to list them to the user. Those that *do* have entries in redis are then enriched using info from there (uploadId and filePaths).
```
(Is this correct? To me your version of the comment seemed like more a description of the code than of intent. What do you think?)

frontend/javascripts/admin/dataset/dataset_upload_view.tsx
```diff
@@ -995,6 +1005,19 @@ class DatasetUploadView extends React.Component<PropsWithFormAndRouter, State> {
                     return archives.length === 0 || archives.length === files.length;
                   }, "Archives cannot be mixed with other files."),
                 },
+                {
+                  validator: syncValidator(
+                    (files: FileWithPath[]) => {
+                      const fileSize = getFileSize(files);
+                      return getLeftOverStorageBytes(this.props.organization) >= fileSize;
+                    },
+                    `The uploaded files exceed the available storage of your organization. Please ${
```
fm3, line 1014:
```suggestion
                    `The selected files exceed the available storage of your organization. Please ${
```

fm3 (approved):
LGTM, just added two suggestions on wording

frontend/javascripts/admin/dataset/dataset_upload_view.tsx
MichaelBuessemeyer: > Does this mean users can still exceed their storage quota if the umcompressed data is (much) larger than the uploaded zip file?

Yes, that's what we agreed upon :+1: 

> I think the uncompressed size of the zip entries is available via metadata, see https://gildas-lormeau.github.io/zip.js/api/interfaces/Entry.html#uncompressedsize
If I read the code correctly we already iterate over all file entries (reader.getEntries()) and could access the uncompressed size there as well.
>
> Feel free to ignore if this doesn't have priority. I just wanted to note :)

@fm3 Do you think I should do this in this PR? I personally would prefer to make this a follow-up issue. The PR is already open quite a long time and this should be easy to do as a follow-up.

fm3: ok let’s go for follow-up issue

webknossos-datastore/app/com/scalableminds/webknossos/datastore/services/uploading/UploadService.scala
MichaelBuessemeyer: Now that I think about this, it is not entirely correct.
> Previously rejected uploads may still appear in this list,

Should usually not be the case as the clean up upon failure of the upload also removes them from postgres (as far as I know). But if this fails (or is forgotten like I did at some stage of this PR), then the datasets would appear in the list.

So this is more like a save guard against missing / failing to delete the postgres entry. 

So what do you think: Is it ok to keep this check here that removes datasets that are not present in the redis as they cannot be resumed anyway? Or should this code be removed as this case should very rarely happen (e.g. wrong code / forgot to remove or failed rpc to the main backend)

webknossos-datastore/app/com/scalableminds/webknossos/datastore/controllers/DataSourceController.scala
```diff
@@ -149,16 +150,17 @@ class DataSourceController @Inject()(
         tuple(
           "resumableChunkNumber" -> number,
           "resumableChunkSize" -> number,
+          "resumableCurrentChunkSize" -> number,
```
MichaelBuessemeyer, line 153: @fm3 this should be fine to add as a param as it is already sent by previous wklibs versions, correct @markbader?

Mroeover, now that I added this, your wklibs pr should hopefully be working now :+1:.

fm3: Yes, should be fine. The old libs do send this already.
If someone lies about this number, the chunk request may wrongly pass, but finishUpload should ultimately be rejected, which seems fine to me. :shipit:

PR #8387, opened 2025-02-06
Support reading Blosc-compressed data chunks with autoshuffle

### LINKED ISSUE #8381: Support decompressing Chunks with Blosc AUTOSHUFFLE (-1)
When decompressing zarr chunks we got this message from JBlosc:

`blosc: shuffle type not supported: '-1'; expected one of 0 (NOSHUFFLE),1 (BYTESHUFFLE),2 (BITSHUFFLE)`

[According to this document](https://numcodecs.readthedocs.io/en/stable/compression/blosc.html), -1 is AUTOSHUFFLE. There, bit-shuffle will be used for buffers with itemsize 1, and byte-shuffle will be used otherwise. The default is SHUFFLE.

If AUTOSHUFFLE, bit-shuffle will be used for buffers with itemsize 1, and byte-shuffle will be used otherwise. The default is SHUFFLE.

We should either update to a Blosc lib that supports this AUTOSHUFFLE directly or mutate its input so that we do the byteshuffle-bitshuffle switch as described by this specification.

Context and example zarr dataset: https://scm.slack.com/archives/C02H5T8Q08P/p1738766454436339

fm3:
Turns out the error didn’t come from JBlosc, but our own implementation. @frcroth do you remember if there was a specific reason to originally reject AUTOSHUFFLE?

- https://autoshuffle.webknossos.xyz

### Steps to test:
- Import dataset from https://scm.slack.com/archives/C02H5T8Q08P/p1738766454436339 – should show data

### Issues:
- fixes #8381 

------


frcroth:
> Turns out the error didn’t come from JBlosc, but our own implementation. @frcroth do you remember if there was a specific reason to originally reject AUTOSHUFFLE?

I have no idea, it was introduced here: https://github.com/scalableminds/webknossos/pull/6466

PR #8392, opened 2025-02-12
Command palette

knollengewaechs:
- https://___.webknossos.xyz

### Steps to test:
- abc

### TODOs:
- [ ] ...

### Issues:
- fixes #

------


frontend/javascripts/oxalis/view/components/command_palette.tsx
```diff
@@ -0,0 +1,198 @@
...
+      isAnnotationLockedByUser: state.tracing.isLockedByOwner,
+      annotationOwner: state.tracing.owner,
+      isDownloadModalOpen: state.uiInformation.showDownloadModal,
+      isRenderAnimationModalOpen: state.uiInformation.showRenderAnimationModal,
+      isShareModalOpen: state.uiInformation.showShareModal,
+      isMergeModalOpen: state.uiInformation.showMergeAnnotationModal,
+      isZarrPrivateLinksModalOpen: state.uiInformation.showZarrPrivateLinksModal,
+      isUserScriptsModalOpen: state.uiInformation.showAddScriptModal,
+    };
+  });
```
philippotto, line 79: this kind of selector will return a new object every time the state changes which is not ideal from a perf perspective. for the old non-hooks-way of doing things (with mapStateToProps), this is not a problem, but here it's not really ideal. this would be better afaik:

```ts
const restrictions = useSelector((state: OxalisState) => state.tracing.restrictions)
const task = useSelector((state: OxalisState) => state.task)
const annotationType = useSelector((state: OxalisState) => state.tracing.annotationType)
const annotationId = useSelector((state: OxalisState) => state.tracing.annotationId)
const activeUser = useSelector((state: OxalisState) => state.activeUser)
const isAnnotationLockedByUser = useSelector((state: OxalisState) => state.tracing.isLockedByOwner)
const annotationOwner = useSelector((state: OxalisState) => state.tracing.owner)
const isDownloadModalOpen = useSelector((state: OxalisState) => state.uiInformation.showDownloadModal)
const isRenderAnimationModalOpen = useSelector((state: OxalisState) => state.uiInformation.showRenderAnimationModal)
const isShareModalOpen = useSelector((state: OxalisState) => state.uiInformation.showShareModal)
const isMergeModalOpen = useSelector((state: OxalisState) => state.uiInformation.showMergeAnnotationModal)
const isZarrPrivateLinksModalOpen = useSelector((state: OxalisState) => state.uiInformation.showZarrPrivateLinksModal)
const isUserScriptsModalOpen = useSelector((state: OxalisState) => state.uiInformation.showAddScriptModal)
 
const props: TracingViewMenuProps = {
  restrictions,
  task,
  annotationType,
  annotationId,
  activeUser,
  isAnnotationLockedByUser,
  annotationOwner,
  isDownloadModalOpen,
  isRenderAnimationModalOpen,
  isShareModalOpen,
  isMergeModalOpen,
  isZarrPrivateLinksModalOpen,
  isUserScriptsModalOpen,
}
```

frontend/javascripts/oxalis/view/components/command_palette.tsx
```diff
@@ -0,0 +1,198 @@
...
+      { name: "Annotations", path: "/dashboard/annotations" },
+      { name: "Datasets", path: "/dashboard/datasets" },
+      { name: "Time Tracking", path: "/timetracking" },
+    ];
+
+    const adminMenu = getAdministrationSubMenu(false, activeUser);
+    const adminCommands: Array<{ name: string; path: string }> = [];
+    adminMenu?.children.map((entry: { key: string }) => {
+      adminCommands.push({ name: getLabelForPath(entry.key), path: entry.key });
+    });
```
philippotto, line 124: a map with side effects is not idiomatic. how about:


```suggestion
    const adminCommands = adminMenu == null ? [] : adminMenu.children.map((entry: { key: string }) => 
      ({ name: getLabelForPath(entry.key), path: entry.key })
    );
```

frontend/javascripts/oxalis/view/components/command_palette.tsx
```diff
@@ -0,0 +1,198 @@
...
+            name: "Available Tasks",
+          },
+        ]
+      : [];
+
+    const navigationEntries = [...basicNavigationEntries, ...adminCommands, ...statisticsCommands];
+
+    navigationEntries.forEach((entry, counter) => {
+      commands.push({
+        id: counter,
```
philippotto, line 143: how about adding a `category: "Navigate"` property here. will look like this: https://main--5c34dd1c22fa8200244c31dd.chromatic.com/?path=/story/command-palette--with-multiple-highlights
the same for settings-related stuff

frontend/javascripts/oxalis/view/components/command_palette.tsx
```diff
@@ -0,0 +1,198 @@
...
+    let availableTools = Object.keys(AnnotationToolEnum) as [keyof typeof AnnotationToolEnum];
+    if (isViewMode || !restrictions.allowUpdate) {
+      availableTools = AvailableToolsInViewMode as [keyof typeof AnnotationToolEnum];
+    }
+    availableTools.forEach((tool, counter) => {
+      commands.push({
+        id: counter,
+        name: `Switch to ${getLabelForTool(tool)} Tool`,
+        command: () => Store.dispatch(setToolAction(tool)),
+        color: "#5660ff",
```
philippotto, line 167: this color appears multiple times in this module. extract into var?

frontend/javascripts/oxalis/view/components/command_palette.tsx
```diff
@@ -0,0 +1,198 @@
...
+    ...mapMenuActionsToCommands(menuActions),
+    ...getTabsAndSettingsMenuItems(),
+  ];
+  return (
+    <div style={{ marginRight: "10px" }}>
+      <ReactCommandPalette
+        commands={allCommands.map((command, counter) => {
+          return {
+            ...command,
+            id: counter,
```
philippotto, line 187: so, this overwrites all previously defined ids (which is good because they would be not unique otherwise). can we avoid defining the other ids in the first place?

frontend/javascripts/oxalis/view/action-bar/tracing_actions_view.tsx
```diff
@@ -261,11 +266,258 @@ export function getLayoutMenu(props: LayoutMenuProps): SubMenuType {
   };
 }
 
+export const getTracingViewModalsAndMenuItems = (
```
philippotto, line 269: cool refactoring :+1:
two things:
1) could you move all handle* variables to the top level? I think they don't depend on the parameters of this function. if they need to access the store, they can do Store.getState() for that (since the handler is only executed at will from the outside, it doesn't need to be reactive).
2) I would remove the modal stuff from this function again. it's not needed by the command palette. either, move it back to TracingActionsView or extract it into another helper function. or is there a reason why they need to stay together here?

afterwards, it might make sense to check again whether it's necessary to require all TracingViewMenuProps for this function. maybe we can do with less (which could also make the useSelector stuff in the command palette module simpler).

frontend/javascripts/libs/utils.ts
```diff
@@ -1300,3 +1300,10 @@ export function generateRandomId(length: number) {
   }
   return result;
 }
+
+export function getPhraseFromCamelCaseString(stringInCamelCase: string): string {
```
philippotto, line 1304: I'm not a big fan of such automatic conversions to make something readable for humans. I'd rather have explicit (and type-enforced) mappings. we can keep it for this iteration, but just fyi :)

philippotto (changes requested):
I'm really looking forward to seeing this in production :) left some comments and would also love to see the styling a bit more tuned to wk colors (as a trade off, it can stay dark even in light mode, I think).

frontend/javascripts/oxalis/view/components/command_palette.tsx
```diff
@@ -0,0 +1,198 @@
...
+  if ("label" in action && action.label != null) {
+    return action.label.toString();
+  }
+  return "";
+};
+
+const mapMenuActionsToCommands = (menuActions: Array<ItemType>): Command[] => {
+  if (menuActions == null) return [];
+  return menuActions
+    .filter((action) => action != null && getLabelForAction(action) != null)
```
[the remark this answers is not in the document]
knollengewaechs: thats working! :raised_hands: 
The last line is not supposed so be in the code (e.g. as a comment), right?

frontend/javascripts/oxalis/view/components/command_palette.tsx
knollengewaechs: I really like the idea and I think the example looks nice. But I feel like this prop might have been removed, because the typechecker doesnt know it and its not rendered. a search within the repo also seemed to yield results in tests and demos https://github.com/search?q=repo%3Aasabaylus%2Freact-command-palette+category&type=code

frontend/javascripts/oxalis/view/action-bar/tracing_actions_view.tsx
knollengewaechs: > or is there a reason why they need to stay together here?

Right. I also thought about separating the modals from the menu items. I thought it was nice that the conditions for menu items and modals were the same, because menu item and modal were added in the same place (e.g. download button and download modal). But you are right that its not clean to create the modals twice and only use them once. I will separate modals and menu items!

frontend/javascripts/libs/utils.ts
knollengewaechs: Alright! I added it to the todo-list for future iterations

knollengewaechs:
@philippotto I have a few open questions that I commented. Other than that, I tried to incorporate all your comments and improved the styling. I tested a bit after the changes, but did not test all entries in every case. If you want me to do that before you review or test again, let me know :) Other than that, feel free to re-review!

frontend/javascripts/oxalis/view/components/command_palette.tsx
philippotto: ok, then let's forget about it for now :)

frontend/javascripts/oxalis/view/components/command_palette.tsx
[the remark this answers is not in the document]
philippotto: sry, the formatting was messed up. it was supposed to be outside of the code block and meant for the actual github comment :)

frontend/javascripts/oxalis/view/action-bar/tracing_actions_view.tsx
```diff
@@ -491,7 +448,7 @@ export const getTracingViewModalsAndMenuItems = (
     });
   }
 
-  menuItems.push(layoutMenu);
+  layoutMenu != null && menuItems.push(layoutMenu);
```
philippotto, line 451: I usually avoid such short hands as I find it harder to read than:
```suggestion
  if (layoutMenu != null) {
    menuItems.push(layoutMenu);
  }
```

philippotto (approved):
great stuff! only see my one nitpick comment :) and don't forget to open the follow-up issue (at least I didn't see it mentioned).

frontend/javascripts/oxalis/view/action-bar/tracing_actions_view.tsx
knollengewaechs: I abbreviated the if-clause by omitting the brackets for it to be in one line, I hope thats okay too :)

PR #8393, opened 2025-02-12
Implement Webauthn

robert-oleynik:
- https://___.webknossos.xyz

### Steps to test:
- abc

### TODOs:
- [ ] ...

### Issues:
- fixes #

------

### DIFF
--- a/app/controllers/AuthenticationController.scala
+++ b/app/controllers/AuthenticationController.scala
@@ -3,6 +3,10 @@ package controllers
 import com.scalableminds.util.accesscontext.{DBAccessContext, GlobalAccessContext}
 import com.scalableminds.util.objectid.ObjectId
 import com.scalableminds.util.tools.{Fox, FoxImplicits, TextUtils}
+import com.scalableminds.webknossos.datastore.storage.TemporaryStore
+import com.yubico.webauthn._
+import com.yubico.webauthn.data._
+import com.yubico.webauthn.exception._
 import mail.{DefaultMails, MailchimpClient, MailchimpTag, Send}
 import models.analytics.{AnalyticsService, InviteEvent, JoinOrganizationEvent, SignupEvent}
 import models.organization.{Organization, OrganizationDAO, OrganizationService}
@@ -29,8 +33,15 @@ import utils.WkConf
 import java.net.URLEncoder
 import java.nio.charset.StandardCharsets
 import java.security.MessageDigest
+import java.util.UUID
 import javax.inject.Inject
+import scala.concurrent.duration.DurationInt
 import scala.concurrent.{ExecutionContext, Future}
+
+case class WebAuthnRegistration(name: String, key: String)
+object WebAuthnRegistration {
+  implicit val jsonFormat: OFormat[WebAuthnRegistration] = Json.format[WebAuthnRegistration]
+}
 
 class AuthenticationController @Inject()(
     actorSystem: ActorSystem,
@@ -52,6 +63,10 @@ class AuthenticationController @Inject()(
     openIdConnectClient: OpenIdConnectClient,
     initialDataService: InitialDataService,
     emailVerificationService: EmailVerificationService,
+    webAuthnCredentialRepository: WebAuthnCredentialRepository,
+    webAuthnCredentialDAO: WebAuthnCredentialDAO,
+    temporaryAssertionStore: TemporaryStore[String, AssertionRequest],
+    temporaryRegistrationStore: TemporaryStore[ObjectId, PublicKeyCredentialCreationOptions],
     sil: Silhouette[WkEnv])(implicit ec: ExecutionContext, bodyParsers: PlayBodyParsers)
     extends Controller
     with AuthForms
@@ -65,6 +80,18 @@ class AuthenticationController @Inject()(
 
   private lazy val ssoKey =
     conf.WebKnossos.User.ssoKey
+
+  private lazy val relyingParty = {
+    var identity = RelyingPartyIdentity
+      .builder()
+      .id("webknossos.local:9000") // TODO: Use Host
+      .name("WebKnossos")
+      .build();
+    RelyingParty.builder()
+      .identity(identity)
+      .credentialRepository(webAuthnCredentialRepository)
+      .build()
+  }
 
   private lazy val isOIDCEnabled = conf.Features.openIdConnectEnabled
 
@@ -405,6 +432,89 @@ class AuthenticationController @Inject()(
       case None => Fox.successful(Redirect("/auth/login?redirectPage=http://discuss.webknossos.org")) // not logged in
     }
   }
+
+  def webauthnAuthStart(): Action[AnyContent] = Action{ implicit request => {
+      val opts = StartAssertionOptions.builder().build();
+      val assertion = relyingParty.startAssertion(opts);
+      val sessionId = UUID.randomUUID().toString;
+      val cookie = Cookie("webauthn-session", sessionId, maxAge = Some(120), httpOnly = true, secure = true)
+      temporaryAssertionStore.insert(sessionId, assertion, Some(2 minutes));
+      Ok(Json.toJson(Json.parse(assertion.toJson))).withCookies(cookie)
+    }
+  }
+
+  def webauthnAuthFinalize(): Action[String] = Action.async(validateJson[String]) { implicit request => {
+      request.cookies.get("webauthn-session") match {
+        case Some(cookie) =>
+          val sessionId = cookie.value
+          val challengeData = temporaryAssertionStore.get(sessionId)
+          temporaryAssertionStore.remove(sessionId)
+          challengeData match {
+            case Some(data) =>
+              try {
+                val keyCredential = PublicKeyCredential.parseAssertionResponseJson(request.body);
+                val opts = FinishAssertionOptions.builder()
+                  .request(data)
+                  .response(keyCredential)
+                  .build();
+                val result = relyingParty.finishAssertion(opts);
+                val userId = WebAuthnCredentialRepository.byteArrayToObjectId(result.getCredential.getUserHandle)
+                val loginInfo = LoginInfo("credentials", userId.toString)
+                loginUser(loginInfo)(request.map(_ => AnyContent()))
+              } catch {
+                case e: AssertionFailedException => Future.successful(Unauthorized("challenge failed"))
+              }
+            case None =>
+              Future.successful(BadRequest("Challenge not found or expired"))
+          }
+      }
+    }
+  }
+
+  def webauthnRegisterStart(): Action[AnyContent] = sil.SecuredAction { implicit request => {
+    val userIdentity = UserIdentity.builder()
+      .name(request.identity.name)
+      .displayName(request.identity.name)
+      .id(WebAuthnCredentialRepository.objectIdToByteArray(request.identity._id))
+      .build();
+    val opts = StartRegistrationOptions.builder()
+      .user(userIdentity)
+      .timeout(120000)
+      .build()
+    val registration = relyingParty.startRegistration(opts);
+    temporaryRegistrationStore.insert(request.identity._multiUser, registration);
+    Ok(Json.toJson(Json.parse(registration.toJson)))
+  }}
+
+  def webauthnRegisterFinalize(): Action[WebAuthnRegistration] = sil.SecuredAction.async(validateJson[WebAuthnRegistration]) { implicit request => {
+    val creationOpts = temporaryRegistrationStore.get(request.identity._multiUser)
+    temporaryRegistrationStore.remove(request.identity._multiUser)
+    creationOpts match {
+      case Some(data) => {
+        val response = PublicKeyCredential.parseRegistrationResponseJson(request.body.key);
+        val opts = FinishRegistrationOptions.builder()
+          .request(data)
+          .response(response)
+          .build();
+        try {
+          val key = relyingParty.finishRegistration(opts)
+          val credential = WebAuthnCredential(
+            ObjectId.generate,
+            request.identity._multiUser,
+            request.body.name,
+            key.getPublicKeyCose.getBytes,
+            key.getSignatureCount.toInt,
+            isDeleted = false,
+          )
+          webAuthnCredentialDAO.insertOne(credential)
+            .map(_ => Ok(""))
+        } catch {
+          case e: RegistrationFailedException => Future.successful(BadRequest("Failed to register key"))
+        }
+      }
+      case None => Future.successful(BadRequest("Challenge not found or expired"))
+    }
+  }}
 
   private lazy val absoluteOpenIdConnectCallbackURL = s"${conf.Http.uri}/api/auth/oidc/callback"
 
--- a/app/models/user/WebAuthnCredentials.scala
+++ b/app/models/user/WebAuthnCredentials.scala
@@ -1,0 +1,47 @@
+import com.scalableminds.util.accesscontext.DBAccessContext
+import com.scalableminds.util.objectid.ObjectId;
+import com.scalableminds.util.tools.Fox;
+import com.scalableminds.webknossos.schema.Tables._;
+
+import utils.sql.{SQLDAO, SqlClient};
+
+import javax.inject.Inject;
+import scala.concurrent.{ExecutionContext, Future}
+
+case class WebAuthnCredential(
+  _id: ObjectId,
+  _multiUser: ObjectId,
+  name: String,
+  publicKeyCose: Array[Byte],
+  signatureCount: Int,
+)
+
+class WebAuthnCredentialDAO @Inject()(sqlClient: SqlClient) (implicit ec: ExecutionContext)
+  extends SQLDAO[WebAuthnCredential, WebauthncredentialsRow, Webauthncredentials](sqlClient) {
+    protected val collection = Webauthncredentials
+
+    protected def parse(r: WebauthncredentialsRow): Fox[WebAuthnCredential] =
+        Fox.future2Fox(Future.successful(
+          WebAuthnCredential(
+            ObjectId(r._Id),
+            ObjectId(r._Multiuser),
+            r.name,
+            r.publickeycose,
+            r.signaturecount,
+          )
+        ))
+
+  def insertOne(c: WebAuthnCredential)(implicit ctx: DBAccessContext): Fox[Unit] =
+    for {
+      _ <- run(q"""INSERT INTO webknossos.webauthnCredentials(_id, _multiUser,
+                                                              name, publicKeyCose,
+                                                              signatureCount)
+                       VALUES(${c._id}, ${c._multiUser}, ${c.name},
+                              ${c.publicKeyCose}, ${c.signatureCount})""".asUpdate)
+    } yield ()
+
+  def removeById(id: ObjectId)(implicit ctx: DBAccessContext): Fox[Unit] =
+    for {
+      _ <- run(q"""DELETE FROM webknossos.webauthnCredentials WHERE _id = ${id}""".asUpdate)
+    } yield()
+}
--- a/app/models/user/WebAuthnCredentials.scala
+++ b/app/models/user/WebAuthnCredentials.scala
@@ -1,11 +1,14 @@
+package models.user
+
 import com.scalableminds.util.accesscontext.DBAccessContext
-import com.scalableminds.util.objectid.ObjectId;
-import com.scalableminds.util.tools.Fox;
-import com.scalableminds.webknossos.schema.Tables._;
+import com.scalableminds.util.objectid.ObjectId
+import com.scalableminds.util.tools.Fox
+import com.scalableminds.webknossos.schema.Tables
+import com.scalableminds.webknossos.schema.Tables._
+import slick.lifted.Rep
+import utils.sql.{SQLDAO, SqlClient}
 
-import utils.sql.{SQLDAO, SqlClient};
-
-import javax.inject.Inject;
+import javax.inject.Inject
 import scala.concurrent.{ExecutionContext, Future}
 
 case class WebAuthnCredential(
@@ -14,24 +17,52 @@ case class WebAuthnCredential(
   name: String,
   publicKeyCose: Array[Byte],
   signatureCount: Int,
+  isDeleted: Boolean,
 )
 
 class WebAuthnCredentialDAO @Inject()(sqlClient: SqlClient) (implicit ec: ExecutionContext)
   extends SQLDAO[WebAuthnCredential, WebauthncredentialsRow, Webauthncredentials](sqlClient) {
     protected val collection = Webauthncredentials
 
+    override protected def idColumn(x: Webauthncredentials): Rep[String] = x._Id
+
+    override protected def isDeletedColumn(x: Webauthncredentials): Rep[Boolean] = x.isdeleted
+
     protected def parse(r: WebauthncredentialsRow): Fox[WebAuthnCredential] =
-        Fox.future2Fox(Future.successful(
+        Fox.successful(
           WebAuthnCredential(
             ObjectId(r._Id),
             ObjectId(r._Multiuser),
             r.name,
             r.publickeycose,
             r.signaturecount,
+            r.isdeleted
           )
-        ))
+        )
 
-  def insertOne(c: WebAuthnCredential)(implicit ctx: DBAccessContext): Fox[Unit] =
+  def findAllForUser(userId: ObjectId)(implicit ctx: DBAccessContext): Fox[List[WebAuthnCredential]] =
+    for {
+      accessQuery <- readAccessQuery
+      r <- run(q"SELECT $columns FROM $existingCollectionName WHERE _multiUser = $userId AND $accessQuery".as[WebauthncredentialsRow])
+      parsed <- parseAll(r)
+    } yield parsed
+
+  def findById(id: ObjectId)(implicit ct: DBAccessContext): Fox[WebAuthnCredential] =
+    for {
+      accessQuery <- readAccessQuery
+      r <- run(q"SELECT $columns FROM $existingCollectionName WHERE _id = $id AND $accessQuery".as[WebauthncredentialsRow])
+      parsed <- parseFirst(r, id)
+    } yield parsed
+
+
+  def findByIdAndUserId(id: ObjectId, userId: ObjectId)(implicit ctx: DBAccessContext): Fox[WebAuthnCredential] =
+    for {
+      accessQuery <- readAccessQuery
+      r <- run(q"SELECT $columns FROM $existingCollectionName WHERE _id = $id AND _multiUser = $userId AND $accessQuery".as[WebauthncredentialsRow])
+      parsed <- parseFirst(r, id)
+    } yield parsed
+
+  def insertOne(c: WebAuthnCredential): Fox[Unit] =
     for {
       _ <- run(q"""INSERT INTO webknossos.webauthnCredentials(_id, _multiUser,
                                                               name, publicKeyCose,
@@ -40,8 +71,9 @@ class WebAuthnCredentialDAO @Inject()(sqlClient: SqlClient) (implicit ec: Execut
                               ${c.publicKeyCose}, ${c.signatureCount})""".asUpdate)
     } yield ()
 
-  def removeById(id: ObjectId)(implicit ctx: DBAccessContext): Fox[Unit] =
+  def removeById(id: ObjectId): Fox[Unit] =
     for {
       _ <- run(q"""DELETE FROM webknossos.webauthnCredentials WHERE _id = ${id}""".asUpdate)
     } yield()
+
 }
--- a/app/security/WebAuthnCredentialRepository.scala
+++ b/app/security/WebAuthnCredentialRepository.scala
@@ -1,0 +1,71 @@
+package security
+
+import models.user.{MultiUserDAO,WebAuthnCredentialDAO};
+
+import com.yubico.webauthn._
+import com.yubico.webauthn.data._
+import com.scalableminds.util.accesscontext.GlobalAccessContext
+import com.scalableminds.util.objectid.ObjectId
+
+import java.util.Optional
+import javax.inject.Inject
+import scala.collection.JavaConverters._;
+
+object WebAuthnCredentialRepository {
+  def objectIdToByteArray(id: ObjectId): ByteArray =
+    new ByteArray(id.toString.getBytes())
+
+  def byteArrayToObjectId(arr: ByteArray): ObjectId =
+    new ObjectId(new String(arr.getBytes))
+}
+
+/*
+ * UserHandle => ObjectId
+ * Username => User's E-Mail address
+ */
+
+class WebAuthnCredentialRepository @Inject()(multiUserDAO: MultiUserDAO, webAuthnCredentialDAO: WebAuthnCredentialDAO) extends CredentialRepository {
+  def getCredentialIdsForUsername(email: String): java.util.Set[PublicKeyCredentialDescriptor] = {
+    val user = multiUserDAO.findOneByEmail(email)(GlobalAccessContext).get("Java interop")
+    val keys = webAuthnCredentialDAO.findAllForUser(user._id)(GlobalAccessContext).get("Java interop");
+    keys.map(key => {
+      PublicKeyCredentialDescriptor.builder()
+        .id(WebAuthnCredentialRepository.objectIdToByteArray(key._id))
+        .build()
+    }).to(Set).asJava
+  }
+
+  def getUserHandleForUsername(email: String): Optional[ByteArray] = {
+    val user = multiUserDAO.findOneByEmail(email)(GlobalAccessContext).get("Java interop")
+    Optional.ofNullable(WebAuthnCredentialRepository.objectIdToByteArray(user._id))
+  }
+
+  def getUsernameForUserHandle(handle: ByteArray): Optional[String] = {
+    val id = WebAuthnCredentialRepository.byteArrayToObjectId(handle)
+    val user = multiUserDAO.findOneById(id)(GlobalAccessContext).get("Java interop")
+    Optional.ofNullable(user.email)
+  }
+
+  def lookup(credentialId: ByteArray, userHandle: ByteArray): Optional[RegisteredCredential] = {
+    val credId = WebAuthnCredentialRepository.byteArrayToObjectId(credentialId)
+    val userId = WebAuthnCredentialRepository.byteArrayToObjectId(userHandle)
+    val credential = webAuthnCredentialDAO.findByIdAndUserId(credId, userId)(GlobalAccessContext).get("Java interop");
+    Optional.ofNullable(RegisteredCredential.builder()
+      .credentialId(WebAuthnCredentialRepository.objectIdToByteArray(credential._id))
+      .userHandle(WebAuthnCredentialRepository.objectIdToByteArray(credential._multiUser))
+      .publicKeyCose(new ByteArray(credential.publicKeyCose))
+      .signatureCount(credential.signatureCount)
+      .build())
+  }
+
+  def lookupAll(credentialId: ByteArray): java.util.Set[RegisteredCredential] = {
+    val credential = webAuthnCredentialDAO.findById(WebAuthnCredentialRepository.byteArrayToObjectId(credentialId))(GlobalAccessContext).get("Java interop");
+    Set(RegisteredCredential.builder()
+      .credentialId(WebAuthnCredentialRepository.objectIdToByteArray(credential._id))
+      .userHandle(WebAuthnCredentialRepository.objectIdToByteArray(credential._multiUser))
+      .publicKeyCose(new ByteArray(credential.publicKeyCose))
+      .signatureCount(credential.signatureCount)
+      .build()).asJava
+  }
+
+}
--- a/conf/evolutions/126-add-webauthn-credentials.sql
+++ b/conf/evolutions/126-add-webauthn-credentials.sql
@@ -1,0 +1,18 @@
+START TRANSACTION;
+
+do $$ begin ASSERT (select schemaVersion from webknossos.releaseInformation) = 126, 'Previous schema version mismatch'; end; $$ LANGUAGE plpgsql;
+
+CREATE TABLE webknossos.webauthnCredentials(
+  _id INTEGER PRIMARY KEY,
+  _multiUser CHAR(24) NOT NULL,
+  name TEXT NOT NULL,
+  publicKeyCode BYTEA NOT NULL,
+  signatureCount INTEGER
+);
+
+ALTER TABLE webknossos.webauthnCredentials
+  ADD FOREIGN KEY (_multiUser) REFERENCES webknossos.multiUsers(_id) ON DELETE CASCADE ON UPDATE CASCADE DEFERRABLE;
+
+UPDATE webknossos.releaseInformation SET schemaVersion = 127;
+
+COMMIT TRANSACTION;
--- a/conf/evolutions/126-add-webauthn-credentials.sql
+++ b/conf/evolutions/126-add-webauthn-credentials.sql
@@ -3,11 +3,11 @@ START TRANSACTION;
 do $$ begin ASSERT (select schemaVersion from webknossos.releaseInformation) = 126, 'Previous schema version mismatch'; end; $$ LANGUAGE plpgsql;
 
 CREATE TABLE webknossos.webauthnCredentials(
-  _id INTEGER PRIMARY KEY,
+  _id TEXT PRIMARY KEY,
   _multiUser CHAR(24) NOT NULL,
   name TEXT NOT NULL,
   publicKeyCode BYTEA NOT NULL,
-  signatureCount INTEGER
+  signatureCount INTEGER NOT NULL
 );
 
 ALTER TABLE webknossos.webauthnCredentials
--- a/conf/evolutions/126-add-webauthn-credentials.sql
+++ b/conf/evolutions/126-add-webauthn-credentials.sql
@@ -7,7 +7,8 @@ CREATE TABLE webknossos.webauthnCredentials(
   _multiUser CHAR(24) NOT NULL,
   name TEXT NOT NULL,
   publicKeyCode BYTEA NOT NULL,
-  signatureCount INTEGER NOT NULL
+  signatureCount INTEGER NOT NULL,
+  isDeleted BOOLEAN NOT NULL DEFAULT false
 );
 
 ALTER TABLE webknossos.webauthnCredentials
--- a/conf/webknossos.latest.routes
+++ b/conf/webknossos.latest.routes
@@ -36,6 +36,14 @@ POST          /auth/resetPassword                                               
 GET           /auth/logout                                                                      controllers.AuthenticationController.logout()
 GET           /auth/sso                                                                         controllers.AuthenticationController.singleSignOn(sso: String, sig: String)
 GET           /auth/oidc/login                                                                  controllers.AuthenticationController.loginViaOpenIdConnect()
+
+# Routes for WebAuthn
+POST          /auth/webauthn/auth/start                                                         controllers.AuthenticationController.webauthnAuthStart()
+POST          /auth/webauthn/auth/finalize                                                      controllers.AuthenticationController.webauthnAuthFinalize()
+
+POST          /auth/webauthn/register/start                                                    controllers.AuthenticationController.webauthnAuthStart()
+POST          /auth/webauthn/register/finalize                                                 controllers.AuthenticationController.webauthnAuthFinalize()
+
 # /auth/oidc/callback route is used literally in code
 GET           /auth/oidc/callback                                                               controllers.AuthenticationController.openIdCallback()
 POST          /auth/createOrganizationWithAdmin                                                 controllers.AuthenticationController.createOrganizationWithAdmin()
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -104,7 +104,9 @@ object Dependencies {
     // SQL Queries class generation. Started with runner as slick.codegen.SourceCodeGenerator
     "com.typesafe.slick" %% "slick-codegen" % "3.5.1",
     // SQL Queries postgres specifics. not imported.
-    "org.postgresql" % "postgresql" % "42.7.3"
+    "org.postgresql" % "postgresql" % "42.7.3",
+    /// WebAuthn Dependencies
+    "com.yubico" % "webauthn-server-core" % "2.6.0",
   )
 
   val dependencyOverrides: Seq[ModuleID] = Seq(
--- a/tools/postgres/schema.sql
+++ b/tools/postgres/schema.sql
@@ -419,11 +419,11 @@ CREATE TABLE webknossos.multiUsers(
 );
 
 CREATE TABLE webknossos.webauthnCredentials(
-  _id INTEGER PRIMARY KEY,
+  _id TEXT PRIMARY KEY,
   _multiUser CHAR(24) NOT NULL,
   name TEXT NOT NULL,
   publicKeyCose BYTEA NOT NULL,
-  signatureCount INTEGER
+  signatureCount INTEGER NOT NULL
 );
 
 
--- a/tools/postgres/schema.sql
+++ b/tools/postgres/schema.sql
@@ -423,7 +423,8 @@ CREATE TABLE webknossos.webauthnCredentials(
   _multiUser CHAR(24) NOT NULL,
   name TEXT NOT NULL,
   publicKeyCose BYTEA NOT NULL,
-  signatureCount INTEGER NOT NULL
+  signatureCount INTEGER NOT NULL,
+  isDeleted BOOLEAN NOT NULL DEFAULT false
 );
 
 
--- a/tools/postgres/schema.sql
+++ b/tools/postgres/schema.sql
@@ -418,6 +418,14 @@ CREATE TABLE webknossos.multiUsers(
   CONSTRAINT nuxInfoIsJsonObject CHECK(jsonb_typeof(novelUserExperienceInfos) = 'object')
 );
 
+CREATE TABLE webknossos.webauthnCredentials(
+  _id INTEGER PRIMARY KEY,
+  _multiUser CHAR(24) NOT NULL,
+  name TEXT NOT NULL,
+  publicKeyCose BYTEA NOT NULL,
+  signatureCount INTEGER
+);
+
 
 CREATE TYPE webknossos.TOKEN_TYPES AS ENUM ('Authentication', 'DataStore', 'ResetPassword');
 CREATE TYPE webknossos.USER_LOGININFO_PROVDERIDS AS ENUM ('credentials');
@@ -871,6 +879,8 @@ ALTER TABLE webknossos.aiInferences
 ALTER TABLE webknossos.aiModel_trainingAnnotations
   ADD FOREIGN KEY (_aiModel) REFERENCES webknossos.aiModels(_id) ON DELETE CASCADE ON UPDATE CASCADE DEFERRABLE,
   ADD FOREIGN KEY (_annotation) REFERENCES webknossos.annotations(_id) ON DELETE CASCADE ON UPDATE CASCADE DEFERRABLE;
+ALTER TABLE webknossos.webauthnCredentials
+  ADD FOREIGN KEY (_multiUser) REFERENCES webknossos.multiUsers(_id) ON DELETE CASCADE ON UPDATE CASCADE DEFERRABLE;
 
 
 CREATE FUNCTION webknossos.countsAsTaskInstance(a webknossos.annotations) RETURNS BOOLEAN AS $$
@@ -943,4 +953,3 @@ $$ LANGUAGE plpgsql;
 CREATE TRIGGER onDeleteAnnotationTrigger
 AFTER DELETE ON webknossos.annotations
 FOR EACH ROW EXECUTE PROCEDURE webknossos.onDeleteAnnotation();
-
--- a/app/models/user/MultiUser.scala
+++ b/app/models/user/MultiUser.scala
@@ -136,6 +136,14 @@ class MultiUserDAO @Inject()(sqlClient: SqlClient)(implicit ec: ExecutionContext
                      (SELECT _id FROM webknossos.users WHERE _organization = $organizationId)""".asUpdate)
     } yield ()
 
+  def findOneById(id: ObjectId)(implicit ctx: DBAccessContext): Fox[MultiUser] =
+    for {
+      accessQuery <- readAccessQuery
+      r <- run(q"SELECT $columns FROM $existingCollectionName WHERE _id = $id AND $accessQuery".as[MultiusersRow])
+      parsed <- parseFirst(r, id)
+    } yield parsed
+
+
   def findOneByEmail(email: String)(implicit ctx: DBAccessContext): Fox[MultiUser] =
     for {
       accessQuery <- readAccessQuery


app/security/WebAuthnCredentialRepository.scala
```diff
@@ -0,0 +1,122 @@
...
+import com.scalableminds.util.objectid.ObjectId
+import com.typesafe.scalalogging.LazyLogging
+import net.liftweb.common.{Empty, Failure, Full}
+
+import java.util.Optional
+import javax.inject.Inject
+import scala.annotation.nowarn
+import scala.jdk.CollectionConverters._
+
+object WebAuthnCredentialRepository {
```
MichaelBuessemeyer, line 16: Please remove unnecessary brackets and semicolons in this file as well :see_no_evil:

app/security/WebAuthnCredentialRepository.scala
```diff
@@ -0,0 +1,122 @@
...
+@nowarn("cat=deprecation")
+class WebAuthnCredentialRepository @Inject()(multiUserDAO: MultiUserDAO, webAuthnCredentialDAO: WebAuthnCredentialDAO)
+    extends CredentialRepository
+    with LazyLogging {
+  def getCredentialIdsForUsername(email: String): java.util.Set[PublicKeyCredentialDescriptor] = {
+    val result = for {
+      user <- multiUserDAO.findOneByEmail(email)(GlobalAccessContext).await("Java interop")
+      keys <- webAuthnCredentialDAO.findAllForUser(user._id)(GlobalAccessContext).await("Java interop")
+      creds = keys.map(key => {
+        PublicKeyCredentialDescriptor.builder().id(WebAuthnCredentialRepository.bytesToByteArray(key.keyId)).build()
```
MichaelBuessemeyer, line 38: Something like `this.bytesToByteArray` doesnt work? or maybe even without the this? :thinking:?

MichaelBuessemeyer: Just tested: the without this is working fine if you import this file itself :see_no_evil:. Try using `import security.WebAuthnCredentialRepository.bytesToByteArray`

MichaelBuessemeyer: Same please for `objectIdToByteArray` and the other used functions :pray:

app/security/WebAuthnCredentialRepository.scala
```diff
@@ -0,0 +1,122 @@
...
+    multiUserDAO.findOneByEmail(email)(GlobalAccessContext).await("Java interop") match {
+      case Full(user) => Optional.ofNullable(WebAuthnCredentialRepository.objectIdToByteArray(user._id))
+      case Empty      => Optional.empty()
+      case Failure(msg, _, _) => {
+        logger.error(msg);
+        Optional.empty();
+      }
+    }
+
+  def getUsernameForUserHandle(handle: ByteArray): Optional[String] = {
```
MichaelBuessemeyer, line 61: Can you please maybe add a comment that the handle is the encoded multiuser id? and needs to be named this way due to extending CredentialRepository?

app/security/WebAuthnCredentialRepository.scala
```diff
@@ -0,0 +1,122 @@
...
+    val userId = WebAuthnCredentialRepository.byteArrayToObjectId(userHandle)
+    val result = for {
+      credential <- webAuthnCredentialDAO
+        .findByKeyIdAndUserId(credId, userId)(GlobalAccessContext)
+        .await("Java interop")
+      registered = RegisteredCredential
+        .builder()
+        .credentialId(WebAuthnCredentialRepository.bytesToByteArray(credential.keyId))
+        .userHandle(WebAuthnCredentialRepository.objectIdToByteArray(credential._multiUser))
+        .publicKeyCose(new ByteArray((credential.publicKeyCose)))
```
MichaelBuessemeyer, line 84: There are too many parantheses here:
```suggestion
        .publicKeyCose(new ByteArray(credential.publicKeyCose))
```

conf/evolutions/reversions/126-add-webauthn-credentials.sql
```diff
@@ -0,0 +1,10 @@
+START TRANSACTION;
+
+-- This reversion might take a while because it needs to search in all annotation layer names for '$' and replace it with ''
+do $$ begin ASSERT (select schemaVersion from webknossos.releaseInformation) = 126, 'Previous schema version mismatch'; end; $$ LANGUAGE plpgsql;
+
+DROP TABLE webknossos.webauthnCredentials;
+
+UPDATE webknossos.releaseInformation SET schemaVersion = 125;
+
```
MichaelBuessemeyer, line 9: Just a reminder: the version numbers here need to be bumped up as the master schema version is currently 128 and maybe even higher when this pr is merged

conf/evolutions/126-add-webauthn-credentials.sql
```diff
@@ -0,0 +1,21 @@
+START TRANSACTION;
+
+do $$ begin ASSERT (select schemaVersion from webknossos.releaseInformation) = 125, 'Previous schema version mismatch'; end; $$ LANGUAGE plpgsql;
```
MichaelBuessemeyer, line 3: same here for the version numbers

app/controllers/AuthenticationController.scala
```diff
@@ -406,6 +445,133 @@ class AuthenticationController @Inject()(
...
+  }
+
+  def webauthnRemoveKey: Action[WebAuthnKeyDescriptor] = sil.SecuredAction.async(validateJson[WebAuthnKeyDescriptor]) {
+    implicit request =>
+      {
+        for {
+          _ <- webAuthnCredentialDAO.removeById(request.body.id, request.identity._multiUser)
+        } yield Ok(Json.obj())
+      }
+  }
```
MichaelBuessemeyer, line 573: It is totally fine to pass the entry that should be deleted in the request body. But I think usually when we send only an ID, this is done via url params. But as your approach is totally fine, feel free to choose one. Just wanted to leave that thought here :D

tools/gen-ssl-dev-certs.sh
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env bash
+
+mkdir -p target/
+openssl req -x509 -newkey rsa:2048 \
+  -keyout target/dev.key.pem \
+  -out target/dev.cert.pem \
+  -sha256 -days 365 \
+  -nodes \
+  -subj "/C=XX/ST=Dev/O=Dev/OU=Dev/CN=webknossos.local"
```
MichaelBuessemeyer, line 9: :thinking: As this is used for the proxy anyway and there is a tools/proxy dir, maybe move that file here?

app/security/WebAuthnCredentialRepository.scala
```diff
@@ -0,0 +1,122 @@
...
+ */
+
+@nowarn("cat=deprecation")
+class WebAuthnCredentialRepository @Inject()(multiUserDAO: MultiUserDAO, webAuthnCredentialDAO: WebAuthnCredentialDAO)
+    extends CredentialRepository
+    with LazyLogging {
+  def getCredentialIdsForUsername(email: String): java.util.Set[PublicKeyCredentialDescriptor] = {
+    val result = for {
+      user <- multiUserDAO.findOneByEmail(email)(GlobalAccessContext).await("Java interop")
+      keys <- webAuthnCredentialDAO.findAllForUser(user._id)(GlobalAccessContext).await("Java interop")
```
fm3, line 36: Any chance of a way to use an asynchronous variant of `getCredentialIdsForUsername` with the webauthn libraries? Maybe you could have another look if something exists there. It would be great if we could avoid the await. Currently we have none of them outside of tests, and it would certainly be a step back if we would have to add them here.

robert-oleynik: I can try https://github.com/webauthn4j/webauthn4j. I do not see any Java interface forcing awaits there.

tools/postgres/schema.sql
```diff
@@ -418,6 +418,17 @@ CREATE TABLE webknossos.multiUsers(
   CONSTRAINT nuxInfoIsJsonObject CHECK(jsonb_typeof(novelUserExperienceInfos) = 'object')
 );
 
+CREATE TABLE webknossos.webauthnCredentials(
```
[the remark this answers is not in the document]
MichaelBuessemeyer: thats still todo as far as I can see

MIGRATIONS.unreleased.md
```diff
@@ -9,4 +9,6 @@ User-facing changes are documented in the [changelog](CHANGELOG.released.md).
 [Commits](https://github.com/scalableminds/webknossos/compare/25.06.1...HEAD)
 
 ### Postgres Evolutions:
+
 - [134-dataset-layer-attachments.sql](conf/evolutions/134-dataset-layer-attachments.sql)
+- [135-add-webauthn-credentials.sql](./conf/evolutions/131-add-webauthn-credentials.sql)
```
MichaelBuessemeyer: the path here also needs to be updated

app/controllers/AuthenticationController.scala
```diff
@@ -29,8 +43,140 @@ import utils.WkConf
...
+  * - `attestation` and `attestationFormat`, because attestation is not implemented.
+  * - `extensions` no extensions in use.
+  */
+case class WebAuthnPublicKeyCredentialCreationOptions(
+    authenticatorSelection: WebAuthnCreationOptionsAuthenticatorSelection,
+    attestation: String = "none",
+    challenge: String,
+    excludeCredentials: Array[WebAuthnCreationOptionsExcludeCredentials],
+    pubKeyCredParams: Array[WebAuthnCreationOptionsPubKeyParam],
+    timeout: Int, // NOTE: timeout in milliseconds
```
MichaelBuessemeyer: For all comments starting with `NOTE:` please remove the `NOTE:` part. The sole reason is that this is uncomment and usually the we do not have a "bycomment" like NOTE in our comments.

app/controllers/AuthenticationController.scala
```diff
@@ -29,8 +43,140 @@ import utils.WkConf
...
 import java.security.MessageDigest
+import java.security.SecureRandom
+import java.util.UUID
 import javax.inject.Inject
+import scala.concurrent.duration.DurationInt
 import scala.concurrent.{ExecutionContext, Future}
+import scala.jdk.CollectionConverters._
+
+/**
+  * Object reference: https://developer.mozilla.org/en-US/docs/Web/API/PublicKeyCredentialCreationOptions
```
MichaelBuessemeyer, line 54: Thanks for adding all the doc references. They might become very helpful in the future :pray:

app/controllers/AuthenticationController.scala
```diff
@@ -405,6 +571,159 @@ class AuthenticationController @Inject()(
...
   }
 
+  def webauthnAuthStart(): Action[AnyContent] = Action {
+    val sessionId = UUID.randomUUID().toString
+    val cookie = Cookie("webauthn-session", sessionId, maxAge = Some(120), httpOnly = true, secure = true)
+    val challenge = new Array[Byte](32) // NOTE: Minimum required length are 16 bytes.
+    secureRandom.nextBytes(challenge)
+    val assertion = WebAuthnPublicKeyCredentialRequestOptions(
+      challenge = Base64.encodeBase64URLSafeString(challenge),
+      timeout = Some(120000),
```
MichaelBuessemeyer: The 120000 and 120 timeout correlate with each other, correct? One is in seconds and the other in milliseconds. Could you please extract the 120 seconds into a constant and if you take a look at [this](https://www.scala-lang.org/api/3.1.3/scala/concurrent/duration/Duration.html): I think you can even do something like:
```
val timeout: FiniteDuration = 120.seconds
val timeoutInSeconds: Int = timeout.toSeconds.toInt
val timeoutInMilliseconds: Int = timeout.toMillis.toInt
```
but you need to import `scala.concurrent.duration._ ` for that.

app/controllers/AuthenticationController.scala
```diff
@@ -405,6 +571,159 @@ class AuthenticationController @Inject()(
...
+      rpId = Some(origin.getHost),
+      userVerification = Some("preferred"),
+      hints = None
+    )
+    temporaryAssertionStore.insert(sessionId, WebAuthnChallenge(challenge), Some(2 minutes))
```
MichaelBuessemeyer: oh and here you have the same timeout again. This could be combined with the comment above by just unsing the `timeout` val here

app/controllers/AuthenticationController.scala
```diff
@@ -405,6 +571,159 @@ class AuthenticationController @Inject()(
...
+  def webauthnAuthFinalize(): Action[WebAuthnAuthentication] = Action.async(validateJson[WebAuthnAuthentication]) {
+    implicit request =>
+      for {
+        cookie <- request.cookies.get("webauthn-session").toFox
+        sessionId = cookie.value
+        challenge <- {
+          val challenge = temporaryAssertionStore.get(sessionId)
+          temporaryAssertionStore.remove(sessionId)
+          challenge.toFox
+        }
```
MichaelBuessemeyer: We usually do not have such nested blocks. Could you please put this at the same level as the sessionId extraction form the cookie?
```
challengeOpt = temporaryAssertionStore.get(sessionId)
_ = temporaryAssertionStore.remove(sessionId)
challenge <- challengeOpt.toFox ?~> "some error message todo"
```

app/controllers/AuthenticationController.scala
```diff
@@ -405,6 +571,159 @@ class AuthenticationController @Inject()(
...
+          credential.credentialRecord,
+          null,
+          false, // NOTE: User verification is not required put preferred.
+          false // NOTE: User presence is not required.
+        )
+        _ <- tryo(webAuthnManager.verify(authData, params)).toFox ?~> "Passkey Authentication Failed" ~> UNAUTHORIZED
+        _ = credential.credentialRecord.setCounter(authData.getAuthenticatorData.getSignCount)
+        _ <- webAuthnCredentialDAO.updateSignCount(credential)
+
+        // TODO: Validate
```
MichaelBuessemeyer: Is this still todo?

app/controllers/AuthenticationController.scala
```diff
@@ -29,8 +43,140 @@ import utils.WkConf
...
+    rpId: Option[String] = None, // Relying party ID
+    userVerification: Option[String] = Some("preferred"), // "required", "preferred", "discouraged"
+    hints: Option[Seq[String]] = None // UI hints for the user-agent
+)
+object WebAuthnPublicKeyCredentialRequestOptions {
+  implicit val jsonFormat: OFormat[WebAuthnPublicKeyCredentialRequestOptions] =
+    Json.format[WebAuthnPublicKeyCredentialRequestOptions]
+}
+
+case class WebAuthnRegistration(name: String, key: JsValue)
```
MichaelBuessemeyer, line 170: It this internal stuff not part of the standart? In case it is part of the standart, could you please maybe also add a docs link :thinking:?

Same for the 2 more case classes below :v:

app/controllers/AuthenticationController.scala
MichaelBuessemeyer: or alternatively you make this a very small private function

app/controllers/AuthenticationController.scala
```diff
@@ -405,6 +571,159 @@ class AuthenticationController @Inject()(
...
+  def webauthnRegisterStart(): Action[AnyContent] = sil.SecuredAction.async { implicit request =>
+    for {
+      email <- userService.emailFor(request.identity)
+      user = WebAuthnCreationOptionsUser(
+        displayName = request.identity.name,
+        id = Base64.encodeBase64URLSafeString(email.getBytes), // NOTE: Not leak database id, but use email instead
+        name = email
+      )
+      credentials <- webAuthnCredentialDAO
+        .findAllForUser(request.identity._multiUser) ?~> "Failed to fetch Passkeys" ~> INTERNAL_SERVER_ERROR
```
MichaelBuessemeyer, line 663: `INTERNAL_SERVER_ERROR` seems a little harsh to me. More like 400 or 404 would be appropriate imo.

app/controllers/AuthenticationController.scala
```diff
@@ -405,6 +571,159 @@ class AuthenticationController @Inject()(
...
+          ))
+        .toArray
+      challenge = {
+        val challenge = new Array[Byte](32) // NOTE: Minimum required length are 16 bytes.
+        secureRandom.nextBytes(challenge)
+        challenge
+      }
+      encodedChallenge = Base64.encodeBase64URLSafeString(challenge)
+      sessionId = UUID.randomUUID().toString
+      cookie = Cookie("webauthn-registration", sessionId, maxAge = Some(120), httpOnly = true, secure = true)
```
MichaelBuessemeyer: Same here. The timeout variable could also be used here. Maybe define these as outside class constants or so?

app/controllers/AuthenticationController.scala
```diff
@@ -405,6 +571,159 @@ class AuthenticationController @Inject()(
...
+    sil.SecuredAction.async(validateJson[WebAuthnRegistration]) { implicit request =>
+      for {
+        registrationData <- tryo(webAuthnManager.parseRegistrationResponseJSON(Json.stringify(request.body.key))).toFox
+        cookie <- request.cookies.get("webauthn-registration").toFox
+        sessionId = cookie.value
+        challenge <- {
+          val challenge = temporaryRegistrationStore.get(sessionId)
+          temporaryRegistrationStore.remove(sessionId)
+          challenge.toFox
+        }
```
MichaelBuessemeyer: Same here. maybe a method would be useful. Or maybe even better: Implement a `pop` function on the class of `temporaryRegistrationStore`

app/controllers/AuthenticationController.scala
```diff
@@ -405,6 +571,159 @@ class AuthenticationController @Inject()(
...
+        serverProperty = new ServerProperty(origin, origin.getHost, challenge)
+        publicKeyParams = webAuthnPubKeyParams.map(k =>
+          new PublicKeyCredentialParameters(PublicKeyCredentialType.PUBLIC_KEY, COSEAlgorithmIdentifier.create(k.alg)))
+        registrationParams = new RegistrationParameters(serverProperty, publicKeyParams.toList.asJava, false, true)
+        _ <- tryo(webAuthnManager.verify(registrationData, registrationParams)).toFox
+        attestation = registrationData.getAttestationObject
+        credentialRecord = new CredentialRecordImpl( // TODO: Rename or minimal custom implementation
```
MichaelBuessemeyer: is this still todo?

app/controllers/AuthenticationController.scala
MichaelBuessemeyer: Alhtough, the logged in user should be able to see their webauthn infos. So maybe it is also alrighty

app/controllers/AuthenticationController.scala
```diff
@@ -405,6 +571,159 @@ class AuthenticationController @Inject()(
...
+        params = new AuthenticationParameters(
+          serverProperty,
+          credential.credentialRecord,
+          null,
+          false, // NOTE: User verification is not required put preferred.
+          false // NOTE: User presence is not required.
+        )
+        _ <- tryo(webAuthnManager.verify(authData, params)).toFox ?~> "Passkey Authentication Failed" ~> UNAUTHORIZED
+        _ = credential.credentialRecord.setCounter(authData.getAuthenticatorData.getSignCount)
+        _ <- webAuthnCredentialDAO.updateSignCount(credential)
```
MichaelBuessemeyer: Was the counting requested by the issue or so to track how much this feature is being used? Or what is the purpose of this?

conf/evolutions/135-add-webauthn-credentials.sql
```diff
@@ -0,0 +1,22 @@
+START TRANSACTION;
+
+do $$ begin ASSERT (select schemaVersion from webknossos.releaseInformation) = 134, 'Previous schema version mismatch'; end; $$ LANGUAGE plpgsql;
+
+CREATE TABLE webknossos.webauthnCredentials(
+  _id TEXT PRIMARY KEY,
+  _multiUser CHAR(24) NOT NULL,
```
MichaelBuessemeyer: A multiuser id is now saved as `TEXT` as well if I am not wrong. Could you please use this type here as well? It is the recommended type according to postgres

conf/evolutions/135-add-webauthn-credentials.sql
```diff
@@ -0,0 +1,22 @@
...
+  serializedAttestedCredential BYTEA NOT NULL,
+  serializedExtensions TEXT NOT NULL,
+  signatureCount INTEGER NOT NULL,
+  isDeleted BOOLEAN NOT NULL DEFAULT false,
+  UNIQUE (_id, credentialId)
+);
+
+ALTER TABLE webknossos.webauthnCredentials
+  ADD FOREIGN KEY (_multiUser) REFERENCES webknossos.multiUsers(_id) ON DELETE CASCADE ON UPDATE CASCADE DEFERRABLE;
+
```
MichaelBuessemeyer: usually for the most tables we also have `<table_name>_` view which list all entries except for the deleted once. Could you please add this view here as well?
```
CREATE VIEW webknossos.webauthnCredentials_ AS SELECT * FROM webknossos.multiUsers WHERE NOT isDeleted;
```

conf/evolutions/reversions/135-add-webauthn-credentials.sql
```diff
@@ -0,0 +1,10 @@
+START TRANSACTION;
+
+-- This reversion might take a while because it needs to search in all annotation layer names for '$' and replace it with ''
+do $$ begin ASSERT (select schemaVersion from webknossos.releaseInformation) = 135, 'Previous schema version mismatch'; end; $$ LANGUAGE plpgsql;
+
+DROP TABLE webknossos.webauthnCredentials;
+
```
MichaelBuessemeyer: same here before dropping `webauthnCredentials`  `webauthnCredentials_` needs to be dropped

frontend/javascripts/admin/api/certificate_validation.ts
MichaelBuessemeyer, line 1: Oh thanks for the renaming / fixing the file type ending :pray:

frontend/javascripts/admin/auth/change_password_view.tsx
```diff
@@ -16,8 +19,19 @@ type Props = {
 const MIN_PASSWORD_LENGTH = 8;
 
 function ChangePasswordView({ history }: Props) {
+  // Password Form
   const [form] = Form.useForm();
 
+  /// Passkeys
+  const [isPasskeyNameModalOpen, setIsPasskeyNameModalOpen] = useState(false);
+  const [newPasskeyName, setNewPasskeyName] = useState("");
+  const [passkeys, setPasskeys] = useState<WebAuthnKeyDescriptor[]>([]);
```
MichaelBuessemeyer: We have a convinience hook for this called `useFetch`. Could you please this here as well?

MichaelBuessemeyer: lthough on the second thought as you manually retrigger the passkeys after a registration, this would be harder with the `useFetch` so better keep your version I'd say

frontend/javascripts/admin/auth/change_password_view.tsx
```diff
@@ -44,117 +58,195 @@ function ChangePasswordView({ history }: Props) {
...
+  async function fetchPasskeys(): Promise<void> {
+    setIsLoadingPasskeys(true);
+    const keys = await listWebAuthnKeys();
+    setPasskeys(keys);
+    setIsLoadingPasskeys(false);
+  }
+
+  const registerNewPasskey = async () => {
+    const passkeyName = newPasskeyName.trim();
+    if (!passkeyName.trim()) {
```
MichaelBuessemeyer: IMO it would be nice to have a min length of 3 or so :thinking:. This is just an idea though

frontend/javascripts/admin/auth/change_password_view.tsx
```diff
@@ -44,117 +58,195 @@ function ChangePasswordView({ history }: Props) {
...
-            />
-          </FormItem>
-          <FormItem>
-            <Button
-              type="primary"
-              htmlType="submit"
-              style={{
-                width: "100%",
-              }}
+    <div>
```
MichaelBuessemeyer: you added the whole diff to accomodate the modal at the end, correct? In that case you can simply use a react fragemnt which will not be rendered into the document and only exist in the reacts internal dom structure as a helper wrapper. A shorthand for this is simply `<>` and `</>` so the wrapping `<div>` `</div>` needs to be swapped with `<>` and `</>`

frontend/javascripts/admin/auth/change_password_view.tsx
```diff
@@ -44,117 +58,195 @@ function ChangePasswordView({ history }: Props) {
...
+          padding: 50,
+        }}
+        align="middle"
+      >
+        <Col span={8}>
+          <h3>Your Passkeys</h3>
+          <p>
+            Passkeys are a new web authentication method that allows you to log in without a
+            password in a secured way. Microsoft Hello and Apple FaceID are examples of technologies
+            that can be used as passkeys to log in in WEBKNOSSOS. If you want to add a new passkey
```
MichaelBuessemeyer:
```suggestion
            that can be used as passkeys to log in to WEBKNOSSOS. If you want to add a new passkey
```

frontend/javascripts/admin/auth/change_password_view.tsx
```diff
@@ -44,117 +58,195 @@ function ChangePasswordView({ history }: Props) {
...
+            to your account use the button below.
+          </p>
+
+          {passkeys.map((passkey) => (
```
MichaelBuessemeyer: Sorry, this UI part here is a little rudimentary. Could you please make this a simple AntD table? I'd leave out the header. Moreover, code styling wise I would move the passkey deletion onClick function into its own function to make the rendering code slimmer.

Kinda like this
```
const deletePasskey = async (passkey) => {
                  await removeWebAuthnKey(passkey);
                  await fetchPasskeys();
                };
const columns = [
  {
    title: 'Name',
    dataIndex: 'name',
    key: 'name',
  },
  {
    title: 'Actions',
    dataIndex: 'id',
    key: 'id',
    render: (_id, passKey) => (<Button
                onClick={() => deletePasskey(passKey)}
              >
                Delete
              </Button>)
  }
];

<Table dataSource={passkey} columns={columns} showHeader={false} />;
```

frontend/javascripts/admin/auth/change_password_view.tsx
```diff
@@ -44,117 +58,195 @@ function ChangePasswordView({ history }: Props) {
...
+              Register Passkey
             </Button>
-          </FormItem>
-        </Form>
-      </Col>
-    </Row>
+          </div>
+        </Col>
+      </Row>
+      <Modal
```
MichaelBuessemeyer: TODO:
- [x] reminder for me: Check during testing whether the theme is respected by this modal as this was a problem in the past quite often

frontend/javascripts/admin/auth/login_form.tsx
```diff
@@ -139,6 +139,27 @@ function LoginForm({ layout, onLoggedIn, hideFooter, style }: Props) {
...
+              onClick={async () => {
+                try {
+                  const [user, organization] = await doWebAuthnLogin();
+                  Store.dispatch(setActiveUserAction(user));
+                  Store.dispatch(setActiveOrganizationAction(organization));
+                  if (onLoggedIn) {
+                    onLoggedIn();
+                  }
+                } catch (error) {
+                  console.error("WebAuthn login failed", error);
```
MichaelBuessemeyer: Could you please also show an error toast here of this?

frontend/javascripts/admin/auth/login_form.tsx
```diff
@@ -139,6 +139,27 @@ function LoginForm({ layout, onLoggedIn, hideFooter, style }: Props) {
...
+                }
+              }}
```
MichaelBuessemeyer: And to make the rendering code a little more compact, could you please consider moving this into an own function which can be called here?

frontend/javascripts/admin/rest_api.ts
```diff
@@ -147,6 +149,55 @@ export async function loginUser(formValues: {
   return [activeUser, organization];
 }
 
+export async function doWebAuthnLogin(): Promise<[APIUser, APIOrganization]> {
```
MichaelBuessemeyer: We started to make this module / file smaller by splitting it up into multiple files. Could you please move these new functions into an own file `admin/api/webauthn.ts` or so? That would be greately appreciated. This file already has 2.4k lines on this branch :see_no_evil:

frontend/javascripts/types/api_types.ts
```diff
@@ -1246,3 +1246,8 @@ export type RenderAnimationOptions = {
   movieResolution: MOVIE_RESOLUTIONS;
   cameraPosition: CAMERA_POSITIONS;
 };
+
+export type WebAuthnKeyDescriptor = {
+  id: string;
+  name: string;
+};
```
MichaelBuessemeyer: This can then also be moved into the new `admin/api/webauthn.ts` file (or however you'll name it)

tools/postgres/schema.sql
```diff
@@ -463,6 +463,18 @@ CREATE TABLE webknossos.multiUsers(
...
+  _id TEXT PRIMARY KEY,
+  _multiUser CHAR(24) NOT NULL,
```
MichaelBuessemeyer: Please use `TEXT` here. See my previous comment

app/controllers/AuthenticationController.scala
MichaelBuessemeyer: IMO a new function on the class side of the `temporaryAssertionStore` would even be better. See
https://github.com/scalableminds/webknossos/pull/8393#discussion_r2128357403

app/controllers/AuthenticationController.scala
robert-oleynik: It is internal stuff and I add docs for clarification.

app/controllers/AuthenticationController.scala
robert-oleynik: This error should only occur if the SQL query fails and therefore, I would prefer an internal server error.

app/controllers/AuthenticationController.scala
robert-oleynik: Yes, because there are some semantic conflicts with the suffix Impl. I've renamed the import to avoid confusions.

conf/evolutions/135-add-webauthn-credentials.sql
robert-oleynik: There are no entries with `isDeleted = true` at the moment. Because of that it is not necessary to create such a view imo.

frontend/javascripts/admin/auth/change_password_view.tsx
robert-oleynik: I agree, and we should also check if the key name already exists, but I am not so familiar with the general UX of WK.

conf/evolutions/135-add-webauthn-credentials.sql
fm3: Is it possible to delete webauthnCredentials from the webknossos code? If not, no isDeleted field is required and thus also no view.

If yes, please add both the isDeleted field and the view, and go for deleting by setting the field to true, and reading from the view. That way it would be consistent with the existing codebase.

frontend/javascripts/admin/auth/login_form.tsx
robert-oleynik: A toast containing the server's error message is already shown. This makes it only easier to debug the exceptions.

app/controllers/AuthenticationController.scala
robert-oleynik: See https://w3c.github.io/webauthn/#signature-counter the section starting with "A relying party...".

It may also be related to the validation TODO.

frontend/javascripts/admin/auth/change_password_view.tsx
robert-oleynik: I've done an implementation with useGuardedFetch and also added a counter to retrigger the update. It may be cleaner to have a useFetch which can be retriggered for updates.

app/controllers/AuthenticationController.scala
```diff
@@ -405,6 +582,160 @@ class AuthenticationController @Inject()(
...
+        cookie <- request.cookies.get("webauthn-session").toFox
+        sessionId = cookie.value
+        challenge <- temporaryAssertionStore
+          .pop(sessionId)
+          .toFox ?~> "Timeout during authentication. Please try again." ~> UNAUTHORIZED
+        authData <- tryo(webAuthnManager.parseAuthenticationResponseJSON(Json.stringify(request.body.key))).toFox ?~> "Bad Request" ~> BAD_REQUEST
+        credentialId = authData.getCredentialId
+        multiUserEmail = new String(authData.getUserHandle)
+        multiUser <- multiUserDAO
+          .findOneByEmail(multiUserEmail)(GlobalAccessContext) ?~> "Passkey Authentication Failed" ~> UNAUTHORIZED
```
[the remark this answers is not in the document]
robert-oleynik: multiUserEmail is already decoded by webauthn4j

app/controllers/AuthenticationController.scala
robert-oleynik: I checked, the verify function of `webAuthnManager` and it already validates all fields.

app/controllers/AuthenticationController.scala
```diff
@@ -405,6 +582,172 @@ class AuthenticationController @Inject()(
...
+          serverProperty,
+          credential.credentialRecord,
+          null,
+          false, // User verification is not required put preferred.
+          false // User presence is not required.
+        )
+        _ <- tryo(webAuthnManager.verify(authData, params)).toFox ?~> "Passkey Authentication Failed" ~> UNAUTHORIZED
+        oldSignCount = credential.credentialRecord.getCounter
+        newSignCount = authData.getAuthenticatorData.getSignCount
+        _ = credential.credentialRecord.setCounter(authData.getAuthenticatorData.getSignCount)
```
MichaelBuessemeyer: You could use `newSignCount` here as a parameter.

frontend/javascripts/admin/auth/change_password_view.tsx
```diff
@@ -23,14 +24,15 @@ function ChangePasswordView({ history }: Props) {
...
+  const [updateCounter, setUpdateCounter] = useState(0);
   const [isPasskeyNameModalOpen, setIsPasskeyNameModalOpen] = useState(false);
   const [newPasskeyName, setNewPasskeyName] = useState("");
-  const [passkeys, setPasskeys] = useState<WebAuthnKeyDescriptor[]>([]);
-  const [_isLoadingPasskeys, setIsLoadingPasskeys] = useState(false);
-
-  useEffect(() => {
-    fetchPasskeys();
-  }, []);
+  const [passkeys, _isLoading] = useGuardedFetch(
```
MichaelBuessemeyer: Oh a `useGuardedFetch`. Awesome :tada:

frontend/javascripts/admin/webauthn.ts
MichaelBuessemeyer, line 1: Please put this into `frontend/javascripts/admin/api/webauthn.ts` as it only includes api request. This make the folder structure a little bit cleaner :innocent: 

Sorry, I think I did not give the best suggested path in the previous review :see_no_evil:

frontend/javascripts/admin/auth/login_form.tsx
```diff
@@ -60,6 +45,19 @@ function LoginForm({ layout, onLoggedIn, hideFooter, style }: Props) {
...
+  const webauthnLogin = async () => {
+    try {
+      const [user, organization] = await doWebAuthnLogin();
+      Store.dispatch(setActiveUserAction(user));
+      Store.dispatch(setActiveOrganizationAction(organization));
+      if (onLoggedIn) {
+        onLoggedIn();
+      }
+    } catch (error) {
+      console.error("WebAuthn login failed", error);
```
MichaelBuessemeyer: As this is quite difficult for me to test: Could you please try what happens in case the authentication fails? In that case a toast should appear informing the user about the failed attempt. Maybe this is already done by `doWebAuthnLogin  which uses our request lib. But I would make sure that's the case. Else the failure is silent and the user doesn't notice.

frontend/javascripts/admin/auth/change_password_view.tsx
```diff
@@ -73,16 +73,16 @@ function ChangePasswordView({ history }: Props) {
 
   const registerNewPasskey = async () => {
     const passkeyName = newPasskeyName.trim();
-    if (!passkeyName.trim()) {
-      Toast.error("Passkey name cannot be empty");
+    if (passkeyName.length < 3) {
```
MichaelBuessemeyer: @philippotto What do you think about imposing the same naming restrictions upon the passkey name as for dataset names? IMO the character set should be enough

conf/webknossos.latest.routes
```diff
@@ -43,7 +43,7 @@ POST          /auth/webauthn/auth/finalize
 POST          /auth/webauthn/register/start                                                     controllers.AuthenticationController.webauthnRegisterStart()
 POST          /auth/webauthn/register/finalize                                                  controllers.AuthenticationController.webauthnRegisterFinalize()
 GET           /auth/webauthn/keys                                                               controllers.AuthenticationController.webauthnListKeys()
-DELETE        /auth/webauthn/keys                                                               controllers.AuthenticationController.webauthnRemoveKey()
+DELETE        /auth/webauthn/key/:id                                                            controllers.AuthenticationController.webauthnRemoveKey(id: ObjectId)
```
MichaelBuessemeyer: Please keep the plural her in the URL, as this is how it is done for other requestable entities' URLs
```suggestion
DELETE        /auth/webauthn/keys/:id                                                            controllers.AuthenticationController.webauthnRemoveKey(id: ObjectId)
```

frontend/javascripts/admin/webauthn.ts
```diff
@@ -50,9 +50,6 @@ export async function listWebAuthnKeys(): Promise<Array<WebAuthnKeyDescriptor>>
   return await Request.receiveJSON("/api/auth/webauthn/keys");
 }
 
-export async function removeWebAuthnKey(key: WebAuthnKeyDescriptor): Promise<void> {
-  await Request.sendJSONReceiveArraybuffer("/api/auth/webauthn/keys", {
-    method: "DELETE",
-    data: key,
-  });
+export async function removeWebAuthnKey(id: string): Promise<void> {
+  await Request.sendJSONReceiveJSON(`/api/auth/webauthn/key/${id}`, { method: "DELETE" });
```
MichaelBuessemeyer:
```suggestion
  await Request.sendJSONReceiveJSON(`/api/auth/webauthn/keys/${id}`, { method: "DELETE" });
```

conf/application.conf
```diff
@@ -171,6 +171,8 @@ features {
   optInTabs = ["ConnectomeView"]
   openIdConnectEnabled = false
   segmentAnythingEnabled = false
+  # Enable Authentication with Passkeys and WebAuthn
+  passkeysEnabled = true
```
MichaelBuessemeyer:
```suggestion
passkeysEnabled = false
```

frontend/javascripts/admin/auth/change_password_view.tsx
```diff
@@ -27,16 +27,25 @@ function ChangePasswordView({ history }: Props) {
...
+    listWebAuthnKeys().then((passkeys) => setPasskeys(passkeys));
+  }
+  useEffect(() => {
+    if (!isMounted) {
+      setIsMounted(true);
+      if (passkeysEnabled) {
+        fetchPasskeys();
+      }
+    }
+  });
```
MichaelBuessemeyer: So this is an useEffect but only once after component mount? In that case
```
useEffect(() => {
        fetchPasskeys();
  }, []);
```
should be sufficient. It should only be execute once the component is mounted first and afterwards never again as the dependency array is empty

frontend/javascripts/admin/auth/login_form.tsx
```diff
@@ -43,18 +43,14 @@ function LoginForm({ layout, onLoggedIn, hideFooter, style }: Props) {
...
-        onLoggedIn();
-      }
-    } catch (error) {
-      console.error("WebAuthn login failed", error);
+    const [user, organization] = await doWebAuthnLogin();
+    Store.dispatch(setActiveUserAction(user));
+    Store.dispatch(setActiveOrganizationAction(organization));
+    if (onLoggedIn) {
+      onLoggedIn();
     }
```
MichaelBuessemeyer: I'd keep the error logging. This can be helpful for debugging in case things go wrong

frontend/javascripts/types/api_types.ts
```diff
@@ -732,6 +732,7 @@ export type APIFeatureToggles = {
   readonly optInTabs?: Array<string>;
   readonly openIdConnectEnabled?: boolean;
   readonly segmentAnythingEnabled?: boolean;
+  readonly passkeysEnabled?: boolean;
```
MichaelBuessemeyer: @fm3 / @philippotto do you know why  `optInTabs`, `openIdConnectEnabled` and `segmentAnythingEnabled` are optional? I think they are always included in our `application.conf`, aren't they?
Because I would say that we could make `passkeysEnabled` non optional here in the typing as it is always present in the `application.conf`

frontend/javascripts/admin/auth/change_password_view.tsx
MichaelBuessemeyer: Nvm this is outdated

frontend/javascripts/types/api_types.ts
fm3: I’d say it’s safe to say they are always present. If someone manually removes them using the config, it is ok if things start to break.

MichaelBuessemeyer: Ok thanks for the input. In that case please make `passkeysEnabled` non-optional @robert-oleynik. The other fields `optInTabs`, `openIdConnectEnabled` and `segmentAnythingEnabled` should be changed in this PR. This should be done in another PR

frontend/javascripts/admin/auth/change_password_view.tsx
robert-oleynik: Switched to useEffect without dependencies. It is simpler (no need of a counter) and `_isLoading` is not used anyways.

frontend/javascripts/admin/auth/login_form.tsx
robert-oleynik: How to test:
1. Create a new passkey
2. Remove the passkey from WebKnossos
3. Try to login with the created but removed passkey. (This should fail with a toast).

My results:
- A toast with `Passkey Authentication Failed` is shown.

This toast is based on the error message returned by the server.

conf/evolutions/reversions/136-add-webauthn-credentials.sql
```diff
@@ -0,0 +1,11 @@
+START TRANSACTION;
+
+-- This reversion might take a while because it needs to search in all annotation layer names for '$' and replace it with ''
```
fm3: comment seems outdated

conf/evolutions/reversions/136-add-webauthn-credentials.sql
```diff
@@ -0,0 +1,11 @@
...
+do $$ begin ASSERT (select schemaVersion from webknossos.releaseInformation) = 135, 'Previous schema version mismatch'; end; $$ LANGUAGE plpgsql;
```
fm3:
```suggestion
do $$ begin ASSERT (select schemaVersion from webknossos.releaseInformation) = 136, 'Previous schema version mismatch'; end; $$ LANGUAGE plpgsql;
```

project/Dependencies.scala
```diff
@@ -103,7 +103,13 @@ object Dependencies {
     // SQL Queries class generation. Started with runner as slick.codegen.SourceCodeGenerator
     "com.typesafe.slick" %% "slick-codegen" % slickVersion,
     // SQL Queries postgres specifics. not imported.
-    "org.postgresql" % "postgresql" % "42.7.5"
+    "org.postgresql" % "postgresql" % "42.7.5",
+    /// WebAuthn Dependencies
```
fm3:
```suggestion
    /// WebAuthn for passkey authentication. import com.webauthn4j
```

tools/proxy/proxy.js
```diff
@@ -1,13 +1,18 @@
...
 const prefixLines = require("prefix-stream-lines");
+const https = require("https");
+
+const time5min = 5 * 60 * 1000;
 
 const proxy = httpProxy.createProxyServer({
-  proxyTimeout: 5 * 60 * 1000, // 5 min
-  timeout: 5 * 60 * 1000, // 5 min
+  proxyTimeout: time5min, // 5 min
+  timeout: time5min, // 5 min
```
fm3:
```suggestion
  proxyTimeout: time5min,
  timeout: time5min,
```

robert-oleynik:
# Security Considerations

https://w3c.github.io/webauthn/#sctn-security-considerations-rp

- [ ] Visibility Considerations for Embedded Usage
- [ ] Cryptographic Challenges
- [ ] Attestation Limitations
- [ ] Revoked Attestation Certificates
- [ ] Credential Loss and Key Mobility
- [ ] Unprotected account detection
- [ ] Code injection attacks
- [ ] Validating the origin of a credential

# Privacy Considerations

https://w3c.github.io/webauthn/#sctn-privacy-considerations-rp

- [ ] User Handle Contents
- [ ] Username Enumeration
- [ ] Privacy leak via credential IDs

frontend/javascripts/admin/auth/passkeys_view.tsx
```diff
@@ -0,0 +1,115 @@
...
+    } else if (passkeys.some((pk) => pk.name.toLowerCase() === passkeyName.toLowerCase())) {
+      Toast.error("A passkey with this name already exists");
+      return;
+    }
+    setIsPasskeyNameModalOpen(false);
+    await doWebAuthnRegistration(passkeyName);
+    Toast.success("Passkey registered successfully");
+    setNewPasskeyName("");
+    fetchPasskeys();
+  };
```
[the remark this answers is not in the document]
robert-oleynik: `doWebAuthnRegistration` and `fetchPasskeys` already do user notifications on error. No need for an additional error message.

frontend/javascripts/admin/auth/passkeys_view.tsx
```diff
@@ -0,0 +1,115 @@
...
+    fetchPasskeys();
+  }, []);
+
+  function webauthnRemoveKey(passkey: WebAuthnKeyDescriptor): () => Promise<void> {
+    return async function () {
+      await removeWebAuthnKey(passkey.id);
+      Toast.success("Passkey '" + passkey.name + "' is removed");
+      await fetchPasskeys();
+    };
+  }
```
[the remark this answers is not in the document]
robert-oleynik: `removeWebAuthnKey` and `fetchPasskeys` already do user notifications on error. No need for an additional error message.

frontend/javascripts/admin/auth/change_password_view.tsx
MichaelBuessemeyer: @philippotto stubs :D

project/Dependencies.scala
```diff
@@ -105,7 +105,7 @@ object Dependencies {
     // SQL Queries postgres specifics. not imported.
     "org.postgresql" % "postgresql" % "42.7.5",
     /// WebAuthn Dependencies
-    "com.webauthn4j" % "webauthn4j-core" % "0.29.2.RELEASE" exclude("com.fasterxml.jackson.core", "jackson-databind"),
+    "com.webauthn4j" % "webauthn4j-core" % "0.29.3.RELEASE" exclude("com.fasterxml.jackson.core", "jackson-databind"),
   )
 
   val dependencyOverrides: Seq[ModuleID] = Seq(
```
MichaelBuessemeyer, line 111: @fm3 / @robert-oleynik I am not familiar with what `dependencyOverrides` does, but if I understand the comment correctly, this overwrite is no longer needed as we do no longer have `netlift common` as a dependency. Thus, can you please delete this and test whether everything still works?
``` 
val dependencyOverrides: Seq[ModuleID] = Seq( 
   // liftweb-commons (used by us for Box/tryo) depends on older scala-xml, but we do not use its xml-related features
    "com.fasterxml.jackson.core" % "jackson-databind" % "2.14.3"	    "com.fasterxml.jackson.core" % "jackson-databind" % "2.14.3"
  )	  )
```

frontend/javascripts/admin/auth/passkeys_view.tsx
```diff
@@ -0,0 +1,115 @@
...
+
+  // biome-ignore lint/correctness/useExhaustiveDependencies: run once
+  useEffect(() => {
+    fetchPasskeys();
+  }, []);
+
+  function webauthnRemoveKey(passkey: WebAuthnKeyDescriptor): () => Promise<void> {
+    return async function () {
+      await removeWebAuthnKey(passkey.id);
+      Toast.success("Passkey '" + passkey.name + "' is removed");
```
MichaelBuessemeyer: A little bit picky, but please end the messages with a  `.` at the end. This is not completely consistent in the code base but it would be better to have correct punctuation :innocent: 
```suggestion
      Toast.success("Passkey '" + passkey.name + "' is removed.");
```

frontend/javascripts/admin/auth/passkeys_view.tsx
```diff
@@ -0,0 +1,115 @@
...
+      await fetchPasskeys();
+    };
+  }
+
+  const registerNewPasskey = async () => {
+    const passkeyName = newPasskeyName.trim();
+    if (passkeyName.length < 3) {
+      Toast.error("Passkey name must be at least 3 characters");
```
MichaelBuessemeyer: Same as above
```suggestion
      Toast.error("Passkey name must be at least 3 characters.");
```

frontend/javascripts/admin/auth/passkeys_view.tsx
```diff
@@ -0,0 +1,115 @@
...
+      return;
+    } else if (passkeys.some((pk) => pk.name.toLowerCase() === passkeyName.toLowerCase())) {
+      Toast.error("A passkey with this name already exists");
```
MichaelBuessemeyer: Same as above
```suggestion
     Toast.error("A passkey with this name already exists.");
```

frontend/javascripts/admin/auth/passkeys_view.tsx
```diff
@@ -0,0 +1,115 @@
...
+      return;
+    }
+    setIsPasskeyNameModalOpen(false);
+    await doWebAuthnRegistration(passkeyName);
+    Toast.success("Passkey registered successfully");
```
MichaelBuessemeyer: Same as above
```suggestion
    Toast.success("Passkey registered successfully.");
```

app/controllers/AuthenticationController.scala
```diff
@@ -585,6 +586,7 @@ class AuthenticationController @Inject()(
   def webauthnAuthStart(): Action[AnyContent] = Action.async { implicit request =>
     for {
       _ <- Fox.fromBool(conf.Features.passkeysEnabled) ?~> "Passkeys Disabled"
+      _ <- Fox.fromBool(usesHttps) ?~> "Passkeys are only supported with HTTPS"
```
MichaelBuessemeyer: As this is repeated quite often, please make this an error message in `messages.conf` and reference it here and below

app/controllers/AuthenticationController.scala
```diff
@@ -617,13 +617,13 @@ class AuthenticationController @Inject()(
         challenge <- temporaryAssertionStore
           .pop(sessionId)
           .toFox ?~> "Timeout during authentication. Please try again." ~> UNAUTHORIZED
-        authData <- tryo(webAuthnManager.parseAuthenticationResponseJSON(Json.stringify(request.body.key))).toFox ?~> "Bad Request" ~> BAD_REQUEST
+        authData <- tryo(webAuthnManager.parseAuthenticationResponseJSON(Json.stringify(request.body.key))).toFox ??~> "Passkey Authentication Failed" ~> UNAUTHORIZED
```
MichaelBuessemeyer: Same here with `"Passkey Authentication Failed"`. Please make it a member of `messages.conf` and use it here and below :pray:

util/src/main/scala/com/scalableminds/util/tools/Fox.scala
```diff
@@ -239,6 +239,15 @@ class Fox[+A](val futureBox: Future[Box[A]])(implicit ec: ExecutionContext) {
...
 
+  // Overwrite error message in case of Failure and Empty (wrapping Empty in a Failure).
+  def ??~>(s: String): Fox[A] =
+    Fox.fromFutureBox {
+      futureBox.map {
+        case Full(value) => Full(value)
+        case _ => Failure(s)
+      }
+    }
+
```
MichaelBuessemeyer: Very neat :+1:

app/models/user/WebAuthnCredentials.scala
```diff
@@ -90,13 +128,19 @@ class WebAuthnCredentialDAO @Inject()(sqlClient: SqlClient)(implicit ec: Executi
...
     val serializedAttestedCredential = c.serializeAttestedCredential(converter)
-    val serializedAuthenticatorExtensions = c.serializedExtensions(converter)
     val credentialId = c.credentialRecord.getAttestedCredentialData.getCredentialId
+    val userVerified = c.credentialRecord.isUvInitialized.booleanValue
+    val backupEligible = c.credentialRecord.isBackupEligible.booleanValue
+    val backupState = c.credentialRecord.isBackedUp.booleanValue
     for {
+      serializedAuthenticatorExtensions <- c.serializedExtensions(converter)
+      attestationStatement <- c.serializeAttestationStatement(converter)
+      _ = println(attestationStatement)
```
MichaelBuessemeyer: please remove this print statement before merging

app/models/user/WebAuthnCredentials.scala
```diff
@@ -90,13 +128,19 @@ class WebAuthnCredentialDAO @Inject()(sqlClient: SqlClient)(implicit ec: Executi
...
     val converter = new ObjectConverter()
     val serializedAttestedCredential = c.serializeAttestedCredential(converter)
-    val serializedAuthenticatorExtensions = c.serializedExtensions(converter)
     val credentialId = c.credentialRecord.getAttestedCredentialData.getCredentialId
+    val userVerified = c.credentialRecord.isUvInitialized.booleanValue
+    val backupEligible = c.credentialRecord.isBackupEligible.booleanValue
+    val backupState = c.credentialRecord.isBackedUp.booleanValue
     for {
+      serializedAuthenticatorExtensions <- c.serializedExtensions(converter)
+      attestationStatement <- c.serializeAttestationStatement(converter)
```
MichaelBuessemeyer: For naming consistency, I would rename this to
```suggestion
      serializedAttestationStatement <- c.serializeAttestationStatement(converter)
```

MichaelBuessemeyer (changes requested):
Neat, thanks for keeping digging throught this :muscle: :pray: :office_worker: 

I found a few left over suggestions regarding your latest changes. These are only minor things like concistent naming & extracting error messages.

I think after this, this PR should be ready to go after. At least code-wise. I have never done any testing

project/Dependencies.scala
robert-oleynik: It is necessary, due to a conflict with the Play framework:

```
backend: [error] com.fasterxml.jackson.databind.JsonMappingException: Scala module 2.14.3 requires Jackson Databind version >= 2.14.0 and < 2.15.0 - Found jackson-databind version 2.19.0
backend: [error]        at com.fasterxml.jackson.module.scala.JacksonModule.setupModule(JacksonModule.scala:61)
backend: [error]        at com.fasterxml.jackson.module.scala.JacksonModule.setupModule$(JacksonModule.scala:46)
backend: [error]        at com.fasterxml.jackson.module.scala.DefaultScalaModule.setupModule(DefaultScalaModule.scala:17
...
```

I will update the comment.

app/controllers/AuthenticationController.scala
```diff
@@ -585,8 +585,8 @@ class AuthenticationController @Inject()(
 
   def webauthnAuthStart(): Action[AnyContent] = Action.async { implicit request =>
     for {
-      _ <- Fox.fromBool(conf.Features.passkeysEnabled) ?~> "Passkeys Disabled"
-      _ <- Fox.fromBool(usesHttps) ?~> "Passkeys are only supported with HTTPS"
+      _ <- Fox.fromBool(conf.Features.passkeysEnabled) ?~> Messages("auth.passkeys.disabled")
```
MichaelBuessemeyer: You can leave away the wrapping `Messages` here. It will be automatically done by the `?~>` operator. It is only necessary in case the message behing `"auth.passkeys.disabled"` is a parameterized message. E.g. when the dataset id should be put into the error message. But here it is not necessary and that the usual coding style we have (might not be 100% consisten though). So please remove it here an in the other cases.

```suggestion
      _ <- Fox.fromBool(conf.Features.passkeysEnabled) ?~> "auth.passkeys.disabled"
```
should do the same :D

MichaelBuessemeyer (commented):
Awesome. Thanks for sticking with this PR :tada:

I found one nitpicky improvement. Please see below. After that I would say that this is ready to be merged code wise. Regarding testing I do not have a valid setup. So maybe ask another person to test or if you are confident that it works, I can also approve this myself :)

app/controllers/AuthenticationController.scala
fm3: FYI, the reason this works without the Messages call is that the resulting Failure will be passed through a Messages call later in https://github.com/scalableminds/webknossos/blob/master/util/src/main/scala/com/scalableminds/util/mvc/ExtendedController.scala#L35 ( so the ?~> operator isn’t strictly speaking responsible for this )

conf/evolutions/138-add-webauthn-credentials.sql
```diff
@@ -0,0 +1,28 @@
...
+  name TEXT NOT NULL,
+  userVerified BOOLEAN NOT NULL,
+  backupEligible BOOLEAN NOT NULL,
+  backupState BOOLEAN NOT NULL,
+  serializedAttestationStatement JSONB NOT NULL,
+  serializedAttestedCredential BYTEA NOT NULL,
+  serializedExtensions JSONB NOT NULL,
+  signatureCount INTEGER NOT NULL DEFAULT 0,
+  isDeleted BOOLEAN NOT NULL DEFAULT false,
+  UNIQUE (_multiUser, credentialId)
```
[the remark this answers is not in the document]
robert-oleynik: Credential IDs are generated by the user and therefore, only unique per user not in the entire system. Credentials are identified by the pair of multi user id and credential id.

app/controllers/AuthenticationController.scala
```diff
@@ -405,6 +583,175 @@ class AuthenticationController @Inject()(
...
+      for {
+        _ <- Fox.fromBool(conf.Features.passkeysEnabled) ?~> "auth.passkeys.disabled"
+        _ <- Fox.fromBool(usesHttps) ?~> "auth.passkeys.requiresHttps"
+        cookie <- request.cookies.get("webauthn-session").toFox
+        sessionId = cookie.value
+        challenge <- temporaryAssertionStore
+          .pop(sessionId)
+          .toFox ?~> "Timeout during authentication. Please try again." ~> UNAUTHORIZED
+        authData <- tryo(webAuthnManager.parseAuthenticationResponseJSON(Json.stringify(request.body.key))).toFox ??~> Messages(
+          "auth.passkeys.unauthorized") ~> UNAUTHORIZED
```
MichaelBuessemeyer: Sorry to be so nitpicky but here are still unnecessary `Message(..)` wraps.
```suggestion
        authData <- tryo(webAuthnManager.parseAuthenticationResponseJSON(Json.stringify(request.body.key))).toFox ??~> "auth.passkeys.unauthorized" ~> UNAUTHORIZED
```

frontend/javascripts/admin/auth/passkeys_view.tsx
```diff
@@ -0,0 +1,106 @@
...
+  type WebAuthnKeyDescriptor,
+  doWebAuthnRegistration,
+  listWebAuthnKeys,
+  removeWebAuthnKey,
+} from "admin/api/webauthn";
+import { Button, Input, Modal, Table } from "antd";
+import Toast from "libs/toast";
+import { useEffect, useState } from "react";
+
+function ChangePasswordView() {
```
hotzenklotz: This should be renamed

hotzenklotz: Perhaps, `PassKeysView`

fm3 (approved):
We re-tested auth flows both with passkeys and with passwords, looks good. Let’s go!

Issue #7360, opened 2023-09-26
Can't download tif stack of dataset


boergens:
I open this link:
https://webknossos.org/links/kGdifWwo28zL_2vF

It shows the dataset.
I select Download, and leave everything on default, except selecting a TIF datastack.
When I try that, I get the message "dataset does not exist or could not be accessed":

![image](https://github.com/scalableminds/webknossos/assets/1429753/ede7ca58-02ce-4a43-a152-da74f8aaffbf)

boergens:
Same btw for creating an annotation

fm3:
Hi and thanks for your report!

This is in fact a permission issue. Unfortunately, the error messages are not very clear about this.
This dataset is not public and you are not in the owning organization.

You can see it, because the link you supplied contains a view sharing token. However, this only grants permission to view the dataset in webknossos, not perform further actions like processing jobs (e.g. tiff export) or annotating. We will try to introduce more explanatory error messages for this case.
To proceed with the dataset, please ask the dataset owners to set it to public or invite you to their webknossos organization.

Hope this helps!

boergens:
Dear Florian,

Thanks for your reply.
However, I'm not convinced that this is the right access model.
In my mind, data should either be unavailable or available.
It seems to me that here the data is available but hassle-moated.
If I really wanted this data I could write a script that flies around the dataset and picks up buckets until I have all of them.

Cheers,
Kevin

PR #8397, opened 2025-02-17
Fix export tiff error in case of other orga’s dataset

fm3:
While allowing this in general is a topic for another day ( https://github.com/scalableminds/webknossos/issues/7360 ) the error message for exporting tiff should certainly not say "Calculating mesh files is only allowed for datasets of your own organization".


MichaelBuessemeyer (approved):
Looks good to me. Did not test as I dont see any bug / error potential here

PR #8400, opened 2025-02-18
Docker cert

MichaelBuessemeyer:
- https://___.webknossos.xyz

### Steps to test:
- abc

### TODOs:
- [ ] ...

### Issues:
- fixes #


util/src/main/scala/com/scalableminds/util/tools/ConfigReader.scala
```diff
@@ -5,7 +5,7 @@ import com.typesafe.config.Config
 import play.api.{ConfigLoader, Configuration}
 
 trait ConfigReader {
-  def raw: Configuration
+  val raw: Configuration
```
MichaelBuessemeyer, line 8: This is the change I'd like you to check @fm3. I changed the raw variable from `def` to `val` as `def` is a deferred evaluation and re-evaluated each time it is acutally used. This was previously fine but now using `val` which is immediately evaluated and only evaluated once is more preferrable imo, as the `raw` field in `WkConf` now is merged with the `featureOverwrites` in a "function" which would needed to be reevaluated each time the `raw` field is accessed. 

Does this make sense?  I am aware that the backend caches the result of the `features` route but I still think as the  `raw` field is calculated an no longer a "plain field" that it makes sense to turn this into a `val`

What do you think?

fm3: if it works, fine by me

PR #8401, opened 2025-02-18
Fix inital loading of mappings

MichaelBuessemeyer:
There is currently a bug in the WEBKNOSSOS frontend that leads to mappings being loaded and set for the wrong layer under certain conditions. this PR aims to fix this. Moreover, I added an "assertion" to the code lines that update the mapping name to have a check whether that mapping acutally exists. Else the update is ignored and  an error is printed to the console as this state should never have been reached by a user and to make debugging easier.  

Regarding implementation details: The cause of the bug was that the reducer was working with segmentation layer retrieved by `getRequestedOrVisibleSegmentationLayer(state, null);` which completely disregarded the `volumeTracingId` potentially part of the action that was dispatched. My PR now "honors" a potentially passed `volumeTracingId` or `layerName` in an volume update action which is handled in the reducer after line 411 (in the new version of the code). 
Moreover, in lines 185ff. in `frontend/javascripts/oxalis/model/reducers/volumetracing_reducer_helpers.ts` before changing the mapping name I added code to ensure that this mapping name exists before accepting to update it for the data layer. 

- https://fixinitalloadingofmappings.webknossos.xyz

### Steps to test:
- You need a dataset with a segmentation layer with an agglomerate mapping.
- View this dataset and configure the segmentation layer's agglomerate mapping as default in the default view config (can easily be done in the dataset view in the left sidebar settings)
- Open an new annotation for that dataset form the dashboard and use the extra menu with the three dots. There configure to create a new layer without the existing fallback layer. 
- On the new empty segmentation layer draw around.
- Reload the annotation: No error about an agglomerate mapping that could not be found should be shown. Moreover, in the left settings the active segmentation layer without the fallback should not look like it has a locked agglomerate mapping.

### Issues:
- fixes this bug report: https://scm.slack.com/archives/C02H5T8Q08P/p1739806596969149

------

### DIFF
--- a/app/utils/WkConf.scala
+++ b/app/utils/WkConf.scala
@@ -132,12 +132,9 @@ class WkConf @Inject()(configuration: Configuration, certificateValidationServic
     val publicDemoDatasetUrl: String = get[String]("features.publicDemoDatasetUrl")
     val exportTiffMaxVolumeMVx: Long = get[Long]("features.exportTiffMaxVolumeMVx")
     val exportTiffMaxEdgeLengthVx: Long = get[Long]("features.exportTiffMaxEdgeLengthVx")
-    val openIdConnectEnabled: Boolean =
-      featureOverrides.getOrElse("openIdConnectEnabled", get[Boolean]("features.openIdConnectEnabled"))
-    val editableMappingsEnabled: Boolean =
-      featureOverrides.getOrElse("editableMappingsEnabled", get[Boolean]("features.editableMappingsEnabled"))
-    val segmentAnythingEnabled: Boolean =
-      featureOverrides.getOrElse("segmentAnythingEnabled", get[Boolean]("features.segmentAnythingEnabled"))
+    val openIdConnectEnabled: Boolean = get[Boolean]("features.openIdConnectEnabled")
+    val editableMappingsEnabled: Boolean = get[Boolean]("features.editableMappingsEnabled")
+    val segmentAnythingEnabled: Boolean = get[Boolean]("features.segmentAnythingEnabled")
   }
 
   object Datastore {
--- a/frontend/javascripts/oxalis/model/reducers/volumetracing_reducer.ts
+++ b/frontend/javascripts/oxalis/model/reducers/volumetracing_reducer.ts
@@ -277,15 +277,29 @@ export function serverVolumeToClientVolumeTracing(tracing: ServerVolumeTracing):
   return volumeTracing;
 }
 
-function VolumeTracingReducer(
-  state: OxalisState,
-  action:
-    | VolumeTracingAction
-    | SetMappingAction
-    | FinishMappingInitializationAction
-    | SetMappingEnabledAction
-    | SetMappingNameAction,
-): OxalisState {
+type VolumeTracingReducerAction =
+  | VolumeTracingAction
+  | SetMappingAction
+  | FinishMappingInitializationAction
+  | SetMappingEnabledAction
+  | SetMappingNameAction;
+
+function retrieveVolumeTracing(state: OxalisState, action: VolumeTracingReducerAction) {
+  if ("tracingId" in action && action.tracingId != null) {
+    return getVolumeTracingById(state.tracing, action.tracingId);
+  }
+  const volumeLayer = getRequestedOrVisibleSegmentationLayer(
+    state,
+    "layerName" in action ? action.layerName : null,
+  );
+
+  if (volumeLayer == null || volumeLayer.tracingId == null) {
+    return null;
+  }
+  return getVolumeTracingById(state.tracing, volumeLayer.tracingId);
+}
+
+function VolumeTracingReducer(state: OxalisState, action: VolumeTracingReducerAction): OxalisState {
   switch (action.type) {
     case "INITIALIZE_VOLUMETRACING": {
       const volumeTracing = serverVolumeToClientVolumeTracing(action.tracing);
@@ -394,13 +408,10 @@ function VolumeTracingReducer(
     return state;
   }
 
-  const volumeLayer = getRequestedOrVisibleSegmentationLayer(state, null);
-
-  if (volumeLayer == null || volumeLayer.tracingId == null) {
+  const volumeTracing = retrieveVolumeTracing(state, action);
+  if (volumeTracing == null) {
     return state;
   }
-
-  const volumeTracing = getVolumeTracingById(state.tracing, volumeLayer.tracingId);
 
   switch (action.type) {
     case "SET_ACTIVE_CELL": {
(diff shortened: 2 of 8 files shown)


app/utils/WkConf.scala
```diff
@@ -132,12 +132,9 @@ class WkConf @Inject()(configuration: Configuration, certificateValidationServic
...
-    val openIdConnectEnabled: Boolean =
-      featureOverrides.getOrElse("openIdConnectEnabled", get[Boolean]("features.openIdConnectEnabled"))
-    val editableMappingsEnabled: Boolean =
-      featureOverrides.getOrElse("editableMappingsEnabled", get[Boolean]("features.editableMappingsEnabled"))
-    val segmentAnythingEnabled: Boolean =
-      featureOverrides.getOrElse("segmentAnythingEnabled", get[Boolean]("features.segmentAnythingEnabled"))
+    val openIdConnectEnabled: Boolean = get[Boolean]("features.openIdConnectEnabled")
+    val editableMappingsEnabled: Boolean = get[Boolean]("features.editableMappingsEnabled")
+    val segmentAnythingEnabled: Boolean = get[Boolean]("features.segmentAnythingEnabled")
   }
```
MichaelBuessemeyer, line 138: @fm3 Sorry to ping you but I need another quick check from your side. I forgot to remove the manual overwrite handling in the `WkConf`'s `Features` object, which is now already done in 18ll of this file. So this is just a small clean-up.

I tested this via println output in the backend to see whether the overwrites were actually applied. So imo these changes should be fine

MichaelBuessemeyer:
@daniel-wer In case you don't have time to review this, please forward the review request to e.g. philipp :handbag:

frontend/javascripts/oxalis/model/reducers/volumetracing_reducer_helpers.ts
```diff
@@ -179,6 +182,23 @@ export function setMappingNameReducer(
...
+      state.dataset,
+      volumeTracing.tracingId,
+    );
+    const listToSearchIn = mappingType === "HDF5" ? agglomerates : mappings;
+    if (
+      mappingType === "HDF5" &&
+      (listToSearchIn == null || listToSearchIn.indexOf(mappingName) === -1)
+    ) {
+      console.error(
+        `Cannot set mapping ${mappingName} of type ${mappingType} for volume tracing ${volumeTracing.tracingId} because it is not included in the agglomerate list of the data layer: ${agglomerates}.`,
```
daniel-wer, line 196:
```suggestion
        `Cannot set mapping ${mappingName} of type ${mappingType} for volume tracing ${volumeTracing.tracingId} because it is not included in the mapping list of the data layer: ${listToSearchIn}.`,
```

frontend/javascripts/oxalis/model/reducers/volumetracing_reducer_helpers.ts
```diff
@@ -179,6 +182,23 @@ export function setMappingNameReducer(
...
+    const { agglomerates, mappings } = getSegmentationLayerByName(
+      state.dataset,
+      volumeTracing.tracingId,
+    );
+    const listToSearchIn = mappingType === "HDF5" ? agglomerates : mappings;
+    if (
+      mappingType === "HDF5" &&
+      (listToSearchIn == null || listToSearchIn.indexOf(mappingName) === -1)
+    ) {
+      console.error(
```
daniel-wer, line 195: If we want to notice when this happens because it indicates a bug, it should also be logged to airbrake

frontend/javascripts/oxalis/model/reducers/volumetracing_reducer_helpers.ts
```diff
@@ -179,6 +182,23 @@ export function setMappingNameReducer(
   if (mappingType !== "HDF5" || !isMappingEnabled) {
     mappingName = null;
   }
+  if (mappingName) {
+    const { agglomerates, mappings } = getSegmentationLayerByName(
+      state.dataset,
+      volumeTracing.tracingId,
+    );
+    const listToSearchIn = mappingType === "HDF5" ? agglomerates : mappings;
```
daniel-wer, line 190: Did you deliberately include both cases here for the code to be future-proof? If yes, that's appreciated :) I'm asking because in line 183 the mappingName is set to null for non-hdf5 mappings and this code is only executed if mappingName is not null (line 185).

daniel-wer (approved):
Nice fixes! Testing also went well :+1:

MichaelBuessemeyer (commented):
Ok nvm, I needed to remove the error handling in this pr. See my next comment

frontend/javascripts/oxalis/model/reducers/volumetracing_reducer_helpers.ts
```diff
@@ -179,6 +182,23 @@ export function setMappingNameReducer(
...
+      mappingType === "HDF5" &&
+      (listToSearchIn == null || listToSearchIn.indexOf(mappingName) === -1)
+    ) {
+      console.error(
+        `Cannot set mapping ${mappingName} of type ${mappingType} for volume tracing ${volumeTracing.tracingId} because it is not included in the mapping list of the data layer: ${listToSearchIn}.`,
+      );
+      return state;
+    }
+  }
+
```
MichaelBuessemeyer, line 201: Ok I did some more testing regarding this error handling. Turns out, when a default mapping is configured, this code is executed before the existing mappings for each layer are loaded by the frontend to set the inital mapping. Thus, this check here would cause the error being logged and then the inital mapping settings would be ignored. Therefore, I removed this error catching mechanism here for now.

To re-add this, I think we need to refactor the loading mechanism of initial mappings. 

@daniel-wer what do you think about this? I think to fix the error of using the correct volume tracing is for now a good enough improvement and should be merged better early than late

MichaelBuessemeyer:
> Turns out, when a default mapping is configured, this code is executed before the existing mappings for each layer are loaded by the frontend to set the inital mapping. Thus, this check here would cause the error being logged and then the inital mapping settings would be ignored. Therefore, I removed this error catching mechanism here for now.

This only occurs when opening an annotation with active mappings configured. The `setMappingNameReducer` seems to be never called in view mode (as its arguments indicate).

frontend/javascripts/oxalis/model/reducers/volumetracing_reducer.ts
```diff
@@ -277,15 +279,33 @@ export function serverVolumeToClientVolumeTracing(tracing: ServerVolumeTracing):
...
-    | SetMappingNameAction,
-): OxalisState {
+type VolumeTracingReducerAction =
+  | VolumeTracingAction
+  | SetMappingAction
+  | FinishMappingInitializationAction
+  | SetMappingEnabledAction
+  | SetMappingNameAction;
+
+function retrieveVolumeTracing(state: OxalisState, action: VolumeTracingReducerAction) {
```
philippotto, line 289: How about `getVolumeTracingFromAction`? I find that way clearer.

PR #8402, opened 2025-02-19
In IntervalScheduler, keep Future-returning ticks overlap-free

fm3:
Before this PR, the tick functions returned Unit, meaning that if they started asynchronous work, the IntervalScheduler couldn’t wait for it to be completed.

Now, tick returns Fox[Unit]. The IntervalScheduler runs with an internal interval of 100ms and checks if the time since the last tick’s completion is long enough to start a new one. That means the real interval will be a little higher than the specified one, because it *includes* the time to compute the tick (plus the time until the next internal 100ms-spaced tick).

### Steps to test:
- To test this, I added a 10s sleep to one of the tick’s asynchronous bodies, and turned down the interval to 2 seconds, and added some logging. The output was like this (note that the internal ticker interval was still set to 20ms then:

```
11:51:20,304 Tick! Sleeping 10s...
11:51:30,304 Sleeping done.

11:51:32,324 Tick! Sleeping 10s...
11:51:42,324 Sleeping done.

11:51:44,344 Tick! Sleeping 10s...
11:51:54,344 Sleeping done.

11:51:56,344 Tick! Sleeping 10s...
11:52:06,344 Sleeping done.

11:52:08,344 Tick! Sleeping 10s...
11:52:18,344 Sleeping done.

11:52:20,364 Tick! Sleeping 10s...
11:52:30,364 Sleeping done.

11:52:32,393 Tick! Sleeping 10s...

```

### Issues:
- Hopefully fixes overloaded postgres when receiving datasource upgrades from the datastore. (My theory is that the next one came, after 1min, before the first one was completely ingested).

------

### DIFF
--- a/app/controllers/WKRemoteDataStoreController.scala
+++ b/app/controllers/WKRemoteDataStoreController.scala
@@ -2,6 +2,7 @@ package controllers
 
 import com.scalableminds.util.accesscontext.{AuthorizedAccessContext, GlobalAccessContext}
 import com.scalableminds.util.objectid.ObjectId
+import com.scalableminds.util.time.Instant
 import com.scalableminds.util.tools.Fox
 import com.scalableminds.webknossos.datastore.controllers.JobExportProperties
 import com.scalableminds.webknossos.datastore.models.UnfinishedUpload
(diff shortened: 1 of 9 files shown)


MichaelBuessemeyer (approved):
Thanks for this PR. :tada: 

Hopefully this fixes the recent errors :crossed_fingers: 

As you renamed the `system: ActorSystem` injects, could you maybe do this as well for the other cases which do not use the interval ticking? I found a few other occurences which still use `system: ActorSystem`.

Besides this and my comment everything looks good and works :tada:

PR #8405, opened 2025-02-24
Look up datasource infos by annotationId rather than tracingId

### LINKED ISSUE #4961: Task Creation: Reduce TaskType Lookups
Task type is looked up in the db for each task twice, even in a bulk creation. Usually they all have the same task type. Some caching could speed up task creation. 
In a quick measurement this took about 2ms per task. So this is not the slowest park of task creation, but maybe one where a relatively easy gain is possible.

### LINKED ISSUE #8403: RevertToVersion failed for annotation across changed layer set
See report at https://scm.slack.com/archives/C5AKLAV0B/p1739973409199019

Noteworthy: Multiple reverts in direct succession in the action log; addLayer and deleteLayer actions present.

Maybe we can have a variant of `getFallbackLayer` that uses the annotation id for the datasource lookup? That way we wouldn’t have to rely on postgres having the current layer info.

fm3:
Reverting annotations (and many operations in tracingstore) relied on looking up the datasource by tracingId. This is not possible if the tracingId does not exist anymore (e.g. because the layer was deleted). Now I use annotationId for those lookups.

That meant that the annotationId needs to be passed to many spots. That sent me into a bit of a rabbit hole. A major change is that now the wk-core side sends new ids (both annotation ids and tracing ids) when saving tracings to the tracingstore. Previously, the tracingstore generated the ids and returned them after saving.

Also I simplified the task creation a bit by asserting that the tasktype is identical for all requested tasks (I cleared this decision with norman). That way, task creation does way fewer postgres queries.

- https://tracingidannotationid.webknossos.xyz

### Steps to test:
- Create Annotation, brush some, add some skeletons
- Reverting to an older version of an annotation should work, even if the volume layer set has changed in the meantime
- Reverting to the same version twice in a row should work
- Ensure I didn’t break any other annotation/task functionality :
  - Download + reupload should work
  - Upload should also work in case of multiple annotations (both with and without volume)
  - Task creation with and without files should work
  - Task creation with multiple files, one of which has an invalid NML, should work and report that only one errored
  - Finish some tasks, view compound project, should work (if all matching tasks are on the same DS)
  - Merging Annotations should work (skeleton-only, volume, editable mapping)
  - Duplicating annotations should work

### Issues:
- fixes #8403
- fixes #4961 

------


webknossos-tracingstore/app/com/scalableminds/webknossos/tracingstore/tracings/volume/VolumeTracingService.scala
```diff
@@ -631,12 +646,16 @@ class VolumeTracingService @Inject()(
...
 
-  def findData(tracingId: String, tracing: VolumeTracing)(implicit tc: TokenContext): Fox[Option[Vec3Int]] =
+  def findData(annotationId: String, tracingId: String, tracing: VolumeTracing)(
+      implicit tc: TokenContext): Fox[Option[Vec3Int]] =
     for {
       _ <- Fox.successful(())
       isTemporaryTracing <- temporaryTracingService.isTemporaryTracing(tracingId)
-      volumeLayer = volumeTracingLayer(tracingId, tracing, isTemporaryTracing = isTemporaryTracing)
-      bucketStream = volumeLayer.bucketProvider.bucketStream(Some(tracing.version))
+      volumeLayer = volumeTracingLayer(annotationId: String,
```
MichaelBuessemeyer, line 654: Why is this explicit type hint needed?
```suggestion
      volumeLayer = volumeTracingLayer(annotationId,
```
I think thats just a copy & :spaghetti: mistake, right?

MichaelBuessemeyer (commented):
Thanks for digging through all this :pray: :pick:.

I did not really find anything. But during review I was wondering why volume annotation layers need such a special care (because they might have been deleted) and skeleton layers don't? Because theoretically it is possible to remove the skeleton layer of an annotation. Wouldn't the bug reoccure in such a case but this time with the skeleton layer as it was deleted and cannot be found by tracing id?

Will do testing now

MichaelBuessemeyer:
Ok here are my testing results / questions:
In short: Everything works but I found one error/problem in the bulk task creation. But I am unsure whether this is related to these changes here

- Download + reupload should work :heavy_check_mark: 
- Upload should also work in case of multiple annotations (both with and without volume) :heavy_check_mark: 
- Task creation with and without files should work :heavy_check_mark: 
- I tested bulk task creation with: `67c06da21db30e5cbae31b4b,63721e2cef0100470266c485,sampleExp,1,0,0,0,0,0,0,1,0,0,0,0,0,0,sampleProject,,67c06e961db30ebabae31b6d` which does not work. Only when I remove the annotation id which should be used as a base. What is maybe even more troubling: The backend error is simply `{  "tasks": [{"status": 400,"error": "Empty"}]}` which does not help to identify what is wrong with supplying this annotation id for the base annotation id.
- Task creation with multiple files, one of which has an invalid NML, should work and report that only one errored :heavy_check_mark: 
- Finish some tasks, view compound project, should work (if all matching tasks are on the same DS) :heavy_check_mark: 
- Merging Annotations should work (skeleton-only, volume, editable mapping) :heavy_check_mark: 
- Duplicating annotations should work :heavy_check_mark:

MichaelBuessemeyer (approved):
> I tested bulk task creation with: 67c06da21db30e5cbae31b4b,63721e2cef0100470266c485,sampleExp,1,0,0,0,0,0,0,1,0,0,0,0,0,0,sampleProject,,67c06e961db30ebabae31b6d which does not work. Only when I remove the annotation id which should be used as a base. What is maybe even more troubling: The backend error is simply {  "tasks": [{"status": 400,"error": "Empty"}]} which does not help to identify what is wrong with supplying this annotation id for the base annotation id.

Nice works now :tada: 

I'd say its merge ready I don't see any more todos

PR #8411, opened 2025-02-26
Release 25.02.1

philippotto:
Contains c4aaa40c31

- [X] Updated [migration guide](../blob/master/MIGRATIONS.unreleased.md#unreleased) if applicable


fm3 (commented):
Not sure I understand the changelog diff :thinking:  Looks like the 25.02.0 part is added here too? Did you not branch off 25.02.0?

philippotto:
> Not sure I understand the changelog diff 🤔 Looks like the 25.02.0 part is added here too? Did you not branch off 25.02.0?

I branched off the tag 25.02.0. The commit in that tag got squashed into the master, so the exact commit is not contained in master. That's why github shows that diff, too. I'm going to tag 25.02.1 now and then pull the master into it. afterwards, only the changelog diff for 25.02.1 should remain.

fm3 (approved):
Got it, thanks!

Issue #8412, opened 2025-02-26
Bulk Task Creation description doesn’t match expected format


fm3:
Bulk task creation page says to start with `datasetName, datasetId`, however the code actually expects just the datasetId.

We should either allow adding the name or adapt the description.

Do you know if we already decided on one of these? @MichaelBuessemeyer

MichaelBuessemeyer:
Well, the `datasetName` is unnecessary and does not add any valuable information as there might be multiple dataset with the same name. Thus, I would avoid having this i the suggested text.

MichaelBuessemeyer:
I opened https://github.com/scalableminds/webknossos/pull/8413

PR #8413, opened 2025-02-26
Remove datasetName from task bulk creation view info

MichaelBuessemeyer:
Remove datasetName from task bulk creation view info as the bulk creation does no longer accept this in its csv format.

- https://___.webknossos.xyz

### Steps to test:
- Shouldn't be needed

### Issues:
- fixes https://github.com/scalableminds/webknossos/issues/8412

------
- ~~[ ] Updated [changelog](../blob/master/CHANGELOG.unreleased.md#unreleased)~~ -> also unnecessary I'd say


fm3:
I’m not sure if we discussed this before, but how are users supposed to find the dataset id in the first place? It’s nowhere in the UI, right? (Except by extracting it from the URI if you open a dataset?)

MichaelBuessemeyer:
> I’m not sure if we discussed this before, but how are users supposed to find the dataset id in the first place? It’s nowhere in the UI, right? (Except by extracting it from the URI if you open a dataset?)

Yeah, that's actually quite the problem... I only identified that it is also possible to get the id after successfully creating a task for the dataset and downloading the csv. This includes the dataset id. But this is not the slightest discoverable.

Therefore maybe something like this: I added a small copy icon to the info view of the tracing view.
image

Other possibilities I can imaging are alike copy id buttons in the edit view of a dataset or in the dataset table itself

What do you think? @philippotto, @normanrz and @daniel-wer might also have opinions on this :thinking:

philippotto:
Your "copy id" button is a bit confusing to me, because such buttons usually copy the text that's next to them (and not something different). The clearest way would be to show the id separately in the sidebar (dataset view and dataset table). However, usually the ID is rather unimportant to the user. Therefor, I wouldn't want to waste space for that.

> Other possibilities I can imaging are alike copy id buttons in the edit view of a dataset [...]

Yes, I think having it in the edit view of the dataset is a good spot. Ideally, the bulk creation view should also explain where the id can be found.

normanrz:
Dataset edit view is fine. I think we should also put it in the sidebar in the dataset table.

MichaelBuessemeyer:
Ok I added an id button the the sidebar of the dataset table:
image
and one to the settings view:
image
(wording of tooltips is now unified to "Copy Dataset ID")

MichaelBuessemeyer:
Maybe now "so much" frontend changes a quick review from @philippotto or @daniel-wer would be nice :thinking: 

@philippotto Do you maybe have time?

MichaelBuessemeyer:
TODO for me: Make UI similar to 
image
in the dataset settings

MichaelBuessemeyer:
Ok done: Looks like this now: 
image

So still ready for review :octocat:

frontend/javascripts/dashboard/dataset/dataset_settings_data_tab.tsx
```diff
@@ -212,10 +223,37 @@ function SimpleDatasetForm({
...
-                      width: 408,
+                      width: LEFT_COLUMN_ITEMS_WIDTH,
                     }}
                   />
                 </FormItemWithInfo>
+                <Space size="large" />
+                <FormItemWithInfo
+                  name={["dataset", "id"]}
+                  label="Dataset ID"
+                  info="The id used to identify the dataset. Needed for e.g. Task bulk creation."
```
philippotto, line 234:
```suggestion
                  info="The ID used to identify the dataset. Needed for e.g. Task bulk creation."
```

frontend/javascripts/dashboard/dataset/dataset_settings_data_tab.tsx
```diff
@@ -212,10 +223,37 @@ function SimpleDatasetForm({
...
+                    </Tooltip>
+                  </Space.Compact>
+                </FormItemWithInfo>
+                {/*<span style={{ marginLeft: 16 }}>
+                    <Tooltip title="Copy dataset ID">
+                      <Button onClick={() => copyDatasetID(dataset?.id)}>
+                        <CopyOutlined />
+                      </Button>
+                    </Tooltip>
+                  </span>*/}
```
philippotto, line 256: delete?

philippotto (approved):
nice! just see my two small comments before merging :)

frontend/javascripts/dashboard/dataset/dataset_settings_data_tab.tsx
MichaelBuessemeyer: Oh damn. Thanks for catching. :pray:

Issue #8234, opened 2024-11-27
Allow to share/copy a ML model with/to another organization


philippotto:
After training a model with WK, there should be a way to share/copy/publish this model to another organization. This could be a dropdown action in the models table.

Use case: training models for customers and sharing these with them afterwards.

PR #8418, opened 2025-02-28
Make AI models cross orga shareable

MichaelBuessemeyer:
- https://sharableaimodels.webknossos.xyz/

### Steps to test:
- Log in as the usual sample user
- open https://sharableaimodels.webknossos.xyz/aiModels
- the default model should be shown
- try editing the shared orgas of the model
- reload, the changes should be persisted
- log in a the other user: asdf@asdf.com pw: same as sample user
- open https://sharableaimodels.webknossos.xyz/aiModels and see the model
  - "problem" currently shown to the user anyway as one needs to be a super user to see the models list and all super users can see all models per default. See my comment https://github.com/scalableminds/webknossos/pull/8418#discussion_r1981524270 regarding this
  - As the asdf user open the dataset https://sharableaimodels.webknossos.xyz/datasets/67c863d7010000c200fe2b32#512,512,512,0,1.3 and try to start a custom inferral job with the shared ai model selected. -> should work 

### TODOs:
- [x] evolution & reversion
- [x] some more testing - try starting an inference with a shared modal
- [x] clean up code
- [x] make sharing resilient if users from different organizations can edit sharing settings

### Issues:
- fixes [#8234](https://github.com/scalableminds/webknossos/issues/8234)

------


app/models/aimodels/AiModel.scala
```diff
@@ -87,7 +112,32 @@ class AiModelDAO @Inject()(sqlClient: SqlClient)(implicit ec: ExecutionContext)
...
+                  SELECT isSuperUser
+                  FROM webknossos.multiUsers_
+                  WHERE _id IN (
+                      SELECT _multiUser
+                      FROM webknossos.users_
+                      WHERE _id = $requestingUserId
+                  )
+              )
+          )
+     """
```
MichaelBuessemeyer, line 140: I adjusted the accessor query so that all models are shown to the user which the user has potentially access to -> all organization in which the user is part of or allowed to join because of being a super user.

While testing I realized that this might be unwanted as e.g. super users can see all models all the time therefore potentially cluttering the view / list of models.

@fm3 Do you think it makes sense to only show models of the organization in which the user is currently logged in?

MichaelBuessemeyer (commented):
I'd say this pr is ready for the first review round :circus_tent:

MichaelBuessemeyer:
@fm3 and @philippotto do you have time to review the backend / frontend.  I'll provide a testing instance with a model and multiple orgas :)

conf/application.conf
```diff
@@ -146,13 +146,13 @@ features {
...
-  isWkorgInstance = false
+  isWkorgInstance = true
   recommendWkorgInstance = true
   taskReopenAllowedInSeconds = 30
   allowDeleteDatasets = true
   # to enable jobs for local development, use "yarn enable-jobs" to also activate it in the database
-  jobsEnabled = false
-  voxelyticsEnabled = false
+  jobsEnabled = true
+  voxelyticsEnabled = true
```
MichaelBuessemeyer, line 155: Undo me

app/controllers/AiModelController.scala
```diff
@@ -156,7 +156,8 @@ class AiModelController @Inject()(
           .submitJob(jobCommand, commandArgs, request.identity, dataStore.name) ?~> "job.couldNotRunTrainModel"
         newAiModel = AiModel(
           _id = modelId,
-          _organization = request.identity._organization,
+          _owningOrganization = request.identity._organization,
+          _organizations = List(request.identity._organization),
```
fm3, line 160: Should probably also use the term sharedOrganizations that you use in code elsewhere

app/controllers/AiModelController.scala
```diff
@@ -213,14 +214,25 @@ class AiModelController @Inject()(
...
-      for {
-        _ <- userService.assertIsSuperUser(request.identity)
-        aiModel <- aiModelDAO.findOne(aiModelId) ?~> "aiModel.notFound" ~> NOT_FOUND
-        _ <- aiModelDAO.updateOne(
-          aiModel.copy(name = request.body.name, comment = request.body.comment, modified = Instant.now))
-        updatedAiModel <- aiModelDAO.findOne(aiModelId) ?~> "aiModel.notFound" ~> NOT_FOUND
-        jsResult <- aiModelService.publicWrites(updatedAiModel)
-      } yield Ok(jsResult)
+      {
+        // Automatically add the owning organization to the shared organizations if it is not already there.
```
fm3, line 218: Maybe we should avoid this redundant saving and just consider the owning organization too when checking the shared orgas? I’m not sure what’s best, but the current variant means that we have to ensure that the owning orga is always in the shared orgas list too. But according to the comment, it could indeed be not already there. What happens in this case?

app/models/aimodels/AiModel.scala
```diff
@@ -138,6 +192,27 @@ class AiModelDAO @Inject()(sqlClient: SqlClient)(implicit ec: ExecutionContext)
           .as[ObjectId])
     } yield rows.toList
 
+  private def insertOrganizationQuery(aiModelId: ObjectId,
+                                      organizationIds: List[String]): List[SqlAction[Int, NoStream, Effect]] = {
+    println("--------------------------------------------------------------------------------")
+    println(s"insertOrganizationQuery($aiModelId, ${organizationIds.mkString(",")})")
+    println("--------------------------------------------------------------------------------")
```
fm3, line 199: looks like dev debug logging

app/models/aimodels/AiModel.scala
```diff
@@ -138,6 +192,27 @@ class AiModelDAO @Inject()(sqlClient: SqlClient)(implicit ec: ExecutionContext)
...
+    organizationIds.map { organizationId =>
+      insertOrganizationQuery(aiModelId, organizationId)
+    }
+  }
+
+  private def insertOrganizationQuery(aiModelId: ObjectId, organizationId: String): SqlAction[Int, NoStream, Effect] =
+    q"""INSERT INTO webknossos.aiModel_organizations(_aiModel, _organization)
+            VALUES($aiModelId, $organizationId)""".asUpdate
+
+  private def findOrganizationsFor(aiModelId: ObjectId): Fox[List[String]] =
```
fm3, line 209: please use the term sharedOrganizations everywhere

tools/postgres/schema.sql
```diff
@@ -553,7 +553,7 @@ CREATE TYPE webknossos.AI_MODEL_CATEGORY AS ENUM ('em_neurons', 'em_nuclei', 'em
 
 CREATE TABLE webknossos.aiModels(
   _id CHAR(24) PRIMARY KEY,
-  _organization VARCHAR(256) NOT NULL,
+  _owningOrganization VARCHAR(256) NOT NULL,
```
fm3, line 556: It might be more consistent with the other code to leave this `_organization` and rather rename the others to `sharedOrganizations`. That would be like what we have with annotations and teams. What do you think?

frontend/javascripts/test/snapshots/public-test/test-bundle/test/backend-snapshot-tests/misc.e2e.js.md
```diff
@@ -76,13 +76,13 @@ Generated by [AVA](https://avajs.dev).
...
+      isWkorgInstance: true,
+      jobsEnabled: true,
       openIdConnectEnabled: false,
       optInTabs: [],
       publicDemoDatasetUrl: 'https://webknossos.org/datasets/scalable_minds/l4dense_motta_et_al_demo',
       recommendWkorgInstance: true,
       segmentAnythingEnabled: false,
       taskReopenAllowedInSeconds: 30,
-      voxelyticsEnabled: false,
+      voxelyticsEnabled: true,
```
fm3, line 87: when the application.conf is reset again, these changes also need to be undone

app/models/aimodels/AiModel.scala
fm3: Yes I agree, only the organization the user is currently logged in to should be considered.

fm3 (commented):
Backend looking good! First test worked for me as well :)

I think “shared organizations” is a bit of a weird term for the UI, since it’s not the organizations that are shared , but the model is shared *with them*. Maybe the frontend reviewers have good suggestions there.

On the other hand, in the code, I’d stick to the term sharedOrganizations, as analogous to annotation’s sharedTeams, and would in fact suggest to use it more consistently, see my comments.

I added a couple more comments.

philippotto (commented):
frontend looks good mostly :+1: cool stuff! left some comments.

I also have one question regarding the permissions: Who exactly may update the shared orgas? from the FE code it looks like I have to be member of the owning orga. but what happens in this scenario:

Orga 1 contains User A and user B
Orga 2 contains User B and user C

if user B creates the model in orga 2, user B and user C can edit the organizations, right? however, if B sets the orgas to 1 and 2, user C won't be able to see/change orga 1. when saving, they will likely remove orga 1 automatically because the FE only fetches `getUsersOrganizations`.

MichaelBuessemeyer:
> I also have one question regarding the permissions: Who exactly may update the shared orgas? from the FE code it looks like I have to be member of the owning orga. but what happens in this scenario:

> Orga 1 contains User A and user B
Orga 2 contains User B and user C

> if user B creates the model in orga 2, user B and user C can edit the organizations, right? however, if B sets the orgas to 1 and 2, user C won't be able to see/change orga 1. when saving, they will likely remove orga 1 automatically because the FE only fetches getUsersOrganizations.

Currently, only super users can change the sharing settings of a model. The issue states that this feature is needed to make trained models from us available to customers. As we are all super user this should work. Additionally, the active user (not mulit user) needs to be in the owning organization (the orga in which the training was started in). 

> 
> if user B creates the model in orga 2, user B and user C can edit the organizations, right? however, if B sets the orgas to 1 and 2, user C won't be able to see/change orga 1. when saving, they will likely remove orga 1 automatically because the FE only fetches getUsersOrganizations.

Oh damn. Very good catch. That could actually become a problem when we allow this for non-super users :thinking:.

TODO:
- [ ] make sharing resilient if users from different organizations can edit sharing settings

app/controllers/AiModelController.scala
```diff
@@ -213,14 +214,40 @@ class AiModelController @Inject()(
...
+                                                 comment = request.body.comment,
+                                                 modified = Instant.now)) ?~> "aiModel.updatingFailed"
+          areSharedOrganizationsUpdated = sharedOrganizationIdsOpt
+            .getOrElse(List())
+            .toSet != aiModel._sharedOrganizations.toSet
+          sharedAndPreservedOrganizationIdsOpt <- if (areSharedOrganizationsUpdated)
+            Fox.runOptional(sharedOrganizationIdsOpt) { newlySharedOrganizationIds =>
+              // Keep organizations with access to the aiModel which the current user has no access to.
+              for {
+                organizationIdsUserCanAccess <- organizationDAO.findAll.flatMap(os => Fox.successful(os.map(_._id)))
```
fm3, line 238:
```suggestion
                organizationIdsUserCanAccess <- organizationDAO.findAll.map(os => os.map(_._id))
```
should be identical

app/controllers/AiModelController.scala
```diff
@@ -213,14 +214,40 @@ class AiModelController @Inject()(
...
+                  organizationIdsUserCanAccess.contains)
+              } yield newlySharedOrganizationIds ++ organizationsToPreserveAccessTo
+            } else Fox.successful(None)
+          _ <- Fox.runOptional(sharedAndPreservedOrganizationIdsOpt)(
+            orgas =>
+              if (orgas.toSet == aiModel._sharedOrganizations.toSet) Fox.successful(())
+              else aiModelDAO.updateSharedOrganizations(aiModel._id, orgas)) ?~> "aiModel.updatingSharedFailed"
+          updatedAiModel <- aiModelDAO.findOne(aiModelId) ?~> "aiModel.notFound" ~> NOT_FOUND
+          jsResult <- aiModelService.publicWrites(updatedAiModel, request.identity)
+        } yield Ok(jsResult)
```
fm3, line 249:
```suggestion
          js <- aiModelService.publicWrites(updatedAiModel, request.identity)
        } yield Ok(js)
```
(or json. but Result is a bit misleading an instance of play’s `Result` class is only built by the `Ok`)

app/models/aimodels/AiModel.scala
```diff
@@ -36,25 +40,50 @@ class AiModelService @Inject()(dataStoreDAO: DataStoreDAO,
...
                                jobDAO: JobDAO,
                                jobService: JobService) {
-  def publicWrites(aiModel: AiModel)(implicit ec: ExecutionContext, ctx: DBAccessContext): Fox[JsObject] =
+  def publicWrites(aiModel: AiModel, requestingUser: User)(implicit ec: ExecutionContext,
+                                                           ctx: DBAccessContext): Fox[JsObject] =
     for {
       dataStore <- dataStoreDAO.findOneByName(aiModel._dataStore)
-      user <- userDAO.findOne(aiModel._user)
-      userJs <- userService.compactWrites(user)
+      user <- userDAO
```
fm3, line 50: maybe `userOpt`? Also, maybe we have a box2Option conversion somewhere that could be used Here? Or if not, maybe create it. Ideally, I would see something like `Fox.runOptional(userBox.toOption)(userService.compactWrites)` but I don’t know if we have everything to make that happen.

tools/postgres/schema.sql
fm3: I see your point, and ultimately it’s not that important. But I still that if the organizations the model is shared with were called webknossos.aiModel_sharedOrganizations (which would be consistent with sharedTeams *and* more descriptive than just organizations), the semantics of `_organization` would be fairly clear. This is the organization the aiModel lives in / belongs to, just like with users, datasets, teams etc..

fm3 (commented):
Backend is getting there :) I added a few more minor comments.

If you prefer the approach of re-adding the user orga to the shared organizations on every update over adding the owning organization to the access query, well fine by me, it makes no big difference ;-)

tools/postgres/schema.sql
MichaelBuessemeyer: Ok I switched back to `_organization`

MichaelBuessemeyer:
> If you prefer the approach of re-adding the user orga to the shared organizations on every update over adding the owning organization to the access query, well fine by me

I adjusted the code to implicitly add the owning orga to the shared orgas. This was your initial intended solution, correct? Please have another look.

fm3 (approved):
Awesome, backend LGTM and I also retested, looking good. I hope I didn’t annoy you too much by somewhat persisting on my suggestions :innocent:

frontend/javascripts/admin/voxelytics/ai_model_list_view.tsx
```diff
@@ -40,6 +42,16 @@ export default function AiModelListView() {
       {isTrainModalVisible ? (
         <TrainNewAiJobModal onClose={() => setIsTrainModalVisible(false)} />
       ) : null}
+      {currentlyEditedModel ? (
+        <EditModelSharedOrganizationsModal
+          model={currentlyEditedModel}
+          onClose={() => {
+            setCurrentlyEditedModel(null);
+            () => setRefreshCounter((val) => val + 1);
```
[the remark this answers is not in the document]
philippotto: yes, the `setRefreshCounter` stuff makes sense. however, it is not triggered unless I'm mistaken? the function invocation is within a function definition (`() => ...`) and that function is not called?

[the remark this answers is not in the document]
philippotto: ah, you fixed it already. then, this is done.

philippotto:
frontend also looks good :+1: by the way, I "resolved" the remaining comments that showed up as "outdated". probably makes more sense for you to do this in the future, @MichaelBuessemeyer, because you have a better overview as the PR author :)

MichaelBuessemeyer:
> Awesome, backend LGTM and I also retested, looking good. I hope I didn’t annoy you too much by somewhat persisting on my suggestions 😇


Nope, everything is alrighty :smile_cat:

PR #8422, opened 2025-03-03
Use temporary store for segment index for compound annotations

### LINKED ISSUE #8204: Compound Annotation’s Segment Index is stored in FossilDB, should be in TemporaryStore
When viewing a compound annotation with volume layers, the merged volume buckts and tracing layers are stored in memory in a TemporaryStore. However, the segment index is still stored in fossildb. This is inconsistent and is a waste of storage. It should also be stored in a TemporaryStore.

Note that this should be tackled after #7839 to avoid merge conflicts.

frcroth:
- https://segmentindexbufferintemporarystore.webknossos.xyz/

### Steps to test:
- Create a compound annotation of volumes (using a task)
- Check that there is a segment volume of > 0

### TODOs:
- [x] ...

### Issues:
- fixes #8204

------


fm3 (approved):
LGTM and appears to work well :)

Issue #8404, opened 2025-02-20
Profile + Optimize performance of saving volume buckets / updating segment index


fm3:
When the server is busy, we see some volume update requests becoming very slow (>10m). Can we do something to improve performance?

fm3:
I’d consider this fixed by https://github.com/scalableminds/webknossos/pull/8460 https://github.com/scalableminds/webknossos/pull/8482 and https://github.com/scalableminds/webknossos/pull/8489 (also vaguely related is the upcoming https://github.com/scalableminds/webknossos/pull/8469 )

Of course we can always optimize further, let’s do so if we notice more slow requests.

PR #8429, opened 2025-03-05
incrase threshold for slow request logging from 30s to 2min

fm3:
Our channel gets a little too many notifications. We want to get below 30s of course, most importantly with #8404 but until then let’s reduce the amount of messages with a higher threshold.


hotzenklotz (approved):

MichaelBuessemeyer (approved):
Just stumbled upon this. LGTM :+1:

Issue #8421, opened 2025-03-03
Fix volume annotation tools for rotated datasets


MichaelBuessemeyer:
## Context

Since #8159 it is possible to rotate datasets. The rotation can be toggled on and off. Annotating is only allowed when the rotation is turned off. But if it is turned off by clicking on a color layer and not the volume layer the volume tools are disabled, although they should be enabled.

## Expected Behavior

The volume tools should be enabled.
## Current Behavior

The volume tools are not enabled

## Steps to Reproduce the bug

- [ ] Cannot reproduce the bug anymore / needs deeper investigation.

1.Perma rotate a dataset via the dataset settings.
2.open an annotation for that dataset
3. turn on (if not already done) the rotation
4. turn off the rotation via a color layer. The volume tools should still be disabled although the layer is rendered natively.

## Your Environment for bug

- nothing special Chrome + Linux

### Current work around
- Toggle of the rotation on the currently visible segmentation layer. This way volume annotation is not blocked by the ui

PR #8432, opened 2025-03-06
Fix volume annotation tool availability on datasets with rotation transforms

MichaelBuessemeyer:
This PR makes volume annotation tools properly available for datasets with rotation transforms where the dataset rotation is turned off but not the active volume layer is the selected "nativelyRenderedLayerName".

The bug: Previously, when having an annotation on a dataset with rotation transforms and deactivating rendering the DS transformed via the color layer, the volume tools were not enabled as the frontend thought that the volume layer was still rendered with transforms. This PR fixes this by letting `getTransformsForLayerOrNull` return the `IdentityTransform` object for all transformations that are equal to the identity transform. Thus, making the `===` or `!==` comparisons with the `IdentityTransform`  now correct :innocent:.

- https://___.webknossos.xyz

### Steps to test:
- Open a DS with rotation transforms enabled
- Toggle the rotation transforms on the color layer
- When the transformations are turned off, not only the skeleton tool but all volume tools should be available.
- Check that the transform icons are also rendered correctly (as the check was also modified in this pr)
- Compare with master -> There the volume tools should be disabled if doing the steps above.

### Issues:
- fixes https://github.com/scalableminds/webknossos/issues/8421

------


frontend/javascripts/oxalis/view/left-border-tabs/layer_settings_tab.tsx
```diff
@@ -208,7 +207,7 @@ function TransformationIcon({ layer }: { layer: APIDataLayer | APISkeletonLayer
   if (!showIcon) {
     return null;
   }
-  const isRenderedNatively = transform == null || isIdentityTransform(transform);
+  const isRenderedNatively = transform == null || transform === IdentityTransform;
```
philippotto, line 210: this is not the actual fix, right? I assume there are other places where the `=== IdentityTransform` is done and that didn't work, because `isIdentityTransform` would have been required. instead of adapting these places, you adapted the accessors to always return the IdentityTransform if possible, so that === works. is this correct? I just want to check whether I got it right :)

MichaelBuessemeyer: > this is not the actual fix, right?

Yes its not. Its just not necessary anymore to use the `isIdentityTransform` comparison function here which does a full matrix comparison while the `=== IdentityTransform` simply makes and "instance equality check". So this "correction" here is simply a mini optimization as due to the `_getTransformsForLayerOrNull` is now using `toIdentityTransformMaybe` to ensure in case the resulting matrix is the identity matrix a reference to `isIdentityTransform` is returned.

> I assume there are other places where the === IdentityTransform is done

Not really. Only in the [nml_helper](https://github.com/scalableminds/webknossos/blob/fd0eb210d03f55976f6dcd8ce0fc606b2528038d/frontend/javascripts/oxalis/model/helpers/nml_helpers.ts#L319). And from 

> that didn't work, because isIdentityTransform would have been required.

Yes, indeed. But I wanted to optimize here a little by allowing an instance comparison in the most part of the code base and only using the manual matrix comparison in the cached function of `dataset_layer_transformation_accessor.ts`

> instead of adapting these places, you adapted the accessors to always return the IdentityTransform if possible, so that === works. is this correct? I

Yes, exactly :tada:

frontend/javascripts/oxalis/model/accessors/dataset_layer_transformation_accessor.ts
```diff
@@ -239,7 +243,7 @@ export function getTransformsForLayer(
   );
 }
 
-export function isIdentityTransform(transform: Transform) {
+function isIdentityTransform(transform: Transform) {
```
philippotto, line 246: how about renaming this to `equalsIdentityTransform` to make it clearer that it checks for equality (because identity can now always be checked with ===).

philippotto (approved):
awesome! only see my one suggestion :)

PR #8437, opened 2025-03-10
When deleting layers, move referenced data

### LINKED ISSUE #8279: Delete layers that are referenced by other datasets
## Context
Symlinks are used to reuse layers in new datasets. When the original dataset is deleted from within wk, these symlinks are broken, making the new datasets broken.

## Expected Behavior
When a dataset is deleted that has layers that other datasets reference via symlinks, these layers should be moved to another dataset and the symlinks repaired. I think it is fine to restrict the symlink search to an organization.

frcroth:
- https://___.webknossos.xyz

### Steps to test:
- For all scenarios create datasets and make sure they have been scanned and reported to WK.
- Layers (all this should work for fully linked layers, so it should not matter if the layer itself is a link or all mags in the layer are linked), note that for locally explored layers you need local non-wkw data
	- DS 1 has Layer A with local data, DS 2 has Layer B with symlink to A. Delete DS 1. Layer B now has local data.
	- DS 1 has Layer A with local data, DS 2 has Layer B with symlink to A, DS 3 has Layer C with symlink to A. Delete DS 1. Layer B (without restriction of generality) has local data, Layer C has symlink to B.
	- DS 1 has Layer A with local data, DS 2 has Layer B locally explored (file:// protocol) with link to A. Delete DS 1. Layer B now exists physically with local data, there is no remote link anymore.
	- DS 1 has layer A with local data, DS 2 has Layer B locally explored with link to A, DS 3 has Layer C with symlink to A. Delete DS 1. **it is currently arbitrary which ds is selected to store local data** If DS 2 is selected, this is now a regular dataset and C is symlink to B, otherwise DS 2 now symlinks to C. In any case, DS 2 is no longer a remote dataset with file protocol.
	- DS 1 has Layer A with local data, DS 2 has Layer B, which exists physically, but every mag in layer B is a symlink to the mag in Layer A. Delete DS 1. Layer B now has local data.
- Mags
	- DS 1 has Layer A with local data. DS 2 has Layer B where one mag is a symlink to a mag of A. Delete DS 1. This mag has now been moved to B.
	- DS 1 has Layer A with local data. DS 2 has Layer B where one mag is a symlink to a mag M of A. DS 3 has layer C where one mag is a symlink to mag M. Delete DS 1. Mag M has been moved to B or C, the other has a symlink to that location.

### Notes

When a layer A symlinks to another layer B that links to a layer C, and you remove layer B, the symlink at layer A is broken. That is because for storage in the DB all symlinks are completely resolved, so A real Path is C and B real Path is C, when B is removed there is no knowledge on the link from A. I hope this works for now. Solving this would probably require a new table only for paths, because every length of link-links need to be resolved.

~~Another problem currently is locally explored datasets (with file://). Because we can't rewrite symlinks there, I think we need to rewrite the datasource properties in this case.~~ Done

Right now everything assumes the usual structure with layers and mags named as usual. Anything else will cause trouble.

### TODOs:

All these cases need to be handled
- [x] Layer with symlinks to it
	- [x] One layer is a symlink -> move layer here
	- [x] Multiple layers with symlinks -> move layer to one, relink everything else
	- [x] (Multiple) layers with symlinks, but also a layer where one mag symlinks
- [x] Mag with symlinks to it
- [x] Locally explored

### Issues:
- fixes #8279 

------


webknossos-datastore/app/com/scalableminds/webknossos/datastore/helpers/DatasetDeleter.scala
```diff
@@ -31,23 +39,244 @@ trait DatasetDeleter extends LazyLogging with DirectoryConstants {
...
+                    dataSourcePath: Path,
+                    reason: Option[String]): Fox[Unit] =
+      if (Files.exists(dataSourcePath)) {
+        val trashPath: Path = dataBaseDir.resolve(organizationId).resolve(trashDir)
+        val targetPath = trashPath.resolve(datasetName)
+        new File(trashPath.toString).mkdirs()
+
+        logger.info(s"Deleting dataset by moving it from $dataSourcePath to $targetPath${if (reason.isDefined)
+          s" because ${reason.getOrElse("")}"
+        else "..."}")
```
fm3, line 53:
```suggestion
        logger.info(s"Deleting dataset by moving it from $dataSourcePath to $targetPath${reason.map(r => s"because $r").getOrElse("...")}")
```

webknossos-datastore/app/com/scalableminds/webknossos/datastore/helpers/DatasetDeleter.scala
```diff
@@ -31,23 +39,244 @@ trait DatasetDeleter extends LazyLogging with DirectoryConstants {
         case e: Exception                                => Fox.failure(s"Deleting dataset failed: ${e.toString}", Full(e))
       }
 
+    def moveToTrash(organizationId: String,
```
fm3, line 42: we should probably make everything in here either private or protected, unless it’s really necessary for it to be public.

webknossos-datastore/app/com/scalableminds/webknossos/datastore/services/BinaryDataService.scala
```diff
@@ -22,7 +22,8 @@ class BinaryDataService(val dataBaseDir: Path,
                         val agglomerateServiceOpt: Option[AgglomerateService],
                         remoteSourceDescriptorServiceOpt: Option[RemoteSourceDescriptorService],
                         sharedChunkContentsCache: Option[AlfuCache[String, MultiArray]],
-                        datasetErrorLoggingService: Option[DatasetErrorLoggingService])(implicit ec: ExecutionContext)
+                        datasetErrorLoggingService: Option[DatasetErrorLoggingService],
+                        val remoteWKClient: Option[DSRemoteWebknossosClient])(implicit ec: ExecutionContext)
```
fm3, line 26: is it really necessary for the binarydataservice to extend the datasetdeleter? I get the feeling it shouldn’t be involved in deletions. If not, this can also be removed again, and the client could be made non-optional in the deleter?

app/models/dataset/DatasetService.scala
```diff
@@ -356,15 +357,17 @@ class DatasetService @Inject()(organizationDAO: OrganizationDAO,
       _ <- Fox.serialCombined(pathInfos)(updateRealPath)
     } yield ()
 
-  def getPathsForDataLayer(datasetId: ObjectId, layerName: String): Fox[List[(DatasetMagInfo, List[DatasetMagInfo])]] =
+  def getPathsForDataLayer(datasetId: ObjectId,
+                           layerName: String): Fox[List[(DataSourceMagInfo, List[DataSourceMagInfo])]] =
```
fm3, line 361: Could youu add a comment explaining the signature of this function? I think it’s not immediately obvious why the return type is this nested list

webknossos-datastore/app/com/scalableminds/webknossos/datastore/helpers/DatasetDeleter.scala
```diff
@@ -31,23 +39,244 @@ trait DatasetDeleter extends LazyLogging with DirectoryConstants {
...
+
+  private def moveSymlinks(organizationId: String, datasetName: String)(implicit ec: ExecutionContext) =
+    for {
+      dataSourceId <- Fox.successful(DataSourceId(datasetName, organizationId))
+      layersAndLinkedMagsOpt <- Fox.runOptional(remoteWKClient)(_.fetchPaths(dataSourceId))
+      exceptionBoxes = layersAndLinkedMagsOpt match {
+        case Some(layersAndLinkedMags) =>
+          layersAndLinkedMags.map(layerMagLinkInfo =>
+            handleLayerSymlinks(dataSourceId, layerMagLinkInfo.layerName, layerMagLinkInfo.magLinkInfos.toList))
+        case None => Seq(tryo {})
```
fm3, line 82: what does the `tryo {}` return? maybe we could make this more explicit?

webknossos-datastore/app/com/scalableminds/webknossos/datastore/helpers/DatasetDeleter.scala
```diff
@@ -31,23 +39,244 @@ trait DatasetDeleter extends LazyLogging with DirectoryConstants {
...
+      val propertiesPath = dataBaseDir
+        .resolve(dataSourceId.organizationId)
+        .resolve(dataSourceId.directoryName)
+        .resolve(GenericDataSource.FILENAME_DATASOURCE_PROPERTIES_JSON)
+      if (Files.exists(propertiesPath)) {
+        JsonHelper.validatedJsonFromFile[DataSource](propertiesPath, dataBaseDir) match {
+          case Full(dataSource) =>
+            val updatedDataSource = dataSource.copy(dataLayers = dataSource.dataLayers.map {
+              case dl: DataLayerWithMagLocators =>
+                if (dl.mags.forall(_.path.exists(_.startsWith("file://")))) {
```
fm3, line 132:
```suggestion
                if (dl.mags.forall(_.path.exists(_.startsWith(s"${DataVaultService.schemeFile}://")))) {
```

That way this will be listed as a usage of this file scheme.

webknossos-datastore/app/com/scalableminds/webknossos/datastore/helpers/DatasetDeleter.scala
```diff
@@ -31,23 +39,244 @@ trait DatasetDeleter extends LazyLogging with DirectoryConstants {
...
+                  // Setting path to None means using resolution of layer/mag directories to access data
+                  dl.mapped(magMapping = _.copy(path = None))
+                } else {
+                  dl
+                }
+              case dl => dl
+            })
+            // Write properties back
+            Files.delete(propertiesPath)
```
fm3, line 141: might throw an exception, maybe we should tryo this too.

fm3 (changes requested):
Wow, quite the complex logic :see_no_evil: But it does seem justified, I don’t see easy ways to simplify.

I’m a little fearful of exceptions, could you have a look which of the used file functions can throw exceptions and add tryos there?

Also, maybe it makes sense to first check for write access so that a readable error message can be given. Also, what happens if for permission reasons this resymlinking does not happen. Can the deletion still go forward? What is shown to the user?

I also added a few small comments on the code.

Did no testing yet, will do that in the next round.

webknossos-datastore/app/com/scalableminds/webknossos/datastore/helpers/DatasetDeleter.scala
frcroth: I don't think I wrote this but happy to change  :smile:

webknossos-datastore/app/com/scalableminds/webknossos/datastore/helpers/DatasetDeleter.scala
frcroth: this is not private because it is defined in the deleteOnDisk method, apart from deleteOnDisk there are no non-private methods here.

webknossos-datastore/app/com/scalableminds/webknossos/datastore/services/BinaryDataService.scala
frcroth: Good idea!

webknossos-datastore/app/com/scalableminds/webknossos/datastore/helpers/DatasetDeleter.scala
frcroth: Removed because remoteWKClient is no longer optional

frcroth:
> Wow, quite the complex logic 🙈 But it does seem justified, I don’t see easy ways to simplify.
> 
> I’m a little fearful of exceptions, could you have a look which of the used file functions can throw exceptions and add tryos there?

I added the suggestion tryo for updateDatasourceProperties deletion, everything else should be covered by the big tryo in handleLayerSymlinks.

> Also, maybe it makes sense to first check for write access so that a readable error message can be given. Also, what happens if for permission reasons this resymlinking does not happen. Can the deletion still go forward? What is shown to the user?

Added an exception before deleting stuff.

frcroth:
Place this script into sample_organization to generate test datasets to test scenarios (not testing locally explored layers)

```sh
#!/bin/bash

set -e  # Exit on error

if [ "$#" -ne 1 ]; then
    echo "Usage: $0 <dataset_directory>"
    exit 1
fi

SOURCE_DIR="$1"
BASE_DIR="$(dirname "$SOURCE_DIR")"

# Function to copy dataset while maintaining symlinks
copy_test_dataset() {
    local src="$1"
    local dest="$2"
    mkdir -p "$dest"
    cp -a "$src"/. "$dest"  # Preserve symlinks and structure
    cp "$src/datasource-properties.json" "$dest/"  # Copy properties file
}

# Test Case 1: DS1 has color with local data, DS2 has color symlinked to DS1
DS1_TC1="$BASE_DIR/DS1_TC1"
DS2_TC1="$BASE_DIR/DS2_TC1"
copy_test_dataset "$SOURCE_DIR" "$DS1_TC1"
mkdir -p "$DS2_TC1"  # Ensure parent directory exists
ln -s "../DS1_TC1/color" "$DS2_TC1/color"
cp "$DS1_TC1/datasource-properties.json" "$DS2_TC1/"  # Copy properties file

# Test Case 2: DS1_TC2 has color with local data, DS3_TC2 has color symlinked to DS1_TC2
DS1_TC2="$BASE_DIR/DS1_TC2"
DS2_TC2="$BASE_DIR/DS2_TC2"
DS3_TC2="$BASE_DIR/DS3_TC2"
copy_test_dataset "$SOURCE_DIR" "$DS1_TC2"
mkdir -p "$DS3_TC2"  # Ensure parent directory exists
ln -s "../DS1_TC2/color" "$DS3_TC2/color"
cp "$DS1_TC2/datasource-properties.json" "$DS3_TC2/"  # Copy properties file
mkdir -p "$DS2_TC2"  # Ensure parent directory exists
ln -s "../DS1_TC2/color" "$DS2_TC2/color"
cp "$DS1_TC2/datasource-properties.json" "$DS2_TC2/"  # Copy properties file

# Test Case 3: DS1_TC3 has color with local data, DS2_TC3 has color with mags symlinked to DS1_TC3
DS1_TC3="$BASE_DIR/DS1_TC3"
DS2_TC3="$BASE_DIR/DS2_TC3"
copy_test_dataset "$SOURCE_DIR" "$DS1_TC3"
mkdir -p "$DS2_TC3/color"
for mag in "$DS1_TC3/color/"*; do
    ln -s "../../$mag" "$DS2_TC3/color/$(basename "$mag")"
done
cp "$DS1_TC3/datasource-properties.json" "$DS2_TC3/"  # Copy properties file

# Test Case 4: DS1_TC4 has color with local data, DS2_TC4 has one mag symlinked to DS1_TC4
DS1_TC4="$BASE_DIR/DS1_TC4"
DS2_TC4="$BASE_DIR/DS2_TC4"
copy_test_dataset "$SOURCE_DIR" "$DS1_TC4"
mkdir -p "$DS2_TC4/color"
ln -s "../../$DS1_TC4/color/1" "$DS2_TC4/color/1"
cp "$DS1_TC4/datasource-properties.json" "$DS2_TC4/"  # Copy properties file

# Test Case 5: DS1_TC5 has color with local data, DS2_TC5 and DS3_TC5 both symlink the same mag
DS1_TC5="$BASE_DIR/DS1_TC5"
DS2_TC5="$BASE_DIR/DS2_TC5"
DS3_TC5="$BASE_DIR/DS3_TC5"
copy_test_dataset "$SOURCE_DIR" "$DS1_TC5"
mkdir -p "$DS2_TC5/color" "$DS3_TC5/color"
ln -s "../../$DS1_TC5/color/1" "$DS2_TC5/color/1"
ln -s "../../$DS1_TC5/color/1" "$DS3_TC5/color/1"
cp "$DS1_TC5/datasource-properties.json" "$DS2_TC5/"  # Copy properties file
cp "$DS1_TC5/datasource-properties.json" "$DS3_TC5/"  # Copy properties file

echo "Test datasets created in $BASE_DIR"
echo "To test deletion scenarios, remove the DS1 datasets"

``

File: webknossos-datastore/app/com/scalableminds/webknossos/datastore/helpers/DatasetDeleter.scala [hunk 8]
```diff
@@ -31,23 +41,264 @@ trait DatasetDeleter extends LazyLogging with DirectoryConstants {
...
+          linkedMags.foreach { magLinkInfo =>
+            val magToDelete = magLinkInfo.mag
+            if (magLinkInfo.linkedMags.nonEmpty) {
+              if (magToDelete.hasLocalData) {
+                // Move mag to a different dataset
+                val magPath = dataBaseDir
+                  .resolve(dataSourceId.organizationId)
+                  .resolve(dataSourceId.directoryName)
+                  .resolve(layerName)
+                  .resolve(magToDelete.mag.toMagLiteral(true))
```
fm3, line 267: Should probably also try with allowScalar=false. If both variants don’t exist, throw exception.

webknossos-datastore/app/com/scalableminds/webknossos/datastore/helpers/DatasetDeleter.scala
```diff
@@ -31,23 +41,264 @@ trait DatasetDeleter extends LazyLogging with DirectoryConstants {
...
+                // Select an arbitrary linked mag to move to
+                val target = magLinkInfo.linkedMags.head
```
fm3, line 269: another `.head` which would lead to non-readable exception. maybe add nonEmpty assertion

fm3 (commented):
Works well, thanks for the setup script!

I like the verbose logging for this stuff, that way we can trace where anything was moved.

I also provoked some errors and I think the handling is ok. (I would prefer smaller tryo units over these big ones, to benefit more from error chaining, but it’s ok like this for the moment)

I added a few more code comments.

util/src/main/scala/com/scalableminds/util/mvc/Formatter.scala
```diff
@@ -108,7 +114,8 @@ trait Formatter {
       }
 
     val serverTimeMsg = if (includeTime) s"[Server Time ${Instant.now}] " else ""
-    serverTimeMsg + formatOneFailure(failure) + formatStackTrace(failure) + formatNextChain(failure.chain)
+    serverTimeMsg + firstException(failure) + formatOneFailure(failure) + formatStackTrace(failure) + formatNextChain(
```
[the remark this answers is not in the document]
frcroth: Yes

```
"java.lang.Error: this is bad: problem Stack trace: java.lang.Error: this is bad
	at <init>(<console>:1)
	at <init>(<console>:6)
	at .<clinit>(<console>:1)
	at .lzycompute(<synthetic>:6)
	at .$print(<synthetic>:5)
	at $print(<synthetic>)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:566)
	at scala.tools.nsc.interpreter.IMain$ReadEvalPrint.call(IMain.scala:678)
	at scala.tools.nsc.interpreter.IMain$Request.loadAndRun(IMain.scala:1027)
	at scala...
>....
```

Should the firstException only be logged when includeStackTraces = false?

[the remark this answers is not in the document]
fm3: I mean it’s not very important, but seems like this would be nicer, yes :)

webknossos-datastore/app/com/scalableminds/webknossos/datastore/helpers/DatasetDeleter.scala
frcroth: Removed the duplication. The different possible mag paths generally make more parts difficult, I will look into it.

frcroth: Handled in 14f8a82965acb236e37f18270b857811d8cb090d. This adds another dimension to test.

fm3 (approved):
Alright let’s go :) Please wait with merging until the current CI problems are resolved, hopefully today.

PR #8439, opened 2025-03-10
Redo: Fix volume annotation tool availability on datasets with rotation transforms

MichaelBuessemeyer:
Redo PR for https://github.com/scalableminds/webknossos/pull/8432 including the fix for report: https://scm.slack.com/archives/C02H5T8Q08P/p1741624272851439

Please see my comments for details on the problem & fix

### How to test:
- Same as before
- In a normal untransformed dataset open a hybrid annotation. Check that all tools are available and not "transformation" is blocking the tools.

Other PRs description:
> This PR makes volume annotation tools properly available for datasets with rotation transforms where the dataset rotation is turned off but not the active volume layer is the selected "nativelyRenderedLayerName".
> 
> The bug: Previously, when having an annotation on a dataset with rotation transforms and deactivating rendering the DS transformed via the color layer, the volume tools were not enabled as the frontend thought that the volume layer was still rendered with transforms. This PR fixes this by letting `getTransformsForLayerOrNull` return the `IdentityTransform` object for all transformations that are equal to the identity transform. Thus, making the `===` or `!==` comparisons with the `IdentityTransform` now correct 😇.
> 
> ### URL of deployed dev instance (used for testing):
> * https://___.webknossos.xyz
> 
> ### Steps to test:
> * Open a DS with rotation transforms enabled
> * Toggle the rotation transforms on the color layer
> * When the transformations are turned off, not only the skeleton tool but all volume tools should be available.
> * Check that the transform icons are also rendered correctly (as the check was also modified in this pr)
> * Compare with master -> There the volume tools should be disabled if doing the steps above.
> 
> ### Issues:
> * fixes [Fix volume annotation tools for rotated datasets #8421](https://github.com/scalableminds/webknossos/issues/8421)
> 
> (Please delete unneeded items, merge only when none are left open)
> 
> * [x]  Updated [changelog](../blob/master/CHANGELOG.unreleased.md#unreleased)

### DIFF
--- a/frontend/javascripts/oxalis/model/accessors/dataset_layer_transformation_accessor.ts
+++ b/frontend/javascripts/oxalis/model/accessors/dataset_layer_transformation_accessor.ts
@@ -191,6 +191,10 @@ export function isLayerWithoutTransformationConfigSupport(layer: APIDataLayer | 
   );
 }
 
+function toIdentityTransformMaybe(transform: Transform | null): Transform {
+  return transform && !equalsIdentityTransform(transform) ? transform : IdentityTransform;
+}
+
 function _getTransformsForLayerOrNull(
   dataset: APIDataset,
   layer: APIDataLayer | APISkeletonLayer,
@@ -210,7 +214,7 @@ function _getTransformsForLayerOrNull(
   const layerTransforms = getOriginalTransformsForLayerOrNull(dataset, layer as APIDataLayer);
   if (nativelyRenderedLayerName == null) {
     // No layer is requested to be rendered natively. -> We can use the layer's transforms as is.
-    return layerTransforms;
+    return toIdentityTransformMaybe(layerTransforms);
   }
 
   // Apply the inverse of the layer that should be rendered natively
@@ -221,11 +225,11 @@ function _getTransformsForLayerOrNull(
   if (transformsOfNativeLayer == null) {
     // The inverse of no transforms, are no transforms. Leave the layer
     // transforms untouched.
-    return layerTransforms;
+    return toIdentityTransformMaybe(layerTransforms);
   }
 
   const inverseNativeTransforms = invertTransform(transformsOfNativeLayer);
-  return chainTransforms(layerTransforms, inverseNativeTransforms);
+  return toIdentityTransformMaybe(chainTransforms(layerTransforms, inverseNativeTransforms));
 }
 
 export const getTransformsForLayerOrNull = memoizeWithThreeKeys(_getTransformsForLayerOrNull);
@@ -239,7 +243,7 @@ export function getTransformsForLayer(
   );
 }
 
-export function isIdentityTransform(transform: Transform) {
+function equalsIdentityTransform(transform: Transform) {
   return transform.type === "affine" && _.isEqual(transform.affineMatrix, Identity4x4);
 }
 
@@ -266,7 +270,7 @@ function _getTransformsForLayerThatDoesNotSupportTransformationConfigOrNull(
     const someLayersTransformsMaybe = usableReferenceLayer
       ? getTransformsForLayerOrNull(dataset, usableReferenceLayer, nativelyRenderedLayerName)
       : null;
-    return someLayersTransformsMaybe;
+    return toIdentityTransformMaybe(someLayersTransformsMaybe);
   } else if (nativelyRenderedLayerName != null && allLayersSameRotation) {
     // If all layers have the same transformations and at least one is rendered natively, this means that all layer should be rendered natively.
     return null;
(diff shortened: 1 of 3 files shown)


MichaelBuessemeyer (commented):
The bug occured due to `getTransformsForLayerThatDoesNotSupportTransformationConfigOrNull` returning the `IdentityTransform` even in case no layer had transformations. This is due to `getTransformsForLayerThatDoesNotSupportTransformationConfigOrNull` not respecting the postfix "...OrNull" in its name anymore. 

I fixed the bug from 2 perspectives:
1. I adjusted the `toIdentityTransformMaybe` to keep null values so that the transformation helper / accessor methods ending with "...OrNull" can still return null in case of null transformations.
2. I fixed `isSkeletonLayerTransformed` to treat the IdentityTransform also as "the skeleton layer has no transformations active" to prevent similar errors to the linked one in the pr description.

See my comments for more details



Moreover, I checked for additional null comparisons with outputs of transformation accessor functions but could not find any relevant (assuming the `...OrNull` functions do return null in case no transformations are configured)

frontend/javascripts/oxalis/model/accessors/dataset_layer_transformation_accessor.ts
```diff
@@ -285,7 +284,7 @@ function _getTransformsForLayerThatDoesNotSupportTransformationConfigOrNull(
     return null;
   }
 
-  return toIdentityTransformMaybe(invertTransform(transformsOfNativeLayer));
```
philippotto: are you sure that `toIdentityTransformMaybe` does not need to be called? I'm wondering because `invertTransform` always returns a new object, but maybe `toIdentityTransformMaybe` happens on the call site? to be honest, it is a bit hard for me to follow where exactly the identity-stuff happens. however, I don't have a great idea for how to improve this.

philippotto (commented):
The fix seems good :+1: I commented one spot where I was unsure about the change.

Could you add a test case for the `getDisabledInfoForTools` function so that we can ensure that untransformed datasets/layers have an enabled skeleton tool (and vice versa)? I feel like this is quite an important logic which should be tested :)
However, I'm not sure how easy it is. I see that `getDisabledInfoForTools` is mocked in `annotation_tool_saga.spec.ts`. This is usually a sign that there is some trouble (probably because it also depends on viewport sizes etc). 
Maybe it is easier to test `isSkeletonLayerTransformed` directly? Let me know if it's too much of a hassle.

frontend/javascripts/oxalis/model/accessors/dataset_layer_transformation_accessor.ts
MichaelBuessemeyer: Oh you are totally correct. `toIdentityTransformMaybe` should be kept here :+1:

MichaelBuessemeyer:
Ok tests for this should work now. Wasn't actually that much of a problem. The main issue was that there was no fixture for a hybrid tracing. I added this now and added 4 tests for the `getDisabledInfoForTools`. A nice bonus it that the test also checks whether the `nativelyRenderedLayerName` is correctly interpreted and that all layers having the same rotation and one being set as rendered natively leads to the skeleton (and all other layers as well) being rendered natively.

What do you think about the tests :)?

frontend/javascripts/test/fixtures/hybridtracing_object.ts
```diff
@@ -0,0 +1,103 @@
...
+import { MISSING_GROUP_ID } from "oxalis/view/right-border-tabs/trees_tab/tree_hierarchy_view_helpers";
+import { TreeTypeEnum } from "oxalis/constants";
+import type { APIColorLayer } from "types/api_flow_types";
+
+mockRequire("app", {
+  currentUser: {
+    firstName: "SCM",
+    lastName: "Boy",
+  },
+});
```
philippotto, line 17: I quickly dug into the rabbit hole and I think this hasn't been necessary since 2017 :sweat_smile: could you remove it here and everywhere else in the repo and check that everything still works?

frontend/javascripts/test/sagas/annotation_tool_disabled_info.spec.ts
```diff
@@ -0,0 +1,181 @@
...
+  for (const toolName in AnnotationToolEnum) {
+    if (toolsDisregardingRotation.includes(toolName as AnnotationToolEnum)) {
+      t.assert(disabledInfo[toolName as AnnotationToolEnum]?.isDisabled === false);
+    } else {
+      t.assert(disabledInfo[toolName as AnnotationToolEnum]?.isDisabled === true);
+    }
+  }
+});
+
+test("Tools should not be  disabled when dataset is rotation is toggled off", (t) => {
```
philippotto, line 167:
```suggestion
test("Tools should not be disabled when dataset rotation is toggled off", (t) => {
```

frontend/javascripts/test/fixtures/hybridtracing_object.ts
MichaelBuessemeyer: still works :D

MichaelBuessemeyer:
So maybe now :green_apple: ?

philippotto (changes requested):
🎉

philippotto (approved):
🎉

PR #8444, opened 2025-03-12
Fix "animation modal" not showing

### LINKED ISSUE #8189: "Create animation" menu entry available but does not do anything
## Context
[Screencast from 13.11.2024 11:21:35.webm](https://github.com/user-attachments/assets/42f22aaf-108f-4a3f-8e19-8118f7f10a3f)

## Expected Behavior
There should be an error message (because I can not do it for a different org) or the menu entry should not be available.

## Current Behavior
Nothing happens

## Steps to Reproduce the bug
1. Have a dataset in org 1, make it public
2. Have a user in org 2 open the dataset
3. Click the Create Animation button

## Your Environment for bug

- Browser name and version: Chromium Version 130.0.6723.69 (Official Build) snap (64-bit)
- Version of WEBKNOSSOS (Release or Commit): 015016bfddebf414ad1ba76bd16125efe355e291

hotzenklotz:
In view mode, the navbar menu entry for "Create Animation" is shown to every user, regardless of there permissions/role/admin status.

The corresponding modal React component, however, is only mounted for super users. In other words, the modal could not be open/displayed for the majority of our user base.

### Steps to test:
- Make sure to open a dataset in 1) view mode and 2) as non-superadmin user.
- Select "Create Animation" from the "Menu" in the navbar
- Modal should appear

### Issues:
- fixes #8189

------


philippotto (commented):
so, only super users have been able to render animations? git blame shows that this isSuperUser part was added in the original pr #7348, but I thought I saw external users creating animations :thinking: then, I must have dreamt that?

--------

regarding the fix:
- I tested this on the dev instance and there I can open the modal, but the "submit" button is disabled (see screenshot). the button does not explain why it's disabled. I'd change the button text to "Start Animation" and put a tooltip on it à la "This feature is not enabled on this webknossos instance. please visit webknossos.org".
- I can open this modal without being logged in now. if the job was enabled on the dev instance, I would probably be able to start the job (not sure whether the backend denies this cc @MichaelBuessemeyer). this should not be possible imo.

image

hotzenklotz:
> so, only super users have been able to render animations? git blame shows that this isSuperUser part was added in the original pr https://github.com/scalableminds/webknossos/pull/7348, but I thought I saw external users creating animations 🤔 then, I must have dreamt that?

In "annotaotion" mode, everyone was able to start animations. The bug occurs only in "viwe" mode.

philippotto:
> > so, only super users have been able to render animations? git blame shows that this isSuperUser part was added in the original pr #7348, but I thought I saw external users creating animations 🤔 then, I must have dreamt that?
> 
> In "annotaotion" mode, everyone was able to start animations. The bug occurs only in "viwe" mode.

ah, thanks for clearing this up :+1:

hotzenklotz:
> put a tooltip on it 

Done.

> I can open this modal without being logged in now. if the job was enabled on the dev instance, I would probably be able to start the job (not sure whether the backend denies this cc @MichaelBuessemeyer). this should not be possible imo.

Yes, one can click on the "Start Animation" button but the backend refuses the API call as unauthorized. It is not the prettiest thing in the world, but at least it prevents any mischief.

Screenshot 2025-03-17 at 13 36 40

philippotto:
Yes, one can click on the "Start Animation" button but the backend refuses the API call as unauthorized. It is not the prettiest thing in the world, but at least it prevents any mischief.

Could you add a simple `activeUser == null` check to the disabled prop of the button?

philippotto (approved):
:tada:

Issue #8427, opened 2025-03-04
Fix up CSS related problems after antd update


philippotto:
There are a couple of CSS regressions that need to be fixed.

https://scm.slack.com/archives/C5AKLAV0B/p1741015478851899
https://scm.slack.com/archives/C5AKLAV0B/p1740727345290209?thread_ts=1740727307.827319&amp;cid=C5AKLAV0B
https://scm.slack.com/archives/C5AKLAV0B/p1739275386286009

PR #8445, opened 2025-03-12
Fix color theme for modals

hotzenklotz:
In this PR, the light/dark color theme for modals available from the "Menu" in the navbar are fixed, e.g. "Download", "Create Animation", "Share" etc.

By default, the navbar is always set to the dark theme. All it's React component children, i.e. all modals are also forced into dark mode. This PR applies correct color theming based on the user settings.

- https://___.webknossos.xyz

### Steps to test:
- Switch between light and dark mode
- Any modal in from the navbar menu should follow suite
- Make sure to test "view" and "annotation" mode

### TODOs:
- [x] Investigate some wrong color in light-mode 

### Issues:
- Contributes to #8427

------


philippotto (approved):
works great :+1:

PR #8450, opened 2025-03-17
Fix click targets in dataset table

normanrz:
Fixes the click targets in the dataset table for selecting datasets.


philippotto (approved):
:+1:

PR #8452, opened 2025-03-17
Fix CSS spacing for actions in Dashboard > Tasks

hotzenklotz:
Fixed the CSS spacing for the task actions in the Dashboard > Tasks view. The same issue is also present in some of the other list views that have a similar "Actions" column.

High-level the underlying issue has to do with several factors:
- a mix of components representing each action (`<span>`, `<ButtonLink>`, `<a>`, `<AsyncLink`) with different stylings
- A mixed use of the `<ButtonLink>` component. Sometimes the icon was passed a prop and had the default antd spacing applied. Other times, the icon was wrapped as child with it's on margin-right. (And worse case, you do both and end up with twice the right spacing :-/ )

Screenshot 2025-03-17 at 16 02 34

- https://___.webknossos.xyz

### Steps to test:
- Create a new task
- Go to Dashboard -> Task and request a task
- Action should neatly line up

### Issues:
- contributes to #8427

------

### DIFF
--- a/frontend/javascripts/dashboard/dashboard_task_list_view.tsx
+++ b/frontend/javascripts/dashboard/dashboard_task_list_view.tsx
@@ -232,7 +232,7 @@ class DashboardTaskListView extends React.PureComponent<Props, State> {
     );
     return task.annotation.state === "Finished" ? (
       <div>
-        <CheckCircleOutlined className="icon-margin-right"/>
+        <CheckCircleOutlined className="icon-margin-right" />
         Finished
       </div>
     ) : (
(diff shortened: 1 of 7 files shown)


frontend/javascripts/dashboard/dashboard_task_list_view.tsx
```diff
@@ -226,37 +226,25 @@ class DashboardTaskListView extends React.PureComponent<Props, State> {
...
     return task.annotation.state === "Finished" ? (
       <div>
-        <CheckCircleOutlined className="icon-margin-right" />
+        <CheckCircleOutlined className="icon-margin-right"/>
         Finished
-        <br />
       </div>
     ) : (
-      <div>
+      <Space direction="vertical" size={1}>
```
hotzenklotz, line 239: IMHO this is cleaner than the `<br>` tags.

philippotto (approved):
muchas gracias!

PR #8459, opened 2025-03-20
Allow histogram editing even if it couldn't be loaded

philippotto:
Normal histogram (as usual):
image

Failed histogram:
image

- https://robusthistogram.webknossos.xyz

### Steps to test:
- open a DS with dev tools open -> histogram should load as usual
- rightclick the histogram request in the network tab and block it with "Block request url"
- in the layer menu, click on "reload data"
- now the histogram should show up like in the picture above. you should still be able to tweak the values
- unblock the request again and click on "reload" -> everything should work as usual

### Issues:
- fixes https://scm.slack.com/archives/C5AKLAV0B/p1742322460170389

------


daniel-wer (approved):
Thanks so much, this will make my life much easier :pray: 

I like the idea of the dummy histogram but for dtypes other than uint8 it is not yet scaled correctly. See this screenshot for the float test dataset:

image

philippotto:
> I like the idea of the dummy histogram but for dtypes other than uint8 it is not yet scaled correctly. See this screenshot for the float test dataset:

thanks for catching that! it's fixed now. another side benefit is that a nice gaussian curve is also shown while the histogram is being downloaded :innocent:

Issue #5498, opened 2021-05-19
Merger Mode cannot deal with multiple trees annotating the same segments


philippotto:
## Context
When enabling merger mode on an existent skeleton tracing, a simple two tree setup à la
![image](https://user-images.githubusercontent.com/2486553/118837586-43c53d80-b8c5-11eb-8131-5f4d99fa9b17.png)

can produce an invalid appearance.

## Expected Behavior
The trees should be interpreted so that all annotated segments are mapped to *one* segment (since both tree have a node within the same segment, an implicit edge between these nodes can be assumed).

## Current Behavior
For each tree, all touched segments get the tree color. Therefore, the merger mode mapping only works correctly if there is not more than one tree per segment. As a known limitation, this might be alright, but right now the behavior is quite confusing. Either we fix it properly or emit a warning.

## Steps to Reproduce the bug
1. Open a hybrid tracing for a dataset with a segmentation layer.
2. Create two trees where both trees have a node in the same segment (see screenshot).
3. Enable merger mode

## Your Environment for bug
chrome/linux

PR #8463, opened 2025-03-21
Fix undo in merger mode

philippotto:
@MichaelBuessemeyer the actual fix is only one commit. the remaining commits only did some renaming to make things easier to understand.

- https://___.webknossos.xyz

### Steps to test:
- create a skeleton with a dataset that has a segmentation
- enable merger mode
- create, delete nodes and use undo

### Issues:
- fixes https://discuss.webknossos.org/t/merger-mode-bug/1911/4

------


MichaelBuessemeyer:
The code looks good to me, but during testing I notice the following undo inconsistencies. Imo we should try to fix these before merging:

https://github.com/user-attachments/assets/6a7124cf-e642-45e4-a300-885c38344c9f
https://github.com/user-attachments/assets/b29eb437-9bd7-4441-8ba2-6ba39bb7b8a9

MichaelBuessemeyer (approved):
Ok the changes look good and the bug I mentioned earlier seems to be triggered only when a segment is part in multiple trees. This is tracked by issue https://github.com/scalableminds/webknossos/issues/5498. 

Thus, this should be fine to merge as it si already an improvement :D

See https://scm.slack.com/archives/C5AKLAV0B/p1742917318478009?thread_ts=1742915131.919449&cid=C5AKLAV0B for additional discussion reference

PR #8466, opened 2025-03-25
In annotation upload, always add skeleton if none is found in NML

### LINKED ISSUE #8143: Download + Reupload drops skeleton layer if it is empty
Looks like no trees in the NML is interpreted as no skeleton layer should exist.

We should probably add a property to NML files to indicate an (empty) skeleton layer

fm3:
We decided that on annotation upload, an empty skeleton layer should always be created, see https://github.com/scalableminds/webknossos/issues/8143#issuecomment-2751144704

- https://uploadalwaysskeleton.webknossos.xyz

### Steps to test:
- Create hybrid, download + reupload, should still contain both layers
- Create volume-only, download + reupload, should contain both layers too
- Note that volume task creation with upload currently fails, fixed in related: https://github.com/scalableminds/webknossos/pull/8468

### Issues:
- fixes #8143 

------


frcroth (approved):
nice!

PR #8474, opened 2025-03-27
Fix ai model list's alignment of items in actions column

MichaelBuessemeyer:
The action items in the ai models list are misaligned if there are multiple. This PR cleans this up. All actions should now have their own row and be left aligned together. Similar to the actions column in the dataset table. See: 
image

Oh and I added some icons as I though it looks nicer this way :D

- https://___.webknossos.xyz

### Steps to test:
- Somehow get an entry into the ai models table
  - test locally & enable jobs (no need to run the worker)
  - create an annotation on a dataset with a volume layer
  - create some bounding boxes
  - start an ai model training job
  - Open the AI model list. The action should now be aligned
  - To add more items to the actions, you could go ahead and edit the code a little: in `ai_model_list_view.tsx` add `<Row>{organizationSharingButton}</Row>` multiple times after line 205. Allowing to check how it would look like when the job was successful.

### Issues:
- follow up of #8418

------


daniel-wer (approved):
Sorry for the wait. Thanks for fixing this :pray:

Issue #3894, opened 2019-03-14
in skeleton tracings, load downsampled segmentation


fm3:
the argument that only segmentation that can also be edited should be loaded does not hold for skeleton-only tracings

philippotto:
Hmm, seems to work in some cases. E.g., https://webknossos.brain.mpg.de/annotations/Explorational/5ee1ffba010000f0131ce406#4941,2651,4096,0,2.200

Not sure what the difference is.

philippotto:
Couldn't reproduce this. And additionally this should also be resolved with the upcoming hybrid + multi-res volume approach.

PR #8476, opened 2025-03-27
Make manual `reserveManualUpload` route return `datasetId` and `directoryName`

MichaelBuessemeyer:
This is a suggestion to help out with vx issue: https://github.com/scalableminds/voxelytics/issues/3894

- https://___.webknossos.xyz

### Steps to test:
- abc

### TODOs:
- [x] Do not allow a dataset id to be passed to `/reserveManualUpload`
- [x] instead allow a parameter like `allowDuplicateDatasetName` to make the request fail in case the organization already has a dataset with that name (and the resulting `directoryName` would be something like `datasetName-datasetId`)
- [x] On success the route should return the new dataset id and directory name.

### Issues:
- related vx issue: [#3894](https://github.com/scalableminds/voxelytics/issues/3894)
- related slack thread: https://scm.slack.com/archives/C5AKLAV0B/p1743087424143489

------
  - Is already on the way: https://github.com/scalableminds/webknossos-libs/pull/1283 


fm3 (commented):
LGTM. Could you explain what use case the requireUniqueName serves? My assumption was that duplicate names are no problem, and the returned directoryName will be used by vx to place the dataset on disk.

MichaelBuessemeyer:
> Could you explain what use case the requireUniqueName serves? My assumption was that duplicate names are no problem, and the returned directoryName will be used by vx to place the dataset on disk.

100'% correct imo. This feature to make the publication task fail in case of a not unique name was requested by @philippotto. Maybe you could elaborate your reasoning for this @philippotto  :thinking:?

philippotto:
> This feature to make the publication task fail in case of a not unique name was requested by @philippotto. Could you please elaborate your reasoning for this @philippotto 🤔 🙏 ?

When we use voxelytics internally for projects, it is good practice to give a new dataset that is about to be published to WK a unique name (so that it can easily be distinguished from other experiments). This is why I voted for being able to have this asserted, even though WK supports non-unique names.
Related to that: personally, I'm not very fond of the idea that the same folder can have two datasets with the same name. I understand that we had to lift this restriction because of cross-orga and multiple-folder use cases. But I think, we went a bit too far... :see_no_evil:

philippotto:
> Related to that: personally, I'm not very fond of the idea that the same folder can have two datasets with the same name. I understand that we had to lift this restriction because of cross-orga and multiple-folder use cases. But I think, we went a bit too far... 🙈

google drive supports non-unique file names in the same folder, too. maybe I'm in the minority my opinion :shrug:

fm3 (approved):
Fair enough. So this should possibly be an option in the PublishTask in vx as well?

MichaelBuessemeyer:
> So this should possibly be an option in the PublishTask in vx as well?

Yeah, I think that's the idea :D

PR #8478, opened 2025-03-31
Credit transactions clean up

MichaelBuessemeyer:
In #8352 I missed enabling all orgas to do credit transactions regardless of whether they have a paid plan or not.

- https://___.webknossos.xyz

### mito inferral should be disabled for non super users:
- I suggest testing locally for simplicity
- Log in as a non super user but an admin.
- Enable jobs
- View a dataset and open the start jobs modal. Mitochondria inferral should be disabled.
- Open the network tab. Start a neuron inferral.
- In the network tab find the neuron inferral request. 
  - either (chrome): right click, copy as fetch, paste into console, in the URL replace `inferNeurons` with `inferMitochondria` and send the request. Check the network tab again. 
  - Or (firefox): right click: Edit & send,  in the URL replace `inferNeurons` with `inferMitochondria` and send the request.
-  The request should have failed with  `"error": "Currently mitochondria inferral is only allowed for super users."`

#### No rapid requesting of credit costs
- Test whether the infinite cost request fix works:
- Open the ai analysis modal, select neuron inferral and select a bounding box. Open the network tab. Check whether the frontend sends frequent cost requests without any input changes from the user. There should be none
- Do the same testing on the current master. There, the frontend should rapidly request costs over and over again without any UI changes by the user.

### Issues:
- follow up fix for: #8352

------


app/models/organization/OrganizationService.scala
```diff
@@ -182,9 +182,6 @@ class OrganizationService @Inject()(organizationDAO: OrganizationDAO,
       _ <- organizationDAO.acceptTermsOfService(organizationId, version, Instant.now)
     } yield ()
 
-  def assertOrganizationHasPaidPlan(organizationId: String): Fox[Unit] =
-    for {
-      organization <- organizationDAO.findOne(organizationId)(GlobalAccessContext)
-      _ <- bool2Fox(PricingPlan.isPaidPlan(organization.pricingPlan)) ?~> "creditTransaction.notPaidPlan"
-    } yield ()
+  def assertOrganizationHasPaidPlan(organizationId: String): Fox[Unit] = Fox.successful(())
+
```
fm3, line 186: Please add a comment explaining why this is not active. Possibly with a link to the issue tracking the eventual activation.

CHANGELOG.unreleased.md
```diff
@@ -32,6 +32,7 @@ For upgrade instructions, please check the [migration guide](MIGRATIONS.released
 - Fixed the alignment of the button that allows restricting floodfill operations to a bounding box. [#8388](https://github.com/scalableminds/webknossos/pull/8388) 
 - Fixed a bug for rotated dataset where volume tools were disabled although the dataset was rendered untransformed. [#8432](https://github.com/scalableminds/webknossos/pull/8432)
 - Fixed rare bug where saving got stuck. [#8409](https://github.com/scalableminds/webknossos/pull/8409)
+- Re-enabled jobs planed to be paid with credits for organizations without a paid plan. [#8478](https://github.com/scalableminds/webknossos/pull/8478)
```
fm3, line 35:
```suggestion
- Re-enabled jobs planned to be paid with credits for organizations without a paid plan. [#8478](https://github.com/scalableminds/webknossos/pull/8478)
```

frontend/javascripts/oxalis/view/action-bar/starting_job_modals.tsx
```diff
@@ -675,13 +675,23 @@ function useCurrentlySelectedBoundingBox(
...
+    const currentSelectedLayer = layers.find((layer) => layer.name === currentlySelectedLayerName);
+    const indexOfLayer = currentSelectedLayer ? layers.indexOf(currentSelectedLayer) : -1;
+    const newCurrentlySelectedBoundingBox = isBoundingBoxConfigurable
+      ? userBoundingBoxes.find((bbox) => bbox.id === selectedBoundingBoxId)
+      : indexOfLayer >= 0
+        ? defaultBBForLayers[indexOfLayer]
+        : undefined;
+    setCurrentlySelectedBoundingBox(newCurrentlySelectedBoundingBox);
+  }, [selectedBoundingBoxId, currentlySelectedLayerName, isBoundingBoxConfigurable]);
+  return currentlySelectedBoundingBox;
```
MichaelBuessemeyer, line 694: This is the fix to avoid infinetly requesting the job costs. This was done due to previously the selected bounding box always being a new object and thus the cache check of whether the costs should be re-requested was always set to true. Now the selected bounding box is correctly stores as a state variable and only updated in case the user edits info in the form.

Some variables used in the effect guarding the re-calculation of the selected bounding box are not included in the dependency list, as they are also newly computed each rendering cycle and thus would always lead to the effect being recalculated

MichaelBuessemeyer: @fm3 do you think you can also review this, or should I ping daniel?

Issue #6962, opened 2023-04-04
Use fossilDB multi-get to optimize performance, e.g. for editable mappings/proofreading


fm3:
The fossilDB should support a get request for a list of keys. This way, grpc overhead could be reduced when requesting many items from fossildb. The same could happen for put and possibly other operations.

PR #8482, opened 2025-03-31
Use FossilDBPutBuffer for more efficient saving of multiple fossil keys

fm3:
Measured duplicate on a volume annotation.
Duplicating 923 volume buckets took 827ms on master, 360ms here (factor 2.2).
Similar speedups are to be expected for the other usages.

- https://putbuffer.webknossos.xyz

### Steps to test:
- Edit a skeleton annotation with multiple trees, reload, should work
- Duplicate annotation, should work (skeleton, volume, editable mapping)
- Merge annotations, should work (skeleton, volume, editable mapping)

### Issues:
- contributes to #6962 

------


fm3:
Thanks for your review! You’re right, this might be a way to speed this up further but as you said, I’d rather avoid the complexity of having to ensure that everything is working correctly in parallel and correctly waiting for everything on the final flush.

Also, since we are serving many users in parallel, maximum parallelization could actually also be a downside. While it would serve to answer individual requests more quickly, it would potentially also use up many threads, CPU, and RPC requests, so that some (possibly smaller) requests by other users might get answered considerably later. This is also why I mostly still use serialCombined, to limit parallelization inside of individual user requests. An alternative would be to use custom thread pools for fine-grained control, but that would add another layer of complexity. I hope this also answers https://github.com/scalableminds/webknossos/pull/8469#issuecomment-2778454127

Issue #8431, opened 2025-03-06
Add Flip in dataset rotation


normanrz:
## Context

* [ ] Specific to long-running jobs (set `jobsEnabled=true` in `application.conf`)
* [ ] Specific to webknossos.org (set `isDemoInstance=true` in `application.conf`)

PR #8485, opened 2025-03-31
Allow to mirror datasets

MichaelBuessemeyer:
This is a follow up for #8159. The rotation feature now allows to mirror datasets along axes as well.

Moreover, the rotation settings no longer save a transformation in case it is "nulled" for every setting resulting in an identity transform.

- https://___.webknossos.xyz

### Steps to test:
- Open your favourite dataset in a tab in e.g. view mode.
- duplicate that tab
- Open the dataset settings in one of the tabs and configure one axis to be mirrored. then view the dataset again
- Go to a specific slice and compare the result with the same position viewed in the tab that does not have the newly configured mirrored axis.
- The data should be rendered mirrored just like configured. Compare some more dataset positions and confirm that this is correct.

### Issues:
- fixes [#8431](https://github.com/scalableminds/webknossos/issues/8431)

------


frontend/javascripts/oxalis/model/accessors/dataset_layer_transformation_accessor.ts
```diff
@@ -364,15 +381,15 @@ function isTranslationOnly(transformation?: AffineTransformation) {
...
+function isRotationAndMirrorMaybeOnly(transformation?: AffineTransformation) {
   if (!transformation) {
     return false;
   }
   const threeMatrix = new THREE.Matrix4()
     .fromArray(nestedToFlatMatrix(transformation.matrix))
     .transpose();
   threeMatrix.decompose(translation, quaternion, scale);
-  return translation.length() === 0 && scale.equals(NON_SCALED_VECTOR);
+  return translation.length() === 0 && scale.length() === NON_SCALED_VECTOR.length();
```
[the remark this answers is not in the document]
MichaelBuessemeyer: Nope this check is intended to allow mirroring. The function name includes "MirrorMaybe" meaning that mirroring should be allowed

frontend/javascripts/dashboard/dataset/dataset_rotation_form_item.tsx
```diff
@@ -80,9 +87,14 @@ export const AxisRotationFormItem: React.FC<AxisRotationFormItemProps> = ({
...
-
-      const rotationInRadians = rotationInDegrees * (Math.PI / 180);
-      const rotationMatrix = getRotationMatrixAroundAxis(axis, rotationInRadians);
+      const rotationValues: RotationSetting = form.getFieldValue(["datasetRotation"])[axis];
+      if (rotationInDegrees) {
+        rotationValues.rotationInDegrees = rotationInDegrees;
+      }
+      if (isMirrored) {
+        rotationValues.isMirrored = isMirrored;
+      }
```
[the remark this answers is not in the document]
MichaelBuessemeyer: oh good catch!

frontend/javascripts/dashboard/dataset/dataset_rotation_form_item.tsx
```diff
@@ -80,9 +87,14 @@ export const AxisRotationFormItem: React.FC<AxisRotationFormItemProps> = ({
...
-      const rotationInRadians = rotationInDegrees * (Math.PI / 180);
-      const rotationMatrix = getRotationMatrixAroundAxis(axis, rotationInRadians);
+      const rotationValues: RotationSetting = form.getFieldValue(["datasetRotation"])[axis];
+      if (rotationInDegrees) {
+        rotationValues.rotationInDegrees = rotationInDegrees;
+      }
+      if (isMirrored !== undefined) {
+        rotationValues.isMirrored = isMirrored;
+      }
+      const rotationMatrix = getRotationMatrixAroundAxis(axis, rotationValues);
```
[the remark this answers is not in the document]
MichaelBuessemeyer: neat thanks for also mentioning this :pray:

frontend/javascripts/oxalis/model/accessors/dataset_layer_transformation_accessor.ts
```diff
@@ -364,15 +381,15 @@ function isTranslationOnly(transformation?: AffineTransformation) {
   return scale.equals(NON_SCALED_VECTOR) && quaternion.equals(IDENTITY_QUATERNION);
 }
 
-function isRotationOnly(transformation?: AffineTransformation) {
+function isRotationAndMirrorMaybeOnly(transformation?: AffineTransformation) {
```
philippotto, line 384: how about: `isOnlyRotatedOrMirrored`? I think "or" makes more sense and the "maybe" part can be omitted?

frontend/javascripts/oxalis/model/accessors/dataset_layer_transformation_accessor.ts
philippotto, line 392: hmm, similarily to the AI, I'm also wondering whether this is the best approach. I think, I would have preferred to take Math.abs of the vector and then check that it equals 1,1,1.
currently, something like (sqrt(3),0,0) could result in the same length as (1,1,1), couldn't it? would be improbable, but still not so clean.

frontend/javascripts/oxalis/model/accessors/dataset_layer_transformation_accessor.ts
```diff
@@ -443,6 +460,28 @@ function _doAllLayersHaveTheSameRotation(dataLayers: Array<APIDataLayer>): boole
 
 export const doAllLayersHaveTheSameRotation = _.memoize(_doAllLayersHaveTheSameRotation);
 
+export function transformationEqualsAffineIdentityTransform(
+  transformations: CoordinateTransformation[],
+): boolean {
+  const hasValidTransformationCount = transformations.length;
```
philippotto, line 466:
```suggestion
  const hasValidTransformationCount = transformations.length > 0;
```

reads a bit nicer imo

frontend/javascripts/oxalis/model/accessors/dataset_layer_transformation_accessor.ts
```diff
@@ -443,6 +460,28 @@ function _doAllLayersHaveTheSameRotation(dataLayers: Array<APIDataLayer>): boole
...
+  const hasOnlyAffineTransformations = transformations.every(
+    (transformation) => transformation.type === "affine",
+  );
+  // There should be a translation to the origin, one transformation for each axis and one translation back. => A total of 5 affine transformations.
```
philippotto, line 470: copy paste error?

frontend/javascripts/dashboard/dataset/dataset_rotation_form_item.tsx
```diff
@@ -80,9 +87,14 @@ export const AxisRotationFormItem: React.FC<AxisRotationFormItemProps> = ({
...
       }
-
-      const rotationInRadians = rotationInDegrees * (Math.PI / 180);
-      const rotationMatrix = getRotationMatrixAroundAxis(axis, rotationInRadians);
+      const rotationValues: RotationSetting = form.getFieldValue(["datasetRotation"])[axis];
+      if (rotationInDegrees !== undefined) {
+        rotationValues.rotationInDegrees = rotationInDegrees;
+      }
+      if (isMirrored !== undefined) {
+        rotationValues.isMirrored = isMirrored;
```
philippotto, line 95: this directly mutates the return value from `form.getFieldValue`. is this how it should be done with antd forms?

frontend/javascripts/dashboard/dataset/dataset_settings_view.tsx
```diff
@@ -204,22 +204,28 @@ class DatasetSettingsView extends React.PureComponent<PropsWithFormAndRouter, St
...
-          };
+          const nulledSetting = { rotationInDegrees: 0, isMirrored: false };
+          initialDatasetRotationSettings = { x: nulledSetting, y: nulledSetting, z: nulledSetting };
         } else {
           initialDatasetRotationSettings = {
             // First transformation is a translation to the coordinate system origin.
-            x: getRotationFromTransformationIn90DegreeSteps(firstLayerTransformations[1], "x"),
-            y: getRotationFromTransformationIn90DegreeSteps(firstLayerTransformations[2], "y"),
-            z: getRotationFromTransformationIn90DegreeSteps(firstLayerTransformations[3], "z"),
+            x: getRotationSettingsFromTransformationIn90DegreeSteps(
```
philippotto, line 217: "rotation settings" now means that this also includes mirroring? it's not super clear but I can't come up with something better.

philippotto (commented):
code looks good overall :+1: only left a few comments :)

MichaelBuessemeyer:
@philippotto

philippotto (approved):
sorry, this slipped through! code looks good and works well, too :)

Issue #7839, opened 2024-05-28
Unified Annotation Versioning


philippotto:
Currently, an annotation can have multiple "tracings" (skeleton, volume, editable mapping). All of these have an isolated version history. The UI also exposes these separately, which can be very confusing to deal with from a user perspective. While we could adapt the UI so that only **one** virtual version history is presented, it might make sense to change the underlying structures. We should draft a design doc and discuss how to proceed.

https://www.notion.so/scalableminds/Design-Doc-Unified-Annotation-Versioning-d43efa620adc4a6080a15e689a82b907

PR #8486, opened 2025-04-01
Fix redundant newer-version-on-server toast

philippotto:
Since #7839, there is no need to check for a newer version per tracing.

- https://___.webknossos.xyz

### Steps to test:
- create an annotation, make it public
- open it again in an incognito window
- mutate the first tab and save
- wait for 1 minute (that's the polling interval)
- one toast should appear telling the user that a newer version is available on the server

### Issues:
- follow-up to #7839 

------

### DIFF
--- a/frontend/javascripts/oxalis/model/sagas/save_saga.ts
+++ b/frontend/javascripts/oxalis/model/sagas/save_saga.ts
@@ -499,39 +499,37 @@ function* watchForSaveConflicts(): Saga<never> {
       annotationId,
     );
 
-    for (const tracing of tracings) {
-      // Read the tracing version again from the store, since the
-      // old reference to tracing might be outdated now due to the
-      // immutability.
-      const versionOnClient = yield* select((state) => {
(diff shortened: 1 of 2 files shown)


frontend/javascripts/oxalis/model/sagas/save_saga.ts
```diff
@@ -499,39 +499,37 @@ function* watchForSaveConflicts(): Saga<never> {
...
-      });
+    // Read the tracing version again from the store, since the
+    // old reference to tracing might be outdated now due to the
+    // immutability.
+    const versionOnClient = yield* select((state) => {
+      return state.tracing.version;
+    });
 
-      const toastKey = `save_conflicts_warning_${tracing.tracingId}`;
-      if (versionOnServer > versionOnClient) {
```
philippotto, line 511: best to review while ignoring whitespace.

MichaelBuessemeyer (approved):
Awesome, thanks for fixing :pray:

PR #8488, opened 2025-04-01
If a split action doesn’t actually split the agglomerate, skip no-op steps

### LINKED ISSUE #8465: In EditableMapping strucutres, skip incrementing largestAgglomerateId if a splitAction doesn’t split anything
in applySplitAction in EditableMappingUpdater:

If there is still another edge holding the agglomerate together, graph2 becomes empty. in this case, we should not store graph2, and not increase largestAgglomerateId.

fm3:
Split actions in our editable mappings remove one edge in an agglomerate. If that edge isn’t the last edge holding the two sides of an agglomerate together, everything remains in one agglomerate. In this case, we can skip incrementing the largestExistingAgglomerateId and skip writing the second (empty) graph to the buffer/db.

This will give a slight performance increase, but more importantly, the code more closely resembles the mental model.

- https://proofreadingpartialsplit.webknossos.xyz

### Steps to test:
- Test splitting/merging agglomerates with proofreading tool, should still work
- Note that when splitting in the frontend, a min-cut will be performed, creating split edges for all the edges holding the two sides together.

### Issues:
- fixes #8465

------


philippotto (approved):
:+1:

Issue #8308, opened 2025-01-08
Clean up (front-end) types


philippotto:
For example:
- [x] `StoreState.tracing` is a `HybridTracing`, but actually this should be an annotation. #8490
- [x] Rename `admin_rest_api` to `rest_api` #8587
- [ ] Rename  `oxalis` to `webknossos`
  - [x] `OxalisState` to `WebknossosState` #8591
  - [x] `OxalisModel` to `WebknossosModel` #8600
  - [x] `OxalisController` to `WebknossosController` #8600
  - [ ] `frontend/javascripts/oxalis` to `frontend/javascripts/webknossos` #8601
- [x] Rename `api_flow_types.ts` to `api_types.ts` #8583

hotzenklotz:
@philippotto Is there anything else that should be cleaned up? Or are we satisfied with all the recent changes and this issue can be considered closed?

philippotto:
I think my main gripes are solved. for everything else, I can open a new issue once I've got something :) thank you for your help on this issue!

PR #8490, opened 2025-04-01
Rename store.tracing to store.annotation in frontend

philippotto:
This PR only renames store.tracing to store.annotation, since this fits our terminology better (a tracing is only a layer of an annotation).
No need to view the diff in my opinion, as it's only renaming.

- https://___.webknossos.xyz

### Steps to test:
- CI should be enough thanks to TS and tests

### Issues:
- contributes to #8308 

------
Did not update changelog because this doesn't affect the user.


hotzenklotz (approved):
Such much renaming. Nice! Green CI ✅ is good to see :-)

Now onwards to the merge hell for everyone 👿 😆

philippotto:
> Now onwards to the merge hell for everyone 👿 😆

yeah, sorry about that! the band-aid had to come off at some point (it's overdue for several years now, I think :see_no_evil:).

PR #8493, opened 2025-04-02
Transform meshes and bboxes according to transformation settings

### LINKED ISSUE #8340: User bounding boxes, meshes don't rotate according to transformations
## Context

WEBKNOSSOS support layer transformations. However, in case the active volume layer is transformed, the meshes do not transform with it. For consistency, this should be the case. Moreover, the user bounding boxes do not adapt to the currently natively rendered layers transformations like the skeleton layer. 

This came up in PR: https://github.com/scalableminds/webknossos/pull/8159#issuecomment-2601949466

## Expected Behavior

When transformations are configured
- Meshes should rotate together with their respective segmentation layer
- User bounding boxes should rotate according to the currently natively rendered layer just like the skeleton layer.

## Current Behavior

The transform settings have no effect to them.

## Steps to Reproduce the bug

1. Configure dataset rotation for a dataset (possible after https://github.com/scalableminds/webknossos/pull/8159 is merged)
2. Create an annotation for the dataset and load meshes and create bounding boxes
3. Toggle the layer transforms and see that the meshes and bounding boxes do not rotate

MichaelBuessemeyer:
- https://___.webknossos.xyz

### Steps to test:
- Configure a rotation /mirroring for a dataset. The dataset should have at least 2 segmentation layers (also achivable via volume annotation layer without fallback)
- open an annotation of the dataset and render some meshes for both segmentation layers. Only the meshes for the active segmentation layer should be visible
- add some user bounding boxes
- Toggle the dataset transformation settings. The bboxes and visible meshes should transform accoridngly.

### TODOs:
- [x] meshes position strangely when rendered with the new rotation feature -> this is very likely due to scale.
  - The segment mesh controller sets the scale of the outer wrapping group to the dataset scale / (scale defined in the meshes file I think).
  - Therefore transformation likely needs to be done on a different level and not the meshes root group. Likely the inner meshes group is a good choice. Thus, this needs to be applied to all meshes groups.  For this I already coded: `forEveryMesh` to easily manipulate all meshes. 
  - This should be used upon changing the transformation view settings and also be set when loading a new mesh
  - As the later is changed in #8106 I would wait for this to be merged so this does not need to be implemented in both the version before and after #8106 is merged. 

### Issues:
- fixes #8340

------


MichaelBuessemeyer:
blocked by #8106 to creating headache merge conflicts

frontend/javascripts/oxalis/controller/segment_mesh_controller.ts
```diff
@@ -196,14 +196,24 @@ export default class SegmentMeshController {
...
     const group = new THREE.Group() as SceneGroupForMeshes;
-
+    const dsScaleFactor = Store.getState().dataset.dataSource.scale.factor;
+    // If the mesh was calculated on a different magnification level,
+    // the backend sends the scale factor of this magnification.
+    // As the meshesLODRootGroup is already scaled by the main rootGroup,
+    // this portion of the scale needs to be take out of the scale applied to the mesh.
+    // If no scale was given, the meshes coordinates are already in scale of dataset and
+    // thus the scaling done by the root group needs to be inverted.
+    scale = scale || [1, 1, 1];
```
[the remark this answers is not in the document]
MichaelBuessemeyer: But on second thought, this would require to change the mesh file descriptor entries which is very likely not what we want, correct @fm3?

frontend/javascripts/oxalis/controller/scene_controller.ts
```diff
@@ -110,8 +113,8 @@ class SceneController {
       new THREE.Vector3(...Store.getState().dataset.dataSource.scale.factor),
     );
     // Add scene to the group, all Geometries are then added to group
```
philippotto, line 115: I feel like this comment could be updated (I think, it was already wrong before; shouldn't it have been "add rootGroup to scene"?). zooming out a bit: this whole imperative `group.add(node)` interface of threejs is not really ideal for understanding the scene graph hierarchy in code, in my opinion.

chatgpt suggested using a helper like this (I tuned it a bit):
```js
function Group(name, children = []) {
  const group = new THREE.Group();
  group.name = name;
  children.forEach(child => group.add(child));
  return group;
}

const scene = new THREE.Scene();
scene.add(
  Group("rootGroup", [
    someMesh,
    Group("meshGroup", [
      someOtherMesh
    ])
  ])
);

this.rootGroup = this.scene.getObjectByName("rootGroup")
...
```

untested. if you feel like it, you could try if integrating something like this is easy and helpful. but it's not a blocker! just thought, I leave my ideas here :)

frontend/javascripts/oxalis/controller/segment_mesh_controller.ts
```diff
@@ -196,14 +196,24 @@ export default class SegmentMeshController {
...
       }
       targetGroup.segmentId = segmentId;
-      if (scale != null) {
-        targetGroup.scale.copy(new THREE.Vector3(...scale));
-      }
+      const dsScaleFactor = Store.getState().dataset.dataSource.scale.factor;
+      // If the mesh was calculated on a different magnification level,
+      // the backend sends the scale factor of this magnification.
+      // As the meshesLODRootGroup is already scaled by the main rootGroup,
+      // this portion of the scale needs to be take out of the scale applied to the mesh.
```
philippotto, line 203:
```suggestion
      // this portion of the scale needs to be taken out of the scale applied to the mesh.
```

frontend/javascripts/oxalis/model/accessors/tool_accessor.ts
```diff
@@ -155,6 +154,27 @@ function _getSkeletonToolInfo(
...
 const getSkeletonToolInfo = memoizeOne(_getSkeletonToolInfo);
 
+function _getBoundingBoxToolInfo(hasSkeleton: boolean, isSkeletonLayerTransformed: boolean) {
+  if (isSkeletonLayerTransformed) {
+    return {
+      [AnnotationToolEnum.BOUNDING_BOX]: {
+        isDisabled: true,
+        explanation: hasSkeleton
+          ? "The bounding box tool is disabled because the bounding boxes are rendered transformed like the skeleton layer. Use the left sidebar to render them without transformations by letting the skeleton layer render without any transformations."
+          : "The bounding box tool is disabled because the bounding boxes are rendered transformed.",
```
philippotto, line 164: when is `isSkeletonLayerTransformed` true and `hasSkeleton` false? seems counter intuitive that this scenario can happen? maybe `isSkeletonLayerTransformed` needs to be renamed to `areGeometriesTransformed` or sth like this?

frontend/javascripts/oxalis/model/accessors/tool_accessor.ts
```diff
@@ -155,6 +154,27 @@ function _getSkeletonToolInfo(
 }
 const getSkeletonToolInfo = memoizeOne(_getSkeletonToolInfo);
 
+function _getBoundingBoxToolInfo(hasSkeleton: boolean, isSkeletonLayerTransformed: boolean) {
+  if (isSkeletonLayerTransformed) {
+    return {
+      [AnnotationToolEnum.BOUNDING_BOX]: {
+        isDisabled: true,
+        explanation: hasSkeleton
+          ? "The bounding box tool is disabled because the bounding boxes are rendered transformed like the skeleton layer. Use the left sidebar to render them without transformations by letting the skeleton layer render without any transformations."
```
philippotto, line 163: I tried to improve the text a bit. maybe like this?

```suggestion
          ? "The bounding box tool is disabled because the bounding boxes are currently transformed according to the skeleton layer. To use the tool, ensure that the skeleton layer is rendered natively in the left sidebar."
```

frontend/javascripts/oxalis/model/accessors/tool_accessor.ts
philippotto, line 164:
```suggestion
          : "The bounding box tool is disabled because the bounding boxes are rendered with transforms.",
```

philippotto (commented):
nice! thanks for making this possible :) left mostly cosmetic feedback.

during review, I noticed another potential problem. bounding boxes can currently be stored for skeleton and volume layers. this is a known "code smell" which bites us every now and then. the current semantics are that there is one set of bboxes and they are stored in "whatever layer is available". in the typical hybrid annotation, they would be stored in the skeleton layer. in a volume-only annotation, the bboxes are stored in the volume layer. does the current PR use the transforms of that layer in that case?

afaik it should never :tm: (cc @fm3) happen that some bboxes are stored in the skeleton layer and some bboxes are stored in a volume layer. so, we should be fine with giving all bboxes the transforms of one layer (deleting a layer might be weird, because the transforms could change then? but maybe this is an okay side-effect).

frontend/javascripts/oxalis/controller/scene_controller.ts
MichaelBuessemeyer: Ok I tried. But in the second commit I got rid of the `Group` function as the normal `add` funciton of threejs already supports multiple arguments at once making the helper function obsolete imo.

Please check out my "clean up commits" d0f26081f5d925a2ac1ef3f235483ba0f443b272 and 00cfb87effeb5aee1fa4f6b8b8ce64fb7587831e. I am unsure whether they are much of an improvement :thinking:

MichaelBuessemeyer:
> during review, I noticed another potential problem. bounding boxes can currently be stored for skeleton and volume layers. this is a known "code smell" which bites us every now and then. the current semantics are that there is one set of bboxes and they are stored in "whatever layer is available". in the typical hybrid annotation, they would be stored in the skeleton layer. in a volume-only annotation, the bboxes are stored in the volume layer. does the current PR use the transforms of that layer in that case?

Oh that's actually quite the problem you noticed there :/. I thought of applying the same to bboxes as to the potential skeleton would keep things consistent as the skeleton layer just rotates according to what is currently rendered natively. Thus, doing the same for bboxes independent of whether the data is stored on a skeleton or volume annotation also sounds consistent to me.

So regarding:
> does the current PR use the transforms of that layer in that case?

That's a no. It always renders like the skeletons would. Even if there are none.

But what's more of a problem now is, that the meshes also rotate as if they were the rendered trees. but the should be transformed like the segmentation layer they belong to. And there can be meshes from different segmentation layers active at the same time, if I am not wrong. So, they should be transformed exactly like their segmentation layers would, but the scene hierarchy does not support retrieving the segmentation layer and render meshes differently depending on that or does it? 

I see 2 possible solutions for this:
1. rotate all meshes according to the transforms of the currently active segmentation layer
2. refactor `SegmentMeshController` to grouping meshes by their layer and set the transformation according to how the layer would be currently rendered

MichaelBuessemeyer (commented):
Thanks for the feedback and mentioning the thought about how bboxes should be rendered. It opened up a new question for me. See above.

Moreover, I gave your refactoring suggestion a try. Not sure how much improvement that is. Happy to hear you comments on that :D

philippotto:
>  It is actually bad to always transform the bounding boxes like the skeleton would be transformed as the attachment to a layer is not even visible to the user? Or is it bad and we should transform them like the associated layer?

the semantical correct approach would be that bounding boxes should always get the transformation of the layer that is returned by `getSomeTracing`. I hope that this is easy to achieve :pray: 

> 1. rotate all meshes according to the transforms of the currently active segmentation layer

yes, I think, this would be a good approach (probably simpler than your second suggestion). meshes from invisible segmentation layers could then be made invisible. this would be nice in general, I think (skeletons are also hidden in the 3D viewport when the skeleton layer is disabled <-- so, it would be similar for meshes then).

frontend/javascripts/oxalis/controller/scene_controller.ts
```diff
@@ -244,10 +225,25 @@ class SceneController {
...
+      this.layerBoundingBoxGroup,
+      this.annotationToolsGeometryGroup.add(
+        ...this.contour.getMeshes(),
+        this.quickSelectGeometry.getMeshGroup(),
+        ...this.lineMeasurementGeometry.getMeshes(),
+        ...this.areaMeasurementGeometry.getMeshes(),
+      ),
+      ...this.datasetBoundingBox.getMeshes(),
+      ...planeMeshes,
+    );
```
philippotto, line 240: nice, I like this :+1:

MichaelBuessemeyer:
> the semantical correct approach would be that bounding boxes should always get the transformation of the layer that is returned by getSomeTracing. I hope that this is easy to achieve 🙏

> yes, I think, this would be a good approach (probably simpler than your second suggestion). meshes from invisible segmentation layers could then be made invisible. this would be nice in general, I think (skeletons are also hidden in the 3D viewport when the skeleton layer is disabled <-- so, it would be similar for meshes then).


Thanks for clarifying :pray:. As far as I can remember, I incorporated all this yesterday, so please give this another review round :eyes: :book: :pencil2:

frontend/javascripts/oxalis/controller/scene_controller.ts
```diff
@@ -430,23 +435,42 @@ class SceneController {
     this.rootNode.add(this.userBoundingBoxGroup);
   }
 
+  private applyTransformToGroup(transform: Transform, group: THREE.Group | CustomLOD) {
+    if (transform?.affineMatrix) {
```
philippotto, line 439: why the `?`? isn't `Transform` always defined? and if it can be null, shouldn't `group.matrix` be set to identity or something like that?

frontend/javascripts/oxalis/controller/scene_controller.ts
```diff
@@ -430,23 +435,42 @@ class SceneController {
...
-      const matrix = new THREE.Matrix4();
-      // @ts-ignore
-      matrix.set(...transformMatrix);
-      // We need to disable matrixAutoUpdate as otherwise the update to the matrix will be lost.
-      this.userBoundingBoxGroup.matrixAutoUpdate = false;
-      this.userBoundingBoxGroup.matrix = matrix;
-      this.segmentMeshController.meshesLODRootGroup.matrixAutoUpdate = false;
-      this.segmentMeshController.meshesLODRootGroup.matrix = matrix;
-    }
+    this.applyTransformToGroup(transformForMeshes, this.segmentMeshController.meshesLODRootGroup);
```
philippotto, line 473: as we discussed [here](https://github.com/scalableminds/webknossos/pull/8493#issuecomment-2801580769), this will transform meshes of all layers with the transforms of the visible segmentation layer :+1: I think, this is fine if meshes of invisible layers also get hidden automatically (which isn't done in master). does this happen in this pr? I think, we need this for correct mesh rendering.
maybe "refactor SegmentMeshController to grouping meshes by their layer" (quote by you) is necessary for doing the invisible stuff, too? so, one could also do the transform per mesh-layer with not a lot of extra work, but I'm fine with either way :see_no_evil: 
what do you think?

frontend/javascripts/oxalis/controller/segment_mesh_controller.ts
```diff
@@ -80,7 +90,8 @@ export default class SegmentMeshController {
   > = {};
 
   constructor() {
-    this.meshesLODRootGroup = new CustomLOD();
+    this.lightsGroup = new THREE.Group();
+    this.meshesLayerLODRootGroup = new THREE.Group();
     this.addLights();
   }
```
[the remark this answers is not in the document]
MichaelBuessemeyer: thanks for noticing. should be fixed now :pray:

frontend/javascripts/oxalis/controller/scene_controller.ts
MichaelBuessemeyer: Ok sorry, I overlooked the part about making meshes of invisible layers invisible as well. The scene hierarchy should now be adjusted to make this relativly easy. The structure is now: Each layer has its own lod group with the respective meshes and the lod group for each layer can be made invisible on demand

frontend/javascripts/oxalis/controller/scene_controller.ts
```diff
@@ -634,6 +656,20 @@ class SceneController {
...
+          const visibleSegmentationLayerNames = getVisibleSegmentationLayers(storeState).map(
+            (l) => l.name,
+          );
+          if (_.isEqual(this.visibleSegmentationLayerNames, visibleSegmentationLayerNames)) {
+            return this.visibleSegmentationLayerNames;
+          } else {
+            this.visibleSegmentationLayerNames = visibleSegmentationLayerNames;
+            return visibleSegmentationLayerNames;
+          }
+        },
```
philippotto, line 670: You can use `reuseInstanceOnEquality` for this if I'm not mistaken. It would look something like this:

```ts
// top level
const getVisibleSegmentationLayerNames = reuseInstanceOnEquality(
          (state) => getVisibleSegmentationLayers(storeState).map(
            (l) => l.name,
          )
)

// in listenToStoreProperty
return getVisibleSegmentationLayerNames(storeState)
```


the instance variable `this.visibleSegmentationLayerNames` can be removed then, too.

philippotto:
The code looks good :+1: however, I cannot load precomputed meshes in view mode. I get:

```
console.js:13 Mesh chunks for segment 5231141 couldn't be loaded due to TypeError: Cannot read properties of null (reading 'meshesLayerLODRootGroup')
    at getLODGroupOfLayer (segment_mesh_controller.ts:308:17)
    at Se (redux-saga-core.esm.js:523:21)
    at En (redux-saga-core.esm.js:1203:7)
    at Rn (redux-saga-core.esm.js:1270:5)
    at gn (redux-saga-core.esm.js:1160:9)
    at en (redux-saga-core.esm.js:1250:7) 
```


can you have a look? :pray:

frontend/javascripts/oxalis/controller/scene_controller.ts
MichaelBuessemeyer: Very awesome thanks :tada:

MichaelBuessemeyer:
@philippotto  could you please give this another look? The bug should be fixed now. It was due to the context not being bonded in two `yield calls` in the mesh saga.

frontend/javascripts/oxalis/controller/scene_controller.ts
```diff
@@ -485,6 +488,7 @@ class SceneController {
 
   updateMeshesAccordingToLayerVisibility(): void {
     const state = Store.getState();
+    console.log("updating according to visible layers");
```
philippotto, line 491: remove

philippotto (approved):
great! only remove the one console.log before merging please :)

frontend/javascripts/oxalis/controller/scene_controller.ts
MichaelBuessemeyer: oh sorry, I used that to confirm no unnecessary recomputation

PR #8494, opened 2025-04-02
Avoid white flash on page reload in dark theme

### LINKED ISSUE #8375: Flash of white when (re)loading the app in dark theme
Using Firefox on Windows, I tried using dark mode and found that each page refresh (ctrl+r) would lead to a split second of the whole page being white before dark content is rendered again. Other pages like github don’t seem to have this problem. Can it be circumvented?

Just guessing, maybe only the body, but not the html element, has the dark background property in css?

fm3:
Our dark theme is rendered dynamically with javascript.
That means that before javascript is initialized, there is a flash of white. This happens on every page reload.
Adding a pure css rule in the main html prevents that.

- https://darkthemeavoidflash.webknossos.xyz

### Steps to test:
- In your browser, select dark theme
- hit some page reloads on wk
- Should not see flash of white

### Issues:
- fixes #8375


knollengewaechs (approved):
the code looks good and testing in a chromium-based browser and firefox was also great :blush: thanks for fixing this! :pray:

PR #8498, opened 2025-04-03
Restructure list annotations query to guide Postgres to a more efficient query plan

fm3:
It appears the new postgres version doesn’t find an efficient query plan anymore for our old WITH query. This PR restructures the query to get the LIMIT in early.

### Steps to test:
- View annotation list, should work
- Share some annotations with multiple teams
- Should be still listed correctly in annotation list view (no duplicates)
- I already tested with a postgres client that the new query is no longer slow on the large production database.

### Issues:
- fixes https://discuss.webknossos.org/t/annotations-tab-loads-forever/1915

------


MichaelBuessemeyer (approved):
Thanks for fixing this :tada:

